PR #1995, opened 2015-05-26
Only skip metadata checkpoint if we really skipped it, not force.

sueloverso:
WT-1944
@michaelcahill I have put the flag back in to only skip idle checkpoints if it was actually skipped, not just based on the modified flag.  Your change in changeset ff56ae5dd686c is causing the second issue in WT-1944.  That change is too broad.  The symptom is that repeated runs of recovery are not forcing a metadata update even though the recovery code uses "force=1".  The change you put in loses the "force" setting and we never updated the metadata, and its checkpoint LSN never advanced.  Please review again.  If you don't like this approach we can discuss alternatives.

### DIFF
--- a/src/include/btree.h
+++ b/src/include/btree.h
@@ -148,8 +148,9 @@ struct __wt_btree {
 #define	WT_BTREE_IN_MEMORY	0x00200	/* Cache-resident object */
 #define	WT_BTREE_NO_EVICTION	0x00400	/* Disable eviction */
 #define	WT_BTREE_SALVAGE	0x00800	/* Handle is for salvage */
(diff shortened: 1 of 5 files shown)


michaelcahill:
Hi @sueloverso, I'm going to merge this to fix the bug.  That said:

1. we're planning to replace the forced log sync with a background sync that we hope will complete concurrent with the checkpoint; and
2. I'm still not a fan of a flag that means "checkpoint skipped this tree".  I don't see a simple alternative, short of parsing out the `"force"` key in `__wt_txn_checkpoint`.

PR #1996, opened 2015-05-27
Calculate a merge generation for bulk loaded LSM chunks.

agorrod:
Otherwise LSM tends to choose poor merges after a bulk load,
which leads to degraded performance.

Refs WT-1947

### DIFF
--- a/src/lsm/lsm_cursor_bulk.c
+++ b/src/lsm/lsm_cursor_bulk.c
@@ -17,11 +17,14 @@ __clsm_close_bulk(WT_CURSOR *cursor)
 {
 	WT_CURSOR_LSM *clsm;
 	WT_CURSOR *bulk_cursor;
+	WT_LSM_CHUNK *chunk;
 	WT_LSM_TREE *lsm_tree;
 	WT_SESSION_IMPL *session;
+	uint64_t avg_chunks, last_size;
 
 	clsm = (WT_CURSOR_LSM *)cursor;
 	lsm_tree = clsm->lsm_tree;
+	chunk = lsm_tree->chunk[0];
 	session = (WT_SESSION_IMPL *)clsm->iface.session;
 
 	/* Close the bulk cursor to ensure the chunk is written to disk. */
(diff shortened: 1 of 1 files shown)


agorrod:
In my local testing this change improves the throughput of the wtperf ```medium-multi-lsm``` workload from:

| Version | Reads | Updates |
|---|---|---|
| branch | 311,000 | 248,000 |
| branch | 262,000 | 246,000 |
| develop | 255,866 | 199,011 |
| develop | 270,000 | 204,000 |

It's a definite improvement - not sure if it's the whole story.

agorrod:
@sueloverso Could you review this one for me please?

src/lsm/lsm_cursor_bulk.c
```diff
@@ -31,7 +34,16 @@ __clsm_close_bulk(WT_CURSOR *cursor)
...
+	F_SET(chunk, WT_LSM_CHUNK_ONDISK);
+	/*
+	 * Setup a generation in our chunk, so that future LSM merges choose
+	 * reasonable sets of chunks.
+	 */
+	avg_chunks = (lsm_tree->merge_min + lsm_tree->merge_max) / 2;
+	for (last_size = avg_chunks * lsm_tree->chunk_size;
+	    last_size < chunk->size;
+	    last_size *= avg_chunks, ++chunk->generation) {}
+
```
michaelcahill: The generation that you are calculating here looks right to me.

I think this code might be a little clearer written like this:

```
avg_chunks = (lsm_tree->merge_min + lsm_tree->merge_max) / 2;
for (total_chunks = chunk->size / lsm_tree->chunk_size;
    total_chunks > 1;
    total_chunks /= avg_chunks)
        ++chunk->generation;
```

agorrod: I guess we think about this in opposite ways: I was thinking "how big would I expect a chunk of generation X to be", your code reads "how many chunk_size pieces fit into a chunk of generation X".

Either way is fine by me.

sueloverso: I find @michaelcahill's suggestion more readable.  I will make that change, test and merge.

PR #1997, opened 2015-05-27
Clear all transaction flags on release, not a subset.

agorrod:
Release is the last thing a transaction does - so all state should
be cleared upon release.


michaelcahill:
fbm, thanks @agorrod.  Merging...

PR #1998, opened 2015-05-27
We were checking the wrong flags value for WT_SKIP_UPDATE_RESTORE

keithbostic:


michaelcahill:
Ouch!  Thanks @keithbostic: I'll merge.

PR #1999, opened 2015-05-27
Evict clean

keithbostic:
@agorrod, @michaelcahill: if the cache is large enough but there are too many dirty pages, don't evict the pages, clean them and leave them in the cache; ref WT-1350.

The real changes are in cff63f7 and ac7ed39.

I spent some time trying to do this work without acquiring an exclusive lock on the page, but eventually gave up on that approach for this (far simpler) solution. There is an advantage to doing it this way, too, we can select which pages we want to keep (I'm still evicting pages that split).

It's only lightly tested, but you both know eviction better than I do, I wanted to give you a chance to point out if there's a reason this approach won't work.


src/include/cache.i
```diff
@@ -91,6 +91,48 @@ __wt_cache_dirty_inuse(WT_CACHE *cache)
 }
 
 /*
+ * __wt_cache_status --
+ *	Return if the cache usage exceeds the eviction or dirty targets.
+ */
+static inline void
+__wt_cache_status(WT_SESSION_IMPL *session, int *evictp, int *dirtyp)
```
agorrod: How would you feel about moving the ```WT_EVICT_PASS_XXX``` flags into the connection handle and basing decisions based on that flag, rather than sharing a check that needs to do actual work around the code?

I'm hoping to move the eviction code towards being more of a state machine, defined by the WT_EVICT_PASS flags, and shuffling this bit of that decision making code into a separate function makes that feel further away.

src/evict/evict_page.c
```diff
@@ -143,9 +144,22 @@ done:	if (((inmem_split && ret == 0) || (forced_eviction && ret == EBUSY)) &&
...
+__wt_evict_page_clean_update(
+    WT_SESSION_IMPL *session, WT_REF *ref, int exclusive)
 {
+	int evict;
+
+	/*
+	 * If doing normal system eviction, but only in the service of reducing
+	 * the number of dirty pages, leave the clean page in cache.
+	 */
+	if (!exclusive) {
```
agorrod: I don't really follow how the exclusive flag maps to "normal system eviction". Can you make that cleaner?

agorrod:
@keithbostic I've made a couple of comments, but generally I think this is a step in the right direction. We can talk about whether it makes sense to keep content in cache when splitting dirty pages as well, but I'd prefer to get this part of the change merged early than make it into something bigger.

Did you run any tests on this branch? Did it help stabilize dirty cache usage?

src/evict/evict_page.c
keithbostic: This value means two things: first, we have exclusive access to the entire file (we don't need to lock down the page), and second, we want to write, evict and discard the page, we're not interested in in-memory splits or retaining clean pages, all the transactions need to be visible, and so on.

In other words, it's "normal system eviction processing" and everything else (currently salvage and close).

How about renaming "exclusive" to "discard", or "closing"?

src/include/cache.i
keithbostic: Doing it this way should result in a better decision, the state of the system may have changed between when the eviction server chose a block and when we're choosing to discard the page in eviction.

It's clearly more work, though.

keithbostic:
Thanks, @agorrod. Let's get @michaelcahill to sign off on the approach as well, in case there's something about eviction you and I haven't thought of.

I believe the change is relatively solid (I've done overnight `test/format` runs with a big cache and `eviction_dirty_target=10`), and I've verified that pages are retained once the eviction target is met, but I have not tested if the change improves performance in real workloads.

src/include/cache.i
agorrod: My vote is to leave the decision making up to the server i.e: keep all the logic for defining what path eviction chooses in the server, rather than sharing it between the server and the thread doing eviction.

agorrod:
@keithbostic I pushed a change into this branch that's been on my todo list for ages: adding a ```eviction_dirty_trigger``` configuration. The ticket is SERVER-16737

src/evict/evict_page.c
michaelcahill: "closing" makes sense to me.

src/include/cache.h
```diff
@@ -44,6 +44,9 @@ struct __wt_evict_worker {
 	uint32_t flags;
 };
 
+#define	WT_EVICT_CHECK_SPLITS	1	/* Boolean to __wt_page_can_evict  */
+#define	WT_EVICT_EXCLUSIVE	1	/* Boolean to __wt_evict */
+
```
michaelcahill: AFAIK, this is the only place we have macros for booleans.  I didn't mind the previous flags arrangement, but this change makes it half flags, half booleans-with-names, which seems strange to me.

Can we just pass 1 if we're going with booleans?  If those calls are unclear, maybe we should go back to flags?

src/evict/evict_page.c
keithbostic: Done, I've switched from "exclusive" to "closing".

src/include/cache.h
keithbostic: Done, I switched back to using booleans.

michaelcahill:

PR #2000, opened 2015-05-27
If we have no config, explicitly check for flushing and clear txn_logsync

sueloverso:
refs  WT-1949

@michaelcahill Although this is a tiny diff, please carefully review.  I explain what is happening in the Jira ticket.  This feels a tiny bit hacky to me, but some of that is due to how the config code assumes the inheritance from the connection, and changes defaults based on that.


michaelcahill:
Thanks, @sueloverso, lgtm.  Merging.

PR #2001, opened 2015-05-28
Compact early exit

agorrod:
refs WT-1953

### DIFF
--- a/src/block/block_compact.c
+++ b/src/block/block_compact.c
@@ -55,7 +55,7 @@ __wt_block_compact_skip(WT_SESSION_IMPL *session, WT_BLOCK *block, int *skipp)
 	WT_EXT *ext;
 	WT_EXTLIST *el;
 	WT_FH *fh;
-	wt_off_t avail_eighty, avail_ninety, avail_total, eighty, ninety;
+	wt_off_t avail_eighty, avail_ninety, eighty, ninety;
 
 	*skipp = 1;				/* Return a default skip. */
 
@@ -76,24 +76,18 @@ __wt_block_compact_skip(WT_SESSION_IMPL *session, WT_BLOCK *block, int *skipp)
 		WT_ERR(__block_dump_avail(session, block));
 
 	/* Sum the available bytes in the first 80% and 90% of the file. */
-	avail_eighty = avail_ninety = avail_total = 0;
+	avail_eighty = avail_ninety = 0;
 	ninety = fh->size - fh->size / 10;
 	eighty = fh->size - ((fh->size / 10) * 2);
 
 	el = &block->live.avail;
-	WT_EXT_FOREACH(ext, el->off) {
-		avail_total += ext->size;
+	WT_EXT_FOREACH(ext, el->off)
 		if (ext->off < ninety) {
 			avail_ninety += ext->size;
 			if (ext->off < eighty)
 				avail_eighty += ext->size;
 		}
-	}
 
-	WT_ERR(__wt_verbose(session, WT_VERB_COMPACT,
-	    "%s: %" PRIuMAX "MB (%" PRIuMAX ") available space in the file",
-	    block->name,
-	    (uintmax_t)avail_total / WT_MEGABYTE, (uintmax_t)avail_total));
 	WT_ERR(__wt_verbose(session, WT_VERB_COMPACT,
 	    "%s: %" PRIuMAX "MB (%" PRIuMAX ") available space in the first "
 	    "80%% of the file",
@@ -112,6 +106,8 @@ __wt_block_compact_skip(WT_SESSION_IMPL *session, WT_BLOCK *block, int *skipp)
 	    *skipp ? "skipped" : "proceeding"));
 
 	/*
+	 * Skip files where we can't recover at least 1MB.
+	 *
 	 * If at least 20% of the total file is available and in the first 80%
 	 * of the file, we'll try compaction on the last 20% of the file; else,
 	 * if at least 10% of the total file is available and in the first 90%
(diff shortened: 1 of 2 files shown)


agorrod:
@keithbostic Could you take a look at this branch please? It's not ready for merge yet - more a point of discussion.

I reproduced the behaviour demonstrated in WT-1953 by making some simple modifications to ```ex_access.c```:

```
--- a/examples/c/ex_access.c
+++ b/examples/c/ex_access.c
@@ -90,6 +90,13 @@ main(void)
                printf("Got record: %s : %s\n", key, value);
        }
        /*! [access example cursor list] */
+       cursor->close(cursor);
+
+       {
+               int i;
+               for (i = 0; i < 100; i++)
+                       ret = session->compact(session, "table:access", NULL);
+       }
 
        /*! [access example close] */
        ret = conn->close(conn, NULL);
```

agorrod:
I attempted to split the commits up so that the different components of this change were obvious. The first change in the pull request is enough to fix the particular use case in the example code above, but I felt as though it was worthwhile pushing a bit further.

There is a more general discussion about my changes - they limit how small a file can be compacted to. I'm not sure if that is a problem for end-users. If it is we could add configuration options to compact so applications can tune how aggressive compact should be.

src/block/block_compact.c
```diff
@@ -67,7 +67,7 @@ __wt_block_compact_skip(WT_SESSION_IMPL *session, WT_BLOCK *block, int *skipp)
 	 * worth doing.  Ignore small files, and files where we are unlikely
 	 * to recover 10% of the file.
 	 */
-	if (fh->size <= 10 * 1024)
+	if (fh->size <= 10 * WT_MEGABYTE)
```
agorrod: Ten megabytes is as arbitrary as 10 K. We could move this test up into the btree layer and make it a page count, or we could make it configurable through the API.

src/block/block_compact.c
```diff
@@ -115,7 +121,7 @@ __wt_block_compact_skip(WT_SESSION_IMPL *session, WT_BLOCK *block, int *skipp)
 	 * empty file can be processed quickly, so more aggressive compaction is
 	 * less useful.
 	 */
-	if (avail_ninety >= fh->size / 10) {
+	if (avail_total > WT_MEGABYTE && avail_ninety >= fh->size / 10) {
```
agorrod: Similarly to the 10MB file size, the 1MB avail list content is arbitrary.

src/btree/bt_compact.c
```diff
@@ -65,6 +65,34 @@ __compact_rewrite(WT_SESSION_IMPL *session, WT_REF *ref, int *skipp)
...
+	*skipp = 1;
+
+	btree = S2BT(session);
+
+	/*
+	 * It's safe to look at the pindex of the root page here. Trees with
+	 * less than 4 level one pages aren't worth compacting.
+	 */
+	pindex = WT_INTL_INDEX_GET_SAFE(btree->root.page);
+	if (pindex->entries < 4)
```
agorrod: I guess trees with 4 pages off the root could be quite large, but that feels like a corner case. I'd be happy to drop this change if you don't like it.

src/block/block_compact.c
keithbostic: I agree it's arbitrary; I suspect this was intended to handle the case of lots and lots of small files where the user wanted to minimize total space.

What if we removed the `fh->size <= ###` test, and instead, did a minimum recoverable space check at the end of the function:
```
/*
 * Skip files where we can't recover at least 1MB.
 *
 * If at least 20% of the total file is available and in the first 80%
 * of the file, we'll try compaction on the last 20% of the file; else,
 * if at least 10% of the total file is available and in the first 90%
 * of the file, we'll try compaction on the last 10% of the file.
 *
 * We could push this further, but there's diminishing returns, a mostly
 * empty file can be processed quickly, so more aggressive compaction is
 * less useful.
 */
if (avail_ninety > WT_MEGABYTE && avail_ninety >= fh->size / 10) {
        ...
}
```
Does that make sense to you, instead of checking an arbitrary file size early on?

I agree we could make the size configurable through the API, but I think that can wait until we have a workload that needs it (and my bet is we never will).

Regardless, I don't think this test belongs in the btree layer, I think it's the block manager's problem.

src/block/block_compact.c
keithbostic: Should have read further before replying!

But why check`avail_total`?

If all the space is at the end of the file, it's not going to help to compact, it has to be at the start of the file. How about:

```
/*
 * Skip files where we can't recover at least 1MB.
 *
 * If at least 20% of the total file is available and in the first 80%
 * of the file, we'll try compaction on the last 20% of the file; else,
 * if at least 10% of the total file is available and in the first 90%
 * of the file, we'll try compaction on the last 10% of the file.
 *
 * We could push this further, but there's diminishing returns, a mostly
 * empty file can be processed quickly, so more aggressive compaction is
 * less useful.
 */
if (avail_eighty > WT_MEGABYTE && avail_eighty >= ((fh->size / 10) * 2)) {
        *skipp = 0;
        block->compact_pct_tenths = 2;
} else if (avail_ninety > WT_MEGABYTE && avail_ninety >= fh->size / 10) {
        *skipp = 0;
        block->compact_pct_tenths = 1;
}
```

src/btree/bt_compact.c
keithbostic: The tree might be small, but the file might be large (imagine a file with 100GB of free blocks followed by a Btree root and 4 level-one pages).

My take is this is a block-manager problem, I don't think the btree check adds much.

keithbostic:
@agorrod, I've added some comments, in general this looks good to me.

src/btree/bt_compact.c
agorrod: My initial thoughts for adding a check in the btree layer was looking at how many checkpoints exist in a file. The current check for whether subsequent checkpoints for compact are making progress is very naive (if a single page is touched by a checkpoint, we continue doing more checkpoints). What I was hoping for by adding a layer here was moving towards being able to track how much space subsequent checkpoints are reclaiming.

Now that I write that down, maybe it doesn't belong here either. I'll make the changes you suggested and open a new issue to enhance compact termination conditions.

PR #2002, opened 2015-05-28
Fix configure checks for posix_memalign

michaelcahill:
Add code from the autoconf archive to detect `posix_memalign` instead of relying on AC_CHECK_FUNCS.

refs WT-1951


agorrod:

PR #2003, opened 2015-05-28
Strip out the "log=(recover=)" configuration from the base configuration file.

keithbostic:
@sueloverso, for your review.

I don't see any easy way to create a configuration string to do this work, but it's only in one place, so it's probably OK to inline it?


sueloverso:
I'll think on what config might do, but for now I think this is okay.

PR #2004, opened 2015-05-28
txn-release resets txn->isolation

keithbostic:
@agorrod, I had a branch lying around from when we were talking about 7b6fe6a.

The one change I had that didn't get in was removing where we're clearing `txn->isolation` in `__wt_txn_release`.

Since the session's isolation can change at any time, resetting here can't be meaningful, so I think it should go.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -397,11 +397,15 @@ __wt_txn_release(WT_SESSION_IMPL *session)
 		__wt_split_stash_discard(session);
 
 	/*
-	 * Reset the transaction state to not running and release the snapshot.
+	 * Release any snapshot.
 	 */
 	__wt_txn_release_snapshot(session);
-	txn->isolation = session->isolation;
-	/* Ensure the transaction flags are cleared on exit */
+
+	/*
+	 * Clear the transaction flags: the one we care about is WT_TXN_RUNNING
+	 * which ensures we don't look at this transaction again, but clearing
+	 * them all is fine.
+	 */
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic This change looks OK to me. I was a bit nervous that we could do something between ```__wt_txn_release``` and ```__wt_txn_begin``` that assumed the isolation level is the session default. The one case I could think where that might happen was auto commit transactions - and I think that's OK, since we call ```__wt_txn_begin``` the first time an operation takes a hazard pointer on a page.

@michaelcahill I think it's worth you taking a look at this one as well.

michaelcahill:
That looks wrong to me.  What problem is it solving?

The reason we need to reset isolation when a transaction completes is if it is followed by a read.  The transaction can override the default isolation level for the session, but once the transaction completes we should go back to the session's isolation level for any subsequent operation.

michaelcahill:
Also, `WT_SESSION::reconfigure` already sets `session->txn.isolation`, so I don't see a problem there.

keithbostic:
The reason I thought it was misleading was I didn't realize that `WT_SESSION.reconfigure` reset `session->txn.isolation`, so I thought that if you did:
```
__wt_txn_release
WT_SESSION.reconfigure
read
```
the read was responsible for setting `session->txn.isolation` to match the value set by `WT_SESSION.reconfigure`, so it couldn't be meaningful.

I see why it's necessary now, I'll discard this branch.

PR #2006, opened 2015-05-29
Add fix for C2054 to lz4_compress

daveh86:
The new LZ4 code will not compile under windows as we are missing an override of "inline" to avoid C2054 errors.

@agorrod and @markbenvenuto can you have a look at this? The logic here is the same as for zlib.


keithbostic:
Thanks @daveh86, looks right to me!

PR #2007, opened 2015-05-29
WT-147 Create indices after populating primary table

ddanderson:
Allow indices to be created at any time.  The hard part about this is the locking, since while the index is being created, we are holding schema lock and we have an exclusive lock on the index data handle. We want to use regular cursor operations to insert into the index; to do so, we need to have a way for a session re-get a lock (that that session is already holding). This is the motivation behind having an 'owner' session for a `dhandle`, and a reference count.  At the same time, we need to drop the schema lock during the inserts since the LSM work thread will occasionally need to create the schema lock during a split.  The locking part is in a separate commit in case this is the wrong approach.


ddanderson:
@michaelcahill, can you review?  I don't love what I needed to do to allow locking to work, but I don't have a better approach.  Perhaps it needs a more complete solution to WT-1598 to set it right.

agorrod:
@ddanderson The schema level locking uses a different approach to the problem of nested acquisitions. I'd prefer if this code did something similar.

The schema level locks use a ```WT_WITH_LOCK``` macro (see ```src/include/schema.h```) and a flag in the session handle. We already have a ```WT_WITH_DHANDLE``` macro, could this code be structured using something like ```WT_WITH_DHANDLE_EXCLUSIVE``` that tracks whether we already have the handle (via a flag in the dhandle), and calls the ```session_get_btree```/```release_btree``` if necessary.

For an example of the schema level lock handling search for ```WT_WITH_HANDLE_LIST_LOCK```.

It's possible that I'm missing something here (maybe the index create code needs to track multiple dhandles at the same time?)

ddanderson:
@agorrod, I've pushed a simple flag-based way to do this.  This is for discussion, as it has some problems, as it fails several tests.

NOT READY FOR MERGE.

agorrod:
@ddanderson Are you waiting on anything from me for this one? Anything I can do to help?

ddanderson:
@agorrod, no I'm not waiting, and sorry this has been dormant.  I was having trouble trying to implement your suggestion. (Since the index file is created deeper in the call chain, the lock for it doesn't exist for me yet to 'prelock' it at the top level).  I hope to have another look at it tomorrow, and maybe I can explain the trouble a little better.  Thanks!

pgk2:
Don,

Is this for the Orc feature request?

-Paul

ddanderson:
@pgk2, this was requested by Orc, although I think they have way to work around this at the moment.  See WT-147.

ddanderson:
Hello @agorrod.  Sorry this has taken so long to get to.  In between other work, I tried to use various flag approaches as I understood them - I was stymied by several factors.

Having a flag on a session e.g. WT_SESSION_LOCKED_BTREE is not specific enough, as there are several Btrees in play.  The code that fills the index needs to open a cursor on the new index, like "index:maintable:x1", and when it does that, the cursor open also gets btrees for the metadata file and any column group files (or the main table file) in order to extract the pieces of the record.

This is one place where the pattern used by WT_WITH_*LOCK breaks down for this situation, there are others.  We'd like to do this:
```
    /* during code that creates index: */
    WT_WITH_BTREE_LOCK(session, dhandle, {
         create index file;
         fill index;
    });
```
The WT_WITH_BTREE_LOCK macro, as I understand, would see if a flag is set, if not, lock the needed btree.  But, at this point in the code, the btree does not exist, nor the dhandle.  Creating a btree and/or dhandle in advance to lock it seems like a difficult solution.

Also, locks should not be necessarily released when the macro completes, at least in current code, meta_track may be used to keep the handle locked until the API call completes.

I've tried several different approaches to getting a purely flag-based solution without success.  Maybe I'm missing something with this approach?  

I came to the conclusion that having a flag WT_SESSION_RELOCK_DATA_HANDLES (to allow relocking) coupled with a reference count approach solves the problem, and passes all tests.  Using a flag like this keeps the solution 'narrow' to the specific situation that we need it.

Anyway, please take another look.

michaelcahill:
@ddanderson, Alex is on vacation so I took a look through this today.  I think I mostly understand, but I'll need to spend some more time doing a careful review.

One question: why not just set `dhandle->excl_session` whenever a handle is locked exclusive, and check whether `session == dhandle->excl_session` rather than have a new flag?

ddanderson:
@michaelcahill, we can do that - we don't necessarily need a new session flag.  I've just pushed a change for that.

michaelcahill:
@ddanderson, this is looking pretty good to me.  Sorry it has languished for so long -- I'll do a more thorough review tomorrow morning, but I'd like to get this merged to use the `excl_session` part to eliminate `__wt_meta_track_find_handle` from checkpoint handle locking.

src/schema/schema_create.c
```diff
@@ -322,6 +322,60 @@ __wt_schema_index_source(WT_SESSION_IMPL *session,
...
+
+	while ((ret = tcur->next(tcur)) == 0) {
+		/*
+		 * For LSM indices, we must drop locks so that the
+		 * LSM worker thread can grab the schema lock during
+		 * a switch.
+		 */
+		WT_WITHOUT_LOCKS(session,
+		    ret = __wt_apply_single_idx(session, idx,
+		    child, ctable, child->insert));
```
michaelcahill: @ddanderson, I understand the LSM issues, but can you talk me through what makes this safe?

Can checkpoint see a partially-completed index (and if so, are there recovery implications)?  Is there any chance of deadlock when reacquiring the locks?

ddanderson: @michaelcahill, I think you are right to question its safety.  Can a checkpoint see a partially-completed index?  Probably, and I guess I would need to ask you about recovery implications!  I think dropping the lock probably has various issues we don't want to explore... 

I took another approach in this pull request: #2127 .  Basically the same as this branch, except for this lock-dropping code.  Instead, we tell LSM that we're holding the schema lock on its behalf (so it doesn't try to take it).  I don't know enough about LSM to know whether this is safe: the worker thread (which is the one that needs the lock) is asynchronous to the application thread, but it seems like file splits (where the lock is needed) are coordinated to be synchronous.  Is this too onerous or fragile?

The only other alternative I can think of is telling LSM not to split while we're filling.  Since an index being filled can be arbitrary size, this seems like it could have pretty bad performance implications for big cases.  But maybe it's okay -- how often do you need LSM indices that are dynamically created?

I suppose we could also punt and disallow dynamic index creation for LSM indices, period.

@agorrod, if you're around, perhaps you can weigh in.

agorrod:
@ddanderson Has this been replaced by #2189 now?

ddanderson:
@agorrod, Yes, this branch has been replaced by #2189, closing.

PR #2008, opened 2015-05-30
Avoid potential races in the PRNG.

keithbostic:
@michaelcahill, @markbenvenuto, this is what I think we need to do in the PRNG.

The only interesting changes are in `src/support/rand.c`, everything else is changing callers of those functions.

Let me know if I'm missing something, or if you think I've slowed things down somehow.


keithbostic:
@lcn2, of possible interest to you.

bench/wtperf/wtperf.c
```diff
@@ -2380,7 +2378,7 @@ wtperf_rand(CONFIG_THREAD *thread)
 	 * Use WiredTiger's random number routine: it's lock-free and fairly
 	 * good.
 	 */
-	rval = (uint64_t)__wt_random(thread->rnd);
+	rval = __wt_random(&thread->rnd);
```
agorrod: Is it OK to remove the cast because the value is going from smaller to bigger?

src/support/rand.c
```diff
@@ -28,43 +28,68 @@
...
-#define	M_Z	(rnd)[1]
+#define	M_Z(p)	((uint32_t *)&(p))[1]
 
 /*
  * __wt_random_init --
  *	Initialize return of a 32-bit pseudo-random number.
  */
 void
-__wt_random_init(uint32_t *rnd)
+__wt_random_init(uint64_t * volatile rnd_state)
```
agorrod: I don't understand why you are declaring the pointer as volatile instead of the variable. Can you explain that for me?

src/support/rand.c
```diff
@@ -28,43 +28,68 @@
...
+	w = M_W(rnd);
+	z = M_Z(rnd);
+
+	/*
+	 * I don't know the period of this PRNG. Do a cheap check, if the value
+	 * goes to 0 (from which we won't recover), reset to the initial state.
+	 * This has additional value if a caller fails to initialize the state,
+	 * or initializes with a seed that results in a short period.
+	 */
+	if (z == 0 || w == 0)
```
agorrod: I don't love this check - though I don't understand enough about the random number generator to say that it's not necessary.

agorrod:
@keithbostic This change looks OK to me. I'm not convinced the volatile declaration is right - so I think the optimizer may get rid of the local variable. Apart from that I'd be OK with merging this.

For the record - I'm not entirely convinced there was a problem before, but treating the read/write of a 64 bit variable as atomic is easier to reason about, so I think this is a good change.

src/support/rand.c
```diff
@@ -28,43 +28,68 @@
...
- * We have to be very careful about races here.  Multiple threads can call
- * __wt_random concurrently, and it is okay if those concurrent calls get the
- * same return value.  What is *not* okay is if reading the shared state races
- * with an update and uses two different values for m_w or m_z.  That could
- * result in a value of zero, in which case they would be stuck on zero
- * forever.  Take local copies of the shared values to avoid this.
  */
 uint32_t
-__wt_random(uint32_t *rnd)
+__wt_random(uint64_t * volatile rnd_state)
```
markbenvenuto: The type annotation is wrong. This means that that you have a volatile pointer to non-volatile memory. Since it is the memory pointed to by `rnd_state` that is volatile, you need to write as `volatile uint64_t* rnd_state` or `uint64_t volatile * rnd_state`.

bench/wtperf/wtperf.c
keithbostic: Yes, I think so.

src/support/rand.c
keithbostic: I think it's necessary: I'm checking on this, but even if the period si as big as we think it is (2^60), it doesn't hurt to do a check of values in a register, does it?

agorrod:
Thanks @keithbostic I'm happy with this change now. I'll merge.

PR #2013, opened 2015-06-05
Throttle worker threads based on eviction targets.

keithbostic:
Reference WT-1744.


src/include/cache.i
```diff
@@ -133,93 +133,6 @@ __wt_cache_status(WT_SESSION_IMPL *session, int *evictp, int *dirtyp)
 }
 
 /*
- * __wt_eviction_check --
- *	Wake the eviction server if necessary.
- */
-static inline int
-__wt_eviction_check(WT_SESSION_IMPL *session, int *fullp, int wake)
```
michaelcahill: Previously, the fast checks (when the cache isn't full) were inlined.  What's the advantage to paying for a function call?

src/evict/evict_lru.c
```diff
@@ -1443,19 +1443,82 @@ __wt_evict_lru_page(WT_SESSION_IMPL *session, int is_server)
 }
 
 /*
- * __wt_cache_wait --
- *	Wait for space in the cache.
+ * __cache_check --
+ *	Return the cache percent-full, and if eviction targets need addressing.
+ */
+static inline void
+__cache_check(WT_CONNECTION_IMPL *conn, int *pctp, int *targetp)
```
michaelcahill: The "target" name doesn't make sense to me.  Maybe "eviction_needed"?

michaelcahill:
@keithbostic, a couple of minor comments, otherwise this lgtm.  Thanks!

src/include/cache.i
keithbostic: No advantage, of course, but threads calling this function have already self-selected as not being terribly busy, right?

They're either about to stall (page-in), have no positioned cursors (cursor-enter), allocating a snapshot (txn-begin), or doing the "transaction idle check". What's the performance path that makes avoiding a function call interesting?

Regardless, happy to inline the front end of `__wt_cache_full_check`, just let me know.

src/evict/evict_lru.c
keithbostic: Thank you -- I meant to go back and look at that function and forgot. I've cleaned it up and simplified it, there's no need for the `targetp` argument at all, the function can return true/false if the application thread should be involved in eviction.

keithbostic:
@michaelcahill, do you want this merged once your comments are addressed, or should we wait for information from the YCSB runs?

src/include/cache.i
michaelcahill: I think of cursor_enter and txn_begin as performance paths where it is likely that no eviction by application threads is required.  That's why this code was structured to inline the fast checks for those cases.

Can you please change that back?

michaelcahill:
Are you running YCSB?  If not, who is going to do that and what is the schedule?

My preference would be to get this in and start getting experience with it (e.g., enabling low dirty targets in some wtperf runs) sooner rather than later.

keithbostic:
@michaelcahill, I've inlined the front-end of the eviction functions.

src/include/cache.i
```diff
@@ -199,49 +230,20 @@ __wt_cache_full_check(WT_SESSION_IMPL *session)
...
+	 * was no cache work to do. After this point, let them skip the sleep.
 	 */
-	if (F_ISSET(session,
-	    WT_SESSION_NO_CACHE_CHECK | WT_SESSION_LOCKED_SCHEMA))
-		return (0);
+	if (didworkp != NULL)
+		*didworkp = 1;
 
-	return (1);
+	return (__wt_cache_eviction_worker(session, busy, pct_full));
```
agorrod: Won't this mean that application threads will keep evicting until the eviction target has been hit? What is stopping some application threads from getting stuck doing eviction for a long time, while other threads get to do work.

keithbostic: "busy" threads only evict a single page before returning from that function, other threads evict 10 pages and then return.

agorrod: My bad - sorry.

agorrod:
lgtm - would you like to do performance testing before this is merged, or are you happy for me to merge it in now?

keithbostic:
I'm working on testing this with YCSB now. It's entirely up to you and Michael, if we're confident this is the right direction, I'm happy to merge now, if we might change our minds depending on the performance measurements, I'm happy to wait.

keithbostic:
@michaelcahill, as we discussed, I switched over to measuring checkpoint duration, using wtperf's update-checkpoint-btree workload with/without `eviction_dirty_target=40,eviction_dirty_trigger=50` configuration.

Checkpoint times for those runs are all over the map, but generally discarding the last checkpoint (which is obviously truncated), and averaging over 4 runs I see the average magnetic disk checkpoint time go from 33 seconds to 22 seconds and the average SSD checkpoint time go from 33 seconds to 24 seconds.

Throughput falls, as expected. On magnetic disk, read/insert/update averages generally drop: from 8M, 93M and 4M, to 5M, 92M and 3M; on SSDs, 53M, 160M, and 35M, to 43M, 164M and 28M, presumably because read and update threads are being pulled to do eviction.

I'm going to do another, longer run to try and iron out the variance, and I have a couple more YCSB runs I want to try, but other than that, I'd like to declare victory and merge this change.

michaelcahill:
Thanks, @keithbostic -- this change is clearly having an effect when the target is set, and the code lgtm.

Longer term, I'd like to try to make sure that inserts are throttled to keep the dirty size under control, and ideally that reads are not.  The problem is that we want to do most of the throttling outside of a transaction, and we currently have no way to know whether the next operation will be an update or a query.

One possibility is to change the API so that applications mark read-only transactions.  Another is to do the heavyweight throttling at transaction commit rather than transaction begin.

PR #2014, opened 2015-06-06
AC_TRY_RUN is obsolete.

keithbostic:
@michaelcahill, did you have a test you ran to integrate the AX_FUNC_POSIX_MEMALIGN macro, could we re-run that test?


michaelcahill:
Without that macro, the build was broken by an upgrade to Fedora 22 because posix_memalign is a compiler builtin with GCC 5+ and autoconf's standard way to detect functions didn't compile.  This change has already passed testing on tinderbox, so it should be fine.

PR #2015, opened 2015-06-06
Move the statistics=(size) support into the block manager

keithbostic:
@michaelcahill, I think this work should be in the block manager, for clarity.

Thoughts?


michaelcahill:

PR #2016, opened 2015-06-06
Change WT_SESSION.verify to not error on leaked blocks, by default.

keithbostic:
Reference WT-1959.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -834,6 +834,11 @@ methods = {
     Config('dump_shape', 'false', r'''
         Display the shape of the tree after verification,
         using the application's message handler, intended for debugging''',
+        type='boolean'),
+    Config('strict', 'false', r'''
+        Treat any verification problem as an error; by default, verify will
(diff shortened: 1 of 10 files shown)


michaelcahill:
This needs a note in the upgrading section of the docs.  Otherwise, lgtm.  Thanks!

keithbostic:
Here's what I added to the upgrading docs:

```
The WT_SESSION.verify method in this release has a new configuration option,
\c strict. By default, with \c strict set to \c false, WT_SESSION.verify will no longer 
return an error for problems that do not impact the future use of the object (for
example, if a leaked block were detected, the application can continue to run). 
WT_SESSION.verify will continue to output an error message whenever an error
is detected, only the final return value is affected.  This change allows
applications to verify objects and continue if at all possible. Applications in
development should configure \c strict to \c true in order to terminate the
application whenever an error is detected.
```

Once the build tests are done, I'll merge the change.

PR #2018, opened 2015-06-08
SERVER-18829 Have pages start in the middle of the LRU queue for eviction

michaelcahill:
Also make sure that when eviction first needs to run, it can find some pages to
evict.

### DIFF
--- a/src/include/cache.i
+++ b/src/include/cache.i
@@ -27,11 +27,11 @@ __wt_cache_read_gen_incr(WT_SESSION_IMPL *session)
 }
 
 /*
- * __wt_cache_read_gen_set --
- *      Get the read generation to store in a page.
+ * __wt_cache_read_gen_bump --
+ *      Get the read generation to keep a page in memory.
  */
 static inline uint64_t
-__wt_cache_read_gen_set(WT_SESSION_IMPL *session)
+__wt_cache_read_gen_bump(WT_SESSION_IMPL *session)
 {
 	/*
 	 * We return read-generations from the future (where "the future" is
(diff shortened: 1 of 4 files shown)


src/include/cache.i
```diff
@@ -46,6 +46,19 @@ __wt_cache_read_gen_set(WT_SESSION_IMPL *session)
...
+ * __wt_cache_read_gen_new --
+ *      Get the read generation for a new page in memory.
+ */
+static inline uint64_t
+__wt_cache_read_gen_new(WT_SESSION_IMPL *session)
+{
+	WT_CACHE *cache;
+
+	cache = S2C(session)->cache;
+	return (__wt_cache_read_gen(session) + cache->read_gen_oldest) / 2;
```
agorrod: Is ```read_gen_oldest``` the right thing to choose here? It's the oldest generation in the queue, not the oldest generation across the cache.

agorrod:
Thanks @michaelcahill looks good to me. I had one question about the chosen read generation, but in general I like this change.

I've also confirmed that this helps the workload I was testing - which was to do a cursor traversal of a very large table, starting with an empty large cache. We were temporarily using more than the available cache without this change, because it takes ~30 seconds for eviction to start choosing candidates after it has reached the trigger size, since no page starts with a read generation, and the read generation starts at 0 at that stage.

src/include/cache.i
michaelcahill: The idea is for new pages to get a priority that puts them "in the middle" of existing pages.  That way, pages that are read once (e.g., in a big sequential scan) will not cause the top 50% of pages to be evicted from cache.

Of course, we don't have a global LRU queue, so we don't know the real oldest read generation.  This is just an attempt to calculate a reasonable value given our current data structures.

michaelcahill:
@agorrod, thanks for the review.  If you are happy merging and opening a new ticket to discuss measuring / improving eviction behavior, that's fine by me.

agorrod:
I'll go ahead and merge this - it's definitely an improvement. We can tweak exactly what we set the initial generation to over time.

PR #2023, opened 2015-06-23
WT-1966 Change how the shared cache assigns priority to participants.

agorrod:
The old code implicitly assumed at least 10 members. The new
code uses a pressure based on a percentage of the most active
member.

refs WT-1966

### DIFF
--- a/src/conn/conn_cache_pool.c
+++ b/src/conn/conn_cache_pool.c
@@ -11,10 +11,16 @@
 /*
  * Tuning constants.
  */
-/* Threshold when a connection is allocated more cache */
-#define	WT_CACHE_POOL_BUMP_THRESHOLD	6
-/* Threshold when a connection is allocated less cache */
-#define	WT_CACHE_POOL_REDUCE_THRESHOLD	2
+/*
+ * Threshold when a connection is allocated more cache, as a percentage of
+ * the amount of pressure the busiest participant has.
+ */
+#define	WT_CACHE_POOL_BUMP_THRESHOLD	60
+/*
+ * Threshold when a connection is allocated less cache, as a percentage of
+ * the amount of pressure the busiest participant has.
+ */
+#define	WT_CACHE_POOL_REDUCE_THRESHOLD	20
 /* Balancing passes after a bump before a connection is a candidate. */
 #define	WT_CACHE_POOL_BUMP_SKIPS	10
 /* Balancing passes after a reduction before a connection is a candidate. */
@@ -444,7 +450,8 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
 	uint64_t entries, highest, new;
 
 	cp = __wt_process.cache_pool;
-	entries = highest = 0;
+	entries = 0;
+	highest = 1; /* Avoid divide by zero */
 
 	/* Generate read pressure information. */
 	TAILQ_FOREACH(entry, &cp->cache_pool_qh, cpq) {
@@ -466,9 +473,6 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
 	WT_RET(__wt_verbose(session, WT_VERB_SHARED_CACHE,
 	    "Highest eviction count: %" PRIu64 ", entries: %" PRIu64,
 	    highest, entries));
-	/* Normalize eviction information across connections. */
-	highest = highest / (entries + 1);
-	++highest; /* Avoid divide by zero. */
 
 	*phighest = highest;
 	return (0);
@@ -487,13 +491,16 @@ __cache_pool_adjust(WT_SESSION_IMPL *session,
 	WT_CACHE_POOL *cp;
 	WT_CACHE *cache;
 	WT_CONNECTION_IMPL *entry;
-	uint64_t adjusted, reserved, read_pressure;
+	uint64_t adjusted, highest_percentile, read_pressure, reserved;
 	int force, grew;
 
 	*adjustedp = 0;
 	cp = __wt_process.cache_pool;
 	force = (cp->currently_used > cp->size);
 	grew = 0;
+	/* Highest as a percentage, avoid 0 */
+	highest_percentile = (highest / 100) + 1;
+
 	if (WT_VERBOSE_ISSET(session, WT_VERB_SHARED_CACHE)) {
 		WT_RET(__wt_verbose(session,
 		    WT_VERB_SHARED_CACHE, "Cache pool distribution: "));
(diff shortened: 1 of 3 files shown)


agorrod:
@michaelcahill Could you review this change please?

@SubuSS I've tested this with a single-participant workload and it improves the shared cache usage characteristics.

src/conn/conn_cache_pool.c
```diff
@@ -456,19 +471,37 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
...
+			cache->cp_pass_read = new - cache->cp_saved_read;
 		else
-			cache->cp_current_read = new;
+			cache->cp_pass_read = new;
 		cache->cp_saved_read = new;
-		if (cache->cp_current_read > highest)
-			highest = cache->cp_current_read;
+
+		/* Update the application eviction count information */
+		new = cache->cp_saved_app_evicts;
```
michaelcahill: I'd avoid the use of C++ operators as C variable names.  It will bite us at some point.

src/conn/conn_cache_pool.c
```diff
@@ -506,7 +542,14 @@ __cache_pool_adjust(WT_SESSION_IMPL *session,
...
 
-		read_pressure = cache->cp_current_read / highest;
+		/*
+		 * The read pressure is calculated as a percentage of how
+		 * much read pressure there is on this participant compared
+		 * to the participant with the most activity. The closer we
+		 * are to the most active the more cache we should get
+		 * assigned.
+		 */
+		read_pressure = cache->cp_pass_read / highest_percentile;
```
michaelcahill: Can you talk me through this part because I'm a little confused.

We're taking into account reads, eviction and application waits to calculate which participants could most benefit from more cache.  But here we're only using reads to calculate the "pressure"?  I'd expected that we would combine the scores and use the combined value everywhere (i.e. `cp_pass_pressure` instead of `cp_pass_read`).

agorrod: Thanks. It was just broken. Then intent was to switch this to use ```cp_pass_pressure```. I'll fix.

src/conn/conn_cache_pool.c
```diff
@@ -453,22 +468,41 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
...
-		cache->cp_saved_read = new;
-		if (cache->cp_current_read > highest)
-			highest = cache->cp_current_read;
+			cache->cp_pass_read = tmp;
+		cache->cp_saved_read = tmp;
+
+		/* Update the application eviction count information */
+		tmp = cache->cp_saved_app_evicts;
+		cache->cp_pass_app_evicts =
+		    tmp - cache->cp_pass_app_evicts;
```
michaelcahill: I don't follow this calculation.  Let's say there are 10 evictions happening between each pass.  Then we would have:

pass 1: saved = 0, pass = 10 --> saved = -10
pass 2: saved = -10, pass = 10 --> saved = -20

Am I misunderstanding the logic?

etc.

src/conn/conn_cache_pool.c
```diff
@@ -453,22 +468,41 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
...
+		cache->cp_saved_app_evicts = tmp;
+
+		/* Update the eviction wait information */
+		tmp = cache->cp_saved_app_waits;
+		cache->cp_pass_app_waits =
+		    tmp - cache->cp_pass_app_waits;
+		cache->cp_saved_app_waits = tmp;
```
michaelcahill: This seems to have the same problem: I don't understand what is being stored in "app_waits" vs "pass_waits".

src/include/cache.h
```diff
@@ -105,8 +108,13 @@ struct __wt_cache {
...
 	 */
-	uint64_t cp_saved_read;		/* Read count from last pass */
-	uint64_t cp_current_read;	/* Read count from current pass */
+	uint64_t cp_saved_app_evicts;	/* User eviction count at last review */
+	uint64_t cp_saved_app_waits;	/* User wait count at last review */
+	uint64_t cp_saved_read;		/* Read count at last review */
+	uint64_t cp_pass_app_evicts;	/* User eviction count from this pass */
+	uint64_t cp_pass_app_waits;	/* User wait count from this pass */
+	uint64_t cp_pass_pressure;	/* Calculated pressure from this pass */
+	uint64_t cp_pass_read;		/* Read count from this pass */
```
michaelcahill: Consider reordering either for alphabetical or to group the "saved" with the "pass" for each category.

src/evict/evict_lru.c
```diff
@@ -1494,6 +1494,7 @@ __wt_cache_wait(WT_SESSION_IMPL *session, int full)
 
 		switch (ret = __wt_evict_lru_page(session, 0)) {
 		case 0:
+			cache->app_evicts++;
```
michaelcahill: Another way to express my confusion with the calculation is that I don't see where this is being read.

src/conn/conn_cache_pool.c
```diff
@@ -453,22 +468,41 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
...
-		if (new >= cache->cp_saved_read)
-			cache->cp_current_read = new - cache->cp_saved_read;
+		if (tmp >= cache->cp_saved_read)
+			cache->cp_pass_read = tmp - cache->cp_saved_read;
 		else
-			cache->cp_current_read = new;
-		cache->cp_saved_read = new;
-		if (cache->cp_current_read > highest)
-			highest = cache->cp_current_read;
+			cache->cp_pass_read = tmp;
```
SubuSS: This *might be losing some information – when wrapping we use tmp directly. (means if you go from int_max-100 to 2, you will count tmp as 2 not 102). Am I missing something?

src/conn/conn_cache_pool.c
```diff
@@ -453,22 +469,54 @@ __cache_pool_assess(WT_SESSION_IMPL *session, uint64_t *phighest)
...
+		 */
+		tmp = cache->bytes_read;
+		if (tmp >= cache->cp_saved_read)
+			reads = tmp - cache->cp_saved_read;
 		else
-			cache->cp_current_read = new;
-		cache->cp_saved_read = new;
-		if (cache->cp_current_read > highest)
-			highest = cache->cp_current_read;
+			reads = (UINT64_MAX - cache->cp_saved_read) + tmp;
```
michaelcahill: How about just `reads = tmp` if there is wrapping?  It's not like it should happen often...

michaelcahill:
@agorrod, this lgtm when you are ready to merge.  I think the wrapping case could be simplified, but that's mostly cosmetic.

src/conn/conn_cache_pool.c
agorrod: I had that originally, but it was more confusing to document why it was OK for the value to be inaccurate than to make it accurate.

michaelcahill: Fair enough, then this lgtm.

PR #2025, opened 2015-06-24
WT-1978 Better checking and tests for index cursor comparison

ddanderson:


ddanderson:
@michaelcahill or @agorrod, this is ready for review.

src/cursor/cur_index.c
```diff
@@ -76,7 +76,8 @@ __curindex_compare(WT_CURSOR *a, WT_CURSOR *b, int *cmpp)
 
 	/* Check both cursors are "index:" type. */
 	if (!WT_PREFIX_MATCH(a->uri, "index:") ||
-	    !WT_PREFIX_MATCH(b->uri, "index:"))
+	    !WT_PREFIX_MATCH(b->uri, "index:") ||
```
agorrod: @ddanderson Is this second check for ```index:``` redundant? Won't the ```strcmp``` catch it?

agorrod:
Thanks @ddanderson nice catch. The change looks correct to me, but I think the check is more complex than necessary.

src/cursor/cur_index.c
ddanderson: @agorrod, you're right.  Thanks, I'll fix.

PR #2026, opened 2015-06-24
WT-1975 Wait for sync_lsn of earlier log files to complete for forced sync.

sueloverso:
@michaelcahill Please review this fix.  As I was not able to repro the failure, I found this by code inspection.


michaelcahill:
I think I understand the fix, but just let me spell it out for posterity.  There was a window after switching to a new log file, where we could sync the new log file and move `sync_lsn` forward before the previous log file sync had completed.  That could cause the previous log file to be removed before the sync completed, which could lead to the test failure.

This lgtm, I'll merge, and if I've misunderstood, please let me know.

PR #2027, opened 2015-06-24
SERVER-18899 Add unit test to simulate fsyncLock.

sueloverso:
@michaelcahill Please review this PR to implement a test for the fsyncLock strategy.  In particular I currently do a shutil.copytree and that may have problems on Windows (based on the comment in test_durability01.py that I based this on).  I updated the doc because if skipping the WiredTiger.lock file is going to be an issue, it'll need to be documented in some clear way (non-Windows can do 'cp -r', Windows will have to walk over each file name and skip that one).

### DIFF
--- a/test/suite/test_backup05.py
+++ b/test/suite/test_backup05.py
@@ -44,32 +44,37 @@ class test_backup05(wttest.WiredTigerTestCase, suite_subprocess):
     create_params = 'key_format=i,value_format=i'
     freq = 5
 
+    def copy_windows(self, olddir, newdir):
+        os.mkdir(newdir)
+        for fname in os.listdir(olddir):
+            fullname = os.path.join(olddir, fname)
+            # Skip lock file on Windows since it is locked
+            if os.path.isfile(fullname) and "WiredTiger.lock" not in fullname:
+                shutil.copy(fullname, newdir)
+
     def check_manual_backup(self, i, olddir, newdir):
         ''' Simulate a manual backup from olddir and restart in newdir. '''
         self.session.checkpoint()
         cbkup = self.session.open_cursor('backup:', None, None)
         # with the connection still open, copy files to new directory
         shutil.rmtree(newdir, ignore_errors=True)
-        # !!! A similar test, test_durability01.py contains the following:
-        #    Skip lock file on Windows since it is locked
-        #    if os.path.isfile(fullname) and "WiredTiger.lock" not in fullname:
-        #        shutil.copy(fullname, newdir)
-        # I suspect most fsyncLock users will do 'cp -r' which is what
-        # shutil.copytree does.
 
-        # Half the time use copytree, the other half use a dd command that
-        # does not align on a block boundary.
-        if i % (self.freq * 2) == 0:
-            os.mkdir(newdir)
-            for fname in os.listdir(olddir):
-                fullname = os.path.join(olddir, fname)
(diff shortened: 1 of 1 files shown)


test/suite/test_backup05.py
```diff
@@ -0,0 +1,108 @@
...
+
+        # Half the time use copytree, the other half use a dd command that
+        # does not align on a block boundary.
+        if i % (self.freq * 2) == 0:
+            os.mkdir(newdir)
+            for fname in os.listdir(olddir):
+                fullname = os.path.join(olddir, fname)
+                inpf = 'if=' + fullname
+                outf = 'of=' + newdir + '/' + fullname
+                cmd_list = ['dd', inpf, outf, 'bs=300']
```
markbenvenuto: `dd` does not exist on Windows machines unless you run under `Cygwin`.

sueloverso: Thanks.  I'll make a windows-specific conditional that avoids the dd and avoids the copytree, but copies each file other than the lock file.

michaelcahill:
@sueloverso, the test logic itself lgtm.

@ddanderson, can you please review with an eye to abstracting out the code to make a copy of a database directory?  It could just be shared code to list the files to copy (excluding the lock file), but it's probably better to have all of this code in one place that can be shared by any tests that need to copy live databases.

test/suite/test_backup05.py
```diff
@@ -0,0 +1,113 @@
...
+        conn.close()
+        
+    def test_backup(self):
+        '''Check manual fsyncLock backup strategy'''
+
+        # Here's the strategy:
+        #    - update the table
+        #    - checkpoint the database
+        #    - open a backup cursor
+        #    - copy the database directory (live, simulating a crash)
```
ddanderson: @sueloverso, this part of the comment ("simulating a crash") doesn't make sense to me.

test/suite/test_backup05.py
```diff
@@ -0,0 +1,113 @@
...
+        #      - use copy tree or non-aligned dd
+        #    - verify in the copy
+        #    - repeat
+        #
+        # If the metadata isn't flushed, eventually the metadata we copy will
+        # be sufficiently out-of-sync with the data file that it won't verify.
```
ddanderson: I don't understand this comment either. If it doesn't verify, should we expect error returns or output from verify?

ddanderson:
@sueloverso, this lgtm, except a couple of nits.  I can do some refactoring as suggested by @michaelcahill, after it's merged.

test/suite/test_backup05.py
michaelcahill: @ddanderson, the principle here is that copying a live database then trying to open it is equivalent to what would happen if the process crashed at that point and had to recover.  There are additional things that can go wrong when the power fails that we aren't testing, but this catches a lot of problems.  We test recovery this way in several places.

test/suite/test_backup05.py
michaelcahill: We have had bugs in the past where the metadata could end up out-of-sync with data files.  This is a regression test for those issues: we don't expect verify to fail unless there is a bug in WiredTiger.  The comment could be reworded to clarify.

michaelcahill:
OK, I'm good with this.  I'll merge.  @ddanderson, can you please do the refactoring on develop?

PR #2030, opened 2015-06-25
WT-1962 Fix usage of log->prep_missed.

sueloverso:
@michaelcahill or @agorrod please review this small diff.  While investigating WT-1962, I discovered we were clearing log->prep_missed at the wrong time.  I also moved where we incremented it to where a clearer, more readable location.


michaelcahill:

PR #2031, opened 2015-06-26
WT-1964: In Java fix session/cursor close in different thread than open

ddanderson:
Limit when we use a saved JNIEnv pointer (basically the JNI abstraction of a Java thread). When we have a JNIEnv pointer available from an explicit close call, use that. When we are implicitly closing a session or cursor, use the connection's JNIEnv pointer.  Add tests for this, and some code comments.


ddanderson:
@agorrod or @michaelcahill, this is ready for review.

agorrod:
Thanks @ddanderson This looks good to me - I'll merge.

PR #2032, opened 2015-06-29
WT-1981 Signed int pack bug

agorrod:


agorrod:
@michaelcahill Could you review the fix please? The change is majority test case, the fix is a single line where we incremented the packed buffer offset by too much.

michaelcahill:
Thanks, @agorrod -- lgtm.  I'll merge.  Can you please follow up on WT-1981?

PR #2033, opened 2015-06-29
WT-1983 Wtperf truncate

daveh86:
WT-1983 Wtperf truncate


agorrod:
@daveh86 Is this ready for review?

daveh86:
@agorrod, yes. It is now passing the PR Tester.

agorrod:
@daveh86 please review for alphabetical ordering

bench/wtperf/config.c
```diff
@@ -253,9 +268,14 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
 		scan = NULL;
 		if (ret != 0)
 			goto err;
-
-		if (workp->insert == 0 &&
-		    workp->read == 0 && workp->update == 0)
+		if (workp->insert == 0 && workp->read == 0 &&
+		    workp->update == 0 && workp->truncate == 0)
+			goto err;
+		/* Why run with truncate if we don't want to truncate anything*/
```
agorrod: Missing space between last character and end of comment.

bench/wtperf/config.c
```diff
@@ -253,9 +268,14 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
...
+		if (workp->truncate != 0 &&
+		    workp->truncate_pct == 0 && workp->truncate_count == 0)
+			goto err;
+		if (workp->truncate_pct < 1 && workp->truncate_pct > 99)
```
agorrod: I think you want an ```||``` here instead of an ```&&```

bench/wtperf/wtperf.c
```diff
@@ -1,4 +1,4 @@
-/*-
+	/*-
```
agorrod: Is this a typo?

bench/wtperf/wtperf.c
```diff
@@ -313,6 +315,9 @@ worker_async(void *arg)
 			if (wtperf_value_range(cfg) < next_val)
 				continue;
 			break;
+		case WORKER_TRUNCATE:
```
agorrod: It would be better to catch an attempt to setup truncate and async ops in the configuration parsing, and return an error, and not have any truncate code in this async loop. Rather than having a partial no-op implementation here.

i.e: If I configure async and truncate, I want to get an error on starting wtperf, rather than it turning the configuration into a no-op.

bench/wtperf/wtperf.c
```diff
@@ -455,6 +465,21 @@ worker(void *arg)
 		lprintf(cfg, ret, 0, "First transaction begin failed");
 		goto err;
 	}
+	/* Setup truncate values */
+	if (thread->workload->truncate > 0) {
```
agorrod: It's not clear anywhere whether truncate is a count or a boolean. I believe it's a boolean - as such it would be good to make that clear in the configuration API, and to treat it as a boolean in the code (i.e: I'd change this to ```!= 0```).

bench/wtperf/wtperf.c
```diff
@@ -455,6 +465,21 @@ worker(void *arg)
...
+		/**
```
agorrod: Why the double star on this comment line?

bench/wtperf/wtperf.c
daveh86: Done.

bench/wtperf/wtperf.c
```diff
@@ -548,6 +572,40 @@ worker(void *arg)
 			if ((ret = cursor->insert(cursor)) == 0)
 				break;
 			goto op_err;
+		case WORKER_TRUNCATE:
+			if ((ret = cursor->reset(cursor)) != 0) {
+				lprintf(cfg, ret, 0, "Cursor reset failed");
+				goto err;
```
agorrod: This should goto op_err, not err.

bench/wtperf/wtperf.c
```diff
@@ -2068,8 +2138,12 @@ main(int argc, char *argv[])
 	 * If the user wants compaction, then we also enable async for
 	 * the compact operation, but not for the workloads.
 	 */
-	if (cfg->async_threads > 0)
+	if (cfg->async_threads > 0) {
+		if (cfg->workload->truncate > 0) {
```
agorrod: If this is a single line, no need for curly braces. It would be better to output an error message for the user, telling them why we are erroring out.

bench/wtperf/wtperf.c
```diff
@@ -548,6 +572,40 @@ worker(void *arg)
 			if ((ret = cursor->insert(cursor)) == 0)
 				break;
 			goto op_err;
+		case WORKER_TRUNCATE:
+			if ((ret = cursor->reset(cursor)) != 0) {
+				lprintf(cfg, ret, 0, "Cursor reset failed");
+				goto err;
+			}
+
+			while ((ret = cursor->next(cursor)) == 0) {
```
agorrod: This loop is going to perform really horribly. We want to match the semantic of oplog as closely as we can. Can you investigate how the oplog works as at https://mongodbcr.appspot.com/6570003/

Then write up a little something about how we could map that onto this truncate implementation in wtperf?

bench/wtperf/wtperf.c
```diff
@@ -63,6 +64,15 @@ static const CONFIG default_cfg = {
...
 };
 
+/* Queue entry for use with the Truncate Logic */
+struct truncate_queue_entry {
+	char *key;
+	STAILQ_ENTRY(truncate_queue_entry) q;
+};
+
+/* Queue head for use with the Truncate Logic */
+STAILQ_HEAD(, truncate_queue_entry) truncate_stone_head;
```
agorrod: You should give the queue a name (first parameter).

bench/wtperf/wtperf.c
```diff
@@ -63,6 +64,15 @@ static const CONFIG default_cfg = {
 #undef OPT_DEFINE_DEFAULT
 };
 
+/* Queue entry for use with the Truncate Logic */
+struct truncate_queue_entry {
+	char *key;
+	STAILQ_ENTRY(truncate_queue_entry) q;
+};
```
agorrod: We'd normally typedef the structure with a capitalised name, like:

```typedef struct __truncate_queue_entry TRUNCATE_QUEUE_ENTRY;```

File: bench/wtperf/wtperf.c [hunk 12]
```diff
@@ -379,6 +391,7 @@ static void *
 worker(void *arg)
 {
 	struct timespec start, stop, interval;
+	struct truncate_queue_entry *item;
```
agorrod: This will become a ```TRUNCATE_QUEUE_ENTRY``` once you have the typedef.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +475,46 @@ worker(void *arg)
 		goto err;
 	}
 
+	/* Setup for truncate */
+	if (thread->workload->truncate != 0) {
+
+		STAILQ_INIT(&truncate_stone_head);
+
+		/* Truncation % value. eg 10% is 0.1 */
```
agorrod: The first % in the comment should be the word ```percentage```

bench/wtperf/wtperf.c
```diff
@@ -456,6 +475,46 @@ worker(void *arg)
...
+		truncation_percentage =
+		    (double)thread->workload->truncate_pct / 100;
+		/* How wide each milestone is */
```
agorrod: I'm not sure what ```wide``` means in this context?

bench/wtperf/wtperf.c
```diff
@@ -456,6 +475,46 @@ worker(void *arg)
...
+			starting_point = 0;
+			/* Shouldn't happen */
+			if (cfg->insert_key > thread->workload->truncate_count)
+				starting_point = cfg->insert_key -
+				    thread->workload->truncate_count;
+			for (i = 0; i < needed_milestones; i++) {
+				truncate_key = malloc(cfg->key_sz);
+				item = malloc(sizeof(item));
+				if (item == NULL) {
+					printf("malloc failed\n");
```
agorrod: Should be real error handling - consistent with the rest of wtperf.

agorrod: Also - no need for braces around a single line statement.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +475,46 @@ worker(void *arg)
...
+		truncate_milestone_gap =
+		    thread->workload->truncate_count*truncation_percentage;
+		/* How many milestones we need */
+		needed_milestones = thread->workload->truncate_count /
+		    truncate_milestone_gap;
+
+		/* Setup the initial milestones if we have data */
+		if (cfg->insert_key != 0) {
+			starting_point = 0;
+			/* Shouldn't happen */
```
agorrod: Why not?

bench/wtperf/wtperf.c
```diff
@@ -456,6 +475,46 @@ worker(void *arg)
...
+			if (cfg->insert_key > thread->workload->truncate_count)
```
agorrod: I think this would be better as something like (pseudo code):

```
start_point = cursor->next();
start_point_key = copy_key(cursor->get_key());
start_point_val = decode_key_from_string(start_point_key);
cursor->reset();
end_point = cursor->prev();
end_point_key = copy_key(cursor->get_key());
start_point_val = decode_key_from_string(end_point_key);

initial_milestone_gap = (end_point_val - start_point_val) / needed_milestones;
/* Deal with pathological case where there is less content than milestones. */
if (initial_milestone_gap == 0)
    initial_milestone_gap = 1;
```

bench/wtperf/wtperf.c
agorrod: A better name for this would be ```truncate_item```

bench/wtperf/config.c
```diff
@@ -243,6 +243,21 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
 					goto err;
 				continue;
 			}
+			if (STRING_MATCH("truncate", k.str, k.len)) {
+				if ((workp->truncate = v.val) <= 0)
```
agorrod: I'd expect ```truncate == 0``` to turn truncate off, not generate an error.

bench/wtperf/wtperf.c
```diff
@@ -27,6 +27,7 @@
  */
 
 #include "wtperf.h"
+#include "../src/include/queue.h"
```
agorrod: I'm surprised this is necessary. ```wtperf.h``` includes ```wt_internal.h```, which includes ```queue.h```

bench/wtperf/wtperf.c
```diff
@@ -63,6 +64,16 @@ static const CONFIG default_cfg = {
...
 
+/* Queue entry for use with the Truncate Logic */
+struct __truncate_queue_entry {
+	char *key;
+	STAILQ_ENTRY(__truncate_queue_entry) q;
+};
+typedef struct __truncate_queue_entry TRUNCATE_QUEUE_ENTRY;
+
+/* Queue head for use with the Truncate Logic */
+STAILQ_HEAD(__truncate_qh, __truncate_queue_entry) truncate_stone_head;
```
agorrod: This belongs in ```wtperf.h``` in the ```__config``` structure, rather than declaring it globally.

bench/wtperf/wtperf.c
```diff
@@ -109,6 +120,16 @@ generate_key(CONFIG *cfg, char *key_buf, uint64_t keyno)
 	sprintf(key_buf, "%0*" PRIu64, cfg->key_sz - 1, keyno);
 }
 
+static inline uint64_t
+decode_key(CONFIG *cfg, char *key_buf)
+{
+	uint32_t i = 0;
```
agorrod: The initializer shouldn't be on the same line as the declaration.

bench/wtperf/wtperf.c
daveh86: Should the struct be moved over to `wtperf.h` also?

bench/wtperf/wtperf.c
```diff
@@ -109,6 +120,16 @@ generate_key(CONFIG *cfg, char *key_buf, uint64_t keyno)
...
+	for (;i < cfg->key_sz; i++)
+		if (key_buf[i] != 0)
+			break;
+	return (atoi (key_buf + i));
```
agorrod: Keys can be 64 bit unsigned values, so ```atoi``` is inadequate here. Some googling revealed ```strtoull``, which is more likely to be correct.

bench/wtperf/wtperf.c
```diff
@@ -445,6 +474,7 @@ worker(void *arg)
 	}
 
 	key_buf = thread->key_buf;
+	truncate_buf = malloc(cfg->key_sz);
```
agorrod: I don't see this getting freed anywhere. Could you use the ```THREAD->key_buf``` instead?

This would show up as a memory leak. It's worth getting into the habit of testing any changes by running in valgrind before pushing changes - that will uncover memory leaks.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
 		goto err;
 	}
 
+	/* Setup for truncate */
+	if (thread->workload->truncate != 0) {
+
+		STAILQ_INIT(&truncate_stone_head);
```
agorrod: Once this moves into the CONFIG structure, the initializer belongs somewhere shared, rather than in a worker. Maybe in ```start_run```.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+
+		cursor = cursors[cfg->icount % cfg->table_count];
```
agorrod: AFAI can see this is the only place you reference the multi-table configuration supported by ```wtperf```. I suspect this won't work properly all the time.

I'd recommend excluding this truncate functionality from the multi-table configuration for now during configuration parsing phase.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+
+		/* Truncation percentage value. eg 10% is 0.1 */
+		truncation_percentage =
+		    (double)thread->workload->truncate_pct / 100;
+		/* How many entries between each milestone */
+		truncate_milestone_gap =
+		    thread->workload->truncate_count*truncation_percentage;
```
agorrod: Spacing is wrong here - you need spaces around the ```*```

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+		/* How many milestones we need */
+		needed_milestones = thread->workload->truncate_count /
+		    truncate_milestone_gap;
+
+		/* Reset this value for use again */
+		truncate_milestone_gap = 0;
+
+		cursor->next(cursor);
```
agorrod: You are ignoring the return value of ```cursor->next```. You should handle the case when the table is empty.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+		ret = cursor->get_key(cursor, &truncate_buf);
+
+		/* We have data */
+		if (ret == 0) {
```
agorrod: I'd switch the order of error checking here. Make it:

```
if (ret != 0) {
    lprintf(...)
    goto err;
}
```

Then you don't need to make the following code conditional on success.

bench/wtperf/wtperf.c
agorrod: A comment explaining what the following code is doing would be worthwhile.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+			start_point_val = decode_key(cfg, truncate_buf);
+			cursor->reset(cursor);
+			if ((ret = cursor->prev(cursor)) != 0) {
+				lprintf(cfg, ret, 0,
+				    "truncate setup end: prev failed");
+				goto err;
+			}
+			if ((ret = cursor->get_key(cursor,
+				    &truncate_buf)) != 0) {
```
agorrod: Indentation is wrong here.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+				lprintf(cfg, ret, 0,
+				    "truncate setup end: get_key failed");
+				goto err;
+			}
+			end_point_val = decode_key(cfg, truncate_buf);
+				/* Not enough documents? */
```
agorrod: Indentation is wrong here.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +486,79 @@ worker(void *arg)
...
+				if (start_point_val + needed_milestones
+				    > end_point_val)
```
agorrod: The ```>``` belongs on the previous line.

agorrod:
@daveh86 I've made a number of comments, could you address them please?

Also I see a number of memory leaks in the new code. I'd recommend testing the code under valgrind to uncover them.

bench/wtperf/wtperf.c
agorrod: Yes - it will need to be for the declaration to work.

daveh86:
@agorrod, can you take another look at this

bench/wtperf/wtperf.c
```diff
@@ -387,21 +405,29 @@ worker(void *arg)
 	WT_SESSION *session;
 	int64_t ops, ops_per_txn, throttle_ops;
 	size_t i;
-	uint64_t next_val, usecs;
+	uint64_t final_milestone_gap, needed_milestones, next_val;
```
agorrod: Could you split the truncate part out into a different function - it makes the ```worker``` function ever more complex.

daveh86:
@agorrod , mind having a look at the latest changes?

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
 static int
+setup_truncate(
+    CONFIG *cfg, CONFIG_THREAD *thread,
+    TRUNCATE_CONFIG *trunc, WT_CURSOR *cursor) {
+
+	int ret;
+	size_t i;
+	uint64_t end_point_val, final_milestone_gap, start_point_val;
+	char *key, *truncate_key;
+	TRUNCATE_QUEUE_ENTRY *truncate_item;
```
agorrod: The local variable declarations aren't ordered properly.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
 }
 
 static int
+setup_truncate(
+    CONFIG *cfg, CONFIG_THREAD *thread,
+    TRUNCATE_CONFIG *trunc, WT_CURSOR *cursor) {
```
agorrod: The line wrapping doesn't look quite right to me here. Could it fit more like the ```start_threads``` method?

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+		ret = cursor->get_key(cursor, &key);
+	}
+	/* We have data */
+	if (ret == 0) {
+		start_point_val = decode_key(cfg, key);
+		cursor->reset(cursor);
+		if ((ret = cursor->prev(cursor)) != 0) {
+			lprintf(cfg, ret, 0,
+			    "truncate setup end: prev failed");
+			return (-1);
```
agorrod: What does ```-1``` mean? You return it in a number of places. We already have a ```ret``` variable with more context - you should return that instead.

There are other places with similar style that discard a return from ```enomem```, then return ```-1```. Those should ```return (enomem(cfg));``` instead.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+		}
+		if ((ret = cursor->get_key(cursor,
+			    &key)) != 0) {
+			lprintf(cfg, ret, 0,
+			    "truncate setup end: get_key failed");
+			return (-1);
+		}
+		end_point_val = decode_key(cfg, key);
+		/* Not enough documents? */
+			if (start_point_val + trunc->needed_milestones
```
agorrod: The indentation is wrong here, and the ```>``` is still on the wrong line.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+				    (end_point_val - start_point_val)
+				    / trunc->needed_milestones;
+	} else if ( ret != WT_NOTFOUND ) {
+		lprintf(cfg, ret, 0,
+			    "truncate setup start: get_key failed");
+		printf("ret was %d\n", ret);
+		return (-1);
+	}
+
+	/* If we have enough data alloc some milestones */
```
agorrod: ```alloc``` should be ```allocate``` - always prefer full words in comments.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+
+	char *truncate_key;
+	int ret;
+	size_t i;
+	TRUNCATE_QUEUE_ENTRY *truncate_item;
+
+	/* Update the total inserts */
+	trunc->total_gross_inserts = 0;
+	for (i=0; i < thread->cfg->workers_cnt; i++) {
+		trunc->total_gross_inserts +=
```
agorrod: Single line within a ```for``` loop doesn't need ```{}```

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,173 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+		    thread->cfg->workers[i].total_inserts;
+	}
+
+	trunc->expected_total +=
+	    (trunc->total_gross_inserts - trunc->last_total_inserts);
+	trunc->last_total_inserts = trunc->total_gross_inserts;
+
+	/* We have enough data, so we can look to truncate */
```
agorrod: This code is setting up the milestones, not doing the truncate. I find the comment confusing.

agorrod: Oh, ignore that. I see it's doing truncates as well.

bench/wtperf/wtperf.c
```diff
@@ -2091,6 +2333,14 @@ main(int argc, char *argv[])
 	if ((ret = config_compress(cfg)) != 0)
 		goto err;
 
+	/* You can't have truncate on a random collection */
+	if (cfg->has_truncate && cfg->random_range)
+		goto err;
```
agorrod: It would be valuable to output error messages in these two cases. It's not obvious from the API.

agorrod:
@daveh86 Could you add another (pair of) wtperf configuration file for testing the truncate. I'd like to be able to populate a table in one pass (one configuration file), then have a workload file that can be run repeatably on the same data set (another configuration file).

ie, the commands will be:

```
./bench/wtperf/wtperf -O ../bench/wtperf/runners/truncate-btree-populate.wtperf
./bench/wtperf/wtperf -O ../bench/wtperf/runners/truncate-btree-workload.wtperf
```

bench/wtperf/wtperf.c
```diff
@@ -382,6 +566,8 @@ worker(void *arg)
 	CONFIG *cfg;
 	CONFIG_THREAD *thread;
 	TRACK *trk;
+	TRUNCATE_QUEUE_ENTRY *truncate_item;
+	TRUNCATE_CONFIG *truncate_config;
```
agorrod: This can be a stack variable. i.e, declare it as:

```
TRUNCATE_CONFIG truncate_config;
```

Then you don't need to malloc it (or free it), and you can pass it as a parameter with ```&truncate_config```

bench/wtperf/runners/truncate-btree-populate.wtperf
```diff
@@ -0,0 +1,7 @@
+# Truncate workload population phase
+conn_config="cache_size=2GB,checkpoint=(wait=60)"
+table_config="type=file"
+# Start with a small set of inserts in the populate phase.
+icount=50000
+report_interval=5
+populate_threads=1
```
agorrod: Missing newline at the end of the file.

bench/wtperf/wtperf.c
```diff
@@ -2091,6 +2335,18 @@ main(int argc, char *argv[])
 	if ((ret = config_compress(cfg)) != 0)
 		goto err;
 
+	/* You can't have truncate on a random collection */
+	if (cfg->has_truncate && cfg->random_range) {
+		fprintf(stderr, "Cannot run truncate and random_range\n");
```
agorrod: This should use the wtperf ```lprintf``` error logging function. Similarly to other occurrences of printf and fprintf in this diff.

bench/wtperf/runners/truncate-btree-workload.wtperf
```diff
@@ -0,0 +1,10 @@
+# truncate workload. work phase
+conn_config="cache_size=2GB,checkpoint=(wait=60)"
+table_config="type=file"
+create=false
+# Start with a small set of inserts in the populate phase.
```
agorrod: This configuration file doesn't have a populate phase.

bench/wtperf/runners/truncate-btree-workload.wtperf
```diff
@@ -0,0 +1,10 @@
...
+report_interval=5
+run_time=500
+# Setup three threads to insert into the oplog (will currently work)
+# Setup one thread to be doing truncates from the oplog (not implemented yet)
```
agorrod: The parts of this comment in parentheses are no longer necessary. Similarly for other config files.

bench/wtperf/wtperf.c
```diff
@@ -109,6 +112,18 @@ generate_key(CONFIG *cfg, char *key_buf, uint64_t keyno)
 	sprintf(key_buf, "%0*" PRIu64, cfg->key_sz - 1, keyno);
 }
 
+static inline uint64_t
+decode_key(CONFIG *cfg, char *key_buf)
+{
+	char *endPtr;
+	uint32_t i;
```
agorrod: There should be a blank line after the local variable declarations.

bench/wtperf/wtperf.c
```diff
@@ -109,6 +112,18 @@ generate_key(CONFIG *cfg, char *key_buf, uint64_t keyno)
...
+	i= 0;
```
agorrod: Why isn't this included in the for loop?

bench/wtperf/wtperf.c
```diff
@@ -109,6 +112,18 @@ generate_key(CONFIG *cfg, char *key_buf, uint64_t keyno)
...
+	for (;i < cfg->key_sz; i++)
```
agorrod: Is this loop necessary? Does strtoull already deal with leading `0` characters? Do we get key inputs with leading ```NULL``` characters?

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,177 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+		ret = cursor->get_key(cursor, &key);
+	}
+	/* We have data */
+	if (ret == 0) {
+		start_point_val = decode_key(cfg, key);
+		cursor->reset(cursor);
+		if ((ret = cursor->prev(cursor)) != 0) {
+			lprintf(cfg, ret, 0,
+			    "truncate setup end: prev failed");
+			return (ret);
```
agorrod: Each return after the cursor is opened should be replaced with a ```goto err;```, and the error cleanup should close the cursor.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,177 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+		}
+		if ((ret = cursor->get_key(cursor,
```
agorrod: This line wrapping doesn't look right. Does it need to be wrapped? If so the two parameters should be on the second line.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,177 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+			    &key)) != 0) {
+			lprintf(cfg, ret, 0,
+			    "truncate setup end: get_key failed");
+			return (ret);
+		}
+		end_point_val = decode_key(cfg, key);
+		/* Not enough documents? */
+			if (start_point_val + trunc->needed_milestones >
```
agorrod: Indentation problem.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,177 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+			    end_point_val)
+				trunc->truncate_milestone_gap = 0;
+			else
+				trunc->truncate_milestone_gap =
+				    (end_point_val - start_point_val)
+				    / trunc->needed_milestones;
+	} else if ( ret != WT_NOTFOUND ) {
```
agorrod: Spacing is wrong.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,177 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+		lprintf(cfg, ret, 0,
+			    "truncate setup start: get_key failed");
```
agorrod: Indentation is wrong.

bench/wtperf/wtperf.c
```diff
@@ -131,6 +146,177 @@ randomize_value(CONFIG_THREAD *thread, char *value_buf)
...
+}
+
+static int
+run_truncate(CONFIG *cfg, CONFIG_THREAD *thread, TRACK **trk,
+    TRUNCATE_CONFIG *trunc, WT_CURSOR *cursor, WT_SESSION *session) {
+
+	char *truncate_key;
+	int ret;
+	size_t i;
+	TRUNCATE_QUEUE_ENTRY *truncate_item;
```
agorrod: Ordering.

bench/wtperf/wtperf.c
```diff
@@ -456,6 +647,14 @@ worker(void *arg)
 		goto err;
 	}
 
+	/* Setup for truncate */
+	if (thread->workload->truncate != 0) {
+		cursor = cursors[cfg->icount % cfg->table_count];
```
agorrod: I don't think you need to retrieve the cursor any more.

bench/wtperf/wtperf.c
```diff
@@ -1420,16 +1651,19 @@ execute_workload(CONFIG *cfg)
...
+	uint64_t last_updates;
 	uint32_t interval, run_ops, run_time;
 	u_int i;
 	int ret, t_ret;
 	void *(*pfunc)(void *);
 
 	cfg->insert_key = 0;
-	cfg->insert_ops = cfg->read_ops = cfg->update_ops = 0;
+	cfg->insert_ops = cfg->read_ops =
+	    cfg->truncate_ops = cfg->update_ops = 0;
```
agorrod: We don't split lines for local variable initialization, this would become:

```
cfg->insert_ops = cfg->read_ops = cfg->update_ops = 0;
cfg->update_ops = 0;
```

bench/wtperf/wtperf.c
```diff
@@ -1454,14 +1688,18 @@ execute_workload(CONFIG *cfg)
...
-		    workp->threads, workp->insert, workp->read, workp->update);
+		    PRId64 ", reads=%" PRId64 ", updates=%" PRId64
+		    ", truncate=%" PRId64, i + 1, workp->threads, workp->insert,
+		    workp->read, workp->update, workp->truncate);
 
 		/* Figure out the workload's schedule. */
 		if ((ret = run_mix_schedule(cfg, workp)) != 0)
 			goto err;
 
+		/* Allocate the truncate threads queue */
```
agorrod: ```Setup``` would be a better word than ```Allocate``` - since allocate infers memory management.

bench/wtperf/wtperf.c
```diff
@@ -2068,8 +2309,11 @@ main(int argc, char *argv[])
 	 * If the user wants compaction, then we also enable async for
 	 * the compact operation, but not for the workloads.
 	 */
-	if (cfg->async_threads > 0)
+	if (cfg->async_threads > 0) {
+		if (cfg->has_truncate > 0)
```
agorrod: Add an error message indicating what the problem is.

daveh86:
@agorrod, i've gone over this and picked up a bunch of little nits in addition to addressing your last comments.

bench/wtperf/wtperf.h
```diff
@@ -223,6 +256,10 @@ struct __config_thread {		/* Per-thread structure */
 	TRACK insert;			/* Insert operations */
 	TRACK read;			/* Read operations */
 	TRACK update;			/* Update operations */
+	TRACK truncate;			/* Truncate operations */
+	TRACK truncate_sleep;		/* Truncate sleep operations */
+
+	uint64_t total_inserts;		/* Inserts done by this thread */
```
agorrod: Why is this different to the ```insert``` field? I believe that the ```TRACK insert``` field already has the behavior you want - i.e: A call to ```sum_insert_ops``` could replace tracking ```total_inserts```

agorrod: p.s: The ```TRACK``` code in ```wtperf``` is a bit esoteric, but it's worth taking the time to figure out what it's doing.

agorrod:
@daveh86 I've pushed a few cleanup changes - could you please take a look at the changes I made, and see if similar changes could be made elsewhere?

Could you please also try testing with different configurations? Including configurations where you'd expect error paths to be taken? For example if you setup a configuration with a 0 size populate phase.

daveh86:
@agorrod, i've made a number of changes along the lines you have suggested. I've also tested with 5 different configurations for:
# When truncate is a non 1 value
# Truncate + async
# Truncate + more than one table
# Truncate + random inserts
# icount of 1
# various different values for truncate count and truncate %

bench/wtperf/wtperf.h
```diff
@@ -251,3 +296,14 @@ void	 lprintf(const CONFIG *, int err, uint32_t, const char *, ...)
 __attribute__((format (printf, 4, 5)))
 #endif
 ;
+
+static inline void
```
agorrod: @keithbostic Can you take a look at this function declaration and let me know if this will be inlined properly? The inline semantics confuse me.

No need to look at the rest of the changes.

Thanks,
Alex

keithbostic: Looks fine to me -- is there something in particular that you're concerned about?

agorrod: Thanks. Nope - I just have difficulty with inline semantics and felt like a sanity check was worthwhile.

bench/wtperf/wtperf.c
```diff
@@ -712,26 +729,35 @@ run_mix_schedule(CONFIG *cfg, WORKLOAD *workp)
 	int64_t pct;
 
 	/* Confirm reads, inserts and updates cannot all be zero. */
```
agorrod: The comment is out of date.

bench/wtperf/wtperf_truncate.c
```diff
@@ -0,0 +1,194 @@
...
+
+	if ((ret = session->truncate(session, NULL, NULL, cursor, NULL)) != 0) {
+		lprintf(cfg, ret, 0, "Truncate: failed");
+		return (ret);
+	}
+	*truncatedp = 1;
+	trunc_cfg->expected_total -= truncate_item->diff;
+	free(truncate_item->key);
+	free(truncate_item);
+	truncate_item = NULL;
```
agorrod: You should ensure that the cursor has been reset prior to returning from this function if anything has been done with the cursor.

bench/wtperf/wtperf.c
```diff
@@ -753,14 +785,18 @@ run_mix_schedule(CONFIG *cfg, WORKLOAD *workp)
 	memset(workp->ops, WORKER_READ, sizeof(workp->ops));
 
 	pct = (workp->insert * 100) /
-	    (workp->insert + workp->read + workp->update);
+	    (workp->insert + workp->read + workp->truncate + workp->update);
```
agorrod: Do you need to factor truncate into these three calculations? I thought specifying truncate for a thread excluded any other operation types.

agorrod:
@daveh86 I pushed a couple of changes - take a look and let me know if you don't understand why I made them.

I've also made a couple of other comments - please address them.

agorrod:
@daveh86 I was just running the following wtperf configuration:

```
conn_config="cache_size=5GB,checkpoint=(wait=60)"
table_config="type=file"
icount=50000
report_interval=5
run_time=5000
populate_threads=1
threads=((count=8,inserts=1,throttle=20000),(count=1,truncate=1,truncate_pct=10,truncate_count=100000000))
```

and I got the following error:

```
Truncate: failed Error: WT_ROLLBACK: conflict between concurrent operations
truncate failed for: 0000000000142024825, range: 152999245 Error: WT_ROLLBACK: conflict between concurrent operations
Worker thread(s) exited without finishing. Error: WT_ERROR: non-specific WiredTiger error
```

Is that a surprise?

bench/wtperf/wtperf.c
daveh86: I have not currently excluded truncate from running alongside another workload.

If you think this limitation is worthwhile, I will add it.

agorrod: If you setup a worker thread to do both truncate and other operations, won't the sleeps in the truncate code cause the workload to behave strangely?

daveh86: I follow. I've added logic to make the truncate thread a single exclusive thread within its own workload.

agorrod:
@daveh86 I've reproduced this once after the latest merge from develop. The config was slightly different:

```
conn_config="cache_size=5GB,checkpoint=(wait=60)"
table_config="type=file"
icount=50000
report_interval=5
run_time=5000
populate_threads=1
threads=((count=16,inserts=1),(count=1,truncate=1,truncate_pct=33,truncate_count=100000000))
```

agorrod:
Interestingly the error message I saw was:

```
Truncate: failed Error: WT_ROLLBACK: conflict between concurrent operations
truncate failed for: 0000000000021086685, range: 141431924 Error: WT_ROLLBACK: conflict between concurrent operations
```

The truncate key is 21 million, the range is 141 million. I wonder if truncate is deciding to remove all the data for some reason? I think it's worth checking the wtperf key formatting - since decoding/re-encoding is new to the truncate code.

agorrod:
My last comment was wrong. I'd assumed that error message came from WiredTiger, but it came from within wtperf. The range referenced in that error message is the total range of values ever inserted by wtperf.

Looking at the code, it makes me realise that truncate should be being reflected in ```wtperf_value_range```. Truncate is the first place I'm aware of where wtperf doesn't start with a key range of zero. I think range will need to move from a single max return to a min and max pair.

agorrod:
Thanks @daveh86 I've merged this now - well done.

daveh86:
:confetti_ball:

michaelcahill:
+1 great work finalising this change. @daveh86.

agorrod:
@daveh86 Looks like there is a test failure related to this change: http://build.wiredtiger.com:8080/job/wiredtiger-gcc47-warnings/3640/console

Could you follow up please?

PR #2034, opened 2015-06-30
Add ability to turn verify and salvage off in test/format.

agorrod:
This can be handy when reproducing failures that aren't uncovered
with verify/salvage, since it decreases the runtime for a single pass.


michaelcahill:
lgtm pending tests.

PR #2035, opened 2015-06-30
WT-1982 Track down cached overflow items that are being freed too early

michaelcahill:
... leading to reconciliation failing with WT_NOTFOUND.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -105,8 +105,7 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
 {
 	WT_BTREE *btree;
 	WT_TXN_GLOBAL *txn_global;
-	uint64_t checkpoint_snap_min, oldest_id;
-	uint32_t checkpoint_id;
+	uint64_t checkpoint_pinned, oldest_id;
 
 	txn_global = &S2C(session)->txn_global;
 	btree = S2BT_SAFE(session);
@@ -115,8 +114,7 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
 	 * Take a local copy of these IDs in case they are updated while we are
 	 * checking visibility.
 	 */
-	checkpoint_id = txn_global->checkpoint_id;
-	checkpoint_snap_min = txn_global->checkpoint_snap_min;
+	checkpoint_pinned = txn_global->checkpoint_pinned;
 	oldest_id = txn_global->oldest_id;
 
 	/*
@@ -129,17 +127,17 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
 	 * checkpoint, or this handle is up to date with the active checkpoint
 	 * then it's safe to ignore the checkpoint ID in the visibility check.
 	 */
-	if (checkpoint_snap_min != WT_TXN_NONE &&
-	    checkpoint_id != session->id && (btree == NULL ||
-	    btree->checkpoint_gen != txn_global->checkpoint_gen) &&
-	    WT_TXNID_LT(checkpoint_snap_min, oldest_id))
-		/*
-		 * Use the checkpoint ID for the visibility check if it is the
-		 * oldest ID in the system.
-		 */
-		oldest_id = checkpoint_snap_min;
+#if 0
+	if (WT_SESSION_IS_CHECKPOINT(session))
+                return (oldest_id);
+#endif
+	if (checkpoint_pinned == WT_TXN_NONE ||
+	    WT_TXNID_LT(oldest_id, checkpoint_pinned) ||
+	    (btree != NULL &&
+	    btree->checkpoint_gen == txn_global->checkpoint_gen))
+                return (oldest_id);
 
-	return (oldest_id);
+	return (checkpoint_pinned);
 }
 
 /*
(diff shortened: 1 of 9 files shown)


src/include/txn.i
```diff
@@ -129,17 +127,17 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
...
-		 * Use the checkpoint ID for the visibility check if it is the
-		 * oldest ID in the system.
-		 */
-		oldest_id = checkpoint_snap_min;
-
-	return (oldest_id);
+#if 0
+	if (WT_SESSION_IS_CHECKPOINT(session))
+                return (oldest_id);
+#endif
```
michaelcahill: This is diagnostic: I would like to turn this on, but doing so makes the bug trigger for me.

michaelcahill:
This is for discussion and to aid debugging only.  There are some cleanups here that I like, but I don't think the underlying bug is fixed properly yet.

agorrod:
For the record, I've had 50 clean runs with the ```#define``` in ```txn.i``` set to both 0 and 1.

agorrod:
I can get it to fail with the conditional code in ```txn.i``` enabled, still haven't seen a failure without that enabled.

agorrod:
@michaelcahill I've had 50 good runs with your latest change. Have you seen a failure with the latest version of this branch?

michaelcahill:
I think the final version is correct.  Apart from sloppiness around setting of `checkpoint_snap_min` and `checkpoint_id`, the main problem was that there was a window after we turned off `btree->checkpointing` and before updating `btree->checkpoint_gen`.

michaelcahill:
@agorrod can you please review?

I've tried to comment where there is trickiness, but I'm happy to add more so we don't hit this kind of problem again.  Also let me know if you would like me to work up a minimal version of the fix for testing.

src/txn/txn.c
```diff
@@ -157,12 +156,7 @@ __wt_txn_get_snapshot(WT_SESSION_IMPL *session)
 
 	/* Walk the array of concurrent transactions. */
 	WT_ORDERED_READ(session_cnt, conn->session_cnt);
-	ckpt_id = txn_global->checkpoint_id;
 	for (i = n = 0, s = txn_global->states; i < session_cnt; i++, s++) {
-		/* Skip the checkpoint transaction; it is never read from. */
```
agorrod: To be sure I understand.. Do we no longer need to skip the checkpoint transaction because it isn't setting a transaction ID in the regular ID table any more? It now only sets it in the ```txn_global->checkpoint_pinned``` field?

src/include/txn.i
```diff
@@ -115,9 +115,9 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
 	 * Take a local copy of these IDs in case they are updated while we are
 	 * checking visibility.
 	 */
-	checkpoint_id = txn_global->checkpoint_id;
-	checkpoint_snap_min = txn_global->checkpoint_snap_min;
-	oldest_id = txn_global->oldest_id;
+	WT_ORDERED_READ(oldest_id, txn_global->oldest_id);
+	WT_ORDERED_READ(checkpoint_gen, txn_global->checkpoint_gen);
```
michaelcahill: I think it would be sufficient to just do a single `WT_ORDERED_READ` here (on `checkpoint_gen`), but we should think that through.

agorrod:
> you would like me to work up a minimal version of the fix for testing

That isn't necessary, I've spent long enough staring at this one that I'm confident I can pick out the bug fix.

src/include/txn.i
agorrod: I think that is right. I think ```checkpoint_gen``` can be a problem since it's used to determine between using ```oldest_id``` and ```checkpoint_pinned```. I don't think ordering is important for the other two - worst case we'll read an ID that is a bit (i.e: older than absolutely required).

src/txn/txn.c
michaelcahill: That's right: instead of making every scan of `txn_global->states` check if each slot is doing a checkpoint, I've changed the checkpoint transaction to clear out its slot once it sets `txn_global->checkpoint_pinned` (and associated state).

src/txn/txn_ckpt.c
```diff
@@ -436,30 +454,37 @@ __wt_txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
 	txn_global->checkpoint_id = session->id;
-	txn_global->checkpoint_snap_min = session->txn.snap_min;
+
+	txn_global->checkpoint_pinned =
+	    WT_MIN(txn_state->id, txn_state->snap_min);
 
 	/*
-	 * No need for this to be atomic it is only written while holding the
-	 * checkpoint lock.
+	 * We're about to clear the checkpoint transaction from the global
```
agorrod: A bit more information about why we are clearing the checkpoint ID would be useful.

agorrod:
Thanks @michaelcahill I've requested one more verbose comment. Apart from that I'm happy with this change (lgtm). If you update the comment I can merge the branch tonight - and we can triple-check it via stress testing.

PR #2036, opened 2015-07-01
WT-1985 integer packing fixes for Python

ddanderson:
Here are some fixes along with some tests for python.


ddanderson:
@agorrod or @michaelcahill, this is ready for review.

There's a performance oddity that I don't have a handle on.  If I modify:
```
        pt.check_range(-10000, 10000)
```
to be `-100000, 100000` there's a big slowdown partway through the test.  Bumping up the cache doesn't help - it's not disk bound or cpu bound. I can look further at that tomorrow and open a ticket.

I think I'll make Java changes into a separate pull request.

agorrod:
Thanks @ddanderson lgtm - I'll merge. It would be good if you can investigate the performance issue.

I also noticed that there are two test files in ```lang/python/wiredtiger``` related to packing ```intpack-test.py``` and ```packing-test.py```. Does your new test case make those obsolete? If not could you move them into the test/suite directory, and make them conform with our other test cases?

PR #2038, opened 2015-07-01
track the tree's max-txn-ID in reconciliation, where the page's max-txn-ID is set.

keithbostic:
@michaelcahill, I was staring at this code this morning, and it seemed like it would be easier (safer?) to set the tree's maximum transaction ID when we set/update the page's maximum transaction ID.

The "safer" part may be a misunderstanding on my part: the WT_BTREE.rec_max_txn field is used by sweep and close, but it's only set as part of the checkpoint operation.

But what if, by the time we get to checkpoint (WT_SYNC_CHECKPOINT), there aren't any leaf pages to write. Is it possible we won't update the tree's maximum transaction ID, since internal pages don't have a maximum transaction ID?

### DIFF
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -109,17 +109,6 @@ __sync_file(WT_SESSION_IMPL *session, int syncop)
 		/* Write all dirty in-cache pages. */
 		flags |= WT_READ_NO_EVICT;
 		for (walk = NULL;;) {
-			/*
-			 * If we have a page, and it was ever modified, track
-			 * the highest transaction ID in the tree.  We do this
-			 * here because we want the value after reconciling
-			 * dirty pages.
-			 */
-			if (walk != NULL && walk->page != NULL &&
-			    (mod = walk->page->modify) != NULL &&
-			    WT_TXNID_LT(btree->rec_max_txn, mod->rec_max_txn))
-				btree->rec_max_txn = mod->rec_max_txn;
-
 			WT_ERR(__wt_tree_walk(session, &walk, NULL, flags));
 			if (walk == NULL)
 				break;
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -5096,7 +5096,13 @@ err:			__wt_scr_free(session, &tkey);
 		btree->modified = 1;
 		WT_FULL_BARRIER();
 	} else {
+		/*
+		 * Set the highest transaction ID for the page, and track the
+		 * highest transaction ID for the tree.
+		 */
 		mod->rec_max_txn = r->max_txn;
+		if (WT_TXNID_LT(btree->rec_max_txn, r->max_txn))
+			btree->rec_max_txn = r->max_txn;
 
 		if (WT_ATOMIC_CAS4(mod->write_gen, r->orig_write_gen, 0))
 			__wt_cache_dirty_decr(session, page);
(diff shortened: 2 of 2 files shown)


michaelcahill:
@keithbostic, I don't think there is actually a bug here, because we only care about `btree->rec_max_txn` on clean trees after a checkpoint has been through (where checkpoint has left some update in memory).  When we consider pages for eviction, they will either be evicted (in which case we don't care about the highest transaction ID that used to be on them), or left dirty in cache, so they will be visited by a future checkpoint before we care about `btree->rec_max_txn`.

My only concern about merging this change is racing: if we were reconciling two pages in a tree around the same time, they could race to update `btree->rec_max_txn`, potentially losing the larger transaction ID.  That shouldn't matter for the reason described above (checkpoint will visit any dirty page before we care about this value), but it makes me nervous.

michaelcahill:
P.S.  I guess we could wrap the update of `btree->rec_max_txn` in `if (!LF_ISSET(WT_EVICTING))`?

keithbostic:
@michaelcahill, first, my bad -- there is definitely a race here.

Is `LF_ISSET(WT_EVICTING)` enough? We could have N eviction worker threads evicting pages in the same tree?

We could switch to an atomic update, but yuck.

keithbostic:
@michaelcahill, second, I'm still not clear on why this is safe, there's something I'm missing. In these two tests:
```
__sweep_expire_handle():
    /* Only sweep clean trees where all updates are visible. */
    if (btree->modified ||
        !__wt_txn_visible_all(session, btree->rec_max_txn))
            goto err;

__wt_checkpoint_close():
    /*
     * If closing an unmodified file, check that no update is required
     * for active readers.
     */
    if (!btree->modified && !bulk) {
            __wt_txn_update_oldest(session, 1);
            return (__wt_txn_visible_all(session, btree->rec_max_txn) ?
                __wt_cache_op(session, NULL, WT_SYNC_DISCARD) : EBUSY);
    }
```

If we have a tree where all of the leaf pages are written during the `WT_SYNC_WRITE_LEAVES` pass, I think the tree is clean, but `btree->rec_max_txn` won't ever have been set, and the tree could be closed while there are still be active readers looking for updates left in the tree.

michaelcahill:
The write leaves phase will skip pages that have any updates that are not globally visible, so it should skip pages that would be problematic for discarding the tree.  However, that check just before calling `__wt_reconcile` could race with a page becoming dirty before `__rec_write_init` takes a copy of `page->modify->write_gen`, so I agree that there is a window.

I think that checking `!LF_ISSET(WT_EVICTING)` is sufficient (note the "!"): i.e., only set `btree->rec_max_txn` in checkpoint operations because they are single-threaded.

keithbostic:
If we were to update a bunch of pages in the file, then evict the ones with higher maximum txn IDs, then checkpoint the rest; doesn't that lead to an incorrect (low) value as our tree maximum txn ID?

michaelcahill:
Yes, and that’s fine.  What we’re tracking here is the highest transaction ID of an update that may still be in cache for a tree.  If eviction succeeded, then by definition we don’t need to keep that page in cache, so it shouldn’t keep the tree in cache.

The field is probably poorly named: it is set based on the highest transaction ID seen during a checkpoint (which is where the name comes from), but it is used to determine whether it is safe to discard all of the pages in a tree without further checking.

keithbostic:
@michaelcahill: I've made that change and pushed it, I think this is ready to merge if/when you're happy.

michaelcahill:

PR #2042, opened 2015-07-03
WT-1992: clean up metadata tracking error handling

keithbostic:
@michaelcahill, @agorrod: for review and merge.

### DIFF
--- a/src/meta/meta_track.c
+++ b/src/meta/meta_track.c
@@ -412,7 +412,6 @@ __wt_meta_track_update(WT_SESSION_IMPL *session, const char *key)
 	WT_ERR(ret);
 	return (0);
 
-
 err:	__meta_track_err(session);
 	return (ret);
 }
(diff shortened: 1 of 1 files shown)


michaelcahill:

src/meta/meta_track.c
```diff
@@ -67,6 +67,35 @@ __meta_track_next(WT_SESSION_IMPL *session, WT_META_TRACK **trkp)
...
 /*
+ * __meta_track_clear --
+ *	Clear the structure.
+ */
+static void
+__meta_track_clear(WT_SESSION_IMPL *session, WT_META_TRACK *trk)
+{
+	__wt_free(session, trk->a);
+	__wt_free(session, trk->b);
+	memset(trk, 0, sizeof(WT_META_TRACK));
```
agorrod: This assumes that ```WT_ST_EMPTY``` is 0 (which it currently is). I'd rather make that explicit in the ```enum``` if the code requires it.

PR #2043, opened 2015-07-06
WT-1988. Added a doc subsection about the interation between cursors and eviction

ddanderson:
, @keithbostic, can you please review?


src/docs/cursors.dox
```diff
@@ -72,6 +72,26 @@ well as any key and value.
...
+active cursors on a memory page that has reached maximum size, the
+WiredTiger library will never have simultaneous control over both cursors to
+move them to smaller, split pages.  This may prevent a large page from ever
+being split and may potentially slow down the application.
+
+To avoid this situation, or any situation where eviction must be deferred,
+applications should call WT_CURSOR::reset during times that it does not need
+to keep a cursor positioned.  A cursor that has been reset is not active and
+cannot inhibit eviction.
+
```
keithbostic: @ddanderson, can we skip the details of eviction? That is, does it matter what kind of eviction is being inhibited by cursors? It seems like unnecessary detail to me, and likely to lead to the wording becoming incorrect as the system evolves.

ddanderson:
@keithbostic, Thanks, I've pushed some more general wording on this, can you take another look?

PR #2044, opened 2015-07-07
WT-1930 Add configuration option to disable sweep of old handles

daveh86:


daveh86:
@agorrod, would you be able to have a look at this?

dist/api_data.py
```diff
@@ -402,7 +402,8 @@ def __cmp__(self, other):
             to close''', min=0),
         Config('close_idle_time', '30', r'''
             amount of time in seconds a file handle needs to be idle
-            before attempting to close it''', min=1, max=100000),
+            before attempting to close it. A setting of 0 will mean that older
+            handles are not closed''', min=0, max=100000),
```
agorrod: replace ```will mean``` with ```means```. Replace ```older``` with ```idle```.

agorrod:
@daveh86 I'm happy to talk about the implementation once you have made the documentation changes.

daveh86:
@agorrod , mind having a look at the latest changes?

test/suite/test_sweep03.py
```diff
@@ -0,0 +1,136 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# test_sweep01.py
```
agorrod: Should be sweep03

test/suite/test_sweep03.py
```diff
@@ -0,0 +1,136 @@
...
+# Test lots of tables, number of open files and sweeping.  Run both
```
agorrod: This description should be updated to reflect the test.

test/suite/test_sweep03.py
```diff
@@ -0,0 +1,136 @@
...
+# with and without checkpoints.
+#
+
+import fnmatch, os, shutil, run, time
+from suite_subprocess import suite_subprocess
+from wiredtiger import wiredtiger_open, stat
+from wtscenario import multiply_scenarios, number_scenarios, prune_scenarios
+import wttest
+
+class test_sweep01(wttest.WiredTigerTestCase, suite_subprocess):
```
agorrod: Should be sweep03, not sweep01

test/suite/test_sweep03.py
```diff
@@ -0,0 +1,136 @@
...
+        ('fix', dict(tabletype='fix',
+                    create_params = 'key_format=r,value_format=8t')),
+    ]
+
+    scenarios = types
+
+    # Overrides WiredTigerTestCase
+    def setUpConnectionOpen(self, dir):
+        self.home = dir
+        self.backup_dir = os.path.join(self.home, "WT_BACKUP")
```
agorrod: Is this necessary?

agorrod:
@daveh86 The code looks OK to me. I think the test could use some work - it would be good to strip out functionality that isn't necessary for the particular case you are targeting (checkpoints, for example).

There is one other thing I'd like to test, that dropped handles are being cleaned up. To test that I'd do:

* Create a table
* Add a lot of data to that table.
* Get statistics for cache usage.
* Drop the table.
* Wait until sweep has done at least one pass
* Get statistics for cache usage.

The second statistics should show a reduction in cache usage.

agorrod:
@daveh86 There is a review comment outstanding, and I'd like to add the cache usage checking I described in the last comment to the test.

I'd like to do the cache checking for both forced and non-forced drops.

src/conn/conn_sweep.c
```diff
@@ -139,7 +140,8 @@ __sweep_expire(WT_SESSION_IMPL *session)
 		    F_ISSET(dhandle, WT_DHANDLE_DEAD))
 			continue;
 		if (dhandle->session_inuse != 0 ||
-		    now <= dhandle->timeofdeath + conn->sweep_idle_time)
+		    now <= dhandle->timeofdeath + conn->sweep_idle_time ||
+		    conn->sweep_idle_time == 0)
```
agorrod: I'd move this test to early in ```__sweep_expire```. The function is a no-op if we aren't expiring handles.

I'd change this to be:

```
/* Never expire handles if idle timeout is disabled. */
if (conn->sweep_idle_time == 0)
    return (0);
```

Just before the call to ```__wt_seconds```.

src/conn/conn_sweep.c
```diff
@@ -266,7 +268,8 @@ __sweep_server(void *arg)
 		WT_ERR(__sweep_mark(session, &dead_handles));
 
 		if (dead_handles == 0 &&
-		    conn->open_file_count < conn->sweep_handles_min)
+		    conn->open_file_count < conn->sweep_handles_min &&
```
agorrod: I think this should be:

```
if (dead_handles == 0 &&
    (conn->open_file_count < conn->sweep_handles_min ||
    conn->sweep_idle_time != 0))
```

i.e: We never want to skip closing handles if we see any dead handles.
We want to skip closing handles if either there aren't enough open, or we've disabled closing idle handles.

The test case should be expanded to capture this:
Setting ```sweep_handles_min``` relatively low, opening and closing lots of handles, and ensuring they never get closed if ```sweep_idle_time``` is set to 0.

src/conn/conn_sweep.c
```diff
@@ -121,6 +122,9 @@ __sweep_expire(WT_SESSION_IMPL *session)
 
 	conn = S2C(session);
 
+	if (conn->sweep_idle_time == 0)
```
agorrod: Add a comment - it's not obvious why an idle time of 0 means there is no work to do.

src/conn/conn_sweep.c
```diff
@@ -266,7 +270,8 @@ __sweep_server(void *arg)
 		WT_ERR(__sweep_mark(session, &dead_handles));
 
 		if (dead_handles == 0 &&
-		    conn->open_file_count < conn->sweep_handles_min)
+		    (conn->open_file_count < conn->sweep_handles_min ||
```
agorrod: Add a comment explaining the conditions in this statement. It's subtle enough that I had to think quite hard about it.

agorrod:
@daveh86 Did the latest version of the test fail before the last change to the ```__sweep_server```? I'd like to be confident that different combinations of configuration options work reliably via Python test suite.

agorrod:
Thanks @daveh86 I made some changes to the Python test case. This lgtm now, I'll merge.

PR #2045, opened 2015-07-07
WT-1970 Speedup s_all

agorrod:
These changes make running ```s_all``` faster by parallelising scripts. Here are the results I saw:

| Machine | develop | speedup-s_all |
|---|---|---|
| tinderbox | 52 | 10 |
| bengal | 40 | 14 |
| AWS box | 60 | 15 |
| Mac | 62 | 40 |

@michaelcahill Could you take a look please? It changes the output from ```s_all``` - it now only reports errors.


dist/s_all
```diff
@@ -51,34 +55,49 @@ test "$reconf" -eq 0 || {
...
+'sh ./s_funcs > ${t_pfx}s_funcs'
+'sh ./s_export > ${t_pfx}s_export'
+'sh ./s_getopt > ${t_pfx}s_getopt'
+'sh ./s_label > ${t_pfx}s_label'
+'sh ./s_lang > ${t_pfx}s_lang'
+'sh ./s_longlines > ${t_pfx}s_longlines'
+'sh ./s_stat > ${t_pfx}_tat'
+'sh ./s_string > ${t_pfx}s_string'
+'python style.py > ${t_pfx}py_style'
+'sh ./s_style > ${t_pfx}s_style'
```
michaelcahill: Since s_style is multi-process internally, it seems a little odd to run it in parallel here: now we're up to ~40 processes running in parallel, and if we do the same to any more of these scripts, that number will grow.  Maybe s_style should be run serially, after the other stages?

michaelcahill: Actually, s_style also rewrites files to fix certain formatting problems automatically.  I think that's another reason for not running it in parallel with other tests.

dist/s_all
```diff
@@ -51,34 +55,50 @@ test "$reconf" -eq 0 || {
...
+'sh ./s_tags > ${t_pfx}tags'
+'sh ./s_define > ${t_pfx}s_define'
+'sh ./s_typedef -c > ${t_pfx}s_typedef_c'
+'sh ./s_funcs > ${t_pfx}s_funcs'
+'sh ./s_export > ${t_pfx}s_export'
+'sh ./s_getopt > ${t_pfx}s_getopt'
+'sh ./s_label > ${t_pfx}s_label'
+'sh ./s_lang > ${t_pfx}s_lang'
+'sh ./s_longlines > ${t_pfx}s_longlines'
+'sh ./s_stat > ${t_pfx}_tat'
```
sueloverso: _tat should be _stat.

michaelcahill:
@agorrod, you've done `s_copyright` as a single re-entrant script and split `s_style` in two.  Is there a reason for the difference?

michaelcahill:
@keithbostic, I like these changes and they help a lot on Linux.  Do you feel strongly about the script changes or running silent on success?

agorrod:
> s_copyright as a single re-entrant script and split s_style in two.

There is no good reason - ```s_style``` was originally more complex, but it would be simple to merge it back together now.

I'm happy to make them consistent, and choose whichever style is preferred.

michaelcahill:
I prefer having the logic in one file (i.e., the {{s_copyright}} style).

keithbostic:
@michaelcahill, This looks great to me (Thanks, Alex!).

I'm OK with silent running -- I'd prefer verbose because the scripts still take awhile to run, but I don't see any obvious way to get there, though.

PR #2047, opened 2015-07-07
WT-1985 More integer packing tests and fixes for python

ddanderson:
@agorrod, here's a followup to #2036.  Can you review?

Your previous comments:
> It would be good if you can investigate the performance issue.

That's covered in WT-1988, I had two cursors open on the same page, they were never both in a state of reset, so the page couldn't be evicted, and there were some enforced pauses. Keith is reviewing some doc modifications extolling the virtues of cursor.reset(), and I added reset into the test case (as part of this pull request).

> I also noticed that there are two test files in lang/python/wiredtiger related to packing intpack-test.py and packing-test.py. Does your new test case make those obsolete? If not could you move them into the test/suite directory, and make them conform with our other test cases?

Thanks for catching that. I moved/rewrote these, they are now covered by test/suite/test_intpack.py and test_pack.py.  They uncovered a few problems - I fixed one in this commit: using 'u' format can expose the hidden 'U' format to the python layer, so python pack/unpack must understand 'U'.  Other problems will be tracked in WT-1998, WT-2000.


agorrod:

PR #2048, opened 2015-07-08
WT-1993 Coverity CID 1310061: Program hangs (LOCK)

michaelcahill:
block_open.c:200: Returning without unlocking "conn->block_lock.lock".


michaelcahill:
@keithbostic can you please review?  The original fix for WT-1993 caused a Coverity error...

keithbostic:
Thanks for fixing -- lgtm, merged.

PR #2049, opened 2015-07-08
WT-1980 Provide an API to query metadata meaningful to WT_SESSION::create

michaelcahill:
Note the connection to WT-1133: this change does not support getting the configuration of connections or sessions, or using a cursor to traverse a config string.


michaelcahill:
@agorrod, can you please review?  There would be an additional change to SERVER-19083 to change "metadata:" to "config:" in the integration layer.

agorrod:
@michaelcahill I don't love that this re-purposes the ```config:``` namespace for cursors. Maybe we could use a specialization on ```metadata:``` instead. Something like ```metadata:create```.

OTOH: I don't have a good vision for how to implement ```config:```, so if you prefer the separation I'm happy to do a full review as is.

michaelcahill:
@agorrod does this look better?

src/include/cursor.h
```diff
@@ -291,8 +291,9 @@ struct __wt_cursor_metadata {
 
 	WT_CURSOR *file_cursor;		/* Queries of regular metadata */
 
-#define	WT_MDC_POSITIONED	0x01
-#define	WT_MDC_ONMETADATA	0x02
+#define	WT_MDC_CREATEONLY	0x01
+#define	WT_MDC_POSITIONED	0x02
+#define	WT_MDC_ONMETADATA	0x04
```
agorrod: Nit: positioned and onmetadata are backward here.

src/cursor/cur_metadata.c
```diff
@@ -30,15 +30,41 @@
...
-	(c)->key.size = (fc)->key.size;					\
-	(c)->value.data = (fc)->value.data;				\
-	(c)->value.size = (fc)->value.size;				\
-	F_SET((c), WT_CURSTD_KEY_EXT | WT_CURSTD_VALUE_EXT);		\
-	F_CLR((mc), WT_MDC_ONMETADATA);					\
-	F_SET((mc), WT_MDC_POSITIONED);					\
-} while (0)
+/*
+ * __curmetadata_setkv --
+ *	Copy the key/value pair from
```
agorrod: The comment is incomplete. It'd be good if it says something about manipulating the content.

agorrod:
Thanks @michaelcahill lgtm modulo a couple of nit-picky comments I made.

PR #2050, opened 2015-07-08
WT-1998 WT-2000 Fixes for indexes with some rarely used key/value formats

michaelcahill:


michaelcahill:
@ddanderson can you please review?  This should fix WT-1998 and WT-2000.

ddanderson:
@michaelcahill, lgtm.  I'm pushing some tests changes to this branch (some tests that now can be enabled with this bug fix). When they clear the build, I'll merge.

PR #2051, opened 2015-07-08
Minor test suite refactor

ddanderson:
Added copy_live_wiredtiger_home function, with two callers.

### DIFF
--- a/test/suite/helper.py
+++ b/test/suite/helper.py
@@ -27,7 +27,7 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 
-import glob, os, string
+import glob, os, shutil, string, subprocess
 import wiredtiger
 
 # python has a filecmp.cmp function, but different versions of python approach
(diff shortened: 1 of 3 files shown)


test/suite/helper.py
```diff
@@ -96,6 +96,29 @@ def confirm_empty(self, uri):
...
+    # unaligned copy requires 'dd', which may not be available on Windows
+    if not aligned and os.name == "nt":
+        raise AssertionError(
+            'copy_live_wiredtiger_home: unaligned copy impossible on Windows')
+    shutil.rmtree(newdir, ignore_errors=True)
+    os.mkdir(newdir)
+    for fname in os.listdir(olddir):
+        fullname = os.path.join(olddir, fname)
+        # Skip lock file, on Windows it is locked.
+        if os.path.isfile(fullname) and "WiredTiger.lock" not in fullname:
```
sueloverso: One difference between the original and the new code is that the original did shutil.copytree on non-windows systems to simulate the expected `cp -rp` that a user would do.  The new code now copies files individually and skips the `WiredTiger.lock` file.  So we're no longer testing what a user is likely to do.  @michaelcahill is that difference important?

sueloverso:
Note this was SERVER-18899 (I thought I had typed this comment already but it disappeared).  Please fix the commit line so it correctly appears in the ticket.

ddanderson:
Yuck.  I already merged it. :-(  Not sure how to fix.

test/suite/helper.py
michaelcahill: This is what a user would have to do on Windows, so I think it's fine.  It's also a reasonable implementation of `cp -rp`, so I don't think the difference will matter.

PR #2054, opened 2015-07-10
WT-1962 Make the hot_backup_lock a read/write lock.

sueloverso:
@michaelcahill please review this branch that changes the hot_backup_lock to a rwlock.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -139,7 +139,7 @@ __log_archive_once(WT_SESSION_IMPL *session, uint32_t backup_file)
 	 * We can only archive files if a hot backup is not in progress or
 	 * if we are the backup.
 	 */
-	__wt_spin_lock(session, &conn->hot_backup_lock);
+	WT_RET(__wt_readlock(session, conn->hot_backup_lock));
 	locked = 1;
 	if (conn->hot_backup == 0 || backup_file != 0) {
 		for (i = 0; i < logcount; i++) {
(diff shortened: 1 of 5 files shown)


src/txn/txn_ckpt.c
```diff
@@ -1099,7 +1099,7 @@ nockpt:			F_SET(btree, WT_BTREE_SKIP_CKPT);
 		btree->modified = 1;
 
 	if (hot_backup_locked)
-		__wt_spin_unlock(session, &conn->hot_backup_lock);
+		WT_RET(__wt_readunlock(session, conn->hot_backup_lock));
```
michaelcahill: This should be `WT_TRET`, I think.

src/conn/conn_log.c
```diff
@@ -167,7 +167,7 @@ __log_archive_once(WT_SESSION_IMPL *session, uint32_t backup_file)
 	if (0)
 err:		__wt_err(session, ret, "log archive server error");
 	if (locked)
-		__wt_spin_unlock(session, &conn->hot_backup_lock);
+		WT_RET(__wt_readunlock(session, conn->hot_backup_lock));
```
michaelcahill: This should be `WT_TRET`, I think.

michaelcahill:
I've made those couple of small changes to error handling (if `__wt_readunlock` fails we're in pretty deep trouble, but I think we should continue with error handling anyway).  I'll merge once it's through testing.

PR #2055, opened 2015-07-13
WT-1985. Integer packing and other fixes for Java

ddanderson:
This set of commits adds more extensive packing tests (PackTest02, PackTest03) for Java and fixes various problems that came up in resolving the new tests:

* The basic PackTest does positive/negative numbers, so needs to use 'q' (signed format).
* Fixed bugs in packing/unpacking bytearrays: we must be aware of 'U' format; when length is provided in format, use it; do padding.
* Fixed signed checking for getInt/Long/Short.
* Added special case encoding for 8256 (to match Python and C), this is the original problem reported in WT-1985.
* Fixed bugs in packing strings: embedded nulls; use length when provided.  Logic copied from packing.py.
* Java's Cursor.reset() more completely resets the cursor - removes the state of any incompletely packed/unpacked items.

### DIFF
--- a/lang/java/Makefile.am
+++ b/lang/java/Makefile.am
@@ -49,6 +49,8 @@ JAVA_JUNIT = \
 		$(JAVATEST)/CursorTest02.java \
 		$(JAVATEST)/ExceptionTest.java \
 		$(JAVATEST)/PackTest.java \
+		$(JAVATEST)/PackTest02.java \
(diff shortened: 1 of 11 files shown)


ddanderson:
@agorrod, can you review please?

agorrod:
Thanks @ddanderson This lgtm - I'll merge.

PR #2058, opened 2015-07-14
WT-2006: s_copyright fixes

keithbostic:
WT-2006

@agorrod, for your review.

`s_copyright` wasn't running on FreeBSD, and checking it over I think I found some problems.

Note, the final commit isn't a bug fix, but it reworks the script so we only create a single copy of the copyright template files.

One thing I didn't change: in the check() function, there's this code:
```
        # Skip files in which WiredTiger holds no rights. 3rdparty and
        # node_modules are used by wtstats
        if expr "$1" : '*3rdparty*' > /dev/null; then
                return;
        fi
        if expr "$1" : '*node_modules*' > /dev/null; then
                return;
        fi
```

There's already code in the parent find that uses sed to remove directories from the copyright list we check:
```
# Search for files, skipping some well-known 3rd party directories.
(cd .. && find [a-z]* -name '*.[chi]' \
        -o -name '*.cxx' \
        -o -name '*.in' \
        -o -name '*.java' \
        -o -name '*.py' |
    sed -e '/Makefile.in/d' \
        -e '/^build_posix\//d' \
        -e '/api\/leveldb\/basho\//d' \
        -e '/api\/leveldb\/hyperleveldb\//d' \
        -e '/api\/leveldb\/leveldb\//d' \
        -e '/api\/leveldb\/rocksdb\//d' \
        -e '/test\/3rdparty\//d' \
        -e '/tools\/wtperf_stats\/3rdparty\//d' \
        -e '/dist\/__/d' \
        -e 's/^\.\///' |
    xargs -P 20 -n 1 -I{} sh dist/s_copyright {})
```

Would it be simpler to put the `3rdparty` and `node_modules` checks with the others, so all the checks are in one place (and it's going to be faster to do that, instead of checking every individual file name for a match).

### DIFF
--- a/dist/s_copyright
+++ b/dist/s_copyright
@@ -9,14 +9,7 @@ c4=__wt.copyright.4
 
 check()
 {
-	# Skip files in which WiredTiger holds no rights. 3rdparty and
-	# node_modules are used by wtstats
-	if expr "$1" : '*3rdparty*' > /dev/null; then
-		return;
-	fi
-	if expr "$1" : '*node_modules*' > /dev/null; then
-		return;
-	fi
+	# Skip files in which WiredTiger holds no rights.
 	if `egrep "skip	$1" dist/s_copyright.list > /dev/null`; then
 		return;
 	fi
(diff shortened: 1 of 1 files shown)


agorrod:
> Would it be simpler to put the 3rdparty and node_modules checks with the others, so all the checks are in one place

Yes - I've made that change (doubly because the check for node_modules with your version wasn't working for me). I didn't include the paths in the sed pattern originally to make it clear that some of those patterns are for specific files, others are generic patterns.

agorrod:
autobuild: retest this please

PR #2059, opened 2015-07-14
WT-1970: read -d command

keithbostic:
WT-1970

@agorrod, `read -d` is a bash extension, and I don't think we actually need it if we use a default xargs configuration.

For your review.

### DIFF
--- a/dist/s_all
+++ b/dist/s_all
@@ -69,26 +69,25 @@ run "sh ./s_typedef -b"
 run "sh ./s_copyright"
 run "sh ./s_style"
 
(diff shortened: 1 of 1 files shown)


keithbostic:
@agorrod, we're not redirecting stderr, shouldn't we be doing something like `2>&1 ./s_tags > ${t_pfx}tags` in the high-level command?

agorrod:
@keithbostic 

Was the reporting the only part that was broken with ```python style.py```? Or do you have another concern - the current implementation works for me where I tested it. Please merge if you are happy.

PR #2060, opened 2015-07-14
WT-1967: Add support for a long-running, isolation=snapshot, reader thread

keithbostic:
WT-1967: Add support for a long-running, isolation=snapshot, reader.

@agorrod, if you want to see the long-running reader in action, this branch can do that.

Here's the CONFIG:
```
cache=2
checkpoints=0
compression=none
data_source=file
file_type=row
internal_page_max=9
leaf_page_max=9
logging=0
long_running_txn=1
ops=1000000
rows=10000
runs=1
salvage=0
threads=2
verify=0
write_pct=80
```
Do a quick mod like this:
```
diff --git a/test/format/lrt.c b/test/format/lrt.c
index a00a4e0..3b9b856 100644
--- a/test/format/lrt.c
+++ b/test/format/lrt.c
@@ -69,6 +69,8 @@ lrt(void *arg)
 
                /* Sleep for some number of seconds. */
                period = mmrand(NULL, 1, 10);
+               fprintf(stderr, "\n" "%s " "%u" "\n",
+                   pinned ? "pinned" : "unpinned", period);
 
                /* Sleep for short periods so we don't make the run wait. */
                while (period > 0 && !g.workers_finished) {
```
and run format, the whole thing freezes up whenever the reading cursor is set.

If you like this change, feel free to merge it into develop -- I've defaulted `long_running_txn` to 0% for now, so it won't affect any runs where we aren't specifically configuring for it.

### DIFF
--- a/test/format/Makefile.am
+++ b/test/format/Makefile.am
@@ -8,7 +8,7 @@ endif
 noinst_PROGRAMS = t
 noinst_SCRIPTS = s_dumpcmp
 t_SOURCES =\
-	config.h format.h backup.c bulk.c compact.c config.c ops.c \
+	config.h format.h backup.c bulk.c compact.c config.c lrt.c ops.c \
 	salvage.c t.c util.c wts.c
 
 if HAVE_BERKELEY_DB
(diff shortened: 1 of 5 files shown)


agorrod:
@keithbostic This code looks like it will be generally useful for reproducing problems, so I'll merge. OTOH I'm not sure we have a solution for reader transactions that keep an ID pinned unless we start allowing eviction to write pages with dirty updates.

PR #2061, opened 2015-07-14
WT-2007 Only grow log slot buffers to  a maximum size.

sueloverso:
@agorrod Here is a branch that implements a simple maximum size.  I chose 256K as the maximum size which uses 32Mb of memory for log slot buffers.  This is the smallest code change.  Please compare this to the other pull request.


agorrod:
We've decided to go with the solution in #2062, closing for now.

PR #2062, opened 2015-07-14
WT-2007 Statically allocate log slot buffers to a maximum size.

sueloverso:
@agorrod, please review this branch for log slot buffer memory.  This one uses the same 256kb max size, but just statically allocates that right away.  It removes a bunch of code.  Please compare to #2061.  Only one (or none if we want another approach for now) should be merged.  The other should be closed and rejected.

I tested both with a wtperf config that used value sizes of 100k to force log slot overflow.

### DIFF
--- a/src/include/log.h
+++ b/src/include/log.h
@@ -12,7 +12,7 @@
 
 /* Logging subsystem declarations. */
 #define	WT_LOG_ALIGN			128
-#define	WT_LOG_SLOT_BUF_INIT_SIZE	64 * 1024
+#define	WT_LOG_SLOT_BUF_SIZE		256 * 1024
 
 #define	WT_INIT_LSN(l)	do {						\
 	(l)->file = 1;							\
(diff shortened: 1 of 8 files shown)


sueloverso:
I chose 256Kb as the maximum size.  That uses 32Mb total for slot buffers.  A slot with small to moderate record size can still get a lot of consolidation on a single slot at that size, it doesn't use vast amounts of memory.  Large record sizes, say a few 10's kb per record, can still get decent consolidation on a single slot.  So that seemed like a reasonable value.  Let me know if you disagree.

src/include/log.h
```diff
@@ -91,11 +91,10 @@ typedef WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) struct {
...
 
-#define	WT_SLOT_BUF_GROW	0x01		/* Grow buffer on release */
-#define	WT_SLOT_BUFFERED	0x02		/* Buffer writes */
-#define	WT_SLOT_CLOSEFH		0x04		/* Close old fh on release */
-#define	WT_SLOT_SYNC		0x08		/* Needs sync on release */
-#define	WT_SLOT_SYNC_DIR	0x10		/* Directory sync on release */
+#define	WT_SLOT_BUFFERED	0x01		/* Buffer writes */
+#define	WT_SLOT_CLOSEFH		0x02		/* Close old fh on release */
+#define	WT_SLOT_SYNC		0x04		/* Needs sync on release */
+#define	WT_SLOT_SYNC_DIR	0x18		/* Directory sync on release */
```
sueloverso: This should be 0x08

src/log/log_slot.c
```diff
@@ -312,23 +319,9 @@ __wt_log_slot_free(WT_SESSION_IMPL *session, WT_LOGSLOT *slot)
 {
 	WT_DECL_RET;
 
+	WT_UNUSED(session);
```
agorrod: This function can be simplified now. The ``ret``` variable is never set. Maybe we should remove the session parameter altogether?

agorrod:
Thanks @sueloverso this change lgtm modulo one cleanup comment I made.

PR #2063, opened 2015-07-14
WT-2007 Add configuration knobs to set slots and memory.

sueloverso:
@agorrod, this pull request adds in configuration knobs to set the slot memory and number of slots.  It is built on #2062.  It is minimally tested (I did run a large wtperf config and then reran changing the knob settings).


agorrod:
@sueloverso @michaelcahill and I talked about this, and decided that exposing this as a configuration isn't justified at the moment.

PR #2064, opened 2015-07-15
WT-2008 Fix a bug in recovery where a file create went missing.

agorrod:
The problem was that a hard system crash could cause a file to not
be created, even though it had been synced. The solution is to do a
directory sync each time a new file is created.


agorrod:
autobuild: retest this please

michaelcahill:
Thanks, @agorrod -- lgtm.  I'll merge.

PR #2066, opened 2015-07-16
WT-2009 Apply tracked metadata operations post-commit

michaelcahill:
In particular, don't remove dropped files until the metadata changes are durable.

### DIFF
--- a/src/meta/meta_track.c
+++ b/src/meta/meta_track.c
@@ -125,31 +125,21 @@ __wt_meta_track_on(WT_SESSION_IMPL *session)
  *	Apply the changes in a metadata tracking record.
  */
 static int
-__meta_track_apply(WT_SESSION_IMPL *session, WT_META_TRACK *trk, int unroll)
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @michaelcahill I'll try this out with my reproducing test case.

michaelcahill:
Thanks @agorrod.  If it helps, there are binaries here: https://evergreen.mongodb.com/version/55a72e883ff122070100001c_0

agorrod:
Thanks @michaelcahill This lgtm - are you happy for me to merge?

michaelcahill:
Yes, please -- let's get this into regular testing.  Should we include the change to check that  `WT_LOG_FLUSH` is set, so that we're actually flushing the log, or leave that for a separate branch?

agorrod:
I'm inclined to leave the log flush change in a separate branch for now - I'm still not sure what the best solution for that part of the problem is.

PR #2067, opened 2015-07-16
WT-1989 Log scalability

sueloverso:
@michaelcahill Please review this diff. It is essentially the "patch2" with a few changes.  With this patch, compared to develop I see the following performance (ops/sec) running Bruce's workload (on the AWS SSD).

Client threads:|Dev|Branch|difference
--------------------|---|-----|---------------
8 | 109 | 92 | -16%
16 | 125 | 151 | +21%
24 | 117 | 143 | +22%
32 | 121 | 126 | +4%
64 | 120 | 111 | -8%

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -241,6 +241,7 @@ connection_stats = [
     LogStat('log_writes', 'log write operations'),
     LogStat('log_write_lsn', 'log server thread advances write LSN'),
 
+    LogStat('log_slot_coalesced', 'written slots coalesced'),
     LogStat('log_slot_consolidated', 'logging bytes consolidated'),
     LogStat('log_slot_closes', 'consolidated slot closures'),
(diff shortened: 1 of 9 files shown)


agorrod:
@sueloverso Isn't <= 8 threads a common case? Is the intent of this case to promote more coalescing when there are lots of slots finishing in parallel? Is there some way to alter how aggressive we are at coalescing slots depending on the current log record throughput to avoid the cost at lower levels of concurrency?

sueloverso:
@agorrod I investigated the 8 thread degradation and found it was due to the changes to no longer yield in the worker thread.  When I fixed that the values at lower numbers of threads increased.  The 8 thread case is now within a couple percent of the original.

michaelcahill:
Thanks @sueloverso, lgtm -- I'll merge.

PR #2069, opened 2015-07-17
WT-1996 If there is an error applying updates during a page rewrite, don't free the first update

michaelcahill:


michaelcahill:

agorrod:
Thanks @michaelcahill lgtm.

PR #2072, opened 2015-07-19
WT-1967: evict pages where updates are still required by older readers.

keithbostic:
WT-1967: evict pages where updates are still required by older readers.

@agorrod, @michaelcahill: this branch is a long way from working, but I think most of the pieces are here, and I'm pushing it so you can get an idea of where I'm heading.


keithbostic:
Here's the CONFIG I've been using to test. It's basically a tiny cache, lots of updates, and format configured with a long-running reader.
```
cache=2
checkpoints=0
compression=none
data_source=file
file_type=row
internal_page_max=9
leaf_page_max=9
logging=0
long_running_txn=1
ops=10000000
rows=100000
runs=1
salvage=0
threads=2
verify=0
write_pct=80
repeat_data_pct=0
```

keithbostic:
@michaelcahill, there's this code in the Btree handle open source:
```
       /*              
         * Set special flags for the metadata file.
         * Eviction; the metadata file is never evicted.
         * Logging; the metadata file is always logged if possible.
         */
        if (WT_IS_METADATA(btree->dhandle)) {
                F_SET(btree, WT_BTREE_IN_MEMORY | WT_BTREE_NO_EVICTION);
                F_CLR(btree, WT_BTREE_NO_LOGGING);
```
That feels like the wrong place to put it, wouldn't it be more reasonable to set that in the `meta_table` code that's opening the handle?

I'm doing something similar for the LAS file at the moment, which is why I noticed.

michaelcahill:
fbm to move that to metadata-specific code (and similarly for the new table).

keithbostic:
@michaelcahill, @agorrod, I pushed this version so you could take a look at `__rec_txn_read()` if you have the time. Basically, it's a rework to handle storing a copy of the on-page value in the WT_UPDATE list.

(This version doesn't work yet, there's a memory overwrite I'm chasing, and when it doesn't hit that, it eventually dies because we lost track of a pinned record.)

keithbostic:
@michaelcahill, @agorrod, here's the primary problem I'm chasing:
```
t: process 94211
pin: snap_min: 3, snap_max: 3
pin-verify: snap_min: 266681, snap_max: 266681
pin: snap_min: 376548, snap_max: 376549
las-insert:
	[85787648-85788160, 512, 3989602656]
	K {0000099084.00/opq}, onpage-txnid=479611
	V {}, txnid=479611
	V {0000099084/LMNOPQRSTUVWXYZABCDEFGHIJ}, txnid=0
read:
	[85787648-85788160, 512, 3989602656]
	instantiate
	K {0000099084.00/opq} 
	V {}, txnid=479611
	V {0000099084/LMNOPQRSTUVWXYZABCDEFGHIJ}, txnid=0
pin-verify: snap_min: 529635, snap_max: 529636
t: read_row 99084: WT_NOTFOUND: item not found
```
* `pin`: cursor's session's snap_min/max values when it's positioned at key 99084
* `pin-verify`: snap_min/max values when it moves back to that key to check if the value's the same
* `las-insert`: write out a page with lookaside records (a delete @ 479611, plus the on-page value @ 0)
* `read`: read the page back in, instantiate those records
* `read_row`: we failed to find the expected value in the snapshot cursor.

So, I think there's something I don't understand about the progression of transaction IDs in the system -- I expected snap_min/max to stay the same while the cursor (in WT_ISO_SNAPSHOT), slept?

Anyway, if you have a few minutes, any direction you can offer would be much appreciated.

michaelcahill:
@keithbostic, I can't see the "pin-verify" code, but without an explicit `begin_transaction`, the snapshot is only kept around while a cursor is positioned.  As soon as the cursor loses its position (including at the beginning of a `search`), a new snapshot will be allocated.

It sounds like `test/format` needs an explicit transaction to do this check?

keithbostic:
@michaelcahill, thanks -- that was the problem, and creating an explicit transaction fixed it.

I can't find any documentation that I think would have prevented me from making this mistake, is there something that covers this I missed?

michaelcahill:
@keithbostic in http://source.wiredtiger.com/develop/cursors.html#cursors_transactions we say:

> If no transaction is active, cursor reads are performed at the isolation level of the session, set with the isolation configuration key to WT_CONNECTION::open_session and successful updates are automatically committed before the update operation completes.
> 
> Any operation that consists of multiple related updates should be enclosed in an explicit transaction to ensure that the updates are applied atomically.
> 
> At read-committed (the default) or snapshot isolation levels, committed changes from concurrent transactions become visible when no cursor is positioned. In other words, at these isolation levels, all cursors in a session read from a stable snapshot while any cursor in the session remains positioned.

This section is linked from the transactions page.  I'm happy to update other places in the documentation: out of interest, where did you look?

keithbostic:
@michaelcahill, the next question.

The problem I'm chasing now is this:
```
==631==ERROR: AddressSanitizer: heap-use-after-free on address 0x603000226c30 at pc 0x0000005b795e bp 0x7f110b73a670 sp 0x7f110b73a668
WRITE of size 8 at 0x603000226c30 thread T7
    #0 0x5b795d in __wt_txn_rollback wiredtiger/src/txn/txn.c:589:4
    #1 0x59dfa6 in __session_rollback_transaction wiredtiger/src/session/session_api.c:861:2
    #2 0x4cd909 in ops wiredtiger/test/format/ops.c:524:16
  
0x603000226c30 is located 0 bytes inside of 20-byte region [0x603000226c30,0x603000226c44)
freed by thread T8 here:
    #0 0x4a4a3b in free (/mnt/fast/bostic/work/wiredtiger/test/format/t+0x4a4a3b)
    #1 0x5dd977 in __wt_free_update_list wiredtiger/src/btree/bt_discard.c:415:3
    #2 0x5dd977 in __free_update wiredtiger/src/btree/bt_discard.c:397
    #3 0x5dd977 in __free_page_row_leaf wiredtiger/src/btree/bt_discard.c:337
    #4 0x5dd977 in __wt_page_out wiredtiger/src/btree/bt_discard.c:122
    #5 0x50ce9b in __evict_page_dirty_update wiredtiger/src/evict/evict_page.c:255:3
    #6 0x50ce9b in __wt_evict wiredtiger/src/evict/evict_page.c:122
    #7 0x50ab42 in __evict_page wiredtiger/src/evict/evict_lru.c:1365:2
```
In short, I'm guessing we evicted a page and then aborted a transaction that wanted to access to an update list we've discarded.

Maybe this means my test of the WT_UPDATE list is flawed? I'm using this:
```
if (*updp == NULL) { 
        if (F_ISSET(r, WT_EVICTING) ?
            __wt_txn_committed(session, txnid) :
            __wt_txn_visible(session, txnid)) 
                *updp = upd;
        else    
                skipped = 1;
} else
        WT_ASSERT(session,
                !F_ISSET(r, WT_EVICTING) || __wt_txn_committed(session, txnid));
```

michaelcahill:
@keithbostic, I'm still working my way through, but I think there are some issues with that logic.

For one thing, there is no guarantee at lower isolation levels than snapshot that transaction IDs are committed in order in the update list.  Because of that, I believe the assertion in the `else` clause can fire.  In other words, just because the first update we saw was committed, that doesn't mean we won't see an uncommitted update later in the list.  That update could still roll back, so depending on whether you are running with diagnostics, maybe this could explain the segfault?

I also see some error paths that don't set `r->leave_dirty`: is that intentional?

keithbostic:
@michaelcahill, that was my guess -- I'll rework the test to EBUSY any pages with an uncommitted update in the list, but the assertion isn't firing, even with the core dump, so that's not the problem.

I think the returns are correct.

keithbostic:
@michaelcahill, are we guaranteed transaction IDs in update lists are always decreasing, that it's not possible to have an update list `a -> b -> c` where b or c have a transaction ID less-than a's?

How about equal-to?

keithbostic:
@michaelcahill, can we talk for a few minutes tonight?

I put in a check for every update my `__rec_txn_read` code thinks is "committed" during eviction, walking the list of sessions and looking for a running transaction that has a pointer to that WT_UPDATE structure, and that hit. Stepping back I got to this:
```
diff --git a/src/txn/txn.c b/src/txn/txn.c
index 9e8def3..1edd83e 100644
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -593,6 +593,9 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
                switch (op->type) {
                case WT_TXN_OP_BASIC:
                case WT_TXN_OP_INMEM:
+                       WT_ASSERT(session,
+                           op->u.upd->txnid >
+                           S2C(session)->txn_global.last_running);
                        op->u.upd->txnid = WT_TXN_ABORTED;
                        break;
                case WT_TXN_OP_REF:
```
and that fires in the develop branch, which isn't what I expected.

michaelcahill:
@keithbostic, on your first question, any ordering of txnids in an update list is possible -- the same txn can do multiple updates, there is no guarantee of any decreasing order, etc.

On the second part, what is the test case to see that assertion firing?  The `last_running` calculation is sloppy, but I expected it to be behind all running transactions.

michaelcahill:
p.s. `last_running == txnid` is permitted.  The assertion I'd expect to work is:

```c
                       WT_ASSERT(session,
                           !WT_TXNID_LT(op->u.upd->txnid,
                           S2C(session)->txn_global.last_running));
```

keithbostic:
The CONFIG I'm using to test this branch appears to fire that assert in the develop branch.

```
pixiebob:format {33} git branch
* develop
  master
  mongodb-3.0
  mongodb-3.2
  mutex-overflow
pixiebob:format {34} git diff
diff --git a/src/txn/txn.c b/src/txn/txn.c
index 9e8def3..0c09940 100644
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -593,6 +593,10 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
                switch (op->type) {
                case WT_TXN_OP_BASIC:
                case WT_TXN_OP_INMEM:
+                       WT_ASSERT(session,
+                           !WT_TXNID_LT(op->u.upd->txnid,
+                           S2C(session)->txn_global.last_running));
+
                        op->u.upd->txnid = WT_TXN_ABORTED;
                        break;
                case WT_TXN_OP_REF:
pixiebob:format {35} cat CONFIG
cache=2
checkpoints=0
compression=none
data_source=file
file_type=row
huffman_key=0
huffman_value=0
internal_page_max=9
leaf_page_max=9
logging=0
long_running_txn=1
#ops=1000
ops=10000000
repeat_data_pct=0
rows=100000
runs=10
salvage=0
threads=2
verify=0
write_pct=80
pixiebob:format {36} ./t 
t: process 65959
[1438216567:32664][65959:0x803406c00], t, WT_SESSION.rollback_transaction: src/txn/txn.c, 598: !WT_TXNID_LT(op->u.upd->txnid, S2C(session)->txn_global.last_running)
[1438216567:32687][65959:0x803406c00], t, WT_SESSION.rollback_transaction: aborting WiredTiger library
[1438216567:32691][65959:0x803406c00], t, WT_SESSION.rollback_transaction: process ID 65959: waiting for debugger...
^C 1: ops: search 1079785, insert 140221, update 615184, remove 324381
124.499u 72.524s 1:34.23 209.0%	1968+213k 0+76io 0pf+0w
```

keithbostic:
This CONFIG fires the assert instantly, and the `long_running_txn` configuration is what makes it fire.
```
cache=2
checkpoints=0
compression=none
data_source=file
file_type=row
huffman_key=0
huffman_value=0
internal_page_max=9
leaf_page_max=9
logging=0
long_running_txn=1
ops=1000
repeat_data_pct=0
rows=100
runs=10
salvage=0
threads=2
verify=0
write_pct=80
```

keithbostic:
@michaelcahill, with the cherry-pick of 476f3d6 and 0e967f5, this branch/CONFIG file appears to run.

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill: something I've been thinking about is some cookie to include in the LAS key or value that would identify page re-use, so we don't have to remove entries from the LAS when freeing backing blocks. We could make the block manager's address cookie unique (a write generation?), but maybe there's something simpler?

keithbostic:
autobuild: test again please

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
@agorrod, it would be great if you could take a look at 6896814 at some point; I know you have strong feelings about how eviction's global state is accessed. I think this is moving in the right direction, but I'm definitely open to something better!

src/btree/bt_read.c
```diff
@@ -8,21 +8,400 @@
...
+		 * dirty transaction to the last transaction currently running.
+		 * However, the updates we installed may be older than that.
+		 * Set the first dirty transaction to an impossibly old value
+		 * so this page is never skipped in a checkpoint.
+		 *
+		 * KEITH: is this correct? We don't care about checkpoints, we
+		 * care about older readers in the system.
+		 */
+		page->modify->first_dirty_txn = WT_TXN_FIRST;
+	}
```
keithbostic: @michaelcahill, would you please review this clause, I'm not sure what's correct. Thanks!

src/conn/conn_las.c
```diff
@@ -0,0 +1,383 @@
...
+
+	/*
+	 * Force discard on release (must be last handle discarded).
+	 *
+	 * KEITH:
+	 * There's a problem, if WT_DHANDLE_DISCARD_FORCE isn't set on the
+	 * lookaside data handle here, we will drop core removing the data
+	 * handle from the connection session's hash list.
+	 */
+	F_SET(conn->las_session->dhandle, WT_DHANDLE_DISCARD_FORCE);
```
keithbostic: @michaelcahill, if the problem here is obvious to you, would you please comment, I've been staring at it awhile without any success.

It's trivial to reproduce, just comment out the `WT_DHANDLE_DISCARD_FORCE` line in `__wt_las_destroy` and run the usual CONFIG with a small number of ops, it hits every time.

This is the stack:
```
0x0000000000485fde in __session_discard_btree (session=0x80343c000, 
    dhandle_cache=0x8054414c0) at src/session/session_dhandle.c:278
278		SLIST_REMOVE(&session->dhhash[bucket],
Current language:  auto; currently minimal
(gdb) where
#0  0x0000000000485fde in __session_discard_btree (session=0x80343c000, 
    dhandle_cache=0x8054414c0) at src/session/session_dhandle.c:278
#1  0x0000000000485ea4 in __wt_session_close_cache (session=0x80343c000)
    at src/session/session_dhandle.c:295
#2  0x00000000004ec2f8 in __wt_conn_dhandle_discard (session=0x80343c000)
    at src/conn/conn_dhandle.c:688
#3  0x000000000041dffb in __wt_connection_close (conn=0x80342c000)
    at src/conn/conn_open.c:125
#4  0x00000000004141b8 in __conn_close (wt_conn=0x80342c000, config=0x0)
    at src/conn/conn_api.c:1005
#5  0x000000000040f7b2 in wts_close () at wts.c:446
#6  0x000000000040ba3b in main (argc=0, argv=0x7fffffffd9d8) at t.c:232
```

keithbostic:
@michaelcahill, I figured out what was wrong with the unverified file chunks, and I'm making a change where I'd appreciate some careful thought.

An earlier change in this branch was to have reconciliation "own" the decision if a page can be evicted, that is, this test from `evict/evict_page.c`:
```     
/*
 * If the page was ever modified, make sure all of the updates
 * on the page are old enough they can be discarded from cache.
 */
if (!closing && mod != NULL &&
    !__wt_txn_visible_all(session, mod->rec_max_txn) &&
    !FLD_ISSET(reconcile_flags, WT_SKIP_UPDATE_RESTORE))
        return (EBUSY);
```
was previously done outside of reconciliation: reconciliation would return 0, doing it's normal successful wrapup (including checkpoints and root splits), and then eviction might potentially overrule that return and block eviction. Since reconciliation already owns most of this testing (and with the lookaside file, will own more and should have better information), I moved this test into reconciliation.

The problem: successful reconciliation used to work like this:
```
ret = reconcile the page, writing some number of blocks
if (ret == 0)
    wrapup function {
        free the old address information
        free the previous backing blocks
        clean up overflow tracking
        checkpoint if necessary
        update the WT_PAGE_MODIFY information (for example, splits)
        mark the page clean and decide if reconciliation was successful
    }
if (ret == 0)
    mark the parent page dirty
```
and now that the wrapup-function is returning EBUSY when it wasn't before, we were swapping the newly written pages into place, then returning an error without marking the parent dirty, and we lost the blocks.

The change in 7194a72 works like this:

ret = reconcile the page, writing some number of blocks
if (ret == 0)
    ret = mark the page clean and decide if reconciliation was successful
if (ret == 0)
    the rest of the wrapup code
else
    wrapup on error, discard the reconciliation work

I like the change because if we are going to fail (EBUSY), we do (a little?) less work in reconciliation, but more interestingly, we release backing blocks allocated/written in the current checkpoint (making them immediately available for reuse), rather than allocating new blocks from the current checkpoint and then discarding blocks from a previous checkpoint, which should minimize our work to maintain checkpoints and general churn in the file.

However, the change is aggressive:

1. If we fail after deciding reconciliation was successful, we're screwed. We can (and probably should) mark the page dirty, but nothing good is going to happen after that.

    That said: we're already not handling these failures: if we get half-way through discarding the previous backing blocks, and then fail, we're screwed. I've been staring at this for awhile, and I suggest we panic if reconciliation ever fails while trying to swap a page into place. The mostly likely failure is either a memory allocation, or a block-manager problem, but regardless, I think our chances of continuing to run after a failure in this code is minimal.

2. Second, reconciliation will modify WT_PAGE_MODIFY fields after marking the page clean. I think we're safe because we haven't released our locks, but I'm concerned reconciliation might touch something in WT_PAGE_MODIFY that isn't entirely owned by reconciliation, after marking the page clean. Or, a problem because reconciliation sets `WT_BTREE.rec_max_txn` before finalizing changes to WT_PAGE_MODIFY.

Thoughts?

keithbostic:
@michaelcahill, compaction plus the lookaside file can lead to self-deadlock. Here's the stack:
```
#0  0x00007f629100af7d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f6291006d41 in _L_lock_790 () from /lib64/libpthread.so.0
#2  0x00007f6291006c47 in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x0000000000b6c11d in __wt_spin_lock (session=0x632000000b00, 
    t=0x61800000fe08) at ./src/include/mutex.i:159
#4  0x0000000000b66700 in __wt_insert_serial (session=0x632000000b00, 
    page=0x61300008b400, ins_head=0x60e000082800, ins_stack=0x6170000ff2e8, 
    new_insp=0x7f628daa4390, new_ins_size=170, skipdepth=1)
    at ./src/include/serial.i:223
#5  0x0000000000b600e8 in __wt_row_modify (session=0x632000000b00, 
    cbt=0x6170000ff180, key=0x6170000ff258, value=0x6170000ff280, upd_arg=0x0, 
    is_remove=0) at src/btree/row_modify.c:206
#6  0x0000000000e7607b in __cursor_row_modify (session=0x632000000b00, 
    cbt=0x6170000ff180, is_remove=0) at src/btree/bt_cursor.c:266
#7  0x0000000000e74b43 in __wt_btcur_insert (cbt=0x6170000ff180)
    at src/btree/bt_cursor.c:547
#8  0x0000000000c2bfc3 in __curfile_insert (cursor=0x6170000ff180)
    at src/cursor/cur_file.c:245
#9  0x00000000007c8e85 in __rec_update_las (session=0x632000000b00, 
    r=0x6170000ff500, btree_id=2, bnd=0x61a000046880)
    at src/reconcile/rec_write.c:3450
#10 0x00000000007c1d3a in __rec_split_write (session=0x632000000b00, 
    r=0x6170000ff500, bnd=0x61a000046880, buf=0x6170000ff518, last_block=1)
    at src/reconcile/rec_write.c:3269
#11 0x00000000007d8e1c in __rec_split_finish_std (session=0x632000000b00, 
    r=0x6170000ff500) at src/reconcile/rec_write.c:2957
#12 0x00000000007743f8 in __rec_split_finish (session=0x632000000b00, 
    r=0x6170000ff500) at src/reconcile/rec_write.c:2972
#13 0x00000000007618fb in __rec_row_leaf (session=0x632000000b00, 
    r=0x6170000ff500, page=0x61a000537080, salvage=0x0)
    at src/reconcile/rec_write.c:5181
#14 0x0000000000741540 in __wt_reconcile (session=0x632000000b00, 
    ref=0x6040001a0a50, salvage=0x0, flags=3) at src/reconcile/rec_write.c:469
#15 0x000000000060e3e1 in __evict_review (session=0x632000000b00, 
    ref=0x6040001a0a50, inmem_splitp=0x7f628daa8f80, closing=0)
    at src/evict/evict_page.c:393
#16 0x0000000000609c5b in __wt_evict (session=0x632000000b00, 
    ref=0x6040001a0a50, closing=0) at src/evict/evict_page.c:79
#17 0x00000000005ebee8 in __evict_page (session=0x632000000b00, is_server=1)
    at src/evict/evict_lru.c:1408
#18 0x000000000060018d in __evict_lru_pages (session=0x632000000b00, 
    is_server=1) at src/evict/evict_lru.c:753
#19 0x00000000005ffc2b in __evict_server_work (session=0x632000000b00)
    at src/evict/evict_lru.c:875
#20 0x00000000005f425c in __evict_pass (session=0x632000000b00)
    at src/evict/evict_lru.c:525
#21 0x00000000005e3222 in __evict_server (arg=0x632000000b00)
    at src/evict/evict_lru.c:168
#22 0x00007f6291004df3 in start_thread () from /lib64/libpthread.so.0
#23 0x00007f62903ff1bd in clone () from /lib64/libc.so.6
(gdb) frame 4
#4  0x0000000000b66700 in __wt_insert_serial (session=0x632000000b00, 
    page=0x61300008b400, ins_head=0x60e000082800, ins_stack=0x6170000ff2e8, 
    new_insp=0x7f628daa4390, new_ins_size=170, skipdepth=1)
    at ./src/include/serial.i:223
223			WT_PAGE_LOCK(session, page);
(gdb) p page->modify->page_lock
$20 = 7 '\a'
(gdb) frame 14
#14 0x0000000000741540 in __wt_reconcile (session=0x632000000b00, 
    ref=0x6040001a0a50, salvage=0x0, flags=3) at src/reconcile/rec_write.c:469
469			ret = __rec_row_leaf(session, r, page, salvage);
(gdb) p page->modify->page_lock
$21 = 7 '\a'
```
Reconciliation locks the page if we're compacting the file to which the page belongs -- we acquire a page lock in order to insert records in the lookaside file, and if that page uses the same page-lock as this page, we can self-deadlock.

The "fix" I pushed in 533b4c4 is to turn off the lookaside file while compaction is running, but I don't like that much. I could give the lookaside file it's own page-lock, that feels cleaner to me. Thoughts?

keithbostic:
@michaelcahill, digging deeper, this change is insufficient, here's another run.

```
(gdb) thread 7
[Switching to thread 7 (Thread 0x7fc1ba7f8700 (LWP 22354))]
#0  0x00007fc1c4d3cf7d in __lll_lock_wait () from /lib64/libpthread.so.0
(gdb) where
#0  0x00007fc1c4d3cf7d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007fc1c4d38d41 in _L_lock_790 () from /lib64/libpthread.so.0
#2  0x00007fc1c4d38c47 in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x0000000000b6c11d in __wt_spin_lock (session=0x632000001700, 
    t=0x61800000ff58) at ./src/include/mutex.i:159
#4  0x0000000000b66700 in __wt_insert_serial (session=0x632000001700, 
    page=0x60c000011140, ins_head=0x60e0000bf7a0, ins_stack=0x61700000b468, 
    new_insp=0x7fc1ba7f0a70, new_ins_size=115, skipdepth=2)
    at ./src/include/serial.i:223
#5  0x0000000000b600e8 in __wt_row_modify (session=0x632000001700, 
    cbt=0x61700000b300, key=0x61700000b3d8, value=0x61700000b400, upd_arg=0x0, 
    is_remove=0) at src/btree/row_modify.c:206
#6  0x0000000000e7607b in __cursor_row_modify (session=0x632000001700, 
    cbt=0x61700000b300, is_remove=0) at src/btree/bt_cursor.c:266
#7  0x0000000000e74b43 in __wt_btcur_insert (cbt=0x61700000b300)
    at src/btree/bt_cursor.c:547
#8  0x0000000000c2bfc3 in __curfile_insert (cursor=0x61700000b300)
    at src/cursor/cur_file.c:245
#9  0x00000000007c8e85 in __rec_update_las (session=0x632000002900, 
    r=0x61700006a900, btree_id=2, bnd=0x61a000060080)
    at src/reconcile/rec_write.c:3450
#10 0x00000000007c1d3a in __rec_split_write (session=0x632000002900, 
    r=0x61700006a900, bnd=0x61a000060080, buf=0x61700006a918, last_block=1)
    at src/reconcile/rec_write.c:3269
#11 0x00000000007d8e1c in __rec_split_finish_std (session=0x632000002900, 
    r=0x61700006a900) at src/reconcile/rec_write.c:2957
#12 0x00000000007743f8 in __rec_split_finish (session=0x632000002900, 
    r=0x61700006a900) at src/reconcile/rec_write.c:2972
#13 0x00000000007618fb in __rec_row_leaf (session=0x632000002900, 
    r=0x61700006a900, page=0x61c0009d3880, salvage=0x0)
    at src/reconcile/rec_write.c:5181
#14 0x0000000000741540 in __wt_reconcile (session=0x632000002900, 
    ref=0x604000011010, salvage=0x0, flags=3) at src/reconcile/rec_write.c:469
#15 0x000000000060e3e1 in __evict_review (session=0x632000002900, 
    ref=0x604000011010, inmem_splitp=0x7fc1ba7f5660, closing=0)
    at src/evict/evict_page.c:393
#16 0x0000000000609c5b in __wt_evict (session=0x632000002900, 
    ref=0x604000011010, closing=0) at src/evict/evict_page.c:79
#17 0x00000000005ebee8 in __evict_page (session=0x632000002900, is_server=0)
    at src/evict/evict_lru.c:1408
#18 0x00000000005e93db in __wt_cache_eviction_worker (session=0x632000002900, 
    busy=0, pct_full=96) at src/evict/evict_lru.c:1485
#19 0x0000000000e91bd1 in __wt_cache_eviction_check (session=0x632000002900, 
    busy=0, didworkp=0x0) at ./src/include/cache.i:242
#20 0x0000000000e91057 in __cursor_enter (session=0x632000002900)
    at ./src/include/cursor.i:63
#21 0x0000000000e90be2 in __curfile_enter (cbt=0x617000017380)
    at ./src/include/cursor.i:96
#22 0x0000000000e6f85e in __cursor_func_init (cbt=0x617000017380, reenter=1)
    at ./src/include/cursor.i:200
#23 0x0000000000e7c3c7 in __wt_btcur_update (cbt=0x617000017380)
    at src/btree/bt_cursor.c:763
#24 0x0000000000c2f6c0 in __curfile_update (cursor=0x617000017380)
    at src/cursor/cur_file.c:284
#25 0x00000000004e51c6 in row_update (tinfo=0x617000005600, 
    cursor=0x617000017380, key=0x7fc1ba7f7be0, value=0x7fc1ba7f7c30, 
    keyno=74913) at ops.c:834
#26 0x00000000004dabff in ops (arg=<optimized out>) at ops.c:465
#27 0x00007fc1c4d36df3 in start_thread () from /lib64/libpthread.so.0
#28 0x00007fc1c41311bd in clone () from /lib64/libc.so.6
(gdb) frame 9
#9  0x00000000007c8e85 in __rec_update_las (session=0x632000002900, 
    r=0x61700006a900, btree_id=2, bnd=0x61a000060080)
    at src/reconcile/rec_write.c:3450
3450				WT_ERR(cursor->insert(cursor));
(gdb) p cursor
$29 = (WT_CURSOR *) 0x61700000b300
(gdb) S2C
$30 = (WT_CONNECTION_IMPL *) 0x630000000400
(gdb) p $30->las_cursor == cursor
$31 = 1
(gdb) frame 14
#14 0x0000000000741540 in __wt_reconcile (session=0x632000002900, 
    ref=0x604000011010, salvage=0x0, flags=3) at src/reconcile/rec_write.c:469
469			ret = __rec_row_leaf(session, r, page, salvage);
(gdb) p page_lock
$32 = 1
(gdb) p page->modify->page_lock
$33 = 10 '\n'
(gdb) frame 4
#4  0x0000000000b66700 in __wt_insert_serial (session=0x632000001700, 
    page=0x60c000011140, ins_head=0x60e0000bf7a0, ins_stack=0x61700000b468, 
    new_insp=0x7fc1ba7f0a70, new_ins_size=115, skipdepth=2)
    at ./src/include/serial.i:223
223			WT_PAGE_LOCK(session, page);
(gdb) p page->modify->page_lock
$34 = 13 '\r'


(gdb) thread 4
[Switching to thread 4 (Thread 0x7fc1b8ff5700 (LWP 22357))]
#0  0x00007fc1c4d3cf7d in __lll_lock_wait () from /lib64/libpthread.so.0
(gdb) where
#0  0x00007fc1c4d3cf7d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007fc1c4d38d41 in _L_lock_790 () from /lib64/libpthread.so.0
#2  0x00007fc1c4d38c47 in pthread_mutex_lock () from /lib64/libpthread.so.0
#3  0x000000000056c89d in __wt_spin_lock (session=0x632000003200, 
    t=0x63000000df40) at ./src/include/mutex.i:159
#4  0x000000000056c3ee in __wt_las_cursor (session=0x632000003200, 
    cursorp=0x7fc1b8feffb0, reset_evict=0x7fc1b8ff0190)
    at src/conn/conn_las.c:206
#5  0x00000000007c5bcd in __rec_update_las (session=0x632000003200, 
    r=0x6170000e3f80, btree_id=2, bnd=0x61a00015f080)
    at src/reconcile/rec_write.c:3310
#6  0x00000000007c1d3a in __rec_split_write (session=0x632000003200, 
    r=0x6170000e3f80, bnd=0x61a00015f080, buf=0x6170000e3f98, last_block=1)
    at src/reconcile/rec_write.c:3269
#7  0x00000000007d8e1c in __rec_split_finish_std (session=0x632000003200, 
    r=0x6170000e3f80) at src/reconcile/rec_write.c:2957
#8  0x00000000007743f8 in __rec_split_finish (session=0x632000003200, 
    r=0x6170000e3f80) at src/reconcile/rec_write.c:2972
#9  0x00000000007618fb in __rec_row_leaf (session=0x632000003200, 
    r=0x6170000e3f80, page=0x61c0007ce080, salvage=0x0)
    at src/reconcile/rec_write.c:5181
#10 0x0000000000741540 in __wt_reconcile (session=0x632000003200, 
    ref=0x60400000ec10, salvage=0x0, flags=3) at src/reconcile/rec_write.c:469
#11 0x000000000060e3e1 in __evict_review (session=0x632000003200, 
    ref=0x60400000ec10, inmem_splitp=0x7fc1b8ff3480, closing=0)
    at src/evict/evict_page.c:393
#12 0x0000000000609c5b in __wt_evict (session=0x632000003200, 
    ref=0x60400000ec10, closing=0) at src/evict/evict_page.c:79
#13 0x00000000005ebee8 in __evict_page (session=0x632000003200, is_server=0)
    at src/evict/evict_lru.c:1408
#14 0x00000000005e93db in __wt_cache_eviction_worker (session=0x632000003200, 
    busy=0, pct_full=108) at src/evict/evict_lru.c:1485
#15 0x000000000086c781 in __wt_cache_eviction_check (session=0x632000003200, 
    busy=0, didworkp=0x0) at ./src/include/cache.i:242
#16 0x000000000086bb71 in __wt_txn_begin (session=0x632000003200, 
    cfg=0x7fc1b8ff44d0) at ./src/include/txn.i:266
#17 0x000000000085ef80 in __session_begin_transaction (
    wt_session=0x632000003200, config=0x0) at src/session/session_api.c:809
#18 0x00000000004d98ce in ops (arg=<optimized out>) at ops.c:394
#19 0x00007fc1c4d36df3 in start_thread () from /lib64/libpthread.so.0
#20 0x00007fc1c41311bd in clone () from /lib64/libc.so.6
(gdb)  frame 10
#10 0x0000000000741540 in __wt_reconcile (session=0x632000003200, 
    ref=0x60400000ec10, salvage=0x0, flags=3) at src/reconcile/rec_write.c:469
469			ret = __rec_row_leaf(session, r, page, salvage);
(gdb) p page_lock
$35 = 1
(gdb) p page->modify->page_lock
$36 = 13 '\r'
(gdb) frame 4
#4  0x000000000056c3ee in __wt_las_cursor (session=0x632000003200, 
    cursorp=0x7fc1b8feffb0, reset_evict=0x7fc1b8ff0190)
    at src/conn/conn_las.c:206
206			__wt_spin_lock(session, &conn->las_lock);
```

In summary, thread 7 is holding page-lock 10, waiting on page-lock 13, and has the shared LAS cursor locked, thread 4 is hold page-lock 13, and is waiting on the shared LAS cursor lock.

The lookaside file is going to need its own page lock, at a minimum -- I think we're safe as long as we don't acquire any overlapping page-locks after acquiring the LAS cursor.

michaelcahill:
@keithbostic, we've talked a few times about naming.  I don't love "lookaside file/table" but haven't come up with anything that I like enough to warrant the work to change it.

src/conn/conn_las.c
```diff
@@ -0,0 +1,400 @@
+/*-
+ * Copyright (c) 2014-2015 MongoDB, Inc.
+ * Copyright (c) 2008-2014 WiredTiger, Inc.
+ *	All rights reserved.
+ *
+ * See the file LICENSE for redistribution information.
+ */
```
michaelcahill: Can you please move this file to `cache/cache_las.c`?

I'd like to reduce the things in `conn` -- the current grouping doesn't make a lot of sense to me.

src/include/btmem.h
```diff
@@ -6,6 +6,8 @@
  * See the file LICENSE for redistribution information.
  */
 
+#define	WT_RECNO_OOB	0		/* Illegal record number */
```
michaelcahill: This is used in exactly one place: there are lots of 'recno == 0' and `recno != 0` checks through the code, should they be updated to use this?

src/include/btmem.h
```diff
@@ -871,6 +893,7 @@ WT_PACKED_STRUCT_BEGIN(__wt_update)
 	 */
 #define	WT_UPDATE_DELETED_ISSET(upd)	((upd)->size == UINT32_MAX)
 #define	WT_UPDATE_DELETED_SET(upd)	((upd)->size = UINT32_MAX)
+#define	WT_UPDATE_DELETED_VALUE		UINT32_MAX
```
michaelcahill: Maybe change `WT_UPDATE_DELETED_ISSET` and `WT_UPDATE_DELETED_SET` in terrms of `WT_UPDATE_DELETED_VALUE`.

src/include/btmem.h
```diff
@@ -239,11 +259,14 @@ struct __wt_page_modify {
 		 * associated disk image.
 		 *
 		 * Skipped updates are either a WT_INSERT, or a row-store leaf
-		 * page entry.
+		 * page entry; in the case of creating lookaside records, there
+		 * is an additional value, the committed item's transaction ID.
 		 */
+#define	WT_LAS_RECONCILE_UPDATE	'1'
```
michaelcahill: Repeat of line 188?

src/include/btmem.h
```diff
@@ -168,6 +171,23 @@ struct __wt_ovfl_txnc {
...
+ * ID plus an address), the on-page item's transaction ID (so we can discard any
+ * update records from the lookaside table once the on-page item's transaction
+ * is globally visible), a counter (used to ensure the update records remain in
+ * the original order, and the key (either a byte-string or a record number).
+ * The value is simpler: the WT_UPDATE structures transaction ID and value.
+ *
+ * K: '1', file ID, address-len, address, on-page txn ID, counter, key/recno
+ * V: transaction ID, value
+ */
+#define	WT_LAS_RECONCILE_UPDATE	'1'
```
michaelcahill: Maybe 'U' for update?

keithbostic:
@michaelcahill, I don't love it either, and changing it's not a lot of work. I'd probably have already changed it to something else, but I was waiting for you to weigh in.

How about, "transient", "shelf", "attic", "side", "temporary", "short-term", "brief", "stash", "sidestore", "reserve", "hide", "tmpstore", "tmptable", "sidetable"?

Of those, I guess I like "stash", but that might confuse with our use of split-stash bytes/objects.

Part of the problem is I've been planning to use this table for other objects (bulk keys and cached overflow keys and values), and you've has been thinking we'd use different tables for them.

The reason I think one table is better is because the eviction-workers only need 1 cursor, sweep only has to clean out 1 table.

Of course, the current "lookaside" doesn't make much sense for bulk keys, while "tmpstore" works for everything that's not durable.

If you're confident this table is for only for old readers, we should use something specific. How about "reader" or "txn", with "store", "cache", "reserve", "stash", "tmp", so, "readercache", or "txncache"?

src/btree/bt_read.c
michaelcahill: `first_dirty_txn` is only relevant for genuinely dirty pages (for checkpoints).  Restored pages with updates in LAS should start clean, right?

src/include/session.h
```diff
@@ -76,6 +76,11 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_session_impl {
 	WT_CURSOR_BACKUP *bkp_cursor;	/* Hot backup cursor */
 	WT_COMPACT	 *compact;	/* Compact state */
 
+	/*
+	 * Lookaside file cursor, sweep thread and eviction worker threads only.
+	 */
+	WT_CURSOR	*las_cursor;	/* Lookaside file cursor */
+
 	WT_DATA_HANDLE *meta_dhandle;	/* Metadata file */
```
michaelcahill: Maybe we should keep a `meta_cursor` around to match what you are doing with `las_cursor`, rather than caching `meta_dhandle`?

dist/stat_data.py
```diff
@@ -210,9 +214,19 @@ def __init__(self, name, desc, flags=''):
...
+        'pages read into cache requiring lookaside entries'),
     CacheStat('cache_write', 'pages written from cache'),
 
     ##########################################
+    # Lookaside table statistics
+    ##########################################
+    LookasideStat('lookaside_cursor_insert', 'lookaside table insert calls'),
+    LookasideStat('lookaside_cursor_insert_bytes',
+        'lookaside table cursor-insert key and value bytes inserted'),
+    LookasideStat('lookaside_cursor_remove', 'lookaside table remove calls'),
```
michaelcahill: Maybe just group these with cache statistics?

dist/stat_data.py
```diff
@@ -168,15 +172,15 @@ def __init__(self, name, desc, flags=''):
         'maximum bytes configured', 'no_clear,no_scale'),
     CacheStat('cache_bytes_overflow',
         'tracked bytes belonging to overflow pages in the cache',
-        'no_clear,no_scale'),
+            'no_clear,no_scale'),
```
michaelcahill: whitespace oddness?

dist/stat_data.py
```diff
@@ -210,9 +214,19 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_pages_inuse',
         'pages currently held in the cache', 'no_clear,no_scale'),
     CacheStat('cache_read', 'pages read into cache'),
+    CacheStat('cache_read_lookaside',
+        'pages read into cache requiring lookaside entries'),
```
michaelcahill: Maybe change this into `lookaside_read` operations and include `lookaside_read_bytes`, grouped along with the inserts.

dist/stat_data.py
```diff
@@ -247,26 +261,29 @@ def __init__(self, name, desc, flags=''):
...
 
     ##########################################
     # Reconciliation statistics
     ##########################################
     RecStat('rec_pages', 'page reconciliation calls'),
     RecStat('rec_pages_eviction', 'page reconciliation calls for eviction'),
+    RecStat('rec_pages_lookaside',
+        'page reconciliation block requires lookaside records'),
+    RecStat('rec_pages_restore',
+        'page reconciliation block requires in-memory restoration'),
```
michaelcahill: What's the point of these?  Why would they differ from the lookaside insert / read counts?

dist/s_string.ok
```diff
@@ -229,6 +232,7 @@ Posix
 Pre
 Preload
 Prepend
+QIu
```
michaelcahill: There is `WT_UNCHECKED_STRING` to get around this.

examples/c/ex_log.c
```diff
@@ -206,10 +206,14 @@ walk_log(WT_SESSION *session)
...
-		 * in this example.  The metadata is fileid 0.
+		 * connection.
+		 *
+		 * !!!
+		 * We cheat by only looking for fileid 2 in this example; the
+		 * metadata is fileid 0, the lookaside file is fileid 1, making
+		 * the file we created fileid 2.
 		 */
-		if (fileid == 1 && rectype == WT_LOGREC_COMMIT &&
+		if (fileid == 2 && rectype == WT_LOGREC_COMMIT &&
```
michaelcahill: It should be sufficient to check `fileid != 0` and avoid mention of the lookaside file here, I think.  There should never be any log records for the lookaside table, right?

src/conn/conn_las.c
keithbostic: Agreed, and done.

Shall I go ahead and move `conn_cache.c` and `conn_cache_pool.c` into the `cache/` directory as well?

src/conn/conn_sweep.c
```diff
@@ -359,8 +366,14 @@ __wt_sweep_create(WT_SESSION_IMPL *session)
...
 	 * operations for the block manager.
+	 *
+	 * The sweep thread sweeps the lookaside table for outdated records,
+	 * it gets its own cursor for that purpose.
+	 *
+	 * Don't tap the sweep thread for eviction.
 	 */
-	F_SET(session, WT_SESSION_CAN_WAIT);
+	F_SET(session, WT_SESSION_CAN_WAIT |
+	    WT_SESSION_LOOKASIDE_CURSOR | WT_SESSION_NO_EVICTION);
```
michaelcahill: The sweep thread does eviction of dead trees and trees that become idle.  What's the case where it can't do eviction?

michaelcahill:
@keithbostic, I don't mind "stash" but I think it's equally vague.  I don't really want to decide what we're doing with bulk keys or overflow items today but I would like to change the key format so there is less encoding / decoding code.  That's obviously easy to change later depending on what we decide to do about these other cases.

tl;dr -- if you really prefer "stash", that's fine by me but I don't think it's significantly better than "lookaside".

I would like to use "table" consistently rather than sometimes calling it the lookaside *file*.

The comment about `conn/conn_cache*.c` went missing -- they should be merged in a separate branch after this change is merged.

src/btree/bt_read.c
```diff
@@ -8,6 +8,333 @@
...
+	*(char *)p = WT_LAS_RECONCILE_UPDATE;
+	p = (uint8_t *)p + sizeof(char);
+	memcpy(p, &btree_id, sizeof(uint32_t));
+	p = (uint8_t *)p + sizeof(uint32_t);
+	*(uint8_t *)p = (uint8_t)addr_size;
+	p = (uint8_t *)p + sizeof(uint8_t);
+	memcpy(p, addr, addr_size);
+	p = (uint8_t *)p + addr_size;
+	len = sizeof(char) + sizeof(uint32_t) + sizeof(uint8_t) + addr_size;
+	WT_ASSERT(session, WT_PTRDIFF(p, prefix) == len);
```
michaelcahill: Setting `key_format=bIuIQ` (I think?) would replace all of this code with a single `set_key` call.  By setting `counter=0` and `txnid=0`, you could get the same effect as calculating the key prefix.

src/include/btmem.h
keithbostic: I've done a pass. I'm sure it's not complete, but I agree it's better than it was, and we can fix the remaining issues over time.

src/include/btmem.h
keithbostic: Done.

src/include/btmem.h
keithbostic: Yes, removed.

src/include/btmem.h
keithbostic: Done.

src/btree/bt_read.c
keithbostic: No, we call `__wt_row_modify` or `__wt_col_modify` to apply the LAS updates to the page, which will mark the pages dirty.

I hadn't thought about leaving the pages clean. What if we explicitly mark the pages clean? That's less work and we can easily evict these pages without writing them, that's a good thing.

If the page is evicted without being dirtied, the previous LAS records are still good for the next page instantiation; if the page is dirtied/written, then we should do the right thing by writing new LAS records.

I think that change makes sense, I've pushed it. Would you please review 3ae90d2 separately?

Regardless, there's still a question of whether or not the freshly instantiated page needs to set `page->modify->first_dirty_txn = WT_TXN_FIRST`. I don't think so, based on your comment: if/when we dirty the instantiated page, `first_dirty_txn` will be set to the current value, and that's the correct value for checkpoint to care about. I've removed that line in commit 7414d48, please let me know if that's incorrect.

src/include/session.h
keithbostic: So, instead of instantiating a new metadata cursor for each operation, have a shared (locked) metadata cursor? Given the locking we have to do to instantiate the metadata cursor, I expect it would be faster for simple (insert/update/remove) operations.

Sure, if you will confirm that's the plan, I'll code it up.

There are some potentially longer-running operations where we might want to continue to instantiate a new metadata cursor: maybe hot backup, meta-btree-apply.

Should we used the shared cursor for those operations, or instantiate a new one? If we instantiate a new one, should we keep the meta_dhandle around to make that faster, or not bother?

dist/stat_data.py
keithbostic: Done.

dist/stat_data.py
keithbostic: Fixed.

dist/stat_data.py
keithbostic: I don't understand this comment:

> Maybe change this into `lookaside_read` operations and include `lookaside_read_bytes`, grouped along with the inserts.

My intent was to track the pages that require LAS record instantiations. We could subdivide that into "required a search of LAS records" and "required LAS records because the readers still existed"?

Anyway, I'm not sure why we're talking about bytes/inserts here?

dist/stat_data.py
keithbostic: `rec_pages_lookaside` is the count of reconciled pages that required us to create any LAS records, the lookaside insert count is the count of LAS records we created, so it's 1-to-N.

`rec_pages_restore` is the count of reconciled pages where we did an update/restore cycle (that is, we didn't write the page, we found uncommitted records so we cleaned up the page and then re-instantiated it in memory). That's orthogonal to all of this work, but I noticed we weren't tracking it anywhere.

dist/s_string.ok
keithbostic: Fixed.

examples/c/ex_log.c
keithbostic: Fixed; there should never be any log records for the lookaside table.

src/conn/conn_sweep.c
keithbostic: > The sweep thread does eviction of dead trees and trees that become idle. What's the case where it can't do eviction?

When it's using its lookaside table cursor, it can't be grabbed for opportunistic eviction, it could deadlock.

`WT_SESSION_NO_EVICTION` is maybe a bad name, since it doesn't mean "don't evict", it means "don't re-purpose this thread to do other eviction work just because it's waiting on something".

I didn't see a reason to rope the sweep thread into doing opportunistic eviction, so I turned it off in all cases; if there's a reason the sweep thread should do opportunistic eviction, I could toggle the flag when the lookaside table cursor is in use, as I do for other threads (for example, eviction worker threads).

keithbostic:
> tl;dr -- if you really prefer "stash", that's fine by me but I don't think it's significantly better than "lookaside".

OK, I'll leave it alone for now.

> I would like to use "table" consistently rather than sometimes calling it the lookaside file.

Done.

What about "metadata file"? Shall I fix those references as well? There are about 40.

> The comment about conn/conn_cache*.c went missing -- they should be merged in a separate branch after this change is merged.

I'm not sure what you mean, I see the comment about the move's commit.

I can revert that move, but it's a new file, so I'm not sure where the problem is?

src/btree/bt_read.c
keithbostic: > Setting `key_format=bIuIQ` (I think?) would replace all of this code with a single `set_key` call.  By setting `counter=0` and `txnid=0`, you could get the same effect as calculating the key prefix.

It's more complicated than that because the row's key is included in the LAS key. I'll take a run at it, and see where I end up.

keithbostic: @michaelcahill, @sueloverso:

It occurs to me we can't use the lookaside table for more than one purpose unless we make the key/value formats simplistic, that is, they both have to be WT_ITEMs that we parse explicitly, there's no reason to believe the key/value formats will be similar for other purposes.

Or is there some way around that problem?

I'd prefer not to crack the key/value pairs myself (and I'm confident Michael feels the same way).

If that's correct:

* Sue, I think we're going to have more internal tables in our future, we might want to move their file IDs somewhere out of the general file ID namespace.

* There's no reason to put a flag byte at the beginning of the record, I should get rid of WT_LAS_RECONCILE_UPDATE?

* We should rename the lookaside table because it can't ever be used for anything other than old reader information, and lookaside is a lousy name for that.

Thoughts?

dist/stat_data.py
michaelcahill: The stats are looking better, but I'm still confused by how many there are and why they are spread out into different grouping.  I saw the count of inserts and bytes inserted and assumed there would be a count of reads and bytes read.  But that's not what is there (maybe because we don't track those in the data source stats?).

Instead, there are is a `rec_*` stat for page writes and a `cache_*` stat for page reads that use the lookaside table.  Plus there are `cache_lookaside_*` stats for operations on the lookaside table.

I'd like to simplify the stats, if possible, so we know what to look for to decide (1) whether updates are being written to the lookaside table, (2) whether reads are having to check it and (3) if it is growing very large.  Maybe just have `cache_read_lookaside` and rename `rec_pages_lookaside` to `cache_write_lookaside`?

If we need finer-grained information about the numbers of operations or bytes written, we could add them later?

src/conn/conn_sweep.c
michaelcahill: Maybe we can rename that flag `WT_SESSION_NO_LRU_EVICTION`?

michaelcahill:
@keithbostic, I suggest worrying about "metadata file" in a separate branch after this is merged.  Similarly for other structural changes (such as moving files from `conn/*`) that are unrelated to the lookaside table.

src/reconcile/rec_write.c
```diff
@@ -489,6 +486,81 @@ __wt_reconcile(WT_SESSION_IMPL *session,
...
+		 * unchanged since reconciliation started, it's clean.
+		 *
+		 * If the write generation changed, the page has been written
+		 * since reconciliation started and remains dirty; if evicting,
+		 * we've failed.
+		 */
+		if (__wt_atomic_cas32(&mod->write_gen, r->orig_write_gen, 0))
+			__wt_cache_dirty_decr(session, page);
+		else if (F_ISSET(r, WT_EVICTING))
+			return (EBUSY);
```
michaelcahill: Can this actually happen?  If we are evicting, shouldn't we have the page exclusive, which means nobody can have updated `mod->write_gen`?

src/reconcile/rec_write.c
```diff
@@ -909,132 +1041,234 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
-		 * reconciled until modified again.
+		 * Find the first update we can use.
 		 */
-		if (*updp == NULL) {
-			if (__wt_txn_visible(session, txnid))
-				*updp = upd;
-			else
+		if (F_ISSET(r, WT_EVICTING)) {
+			/*
+			 * Eviction can write any committed update.
```
michaelcahill: I think this comment is only true if `WT_EVICT_UPDATE_RESTORE` and `WT_EVICT_LOOKASIDE` are both set?

src/reconcile/rec_write.c
```diff
@@ -909,132 +1041,234 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+			 * snapshot.
+			 *
+			 * When reconciling for a checkpoint, track whether any
+			 * updates were skipped on the way to finding the first
+			 * visible update.
+			 *
+			 * KEITH: I think this may be wrong, we aren't checking
+			 * that we can't write updates committed more recently
+			 * than in-progress checkpoints.
+			 */
```
michaelcahill: Updates more recent than the checkpoint will not be visible, so I think this is okay.  Is there some case I'm missing?

src/reconcile/rec_write.c
```diff
@@ -909,132 +1041,234 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+	 * update, but then find all updates are stable at the end of the scan.
+	 *
+	 * Skip the visibility check for the lookaside table as a special-case,
+	 * we know there are no older readers of that table.
 	 */
-	if (!skipped && __wt_txn_visible_all(session, max_txn))
+	if (!skipped &&
+	    (F_ISSET(btree, WT_BTREE_LAS_FILE) ||
+	    __wt_txn_visible_all(session, max_txn)))
 		return (0);
```
michaelcahill: Is this skipping of the lookaside table just an optimization?  The claim in the comment that "there are no older readers" doesn't make a lot of sense to me: can't there be multiple sessions reading from the lookaside table that started at different points in time?

src/conn/conn_sweep.c
keithbostic: I don't think that's better, "LRU" eviction happens at the top level too.

How about `WT_SESSION_NO_WAIT_EVICTION`, or `WT_SESSION_NO_OPPORTUNISTIC_EVICTION`?

The latter is awfully long, but maybe something explicit that grabs attention is right thing to do here.

src/cache/cache_las.c
```diff
@@ -0,0 +1,390 @@
...
+	 */
+	if (conn->las_cursor == NULL)
+		return;
+
+	/*
+	 * KEITH: is there a cleaner way to get a reference to the underlying
+	 * data handle?
+	 */
+	cstats = conn->stats;
+	dstats = ((WT_CURSOR_BTREE *)conn->las_cursor)->btree->dhandle->stats;
```
michaelcahill: We could cache the LAS btree handle in the connection, if that helps?  Or do a `__wt_session_get_btree` for the LAS?

src/reconcile/rec_write.c
keithbostic: Thanks! I'll fix.

src/include/btree.h
```diff
@@ -146,12 +146,14 @@ struct __wt_btree {
 	/* Flags values up to 0xff are reserved for WT_DHANDLE_* */
 #define	WT_BTREE_BULK		0x00100	/* Bulk-load handle */
 #define	WT_BTREE_IN_MEMORY	0x00200	/* Cache-resident object */
-#define	WT_BTREE_NO_EVICTION	0x00400	/* Disable eviction */
-#define	WT_BTREE_NO_LOGGING	0x00800	/* Disable logging */
-#define	WT_BTREE_SALVAGE	0x01000	/* Handle is for salvage */
-#define	WT_BTREE_SKIP_CKPT	0x02000	/* Handle skipped checkpoint */
-#define	WT_BTREE_UPGRADE	0x04000	/* Handle is for upgrade */
-#define	WT_BTREE_VERIFY		0x08000	/* Handle is for verify */
+#define	WT_BTREE_LAS_FILE	0x00400	/* Look-aside object */
```
michaelcahill: Naming -- can this be `WT_BTREE_LOOKASIDE`?

src/reconcile/rec_write.c
keithbostic: I think it's true if `WT_EVICT_LOOKASIDE` is set; `WT_EVICT_UPDATE_RESTORE` is a way to handle updates we can't write.

src/reconcile/rec_write.c
keithbostic: I wasn't sure if this was right -- this is one of the clauses I really wanted to make sure you reviewed closely (well, this whole function, actually).

michaelcahill:
@keithbostic, I've made it through my review and asked a few more questions, but I'm happy to merge this tomorrow morning my time, if that works for you?

I'd like to bump the mongodb-3.2 branch before merging this one, and I suspect there may be some more tuning of eviction required, but I don't see any problems that should hold up merging any longer.

src/reconcile/rec_write.c
keithbostic: I think it's an optimization: the only readers of the lookaside table are the sweep thread and threads reading a page into cache, that is, there aren't any readers of the table that want anything other than the current data. That's what I meant by "older readers". In other words, there's no point in checking visibility, everything is globally visible by definition. Is there a better way to say that?

src/conn/conn_sweep.c
michaelcahill: What do you mean by "at the top level"?

I think of LRU eviction as exactly the "opportunistic" case you are talking about -- threads other than the eviction workers choose a page from the LRU queue to evict if they can't make progress when the cache is full.

Or we could flip this around, and mark ordinary application sessions with `WT_SESSION_CAN_EVICT`?

src/cache/cache_las.c
keithbostic: I think the call is safe, it just feels ugly to be reaching down through a cursor to a btree cursor to a btree, to a data handle.

I don't think `__wt_session_get_btree` is worth the effort.

src/include/btree.h
keithbostic: Sure, done.

src/reconcile/rec_write.c
michaelcahill: Maybe this logic should go into `__wt_txn_oldest_id` (where we already check things in the btree handle).  That way, it will apply to all visibility checks done when deciding whether to try evicting pages, not just this one decision inside reconciliation.

src/conn/conn_sweep.c
keithbostic: There's a file named `evict_lru.c`, and it implements LRU eviction, I think it would be confusing to refer to the opportunistic case as LRU eviction. The point isn't if the thread can evict or not, the point is if the thread can be re-purposed from whatever it's doing to working on eviction at a particular point in time, right?

I'm happy to flip this around if that's a better approach.

keithbostic:
@michaelcahill, I'm fine with a merge -- I think this branch works, but I also think we've got tuning and some bugs to chase. (To be clear, I've not seen a failure in a few days, but we could let the stress testing pound on the new logging code until Monday, too, and merge then.)

keithbostic:
@michaelcahill, did `evict_lru.c` look OK to you? 

There were some tuning changes there (that went in and then came back out, when we decided the lookaside file would only configure after eviction hit "aggressive"), plus the whole removal of `__wt_evict_page`.

src/cache/cache_las.c
```diff
@@ -0,0 +1,390 @@
...
+	 * moving thread. Try to review the entire lookaside table once every 5
+	 * minutes, or every 30 calls.
+	 *
+	 * The reason is because the lookaside table exists because we're seeing
+	 * cache/eviction pressure (it allows us to trade performance and disk
+	 * space for cache space), and it's likely lookaside blocks are being
+	 * evicted, and reading them back in doesn't help things. A trickier,
+	 * but possibly better, alternative might be to review all lookaside
+	 * blocks in the cache in order to get rid of them, and slowly review
+	 * lookaside blocks that have already been evicted.
```
michaelcahill: We should also set `WT_SESSION_NO_CACHE` through here to avoid polluting the cache with pages from the lookaside table.

src/reconcile/rec_write.c
keithbostic: Sure, I'll push that change.

michaelcahill:
I check `evict_lru.c` around the removal of the `WT_ISO_EVICTION`, which I think is okay because you're now doing explicit `visible_all` calls for eviction and `visible` calls for checkpoints.

One thing I think is missing with the removal of `__wt_evict_page` is a call to `__wt_txn_update_oldest` before starting eviction -- that's going to cause wasted effort if the oldest ID has fallen behind current.

src/cache/cache_las.c
keithbostic: Good point -- do we ever want to cache lookaside table pages we read? I'm guessing not, there's no reason to believe they'll be useful (unless we have a workload where reading one page that needs LAS records implies we're likely to read other pages that need LAS records, and that seems like a stretch).

michaelcahill: I think the less caching of lookaside pages we do, the better.

src/conn/conn_sweep.c
michaelcahill: `evict_lru.c` manages the LRU queue, which application threads can use if they are stuck when the cache is full.  I don't see them as separate things, and I don't think any thread should be "repurposed from whatever it's doing" -- threads only get involved in eviction if they would otherwise be blocked by a full cache (at least with high probability).

Regardless, we're spending a lot of time discussing the naming of one flag.  I've been through and reviewed the use of `WT_SESSION_NO_EVICTION` again, including the pre-existing cases where we had `WT_SESSION_NO_CACHE_CHECK` plus the new cases to avoid re-entering eviction in places that could cause deadlocks.  I don't think changing the name will help at this point since there isn't an obviously better candidate, and `WT_SESSION_NO_EVICTION` matches the existing `WT_BTREE_NO_EVICTION` naming.

Let's leave this one alone.

keithbostic:
> One thing I think is missing with the removal of `__wt_evict_page` is a call to `__wt_txn_update_oldest` before starting eviction -- that's going to cause wasted effort if the oldest ID has fallen behind current.

Should we add a call to `__wt_txn_update_oldest` in `__wt_evict`?

michaelcahill:
> Should we add a call to __wt_txn_update_oldest in __wt_evict?

That's potentially some wasted effort (e.g., when sweep is discarding a whole tree from cache it will update the oldest ID for every page), but it would be better than not doing it at all.  There's no need to call it when evicting clean pages, at least.

src/include/txn.i
```diff
@@ -110,6 +110,10 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
 	txn_global = &S2C(session)->txn_global;
 	btree = S2BT_SAFE(session);
 
+	/* All updates to the lookaside table are immediately visible. */
+	if (btree != NULL && F_ISSET(btree, WT_BTREE_LOOKASIDE))
+		return (WT_TXN_ABORTED);
+
```
michaelcahill: This is causing assertion failures -- `session->dhandle` must be set to the LAS table in some situation where we are committing a transaction.  Rather than attempt to debug that, can I suggest that we leave this out for now and come back to it later?

Another way to get the same effect would be to arrange for all updates to the LAS to have txnid = WT_TXN_NONE: that wouldn't require any special visibility checks for LAS reads.

keithbostic:
>> Should we add a call to __wt_txn_update_oldest in __wt_evict?

> That's potentially some wasted effort (e.g., when sweep is discarding a whole tree from cache it will update the oldest ID for every page), but it would be better than not doing it at all.  There's no need to call it when evicting clean pages, at least.

Take a look at 1e679b0, I tucked the call into `__evict_review`, so we do it after we have exclusive access (that seemed preferable), and only if we're not closing and the page is dirty, immediately before the first visibility tests.

src/include/txn.i
keithbostic: Agreed, let's pull it out for now.

dist/stat_data.py
keithbostic: > I'd like to simplify the stats, if possible, so we know what to look for to decide (1) whether updates are being written to the lookaside table, (2) whether reads are having to check it and (3) if it is growing very large.  Maybe just have `cache_read_lookaside` and rename `rec_pages_lookaside` to `cache_write_lookaside`?

I've reduced it to:

```
Connection only:
    cache_lookaside_insert
    cache_lookaside_remove

Connection & data-source:
    cache_read_lookaside
    cache_write_lookaside
    cache_write_restore
```

For (1), cache_write_lookaside, for (2) cache_read_lookaside, and (3) cache_lookaside_insert vs. cache_lookaside_remove.

I've left cache_write_restore even though it's not related to the lookaside table. That's a new statistics that tells us if we're updating/restoring pages because we had uncommitted updates on the page.

Let me know if that's still not right.

michaelcahill:
Thanks, @keithbostic -- I'm happy with this now, I'll merge into develop and switch the sanitizer stress test back as well.

PR #2073, opened 2015-07-20
WT-1865 - Performance degrades as more files are open

daveh86:


daveh86:
@michaelcahill, would you mind having a look at this and see what you think of the direction this patch is taking?

src/conn/conn_dhandle.c
```diff
@@ -52,6 +52,9 @@ __conn_dhandle_alloc(WT_SESSION_IMPL *session,
 	dhandle->handle = btree;
 	btree->dhandle = dhandle;
 
+	if (WT_PREFIX_MATCH(uri, "file:"))
+		F_SET(dhandle, WT_DHANDLE_IS_FILE);
+
```
michaelcahill: This probably belongs in `bt_handle.c`.

src/evict/evict_lru.c
```diff
@@ -160,14 +160,16 @@ __evict_server(void *arg)
 	WT_DECL_RET;
 	WT_EVICT_WORKER *worker;
 	WT_SESSION_IMPL *session;
+	uint64_t laps;
 
 	session = arg;
 	conn = S2C(session);
 	cache = conn->cache;
+	laps = 0;
```
michaelcahill: This appears to be a constant -- the local variable isn't updated by `__evict_pass`.  What's the idea of counting laps here?

src/evict/evict_lru.c
```diff
@@ -983,24 +985,34 @@ retry:	while (slot < max_entries && ret == 0) {
...
-		if (!WT_PREFIX_MATCH(dhandle->name, "file:") ||
+		/*
+		 * Once every 15 runs we do a full walk. If we aren't doing a
+		 * full walk, then we can declare ourselves finished when we
+		 * find the first entry that isn't to be read.
+		 */
+		if (laps % 15 != 0 && dhandle->evict_skip_until > laps) {
+			cache->evict_file_next = NULL;
+			break;
+		}
```
michaelcahill: What is the point of checking all handles every 15 laps?  Won't that fall out of setting `dhandle->evict_skip_until = laps + 15`?

Or is here because you're not counting the absolute number of laps in the calling code?

src/evict/evict_lru.c
```diff
@@ -1057,17 +1058,31 @@ retry:	while (slot < max_entries && ret == 0) {
...
+		 * skipped for a few runs and remove it and insert to the tail.
 		 */
-		if (slot == prev_slot)
-			btree->evict_walk_period = WT_MIN(
-			    WT_MAX(1, 2 * btree->evict_walk_period), 100);
-		else
-			btree->evict_walk_period = 0;
+		if (slot == prev_slot) {
+skip:			if (dhandle->evict_empty_passes++ > 100 ) {
+				dhandle->evict_skip_until = laps + 15;
```
michaelcahill: The magic number 15 should have a `#define` in `cache.h`.

src/evict/evict_lru.c
```diff
@@ -1057,17 +1058,31 @@ retry:	while (slot < max_entries && ret == 0) {
...
+		 * If we didn't find any candidates in the file, mark it to be
+		 * skipped for a few runs and remove it and insert to the tail.
 		 */
-		if (slot == prev_slot)
-			btree->evict_walk_period = WT_MIN(
-			    WT_MAX(1, 2 * btree->evict_walk_period), 100);
-		else
-			btree->evict_walk_period = 0;
+		if (slot == prev_slot) {
+skip:			if (dhandle->evict_empty_passes++ > 100 ) {
```
michaelcahill: Why try 100 times when finding no candidates before skipping?

src/evict/evict_lru.c
```diff
@@ -1057,17 +1058,31 @@ retry:	while (slot < max_entries && ret == 0) {
...
+				dhandle->evict_skip_until = laps + 15;
+				STAILQ_REMOVE(&conn->dhlh,
+				    dhandle, __wt_data_handle, l);
+				STAILQ_INSERT_TAIL(&conn->dhlh, dhandle, l);
+			}
+		} else {
+			dhandle->evict_empty_passes = 0;
+			/* If we find something of value, move it back up. */
+			if (dhandle->evict_skip_until > laps) {
```
michaelcahill: This seems weird to me: I don't think we should visit files if we've set their `skip_until` in the future.

src/evict/evict_lru.c
```diff
@@ -1057,17 +1058,31 @@ retry:	while (slot < max_entries && ret == 0) {
 			WT_ASSERT(session, session->split_gen == 0);
 		}
 
-		__wt_spin_unlock(session, &cache->evict_walk_lock);
```
michaelcahill: Why does the `evict_walk_lock` need to be held longer?

src/evict/evict_lru.c
```diff
@@ -1029,21 +1041,10 @@ retry:	while (slot < max_entries && ret == 0) {
...
-		    cache->evict_entries >= WT_EVICT_WALK_INCR &&
-		    btree->evict_walk_skips++ < btree->evict_walk_period)
-			continue;
-		btree->evict_walk_skips = 0;
 		prev_slot = slot;
 
 		(void)WT_ATOMIC_ADD4(dhandle->session_inuse, 1);
 		incr = 1;
-		__wt_spin_unlock(session, &conn->dhandle_lock);
-		dhandle_locked = 0;
```
michaelcahill: This change isn't feasible: we do a lot of work to avoid holding `dhandle_lock` when not actively accessing the list of data handles.  I think it should be dropped here and acquired again if the handle needs to be moved to a different place in the list.

michaelcahill:
@daveh86, I've asked a bunch of questions about this change.  I think the `laps` counter should be tracked across calls to `evict_lru_walk` so it keeps moving forward, then there shouldn't be any need to scan all handles every 15 passes.  I'm unclear on why you want to visit a file 100 times finding no candidate pages before starting to skip it.

Also, the handle list lock can't be held across the call to `__evict_walk_file`.

After those changes are made, I think this looks basically right.

daveh86:
@michaelcahill, have made the changes you suggested. The patch runs on-par in the case of the multi-btree test and at a slight performance degradation with the example config provided in the bug report.

Would be interested to know of any other performance test suites you can suggest running this patch against.

src/evict/evict_lru.c
```diff
@@ -927,12 +929,12 @@ __evict_walk(WT_SESSION_IMPL *session, uint32_t flags)
 	WT_DATA_HANDLE *dhandle;
 	WT_DECL_RET;
 	u_int max_entries, prev_slot, retries, slot, start_slot, spins;
-	int incr, dhandle_locked;
+	int incr, dhandle_locked, move_dhandle;
```
michaelcahill: I suggest you make this `enum { NONE, HEAD, TAIL } move_dhandle`.

src/evict/evict_lru.c
```diff
@@ -1060,14 +1079,15 @@ retry:	while (slot < max_entries && ret == 0) {
 		__wt_spin_unlock(session, &cache->evict_walk_lock);
 
 		/*
-		 * If we didn't find any candidates in the file, skip it next
-		 * time.
+		 * If we didn't find any candidates in the file, mark it to be
+		 * skipped for a few runs and remove it and insert to the tail.
```
michaelcahill: I'd say "mark it to be skipped and move it to the end of the handle list"

src/include/cache.h
```diff
@@ -13,7 +13,7 @@
 #define	WT_EVICT_INT_SKEW  (1<<20)	/* Prefer leaf pages over internal
 					   pages by this many increments of the
 					   read generation. */
-#define	WT_EVICT_WALK_PER_FILE	 10	/* Pages to queue per file */
+#define	WT_EVICT_WALK_PER_FILE	 50	/* Pages to queue per file */
```
michaelcahill: Is this change still necessary?  How much of the speedup in the original workload is due to this change?

The reason I ask is that there are workloads where eviction spends too much time in files that don't have many candidate pages.  That was the motivation for keeping this value small.

daveh86: Over a number of runs with the current logic there isn't much noticeable speed up.

I've reverted this part of the change.

michaelcahill:
For the record, I was uneasy about this branch, and after spending a week working on sweeps, I realized why: there are sweep server passes through the handle list that don't lock (because in develop, the sweep server is the only thread that ever removes from the handle list).  That is no longer true on this branch :-(

One solution would be to have sweep reorder the handle list when it sees handles with high eviction skip counts.  That would only be advisory, though -- eviction would still have to walk the whole list (at least when aggressive).

michaelcahill:
We're going to need a different approach, closing this PR.

PR #2074, opened 2015-07-20
Encryption mismatch checks

ddanderson:
Fix to detect the case when encryption is specified when reopening, but the system was originally started without encryption.  Also added tests for the same.


ddanderson:

sueloverso:
Thanks Don lgtm.  I've confirmed it works with the examples changes I made too.

PR #2077, opened 2015-07-22
WT-2014 Checkpoint file fixes

agorrod:
For WT-2014 and WT-2009, which came from durability testing with MongoDB.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -1202,11 +1202,9 @@ __wt_checkpoint_close(WT_SESSION_IMPL *session, int final)
 	 * Turn on metadata tracking if:
 	 * * The session isn't already doing metadata tracking
 	 * * The file isn't bulk loadable
-	 * * The close isn't during connection close or the update is for the
-	 *   metadata file itself (when we need to track to ensure durability).
+	 * * The close isn't during connection close
 	 */
-	need_tracking = !WT_META_TRACKING(session) && !bulk &&
-	   (!final || WT_IS_METADATA(session->dhandle));
+	need_tracking = !WT_META_TRACKING(session) && !bulk && !final;
 
 	if (need_tracking)
 		WT_RET(__wt_meta_track_on(session));
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -531,26 +531,16 @@ __wt_txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_ERR(__wt_txn_commit(session, NULL));
 
 	/*
-	 * Checkpoint the metadata file if this is a system wide checkpoint or
-	 * logging is disabled. Otherwise ensure that the log file has been
-	 * flushed so that the checkpoint is stable on disk.
-	 * It is not OK to checkpoint the metadata if logging is enabled and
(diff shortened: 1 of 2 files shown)


agorrod:
@michaelcahill Would you mind reviewing these changes please? I've split the commits out to make the different aspects of the fix clear.

agorrod:
@michaelcahill I've done some more local testing and I don't think this is ready yet. Please hold off on the review for now.

src/txn/txn_ckpt.c
```diff
@@ -532,26 +531,44 @@ __wt_txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
+		saved_meta_next = session->meta_track_next;
+		session->meta_track_next = NULL;
+		WT_WITH_DHANDLE(session,
+		    session->meta_dhandle, ret = __wt_checkpoint(session, cfg));
+		session->meta_track_next = saved_meta_next;
+		WT_ERR(ret);
+
+		WT_ERR(__checkpoint_verbose_track(session,
+		    "metadata sync completed", &verb_timer));
+	} else if (logging)
```
michaelcahill: fwiw, I think the condition here is redundant.  That is, `logging` will always be true in the else case.

agorrod:
@michaelcahill I haven't been able to reproduce the failure - I suspect I was testing the wrong binary. Please go ahead and review/merge.

src/txn/txn_ckpt.c
```diff
@@ -1174,19 +1191,21 @@ __wt_checkpoint_close(WT_SESSION_IMPL *session, int final)
...
-		WT_ASSERT(session,
-		    bulk || F_ISSET(session, WT_SESSION_LOCKED_SCHEMA));
+	WT_ASSERT(session,
+	    final || bulk || F_ISSET(session, WT_SESSION_LOCKED_SCHEMA));
+
+	/*
+	 * Turn on metadata tracking if:
+	 * * The session isn't already doing metadata tracking
+	 * * The file isn't bulk loadable
+	 * * The close isn't during connection close
```
michaelcahill: Minor nit: "bulk loadable" -> "being bulk loaded" and punctuation.

src/txn/txn_ckpt.c
```diff
@@ -532,26 +531,44 @@ __wt_txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
+		 * If any tree was dirty, we will have updated the metadata
+		 * with the new checkpoint information.  If the metadata is
+		 * clean, all other trees must have been clean.
+		 *
+		 * Disable metadata tracking during the metadata checkpoint.
+		 *
+		 * We don't lock old checkpoints in the metadata file: there is
+		 * no way to open one.  We are holding other handle locks, it
+		 * is not safe to lock conn->spinlock.
+		 */
```
michaelcahill: This comment is pretty stale: we don't have a `conn->spinlock` any more.  Consider merging with the comment in the `else` block and moving up a level.  The main point is "we need to make sure the metadata changes are durable before the checkpoint is resolved (making blocks from old checkpoints available for reallocation)" plus "recovery relies on the metadata's checkpoint LSN only being updated by full checkpoints".

michaelcahill:
Thanks @agorrod, lgtm % comment nits.

michaelcahill:

PR #2078, opened 2015-07-22
WT-2016: wt_rwlock_t union packing

markbenvenuto:
@keithbostic Here is a really trivial change to nest the `us` member in a struct in the union. I need this so that I can add padding into the struct for big-endian systems.


michaelcahill:
@markbenvenuto, lgtm, I'll merge.

PR #2079, opened 2015-07-22
WT-1967 some fixes and suggestions

michaelcahill:
@keithbostic, I've been trying out the multi-cursor branch, here are a few fixes and suggestions (in the last four commits).  It includes a fix for the assertion failure you were seeing: that assertion is too strong now that we can do updates in the service of a page read.

Take all of these changes with a grain of salt: feel free to pick and choose anything here that you like.

This branch does fix connection close, where there were some ordering problems (e.g., dropping the LAS table after closing the metadata), and also fixes the restore code where I believe the updates were loaded in reverse order.  It also includes a suggestion to use `value_format` to avoid a lot of serialization code that you might not like so much.

I'm still seeing various failures after these changes -- I used this as an exercise to see how the branch was working and understand what is left to do.


keithbostic:
@michaelcahill, I merged all of your changes except for 9549aec (it's not I disagree with it, but I haven't yet merged the changes to create a lookaside table cursor per session, I'm still using a global lock in the WT_CONNECTION.impl structure around a single cursor, in the main branch).

keithbostic:
@michaelcahill, I think we're done with this branch -- the only thing left here is the changes to create cursors on demand, and I think we've decided against doing that work.

PR #2082, opened 2015-07-23
WT-2017 Don't shutdown eviction server threads during runtime.

agorrod:
Once there has been enough workload for threads to be started leave
them running. Stopping/starting is expensive and distracting in a
debugger.


michaelcahill:
@agorrod, I'm happy with this change: should I merge?

agorrod:
Yes please. I did some performance measurements with wtperf, and didn't see any measurable difference in throughput.

PR #2084, opened 2015-07-24
WT-2012 Fix a bug updating the oldest ID

agorrod:
We weren't checking that the oldest ID doesn't move past the oldest named snapshot ID.

Found by inspection.


agorrod:
@michaelcahill Could you review this change please? We were checking this before outside of the loop that calculates the ID that is used to bump the global value, then ignoring it when doing the actual update.

PR #2088, opened 2015-07-27
SERVER-19340 Avoid type aliasing in the random number generator.

michaelcahill:


michaelcahill:
@daveh86, can you please review?  This unfortunately touches a lot more code than I'd like...

src/include/misc.h
```diff
@@ -255,3 +255,11 @@
...
 	__wt_page_swap_func(session, held, want, flags)
 #endif
+
+/* Random number generator state. */
+union __wt_rand_state {
+	uint64_t v;
+	struct {
+		uint32_t w, z;
+	} x;
+};
```
keithbostic: @michaelcahill, just sheer curiosity, but why did we have to rename the union to `x`?

michaelcahill: @keithbostic, I screwed up the struct first time around when moving it into the header and wanted to avoid doing the it again in the future.  In other places we use "u" for a union: I'm trying to be consistent here because the "x" here is a struct not a union.

daveh86:
@michaelcahill, using this patch I was able to successfully compile without -fno-strict-aliasing using GCC 5.2. Overall the patch looks fine

PR #2090, opened 2015-07-27
WT-2023: Keep 16-bit values from overflowing into other fields.

ddanderson:


ddanderson:
This is ready for anyone to review.  I also created a WIKI page describing how I think the algorithm is supposed to work: https://github.com/wiredtiger/wiredtiger/wiki/Low-level-read-write-locking-algorithms-for-the-WT_RWLOCK-type. Hope that will help in understanding the code. Please update that and/or let me know if something seems incorrect.

Note that the fixes proposed here don't fix any current bugs (see WT-2023), but just avoiding future problems.

ddanderson:
Please hold off on review for now.

@keithbostic, I just saw your comment on http://locklessinc.com/articles/locks/, and I see that overflow from `users` into `pad` may be intentional.  I still think there's a problem in try_readlock where `readers` can overflow into `users`, I'll modify my branch and comment when I'm ready.

michaelcahill:
@keithbostic, can you please chime in on `try_readlock`?

ddanderson:
@keithbostic, this is ready for review.

src/os_posix/os_mtx_rw.c
```diff
@@ -79,9 +80,11 @@ __wt_try_readlock(WT_SESSION_IMPL *session, WT_RWLOCK *rwlock)
 	l = &rwlock->rwlock;
 	pad = l->s.pad;
 	users = l->s.users;
+	nusers = (uint16_t)users + 1;
 	writers = l->s.writers;
 	old = (pad << 48) + (users << 32) + (users << 16) + writers;
-	new = (pad << 48) + ((users + 1) << 32) + ((users + 1) << 16) + writers;
+	/* Intentionally overflow (users + 1) into pad to handle 64K lockers. */
+	new = (pad << 48) + ((users + 1) << 32) + (nusers << 16) + writers;
```
keithbostic: @ddanderson, do we still need the `nusers` variable?

ddanderson: @keithbostic, perhaps it could be done with casting or bit masking:
```
new = (pad << 48) + ((users + 1) << 32) + (((users + 1) & 0xffff) << 16) + writers;
```

keithbostic:
@ddanderson, I've loaded all this back into my head, and starting with your Wiki page, commented the source code.

First, would you please review 8b260c3 (there should be no source changes other than I used your suggested 16-bit mask instead of the separate variable, I thought it was clearer that way), and then would you please read through the comments for errors?

One specific thing about the comment:
```
/* Intentionally overflow (users + 1) into pad to handle 64K lockers. */
```
I don't think those lines handle 64K lockers. I don't think anything handles 64K lockers because the problem there is potential for two threads to be allocated the same 16-bit ticket, and the overflow field doesn't help with that problem.

I believe the point of the overflow field is to allow us to atomically update an 8B quantity without worrying about the user field overflowing -- I think we could make "overflow" a simple padding byte again, by masking off the "users" field when we increment it so it never overflows.

Frankly, this stuff just makes my head hurt, so I may be wrong here... the next time we're both online, maybe we can do a call and talk it through?

keithbostic:
@ddanderson, one last thought: I've been staring at this for a little while now, and I can't think of any reason for the barrier instruction in `__wt_writeunlock`. Can you? I think I understand why the update of readers/writers needs to be atomic, but I don't understand why there needs to be a barrier between the read and update, no other thread of control can modify them.

keithbostic:
@ddanderson, you should update the Lockless web site -- you found the bug, you should get the credit! :)

keithbostic:
@ddanderson, the only reason I can think of for the barrier in `__wt_writeunlock` is to avoid the compiler figuring out what we're doing and updating the readers and writers fields separately. I think that's unlikely, but possible.

ddanderson:
@keithbostic, Thanks, these are great changes!  It really makes the code much clearer. I'll merge in, along with improvements from @michaelcahill.

ddanderson:
@keithbostic, you said:
> you should update the Lockless web site

I don't think Lockless has the bug, they increment a separate variable, so it wraps without overflow.

PR #2091, opened 2015-07-28
SERVER-19522 Try to evict internal pages with no useful child pages.

michaelcahill:


michaelcahill:
Thanks for the review @keithbostic, I've merged your changes.  I'll check in with @agorrod this morning, but I think we're good to go.

src/reconcile/rec_write.c
```diff
@@ -386,22 +387,38 @@ __wt_reconcile(WT_SESSION_IMPL *session,
...
-	} else
+		page_lock = 1;
+	}
+
+	/*
+	 * Reconciliation reads the lists of updates, so obsolete updates cannot
+	 * be discarded while reconciliation is in progress.
+	 */
+	for (;;) {
+		F_CAS_ATOMIC(page, WT_PAGE_SCANNING, ret);
```
agorrod: @keithbostic Is there any disadvantage to setting the WT_PAGE_SCANNING flag when ```compact_in_memory_pass``` is true? I don't see it (i.e: I don't see a path from either update_serial to __wt_reconcile or vice versa) but I thought it worth asking explicitly.

agorrod:
I had one question - apart from that this lgtm - I'd be happy to merge it.

src/reconcile/rec_write.c
keithbostic: @agorrod, I didn't see anything (and I looked); I think they're entirely orthogonal.

PR #2092, opened 2015-07-28
Minor lint and clarification.

keithbostic:
@michaelcahill, this is for your review, take it if you like it, discard it if you don't.

In short, I had to stare at this to convince myself `page` was safe, and I can see a future change where using `page` in some other way in this function would get us into trouble.


michaelcahill:
Sure, fbm.  I'll merge.

PR #2093, opened 2015-07-28
reconciliation locking

keithbostic:
SERVER-19522, WT-2024

Alex, Michael: I think there's a bug in reconciliation locking. I opened WT-2024 to explain it, and this is what I'd suggest as a fix. (I might be completely wrong about the bug, though, I may have missed something?)

The other change here is to lift the split-locking out of the internal-page reconciliation functions and into the upper-level function, so it's done with the rest of the locking, that seems safer and cleaner to me.

For your review and consideration, feel free to just discard all of this if I'm wrong.

@michaelcahill, @agorrod


michaelcahill:
OK, I agree that it looks like there was a potential race between compact removing obsolete updates.  I'm not sure if compact ever walks an update list, but it's better to be safe and consistent.

PR #2094, opened 2015-07-29
WT-2025 SERVER-19573 Change row-store inserts to avoid page locking.

michaelcahill:
Also switch to always using an atomic operation to swap inserts into the skiplist.  This fixes a potential race where readers may see a partially initialized WT_INSERT object, because the spinlock we used in the past for all inserts did not imply a write barrier.


michaelcahill:

michaelcahill:
This lgtm with @keithbostic's changes.  I'll merge.

PR #2095, opened 2015-07-30
WT-1967 Don't allow last_running to move past a running transaction.

michaelcahill:


michaelcahill:
@keithbostic, can you check if this fixes the assertion failure you've been seeing on develop?

keithbostic:
Yes, it does -- thanks!

PR #2096, opened 2015-07-30
WT-1983 wtperf_truncate - fix issues with uint64 -> int64 conversion

daveh86:


michaelcahill:
OK, thanks @daveh86 -- I believe this is cleanup from WT-1983, so I've marked it as such.  I'll merge.

PR #2101, opened 2015-08-02
WT-2023: Mutex overflow simplify

keithbostic:
@ddanderson, @michaelcahill, this change is the follow up on my comment in the conversation:

> I think we could make "overflow" a simple padding byte again, by masking off the "users" field when we increment it so it never overflows.

Rather than masking when allocating tickets, I changed the code to atomically update just the ticket which should give us the correct overflow behavior (we already depend on that behavior elsewhere in this code). I **think** that's safe, emphasis on **think**.


michaelcahill:
I like not having to think about padding, and I agree that we rely on incrementing 16-bit fields elsewhere, so this lgtm.  I'll merge.

PR #2102, opened 2015-08-03
WT-2029 Improve scalability of statistics

michaelcahill:
An implementation for WT_STAT that uses 'per-thread' array counters to avoid scalability bottlenecks in some workloads.


src/include/stat.h
```diff
@@ -6,35 +6,169 @@
...
+
+/* Set all the values in the array counter slots to zero. We are
+ * doing more work than needed by using memset, because we are setting
+ * both the values and the padding. However, resetting the counters is
+ * not a common case operation, so we use memset for compactness.
+ */
+#define	WT_STAT_ALL_RESET(stats, fld)  do {				\
+		memset((stats)->fld.array_v, 0,                         \
+		       sizeof(padded_counter_t) * WT_COUNTER_SLOTS);	\
+} while (0)
```
keithbostic: @fedorova, did you measure this, I'm wondering if memset might actually be faster than explicit strides through the array?

Anyway, we clear memory relatively often, so if memset is slower, I'd be inclined to write the inline function instead, since it's trivial to do.

Thoughts?

src/include/stat.h
```diff
@@ -6,35 +6,169 @@
...
+ * a heuristic for the number of CPUs. Unfortunately, However, inside WT we do
+ * not know when the application creates its threads.
+ *
+ * Our solution is to simply use a fixed number of slots that roughly
+ * approximates the largest number of cores we expect to see on the machine
+ * where WT is used. All we want is to avoid all threads writing the same
+ * variable at the same time, and that solution does the trick.
+ *
+ */
+#define	WT_COUNTER_SLOTS 24
```
keithbostic: @fedorova, how did you determine to use 24, did you run any tests on larger/smaller numbers?

fedorova: @keithbostic, using a smaller number would create more conflicts, using a larger number will use more memory. I have not run the tests to determine when these effects become prominent. I picked 24, because at this point in time it is not very likely that we will encounter lots of machine with more than 24 cores in production. (They do exist, but they are expensive and few people use them.) Would you like me to run tests across the range of values so we better understand the implications?

src/include/stat.h
fedorova: @keithbostic: Haven't measured, because did not realize that resetting counters is a common operation (based on code reading). I can certainly measure. Could you recommend tests that perform lots of counter resetting? Are there any in the tree? My test that suffers from scalability loss due to stats counters, which I've been primarily using to test this branch, does not reset counters, so it's not the right test to evaluate this effect.

src/include/stat.h
keithbostic: @fedorova, I don't think there's a reason to run tests now -- I agree with your view, but I wanted to add something to stat.h that explains how we got to 24 if/when we revisit in the future.

src/include/stat.h
keithbostic: @fedorova, applications can configure `statistics=clear` to `wiredtiger_open`, which resets statistics after they are gathered, where appropriate (for example, a cache size statistic is not cleared, while the count of cursor insert operations will be cleared). So, those applications are clearing the stats every time they're gathered.

It's probably simplest to build a standalone C program that just compares memset to clearing each value, and run it for a million loops? I'm also suspicious that different compilers and different library implementations of memset (and even different underlying instruction sets on different x86 and other architectures, we're not limited to the x86), will come into play.

Since figuring out the "right" answer sounds like a ton of work, and I doubt it's worth that much effort, I'd be happy to go with your best guess on this one for now -- if you think memset is likely faster, we'll go with it, if not, we can write the inline function to clear the values.

src/include/stat.h
```diff
@@ -6,34 +6,160 @@
...
+	 * Additionally, the aggregation can go negative (imagine a thread
+	 * incrementing a value after aggregation has passed its slot and a
+	 * second thread decrementing a value before aggregation has reached
+	 * its slot). Limit the return to 0, negative numbers will just look
+	 * really, really large.
+	 */
+	if (aggr_v < 0)
+		aggr_v = 0;
+	return ((uint64_t)aggr_v);
+}
```
keithbostic: @fedorova, I think we can sum to a negative value, so I think we should bound the return to 0. Does that look correct to you?

fedorova: lgtm

src/include/stat.h
fedorova: Keith, 

My intuition fails me here. I’ll run tests on a couple of machines this week-end to get more insight.

src/include/stat.h
fedorova: Would you like me to change the comment before this variable or are you comfortable with changing the wording yourself?

keithbostic: @fedorova, I'll update it -- once I think we're ready to pull, I'll ask you to do one more pass to make sure I've not broken anything.

src/include/stat.h
keithbostic: @fedorova, don't bother running those tests for now -- I'm going to see if I can get rid of all of that stuff, WT_STATS is a fixed-size object, it shouldn't be too difficult.

keithbostic:
@fedorova, @michaelcahill: I think I'm done in this branch. It passes the test suite, but I haven't tested it against the graphing scripts yet.

Sasha, we got rid of the need for the atomic variants of the statistics function, which removed the need for the slot-0 magic, so things are bit simpler. Would you please do a review and confirm you're happy with my changes in stat.h?

Michael, because we're locking down 23, 64B aligned slots per WT_STATS structure, per data-source, it's roughly 130KB per data-source. Is that acceptable? And, we might may need to include that in the cache size calculations, which I'm not currently doing.

This will need to be backported, I'm afraid -- I'm happy to do that work if you approve the change.

keithbostic:
@michaelcahill, @fedorova: there's a problem here: we copy the statistics into a cursor when returning statistics, and that's not working, we have to aggregate in order to make that work. It's not a trivial fix, and I'm looking at it.

@fedorova, something Michael suggested: instead of padding each value and having a statistics structure with N values, what would you think of doing it in reverse, that is, instead of creating N, padded values, create N, padded statistics structures, each with one value? The reason this might be interesting is the N, padded values requires about 130KB per data-source handle, and we have lots (and lots!) of data source handles. If we created N, padded statistics structures, we'd need less memory. Do you think that approach would work?

fedorova:
@keithbostic: Creating N padded statistics structures for each counter will work: things will scale as long as threads update different structures and those structures are not on the same cache line. I just don’t see how this will save memory: we are still using N * CACHE_LINE_SIZE bytes for every counter in both cases.

keithbostic:
I explained that badly:

N padded counters is 1 statistics structure, with roughly 100 statistics, each of which is `(24 slots * 64B each slot)`, so about 150KB.

N padded statistics structures would be 24 statistics structures, each with roughly 100 statistics, each of which is 8B (the counter isn't padded in any way), so about 20KB.

Instead of using the session ID to select a slot, we use the session ID to select a statistic structure.

fedorova:
Makes sense. I don’t see why this would work any differently from the existing solution in terms of cache conflicts. We should definitely try this.

fedorova:
@keithbostic: Keith, looked at your changes in stat.h. Happy with them. Re-ran my performance test. All good: same scaling as my original implementation.

keithbostic:
I'm taking a run at Michael's idea now -- it's not simple, I'm afraid.

I think what the second design argues is updates to statistics aren't uniform, that is, two adjacent statistics values aren't equally likely to be touched. In which case, having two different statistics in the same cache line isn't necessarily bad. Further it argues that only a few statistics are hot (cursor search count), and by creating multiple statistics structures, we ensure multiple sessions can update the hot statistics without colliding.

The first design works in all cases; the second design works as well as the first if only one of each group of 8 statistics sharing a cache line is hot. The second design degrades as the number of hot statistics on a cache line increase.

I suspect related stats are likely to be roughly equivalent in hotness, we may do the wrong thing at the moment by trying to group related stats together. I wonder if we'd do better by randomizing the order of statistics in the structure to randomize access across the memory.

fedorova:
Keith, 

The way I look at it is as follows:

We want to minimize the number of simultaneous updates of the same cache line by different threads — let’s call them collisions. 

With the first design there is a collision only when the following two conditions are met:
1) Different threads’ session IDs hash to the same slot
2) The threads are updating the same statistic value. 

With the second design, there is a collision when:
1) Different threads’ session IDs hash to the same slot
2) The threads are updating one of the eight values that happen to live in the same cache line. 

So the probability of a collision is slightly higher in the second case, but how much higher — that’s for us to guess. It all depends on the workload. 

I can’t make a call for one design over the other using just that information. For all practical purposes they might be equivalent in terms of scalability. However the second design definitely saves us memory. How important is it to save 100KB of memory per data source handle? If it is very important, I would go with the second design.

keithbostic:
Close #2102, we're going with #2122 instead.

PR #2103, opened 2015-08-03
WT-2023 Avoid bit shifts during read-write lock operations.

michaelcahill:
I hope this doesn't muddy the waters, but I've wondered this before.  Is there any reason not to avoid the shift operations entirely?

### DIFF
--- a/src/os_posix/os_mtx_rw.c
+++ b/src/os_posix/os_mtx_rw.c
@@ -63,7 +63,7 @@
  * next number available. However, instead of waiting for 'writers' to equal
  * their number, they wait for 'readers' to equal their number.
  *
- * This has the effect of queueing lock requests in the order they arrive
+ * This has the effect of queuing lock requests in the order they arrive
  * (incidentally avoiding starvation).
  *
  * Each lock/unlock pair requires incrementing both 'readers' and 'writers'.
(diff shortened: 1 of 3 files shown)


michaelcahill:
@keithbostic, @ddanderson can you please both take a look?

src/os_posix/os_mtx_rw.c
```diff
@@ -63,7 +63,7 @@
  * next number available. However, instead of waiting for 'writers' to equal
  * their number, they wait for 'readers' to equal their number.
  *
- * This has the effect of queueing lock requests in the order they arrive
+ * This has the effect of queuing lock requests in the order they arrive
```
michaelcahill: this is according to `aspell` on my system...

michaelcahill:
p.s. this should also make the code endian-independent.

src/os_posix/os_mtx_rw.c
```diff
@@ -159,22 +156,17 @@ __wt_try_readlock(WT_SESSION_IMPL *session, WT_RWLOCK *rwlock)
...
 	/*
 	 * The old and new values for the lock: note the use of "users" instead
-	 * of "writers", this may not be the lock's current value, rather it's
-	 * the value the lock must have if we are to grant this write lock.
-	 *
-	 * Note the masking of "users + 1" in the calculation of the new value:
-	 * we want to set the readers field to the next readers value, not the
-	 * next users value, so it has to wrap, not overflow.
+	 * of "readers", this may not be the lock's current value, rather it's
+	 * the value the lock must have if we are to grant this read lock.
```
michaelcahill: I think we should make this comment change regardless of the code changes: there was a copy/paste swap of "write" for "read" that I didn't spot on my review.

keithbostic:
@michaelcahill, @ddanderson, this looks correct to me, and I think it's worth the price of admission just remove the endian dependency for read-write locks.

I can't imagine it's slower either, the compiler should be generating the same masking instructions that we were doing by hand. (And, we may do less instructions because the fast test from c76a969 means the old value is correct, so there's no need to touch it, we just have to create the new value.)

PR #2104, opened 2015-08-03
WT-1481 Fix and clarify some sweep statistics.

sueloverso:
@michaelcahill Please review these sweep stats changes.  See if the names/descriptions are more informative.


sueloverso:
@michaelcahill The statistics (`dh_conn_handles_removed`) are indicating that we only remove them from the list on close (according to test_sweep01).

michaelcahill:
Thanks, @sueloverso, I fixed the test case and did a little more tweaking of the names so they all have a "sweep" prefix.  LGTM now, I'll merge...

PR #2105, opened 2015-08-03
row-insert-nolock review

keithbostic:


michaelcahill:
Thanks, @keithbostic -- lgtm.

PR #2106, opened 2015-08-04
WT-2025 Assert that cursors are positioned for inserts

michaelcahill:
Fix a possible race between two threads inserting into a clean page.


michaelcahill:
@keithbostic, as discussed, for your review.

keithbostic:
autobuild: please retest this

keithbostic:
@michaelcahill, this is another example of interesting dependencies: we depend on `WT_PAGE_UNLOCK` being a write-barrier (it's what protects our increment of the last record number in the tree), but we don't document that anywhere. I've just spent some time staring at the serial functions and I'm pretty confident they're correct, but there are a few things I want to talk through with you next chance we get.

PR #2108, opened 2015-08-05
WT-2032: WT_CURSOR.next configured with next_random=true on insert-only pages

keithbostic:


michaelcahill:

PR #2112, opened 2015-08-06
WT-2025: inline atomic functions.

keithbostic:
@michaelcahill, @agorrod, @sue, @markbenvenuto:

I thought it would be interesting to see how bad/good it was to inline the atomic functions, this branch builds pretty cleanly and appears to run.

The interesting things:
* It's not ready to merge, I just hacked my way through it -- the macro-to-function-call changes are pretty easily reviewed, but the structure field changes would need to be carefully looked at.
* We've made some mistakes, we're passing enum and ints to atomic functions, assuming they're 4B in size. **[Update: that's not correct, we asserted the size of the objects, so we'd catch a problem during the build on any machine/compiler where there was a problem.]**
* We could probably use defines, or auto-build the inline functions, I didn't bother trying.
* We need int32/uint32 variants because we do atomic instructions on both.
* Clang doesn't complain if you pass a `volatile XXX` to a function expecting `XXX`, but flexelint does, so I'm casting those fields as part of passing them into the atomic functions. I'd probably want to fix that in the mix, there's no reason we can't have atomic variants that take `volatile XXX` as arguments.
* I also didn't touch the Windows msvc.h file, it's scary inside.

In general, I like this change, it found several problems, and will ensure we don't get caught again on #2111, multiple macro evaluations.


keithbostic:
@markbenvenuto, would you be willing to review 09424e9?  No rush or anything, it's not even merged into develop yet.

michaelcahill:
@keithbostic, while we're considering a big change, how would you feel about naming the functions to match the types, with sizes in bits rather than bytes?

In other words, we'd have `__wt_atomic_add64` for 64-bit ints, rather than `__wt_atomic_add8`.

keithbostic:
> In other words, we'd have __wt_atomic_add64 for 64-bit ints, rather than __wt_atomic_add8.

Sure, I'll push that.

markbenvenuto:
@keithbostic The first argument to `_InterlockedCompareExchange64` was correct.
This is the correct line:
```
	return (_InterlockedCompareExchange64(vp, (int64_t)new, (int64_t)old) == ((int64_t)old));
````

keithbostic:
@markbenvenuto, thank you!

Pushed.

michaelcahill:
@keithbostic, this lgtm, I'm happy to merge if you're prepared to backport?

keithbostic:
Sure -- in what branch should I make the change? And is there any specific testing to do for backports?

michaelcahill:
FTR, please backport to `mongodb-3.0` using the command `git cherry-pick -x -e -m 1 <merge commit>` and make sure the commit message still has the jira ticket at the beginning.

For testing, please smoke test locally, then after pushing, make sure "wiredtiger-test-mongodb-3.0" is running on Jenkins.

keithbostic:
@michaelcahill, merged, pushed, and "wiredtiger-test-mongodb-3.0" is queued up on Jenkins.

PR #2113, opened 2015-08-08
WT-2036 Make handle sweeps more robust

michaelcahill:
Clear time-of-death in handles with more than one reference, don't sweep handles with the exclusive flag set.


michaelcahill:
I'm going to merge these changes so they get some stress testing, then backport to the mongodb-3.0 branch.

PR #2114, opened 2015-08-10
WT-2037 Only write a checkpoint to the log on close if it wasn't

sueloverso:
recovery with an error.

@michaelcahill Please review.


michaelcahill:

PR #2115, opened 2015-08-11
WT-2038 Avoid long scans holding the handle list lock.

michaelcahill:


michaelcahill:
This has passed tests, I'm going to merge so it gets stress testing, then backport to the mongodb-3.0 branch.

PR #2118, opened 2015-08-13
WT-2038 Use TAILQ for all lists.

michaelcahill:
Otherwise remove is O(N), which may not show up until running real workloads (while holding locks).

### DIFF
--- a/dist/s_style
+++ b/dist/s_style
@@ -46,6 +46,11 @@ else
 		cat $t
 	fi
 
+	if ! expr "$f" : 'src/include/queue\.h' > /dev/null &&
+	    egrep 'STAIL_|SLIST_|\bLIST_' $f ; then
+		echo "$f: use TAILQ for all lists"
+	fi
+
 	if ! expr "$f" : 'src/os_posix/.*' > /dev/null &&
 	   ! expr "$f" : 'src/os_win/.*' > /dev/null &&
 	   ! expr "$f" : 'src/include/extern.h' > /dev/null &&
--- a/dist/s_style
+++ b/dist/s_style
@@ -47,7 +47,7 @@ else
 	fi
 
 	if ! expr "$f" : 'src/include/queue\.h' > /dev/null &&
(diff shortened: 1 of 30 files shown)


michaelcahill:
@keithbostic, @sueloverso, I'd appreciate it if you can both take a look at this.  The code changes are all boilerplate: I've converted every linked list to a `TAILQ`.

The immediate motivation here is the sweep server: with MongoDB's settings, it's easy to end up with hundreds of thousands of handles by the time 100K seconds have elapsed, and sweep tries to remove many of them (partly because checkpoint "lock" handles are not being removed sooner, but that's a separate problem for now).

It has taken me over a week to realize that we were using `SLIST` for the list containing all of the handles, and that removing a handle from that list involves a linear scan.  I don't want to have to do that again, so what I'm proposing here is that we use `TAILQ` for everything.  I'm prepared to pay the extra pointer.  Let me know if you disagree.

michaelcahill:
P.S.  This change has to be backported, both because we're fixing a problem in 3.0 and because this touches enough code that it will conflict with most patches we'll choose to do in future.

dist/s_style
```diff
@@ -46,6 +46,11 @@ else
 		cat $t
 	fi
 
+	if ! expr "$f" : 'src/include/queue\.h' > /dev/null &&
+	    egrep 'STAIL_|SLIST_|\bLIST_' $f ; then
```
sueloverso: That needs to be STAILQ_.

src/include/connection.h
```diff
@@ -22,7 +22,7 @@
 struct __wt_process {
 	WT_SPINLOCK spinlock;		/* Per-process spinlock */
 
-					/* Locked: connection queue */
+					/* Locked: connection list */
```
sueloverso: This comment should have stayed as 'queue'.

sueloverso:
Ouch!  I pushed a change for my minor comments.  @keithbostic once you review, you can merge.

PR #2120, opened 2015-08-14
WT-2038 Discard checkpoint lock handles asap

michaelcahill:


michaelcahill:
I ran a pretty simple workload to see what was going on.  With 1000 tables all being frequently updated, with `close_idle_time=60` and `checkpoint.wait=5`, we were seeing:

handle-count-before

(i.e., # handles = 12x # tables).

With these changes, I see: 

handle-count-after

(i.e., # handles = 1-2x # tables).

michaelcahill:
Note that in MongoDB, instead of 12x, we may see (100,000 seconds timeout / 60 seconds per checkpoint) == 1666x more handles than active tables.

src/conn/conn_sweep.c
```diff
@@ -235,21 +239,15 @@ __sweep_remove_handles(WT_SESSION_IMPL *session, time_t now)
...
-		    now <= dhandle->timeofdiscard + conn->sweep_idle_time)
+		if (!WT_DHANDLE_CAN_DISCARD(dhandle))
 			continue;
 
 		WT_WITH_HANDLE_LIST_LOCK(session,
 		    ret = __sweep_remove_one(session, dhandle));
 		if (ret == 0)
 			WT_STAT_FAST_CONN_INCR(session, dh_sweep_remove);
-		else {
+		else
```
sueloverso: We should add a comment why non-zero return means we should increment the sweep reference stat (i.e. the handle was idle and by the time we grabbed the lock to remove it, someone started using it again).  Do we only expect EBUSY in that case?  If so maybe we should assert that or only increment the stat on that error?

src/conn/conn_sweep.c
```diff
@@ -8,6 +8,10 @@
 
 #include "wt_internal.h"
 
+#define	WT_DHANDLE_CAN_DISCARD(dhandle)					\
```
sueloverso: Maybe move to dhandle.h - possibly can be combined in part with WT_DHANDLE_INACTIVE.

sueloverso:
Thanks Michael.  These changes reflect our conversation last night and those numbers look so much better.  I'll merge.

PR #2122, opened 2015-08-15
WT-2029, improve scalability of statistics

keithbostic:
Instead of using per-thread arrays of counters (see #2102), use arrays
of statistics structures. I started with the scalable_stats branch, so
this incorporates most of the commits on that branch, but the diff is
against the develop branch to make it easier to review.

Specifically, this includes a9675d0:
	There are only 3 statistics we maintain atomically: the count
	of open cursors and the split-stash bytes and object counts.
	Move them to WT_CONNECTION_IMPL fields, and copy them into the
	statistics on demand, this allows us to remove support for
	atomically maintaining statistics values.

And 6bda287:
	Get rid of the WT_CONN_{STAT,STAT_GET,STAT_SET} macros, instead
	maintain a few additional fields in the WT_CONNECTION_IMPL
	structure (checkpoint min/max, recent and total time).

And 7590930:
	Use a prime number of buckets rather than assuming a good hash
	(Reference Sedgewick, Algorithms in C, "Hash Functions").

And 63a2082:
	I'm seeing a roughly 20% advantage of using a loop to clear the
	stats values, vs. a memset to clear the whole padded chunk, and
	I'd rather err on the side of writing fewer cache lines, anyway.

Additionally:

Now we have arrays of statistics structures, separate out each statistic
value's description into a separate array, a struct of strings. We could
access it directly, and maybe should, but I'm using a function call at
the moment.

Now the statistic's descriptions are separate from the statistics values,
we no longer need macros to access the fields, in general we can simply
use the stat structure reference and the field name and set the int64_t
directly. This includes removal of the WT_STATS typedef.

Change the LSM code to not maintain a local data-source statistics
structure in each LSM-tree structure -- that allows us to remove all of
the macros to update a single stats structure, and I don't think it
complicates things that much. The changes to the LSM statistics code
should be closely reviewed.


keithbostic:
@michaelcahill, @fedorova, this is a branch that creates multiple statistics structures instead of multiple counters. As Michael expected, this approach reduces the memory required by a lot: it's roughly 16KB per data handle, 25KB per connection handle.

Sasha, can you please try running your benchmark against this branch?

fedorova:
Keith, retested the latest in scalable_stats_keith. Overall, we still see good scaling. On my machine, the 24-thread test is 15% slower than my original branch (5% stdev), but that's probably because the slots are set to 23 and I have one of those rare machines with 24 cores. So I wouldn't worry about this. A couple of data points are significantly faster (by 8%) than the previous design, possibly because we are using less memory.

michaelcahill:
I'm happier with this approach, the code changes look good to me.  If you are both agreed, let's close down #2102 and merge this one.

PR #2123, opened 2015-08-16
WT-2045 fix a deadlock in the eviction server

keithbostic:


src/evict/evict_lru.c
```diff
@@ -1521,7 +1521,18 @@ __wt_cache_eviction_worker(WT_SESSION_IMPL *session, int busy, int pct_full)
...
+		 * again; if we've waited for a second, decrement the counter.
+		 * The decrement is defensive (otherwise the loop might never
+		 * terminate), and there have been bugs in other parts of the
+		 * code causing the wrong thread to infinitely loop here.
+		 */
+		if (++wait_count > 10) {
+			wait_count = 0;
+			if (--count == 0)
+				return (0);
+		}
```
michaelcahill: I understand the motivation, but it seems like this will allow the cache to grow in situations where it is bounded today.  It will grow slowly (1 operation per second per thread), but we're going over the allocated cache size.

How much testing of this have you done?  Was there a particular workload that prompted this part of the change?

michaelcahill:
@keithbostic, the changes to `bt_page.c` lgtm.  I have some questions about the changes to `__wt_cache_eviction_worker` that I'd like to discuss before those changes are merged.

src/evict/evict_lru.c
keithbostic: Well, the cache would only grow if there are no pages to evict in the queue, that is, the queue is empty for a second.

There was no workload that prompted this change, it's purely defensive in case the eviction server ever gets down here again with an empty queue.

I'm only interested in avoiding the infinite loop.

We could clear the wait-count if we find anything at all to evict, or we could increase the wait time to something much longer, maybe 10 seconds?

keithbostic:
I've cherry-picked 7c8ce54 and 7540ec4 into develop.

michaelcahill:
As discussed, we'll close this now that the `bt_page.c` changes are cherry-picked.  We talked about setting a session flag in the eviction server to avoid getting stuck in here if we ever see that happen.

PR #2126, opened 2015-08-18
SERVER-19989 Different approach to dhandle-barrier.

keithbostic:
@michaelcahill, I think we could potentially have this problem in other places. For example, `__conn_add_collator` builds a `WT_COLLATOR` structure and then wraps the insert on the TAILQ list inside the connection's API spinlock. If that spinlock doesn't imply a write-barrier, I think we could theoretically lose.

I don't believe we have any tailq inserts in performance paths, what would you think of including a write-barrier in the TAILQ insert macros? It's kind of brute force, but now that we've standardized on TAILQs in all cases, it might be a reasonable solution.

There are two commits in this branch, only the second one is interesting, the first one just removes everything but TAILQ from queue.h, which I think is reasonable if our queue.h is going to diverge.


keithbostic:
Reference #2124.

michaelcahill:

FTR, I did check other uses of TAILQ, and I don't believe they have problems today.  Spinlocks are sufficient if readers and writers always use them -- the problem with the list of data handles is that there are readers including sweep that don't acquire the spinlock.  That said, this approach will avoid reintroducing similar problems in future and as you note, we shouldn't be doing this in performance paths.

PR #2127, opened 2015-08-19
WT-147. For dynamic index creation, solve a locking issue with LSM.

ddanderson:
This branch was built upon index-create, which does dynamic index creation.

For dynamic index creation, solve a locking issue with LSM.  The issue is that the application thread is holding the schema lock while we are filling the index.  If the index is LSM, then a different LSM thread wants to split the file eventually, but it needs the schema lock to do that.

Coordinate the locking by tagging the LSM cursor with a special flag: "I'm holding the schema lock while filling this lsm: cursor". That in turn sets a flag on the LSM file, so when the LSM worker thread comes along, it pretends it's already holding the schema lock - it's being held on its behalf by the app thread.


ddanderson:
Note: this is a variation of pull request #2007 intended to fix a problem brought up in review.

michaelcahill:
Replaced this pull request with #2134, merging into `index-create`.

PR #2129, opened 2015-08-19
SERVER-18356 - Avoid reading baseconfig file when config_base is false

daveh86:


src/conn/conn_api.c
```diff
@@ -1872,7 +1877,10 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
 	    WIREDTIGER_VERSION_MAJOR, WIREDTIGER_VERSION_MINOR) >=
 	    (int)sizeof(version), ENOMEM);
 	__conn_config_append(cfg, version);
-	WT_ERR(__conn_config_file(session, WT_BASECONFIG, 0, cfg, i1));
+
+	/* Ignore the base_config file if we config_base set to false. */
+	if (config_base_set)
```
michaelcahill: Our style is to only do `if (var)` for simple `int` vars.  So either make `config_base_set` an `int`, and set it above with `config_base_set = (cval.val != 0);` or just change this to `if (config_base_set != 0)`.

michaelcahill:
Thanks, @daveh86 -- one style nit, otherwise lgtm.

michaelcahill:
Two further points on a second reading.

1. I think we need to change `dist/api_data.py` to include a sentence explaining that `config_base=false` suppresses reading `WiredTiger.basecfg` but only when passed to `wiredtiger_open` directly -- i.e., not via the `WIREDTIGER_CONFIG` environment variable.
2. We need a sentence in `src/docs/upgrading.dox` explaining that this is a behavior change.

michaelcahill:
Thanks, @daveh86 -- I've tweaked the docs slightly, I'll merge...

PR #2131, opened 2015-08-19
Pythonic way for samples

JulienBalestra:
Light changes to fit PEP8 (py-RFC) and best practices of Python language
More about it here: http://legacy.python.org/dev/peps/pep-0008/

### DIFF
--- a/examples/python/ex_access.py
+++ b/examples/python/ex_access.py
@@ -50,6 +50,6 @@ cursor.insert()
 # Iterate through the records
 cursor.reset()
 for key, value in cursor:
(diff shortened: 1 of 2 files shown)


wtbuild:

JulienBalestra:
You also may use "%d" instead of recast value with str( int or float )

michaelcahill:
autobuild: ok to test

michaelcahill:
Thanks for these cleanups, @JulienBalestra -- they look good to me, I will merge them now.

PR #2133, opened 2015-08-20
WT-2049 don't start format's RNG at the same place every time.

keithbostic:


michaelcahill:

PR #2134, opened 2015-08-20
WT-147. For dynamic index creation, solve a locking issue with LSM

michaelcahill:
From @ddanderson (replacing #2127 with a pull request merging into the original branch):

This branch was built upon index-create, which does dynamic index creation.

For dynamic index creation, solve a locking issue with LSM. The issue is that the application thread is holding the schema lock while we are filling the index. If the index is LSM, then a different LSM thread wants to split the file eventually, but it needs the schema lock to do that.

Coordinate the locking by tagging the LSM cursor with a special flag: "I'm holding the schema lock while filling this lsm: cursor". That in turn sets a flag on the LSM file, so when the LSM worker thread comes along, it pretends it's already holding the schema lock - it's being held on its behalf by the app thread.

### DIFF
--- a/src/include/lsm.h
+++ b/src/include/lsm.h
@@ -59,14 +59,15 @@ struct __wt_cursor_lsm {
 
 #define	WT_CLSM_ACTIVE		0x001   /* Incremented the session count */
 #define	WT_CLSM_BULK		0x002   /* Open for snapshot isolation */
-#define	WT_CLSM_ITERATE_NEXT    0x004   /* Forward iteration */
-#define	WT_CLSM_ITERATE_PREV    0x008   /* Backward iteration */
-#define	WT_CLSM_MERGE           0x010   /* Merge cursor, don't update */
-#define	WT_CLSM_MINOR_MERGE	0x020   /* Minor merge, include tombstones */
-#define	WT_CLSM_MULTIPLE        0x040   /* Multiple cursors have values for the
+#define	WT_CLSM_SCHEMA_LOCK_FILL 0x004	/* Schema locked while index filled */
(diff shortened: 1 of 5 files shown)


michaelcahill:
@ddanderson, this seems better, but I think there are still some problems with this approach:

1. Do we need any additional coordination between LSM workers now that they aren't actually holding the schema lock?  In other words, could two workers race to service the switch request?
2. If the index is big enough, LSM throttling will kick in once it loads enough chunks because merges will be blocked.  I think we either need to disable throttling (the easy solution) or enable merges by having them inherit the schema lock as well (the scary solution).
3. I am concerned that there could be a switch operation in progress after the index build creates that isn't holding the schema lock.  In other words, the index building thread releases the schema lock but a worker thread continues on afterward unprotected against real conflicts.  Are you confident that can't happen?
4. Any ideas about making the layering violation less egregious?

agorrod:
@ddanderson A potential alternative to this might be to have index-create into LSM trees use an approach similar to bulk loading into an LSM tree. A bulk load inserts all content into a single chunk - meaning there is no background maintenance work to do.

Index create differs from bulk load since the data is out-of-order, but I think we could re-use most of the bulk load logic in LSM. Do you think that would be a reasonable approach?

ddanderson:
@agorrod, thanks for that idea.  It does makes sense to me - it's really the same problem other than the out-of-order.  Do I understand that there are no splits during a bulk load?

I'll take a closer look to see what's involved.

michaelcahill:
@ddanderson, that's right -- this will just create a single btree that becomes the first chunk in the LSM tree after the index is built.

FWIW, @agorrod's #2150 looks like a safer solution than this branch to me.

PR #2137, opened 2015-08-20
WT-2053: Fix a bug in disk verify messages.

keithbostic:

@ddanderson, here's the fix for WT-2053.


michaelcahill:

PR #2138, opened 2015-08-21
WT-2051 - add a nice error message when failing to open metadata

daveh86:
Note: changes error handling to not panic immediately on seeing an invalid value.

### DIFF
--- a/src/meta/meta_table.c
+++ b/src/meta/meta_table.c
@@ -37,10 +37,23 @@ __metadata_turtle(const char *key)
 int
 __wt_metadata_open(WT_SESSION_IMPL *session)
 {
+	int tret;
+
 	if (session->meta_dhandle != NULL)
 		return (0);
 
-	WT_RET(__wt_session_get_btree(session, WT_METAFILE_URI, NULL, NULL, 0));
+	tret = __wt_session_get_btree(session, WT_METAFILE_URI, NULL, NULL, 0);
+	if (tret != 0) {
+		__wt_errx(session, "WiredTiger has failed to open its "
+		    "metadata file.");
+		__wt_errx(session, "This may be due to the database "
+		    "files being encrypted, being from an older "
+		    "version or due to corruption on disk.");
+		__wt_errx(session, "You should confirm that you have "
+		    "started with the correct options including all encryption "
+		    " and compression options.");
+		return (tret);
+	}
 
 	session->meta_dhandle = session->dhandle;
 	WT_ASSERT(session, session->meta_dhandle != NULL);
--- a/src/support/err.c
+++ b/src/support/err.c
@@ -480,17 +480,15 @@ __wt_panic(WT_SESSION_IMPL *session)
 	F_SET(S2C(session), WT_CONN_PANIC);
 	__wt_err(session, WT_PANIC, "the process must exit and restart");
 
-#if !defined(HAVE_DIAGNOSTIC)
+#if defined(HAVE_DIAGNOSTIC)
+	__wt_abort(session);                    /* Drop core if testing. */
+#endif
 	/*
 	 * Chaos reigns within.
 	 * Reflect, repent, and reboot.
 	 * Order shall return.
 	 */
 	return (WT_PANIC);
-#endif
-
-	__wt_abort(session);			/* Drop core if testing. */
-	/* NOTREACHED */
 }
 
 /*
@@ -504,12 +502,11 @@ __wt_illegal_value(WT_SESSION_IMPL *session, const char *name)
 	    name == NULL ? "" : name, name == NULL ? "" : ": ",
 	    "encountered an illegal file format or internal value");
 
-#if !defined(HAVE_DIAGNOSTIC)
-	return (__wt_panic(session));
+#if defined(HAVE_DIAGNOSTIC)
+	__wt_abort(session);                    /* Drop core if testing. */
 #endif
 
-	__wt_abort(session);			/* Drop core if testing. */
-	/* NOTREACHED */
+	return (WT_ERROR);
 }
 
 /*


daveh86:
@michaelcahill, would you mind having a look at this?

src/meta/meta_table.c
```diff
@@ -37,10 +37,23 @@ __metadata_turtle(const char *key)
 int
 __wt_metadata_open(WT_SESSION_IMPL *session)
 {
+	int tret;
```
michaelcahill: Prefer `WT_DECL_RET` here, and a `ret` variable.  We only use `tret` in complicated situations where there may already be a non-zero `ret`.

src/support/err.c
```diff
@@ -504,12 +502,11 @@ __wt_illegal_value(WT_SESSION_IMPL *session, const char *name)
...
 
-#if !defined(HAVE_DIAGNOSTIC)
-	return (__wt_panic(session));
+#if defined(HAVE_DIAGNOSTIC)
+	__wt_abort(session);                    /* Drop core if testing. */
 #endif
 
-	__wt_abort(session);			/* Drop core if testing. */
-	/* NOTREACHED */
+	return (WT_ERROR);
```
michaelcahill: @keithbostic, you should also take a look at this -- it changes `__wt_illegal_value` to not panic, which is necessary in MongoDB to get a decent error message out.  (One problem with panic is that as soon as MongoDB sees a `WT_PANIC` passed to its error handler, it shuts down).

I was actually a little surprised that this was a panic: are we sure we can never get here from verify or salvage, for example?

michaelcahill:
@daveh86, let's wait for @keithbostic to have a chance to look at this change: he may feel strongly about the existing behavior.

src/support/err.c
keithbostic: `__wt_illegal_value` is only called when we've hit a truly fatal error (for example, the default case in a switch statement that "can't happen", or we failed checksum on a block we read), it should not be hit by salvage or verify (under normal conditions).

In short, panic is the right behavior for WiredTiger internally, we can't proceed further.

@daveh86 could change the name from `__wt_illegal_value` to `__wt_panic_something`, to clarify that?

Not sure what to suggest about MongoDB: could it push out the error message before it shuts down?

michaelcahill: @keithbostic, `__wt_illegal_value` still returns an error, the only code that will run afterwards is error handling.  In this case, a block read should fail, but that shouldn't bring down the system: it just means there is a file we can't open.  That seems very different to me to a programming error where we control all of the possible values that could go into a `switch`.

My concern about the panic here (and more generally, the number of calls to panic through the code) is that I can't reason about whether an operation might cause a panic or not.  In this case, that means it's impossible to handle this failure in a more graceful way.  It isn't possible for MongoDB to figure out what has caused this kind of failure short of parsing WiredTiger's error log...

keithbostic: @michaelcahill,

> __wt_illegal_value still returns an error, the only code that will run afterwards is error handling.

We have to call `__wt_panic` in order to shut the system down (set `WT_CONN_PANIC`), what prevents other code from running?

> In this case, a block read should fail, but that shouldn't bring down the system: it just means there is a file we can't open. That seems very different to me to a programming error where we control all of the possible values that could go into a switch.

Yes and no. If we can't open a file and read the first block, I agree we want to fail gracefully, it's likely a configuration/encryption problem. If we're running for a day and suddenly a block read fails, I think it's a pretty big problem and we want to shut down fast, before more damage occurs.

> My concern about the panic here (and more generally, the number of calls to panic through the code) is that I can't reason about whether an operation might cause a panic or not.

I know it's not helpful, but I think the rule is that any significant operation can panic: if a memory error ever corrupts `WT_PAGE.type`, we're almost certain to do a switch on the page type somewhere in every significant operation, fall into `WT_ILLEGAL_VALUE_ERR` in a switch statement, and panic. I can't think of any alternative, either, if there's a corrupted page in memory, we're done.

The way we handled this for encryption was to explicitly call verify on the root page when we opened each new file, and fail (relatively) gracefully. (It occurs to me we're probably still outputting a scary message into MongoDB's log, which may be a problem? I'll take a look in a minute.)

Is it the metadata file is special and we need to do something similar there?

michaelcahill: > Yes and no. If we can't open a file and read the first block, I agree we want to fail gracefully, it's likely a configuration/encryption problem. If we're running for a day and suddenly a block read fails, I think it's a pretty big problem and we want to shut down fast, before more damage occurs.

That's why I feel that panic should be handled at the top level rather than deep down in a macro like `WT_ILLEGAL_VALUE` -- if instead it returned a `WT_ILLEGAL_VALUE` error (or just `EINVAL`), we can decide above whether that warrants a panic or not.  There is no way for the low-level code to know the difference between the two situations short of more special session flags.

src/meta/meta_table.c
```diff
@@ -37,10 +37,23 @@ __metadata_turtle(const char *key)
...
+		__wt_errx(session, "WiredTiger has failed to open its "
+		    "metadata file.");
+		__wt_errx(session, "This may be due to the database "
+		    "files being encrypted, being from an older "
+		    "version or due to corruption on disk.");
+		__wt_errx(session, "You should confirm that you have "
+		    "started with the correct options including all encryption "
+		    " and compression options.");
+		return (tret);
+	}
```
michaelcahill: @daveh86, now that #2114 is open, I think we should close this branch down and copy the error message to https://github.com/wiredtiger/wiredtiger/pull/2144/files#diff-e3f17d390ac7c334303e82cbc2a5900fR416

Does that make sense to you?

daveh86: @michaelcahill, Sounds sensible to me.

michaelcahill:
Closing this PR -- merging into #2144.

PR #2139, opened 2015-08-21
WT-2050 - when memory allocation fails, say how much memory we were t…

daveh86:
…rying to allocate

### DIFF
--- a/src/os_posix/os_alloc.c
+++ b/src/os_posix/os_alloc.c
@@ -58,7 +58,8 @@ __wt_calloc(WT_SESSION_IMPL *session, size_t number, size_t size, void *retp)
 		WT_STAT_FAST_CONN_INCR(session, memory_allocation);
 
 	if ((p = calloc(number, size)) == NULL)
-		WT_RET_MSG(session, __wt_errno(), "memory allocation");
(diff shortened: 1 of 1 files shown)


src/os_posix/os_alloc.c
```diff
@@ -58,7 +58,8 @@ __wt_calloc(WT_SESSION_IMPL *session, size_t number, size_t size, void *retp)
 		WT_STAT_FAST_CONN_INCR(session, memory_allocation);
 
 	if ((p = calloc(number, size)) == NULL)
-		WT_RET_MSG(session, __wt_errno(), "memory allocation");
+		WT_RET_MSG(session, __wt_errno(), "memory allocation of %"
+		    PRIu64 " bytes failed", size * number);
```
michaelcahill: This should be `SIZET_FMT` instead of `PRIu64`.  Similarly for the others.

michaelcahill: Also, I know what you've got is syntactically correct, but please get the "%" on the same line as the format specifier.

michaelcahill:
Thanks, @daveh86, lgtm.  I'll merge...

PR #2141, opened 2015-08-21
WT-2057: Strip the verbose configuration as part of writing the base configuration file.

keithbostic:
WT-2057: The verbose configuration is being written to the base configuration file, which means all subsequent runs (including wt command runs), inherit the verbose flags from the initial run.

Strip the verbose configuration as part of writing the base configuration file.

### DIFF
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1724,7 +1724,8 @@ __conn_write_base_config(WT_SESSION_IMPL *session, const char *cfg[])
 	    "encryption=(secretkey=),"
 	    "exclusive=,"
 	    "log=(recover=),"
-	    "use_environment_priv=,", &base_config));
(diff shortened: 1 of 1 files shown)


michaelcahill:
Thanks, @keithbostic.  I think we've discussed this before, but that list of things to strip is long enough that I think we should consider listing just the things that should go into the base config file (in `api_data.py`).

keithbostic:
@michaelcahill, I vaguely recall there wasn't a simple way to turn a list from `api_data.py` into a list of things to include (without writing the config parsing loop), but I don't remember for sure.

I don't feel strongly; currently, the default is creating a new config always adds it to the base config and the alternative is to change it so a new config doesn't default into the base config, and either choice can cause problems.

Anyway, I'm happy to take a run at this if you like, shall I?

PR #2142, opened 2015-08-22
WT-2058: Fix mutex and log-slot alignment.

keithbostic:
@michaelcahill, I think this is ready for review and merge.

One thing: we currently align the WT_LSM_CHUNK structure, and the declaration syntax is correct so it's being correctly aligned. However, I couldn't figure out exactly where/how it's being allocated, so that made me wonder if it really needs to be aligned in any interesting way? If it does and there's an obvious place to do it, we might want to add a `WT_CACHE_LINE_ALIGNMENT_VERIFY` call to verify that we get the alignment we want.


michaelcahill:

I doubt `WT_LSM_CHUNK` needs special alignment: we only ever allocate one at a time, and keep an array of pointers to chunks in `WT_LSM_TREE`.  I vaguely remember that we used to allocate the array of chunks directly, so maybe alignment used to make sense.

PR #2143, opened 2015-08-23
Replace the WT_PAGE_SPLIT_LOCKED lock with WT_PAGE_RECONCILIATION.

keithbostic:
@michaelcahill, staring at the page locks this morning, it occurred to me we don't need `WT_PAGE_SPLIT_LOCKED` -- reconciliation has to acquire it on every internal page, which means `WT_PAGE_RECONCILIATION` is just as good.

As of 3a41ccb, `WT_PAGE_RECONCILIATION` is used to block the trimming of obsolete updates and to protect compaction when it's reading page modify information. Updates don't happen on internal pages, and compaction of internal pages doesn't seem like much of a blocker.

The gain isn't that (one less atomic operation when reconciling every internal page), but I think it's a bit simpler/cleaner.

Feel free to discard this branch if you disagree, of course.


keithbostic:
@michaelcahill, one other thing. The comment in the code said something like `There's a deadlock when performing an in-memory split during a checkpoint`, and obviously it has something to do with multi-block splits in reconciliation.

I've stared at it a bit and the commit when the locking went in, and I don't see the deadlock. If you remember what it was, it'd be great if there was a better explanation in the comment, or maybe @daveh86 remembers?

keithbostic:
autobuild: retest this please

michaelcahill:
@keithbostic, as I remember it, the deadlock was between a checkpoint traversing an internal node locking each child, and an in-memory split of a leaf page that has the child exclusive (blocking the checkpoint) and is spinning waiting to set "splitting" on the parent.

These changes have caused some test failures and I didn't get a chance today to debug: can you please take a look?

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill, sorry for the test failure. I asserted `WT_PAGE_SPLIT_INSERT` isn't set when pages are discarded, and that was wrong, that flag is persistent.

keithbostic:
@michaelcahill, I'm still missing it.

The checkpoint is walking the parent/internal page, reviewing the child page; it hits the child page held exclusively for the split/eviction, and then loops in `__rec_child_modify` waiting for `WT_REF.state == WT_REF_LOCKED` to be resolved.

The split/eviction acquires the child page's split-lock (but that's just to single-thread splits into the parent), does the split and releases the child page.

I would agree there's a problem with checkpoint reconciliation racing with a split, there's this code in `__rec_child_modify`:
```
case WT_REF_SPLIT:
        /*
         * The page was split out from under us.
         *
         * We should never be here during eviction, a child page
         * in this state within an evicted page's subtree would
         * have caused eviction to fail.
         *
         * We should never be here during checkpoint, dirty page
         * eviction is shutout during checkpoint, all splits in
         * process will have completed before we walk any pages
         * for checkpoint.
         */
        WT_ASSERT(session, ref->state != WT_REF_SPLIT);
```

but that's not a deadlock?

michaelcahill:
@keithbostic, a split tries to acquire the *parent's* split-lock.  As you noted, the checkpoint is blocked at that point -- that's where the deadlock could come from.  It is avoided by having in-memory splits only do a try-lock on the parent (in `bt_split.c`):

```c
 894                 /*
 895                  * If we're attempting an in-memory split and we can't lock the
 896                  * parent, give up.  This avoids an infinite loop where we are
 897                  * trying to split a page while its parent is being
 898                  * checkpointed.
 899                  */
 900                 if (LF_ISSET(WT_SPLIT_INMEM))
 901                         return (EBUSY);
```

keithbostic:
@michaelcahill, OK -- when I read the comment I thought there was some underlying deadlock we were avoiding, not the deadlock created by the locks themselves! I've reworked some comments, and pushed those changes, I think this change is ready to merge whenever you want.

src/btree/bt_split.c
```diff
@@ -876,24 +876,35 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref,
...
-		 * checkpointed.
+		 * A checkpoint reconciling this parent page can deadlock with
+		 * our in-memory split. We have an exclusive page lock on the
+		 * child before we acquire the page's reconciliation lock, and
+		 * reconciliation acquires the page's reconciliation lock before
+		 * it will encounter the child's exclusive lock (which causes
+		 * reconciliation to loop until the exclusive lock is resolved).
+		 * If we can't lock the parent, give up to avoid that deadlock.
 		 */
 		if (LF_ISSET(WT_SPLIT_INMEM))
```
michaelcahill: Unrelated, but while reviewing, I think we could tighten this down and only give up in-memory splits if `btree->checkpointing` is set.  @keithbostic do you agree?

keithbostic: @michaelcahill, yes, I agree; as we discussed, we could tighten it down even further, by flagging in the page that reconciliation is holding the reconciliation lock, it's not another in-memory split that's got that lock.

PR #2144, opened 2015-08-24
WT-2051 Encryption/metadata error cleanup.

keithbostic:
@ddanderson, @michaelcahill, @daveh86:

Here is a branch to show what I'm thinking about with respect to WT-2051 (#2138).

Additionally, Andreas wrote to Don:

> So I tried returning a unique error code from the decrypt callback but it doesn't help me. I still get the the following errors:

> WiredTiger (0) [1440422383:942164][87246:0x7fff7477a300], file:WiredTiger.wt,
connection: file:WiredTiger.wt: encountered an illegal file format or internal value

> WiredTiger (-31804) [1440422383:942200][87246:0x7fff7477a300],
file:WiredTiger.wt, connection: the process must exit and restart:
WT_PANIC: WiredTiger library panic

In summary, all I did here was special-case the read/verify of the btree root page when a file is first opened, so errors from that read/verify don't output scary messages into the MongoDB log, nor will they panic the system.

I believe this branch fixes Andreas' concern as well (although I didn't test it); the btree read function no longer overwrites the result of the decryption function if "corruption" is anticipated.

I'm not sure this is the correct solution and I'm happy to discard this branch in favor of something better, but I thought it might be easier to code it than to explain it in GitHub comments.

Don, I think you said you would write up a test of the behavior that Andreas is looking for?


ddanderson:
@keithbostic, this looks good to me.  To be clear, with this change, when a decrypt function returns non-zero when verifying the first block of a file (the WT_SESSION_QUIET_CORRUPT_FILE is on), WT does not emit any message, that's up to the decrypt function or the application.

I'm almost done with a test, will push it to a subordinate branch later this evening.

keithbostic:
@ddanderson, that's a good point.

If we choose to use this approach, we should add an error message in the btree open call saying "unable to read file XXX", and additionally there's @daveh86's messages when the file we can't open is the metadata file.

ddanderson:
@keithbostic, the test is in pull request #2147.  Not as clean as I would like, but it confirms your code is correct for the case Andreas describes.

michaelcahill:
Thanks @keithbostic, @ddanderson and @hows -- this lgtm.  I'll merge.

PR #2146, opened 2015-08-25
WT-2051, WT-1822 encryption tests

ddanderson:
Added encryption test covering the case in which decrypt returns an error.  The error code must be returned by wiredtiger_open.


ddanderson:
Made the pull request against master by mistake.  Ignore this branch and pull request.

PR #2147, opened 2015-08-25
WT-2051, WT-1822.  Add test for decrypt returning an error

ddanderson:


ddanderson:
@keithbostic, here's the test I promised.

PR #2148, opened 2015-08-25
SERVER-20008 Don't reset eviction walks when hitting a busy page.

michaelcahill:
Also, when clearing eviction walk points before the eviction server goes to sleep, clear *all* walks, not just ones that sessions are waiting to be have cleared.


agorrod:
Thanks @michaelcahill I like this change (as discussed in person). It makes sense to avoid resetting the eviction walk where possible. Should I go ahead and merge, or would you like confirmation that it resolves SERVER-20008 first?

michaelcahill:
@agorrod, I'd be happy to have it merged so it gets more standalone WT testing.

Also, I think this is a backport candidate, do you agree?

agorrod:
Merged. I agree - this should be back ported.

PR #2150, opened 2015-08-26
WT-147 Add undocumented bulk=unordered for LSM cursors.

agorrod:
The feature allows us to stop worrying about schema lock interactions when building LSM indexes on existing tables.

### DIFF
--- a/src/lsm/lsm_cursor_bulk.c
+++ b/src/lsm/lsm_cursor_bulk.c
@@ -92,6 +92,7 @@ __wt_clsm_open_bulk(WT_CURSOR_LSM *clsm, const char *cfg[])
 {
 	WT_CONFIG_ITEM cval;
 	WT_CURSOR *cursor, *bulk_cursor;
+	WT_DECL_RET;
 	WT_LSM_TREE *lsm_tree;
 	WT_SESSION_IMPL *session;
 	int ordered;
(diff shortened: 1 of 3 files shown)


agorrod:
@ddanderson This is the LSM unordered-bulk load change we talked about in #2134 (it should replace the PR). To create a cursor with this semantic pass in a configuration string of: ```bulk=unordered```.

For example, the change I made to test out the functionality was:

```
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -840,7 +840,7 @@ populate_thread(void *arg)
 
        /* Do bulk loads if populate is single-threaded. */
        cursor_config =
-           (cfg->populate_threads == 1 && !cfg->index) ? "bulk" : NULL;
+           (cfg->populate_threads == 1 && !cfg->index) ? "bulk=unordered" : NULL;
        /* Create the cursors. */
        cursors = calloc(cfg->table_count, sizeof(WT_CURSOR *));
        if (cursors == NULL) {
```

src/lsm/lsm_cursor_bulk.c
```diff
@@ -90,17 +90,29 @@ __clsm_insert_bulk(WT_CURSOR *cursor)
...
 
+	/*
+	 * Check for the undocumented unordered bulk flag, which is used when
+	 * doing index builds into LSM for existing trees.
+	 */
+	WT_RET(__wt_config_gets_def(session, cfg, "bulk", 0, &cval));
+	WT_ASSERT(session, cval.val != 0);
+	if (strncmp(cval.str, "unordered", strlen("unordered")) == 0)
+		ordered = 0;
+
```
michaelcahill: @agorrod, another solution would be to pass this through LSM and handle `bulk=unordered` in `__wt_curfile_open` (by turning on `WT_DHANDLE_EXCLUSIVE` but not `WT_BTREE_BULK`.  Did you consider that approach?

agorrod: @michaelcahill I did think about that, and it generalizes the solution. I decided to keep it within LSM for now - since I figured an un-ordered bulk load into a btree muddies the ```bulk``` semantic (though probably less than having a special option that is only supported by LSM).

michaelcahill: Fair enough, I'm still trying to connect the dots to see what the `__fill_index` code will look like to take advantage of this feature: my hope was that there wouldn't have to be LSM-specific code in there.

agorrod: Sure - if it makes the index-create code more consistent, I'm happy to push
the functionality into btree.

ddanderson:
@agorrod, I tried this and there still seems to be an initial switch when bulk=unordered is specified.  The LSM worker threads can't switch because they are still waiting on the schema lock. (I made a reference pull request here: #2158, hangs when running test_schema02) Is there a way to make the file switch happen in the same thread for this case?

agorrod:
Thanks @ddanderson I see the failure. I'll dig into it.

agorrod:
@ddanderson I've pushed a change that makes the switch happen in the thread opening the cursor. I've verified that this change means test_schema02 no longer hangs with the index create changes.

Having said that.. I'm concerned about the semantic of holding the schema lock for the duration of an index create. Has there been a discussion about possibly reducing how long the schema lock is held from?

agorrod:
@ddanderson Could you review this please? I'm happy for it to be merged if it provides enough functionality to allow forward progress on WT-147.

PR #2152, opened 2015-08-26
WT-2062 Try harder to make progress on in-memory splits

michaelcahill:
Only give up if a checkpoint is in progress.


michaelcahill:
@agorrod can you please review -- this is a simple change, but I think it might help a lot in some workloads...

agorrod:
Thanks @michaelcahill lgtm

PR #2153, opened 2015-08-26
Specify a key-format to simplify creating lookaside table records.

keithbostic:
@michaelcahill, I haven't tested this extensively yet, but this is the direction I'm heading (it seems to run, fwiw).

This removes the flag byte from the beginning of the lookaside table records, I don't see any way to support multiple formats in that table.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -112,8 +112,7 @@ __wt_las_create(WT_SESSION_IMPL *session)
 	WT_RET(__wt_session_drop(session, WT_LAS_URI, drop_cfg));
 
 	/* Re-create the file. */
-	WT_RET(__wt_session_create(session, WT_LAS_URI,
-	    "key_format=u,value_format=" WT_UNCHECKED_STRING(QIu)));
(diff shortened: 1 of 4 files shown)


src/cache/cache_las.c
```diff
@@ -352,24 +337,22 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
-				 * record before we do, and the cursor remains
-				 * positioned in that case.
-				 */
-				WT_ERR(cursor->remove(cursor));
-			}
-			break;
-		WT_ILLEGAL_VALUE_ERR(session);
-		}
+		if (cnt == 1)
+			WT_ERR(__wt_cursor_get_raw_key(cursor, key));
```
michaelcahill: Don't you still need to copy the key out of the cursor with a `__wt_buf_set`?

keithbostic: That was my guess at the last core dump... :)

michaelcahill:
Thanks, @keithbostic, lgtm.

PR #2154, opened 2015-08-27
WT-2065 Add a quota to shared cache configuration.

agorrod:
That provides the ability to limit how much cache a single participant in a shared cache can be allocated.

### DIFF
--- a/src/conn/conn_cache_pool.c
+++ b/src/conn/conn_cache_pool.c
@@ -52,7 +52,7 @@ __wt_cache_pool_config(WT_SESSION_IMPL *session, const char **cfg)
 	WT_DECL_RET;
 	char *pool_name;
 	int created, updating;
-	uint64_t chunk, reserve, size, used_cache;
+	uint64_t chunk, quota, reserve, size, used_cache;
 
 	conn = S2C(session);
 	created = updating = 0;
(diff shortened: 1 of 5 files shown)


agorrod:
@michaelcahill Would you mind reviewing change?

src/conn/conn_cache_pool.c
```diff
@@ -156,6 +161,9 @@ __wt_cache_pool_config(WT_SESSION_IMPL *session, const char **cfg)
 		    session, cfg, "shared_cache.chunk", &cval));
 		WT_ASSERT(session, cval.val != 0);
 		chunk = (uint64_t)cval.val;
+		WT_ERR(__wt_config_gets(
+		    session, cfg, "shared_cache.quota", &cval));
+		quota = (uint64_t)cval.val;
 	}
```
michaelcahill: Three is about as many options as I'd like to repeat here -- is it reasonable to create an inlined function that handles both cases and call it three times?

michaelcahill:
@agorrod, LGTM -- we can worry about the config code separately if you prefer.

agorrod:
> we can worry about the config code separately if you prefer

@michaelcahill Yes please. I coded up a macro wrapper and it's not an obvious improvement. I'd rather push it into a different pull request.

PR #2155, opened 2015-08-27
WT-2064 Don't spin indefinitely waiting for the handle list lock in eviction

michaelcahill:
A thread may be holding that lock (e.g., verify) and waiting for the eviction server to clear a walk, leading to deadlock.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -159,6 +159,7 @@ __evict_server(void *arg)
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
 	WT_SESSION_IMPL *session;
+	u_int spins;
 
 	session = arg;
 	conn = S2C(session);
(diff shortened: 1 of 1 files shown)


michaelcahill:
@agorrod, @sueloverso -- I haven't been able to reproduce the hang in order to verify this fix, but based on the stack traces, this is where eviction is getting stuck.  Can you both please take a look?

agorrod:
Thanks @michaelcahill this seems reasonable to me. That dhandle try-loop is repeated a couple of times in the code now, if we find more uses it might be worth creating another macro (or pair of macros).

I'm happy for this to be merged if @sueloverso is happy with the change.

michaelcahill:
@sueloverso, I've poked at a couple more hangs and I believe this change will fix them.  Can you merge if you are happy with this?

@keithbostic, you'll want to pick up this change in WT-1967: it is causes the pull request tester to timeout frequently.

sueloverso:
I am testing this on the source tree where I was able to reproduce the problem.  If it runs, I'll merge.

PR #2156, opened 2015-08-27
WT-2066 - Update the oldest transaction ID from eviction

michaelcahill:


agorrod:
@michaelcahill When the eviction server isn't working it wakes every 100 ms, is that often enough?

michaelcahill:
@agorrod, I agree that we may want to update the oldest ID more frequently than 10 times per second if there are very high update rates to individual records.  I'm inclined to try this approach and add back in an update from applications threads selectively if we find a workload where the eviction server is too slow.  We could also switch the eviction server to wake more frequently as an intermediate approach.

agorrod:
Thanks, I'm happy with this - I'll merge.

PR #2160, opened 2015-08-28
WT-2056 Reorder btree cursor close so stats are maintained correctly.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'm happy for this to be merged once the fallout of the las changes has settled.

PR #2161, opened 2015-08-28
WT-1967 fix an "uninitialized variable" warning.

michaelcahill:
../src/reconcile/rec_write.c: In function '__rec_split_write':
../src/reconcile/rec_write.c:3252: warning: 'upd' may be used uninitialized in this function

Make it really obvious to dumb compilers that we're not going to continue after an illegal value.


agorrod:
Thanks @michaelcahill lgtm - I'm happy to merge when testing passes.

keithbostic:
Nice fix! That one has annoyed me forever, and it never occurred to me to change the macro.

PR #2162, opened 2015-08-29
WT-1967: don't dirty the tree/page when instantiating lookaside table records

keithbostic:
@michaelcahill, this change fixes some sporadic errors we've been seeing, generally I see it as an eviction server panic or `write operation on read-only checkpoint handle` error.

In short, we can't allow instantiating lookaside table records on a page being read into the cache dirty the tree/page, because if it's a checkpoint handle, we're never going to be able to write it, and it eventually all goes south during close or eviction.

I'm putting this on a branch, it needs review.

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -244,6 +244,35 @@ __cursor_row_search(
 }
 
 /*
+ * __cursor_modify_init --
+ *	Initialize the page's modification structure, mark the cursor's page
+ * and tree dirty.
+ */
+static inline int
+__cursor_modify_init(WT_SESSION_IMPL *session, WT_PAGE *page)
+{
+	/* If we don't yet have a modify structure, allocate one. */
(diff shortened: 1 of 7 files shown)


michaelcahill:
@keithbostic, I'll review the code properly once I'm on the ground in Hawaii, but my first question is why do we need to apply changes from the lookaside table in checkpoint handles?  The only thing I'd expect to be in a checkpoint tree are the page images that were created by the checkpoint...  Am I missing something?

keithbostic:
You're right! I fixed the symptom instead of thinking through the cause, trees meet forest.

I've pushed the simpler fix: if we're reading a page in the service of a checkpoint, we can ignore the cases where we fix up that page in memory. That's in 9607bb9, and is just a check of `session->dhandle->checkpoint`: it seems odd to be checking that down in the btree read function, is there a better test?

keithbostic:
This branch is the wrong fix, the right fix is in 9607bb9, closing.

PR #2163, opened 2015-08-31
WT-2069 - Add truncate to wtperf_run.sh

daveh86:


agorrod:
@daveh86 Could you try switching one of the Jenkins wtperf branches across to this branch before it gets merged - I'd like to verify that the jobs aren't going to be affected before pulling in the change.

agorrod:

PR #2165, opened 2015-09-01
WT-2076 wtperf slowdown.

keithbostic:


keithbostic:
@michaelcahill, in 6b2f6fb, you added this code:
```
/* Don't split a page that is pending a multi-block split. */
if (F_ISSET(page->modify, WT_PM_REC_MULTIBLOCK))
        return (0);
```
Do you remember what was going on, the commit doesn't say much.

keithbostic:
There are some workloads that still aren't fixed by this change (`log-append.wtperf` fails for me, `parallel-pop-stress` fails on OS X for Sue); however, they exhibit the same symptoms: we are failing to do in-memory splits on pages where appends are happening. I'm becoming convinced that what happened is we made it more likely for other types of eviction to succeed and so now regular eviction is happening when it didn't before, and it's significantly slower for these workloads than in-memory splits.

michaelcahill:
@keithbostic, the basic problem with doing in-memory splits on a page that was split during its last reconciliation is that as part of the in-memory split, we end up creating two pages in memory.  It wasn't clear to me what to do with the original `page->modify->mod_multi` at that point, so I avoided doing in-memory splits at all in that case.

More generally, in-memory splits were designed to be done in preference to ordinary eviction, because they move I/O out of application operations.  I'm happy to be convinced otherwise, but I think these performance results are telling us that trying an in-memory split before eviction was the right call.

michaelcahill:
I'm not in a good state to review this today -- @keithbostic, if there are no more changes to make on this branch, can you please have @agorrod review?

agorrod:
@keithbostic I've been digging into this today. I don't think it's the right solution for the performance degradation. It appears to me as though the earlier code was succeeding in doing an in-memory split via ```EVICT_UPDATE_RESTORE```, wheras this workload is failing to let ```EVICT_UPDATE_RESTORE``` evictions finish. I think I've figured out why - I'll push it into a separate branch.

agorrod:
I've pushed an alternative code change into https://github.com/wiredtiger/wiredtiger/pull/2167

PR #2167, opened 2015-09-02
WT-2076 Fix a bug that stops us writing out large pages with updates

agorrod:
Fix a bug that stops us writing out large pages with updates  that are new enough that they need to stay visible.

This is a workaround - the solution is hacky and incomplete. See comments in the patch for a description of the issue.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -433,11 +433,19 @@ __wt_reconcile(WT_SESSION_IMPL *session,
 	}
 
 	/* Get the final status for the reconciliation. */
+#if 0
+	The call to __rec_row_leaf is returning EBUSY for UPDATE_RESTORE calls
+	which means we are skipping the status call, and need the "special"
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic Could you take a look at this please? Tracing the ```WT_EVICT_UPDATE_RESTORE``` code path via the different variables it sets makes my head hurt - but this branch fixes the issue in my testing.

The short version is that the older code was doing a split-keep-updates, wheras the new code is bailing out of split-keep-updates evictions with an EBUSY return.

keithbostic:
Thank you Alex!!  Next time we meet, you're drinking for free. :)

The fix isn't too complicated, so I've pushed bf29f16 into develop so we can get some experience with it. Would you please do the review?

I don't think there's anything in #2167 you wanted to save, so I'll go ahead and close it.

@agorrod, @michaelcahill, @sueloverso

agorrod:
Thanks @keithbostic - I'll take a closer look at the fix today.

PR #2168, opened 2015-09-02
WT-2074: fix a race between lookaside table reconciliation and checkpoints.

keithbostic:
@michaelcahill, I hate to maintain a count of LAS evictions in progress, and wait for that to drain.

What do you think of checking the generation values before/after the reconciliation?

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -298,12 +298,9 @@ static int
 __evict_review(
     WT_SESSION_IMPL *session, WT_REF *ref, int *inmem_splitp, int closing)
 {
-	WT_BTREE *btree;
 	WT_DECL_RET;
 	WT_PAGE *page;
 	uint32_t flags;
-
-	btree = S2BT(session);
 
 	/*
 	 * Get exclusive access to the page if our caller doesn't have the tree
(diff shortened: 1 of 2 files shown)


keithbostic:
autobuild: retest this please

agorrod:
@keithbostic I'm happy to do a review of this one tomorrow.

michaelcahill:
@keithbostic, @agorrod, I guess I'm okay with this approach, but we are already counting LRU evictions in each file and waiting for them to drain in various cases.

My only concern with this change is that we might do a lot of work to split and evict a 10MB page when a checkpoint starts, and now have to throw it away, versus adding a little latency to the checkpoint itself.  If it doesn't show up in the parallel populate workload, I'm happy for this to be merged.

keithbostic:
@michaelcahill, we can wait for the count of LAS evictions in progress without holding anything other than the checkpoint lock, of course, but that seemed heavier-weight than checking for the race, plus it's a shared cache line. Anyway, just a suggestion -- if you'd rather go with a count/drain approach, let me know and I'll code it up.

michaelcahill:
@keithbostic, let's merge this and watch the automated perf tests.

PR #2169, opened 2015-09-03
WT-2076 cleanup

keithbostic:
@agorrod, here are most of the changes from the `wt-2076-append-slowdown` branch.

I've tried to do the commits in a step-wise fashion for clarity, would you please do the review?

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -311,7 +311,7 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_PAGE *page)
 	__wt_txn_update_oldest(session, 0);
 
 	/* If eviction cannot succeed, don't try. */
-	return (__wt_page_can_evict(session, page, 1, NULL));
+	return (__wt_page_can_evict(session, page, 1));
 }
 
 /*
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -311,7 +311,7 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_PAGE *page)
 	__wt_txn_update_oldest(session, 0);
 
 	/* If eviction cannot succeed, don't try. */
-	return (__wt_page_can_evict(session, page, 1));
+	return (__wt_page_can_evict(session, page, 1, NULL));
 }
 
 /*
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -442,7 +442,7 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
 
 	btree = S2BT(session);
 
-	for (force_attempts = oldgen = 0, wait_cnt = 0;;) {
+	for (force_attempts = oldgen = 0, sleep_cnt = wait_cnt = 0;;) {
 		switch (ref->state) {
 		case WT_REF_DISK:
 		case WT_REF_DELETED:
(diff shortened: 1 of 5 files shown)


src/btree/bt_read.c
```diff
@@ -311,7 +311,7 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_PAGE *page)
 	__wt_txn_update_oldest(session, 0);
 
 	/* If eviction cannot succeed, don't try. */
-	return (__wt_page_can_evict(session, page, 1, NULL));
+	return (__wt_page_can_evict(session, page, 1));
```
michaelcahill: I'm missing something about this change -- the common case of forced eviction is appends, and in that case we want to do an in-memory split before trying ordinary eviction.  Won't this change prevent some of those pages being split, because we're no longer checking if they can be?

keithbostic: @michaelcahill, @agorrod, OK, I think the light just dawned.

So, the point of this test in the middle of `__wt_page_can_evict` is we explicitly skip the other tests, because in-memory splits don't care about whether or not the page is dirty because we're not really evicting?

I can understand skipping the ` if (btree->checkpointing && (__wt_page_is_modified(page) || F_ISSET(mod, WT_PM_REC_MULTIBLOCK))) ` test, we want to split even if we're checkpointing the tree.

OK, how about moving the "page can split" test FIRST in `__wt_page_can_evict`, and adding a big comment, something like "see if we're going to split, otherwise test for the possibility of actual eviction".

michaelcahill: Hey @keithbostic, yes, that was the idea of having `__wt_page_can_split` in `__wt_page_can_evict`.  Making that clearer works for me.

agorrod:
@keithbostic I was going to shuffle this around today, but it feels like you've got a plan so I've left it alone.

I liked the idea of reducing the number of items required on a page before we consider it for in-memory splitting. The current minimum is ~4k, I thought you'd reduced that in (https://github.com/wiredtiger/wiredtiger/commit/0352823088b8e4ae83253be845e04990a566f4e3), but it looks like you actually increased it to ~8k. How would you feel about reducing it to 1k - it's only splitting into other in-memory pages, right?

An example:
* 5MB memory page max
* 2K key/value size per record
* 2560 items per 5MB (ignoring overhead)

The reason I'm pushing for this is that for append workloads we should try hard to do in-memory splits rather than pushing a reconciliation into an application thread (and locking down the page while we do it, thus stalling all insert threads until the eviction is done).

keithbostic:
> I liked the idea of reducing the number of items required on a page before we consider it for in-memory splitting. The current minimum is ~4k, I thought you'd reduced that in (0352823), but it looks like you actually increased it to ~8k.

I think I left it unchanged at 8K:
```
L0 = P(1)
L1 = P(1/4)
L2 = P(1/16)
L3 = P(1/64)
L4 = P(1/256)
L5 = P(1/1024)
L6 = P(1/4096)
```
So 2 L6 slots is 8K, the same as 8 L5 slots.

What I was seeing was we were getting skiplists with 32K entries, so the expected L6 entries was around 8 but the actual L6 entries were 0. I'll try and re-test that and see if I can reproduce it.

keithbostic:
> How would you feel about reducing it to 1k - it's only splitting into other in-memory pages, right?
> An example:
```
5MB memory page max
2K key/value size per record
2560 items per 5MB (ignoring overhead)
```
> The reason I'm pushing for this is that for append workloads we should try hard to do in-memory splits rather than pushing a reconciliation into an application thread (and locking down the page while we do it, thus stalling all insert threads until the eviction is done).

I generally agree with you, but average 2KB key/value sizes per record seems high. If it's a 100B key/value pair, then we'd split in-memory at 100KB which seems too quick.

That's probably irrelevant though, we aren't going to split at all unless we've first hit the memory page footprint limit, which should prevent us from splitting too aggressively. (It might also be interesting to split before the memory page footprint limit hits, if there are enough entries on the skiplist.)

Maybe commit a change to look for 4 L5 slots, so we split if there are at least 4K items on the list, and we can see how that behaves?

keithbostic:
@agorrod, I've reverted the mistakes, and added in the decrease in items required for an in-memory split.

Would you please do the review & merge if you're happy with it?

agorrod:

PR #2170, opened 2015-09-03
WT-2081 Make verify progress reporting less verbose.

agorrod:


agorrod:
This was triggered by SERVER-20179 - a test was generating too many progress messages. This change reduces the number of progress messages from ~1500 to 5 in my runs.

agorrod:
@keithbostic Can you review this change please?

keithbostic:
@agorrod, lgtm.

I removed the progress message wrapup (I don't see the point if we're not going to always do it) and I went ahead and made the same changes to salvage, since it has a similar structure.

PR #2172, opened 2015-09-04
WT-2085 Run some of the log_server threads operations more frequently

sueloverso:
than others.  In particular, flush the log slot 20x per second.

@keithbostic Would you please review?  I added a new API for waiting on conditions.  This change allows the log slot to be forced out more frequently than other operations in the same worker thread rather than having to create a new thread.

### DIFF
--- a/src/os_win/os_mtx_cond.c
+++ b/src/os_win/os_mtx_cond.c
@@ -42,7 +42,7 @@ __wt_cond_alloc(WT_SESSION_IMPL *session,
  *	before the time out period expires, let the caller know.
  */
 int
-__wt_cond_wait_isngal(WT_SESSION_IMPL *session,
+__wt_cond_wait_signal(WT_SESSION_IMPL *session,
     WT_CONDVAR *cond, uint64_t usecs, int *signalled)
 {
 	DWORD milliseconds;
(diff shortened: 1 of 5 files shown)


src/os_win/os_mtx_cond.c
```diff
@@ -37,11 +37,13 @@ __wt_cond_alloc(WT_SESSION_IMPL *session,
...
 /*
- * __wt_cond_wait --
- *	Wait on a mutex, optionally timing out.
+ * __wt_cond_wait_signal --
+ *	Wait on a mutex, optionally timing out.  If we get it
+ *	before the time out period expires, let the caller know.
  */
 int
-__wt_cond_wait(WT_SESSION_IMPL *session, WT_CONDVAR *cond, uint64_t usecs)
+__wt_cond_wait_isngal(WT_SESSION_IMPL *session,
```
markbenvenuto: Fix the function spelling: `__wt_cond_wait_signal` instead of `__wt_cond_wait_isngal`.

sueloverso: Thanks Mark!  Fix is pushed.  I really can type quite well usually.

sueloverso:
@michaelcahill Although this is merged, it only narrows the window of how long log records sit in memory.  Some tests may write data, read it back, crash and not find that data on restart.  We should talk about that.

PR #2173, opened 2015-09-04
Avoid losing errors from the underlying Windows functions.

keithbostic:
@markbenvenuto, can you review & merge if this is correct?

### DIFF
--- a/src/os_win/os_mtx_cond.c
+++ b/src/os_win/os_mtx_cond.c
@@ -45,7 +45,7 @@ int
 __wt_cond_wait_signal(
     WT_SESSION_IMPL *session, WT_CONDVAR *cond, uint64_t usecs, int *signalled)
 {
-	DWORD milliseconds;
+	DWORD err, milliseconds;
 	WT_DECL_RET;
 	uint64_t milliseconds64;
 	int locked;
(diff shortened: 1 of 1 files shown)


src/os_win/os_mtx_cond.c
```diff
@@ -94,18 +94,14 @@ __wt_cond_wait_signal(
...
-	}
+	if (ret == 0 && GetLastError() == ERROR_TIMEOUT)
+		*signalled = 0;
 
 	(void)__wt_atomic_subi32(&cond->waiters, 1);
 
 	if (locked)
 		LeaveCriticalSection(&cond->mtx);
-	if (ret != 0)
+	if (ret == 0)
```
markbenvenuto: `SleepConditionVariableCS` returns non-zero on success. See https://msdn.microsoft.com/en-us/library/windows/desktop/ms686301(v=vs.85).aspx

keithbostic:
@markbenvenuto, I tried again, please let me know what you think.

markbenvenuto:
@keithbostic The control flow is not a little ugly on the error case, but I wanted to be sure we set the state correctly on error. Let me know what you think.

keithbostic:
Thanks, @markbenvenuto, looks great!

keithbostic:
Reference WT-2090.

PR #2174, opened 2015-09-05
WT-147. Dynamic index creation, using LSM bulk=unordered

ddanderson:
This branch uses LSM bulk=unordered. Because I'm opening an LSM file indirectly fix via an index cursor, we need to do a little footwork to keep the bulk flag on the index file, and not the column group files.


ddanderson:
autobuild: retest this please

ddanderson:

agorrod:
@ddanderson Sorry to leave you hanging on this one. I did take a look.

I see that you made a similar change to the one I made in #2150, but your implementation is somewhat different. You are setting the ```WT_BTREE_BULK``` flag on the underlying btree when creating one of these unordered, bulk loaded files. That makes me nervous - WiredTiger has special treatment of bulk handles during checkpoints, and I'm not sure it's the semantic we want for these index builds.

Did you encounter problems when using the version in #2150? If not.. Maybe we could merge that into develop and then build this branch on top of those changes? Sorry - I know this has been a puzzle of a change, and I'm the cause..

ddanderson:
@agorrod, no need for apologies - your help thus far has been very helpful!

This change does not set `WT_BTREE_BULK`, but it introduces a new flag `WT_BTREE_BULK_UNORDERED` which is set.  I needed a way to test for this specific case, without setting `WT_BTREE_BULK` (we agree that setting that is wrong).

I think I may be able to eliminate some of the uses for `WT_BTREE_BULK_UNORDERED` (I can see that the reference to it in `txn_ckpt.c` is a no-op - I'll see if other uses can be eliminated - or at least clarify why not.

Also you probably noticed that I'm opening an index cursor with `bulk=index` -- I do that because I want the *index* file to be opened with `bulk=ordered`, but if I have a blanket `bulk=ordered` then I also get the *column groups* opened with `bulk=ordered` - that causes its own locking headaches.  So I created `bulk=index` to signal that just the index file should be opened specially. To back up a bit, why don't I open the underlying `lsm:` directly? Because a cursor on the higher level `index:` allows code to be shared more readily.  I have a thought about cleaning this up - opening a cursor using a projection that includes just the index columns shouldn't ever use the column groups.

Let me try those changes and get back to you - hopefully will be simplified.  Thanks!

If this doesn't make sense, I'll be around next week, we can schedule a call.

ddanderson:
@agorrod.  Hmm, hold off on reviewing this for the moment.  This has some of what I want, but something went wrong with my git push.

ddanderson:
See #2189 - another attempt to address locking issues.

agorrod:
@ddanderson Can we close this pull request, or does it have some content that #2189 doesn't?

ddanderson:
Closing this branch, #2189 is the current version.

PR #2176, opened 2015-09-07
WT-2092 Free log condition variables after all threads are joined.

agorrod:
Prevents any possible races when shutting down log utility threads.


agorrod:
@michaelcahill Could you review please?

michaelcahill:
Thanks, @agorrod, lgtm.

@sueloverso, FYI, we think this should fix the recent test segfaults.

PR #2177, opened 2015-09-08
SERVER-20193 - Performance regression fix

daveh86:


agorrod:
@daveh86 Could you add more information about the regression to the pull request?

src/include/serial.i
```diff
@@ -298,18 +298,27 @@ __wt_update_serial(WT_SESSION_IMPL *session, WT_PAGE *page,
...
-		F_CAS_ATOMIC(page, WT_PAGE_RECONCILIATION, ret);
-		/* If we can't lock it, don't scan, that's okay. */
-		if (ret != 0)
-			return (0);
-		obsolete = __wt_update_obsolete_check(session, page, upd->next);
-		F_CLR_ATOMIC(page, WT_PAGE_RECONCILIATION);
-		if (obsolete != NULL) {
-			page->modify->obsolete_check_txn = WT_TXN_NONE;
-			__wt_update_obsolete_free(session, page, obsolete);
+	if (upd->next == NULL)
```
agorrod: The comment no longer matches the control flow in the code. It should be updated.

src/include/serial.i
```diff
@@ -298,18 +298,27 @@ __wt_update_serial(WT_SESSION_IMPL *session, WT_PAGE *page,
...
+		return (0);
+
+	if (page->modify->obsolete_check_txn != WT_TXN_NONE) {
+		if (!__wt_txn_visible_all(session, page->modify->obsolete_check_txn)) {
```
agorrod: The code guideline is to test explicitly against 0 for function returns, rather than ```!```.

daveh86:
Sure, most of it is what is covered in SERVER-20193.

The following single threaded tests in Evergreen showed a slowdown https://evergreen.mongodb.com/task/performance_linux_wt_repl_singleThreaded_a7338c436160e259d0b20058f0a9ae4b442ca102_15_08_28_06_06_32

All of the Update.MmsIncDeep* tests saw a significant drop off. It was suggested by Michael that WT-2066, introduced this regression.

src/include/serial.i
```diff
@@ -298,18 +298,27 @@ __wt_update_serial(WT_SESSION_IMPL *session, WT_PAGE *page,
...
-			return (0);
-		obsolete = __wt_update_obsolete_check(session, page, upd->next);
-		F_CLR_ATOMIC(page, WT_PAGE_RECONCILIATION);
-		if (obsolete != NULL) {
-			page->modify->obsolete_check_txn = WT_TXN_NONE;
-			__wt_update_obsolete_free(session, page, obsolete);
+	if (upd->next == NULL)
+		return (0);
+
+	if (page->modify->obsolete_check_txn != WT_TXN_NONE) {
```
agorrod: Add a comment describing why this check is necessary? I'd have thought the ```__wt_txn_visible_all``` call would capture the ```WT_TXN_NONE``` case.

agorrod:
> Sure, most of it is what is covered in SERVER-20193.

It makes my life easier if the information is contained in the pull request (specifically in either the commit messages, or the pull request description) - that saves me dereferencing into JIRA when creating change logs.

src/include/serial.i
michaelcahill: We're in a bind here because C (specifically C90) can't distinguish between integers and booleans.  `__wt_txn_visible_all` would return a boolean if it was part of our standard, so using `!` here is reasonable, I think.

src/include/serial.i
michaelcahill: This is my fault -- I handed @daveh86 a diff with no comments...

In this case, I'd like to call `__wt_txn_update_oldest`, but only if we've already noticed that there are many updates on the page.  We're using `obsolete_check_txn != WT_TXN_NONE` to indicate that.

src/include/serial.i
```diff
@@ -298,18 +298,27 @@ __wt_update_serial(WT_SESSION_IMPL *session, WT_PAGE *page,
...
+		if (!__wt_txn_visible_all(session, page->modify->obsolete_check_txn)) {
+			/* Check if the oldest ID can move forward and re-check. */
+			__wt_txn_update_oldest(session,0);
 		}
+		if (!__wt_txn_visible_all(session, page->modify->obsolete_check_txn))
+			return (0);
```
michaelcahill: We should set `page->modify->obsolete_check_txn = WT_TXN_NONE` here, so we don't keep doing the slow path in future.

michaelcahill: That would replace line 320.

agorrod:

PR #2178, opened 2015-09-08
WT-2093 Use the C99 bool type to clarify when functions return true/false

michaelcahill:


michaelcahill:
@keithbostic, you are most likely to have a strong reaction to this change -- what do you think?

keithbostic:
autobuild: retest this please

keithbostic:
I have no objection to this change, it makes sense to me.

I think we should probably agree to including any C99 features, rather than discussing them one-by-one?

agorrod:
> I think we should probably agree to including any C99 features

I've opened https://jira.mongodb.org/browse/WT-2095 to cover a discussion about that.

michaelcahill:
autobuild: retest this please

michaelcahill:

src/cache/cache_las.c
```diff
@@ -169,7 +169,7 @@ __wt_las_set_written(WT_SESSION_IMPL *session)
  * __wt_las_is_written --
  *	Return if the lookaside table has been written.
  */
-int
+bool
 __wt_las_is_written(WT_SESSION_IMPL *session)
 {
 	return (S2C(session)->las_written);
```
agorrod: Should ```las_written``` switch to a bool?

agorrod:
@michaelcahill I had one nit, otherwise lgtm.

michaelcahill:
Seems like we're all in agreement, let's merge to develop and see where we end up.

Note that the two bugs identified in the ticket are not fixed by this commit: there is more work to do before it can be resolved.

PR #2180, opened 2015-09-08
WT-2094 Eliminate direct write and record unbuffered log records that

sueloverso:
are arbitrarily large.

@michaelcahill Please review these changes.  I think this is a much better implementation of handling log records that are larger than the memory buffer.  This code does have a similar "force" setting enabled at this time, but it is much less intrusive than the old code.

I'm going to start up test/format on AWS against this code, but it passes the normal tests.


sueloverso:
@michaelcahill Although it passes on my machine, Jenkins is seeing a failure in `ex_backup`.  I will look into that, as that program does depend on logging.

src/include/log.h
```diff
@@ -63,21 +63,22 @@
  */
 
 /*
+ * The high bit is reserved for the special states.  If the high bit is
+ * set (WT_LOG_SLOT_RESERVED) then we are guaranteed to be in a special state.
+ */
+#define	WT_LOG_SLOT_FREE	-1	/* Not in use */
```
michaelcahill: What's going on here with the "-"?

michaelcahill: That seems to be a github oddity: it isn't in the code.  I'll delete...

michaelcahill:

sueloverso:
@michaelcahill I believe this is ready for review and merging.  It has run lots of iterations of random logging workloads as well as many different wtperf configurations.  

I ran the {{log-append.wtperf}} workload with varying value sizes.  In the case of using 130000 value sizes, where every record will exceed our buffer, the new tree is slower by about 40% compared against an old-develop (changeset 12044d22 - before any logging changes at all).  In the case of moderate sized records (24000) the two runs are within 1-2% of each other.  The current {{log-append.wtperf}} using default value sizes (100 bytes) runs about 40% faster in this branch.  These were run on my Mac.

michaelcahill:
@sueloverso, thank you, this lgtm.  I'll merge and follow up on jira.

PR #2182, opened 2015-09-10
WT-2075 When closing a slot, detect, return and handle different values.

sueloverso:
This avoids hangs when rapidly filling log files so that in progress
writes can actually finish.

@michaelcahill Please review this branch to solve the hang with many threads and very few records per log file.  There are a couple things of note:
1.  Performance generally is a bit better than currently because by detecting these different conditions, there actually is less slot churn.
2.  Please consider the places where I used `*_BUSY_OK` and if it would be otherwise appropriate to loop/wait for a forced write to complete.  In particular, log cursors could still see a written log record still buffered if it happens when another thread is in the middle of its write.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -591,8 +591,7 @@ __log_wrlsn_server(void *arg)
 	 * On close we need to do this one more time because there could
 	 * be straggling log writes that need to be written.
 	 */
-	while (__wt_log_force_write(session) == EBUSY)
-		__wt_yield();
+	WT_ERR(__wt_log_force_write(session, 1));
 	WT_ERR(__wt_log_wrlsn(session));
 	if (0) {
 err:		__wt_err(session, ret, "log wrlsn server error");
@@ -644,7 +643,7 @@ __log_server(void *arg)
 		 * and a buffer may need to wait for the write_lsn to advance
 		 * in the case of a synchronous buffer.  We end up with a hang.
 		 */
-		WT_ERR_BUSY_OK(__wt_log_force_write(session));
+		WT_ERR_BUSY_OK(__wt_log_force_write(session, 0));
 
 		/*
 		 * We don't want to archive or pre-allocate files as often as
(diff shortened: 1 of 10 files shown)


sueloverso:
I have added a statistic and restarted the wtperf run, looking for the stat to catch the case I found earlier today after 3 hours of successful runs.  I've also restarted the Jenkins job.

src/conn/conn_log.c
```diff
@@ -591,7 +591,8 @@ __log_wrlsn_server(void *arg)
 	 * On close we need to do this one more time because there could
 	 * be straggling log writes that need to be written.
 	 */
-	WT_ERR(__wt_log_force_write(session, 0));
+	while (__wt_log_force_write(session) == EBUSY)
+		__wt_yield();
```
michaelcahill: What about other errors from `__wt_log_force_write`?

src/log/log.c
```diff
@@ -45,7 +45,8 @@ __wt_log_ckpt_lsn(WT_SESSION_IMPL *session, WT_LSN *ckp_lsn)
 
 	conn = S2C(session);
 	log = conn->log;
-	WT_RET(__wt_log_force_write(session, 1));
+	while (__wt_log_force_write(session) == EBUSY)
+		__wt_yield();
```
michaelcahill: likewise we should `WT_RET_BUSY_OK` here.

src/log/log.c
```diff
@@ -1636,20 +1641,22 @@ __log_force_write_internal(WT_SESSION_IMPL *session, int new_slot)
...
+	 * else is processing the slot change.  We're done.  If we get
+	 * EBUSY, then return that so the caller can decide what to do.
 	 */
-	__wt_log_slot_close(session, slot, &release);
+	ret = __wt_log_slot_close(session, slot, &release, 1);
+	if (ret == WT_NOTFOUND)
+		return (0);
+	if (ret == EBUSY)
+		return (ret);
+	WT_ASSERT(session, ret == 0);
```
michaelcahill: Should this just be `WT_RET`?

michaelcahill:

PR #2183, opened 2015-09-10
Fix incorrect assert in Windows error handling

markbenvenuto:
This assert was wrong when I shifted the WIndows error range. It should just be < 0.

Also fixed an issue where a buffer was not correctly null-terminated by switching from `__wt_buf_set` to  `__wt_buf_fmt` during this testing.

Tests:
Windows build
Adhoc fault injection into __wt_read to ensure it is printed correctly.


agorrod:
Thanks @markbenvenuto lgtm - I'll merge.

PR #2184, opened 2015-09-11
WT-2100 Rename evict to evict_queue so it's easier to search for.

agorrod:


michaelcahill:
Sure, @agorrod, lgtm.  I'll merge...

PR #2185, opened 2015-09-11
WT-2093 Fix bugs treating error returns as booleans

keithbostic:
@michaelcahill, here are the fixes for the two bugs you identified in WT-2093.

Ready for review & merge.


src/btree/bt_vrfy_dsk.c
```diff
@@ -293,7 +293,7 @@ __verify_dsk_row(
 		case WT_CELL_ADDR_LEAF_NO:
 		case WT_CELL_KEY_OVFL:
 		case WT_CELL_VALUE_OVFL:
-			if (!bm->addr_valid(bm,
+			if (bm->addr_invalid(bm,
 			    session, unpack->data, unpack->size))
```
michaelcahill: Do we need to distinguish between `EINVAL` and other errors here (e.g., `ENOMEM`, `WT_PANIC`)?

src/btree/bt_vrfy_dsk.c
```diff
@@ -500,7 +500,7 @@ __verify_dsk_col_int(
 		    session, cell_num, tag, unpack->type, dsk->type));
 
 		/* Check if any referenced item is entirely in the file. */
-		if (!bm->addr_valid(bm, session, unpack->data, unpack->size))
+		if (bm->addr_invalid(bm, session, unpack->data, unpack->size))
```
michaelcahill: similarly, discarding the return value may lose information

src/btree/bt_vrfy_dsk.c
```diff
@@ -569,7 +569,7 @@ __verify_dsk_col_var(
 
 		/* Check if any referenced item is entirely in the file. */
 		if (cell_type == WT_CELL_VALUE_OVFL &&
-		    !bm->addr_valid(bm, session, unpack->data, unpack->size))
+		    bm->addr_invalid(bm, session, unpack->data, unpack->size))
```
michaelcahill: similarly

src/reconcile/rec_track.c
```diff
@@ -473,7 +473,7 @@ __wt_ovfl_reuse_search(WT_SESSION_IMPL *session, WT_PAGE *page,
 
 	if (WT_VERBOSE_ISSET(session, WT_VERB_OVERFLOW))
 		WT_RET(__ovfl_reuse_verbose(session, page, reuse, "reclaim"));
-	return (1);
+	return (0);
```
michaelcahill: Would it be clearer to return `WT_NOTFOUND` if the search fails, rather than having calling code check for `addr == NULL`?

src/btree/bt_vrfy_dsk.c
keithbostic: I don't think so.

The only possible error returns (in non-diagnostic mode) are EINVAL and ENOMEM: `__wt_block_addr_invalid` calls `__wt_block_buffer_to_addr`, calls `__block_buffer_to_addr`, calls `__wt_vunpack_uint`, which returns either EINVAL or ENOMEM. Neither of those two errors means anything more than a corrupted value, as far as I can tell.

It would be great to try and defend against future code changes breaking this path, but I don't see any obvious way to do that. Suggestions?

src/reconcile/rec_track.c
keithbostic: Then if any function called by `__ovfl_reuse_verbose` were ever to return `WT_NOTFOUND` for any reason, we'd have reintroduced the same bug this change is supposed to fix.

I could add `int *foundp` to the `__ovfl_reuse_verbose` signature if you'd prefer something more explicit?

src/btree/bt_vrfy_dsk.c
michaelcahill: It seems wrong for `__wt_vunpack_uint` to return `ENOMEM`.  That makes sense when packing an integer if the supplied buffer isn't big enough, but not for unpacking.  If we change `__wt_vunpack_uint` to return `EINVAL` in that case, are there any other cases you're aware of where this code has to handle any error return other than `EINVAL`?

My feeling in general is that we should check for specific returns (which if we change `__wt_vunpack_uint` should just be `EINVAL`) and otherwise use `WT_ERR` / `WT_RET` to handle errors without building in assumptions about what the called functions return.

src/reconcile/rec_track.c
michaelcahill: np, it was just a query -- adding a `*foundp` isn't any clearer than checking `addr != NULL`.

src/btree/bt_vrfy_dsk.c
keithbostic: With that change, the only error return from `WT_BM::addr_invalid` is `EINVAL`.

I mostly agree about checking for specific returns. Obviously, the problem is a future change to `__wt_vunpack_uint` can now break the verify code, and it's impossible to check every possible consumer every time an underlying function changes its set of error values. I agree in this case that's pretty unlikely -- the trick is drawing the line between places it's OK to do this, and places it's not.

keithbostic:
@michaelcahill, I've made the changes you suggested, this one is ready for review & merge if you're happy with it.

michaelcahill:

PR #2186, opened 2015-09-11
WT-2102 Combine the switch and force_write functions.

sueloverso:
All paths check whether they have closed the slot and need to retry
creating a new slot if they get EBUSY.

@michaelcahill Please review this.  This solves the hang and it is better to have a single function doing this work.


michaelcahill:
Thanks, @sueloverso, lgtm.  I'll merge...

PR #2187, opened 2015-09-12
WT-2104 Change log_printf to be synchronous.

sueloverso:
@keithbostic and @michaelcahill  Here is a simple set of changes to accomplish a synchronous flushing of the log via log_printf.  Feel free to toss if you don't like it.  I confirmed by turning on statistics and adding a couple more log_printf calls in ex_log.c and watch the log sync stat go up.

### DIFF
--- a/src/include/wiredtiger.in
+++ b/src/include/wiredtiger.in
@@ -1238,7 +1238,7 @@ struct __wt_session {
 	/*!
 	 * Insert a ::WT_LOGREC_MESSAGE type record in the database log files
 	 * (the database must be configured for logging when this method is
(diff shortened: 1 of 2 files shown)


keithbostic:
In my experience, when people are writing their own log records, they're potentially writing a lot of them, so I think this behavior would have to be configurable, we don't want to flush the log on every log-printf.

(I think this is another reason to have an "official" way to flush the log, it seems like an application might want to printf some number of log records and then have a way to know they're flushed.)

sueloverso:
We've agreed this is not the path we want.  Tossing this.

PR #2190, opened 2015-09-14
WT-2101 Don't update the logging ckpt_lsn on clean shutdown.

sueloverso:
It can race with archive.  Metadata LSNs may not be updated on clean shutdown.

@michaelcahill Please review.  This fixes the missing file errors with `test_txn11`.  Also a minor KNF I saw along the way.


michaelcahill:
Thanks, @sueloverso!  LGTM.

PR #2191, opened 2015-09-15
WT-2075 - Add test for log hang with large parallel workload

daveh86:


bench/wtperf/runners/large-doc-log-hang.wtperf
```diff
@@ -0,0 +1,10 @@
+# wtperf options file: Test a log file with a multi-threaded
+# append workload.
```
sueloverso: @daveh86 I have two comments:
1.  Please name the file describing its content, derivative and purpose.  I suggest log-append-large.wtperf.  
2.  Please add more to the comment above in the config file to state we want to force a  lot of log file switches and we want to fit fewer records per log file than we have populate threads, etc.
Otherwise thanks for adding this!

michaelcahill:
LGTM, merging.

PR #2195, opened 2015-09-16
WT-2109 Slow op tracker

agorrod:


agorrod:
@daveh86 Could you take a look through this change please? The idea behind this change is that it will log information about any API call that takes longer than a configured amount of time. The information consists of a set of "events" that can be nested.

The logging happens via the WiredTiger message event handler.

I'd like general review comments, but I've also noticed that the timers aren't working as I expect. Each operation consists of a total time and a self time field. I'm seeing that the self time is often longer than the total time.

For example, when I change the wtperf medium btree configuration as follows:

```
--- a/bench/wtperf/runners/medium-btree.wtperf
+++ b/bench/wtperf/runners/medium-btree.wtperf
@@ -1,6 +1,7 @@
 # wtperf options file: medium btree configuration
 conn_config="cache_size=1G"
 table_config="type=file"
+sess_config="op_trace_min=5"
 icount=50000000
 report_interval=5
 run_time=120
```

I see some outputs that look like (I suspect you'll need a diagnostic build, otherwise you might not get any messages):

```
Operation tracking message:
{
"slow_op" : [
{
"elapsed" : 6,
"self_time" : 62,
"type" : 9,
"done" : 1,
},
{
"elapsed" : 0,
"self_time" : 19,
"type" : 1004,
"done" : 1,
},
]
}
```

I'm working on making those messages more useful - at least giving the type a name rather than a number. The type numbers aren't relevant to figuring out the timer issue.

agorrod:
autobuild: retest this please

src/conn/conn_cache_pool.c
```diff
@@ -647,6 +647,7 @@ __cache_pool_adjust(WT_SESSION_IMPL *session,
 				    (entry->cache_size - smallest) / 2);
 			adjustment =
 			    WT_MIN(adjustment, entry->cache_size - reserved);
+
```
daveh86: Just a newline

src/cursor/cur_table.c
```diff
@@ -867,6 +867,7 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 	tablename = uri;
 	if (!WT_PREFIX_SKIP(tablename, "table:"))
 		return (EINVAL);
+
```
daveh86: Just a newline

src/session/session_op_tracker.c
```diff
@@ -0,0 +1,252 @@
...
+	 */
+	if (session->iface.connection == NULL)
+		return (0);
+
+	if (entry->api_boundary)
+		--session->api_call_depth;
+
+	WT_RET(__wt_epoch(session, &entry->end));
+	entry->done = 1;
+	/* Start the sub-entry timer if this is a nested operation. */
```
daveh86: To be clear, this is the sub-entry timer within the "parent" object. Not the timer object of a sub-entry

sueloverso:
@agorrod I didn't see this before commenting in the Jira ticket.  Can you tell me the difference between this new wtperf and underlying code and the `max_latency` support that is already in wtperf (other than exceeding the latency is currently always fatal)?  It seems very similar functionally but also a lot more complex.

daveh86:
retest this please

src/os_posix/os_fsync.c
```diff
@@ -15,6 +15,11 @@
 static int
 __wt_handle_sync(int fd)
 {
+#define	DISABLE_SYNC 1
```
daveh86: Looks like this is debug code of some description.

src/session/session_op_tracker.c
```diff
@@ -0,0 +1,252 @@
...
+		prev_entry->self_time_us +=
+		    WT_TIMEDIFF(entry->start, prev_entry->last_stop);
+		break;
+	}
+
+	/* Create and insert the first entry into the slow op tracker. */
+	entry->type = type;
+	*entryp = entry;
+
+	if (api_boundary)
```
daveh86: It seems like this value will only ever be 0 or 1 as it is only incremented when we are on API boundaries. What happens if we have some nesting thats more than 2 calls deep?

agorrod: The ```api_boundary``` is a flag passed into ```__wt_session_op_tracker_create_entry```. The ```session->api_call_depth``` variable is used to track how deep in the API call stack we are.

We need to track those since WiredTiger makes calls via it's own APIs, and we want to track when we are at the top level (user) API call.

src/session/session_op_tracker.c
```diff
@@ -0,0 +1,270 @@
...
+		memcpy(&prev_entry->last_start,
+		    &entry->end, sizeof(struct timespec));
+		break;
+	}
+
+	/* Reporting is done as we are returning from the API call. */
+	if (entry->api_boundary && session->api_call_depth == 0) {
+		/* Update the self timer since this is the end of the trace. */
+		entry->self_time_us +=
+		    WT_TIMEDIFF(entry->end, entry->last_start);
```
sueloverso: WT_TIMEDIFF returns the time difference in nanoseconds so divide by 1000.  I've generally been dividing by WT_MILLION.  Do we need less than 1 msec granularity?

src/session/session_op_tracker.c
```diff
@@ -0,0 +1,270 @@
...
+{
+	WT_DECL_RET;
+	WT_ITEM *buffer;
+	WT_OP_TRACKER_ENTRY *entry;
+	uint64_t cumulative_skipped_time, optime_us, self_time_us;
+
+	cumulative_skipped_time = 0;
+
+	if ((entry = TAILQ_FIRST(&session->op_trackerq)) == NULL)
+		return (WT_NOTFOUND);
```
sueloverso: Do we really want to return WT_NOTFOUND here?  It seems like if there is nothing to do, we should just return 0, or if we really expect something to always be there, assert entry != NULL.  I don't think we want successful operations to suddenly start returning WT_NOTFOUND (like commit or rollback for example).

PR #2196, opened 2015-09-16
WT-2042 Only try to evict tombstones that are visible to all readers.

michaelcahill:


michaelcahill:
@keithbostic, as discussed, this change does help with the workload described in the ticket.  To get the rest of the performance, we need a change to the integration layer which I will push into code review soon.

keithbostic:
@michaelcahill, looks good to me, of course, might as well merge it?

michaelcahill:
Sure, if you're happy, please merge.  We're having trouble with the pull request tester job at the moment, I'm trying to get to the bottom of it, but I think this one is pretty safe.

PR #2197, opened 2015-09-16
SERVER-20159 Make all readers wait while the cache is full.

michaelcahill:


michaelcahill:
autobuild: test this please

agorrod:
This check was designed to allow old readers to continue making progress even if the cache was full (so they didn't need to be rolled back). The new look aside table should mean this isn't necessary any longer - since old readers no longer pin pages in cache.
Stop allowing old readers to pull in pages over the cache limit.

agorrod:
Thanks @michaelcahill lgtm - I'd like to get this into stress testing to ensure that it doesn't hurt configurations with small caches. Should I go ahead and merge?

michaelcahill:
Thanks, @agorrod, please go ahead and merge.

PR #2199, opened 2015-09-16
WT-2104 Write up commit and log_flush options.

sueloverso:
Also add large default value for transaction_sync.

@keithbostic and @michaelcahill Please review these changes.  I added a large, real default value for the `transaction_sync` method.  I also wrote up the options for `commit_transaction` and `log_flush`.  They're a bit redundant.  Let me know if there is enough information about the `background` setting.

### DIFF
--- a/src/docs/tune-durability.dox
+++ b/src/docs/tune-durability.dox
@@ -66,6 +66,40 @@ most durability guarantees).
 	logging configured; updates durable on application or system
 	failure}
 </table>
+
+The durability setting can also be controlled directly on a per-transaction
+basis via the WT_SESSION::commit_transaction method.
+The WT_SESSION::commit_transaction supports several durability modes with
+the \c sync configuration that override the connection level settings.
+
+If \c sync=on is configured then this commit operation will wait for its
+log records, and all earlier ones, to be durable on disk before returning.
+
+If \c sync=off is configured then this commit operation will write its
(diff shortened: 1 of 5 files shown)


src/docs/tune-durability.dox
```diff
@@ -66,6 +66,40 @@ most durability guarantees).
...
 	failure}
 </table>
+
+The durability setting can also be controlled directly on a per-transaction
+basis via the WT_SESSION::commit_transaction method.
+The WT_SESSION::commit_transaction supports several durability modes with
+the \c sync configuration that override the connection level settings.
+
+If \c sync=on is configured then this commit operation will wait for its
+log records, and all earlier ones, to be durable on disk before returning.
```
michaelcahill: My only quibble is with "durable on disk" -- unless I'm missing something, that depends on the `transaction_sync.method` being set to `fsync`.

michaelcahill:
Thanks, @sueloverso, apart from a quibble about "durable on disk", this lgtm.

src/docs/tune-durability.dox
sueloverso: I fixed it for the commit part and added a sentence on `log_flush` to say it overrides that and forces an `fsync`.  Please review.

michaelcahill:
Thanks, @sueloverso, lgtm.

PR #2200, opened 2015-09-17
Re-organize and rename WT riak/basho tests.

daveh86:
As per the email discussion with Sue. Renamed the test1-4 suite and added a few extra comments


daveh86:
@sueloverso, can you review this test? With your OK I will add 500m-btree suite into Jenkins.

sueloverso:
Thanks @daveh86 I have merged this.

PR #2201, opened 2015-09-17
WT-2111: coverity complaint

keithbostic:
@sueloverso, would you please do the review and merge on this one?

One note, it looked to me like it was OK to simply set `WT_CURSOR.saved_err` (potentially clearing it), in this case.


sueloverso:
Thanks @keithbostic.  LGTM.  Once the tester passes, I'll merge.

PR #2202, opened 2015-09-17
WT-2112: Coverity 73704, 73705, memory leak in salvage

keithbostic:
@sueloverso, another Coverity fix for review & merge, please!

### DIFF
--- a/src/btree/bt_slvg.c
+++ b/src/btree/bt_slvg.c
@@ -153,7 +153,6 @@ static int  __slvg_trk_leaf_ovfl(
 		WT_SESSION_IMPL *, const WT_PAGE_HEADER *, WT_TRACK *);
 static int  __slvg_trk_ovfl(WT_SESSION_IMPL *,
 		const WT_PAGE_HEADER *, uint8_t *, size_t, WT_STUFF *);
-static int  __slvg_trk_split(WT_SESSION_IMPL *, WT_TRACK *, WT_TRACK **);
 
 /*
  * __wt_bt_salvage --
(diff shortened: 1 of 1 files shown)


src/btree/bt_slvg.c
```diff
@@ -512,26 +511,6 @@ err:	__wt_free(session, trk->trk_addr);
 }
 
 /*
- * __slvg_trk_split --
- *	Split a tracked chunk.
- */
-static int
-__slvg_trk_split(WT_SESSION_IMPL *session, WT_TRACK *orig, WT_TRACK **newp)
```
sueloverso: Hmm, I'm more a fan of renaming the function to make it clear we allocated, like `__slvg_trk_alloc` or `__slvg_trk_split_alloc` than duplicating the code in two places.  Otherwise these LGTM.

keithbostic: I thought about that, but `++a_trk->shared->ref;` is part of the work we're doing to split the WT_TRACK object, and I'd rather do that inline instead of in some other place.

src/btree/bt_slvg.c
```diff
@@ -1017,15 +997,21 @@ __slvg_col_range_overlap(
...
 	/*
 	 * Second, reallocate the array of pages if necessary, and then insert
 	 * the new element into the array after the existing element (that's
 	 * probably wrong, but we'll fix it up in a second).
 	 */
-	WT_RET(__wt_realloc_def(
-	    session, &ss->pages_allocated, ss->pages_next + 1, &ss->pages));
+	if ((ret = __wt_realloc_def(session,
+	    &ss->pages_allocated, ss->pages_next + 1, &ss->pages)) != 0) {
+		__wt_free(session, new);
```
sueloverso: Based on what you pointed out in your last comment, don't we need to decrement `a_trk->shared->ref` on the error paths too?

keithbostic: Thanks!

Fixed in 744f8d6.

PR #2203, opened 2015-09-18
WT-2114 Make application eviction fairer

agorrod:
Have it use a global evict count rather than a count of evictions done by this thread.


michaelcahill:
@agorrod, is gthis ready for review and merge?

agorrod:
I ran the same workload I used in #2213 comparing this branch to the develop branch. On develop I see 357 operations that take longer than 1 second, on this branch I don't see any.

Total throughput across the 5 connections was slightly improved on this branch as well - though within expected run-to-run variation.

michaelcahill:
@agorrod, FTR I'd like to merge `txn_busy` with `busy` in this code, but that's orthogonal to your change.  I'll merge...

PR #2204, opened 2015-09-20
WT-2117 Set a minimum of 1MB cache per thread in test/format

agorrod:


agorrod:
@keithbostic What do you think of this change?

keithbostic:
@agorrod, I think I'm in favor of it -- otherwise, we end up with "bug fixes" that lead to other problems, and we waste time on workloads that just aren't sensible.

michaelcahill:
+1 from me, too -- shall I merge?

agorrod:
Yes please.

PR #2205, opened 2015-09-21
WT-2115: fast-delete pages can be incorrectly lost due to a crash.

keithbostic:
@michaelcahill, this is what I'm thinking for WT-2115.

It's untested (I just wrote it up on the plane ride home), so probably not ready to merge.

One interesting change: I'm making it more likely we'll write proxy cells, which allows us to evict internal pages we'd not otherwise evict. Seems reasonable, if we're trying to evict internal pages, things are already bad.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1357,7 +1357,7 @@ __rec_child_modify(WT_SESSION_IMPL *session,
 #define	WT_CHILD_IGNORE		1		/* Deleted child: ignore */
 #define	WT_CHILD_MODIFIED	2		/* Modified child */
 #define	WT_CHILD_ORIGINAL	3		/* Original address */
-#define	WT_CHILD_PROXY	`	4		/* Deleted: proxy cell */
+#define	WT_CHILD_PROXY		4		/* Deleted: proxy cell */
 	*statep = WT_CHILD_ORIGINAL;
 
 	/*
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1522,7 +1522,7 @@ __rec_child_deleted(
 	WT_PAGE_DELETED *page_del;
 	size_t addr_size;
 	const uint8_t *addr;
-	int visible, visible_all;
+	bool visible, visible_all;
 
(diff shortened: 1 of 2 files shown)


src/reconcile/rec_write.c
```diff
@@ -1580,18 +1584,6 @@ __rec_child_deleted(
...
-	 */
-	if (ref->addr != NULL) {
-		r->leave_dirty = 1;
-
-		/* This page cannot be evicted, quit now. */
-		if (F_ISSET(r, WT_EVICTING))
-			return (EBUSY);
-	}
-
-	/*
```
michaelcahill: Continuing to evict seems wrong to me: we're going to throw away version information, so future readers can't figure out whether to see the deleted data or not.

keithbostic: I'm continuing to evict if the change is visible to the eviction thread, in other words, there's an older reader. That's probably wrong: in `rec_txn_read`, we switched to checking `wt_txn_committed` instead of "visibility" in the case of eviction.

I also dropped `r->leave_dirty`: I don't think that has any meaning for internal pages.

keithbostic:
autobuild: retest this please

michaelcahill:
@keithbostic, this still seems wrong to me -- it only makes sense to use `__wt_txn_committed` in the context of the lookaside table (i.e., if we are going to evict a page and mark it as needing lookaside when it is read back in, which will allow reads to figure out correct versions).

I don't want the lookaside table anywhere near truncates: that would copy around pages that we're going to free as soon as we can.

keithbostic:
@michaelcahill, let's talk this over?

You know this stuff better than I do, but it seems to me the internal page's proxy cell acts pretty much like the lookaside table. but without needing it. When we read an internal page with a proxy cell, we create the state as `WT_REF_DELETED`: https://github.com/wiredtiger/wiredtiger/blob/develop/src/btree/bt_page.c#L446. We don't need the lookaside table because all of the rows are in a single known state.

keithbostic:
This branch isn't correct, it evicts the internal page in some cases where we can't safely evict.

Discard in favor of #2206.

PR #2206, opened 2015-09-21
WT-2115 Don't skip truncated pages that are part of a checkpoint

michaelcahill:


michaelcahill:
@keithbostic, on first look through, here's what I was thinking: looks like we overlapped pushing changes for this one...

michaelcahill:
P.S. Apologies for screwing up the commit message: I need to move some checks from my regular development machine to my laptop.

PR #2207, opened 2015-09-21
WT-2039 Add error check and unit test for log records over 4Gb.

sueloverso:
@michaelcahill Please review this change and unit test for handling attempts to log > 4Gb.  In particular I'm open to modifying which error is returned (per comment, which will need to be removed once decided).  If the error changes, the `msg` in the unit test needs to be updated too.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1798,6 +1798,15 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
 
 	conn = S2C(session);
 	log = conn->log;
+	if (record->size > UINT32_MAX) {
+		__wt_errx(session, "Record size %" PRIu64
+		    "too large to write into log.  Maximum %" PRIu32,
+		    (uint64_t)record->size, UINT32_MAX);
+		/*
+		 * Other possible errors: E2BIG, EPERM, ENOMEM, EINVAL
+		 */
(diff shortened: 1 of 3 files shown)


src/log/log.c
```diff
@@ -1798,6 +1798,12 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
 
 	conn = S2C(session);
 	log = conn->log;
+	if (record->size > UINT32_MAX) {
+		__wt_errx(session, "Record size %" PRIu64
+		    "too large to write into log.  Maximum %" PRIu32,
+		    (uint64_t)record->size, UINT32_MAX);
```
michaelcahill: Minor nit: we have `WT_SIZET_FMT` to avoid this cast to `uint64_t`.

src/txn/txn.c
```diff
@@ -519,7 +519,6 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		 */
 		__wt_txn_release_snapshot(session);
 		ret = __wt_txn_log_commit(session, cfg);
-		WT_ASSERT(session, ret == 0);
```
michaelcahill: Ouch, nice catch.

michaelcahill:
@sueloverso, thanks, lgtm if you can fix the `printf` nit.

PR #2208, opened 2015-09-21
WT-2122 Convert more expressions to use bool

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm. @keithbostic should do a full review as well - there is some subtlety that could benefit from another set of eyes.

keithbostic:
@michaelcahill, @agorrod, I've reviewed these changes, pushed a couple of minor cleanups, and I'm fine with merging it. (To be clear, I like this change, I wish we'd done it a few years ago!)

That said...

I wanted to get a better feel for the change, so I took a chunk of code (reconciliation), and pushed bool down into it (there were ten or so "ints" in the underlying reconciliation structure that were all really booleans). It was a pretty straight-forward change, but as you can see in #2209, it's a thousand lines of diff, and that's just the set of changes for a single chunk of code.

I'm concerned that using diffs for backporting is going to become impossible, if we continue to convert parts of the system to bool.  Remember, I am the person least experienced with backports here, so you two should just feel free to ignore me, but maybe backporting this isn't worth the effort?

PR #2209, opened 2015-09-22
WT-2122 more boolean conversion

keithbostic:
@michaelcahill, there are two changes on this branch (on separate commits, it should be easy to take one and not the other).

The first change converts `WT_PAGE_MODIFY.flags` from a "flags" field to an "enum" so we don't need all the masking we were doing. That field being a "flags" field was just a holdover from long-ago when there were flags there, or we thought there might be flags there, I don't even remember which it was.

The second change drives bool further down into reconciliation and the block manager, discussion in #2208.


src/include/btmem.h
```diff
@@ -387,14 +387,11 @@ struct __wt_page_modify {
...
-#define	WT_PM_REC_REWRITE	0x08	/* Reconciliation: rewrite in place */
-#define	WT_PM_REC_MASK							\
-	(WT_PM_REC_EMPTY | WT_PM_REC_MULTIBLOCK |			\
-	 WT_PM_REC_REPLACE | WT_PM_REC_REWRITE)
-	uint8_t flags;			/* Page flags */
+#define	WT_PM_REC_EMPTY		1	/* Reconciliation: no replacement */
+#define	WT_PM_REC_MULTIBLOCK	2	/* Reconciliation: multiple blocks */
+#define	WT_PM_REC_REPLACE	3	/* Reconciliation: single block */
+#define	WT_PM_REC_REWRITE	4	/* Reconciliation: rewrite in place */
+	uint8_t recon_result;		/* Reconciliation state */
```
michaelcahill: @keithbostic, we use `rec_*` everywhere else, can we call this `rec_result` to match?

Also, should we make this an `enum` for real to match the commit comment?

keithbostic: Sure, renamed and pushed.

I left the field as a uint8_t instead of an enum because an enum is 4B, and I hate to grow WT_PAGE_MODIFY.

michaelcahill:

BTW, you can make 1 byte `enum`s using `__attribute__ ((__packed__))` if that's the only reason to keep lists of `#define` constants around.  The advantage is that we get compile-time checking of `switch` and easier debugging...

keithbostic:
@michaelcahill,

> BTW, you can make 1 byte enums using __attribute__ ((__packed__)) if that's the only reason to keep lists of #define constants around. The advantage is that we get compile-time checking of switch and easier debugging...

I agree, but as far as I can tell, there's no way to do 1 byte enums in MSVC.

PR #2210, opened 2015-09-23
WT-2126: page split locking

keithbostic:
@michaelcahill, thinking about this: it reduces the chance of our doing any work before colliding with a checkpoint, at the cost of holding the parent page's reconciliation lock for a longer time.

This doesn't fix WT-2126, we can still be forced to unroll the split preparation, it's just far less likely.


agorrod:
@keithbostic This change makes me nervous - I'm chasing an issue now where operations are taking over a second to acquire the ```WT_PAGE_RECONCILIATION``` lock to update their parent after a reconciliation.

michaelcahill:
Thanks, @keithbostic, I understand the intent here after talking it through today.

FTR, we can potentially leak memory if there is a failure after we start allocating pages during a split, but the bigger problem is that updates are *moved* (not copied) from the original page onto split pages, if we are restoring updates.

As discussed, we should panic if any updates were removed from the original page: we could put them back using the same process we're using to apply the updates to the new pages, but that code will be fiddly to get right.

michaelcahill:
I'm going to merge, we can follow up separately to make failures panic, but I think they can now only happen due to `ENOMEM`.

agorrod:
This was back ported to MongoDB 3.0.7

PR #2211, opened 2015-09-23
WT-2127 Deepen the tree more regularly to avoid wide internal pages.

agorrod:
I was seeing internal pages grow to contain over 70,000 entries. Which makes reconciliation of child pages very expensive. Be more aggressive about splitting internal pages.


michaelcahill:
I like this change in principle, can we talk about what tests to run to make sure it doesn't introduce performance regressions?

dist/api_data.py
```diff
@@ -271,7 +271,9 @@ def __cmp__(self, other):
         leaf pages''',
         min=0),
     Config('split_deepen_min_child', '0', r'''
-        minimum entries in a page to consider deepening the tree''',
+        minimum entries in a page to consider deepening the tree. Once
+        there are more than this many entries pages will be immediately
+        considered for splitting''',
```
keithbostic: @agorrod, I'd rephrase this to talk about deepening the tree, not `considered for splitting`.

agorrod:
> can we talk about what tests to run to make sure it doesn't introduce performance regressions?

Yes. I'll start by running the set of wtperf configurations. I'll also try a couple of longer YCSB runs. Do you have any other suggestions?

michaelcahill:
FTR, we have run various workloads today including the current MongoDB system perf tests.  We haven't seen any regressions from these changes.  Given that, I'll merge and we'll see what happens in our full perf tests and in MongoDB's.

PR #2213, opened 2015-09-24
WT-2131 Switch to using a lock to control page splits.

agorrod:
We used to spin on setting an atomic flag, but we saw cases where that wasn't fair (and burned CPU).

This implementation adds a lock into the page structure - we could probably find a better place for that if the change is a general improvement.


keithbostic:
@agorrod, have you seen this make a difference in practice?

We were yielding on each loop spin, it might be interesting to try backing off instead of yielding immediately? Burns more CPU, of course, but that made a difference in the page acquisition loop, IIRC, yields were waiting too long.

I don't think there's any place we use WT_PAGE.modify for internal pages, we might be able to create a union there to avoid growing the WT_PAGE structure (or, instead, move the WT_PAGE.modify pointer into the leaf-page unions).

Not excited, to be honest: we used to have 3 locks around this stuff, I was pretty happy when we got it down to 1.

agorrod:
> @agorrod, have you seen this make a difference in practice?

Yes - I've seen places where a thread has waited for over 3 seconds for a chance to update the parent. Details are in https://jira.mongodb.org/browse/WT-2131

keithbostic:
Does decreasing the page size at which we split help?

agorrod:
> Does decreasing the page size at which we split help?

Decreasing the page size means that I no longer see threads waiting for over a second, and should decrease the amount of splitting into a single parent in a pure-random workload with a large data set.

It doesn't change that using a spinlock approach can result in unfairness if there is a lot of contention.

I should have been clearer - I'm not tied to using a read/write lock here. It is just a mechanism to ensure threads get to update the parent in the same order they started waiting. We could create a new lock type or data structure to achieve the same result.

keithbostic:
My assumption has been that this doesn't happen often enough that fairness is an issue, I'm still adjusting to the new reality. :)

agorrod:
> My assumption has been that this doesn't happen often enough that fairness is an issue, I'm still adjusting to the new reality

I spent pretty much a week looking for this problem because it didn't even occur to me either. Out of cache random update workloads do split into their parents a lot - especially if configured with small pages.

agorrod:
autobuild: retest this please

src/include/mutex.i
```diff
@@ -251,3 +251,54 @@ __wt_spin_unlock(WT_SESSION_IMPL *session, WT_SPINLOCK *t)
...
+
+        WT_UNUSED(session);
+
+	/*
+	 * Possibly wrap: if we have more than 64K lockers waiting, the ticket
+	 * value will wrap and two lockers will simultaneously be granted the
+	 * lock.
+	 */
+	ticket = __wt_atomic_fetch_add16(&lock->waiter, 1);
+	for (pause_cnt = 0; ticket != lock->owner;) {
```
agorrod: @michaelcahill I'm a bit concerned that the compiler will choose not to update lock->owner on it's way through this loop. Is that a valid concern?

michaelcahill: @agorrod, what we're relying on in this loop is that the `WT_PAUSE` and/or `__wt_sleep` inside the loop prevent the compiler from caching `lock->owner` in a register.  We could further call that out with `volatile`, but I don't think that is strictly necessary.

src/btree/bt_split.c
```diff
@@ -1183,6 +1174,7 @@ err:	if (!complete)
 			if (next_ref->state == WT_REF_SPLIT)
 				next_ref->state = WT_REF_DELETED;
 		}
+	WT_TRET(__wt_fair_unlock(session, &parent->pg_intl_split_lock));
```
agorrod: Self review - this shouldn't be here. I've removed it from #2213 - I'll wait for a decision as to whether that branch is a preferred solution before fixing it here.

keithbostic:
@agorrod, I'm missing the reason for the new lock structure, I thought the existing r/w locks are fair?

agorrod:
The existing rw locks are fair, but in this case we don't need a two mode
lock. That allows for a simpler implementation. An advantage of the fair
lock over the rw lock is that it only requires 4 bytes.

It might make sense to switch most of our spin lock usages to be fair locks.

keithbostic:
> The existing rw locks are fair, but in this case we don't need a two mode lock. That allows for a simpler implementation.

As far as I can tell, the only thing the new implementation saves is a single barrier instruction in the unlock code, is it really any simpler? (Although I think I now see your point about why the "single lock" branch is more complex, otherwise you don't need a trylock function.)

A somewhat orthogonal issue: we should rename the lock if we keep it, "fair" doesn't distinguish it from the existing read/write locks. Maybe it's the "wlock" instead of the "rw" lock?

> An advantage of the fair lock over the rw lock is that it only requires 4 bytes.

Yes, it keeps us from growing `WT_PAGE` from 72 to 80 bytes.

agorrod:
A concrete example of the benefit of this branch. Using the following wtperf configuration file:

```
conn_config="cache_size=2GB,eviction=(threads_min=2,threads_max=2),log=(enabled=false),direct_io=(data,checkpoint),buffer_alignment=4096,checkpoint_sync=true,checkpoint=(wait=60)"
database_count=5
table_config="allocation_size=4k,prefix_compression=false,split_pct=75,leaf_page_max=4k,internal_page_max=16k,leaf_item_max=1433,internal_item_max=3100,type=file"
icount=50000
populate_threads=1
random_range=50000000
report_interval=5
run_time=3600
threads=((count=1,inserts=1),(count=10,reads=1))
value_sz=100
max_latency=1000
sample_interval=5
sample_rate=1
```

i.e: 5 different databases each with 10 read threads and one writer thread, checkpointing every 60 seconds. With direct I/O configured to avoid file system buffering consequences.

I see the following maximum latencies (us) for insert operations:

| Develop | split-fairer |
|---|---|
| 6.5|4.7 |
| 7.9|4.8 |
| 13.6|4.9 |
| 17.1|4.9 |
| 19.0|5.0 |
| 20.2|5.2 |
| 21.0|5.4 |
| 21.5|5.6 |
| 25.5|5.9 |
| 27.7|6.0 |
| 34.7|6.2 |
| 35.3|7.0 |
| 37.6|12.0 |

I see the following maximum latencies (seconds) for read operations:

| Develop | split-fairer |
|---|---|
|6.5 | 5.0 |
|7.6 | 5.2 |
|8.0| 5.4|
|13.6|5.4|
|17.8|5.5|
|19.0|5.6|
|20.2|5.7|
|21.0|6.3|
|25.5|7.0|
|27|7.1|
|34.7|7.4|
|35.4|7.9|
|37.6|12.0|

agorrod:
@michaelcahill Could you do a review of this change please? Firstly I'd like to know that the high level approach (using a new lock type) is right for now.

src/btree/bt_discard.c
```diff
@@ -55,7 +55,11 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	 */
 	WT_ASSERT(session, !__wt_page_is_modified(page));
 	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_EVICT_LRU));
-	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_RECONCILIATION));
+	/*
+	 * This will leave the page lock held - but that doesn't matter - we
+	 * are about to free the memory anyway.
+	 */
+	WT_ASSERT(session, __wt_fair_trylock(session, &page->page_lock) == 0);
```
michaelcahill: This isn't okay -- we're going to grab the lock in diagnostic mode and not release it.  Could we add a read-only `__wt_fair_islocked` for diagnostics?

src/btree/bt_debug.c
```diff
@@ -647,11 +647,14 @@ __debug_page_metadata(WT_DBG *ds, WT_PAGE *page)
...
-	if (F_ISSET_ATOMIC(page, WT_PAGE_RECONCILIATION))
-		__dmsg(ds, ", reconciliation");
 	if (F_ISSET_ATOMIC(page, WT_PAGE_SPLIT_INSERT))
 		__dmsg(ds, ", split-insert");
 
+	if (__wt_fair_trylock(session, &page->page_lock) != 0)
+		__dmsg(ds, ", locked");
+	else
+		WT_RET(__wt_fair_unlock(session, &page->page_lock));
+
```
michaelcahill: I tend to agree with the previous comment that this code is dumping flags, and if we switch to a lock we shouldn't try to dump its state.

agorrod:
@michaelcahill I've pushed fixes for your specific review comments. Do you have other general concerns with this pull request?

michaelcahill:
Thanks, @agorrod, this lgtm and I will merge.

To be clear, I think this is a step, and we will only temporarily be adding a new lock type.  My preference is to rename `WT_FAIR_LOCK` to `WT_SPIN_LOCK` and replace the current variants with this, but let's have that discussion separately to fixing the performance issues with contended splits.

PR #2214, opened 2015-09-24
WT-2119: don't evict clean multiblock pages with overflow items during checkpoints.

keithbostic:
@michaelcahill, this change looks solid in my test runs, I think it's ready to merge.


michaelcahill:
Thanks, @keithbostic, lgtm, merging.

PR #2215, opened 2015-09-24
WT-2134 Flush all buffered log records in log_flush

sueloverso:
Force out pending log slot buffer to advance log->alloc_lsn for log_flush.  Fix example to use the correct config signature.

@michaelcahill Please review this change to fix the bug in log_flush.  I also plan to write a new python unit test to simulate an "abort" and restart, but this merge does not need to wait on that new test.


sueloverso:
I wrote a unit test that fails currently on develop (in the expected way) and passes every time on this branch.

src/log/log.c
```diff
@@ -1978,6 +1978,11 @@ __wt_log_flush(WT_SESSION_IMPL *session, uint32_t flags)
 	conn = S2C(session);
 	WT_ASSERT(session, FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED));
 	log = conn->log;
+	/*
+	 * We need to flush out the current slot first to get the real
+	 * end of log LSN in log->alloc_lsn.
+	 */
+	WT_RET(__wt_log_flush_lsn(session, &lsn, 0));
 	last_lsn = log->alloc_lsn;
 	lsn = log->write_lsn;
```
michaelcahill: Do we need to set `lsn` again here?  If we're not using the LSN returned from `__wt_log_flush_lsn`, can we pass `NULL`?

sueloverso: Good catch.  I removed the line to set lsn the second time.

michaelcahill:
Thanks, @sueloverso, lgtm.  I'll merge...

PR #2216, opened 2015-09-24
WT-2112 Use bool in more places

keithbostic:
@michaelcahill, here's a boolean conversion pass over the rest of the source tree, except for src/include.


michaelcahill:
Thanks, @keithbostic, lgtm.  The `lsm_tree->exclusive` change is problematic, but I'll merge and fix that in the mix.

PR #2217, opened 2015-09-24
WT-2132 Be more defensive in code used to debug address structures.

agorrod:
The code is used via diagnostics and salvage.


agorrod:
@keithbostic Could you review this for me please? I've only ever hit it via the bt_debug functions, so I don't think it's an issue for other callers, but better safe than segfaulted.

keithbostic:
@agorrod, yes, that seems like a good change.

PR #2219, opened 2015-09-25
WT-2119: don't evict clean multiblock pages with overflow items during checkpoints.

keithbostic:
Another approach: track if an internal page has any on-page overflow keys.


keithbostic:
@michaelcahill, I've got a couple hundred runs of this branch now, it seems solid.

michaelcahill:
@keithbostic, thanks, I've reviewed this and I'm going to merge.  I've also opened WT-2144 to have some more discussion about overflow items...

PR #2220, opened 2015-09-27
WT-2140: WT_SESSION.reset can release more memory

keithbostic:
@daveh86, @michaelcahill: it occurred to me we should be discarding reconciliation memory as part of `WT_SESSION.reset`, it can be real memory if the session reconciles a big page. (Imagine an application that hasn't configured enough eviction threads and is round-robining through sessions, each of which gets involved in eviction.)


michaelcahill:
@keithbostic, this is fine except that it's expensive for MongoDB to call `WT_SESSION::reset` in current workloads (10% overhead in the graph in https://jira.mongodb.org/browse/SERVER-19966).

Can we quantify what it means to evict a "big" page, or have application threads always free after eviction?

keithbostic:
@michaelcahill,

> this is fine except that it's expensive for MongoDB to call `WT_SESSION::reset` in current workloads (10% overhead in the graph in https://jira.mongodb.org/browse/SERVER-19966).

Based on that graph, I'd agree, calling `WT_SESSION.reset` at all seems like a bad idea for MongoDB.

That said, once an application makes the decision to reset the session, don't we want to be thorough about discarding any memory we're holding?

> Can we quantify what it means to evict a "big" page, or have application threads always free after eviction?

It's a little tricky to quantify the reconciliation cleanup, but we can, of course. We'd want to release any memory based on evicting a large page (the number of page-split structures and keys per split point can be large), and any memory based on writing a large key or value (the in-memory copies of the biggest key/value seen in the page can be large).

Just to be clear, the goal is to avoid calling free on small chunks of memory?

We already quantify the block manager cleanup work. The issue there is if a thread does a checkpoint, it can potentially allocate large numbers of `WT_EXT` structures for extent list manipulation. But, we already discard all but 250 of the allocated `WT_EXT` structures when the checkpoint finishes, so our exposure is already limited.

As for having application threads threads free their memory after eviction. We could easily do that, two possible concerns: (1) if an application thread is doing eviction work for some period of time, we're slowing down the eviction path, and (2) we flag internal sessions with `WT_SESSION_INTERNAL`, but I'm pretty sure we don't always set that flag on every internal session, so there may be a little trickiness there. We already do some cleanup based on this (if we have more than 10K boundary points, we discard all of that memory when finishing a reconciliation), we could be more aggressive about that in the case of an application session.

All that said, I suspect reconciliation can tie down more memory than the typical cursor memory and even the scratch buffers. If we have to make a choice, I'd go for reconciliation memory before cursor resets.

keithbostic:
@michaelcahill, I came back to this one, and have two questions around using the `WT_SESSION_INTERNAL` flag to decide what is/is-not an application thread. 

1. Checkpoint does a lot of reconciliation work, and we don't want to slow down checkpoint threads by making them re-allocate reconciliation memory for every block they reconcile. Lots of different solutions here: change the checkpoint function set the `WT_SESSION_INTERNAL` flag, or add a new checkpoint flag to the session that's checked during reconciliation cleanup,  or look at the checkpoint lock? And plus call WT_SESSION.reset explicitly at the end of the checkpoint so the checkpoint session doesn't tie down a lot of resources. Anything you'd prefer, or do you have a better idea?

1. WT-2167: Recovery creates a normal (not internal) session, which means it would be considered an application thread. My guess is recovery should create an internal session and this problem goes away, but if there's a reason recovery is using a normal session, it has the same problem as a checkpoint thread.

michaelcahill:
@keithbostic, use `WT_SESSION_IS_CHECKPOINT` for 1.  WT-2167 has been fixed now, so 2 should no longer be an issue.

keithbostic:
@michaelcahill, calling `WT_SESSION.reset` at the end of checkpoint turned out to be a little harder than anticipated, it required some shuffling of the checkpoint and compaction code. There weren't any changes to locking order though, so the change should have minimal impact.

This change is ready for final review/merge.

michaelcahill:
Thanks, @keithbostic, this lgtm.  I'll merge.

PR #2222, opened 2015-09-27
WT-2141: __meta_track_apply, __meta_track_unroll have unnecessarily complex error handling.

keithbostic:
@michaelcahill, just a minor cleanup, but ought to be reviewed.


michaelcahill:
Thanks, @keithbostic, I think the error handling logic was left over from before that code was refactored.  I'll merge...

PR #2223, opened 2015-09-28
WT-2146: fast q search.

keithbostic:
@michaelcahill, I was having difficulty explaining what I was talking about when we were talking about speeding up row-store searches with Q/q formats as keys, so I coded it up.

I see a win, but not much of one, it's between 2-4% on searches, presumably from avoiding the for loop setup.

I'm using clang version 3.3, -O3.

Inlining the return in the switch statement slows things down, as does combining the whole thing into a single switch statement (where the default switch case does the work for keys larger than 9 bytes), which makes me think the win here is a tiny/simple switch statement.

Anyway, I'm pushing the branch, maybe you and @fedorova could take a look and see if you see any improvements?

Otherwise, I'd guess there's not enough of a win to keep the change.

Here's my test code:
```
#include <assert.h>
#include <inttypes.h>
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <time.h>
#include <unistd.h>

#include <wiredtiger.h>

#define check(x) do {                                                   \
        int __r;                                                        \
        if ((__r = (x)) != 0) {                                         \
                fprintf(stderr,                                         \
                    "%s: %d (%s)\n",                                    \
                    #x, __r, wiredtiger_strerror(__r));                 \
                assert(__r == 0);                                       \
        }                                                               \
} while (0)

#define MILLION 1000000

int
main(int argc, char *argv[])
{
        WT_CONNECTION *conn;
        WT_CURSOR *cursor;
        WT_SESSION *session;
        clock_t ce, cs;
        uint64_t key;
        int i, ops;
        const char *obj = "table:xx";

        /* Remove everything. */
        (void)system("rm -rf WT_TEST && mkdir WT_TEST");

        /* Create a single table, populate it with a million key/value pairs. */
        check(wiredtiger_open("WT_TEST", NULL, "create", &conn));
        check(conn->open_session(conn, NULL, NULL, &session));
        check(session->create(session, obj,
            "allocation_size=512,leaf_page_max=512,"
            "key_format=q,value_format=S"));
        check(session->open_cursor(session, obj, NULL, NULL, &cursor));
        for (key = 0; key < MILLION; ++key) {
                cursor->set_key(cursor, key);
                cursor->set_value(cursor, "========== VALUE =======");
                check(cursor->insert(cursor));
        }

        /* Close and re-open the database, so we get a tree. */
        check(conn->close(conn, NULL));
        check(wiredtiger_open("WT_TEST", NULL, "create", &conn));

        /* Repeated searches of a key in the middle. */
        check(conn->open_session(conn, NULL, NULL, &session));
        check(session->create(session, obj, NULL));
        check(session->open_cursor(session, obj, NULL, NULL, &cursor));
        key = MILLION / 2;
        cursor->set_key(cursor, key);

        ops = 10 * MILLION;
        printf("start run... "); fflush(stdout);
        cs = clock();
        for (i = 0; i < ops; ++i)
                check(cursor->search(cursor));
        ce = clock();
        printf("%dM: %.2lf\n",
            ops / MILLION, (ce - cs) / (double)CLOCKS_PER_SEC);

        return (0);
}
```


keithbostic:
@michaelcahill, @fedorova: it occurred to me I saw similar results when working with the collator/non-collator versions of the lexicographic sort functions, that is, branching is the problem more than anything else. Giving row-store record-number search its own inline function gives me a solid 10% faster search for record-number keys, and it's a simpler change. I'm checking for just Q and q at the moment, but presumably this would work for all of the  integer key formats?

src/btree/row_srch.c
```diff
@@ -218,10 +218,10 @@ restart:	page = current->page;
...
 
 		/*
-		 * Binary search of the internal page.  There are two versions
-		 * (a default loop and an application-specified collation loop),
-		 * because moving the collation test and error handling inside
-		 * the loop costs about 5%.
+		 * Binary search of the internal page. Three versions (record
+		 * numbers, no application-specified collation order and
+		 * application-specified collation order), because moving those
+		 * tests and the error handling inside the loop costs about 5%.
```
michaelcahill: I didn't catch this comment change in my branch, sorry -- it should be updated to say something like "short keys with no application-specified collation order" instead of "record numbers".

src/btree/row_srch.c
```diff
@@ -356,13 +370,28 @@ restart:	page = current->page;
 	}
 
 	/*
-	 * Binary search of the leaf page.  There are two versions (a default
-	 * loop and an application-specified collation loop), because moving
-	 * the collation test and error handling inside the loop costs about 5%.
+	 * Binary search of the leaf page. Three versions (record numbers, no
+	 * application-specified collation order and application-specified
+	 * collation order), because moving those tests and the error handling
+	 * inside the loop costs about 5%.
```
michaelcahill: Similarly, this comment should be updated if you take the "short key comparison" change.

keithbostic: Done, and I'll merge this when it passes tests.

This was a good idea, I hadn't thought of just sweeping up all small keys, nice.

PR #2224, opened 2015-09-28
WT-2131 Switch all WT_PAGE_RECONCILIATION locks to fair locks.

agorrod:
The main purpose is to keep a single lock for all page operations.

### DIFF
--- a/src/btree/bt_debug.c
+++ b/src/btree/bt_debug.c
@@ -645,10 +645,13 @@ __debug_page_metadata(WT_DBG *ds, WT_PAGE *page)
 		__dmsg(ds, ", disk-mapped");
 	if (F_ISSET_ATOMIC(page, WT_PAGE_EVICT_LRU))
 		__dmsg(ds, ", evict-lru");
-	if (F_ISSET_ATOMIC(page, WT_PAGE_RECONCILIATION))
-		__dmsg(ds, ", reconciliation");
 	if (F_ISSET_ATOMIC(page, WT_PAGE_SPLIT_INSERT))
 		__dmsg(ds, ", split-insert");
+
+	if (__wt_fair_trylock(session, &page->page_lock) != 0)
+		__dmsg(ds, ", locked");
+	else
+		WT_RET(__wt_fair_unlock(session, &page->page_lock));
 
 	if (mod != NULL)
 		switch (mod->rec_result) {
--- a/src/btree/bt_discard.c
+++ b/src/btree/bt_discard.c
@@ -55,7 +55,11 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	 */
 	WT_ASSERT(session, !__wt_page_is_modified(page));
 	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_EVICT_LRU));
-	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_RECONCILIATION));
+	/*
+	 * This will leave the page lock held - but that doesn't matter - we
+	 * are about to free the memory anyway.
+	 */
+	WT_ASSERT(session, __wt_fair_trylock(session, &page->page_lock) == 0);
 
 #ifdef HAVE_DIAGNOSTIC
 	{
--- a/src/include/mutex.i
+++ b/src/include/mutex.i
@@ -253,6 +253,26 @@ __wt_spin_unlock(WT_SESSION_IMPL *session, WT_SPINLOCK *t)
 #endif
 
 /*
+ * __wt_fair_trylock --
+ *	Try to get a lock - give up if it is not immediately available.
+ */
+static inline int
+__wt_fair_trylock(WT_SESSION_IMPL *session, WT_FAIR_LOCK *lock)
+{
+	uint16_t ticket;
+
+	WT_UNUSED(session);
+
+	/* Do the cheap test first. */
+	if (lock->waiter != lock->owner)
+		return (EBUSY);
+
+	ticket = lock->owner;
+	return (__wt_atomic_cas16(
+	    &lock->waiter, ticket, ticket + 1) ? 0 : EBUSY);
+}
+
+/*
  * __wt_fair_lock --
  *	Get a lock.
  */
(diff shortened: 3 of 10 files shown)


agorrod:
@keithbostic You suggested:

> we used to have 3 locks around this stuff, I was pretty happy when we got it down to 1.

This branch does that. I personally don't love it - I think it adds complexity, but it also grows the page structure for all pages sizes.

I'm happy to pull this in if you prefer it to a mixture of ```WT_PAGE_RECONCILIATION``` and locks - you have more experience debugging issues in the code than I do.

src/include/mutex.i
```diff
@@ -253,6 +253,26 @@ __wt_spin_unlock(WT_SESSION_IMPL *session, WT_SPINLOCK *t)
...
+
+	WT_UNUSED(session);
+
+	/* Do the cheap test first. */
+	if (lock->waiter != lock->owner)
+		return (EBUSY);
+
+	ticket = lock->owner;
+	return (__wt_atomic_cas16(
+	    &lock->waiter, ticket, ticket + 1) ? 0 : EBUSY);
```
michaelcahill: I don't think this is sufficient -- I think we need a `cas32` where we also check that `lock->owner` hasn't raced.

src/btree/bt_discard.c
```diff
@@ -55,7 +55,11 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	 */
 	WT_ASSERT(session, !__wt_page_is_modified(page));
 	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_EVICT_LRU));
-	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_RECONCILIATION));
+	/*
+	 * This will leave the page lock held - but that doesn't matter - we
+	 * are about to free the memory anyway.
+	 */
+	WT_ASSERT(session, __wt_fair_trylock(session, &page->page_lock) == 0);
```
keithbostic: @agorrod, I'd just toss this discard check, it's really not telling us anything useful.

src/btree/bt_debug.c
```diff
@@ -645,11 +645,14 @@ __debug_page_metadata(WT_DBG *ds, WT_PAGE *page)
...
-	if (F_ISSET_ATOMIC(page, WT_PAGE_RECONCILIATION))
-		__dmsg(ds, ", reconciliation");
 	if (F_ISSET_ATOMIC(page, WT_PAGE_SPLIT_INSERT))
 		__dmsg(ds, ", split-insert");
 
+	if (__wt_fair_trylock(session, &page->page_lock) != 0)
+		__dmsg(ds, ", locked");
+	else
+		WT_RET(__wt_fair_unlock(session, &page->page_lock));
+
```
keithbostic: @agorrod, I would remove this code, since the lock is no longer a flag, it shouldn't be in the list of page flags.

keithbostic:
@agorrod

>> we used to have 3 locks around this stuff, I was pretty happy when we got it down to 1.

> This branch does that. I personally don't love it - I think it adds complexity, but it also grows the page structure for all pages sizes.

It feels awkward to me to do split/reconciliation locking differently for internal vs. leaf pages and I don't understand where it's more complex to do it one way instead of two? Can you please clarify where you see the additional complexity?

With respect to growing the page size, both approaches grow the page size equally:
```
                                sizeof(WT_PAGE) sizeof(WT_PAGE.u)
develop                         72              32
split-fairer                    80              40
split-fairer-single-lock        80              32
split-fairer-single-lock (fix)  72              32
```

The reason is the `WT_PAGE.u` union is currently 32B, and since the `WT_PAGE_.u.intl` member is 32B, adding `WT_FAIR_LOCK` to the union grows the whole page.

However, there's some wasted space in WT_PAGE at the moment (outside of the union), so moving the `WT_FAIR_LOCK` structure up a few lines keeps the page size from growing:
```
diff --git a/src/include/btmem.h b/src/include/btmem.h
index 04f8b77..c626f88 100644
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -583,6 +583,12 @@ struct __wt_page {
        uint8_t flags_atomic;           /* Atomic flags, use F_*_ATOMIC */
 
        /*
+        * Used to protect and co-ordinate splits for internal pages and
+        * reconciliation for all pages.
+        */
+       WT_FAIR_LOCK page_lock;
+
+       /*
         * The page's read generation acts as an LRU value for each page in the
         * tree; it is used by the eviction server thread to select pages to be
         * discarded from the in-memory tree.
@@ -603,12 +609,6 @@ struct __wt_page {
 #define        WT_READGEN_STEP         100
        uint64_t read_gen;
 
-       /*
-        * Used to protect and co-ordinate splits for internal pages and
-        * reconciliation for all pages.
-        */
-       WT_FAIR_LOCK page_lock;
-
        size_t memory_footprint;        /* Memory attached to the page */
 
        /* Page's on-disk representation: NULL for pages created in memory. */
```

agorrod:
On 28 Sep 2015 11:15 pm, "Keith Bostic" <notifications@github.com> wrote:
> It feels awkward to me to do split/reconciliation locking differently for
internal vs. leaf pages and I don't understand where it's more complex to
do it one way instead of two? Can you please clarify where you see the
additional complexity?

The complexity I see is from using the same lock for two different
purposes. It took me a fair while to untangle splits from reconciliation in
my head when reasoning about the WT_PAGE_RECONCILIATION flag, and I got the
separation wrong a couple of times.

> With respect to growing the page size, both approaches grow the page size
equally:

I should have been clearer - that was assuming the internal specific change
(split-fairer) shuffled the page content so the lock didn't grow the union.
I'm glad you are clever enough to get the same result in this branch.

agorrod:
I should also note that my primary concern was with growing the page size,
not the complexity. Since you pointed out that growing the page isn't
necessary I like this branch a lot more.

I'd like to choose a better name than page_lock, since we have one of those
in the page modify structure anyway. With a variable name change and the
bit of cleanup as you suggested to the debug/diagnostic code I'm liking
this branch more.

agorrod:
Combining the WT_PAGE_RECONCILIATION flag lock and the page lock are the way to go - I'll merge this into the other branch and we can continue refining this change there.

PR #2225, opened 2015-09-28
WT-2143 Revert part of the recent log cleanup changes.

agorrod:
They introduced a performance regression, due to log slot states
being setup at the wrong time.


agorrod:
@sueloverso I didn't dig too deeply into why this change caused WT-2143 apart from reproducing the regression. I re-instated the comment in ```__wt_log_slot_free``` - since I suspect that was the reason.

sueloverso:
Thanks @agorrod.  Looks like I was overzealous in cleanup.

PR #2226, opened 2015-09-28
WT-2137 Check the sync_lsn is in the correct file before moving it forward.

sueloverso:
@michaelcahill Please review.  After being able to repro the failure, I have fixed it similarly to your analysis.  I also did reduce the wait time to 100msec.  I repro'ed in about 20 iterations of running the test and this has run successfully a few hundred.


michaelcahill:

PR #2227, opened 2015-09-29
WT-2148 Change encoding functions to avoid -Wshift-negative-value

daveh86:


michaelcahill:
Thanks, @daveh86, lgtm.  I'll merge.

PR #2228, opened 2015-09-29
SERVER-20543: Coverity/lint cleanups.

keithbostic:
@sueloverso, I'm hoping we can shut Coverity/lint up about some log.c code.

Feel free to reject any of these changes you don't want, there's nothing wrong here that I can see.


sueloverso:
Thanks @keithbostic.  lgtm.

PR #2229, opened 2015-09-29
WT-2151 Add zero fill config

sueloverso:
@michaelcahill Please review the change for zero-filling logs via a configuration knob.  I'm seeing mixed results on performance.  I see no performance difference on the AWS SSD or my Mac laptop.  On my AWS HDD box I saw no difference running to the HDD file system with small (20Mb) log files.  When I ran to the HDD box with 100Mb files or when I ran to the local disk, I saw about an 80% performance hit.

I also added a log-append-zero workload and I'll add it to the logging jobs once this is merged and plot it against the normal log-append workload.

Of course the trick will be having someone know when they need to configure this.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -74,6 +74,10 @@ __logmgr_config(WT_SESSION_IMPL *session, const char **cfg, bool *runp)
 	if (cval.val != 0)
 		FLD_SET(conn->log_flags, WT_CONN_LOG_ARCHIVE);
 
+	WT_RET(__wt_config_gets(session, cfg, "log.zero_fill", &cval));
+	if (cval.val != 0)
+		FLD_SET(conn->log_flags, WT_CONN_LOG_ZERO_FILL);
+
 	WT_RET(__wt_config_gets(session, cfg, "log.file_max", &cval));
 	conn->log_file_max = (wt_off_t)cval.val;
 	WT_STAT_FAST_CONN_SET(session, log_max_filesize, conn->log_file_max);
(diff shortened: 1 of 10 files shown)


sueloverso:
@michaelcahill This is ready for review.  However this branch needs to be merged with the bug fix from PR #2235 before the reconfigure test will pass.  So please review that one first.  There will also be a conflict in the flags once that is merged into here.

src/conn/conn_log.c
```diff
@@ -74,26 +90,56 @@ __logmgr_config(WT_SESSION_IMPL *session, const char **cfg, bool *runp)
...
-	WT_RET(__wt_config_gets(session, cfg, "log.prealloc", &cval));
 	/*
-	 * If pre-allocation is configured, set the initial number to one.
+	 * If pre-allocation is configured, set the initial number to a few.
 	 * We'll adapt as load dictates.
 	 */
+	WT_RET(__wt_config_gets(session, cfg, "log.prealloc", &cval));
 	if (cval.val != 0)
-		conn->log_prealloc = 1;
+		conn->log_prealloc = 5;
```
michaelcahill: @sueloverso, what's driving this change?  Won't preallocating 500MB of log files by default surprise some users?

michaelcahill:
@sueloverso, I've done the merge and testing is now looking good.  I have a query about the change to the initial value of `log_prealloc`, but the rest of the changes lgtm.

sueloverso:
I ran tests with the current version of the code on the HDD box comparing 1 versus 5 pre-allocated files and it didn't affect the performance of the log-append.wtperf workload when zero'ing.  The idea behind the change was that, when zero'ing, it is better to try really hard to have pre-allocated files so that the I/O penalty does not happen on the log_write/transaction_commit path if we're changing log files.  I have reset it back to 1.

I made other changes recently that may be helping, such as the pre-allocation work thread is now running more frequently so it may be keeping up better than when I iniitally tested zero'ing a few weeks ago.

PR #2230, opened 2015-09-30
WT-2145 Reconciliation code reorganization

agorrod:


agorrod:
@keithbostic This is a start to what I was thinking about with restructuring the reconciliation code.

There are a couple of other things I'd like to do as well:
* Split up some of the very long functions into smaller units.
* Review some of the parameter passing - the functions are tracking a lot of different conditions. A lot of the complexity is necessary, but I'd like to see if we can simplify anything.

I thought I'd get what I have so far into GitHub so you could take a look and let me know if you don't like the direction I'm going in. i.e: Would you mind taking a look through the shape of this change, but not necessarily doing a full review right now.

agorrod:
I realized that I did not describe the changes I have made so far - and the diff is likely to be unhelpful.

I split the content of `rec_write.c` into several different source files. I renamed the functions to reflect their new homes, and adjusted the visibility of those functions as required.

I created one new function: `__rec_row_bnd_split_or_write` because otherwise I couldn't fit a single function call on an 80 character line due to the level of indenting.

There was some conflict between the way I split up the files - for example `rec_bulk.c` has some code that is specific to row vs column stores - that may have been better in `rec_row.c/rec_col.c`. I'm happy to talk about different arrangements.

src/include/extern.h
```diff
@@ -541,12 +569,21 @@ extern void __wt_ovfl_txnc_free(WT_SESSION_IMPL *session, WT_PAGE *page);
...
-extern int __wt_bulk_wrapup(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk);
-extern int __wt_bulk_insert_row(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk);
-extern int __wt_bulk_insert_fix(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk);
-extern int __wt_bulk_insert_var(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk);
+extern int __wt_rec_write_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page);
+extern int __wt_rec_write_init(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags, WT_SALVAGE_COOKIE *salvage, void *reconcilep);
+extern void __wt_rec_destroy(WT_SESSION_IMPL *session, void *reconcilep);
+extern int __wt_rec_block_free( WT_SESSION_IMPL *session, const uint8_t *addr, size_t addr_size);
+extern int __wt_rec_update_move( WT_SESSION_IMPL *session, WT_BOUNDARY *bnd, WT_SAVE_UPD *supd);
+extern int __wt_rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, WT_ROW *rip, WT_CELL_UNPACK *vpack, WT_UPDATE **updp);
```
keithbostic: @agorrod, the functions that were inlined are now real function calls, which worries me. A few years ago, we inlined the calls that were called one or more times for each key/value pair, and I'm not sure about discarding those inlines for function calls.

keithbostic: @agorrod `__rec_row_bnd_split_or_write` is another example, we're doing an additional function call per element on the page, just to decide we don't have to split? (I'd rather the `__rec_row_bnd_split_or_write` code stayed inlined, I think.)

agorrod:
@michaelcahill I talked to Keith about this pull request. He said that he didn't really like the change, but thinks that is because he is too close to the current version. He said that you and I should discuss and decide if it was a worthwhile change. Can talk about it next week?

src/include/extern.h
agorrod: If we decide to go ahead with this change (or some version of it) I'll pull the relevant functions out into a new `.i` file.

agorrod:
@sueloverso Would you mind taking a look at this branch and letting me know if you think it's enough of an improvement to finish and get merged?

To be clear there is more work here before it would be ready to merge:
* It needs a reconcile.i file with a number of inline functions that are no longer inlined.
* I need to resolve the merge conflicts between this and develop.

i.e: I'm not looking for a lgtm - but rather a "I think this is worth pursuing" or "I don't see enough benefit to justify the disruption".

sueloverso:
I do think it is an improvement and you have split the code in logical places.  There are a couple places where I might have something slightly different but rec_write.c is very large.  My mental image of how the `.i` files work, per the discussion above, is that structures and defines go into the `.h`, functions, including static inlined ones, go into a `.c` and non-static inlined functions go into a `.i`.

agorrod:
Thanks Sue.  I'll work on getting the change ready for a real review.

agorrod:
I'm going to close this for now. It's far enough out of date that it makes more sense to start from scratch rather than to backport.

PR #2231, opened 2015-09-30
WT-2152 Add support for WTPERF operations recieving a rollback

daveh86:


daveh86:
@agorrod , would you mind having a look over this one.

This patch is to allow wtperf workloads to continue in the event that they receive a rollback.

bench/wtperf/wtperf.c
```diff
@@ -603,7 +604,20 @@ worker(void *arg)
...
+			 * If we get a rollback error we rollback and continue
+			 */
+			if (ret == WT_ROLLBACK) {
+				if ((ret = session->rollback_transaction(
+				    session, NULL)) != 0) {
+					lprintf(cfg, ret, 0,
+					     "Failed rollback_transaction");
+					goto err;
+				}
+				hit_rollback = 1;
```
agorrod: I'd be inclined to call `session->begin_transaction` here directly after the rollback, rather than tracking whether we had a rollback return in `hit_rollback`.

Otherwise we'll end up doing auto-commit transactions until `ops_per_txn` rolls over, which doesn't feel right to me.

bench/wtperf/wtperf.c
```diff
@@ -603,7 +604,20 @@ worker(void *arg)
 op_err:			lprintf(cfg, ret, 0,
 			    "%s failed for: %s, range: %"PRIu64,
 			    op_name(op), key_buf, wtperf_value_range(cfg));
-			goto err;
+			/*
+			 * If we get a rollback error we rollback and continue
```
agorrod: This comment deserves expansion. I'd like it to mention:
* We only expect to see WT_ROLLBACK if explicit transactions are configured.
* This potentially changes the characteristics of the data set WiredTiger generates. If we don't retry operations after a rollback the data range can have gaps (random-range mode is the other case where this is true).

bench/wtperf/wtperf.c
```diff
@@ -603,7 +604,20 @@ worker(void *arg)
 op_err:			lprintf(cfg, ret, 0,
```
agorrod: I don't think a `WT_ROLLBACK` deserves a first-class error message from wtperf. If it is an expected case, it should be at least level 1.

agorrod:
@daveh86 Can you add a JIRA ticket reference to the title of this pull request?

bench/wtperf/wtperf.c
```diff
@@ -600,7 +600,33 @@ worker(void *arg)
 			if (ret == WT_NOTFOUND)
 				break;
 
-op_err:			lprintf(cfg, ret, 0,
+op_err:			if (ret == WT_ROLLBACK && ops_per_txn != 0) {
+			/*
```
agorrod: The indenting of this comment is wrong - it should match the following code.

bench/wtperf/wtperf.c
```diff
@@ -600,7 +600,33 @@ worker(void *arg)
...
+				if ((ret = session->rollback_transaction(
+				    session, NULL)) != 0) {
+					lprintf(cfg, ret, 0,
+					     "Failed rollback_transaction");
+					goto err;
+				}
+				if ((ret = session->begin_transaction(
+				    session, NULL)) != 0) {
+					lprintf(cfg, ret, 0,
+					    "Worker transaction commit failed");
```
agorrod: The error message doesn't match the failure.

agorrod:
Thanks @daveh86 the shape of this change looks good. I made a couple of style comments.

agorrod:
Thanks @daveh86 lgtm.

PR #2232, opened 2015-09-30
WT-2129 Make cell dump a little safer

daveh86:


daveh86:
@michaelcahill, do you mind having a further look at this?

michaelcahill:
@daveh86 you mention "dump" in the title but this is changing a low-level routine -- what is driving this change?  Also, is there a ticket?

daveh86:
As part of chasing WT-2129.

When dumping a LAS file its possible that there is no dhandle set, so we can attempt to de-reference a null when calling S2BT.

michaelcahill:
@daveh86 I've updated the title to include the ticket.

Can you get a stack trace when this was failing?

The reason I ask is that this will add a little overhead to some very common cases.

agorrod:
@daveh86 I see what this change was solving, but I'm inclined to push the solution into `bt_debug.c` instead of the cell code. Can we do that face to face tomorrow?

daveh86:
@agorrod, that sounds like a good idea.

PR #2233, opened 2015-09-30
SERVER-20303: tuning in-memory splits for when inserting large objects

keithbostic:
@michaelcahill, @agorrod: ready for review and merge.

### DIFF
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -1008,7 +1008,7 @@ __wt_page_can_split(WT_SESSION_IMPL *session, WT_PAGE *page)
 		return (false);
 
 	/*
-	 * There is no point in splitting in-memory if the insert list doesn't 
+	 * There is no point in splitting in-memory if the insert list doesn't
 	 * represent most of the page footprint. Split if there are many items,
 	 * or if there are enough items and the items are a significant part
 	 * of the page's footprint. A 1/4 probability of adding a new skiplist
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -977,6 +977,7 @@ __wt_page_can_split(WT_SESSION_IMPL *session, WT_PAGE *page)
 	WT_BTREE *btree;
 	WT_INSERT_HEAD *ins_head;
 	WT_INSERT *ins;
+	uint64_t mem;
 	int i;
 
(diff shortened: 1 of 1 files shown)


src/include/btree.i
```diff
@@ -1007,25 +1008,33 @@ __wt_page_can_split(WT_SESSION_IMPL *session, WT_PAGE *page)
...
 	    WT_ROW_INSERT_SLOT(page, page->pg_row_entries - 1);
 	if (ins_head == NULL)
 		return (false);
-	for (i = 0, ins = ins_head->head[WT_MIN_SPLIT_SKIPLIST_DEPTH];
-	    ins != NULL; ins = ins->next[WT_MIN_SPLIT_SKIPLIST_DEPTH])
-		if (++i == 4) {
+	for (i = 0, mem = 0, ins = ins_head->head[WT_MIN_SPLIT_SKIPLIST_DEPTH];
+	    ins != NULL; ins = ins->next[WT_MIN_SPLIT_SKIPLIST_DEPTH]) {
+		mem += WT_UPDATE_MEMSIZE(ins->upd);
+		if (++i == 256 || mem > btree->maxleafpage) {
```
agorrod: @keithbostic Could we make this test compare to half the maximum leaf page size?

I'm not sure why we are keeping the test for a certain number of entries at all - if the test here fails, then we'll go on to actually evict the page, which can stall application threads waiting for writing out the large page. I'd have thought that it was always better to do an in-memory split, then have eviction deal with the page when it's no longer hot.

agorrod: Thinking about this some more.. We only want to split when doing an append (mostly) workload, so I understand why we want to keep a test.

I also thought the test was against memory page max, not leaf page max (lazy, sorry). So with the code above a random workload with large items is going to trigger a split.

michaelcahill:
@keithbostic, I've tweaked this a little in #2236 -- can you please review?

src/include/btree.i
keithbostic: @agorrod, generally I agree, and I think the question is if we should err on the side of doing an in-memory split, even if we're not sure it will help? We'll only try it once, so it's not terrible if we're wrong.

PR #2234, opened 2015-09-30
WT-2149: Deadlock opening lookaside table cursor

keithbostic:


michaelcahill:
@keithbostic, there was a problem with the order of creation of the lookaside table in this branch relative to recovery and eviction.  I've fixed that in #2237, can you please take a look?

keithbostic:
@agorrod, @michaelcahill, I've reviewed this again, and it looks correct to me.

I apologize for calling out `session_flags` all over the place, I kept getting it wrong, so figured I'd be explicit.

Anyway, I'm good to go with this change if the two of you agree.

agorrod:
@michaelcahill I'm happy with this change - should I go ahead and merge?

michaelcahill:
@agorrod, thanks -- yes, please!

PR #2235, opened 2015-09-30
WT-2153 Fix bug. Now we always need to start the log_server thread.

sueloverso:
@michaelcahill Please review this bug fix.  I found it when adding tests for reconfiguration in PR #2229.  But it is a bug on its own and deserved its own ticket.


michaelcahill:

PR #2236, opened 2015-10-01
SERVER-20303 Require a minimum item count for in-memory splits.

michaelcahill:
I'm concerned that the first version of this change could split if a single large item is appended to a page.  This version reduces the required count, but only splits if we hit the count *and* the size condition.  

Further, it makes it more explicit that we are sampling by looking at level 2 in the tree.

### DIFF
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -977,8 +977,8 @@ __wt_page_can_split(WT_SESSION_IMPL *session, WT_PAGE *page)
 	WT_BTREE *btree;
 	WT_INSERT_HEAD *ins_head;
 	WT_INSERT *ins;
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, I was concerned that the changes in #2233 would split if a single large entry was appended to a page.  I think this version also makes it clearer what is going on by scanning level 2 of the list.  Can you please review?

michaelcahill:
autobuild: retest this please

keithbostic:
@michaelcahill, lgtm, I'll merge it in.

agorrod:
This was back ported to MongoDB 3.0.8

PR #2237, opened 2015-10-01
WT-2149 Fix the order of creation of the lookaside table

michaelcahill:
The lookaside table has to be created after recovery has completed because it needs a file ID and otherwise recovery could overwrite its entry in the metadata.  However, recovery needs eviction threads running because recovery could touch more data than fits in cache.

Resolve this ordering problem by having recovery start a special set of eviction threads without lookaside table cursors.  Once it is done, they are shut down, the lookaside table can be created, and a new set of eviction threads started.

### DIFF
--- a/src/txn/txn_recover.c
+++ b/src/txn/txn_recover.c
@@ -412,11 +412,12 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
 	WT_RECOVERY r;
 	struct WT_RECOVERY_FILE *metafile;
 	char *config;
-	bool needs_rec, was_backup;
+	bool eviction_started, needs_rec, was_backup;
 
 	conn = S2C(session);
 	WT_CLEAR(r);
 	WT_INIT_LSN(&r.ckpt_lsn);
+	eviction_started = false;
 	was_backup = F_ISSET(conn, WT_CONN_WAS_BACKUP);
 
 	/* We need a real session for recovery. */
@@ -494,6 +495,15 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
 	 */
 	if (needs_rec && FLD_ISSET(conn->log_flags, WT_CONN_LOG_RECOVER_ERR))
 		WT_ERR(WT_RUN_RECOVERY);
+
+	/*
+	 * Recovery can touch more data than fits in cache, so it relies on
+	 * regular eviction to manage paging.  Start eviction threads for
+	 * recovery without LAS cursors.
+	 */
+	WT_ERR(__wt_evict_create(session, false));
+	eviction_started = true;
+
 	/*
 	 * Always run recovery even if it was a clean shutdown.
 	 * We can consider skipping it in the future.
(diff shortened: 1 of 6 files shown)


michaelcahill:
@keithbostic, fixes for #2234, can you please review?

michaelcahill:
autobuild: retest this please

src/txn/txn_recover.c
```diff
@@ -522,6 +532,15 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
...
 	__wt_free(session, config);
+
+	/*
+	 * Destroy the eviction threads that were started in support of
+	 * recovery.  They will be restarted once the lookaside table is
+	 * created.
+	 */
+	if (eviction_started)
+		WT_TRET(__wt_evict_destroy(session));
+
```
keithbostic: @michaelcahill, I'd suggest moving these calls outside of this function, wrapping `__wt_txn_recover`, if that's possible. Looking over the code, I don't see why it wouldn't be possible, but I could easily be missing something.

keithbostic:
@michaelcahill, aside from one comment, lgtm.

I really wish there was a better option, but I can't think of one.

src/txn/txn_recover.c
agorrod: @keithbostic It was me that pushed @michaelcahill to encapsulate recovery in a single function. Since the eviction threads are being started and stopped solely for recovery I thought it made more sense to do that within the `__wt_txn_recover` method.

Why would you prefer the calls to be outside the function? Are you thinking something like `WT_WITH_EVICTION_THREADS(__wt_txn_recover(...));`  (psuedo code, I doubt we'd actually use a macro) is clearer?

keithbostic: It hides information that's better exposed -- it would never occur to me to look in a function named `__wt_txn_recover` to reason about when eviction is running. When looking at `__wt_connection_workers` to understand the startup code, it's easy to see a start/stop of eviction, but I'm never going to think one of the underlying functions also starts/stops eviction threads.

I worry somebody will shuffle the function calls in `__wt_connection_workers` for some reason, and things will stop working because they don't realize `__wt_txn_recover` isn't just recovery.

Secondarily, it's more complicated to do that work in `__wt_txn_recover` than to wrap that function in calls to start/stop eviction, so it's more error-prone.

michaelcahill: We'd have to split recovery into pieces to do what you're suggesting -- we don't want the overhead of starting and stopping eviction every time we start, only if a real recovery pass is required.  At the moment, we only determine that in the middle of `__wt_txn_recover`.

And I disagree about `__wt_connection_workers` -- putting the start and stop in one place *avoids* the kind of ordering issues you are worried about.

To be clear, I don't love this change -- it made me sad that something like this is required.  That said, I started with the kind of solution you're describing and I don't think it's better than this.

keithbostic: I see what you mean about "avoiding ordering issues", I hadn't thought of it that way and I'd agree there are similar risks with both approaches.

The only solution I see is passing back a side-effect, that eviction was started inside recovery, and leaving the eviction-destroy outside of the recovery code, that guarantees anyone working in `__wt_connection_workers` will realize what's going on. But there are some obviously ugly things there, too.

So: if Alex and you have tried it the other way and this is better, I'm good to go.

PR #2238, opened 2015-10-01
WT-2154 - Make btree dump safer

daveh86:


daveh86:
@agorrod, when you get a moment would you mind taking a look?

agorrod:
@daveh86 I don't see any comments. We'd talked about encapsulating all of the changes in `bt_debug.c` - but I see a change to `bt_vrfy.c`.

Could you switch the asserts from being `session->dhandle != NULL` to `S2BT_SAFE(session) != NULL`

agorrod:
@daveh86 I'm not sure if you wanted another review for this. A couple more comments:

* Some of the comments appear incomplete.
* Re the comments referencing why a btree handle is in the parameter list. The should mention that the functions may be called from within a debugger - so setting a btree handle in calling code is not  practical.

daveh86:
please retest this

daveh86:
retest this please

daveh86:
@agorrod, can you have a further look at this?

src/btree/bt_debug.c
```diff
@@ -524,11 +547,16 @@ __wt_debug_page(WT_SESSION_IMPL *session, WT_PAGE *page, const char *ofile)
 
 /*
  * __debug_tree --
- *	Dump the in-memory information for a tree.
+ *	Dump the in-memory information for a tree. Takes an explicit btree
```
agorrod: Why is this comment different to the two equivalent comments above?

src/btree/bt_debug.c
```diff
@@ -524,11 +547,16 @@ __wt_debug_page(WT_SESSION_IMPL *session, WT_PAGE *page, const char *ofile)
...
+ *	as one may not be set on the session. This is often the case as this
+ *	function will be called from within a debugger, which makes setting a
```
agorrod: This version of the comment reads: 
"Takes an explicit btree as one may not be set on the session."

The other version reads:
"Takes an explicit btree as an argument, as one may not yet be set on the session."

I also don't think the final sentence in the comment adds value - it's obvious from the implementation that it's using ```WT_WITH_BTREE```

agorrod:
lgtm

PR #2239, opened 2015-10-01
WT-2155 Remove last use of F_CAS_ATOMIC and the associated macro.

agorrod:
It's not trivial to remove all atomic flags primitives, since the page flags rely on them. This part of the change is simple, so it might as well be done quickly.


michaelcahill:

PR #2241, opened 2015-10-04
WT-2126: clean up if there is an error during splits

keithbostic:
@michaelcahill, here's the change to panic if we error after corrupting the original page during a split.


michaelcahill:
@keithbostic, I think the panic is necessary now, but on re-reading the comment that you're removing here about "copying" changes to the new pages, I wondered whether we could make that work for real.

In other words, how about we don't clear the update from the original page, so it does remain valid?  Then we could plumb a flag down through `__wt_page_out` specifying whether to free the updates (normally true, but in this split path, false).

That way, we could discard the original page (but not the updates) if the split succeeds, or discard the split pages (but not the updates) if the split fails.  I know it won't be quite as simple as this, but this might make it feasible to get the error handling "right"?

keithbostic:
@michaelcahill, I thought of the same thing when I was coding up this change, and I agree there's a path there. I think what we have to do though, is split the update chain handling into two passes:

1. Walk the list of updates and re-create them on the new in-memory page (if fail, discard the new page with the flag to ignore update chains)
2. Walk the list of updates and NULL-terminate the reference from the old page (can't fail)
3. Discard the old page

Is it worth doing?

michaelcahill:
@keithbostic, isn't 2 + 3 the same as "discard the old page with the flag to ignore update chains"?

I think this is worth a try, but if it gets too ugly I'll merge the panic change rather than continuing with corrupt data.

keithbostic:
@michaelcahill, the problem is the old page has update chains we didn't have to instantiate on the new page, and we have to free those when discarding the old page -- it's only the selected set of update chains that are in motion.

keithbostic:
@michaelcahill, commit c6be1a0 gets half the way there, we don't have to panic, but we do leak some memory if we fail in the middle of creating the new pages and splitting the parent. We could (1) drill the API hole you suggested into the page-discard functions and not discard any update chains from the newly created pages (this will work, unlike using that technique for the original page, because all of the update chains on the new page are ones we've put there); (2) use a page flag instead of an API hole, (3) add functions to explicitly discard update chains from the new page; (4) leak the memory.

I admit to a preference for anything but (1), that hole is going to be deep/ugly.

keithbostic:
autobuild: retest this please

michaelcahill:
@keithbostic, let's go with door (3) to clear all updates from the newly allocated pages before doing an ordinary discard.

keithbostic:
@michaelcahill, I coded up (3), but it was scary, handling all of the possible update lists for the different page types is a chunk of new code that mostly looks like code in the page discard functions, and it's going to be tricky to maintain outside of that code.

I kept that code around if you prefer it, but what I pushed is a new page flag that causes the page discard code to ignore any updates it finds. That's not as localized a change, but I have a lot more confidence it's correct/maintainable.

Let me know what you think.

There's also some work we need to do to survive a failed split. Here's the interesting stack:
```
(gdb) where
#0  __evict_lru_pages (session=0x803455f00, is_server=true)
    at src/evict/evict_lru.c:803
#1  0x00000000004367f5 in __evict_server_work (session=0x803455f00)
    at src/evict/evict_lru.c:923
#2  0x0000000000434f55 in __evict_pass (session=0x803455f00)
    at src/evict/evict_lru.c:573
#3  0x0000000000432c7f in __evict_server (arg=0x803455f00)
    at src/evict/evict_lru.c:169
```
We return from `__evict_page` with ENOMEM, which then returns through `__evict_lru_pages` and `__evict_server_work` into `__evict_pass`, which panics:
```
WT_PANIC_MSG(session, ret, "cache eviction server error");
```
In other words, this change is necessary, but not sufficient. We could convert non-panic split-function errors into `EBUSY`, or the server could ignore non-panic errors, or something else?

michaelcahill:
Thanks, @keithbostic -- I don't love the page flag to change discard behavior, but it looks correct to me, and I don't see an obviously better alternative.  I'll merge...

keithbostic:
@michaelcahill, should we create a ticket to make this change work out through eviction? Or just let sleeping dogs lie...?

PR #2242, opened 2015-10-05
WT-2159: don't check the config twice in one path.

keithbostic:
@agorrod or @sueloverso, ready for review & merge.


sueloverso:
@keithbostic lgtm.  Once the tester is done, I'll merge.

PR #2243, opened 2015-10-07
WT-2157 If we give up trying to split a page, make sure it is written by the next checkpoint

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm. I'll merge.

PR #2244, opened 2015-10-07
WT-2149 Track if the lookaside table is open when starting eviction workers

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm. As discussed this change should fix the current Jenkins compat-r1 test failures as well - since we'll no longer attempt to open the LAS file during recovery (which is a problem if it doesn't exist because we are recovering a database created by a pre-LAS WiredTiger).

PR #2245, opened 2015-10-08
WT-2162.  Add null pointer check, needed after an index is dropped.

ddanderson:
The WT_TABLE.indices array is zeroed after an index is dropped, and is filled on demand. Null pointer checks are needed unless __wt_schema_open_indices() is called first, which repopulates the array.

Did a review to verify that all uses of WT_TABLE.indices guard against null pointers or call __wt_schema_open_indices first.


ddanderson:
@michaelcahill, this is a quick fix, can you review?

agorrod:

PR #2246, opened 2015-10-09
WT-2139 At read-uncommitted isolation, keep the same ID pinned while cursors are positioned

michaelcahill:


michaelcahill:
After a lot of debugging, @agorrod and I realized that this issue isn't actually about LSM as such: that's just the place where we stress out having multiple cursors positioned simultaneously.

What's going on during an LSM WT_CURSOR::next operation is that we have a set of chunk cursors positioned, then we move the earliest one(s) forward, setting `cursor->value` at this point.  Then take the key/value pair from the smallest remaining cursor.

The session's `snap_min` value was being updated between every operation, so by the time we examined the value in one of the chunk cursors, it was possible for the referenced value to be obsolete and for another thread to free it.

michaelcahill:
Note that this change is not ideal.  In particular, it means that a read-uncommitted cursor will keep a single transaction ID pinned for an entire scan.  That isn't required for a single cursor, but it isn't simple to detect this case.

michaelcahill:

keithbostic:
@michaelcahill, @agorrod, lgtm, merged.

While looking at this change, I noticed the comment in `__wt_txn_cursor_op()` is out of date:
```
/*
 * If there is no transaction running (so we don't have an ID), and no
 * snapshot allocated, put an ID in the global table to prevent any
 * update that we are reading from being trimmed to save memory.  Do a
 * read before the write because this shared data is accessed a lot.
 *
 * !!!
 * Note:  We are updating the global table unprotected, so the
 * oldest_id may move past this ID if a scan races with this
 * value being published.  That said, read-uncommitted operations
 * always take the most recent version of a value, so for that version
 * to be freed, two newer versions would have to be committed.  Putting
 * this snap_min ID in the table prevents the oldest ID from moving
 * further forward, so that once a read-uncommitted cursor is
 * positioned on a value, it can't be freed.
 */
```
The part "If there is no transaction running" comes from a version where `F_ISSET(txn, TXN_RUNNING)` was part of the test, and "the oldest_id may move past this ID" comes from a version where we compared `snap_min` against `txn_global->oldest_id`, neither of which is still true.

Can somebody please  update the comment? I would, but my chances of getting it right aren't good.

michaelcahill:
Done: ebec2197e7f67be432e4b3c7c2eabfc4f3968bb1.

The second paragraph was still valid, it had nothing to do with comparing `snap_min` against `txn_global->oldest_id` -- I've tried to make that clearer.  See the matching comment in `__wt_txn_update_oldest` for more context.

agorrod:
MongoDB doesn't use read-uncommitted, there is no value in back porting this change.

PR #2247, opened 2015-10-12
SERVER-19954 Don't scan tracked handles during checkpoints.

michaelcahill:
The change in WT-147 to allow a session to lock the same handle exclusive multiple times makes this unnecessary.


agorrod:
Thanks @michaelcahill lgtm since this branch passes the test case the workaround was originally introduced for.

PR #2248, opened 2015-10-12
WT-2068: complain if the live system is re-opened.

keithbostic:
@sueloverso, here's the change to detect if the live system is re-opened. I've smoke-tested it, but not much more.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -52,6 +52,7 @@ __wt_block_checkpoint_load(WT_SESSION_IMPL *session, WT_BLOCK *block,
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	uint8_t *endp;
+	bool is_open;
 
 	WT_UNUSED(addr_size);
 	ci = NULL;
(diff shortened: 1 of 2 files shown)


src/block/block_ckpt.c
```diff
@@ -83,11 +84,19 @@ __wt_block_checkpoint_load(WT_SESSION_IMPL *session, WT_BLOCK *block,
...
 		 */
+		__wt_spin_lock(session, &block->live_lock);
+		is_open = block->live_open;
+		if (!is_open)
+			block->live_open = true;
+		__wt_spin_unlock(session, &block->live_lock);
+		if (is_open)
+			WT_ERR_MSG(session, WT_ERROR,
+			    "%s: open failed, live system already open",
+			    block->name);
```
sueloverso: I have two comments. 1. Should we put all of this change under HAVE_DIAGNOSTICS?  2.  Given the comment, why not just `WT_ASSERT(session, block->live_open == false)` inside the spin lock instead of returning an error message?

keithbostic: Given the way we end up debugging MongoDB failures, I think we should move away from HAVE_DIAGNOSTICS and asserts, and more towards error paths that dump out a message into the log, even if it's "not supposed to happen".

sueloverso: It just seems a shame to have to take a lock unconditionally, every time, for diagnostics and something that isn't supposed to happen.

keithbostic: Yes, but it's a file open, not a performance path. We could probably get rid of the lock and use a CAS instruction instead, but shrug.

PR #2249, opened 2015-10-12
WT-2126: __split_deepen() doesn't really need to panic

keithbostic:
@michaelcahill, I'm open to going the other way, doing a panic in both `__split_parent` and `__split_deepen`. It's unclear how long we're going to run after this failure anyway.

Maybe we should just quit when malloc fails? That would simplify a lot of error handling, and once we're out of memory we're almost certainly taking rarely tested error paths, nothing good is going to happen.


michaelcahill:
This lgtm, thanks @keithbostic.  I'll merge.

PR #2250, opened 2015-10-12
WT-2167 Switch recovery to using an internal session.

agorrod:
It used to open and close a session handle that wasn't marked
internal. No functional change since recovery closes the handle
when it is done and the primary difference between internal and
non-internal sessions is during connection close.


michaelcahill:
Thanks, @agorrod, lgtm.  I'll merge...

PR #2251, opened 2015-10-13
WT-2068 Use dhandle->close_lock to prevent checkpoint re-opening a file in the process of being closed

michaelcahill:


michaelcahill:
This approach doesn't work -- while we might be able to work around the deadlock this hits, it feels like the wrong direction.  Closing.

PR #2253, opened 2015-10-13
WT-2068 Put live_open usage under HAVE_DIAGNOSTIC and an assert.

sueloverso:
@keithbostic and @michaelcahill - after thinking about this for a while, I still feel that this code really should be under HAVE_DIAGNOSTIC and an assertion.  I have not heard that we want to move away from that for things like this.  But this seems cleaner to me.  I made it a sub-branch off Keith's so it can be tossed if you want.

I definitely think we want some form of this because it quickly uncovered WT-2068 and we may need it again as we come up with a fix.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -52,7 +52,6 @@ __wt_block_checkpoint_load(WT_SESSION_IMPL *session, WT_BLOCK *block,
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	uint8_t *endp;
-	bool is_open;
 
 	WT_UNUSED(addr_size);
 	ci = NULL;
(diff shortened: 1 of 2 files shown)


keithbostic:
@sueloverso, @michaelcahill: I think I disagree.

My impression is we're seeing an increasing number of reports that boil down to "a problem happened, here's the MongoDB log". I think we should make an effort to move toward verbose error messages in the log, rather than HAVE_DIAGNOSTIC tests and WT_ASSERTs.

I would also support more builds with WT_ASSERT set (WT_ASSERT depends on configuring with HAVE_DIAGNOSTIC, and we're the only ones that do that); is there some way we could get builds other than release builds to configure with WT_ASSERT? That would require review to make sure we don't have any slow WT_ASSERT calls, but that's pretty easy to do.

sueloverso:
I am running a production build of develop versus a production build of develop with WT_ASSERT outside HAVE_DIAGNOSTIC with a representative sample of 8 wtperf configs on the AWS HDD instance.  I'll let you know what I see.  I did not do any kind of review to find any slow ones.

michaelcahill:
@keithbostic, asserts are often more useful than log messages.  I'm not sure what MongoDB tickets you're referring to, but a stack trace is the most useful thing we get from logs and error messages don't give us that.  The full MongoDB test sure is run against debug builds of MongoDB, so asserts will fire for anything covered by the test suite.

keithbostic:
@michaelcahill, we've had stack traces based on an assert for this one for weeks; this is an easy mistake to make, a difficult issue to debug, and I'd like to leave the error message in place, in all cases.

sueloverso:
My run of 8 wtperf configs in both a develop tree and a with-assert tree shows that with the asserts in a production build it ran as fast or even faster in all cases.  (Sometimes significantly faster.)  I did only run each config once on each tree so there would likely be variation.  

So production asserts don't seem to be a performance drag.

sueloverso:
I agree it was a painful and slow path from the original assert to the assert I put into the debug branch based on Keith's change.

It was only when this particular error case became an assert that we got stack traces for the racing threads.  The original assertion was only detected after-the-fact.  If we only return an error back up to the caller we lose the stack when it was detected and likely the racing thread by the time it errors at the top-level.  Other than letting us know that the bug still exists in the system, the error message itself does not actually help diagnose or debug the problem.

keithbostic:
@sueloverso, fair enough -- we did need stacks after the message, and as @michaelcahill points out, we do get stacks from MongoDB.

OK, I agree, let's go with the asserts.

PR #2254, opened 2015-10-14
WT-2068 Protect discarding handles with the handle list lock.

michaelcahill:


sueloverso:
Thanks @michaelcahill.  This lgtm.  I've backported to my debugging branch.  With the assert that branch hit the problem in about 90 minutes.  I'll merge after that runs successfully for a few hours.

PR #2255, opened 2015-10-14
WT-2170 Add turtle file lock.

sueloverso:
@michaelcahill Please review this change.  It contains only the turtle file lock portion discussed in the ticket.  I have not yet been able to reproduce the original problem without the fix (even adding in a 1 msec sleep call before the rename).


michaelcahill:
Thanks, @sueloverso: lgtm, I'll merge.

Let's talk tomorrow about the related issue of flushing the metadata during a checkpoint.

PR #2256, opened 2015-10-16
WT-2174 Avoid the table list lock when creating a size only statistics cursor.

agorrod:
When there are lots of tables being created and dropped it can take a long time to get the table lock.

### DIFF
--- a/src/schema/schema_stat.c
+++ b/src/schema/schema_stat.c
@@ -73,27 +73,30 @@ __curstat_size_only(WT_SESSION_IMPL *session,
 	WT_CLEAR(namebuf);
 	*was_fast = false;
 
-	if (!F_ISSET(cst, WT_CONN_STAT_SIZE))
-		return (0);
-
+	/* Retrieve the metadata for this table. */
 	WT_RET(__wt_metadata_search(session, uri, &tableconf));
-	WT_ERR(__wt_config_getones(session, tableconf, "columns", &colconf));
 
+	/*
+	 * The fast path only works if the table consists of a single file
+	 * and does not have any indexes. The absence of named columns is how
+	 * we determine that neither of those conditions can be satisfied.
+	 */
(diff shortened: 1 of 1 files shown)


src/schema/schema_stat.c
```diff
@@ -53,6 +53,60 @@ err:	__wt_scr_free(session, &buf);
...
+
+	if (!F_ISSET(cst, WT_CONN_STAT_SIZE))
+		return (0);
+
+	WT_RET(__wt_metadata_search(session, uri, &tableconf));
+	WT_ERR(__wt_config_getones(session, tableconf, "columns", &colconf));
+
+	WT_ERR(__wt_config_subinit(session, &cparser, &colconf));
+	if ((ret = __wt_config_next(&cparser, &ckey, &cval)) == 0)
+		goto err;
```
michaelcahill: This could use a comment: "Checking for a simple table with no column names.  Otherwise, there may be multiple sources involved and we have to take the slow path."

src/schema/schema_stat.c
```diff
@@ -53,6 +53,60 @@ err:	__wt_scr_free(session, &buf);
...
+	WT_DECL_RET;
+	WT_ITEM namebuf;
+	wt_off_t filesize;
+	char *tableconf;
+
+	WT_CLEAR(namebuf);
+	*was_fast = false;
+
+	if (!F_ISSET(cst, WT_CONN_STAT_SIZE))
+		return (0);
```
michaelcahill: I understand that this test will probably be inlined, but I think it would be clearer to check in the calling code.

agorrod:
Thanks for the review @michaelcahill . I've made the changes you suggested @daveh86 confirmed that this change improves https://jira.mongodb.org/browse/SERVER-20961 - could youmerge if you are happy now?

michaelcahill:
Thanks, @agorrod, lgtm.  I'll merge...

PR #2257, opened 2015-10-20
WT-2178 In-memory storage engine support

michaelcahill:
Not ready for merge, this is the working branch for in-memory behavior.


agorrod:
I've made a change to wtperf so that i can test out in memory configurations. I've been using the following configuration file:

```
conn_config="cache_size=500MB,in_memory=true,file_manager=(close_idle_time=0),log=(enabled=false)"
table_config="type=file"
icount=500000
reopen_connection=false
report_interval=5
run_time=20
populate_threads=1
threads=((count=8,reads=1))
```

agorrod:
@keithbostic Would you mind taking a look at this branch? I'm happy to chat with you tomorrow if you have questions.

agorrod:
For the record the test failure isn't completely unexpected.  I added a new test case that is timing dependant, I need to make it more reliable before the branch is merged. 

There are two things I thought you might have an opinion about. The new error return code and the new method that does configuration checking.

src/include/api.h
```diff
@@ -116,10 +116,17 @@
...
+	TXN_API_CALL_NOCONF(s, WT_CURSOR, remove, cur,			\
+	    ((bt) == NULL) ? NULL : ((WT_BTREE *)(bt))->dhandle);
+
 #define	CURSOR_UPDATE_API_CALL(cur, s, n, bt)				\
 	(s) = (WT_SESSION_IMPL *)(cur)->session;			\
 	TXN_API_CALL_NOCONF(s, WT_CURSOR, n, cur,			\
-	    ((bt) == NULL) ? NULL : ((WT_BTREE *)(bt))->dhandle)
+	    ((bt) == NULL) ? NULL : ((WT_BTREE *)(bt))->dhandle);	\
+	if (F_ISSET(S2C(s), WT_CONN_IN_MEMORY) && __wt_cache_full(s))	\
+		return (WT_CACHE_FULL);
```
michaelcahill: Can this be turned into `WT_ERR`.

dist/api_err.py
```diff
@@ -51,6 +51,11 @@ def __init__(self, name, value, desc, long_desc=None, **flags):
         'recovery must be run to continue', '''
         This error is generated when wiredtiger_open is configured
         to return an error if recovery is required to use the database.'''),
+    Error('WT_CACHE_FULL', -31807,
+        'operation would overflow cache', '''
+        This error is generated when wiredtiger_open is configured
+        to run in-memory, and an insert or update operation requires more
+        than the configured cache size to complete.'''),
```
keithbostic: @agorrod, should we avoid the word "cache" as much as possible, because we have so many "cache" statistics and information already? Here, we could do `WT_INMEMORY_FULL: operation would overflow in-memory configuration`.

src/btree/bt_split.c
```diff
@@ -773,7 +773,8 @@ __split_multi_inmem(
 	 * transaction to an impossibly old value so this page is never skipped
 	 * in a checkpoint.
 	 */
-	page->modify->first_dirty_txn = WT_TXN_FIRST;
+	if (page->modify != NULL)
+		page->modify->first_dirty_txn = WT_TXN_FIRST;
```
keithbostic: @agorrod, I don't understand why this change would be necessary; how can we be here if there's no `page->modify` structure?

src/conn/conn_api.c
```diff
@@ -1643,6 +1643,45 @@ __wt_verbose_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+			/*
+			 * It is not necessary to check for log based
+			 * automatic checkpoints since logging must be
+			 * disabled.
+			 */
+			WT_RET_MSG(session, EINVAL,
+			    WT_CONN_CONFCHK_PREFIX "checkpoints");
+	}
+	return (0);
+}
```
keithbostic: @agorrod, I think each subsystem should be responsible for doing its own configuration validation, that is, `__wt_sweep_config()` should own checking if something else is configured that is a problem for the sweep configuration.

And, I think it should be a configuration check, that is, we should check for problematic configuration strings instead of checking some other subsystem's global configuration information. Arguably, `WT_CONN_IN_MEMORY` is global information, though, I don't mind checking it in other places.

src/evict/evict_lru.c
```diff
@@ -1216,6 +1216,10 @@ __evict_walk_file(WT_SESSION_IMPL *session, u_int *slotp)
 		page = ref->page;
 		modified = __wt_page_is_modified(page);
 
+		/* Don't add clean pages if we are running in-memory */
+		if (F_ISSET(S2C(session), WT_CONN_IN_MEMORY) && !modified)
+			continue;
+
```
keithbostic: @agorrod, I'd move this down about 10 lines and merge with the existing test for skipping clean pages, doesn't that allow us to get rid of empty pages, which is a good thing?

keithbostic: @agorrod, there's a local `conn` variable.

src/evict/evict_page.c
```diff
@@ -106,7 +106,8 @@ __wt_evict(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 		conn->cache->evict_max_page_size = page->memory_footprint;
 
 	/* Update the reference and discard the page. */
-	if (mod == NULL || mod->rec_result == 0) {
+	if ((mod == NULL || mod->rec_result == 0) &&
+	    !F_ISSET(S2C(session), WT_CONN_IN_MEMORY)) {
```
keithbostic: @agorrod, there's a local `conn` variable.

src/evict/evict_page.c
```diff
@@ -324,6 +326,15 @@ __evict_review(
...
 
+	modified = __wt_page_is_modified(ref->page);
+
+	/*
+	 * Clean pages can't be evicted when running in memory only. This
+	 * should be uncommon - we don't add clean pages to the queue.
+	 */
+	if (F_ISSET(S2C(session), WT_CONN_IN_MEMORY) && !modified)
+		return (EBUSY);
+
```
keithbostic: @agorrod, there's another call to `__wt_page_is_modified` about 10 lines farther on that could be replaced with `modified`.

src/reconcile/rec_write.c
```diff
@@ -5491,7 +5494,8 @@ __rec_write_wrapup(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
 		 * of forced eviction: set up a single block as if to split,
 		 * then use that block to rewrite the page in memory.
 		 */
-		if (F_ISSET(r, WT_EVICT_UPDATE_RESTORE) && bnd->supd != NULL)
+		if (bnd->supd != NULL ||
+		    F_ISSET(S2C(session), WT_CONN_IN_MEMORY))
```
keithbostic: @agorrod: why aren't we checking `WT_EVICT_IN_MEMORY` here?

src/reconcile/rec_write.c
```diff
@@ -5639,7 +5643,8 @@ __rec_split_row(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
 		WT_RET(__wt_row_ikey_alloc(session, 0,
 		    bnd->key.data, bnd->key.size, &multi->key.ikey));
 
-		if (F_ISSET(r, WT_EVICT_UPDATE_RESTORE) && bnd->supd != NULL) {
+		if (bnd->supd != NULL ||
+		    F_ISSET(S2C(session), WT_CONN_IN_MEMORY)) {
```
keithbostic: @agorrod: why aren't we checking `WT_EVICT_IN_MEMORY` here?

src/reconcile/rec_write.c
```diff
@@ -5679,7 +5684,8 @@ __rec_split_col(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
 	    bnd = r->bnd, i = 0; i < r->bnd_next; ++multi, ++bnd, ++i) {
 		multi->key.recno = bnd->recno;
 
-		if (F_ISSET(r, WT_EVICT_UPDATE_RESTORE) && bnd->supd != NULL) {
+		if (bnd->supd != NULL ||
+		    F_ISSET(S2C(session), WT_CONN_IN_MEMORY)) {
```
keithbostic: @agorrod: why aren't we checking `WT_EVICT_IN_MEMORY` here?

keithbostic:
@agorrod: in two places, `multi->supd == NULL` changed to `multi->supd == NULL && multi->supd_dsk == NULL`, and I don't understand that change. Why was it necessary, what's going on?

keithbostic:
@agorrod, @michaelcahill:

I think we should be setting `WT_BTREE_IN_MEMORY`, so we don't bother with hazard locks.

There's also clear overlap with `WT_BTREE_NO_EVICTION`. Ideally, I think we should be able to replace most of the tests we've added for `WT_CONN_IN_MEMORY` with `WT_BTREE_NO_EVICTION`, and where that's wrong, it's a problem with `WT_BTREE_NO_EVICTION` that we should fix.

keithbostic:
@agorrod, @michaelcahill: I've added some comments, I'm happy to see this merged.

michaelcahill:
As discussed, `WT_BTREE_IN_MEMORY` is too restrictive -- we need "eviction" in these trees to clean up obsolete updates and rewrite pages with lots of removes.  I agree that we could set a btree flag rather than checking the connection flag everywhere (and applying the same logic to metadata so it can also be cleaned up by eviction threads).  I'll take a look at that change.

src/btree/bt_split.c
agorrod: The modify structure can be NULL, because in this brave new world we may be creating clean pages when doing an in-memory split. It used to be the case that we only re-instantiated pages when updates couldn't be evicted, a ``WT_EVICT_IN_MEMORY`` split potentially (hopefully) creates a number of clean pages.

I'll try to get that context into a comment.

src/evict/evict_page.c
agorrod: fixed

dist/api_err.py
michaelcahill: I'm sympathetic to both points of view, but the setting that configures the amount of memory available is going to remain "cache_size", and all of the relevant statistics will be "cache" statistics.  I don't think we can hide that from in-memory users without major surgery.

How about we make this `undoc=True` for now, since the whole feature is currently undocumented: that buys us some time to figure out how we're going to deal with this in MongoDB?

agorrod:
@michaelcahill Would you mind reviewing this change?

michaelcahill:
Thanks, @agorrod -- I am happy with this change for now, I'll merge.  I'm sure there will be more in-memory work post-integration.

PR #2258, opened 2015-10-20
WT-2073: metadata cleanup

keithbostic:
@michaelcahill, @agorrod, for your review.

There's a nasty potential bug this change introduces, where code is doing a loop using the session's cached metadata cursor, and some underlying function uses the metadata cursor itself, causing the top-level's use of the cursor to be mis-positioned in the loop. I don't have a good solution for that, or even a way to detect it -- I'd like to do something better in that case than what I do now, which is fail at some indeterminate point in the future, and that's not good. Looking for a good idea on that one, maybe we can talk it over.


michaelcahill:
@keithbostic, it sounds like you'll need `__wt_metadata_cursor_release` replacing the existing calls to `cursor->close` for metadata cursors.  Then add some session state tracking whether the session-wide metadata cursor is in use, and assert that it isn't in use at the start of  `__wt_session_metadata_cursor`.

src/meta/meta_table.c
```diff
@@ -64,44 +64,22 @@ __wt_metadata_open(WT_SESSION_IMPL *session)
...
 /*
- * __wt_metadata_cursor --
- *	Opens a cursor on the metadata.
+ * __wt_metadata_session_cursor --
+ *	Opens the session's cached metadata cursor.
  */
 int
-__wt_metadata_cursor(
-    WT_SESSION_IMPL *session, const char *config, WT_CURSOR **cursorp)
+__wt_metadata_session_cursor(WT_SESSION_IMPL *session, WT_CURSOR **cursorp)
```
michaelcahill: This name change seems wrong to me.  Every cursor is already tied to a session, so adding "session" to the name doesn't make the intent clear.  How about calling this `__wt_metadata_cursor` and the other one `__wt_metadata_cursor_open`?

keithbostic: Agreed & pushed.

keithbostic:
@michaelcahill, I don't want to go to an open/close pair, that makes all the callers ugly. If there's no other solution... (and I admit I can't think of one).

keithbostic:
@michaelcahill, I left `__wt_txn_recover` opening a new metadata cursor rather than using the cached one, there's a comment:
```
/*
 * We no longer need the metadata cursor: close it to avoid pinning any
 * resources that could block eviction during recovery.
 */
r.files[0].c = NULL;
WT_ERR(metac->close(metac));
```
which doesn't make much sense to me, I don't know why it would be pinning resources here. Should I convert this code to using the cached cursor?

src/conn/conn_dhandle.c
```diff
@@ -708,6 +708,15 @@ __wt_conn_dhandle_discard(WT_SESSION_IMPL *session)
...
 
+	/*
+	 * The connection may have an open metadata cursor handle. We cannot
+	 * close it before now because it's potentially used when discarding
+	 * other open data handles. Close it before discarding the underlying
+	 * metadata handle.
+	 */
+	if (session->meta_cursor != NULL)
+		WT_TRET(session->meta_cursor->close(session->meta_cursor));
+
```
keithbostic: @michaelcahill, I meant to flag this change for your review. I think it's correct, but maybe there's something that's less fragile?

michaelcahill:
@keithbostic, I thought there was still work to do here to make this change safe (by checking that there is no use of `session->meta_cursor` during a loop).

I can see the special cases that call `__wt_metadata_cursor_open`: maybe we should stop exporting `__wt_metadata_cursor` if it isn't safe?

keithbostic:
@michaelcahill, assuming this passes testing, I think it's ready for review and merge.

A couple of open questions:

* I flagged one change for close review in `__wt_conn_dhandle_discard()`.
* There's an open question about `__wt_txn_recover()` in the coversation above this comment.

keithbostic:
autobuild: retest this please

src/conn/conn_dhandle.c
michaelcahill: That seems fine to me -- this code is already special-casing metadata.

src/include/dhandle.h
```diff
@@ -33,6 +33,10 @@
 	(F_ISSET(dhandle, WT_DHANDLE_DEAD) ||				\
 	!F_ISSET(dhandle, WT_DHANDLE_EXCLUSIVE | WT_DHANDLE_OPEN))
 
+/* Cursor's data handle. */
+#define	WT_CURSOR_DHANDLE(c)						\
+	(((WT_CURSOR_BTREE *)(c))->btree->dhandle)
+
```
michaelcahill: I'd prefer a macro called `WT_SESSION_META_DHANDLE` that takes a session and looks up the data handle `session->meta_cursor` to find the data handle for this one case.  This only works on btree cursors (so the name is misleading) and is only ever used on `session->meta_cursor`.

src/conn/conn_dhandle.c
michaelcahill: On second thought, does it make sense to move this into `__wt_session_close_cache`?

keithbostic: I think that gets messy -- but I may just not understand the plan.

First, `__session_close` has a separate loop that closes open cursors before it calls `__wt_session_close_cache`, and so we'd attempt to close the metadata cursor twice, so I think there's an issue there.

Second, `__wt_conn_dhandle_discard` calls `__wt_session_close_cache` twice (before and after closing other open data handles), so wouldn't  that mean we'd close the metadata cursor, then close some other handle that would re-open the metadata cursor, and then close it again?

src/include/dhandle.h
keithbostic: > I'd prefer a macro called `WT_SESSION_META_DHANDLE` ...

Yes, that  makes sense to me, too, it's better that way.

Done and pushed.

src/cache/cache_las.c
```diff
@@ -161,7 +163,7 @@ __wt_las_cursor_create(WT_SESSION_IMPL *session, WT_CURSOR **cursorp)
 	 * opens (the first update is safe because it's single-threaded from
 	 * wiredtiger_open).
 	 */
-	btree = S2BT(session);
+	btree = ((WT_CURSOR_BTREE *)(*cursorp))->btree;
```
agorrod: This change appears a few times, is it a consequence of no longer saving and restoring the ``session_dhandle`` in ``__meta_btree_apply``? I'd rather keep the ``S2BT`` semantic if possible.

src/meta/meta_table.c
```diff
@@ -64,44 +64,79 @@ __wt_metadata_open(WT_SESSION_IMPL *session)
...
-	return (ret);
+	if ((cursor = *cursorp) == NULL)
+		return (0);
+	*cursorp = NULL;
+
+	/*
+	 * If using the session's cached metadata cursor, clear the in-use flag
+	 * and reset it, otherwise, discard the cursor.
+	 */
+	if (F_ISSET(cursor, WT_CURSTD_META_INUSE)) {
```
agorrod: I'd feel more comfortable comparing the cursor pointers than the flag here, or at least asserting that they are the same if we expect them to be. i.e: This code assumes that acquisition/release are entirely nested.

src/meta/meta_table.c
```diff
@@ -160,12 +194,11 @@ __wt_metadata_update(
 	if (WT_META_TRACKING(session))
 		WT_RET(__wt_meta_track_update(session, key));
 
-	WT_RET(__wt_metadata_cursor(session, "overwrite", &cursor));
+	WT_RET(__wt_metadata_cursor(session, &cursor));
```
agorrod: I'd rather keep this as an explicit overwrite setting, I guess there is an overhead to doing that, but relying on the default not changing makes me nervous.

src/conn/conn_dhandle.c
agorrod: This question makes me wonder.. Where do we close the new cursor when closing a session handle? I don't see it anywhere.

keithbostic: > Where do we close the new cursor when closing a session handle?

The metadata cursor is on the cursor list, so we close it in `__session_close()`:
```
/* Close all open cursors. */
while ((cursor = TAILQ_FIRST(&session->cursors)) != NULL) {
        /*
         * Notify the user that we are closing the cursor handle
         * via the registered close callback.
         */
        if (session->event_handler->handle_close != NULL &&
            !WT_STREQ(cursor->internal_uri, WT_LAS_URI))
                WT_TRET(session->event_handler->handle_close(
                    session->event_handler, wt_session, cursor));
        WT_TRET(cursor->close(cursor));
}
```
Or is that not what you're asking?

src/cache/cache_las.c
keithbostic: @agorrod:
> This change appears a few times, is it a consequence of no longer saving and restoring the session_dhandle in __meta_btree_apply? I'd rather keep the S2BT semantic if possible.

The locations where that happens are when opening cursors on the LAS file, the metadata file and the LSM primary chunk. I'm not sure about the LSM primary chunk, but the other two are deliberately not setting the WT_SESSION handle (although I didn't trace back why that's the case).

Anyway, to make S2BT work, we'd have to save/restore the WT_SESSION handle around the S2BT macro call, which seems like more work than it's worth to me.

src/meta/meta_table.c
keithbostic: @agorrod:
> I'd feel more comfortable comparing the cursor pointers than the flag here, or at least asserting that they are the same if we expect them to be.

We have to clear the flag here, which is why I tested it -- I'm happy to assert they're the same, though.

src/meta/meta_table.c
keithbostic: @agorrod:
> I'd rather keep this as an explicit overwrite setting, I guess there is an overhead to doing that, but relying on the default not changing makes me nervous.

I disagree strongly on this one, parsing configuration strings is pretty slow, we've had problems making that work fast.

agorrod:
> I don't know why it would be pinning resources here. Should I convert this code to using the cached cursor?

It was a very deliberate commit: https://github.com/wiredtiger/wiredtiger/commit/ed8b6bd69e8609215cdc36b8869a5464c0519313

I can't find the motivation for this change, but I'm inclined to leave it as is.

keithbostic:
@agorrod, @michaelcahill:

I think there are three remaining questions on this branch:

First, I left `__wt_txn_recover` opening a new metadata cursor rather than using the cached one because there's a comment which I don't understand.
```
/*
 * We no longer need the metadata cursor: close it to avoid pinning any
 * resources that could block eviction during recovery.
 */
r.files[0].c = NULL;
WT_ERR(metac->close(metac));
```
It doesn't matter much, but it's the last place opening/closing its own metadata cursor open for no reason I can see. Removing that code doesn't seem to break anything, but I could be wrong.

Second, Michael said: I'd prefer a macro called WT_SESSION_META_DHANDLE that takes a session and looks up the data handle session->meta_cursor to find the data handle for this one case.

I made and pushed that change, Michael should probably sign off on it.

Third, Michael said: On second thought, does it make sense to move [the close of the open metadata handle] into `__wt_session_close_cache`?

I responded in the comment chain that I don't see a path to that change, Michael should probably agree before we merge.

agorrod:
@keithbostic I've pushed a change with some minor nits. Once it's passed testing I'm OK with merging this branch.

PR #2259, opened 2015-10-22
SERVER-21027 Reverse split if there are many deleted pages (3.0)

michaelcahill:
This code was developed on the `mongodb-3.0` branch, and will be cherry-picked to create a set of changes against `develop`.


agorrod:
This has been replaced by #2330

PR #2260, opened 2015-10-22
SERVER-21027 Reverse split if there are many deleted pages

michaelcahill:
This is the version of the reverse split changes for develop.


src/evict/evict_page.c
```diff
@@ -142,6 +142,29 @@ done:	if (((inmem_split && ret == 0) || (forced_eviction && ret == EBUSY)) &&
...
+		return (0);
+
+	parent = ref->home;
+	WT_INTL_INDEX_GET(session, parent, pindex);
+	deleted_entries = __wt_atomic_addv32(&pindex->deleted_entries, 1);
+	if (deleted_entries > pindex->entries / 10)
+		WT_RET(__wt_split_reverse(session, parent->pg_intl_parent_ref));
+
+	return (0);
+}
```
keithbostic: @michaelcahill, I'd leave the set of `WT_REF_DELETED` in the callers and move the `WT_WITH_PAGE_INDEX` work into `__evict_ref_delete()`. That keeps the work to set the fields of the `WT_REF` in one place, rather than splitting it into two parts, and puts the page-index work in a single place as well.

keithbostic:
@michaelcahill, I fixed a couple of comments, and there's one code shuffled I'd recommend (but don't feel terribly strongly about). In short, this looks good to me.

src/evict/evict_page.c
michaelcahill: I've moved setting `WT_REF_DELETED` out (and renamed the function) -- I don't think it makes much sense to move `WT_WITH_PAGE_INDEX` out, or we'd need another function, where the only purpose of the first is to wrap the second in `WT_WITH_PAGE_INDEX`.

agorrod:
Thanks @michaelcahill lgtm. I'll merge and get this into stress testing.

@keithbostic I'm not precluding further discussion around changes to the control flow around ``__evict_ref_delete``.

PR #2261, opened 2015-10-22
WT-2179 Added decorator to mark txn13 as part of the --long test suite.

ddanderson:


ddanderson:
@agorrod, please review.  I think this is all that's needed for moving tests into --long.

agorrod:
Thanks @ddanderson lgtm. I'm glad it wasn't more complex.

PR #2262, opened 2015-10-22
WT-2164 Prevent another chunk checkpoint while the first is still in progress.

sueloverso:
@agorrod Please review this change.  It avoids the bus error by not allowing two LSM workers to checkpoint the same chunk.


sueloverso:
The test has run over 200 iterations over all the trees I have with the fix running.  That is more than it ever had to reproduce the bus error.  I'll let it keep running until we chat but it is looking solid.

PR #2264, opened 2015-10-23
WT-2164 Have LSM enable metadata tracking prior to checkpointing a chunk.

agorrod:
Otherwise insufficient locks are held to keep checkpoint handles alive.

Also add assertions into checkpoint to avoid running into this again
in the future.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -59,6 +59,7 @@ CSV
 CURSORs
 CURSTD
 CallsCustDate
+Checkpointing
 Checksum
 Checksums
 CityHash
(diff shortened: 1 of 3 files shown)


agorrod:
@sueloverso Could you take a look at this change please? It is the change to add metadata tracking when LSM flushes a chunk.

I haven't been able to reproduce the failure, so it would be great if you could run the reproducer against this branch or tell me how you are reproducing the issue.

I think both this and the change in WT-2262 are good changes. I kept them separate because I would like to know whether this change resolves the sigbus independent of the LSM exclusive flushing change.

sueloverso:
I have run this over 600 iterations without the bus error also.  So both of these fix it.  I agree both are good changes as there is no reason for an LSM worker to sit waiting for a lock when another is already doing that work.  I will merge this and then fix up the other branch from the conflict and merge that too.

PR #2268, opened 2015-10-27
WT-2146 Do quick searches based on search key size

michaelcahill:
This makes the branch more generally applicable, rather than being a special case for one type of key.  It was just as fast in my testing.

I tried several variations on the core of #2223 but didn't find any way to make it faster, and my attempts to simplify it made it slower.

@keithbostic, if you're happy with this change, can you please merge it, then merge #2223?


keithbostic:
@michaelcahill, lgtm, merged.

PR #2269, opened 2015-10-28
WT-2170 Add a macro perform an operation at an isolation level.

michaelcahill:
Lock down visibility checks so we never try to use a snapshot without one being allocated.

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -443,6 +443,7 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
 	} else if ((ret = __wt_btcur_next(cbt, false)) != WT_NOTFOUND)
 		exact = 1;
 	else {
+		WT_ERR(__cursor_func_init(cbt, true));
 		WT_ERR(btree->type == BTREE_ROW ?
 		    __cursor_row_search(session, cbt, NULL, true) :
 		    __cursor_col_search(session, cbt, NULL));
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -152,6 +152,8 @@ __sync_file(WT_SESSION_IMPL *session, int syncop)
 				leaf_bytes += page->memory_footprint;
 				++leaf_pages;
 			}
+			if (txn->isolation == WT_ISO_READ_COMMITTED)
+				__wt_txn_get_snapshot(session);
 			WT_ERR(__wt_reconcile(session, walk, NULL, 0));
 		}
 		break;
--- a/src/evict/evict_file.c
+++ b/src/evict/evict_file.c
@@ -18,7 +18,10 @@ __wt_evict_file(WT_SESSION_IMPL *session, int syncop)
 	WT_DECL_RET;
 	WT_PAGE *page;
 	WT_REF *next_ref, *ref;
+	WT_TXN *txn;
 	bool evict_reset;
+
+	txn = &session->txn;
 
 	/*
 	 * We need exclusive access to the file -- disable ordinary eviction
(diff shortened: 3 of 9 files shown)


michaelcahill:
@sueloverso, this is significantly more complex than I had hoped for, but I believe it also fixes some real bugs in the process of making a general-purpose `WT_WITH_TXN_ISOLATION` macro...

Can you please take a look?

src/btree/bt_cursor.c
```diff
@@ -443,6 +443,7 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
 	} else if ((ret = __wt_btcur_next(cbt, false)) != WT_NOTFOUND)
 		exact = 1;
 	else {
+		WT_ERR(__cursor_func_init(cbt, true));
```
michaelcahill: The cursor is not positioned at this point, so we may have released our snapshot.  We need a new one to restart the search.

src/btree/bt_sync.c
```diff
@@ -152,6 +152,8 @@ __sync_file(WT_SESSION_IMPL *session, int syncop)
 				leaf_bytes += page->memory_footprint;
 				++leaf_pages;
 			}
+			if (txn->isolation == WT_ISO_READ_COMMITTED)
+				__wt_txn_get_snapshot(session);
```
michaelcahill: We need this to make the "read-committed" checkpoint of the metadata work.  Otherwise there are paths into `__wt_meta_track_off` without a snapshot that try to do visibility checks.

src/evict/evict_file.c
```diff
@@ -56,8 +59,11 @@ __wt_evict_file(WT_SESSION_IMPL *session, int syncop)
 		 * and the write will fail with EBUSY.  Our caller handles that
 		 * error, retrying later.
 		 */
-		if (syncop == WT_SYNC_CLOSE && __wt_page_is_modified(page))
+		if (syncop == WT_SYNC_CLOSE && __wt_page_is_modified(page)) {
+			if (txn->isolation == WT_ISO_READ_COMMITTED)
+				__wt_txn_get_snapshot(session);
```
michaelcahill: Again, we can get here without a snapshot (e.g., when dropping a dirty file).

src/include/txn.h
```diff
@@ -153,13 +181,13 @@ struct __wt_txn {
 
 	/* Checkpoint status. */
 	WT_LSN		ckpt_lsn;
-	bool		full_ckpt;
 	uint32_t	ckpt_nsnapshot;
 	WT_ITEM		*ckpt_snapshot;
+	bool		full_ckpt;
```
michaelcahill: This is irrelevant to the rest of the change: I started reordering fields to save space.  If we moved `ckpt_nsnapshot` after `ckpt_snapshot`, I think it would save 8 bytes per session...

src/include/txn.i
```diff
@@ -186,11 +186,14 @@ __wt_txn_visible(WT_SESSION_IMPL *session, uint64_t id)
...
 	/*
-	 * Read-uncommitted transactions see all other changes.
+	 * If we don't have a snapshot or we are running at read-uncommitted
+	 * isolation, all other changes are visible.
 	 */
 	if (txn->isolation == WT_ISO_READ_UNCOMMITTED)
 		return (true);
 
+	WT_ASSERT(session, F_ISSET(txn, WT_TXN_HAS_SNAPSHOT));
+
```
michaelcahill: Any visibility check that  isn't read-uncommitted should have an active snapshot.  This assertion found many paths that could give unexpected results (either when checkpointing the metadata or when evicting dirty files, which used to use the special `WT_ISO_EVICTION` isolation level.

src/meta/meta_track.c
```diff
@@ -275,7 +275,8 @@ __wt_meta_track_off(WT_SESSION_IMPL *session, bool need_sync, bool unroll)
 		WT_RET(ret);
 	} else {
 		WT_WITH_DHANDLE(session, session->meta_dhandle,
-		    ret = __wt_checkpoint(session, NULL));
+		    WT_WITH_TXN_ISOLATION(session, WT_ISO_READ_COMMITTED,
+			ret = __wt_checkpoint(session, NULL)));
```
michaelcahill: Without this, flushing the metadata gives unpredictable results depending on the configuration of the session.

src/include/txn.i
sueloverso: It would make more sense to me to restore the original comment and add another one at the assert that says this.

PR #2270, opened 2015-10-29
WT-2185: Don't do reverse splits when closing a file.

agorrod:
The ovfl_lock may already have been freed.


michaelcahill:
@agorrod, thanks for fixing this -- I agree that there's no point trying reverse splits when evicting a whole file.  I'll merge.

PR #2271, opened 2015-10-29
SERVER-21027 Fix reverse splits to keep the original child ref locked

michaelcahill:
Fix reverse splits to keep the original child ref locked until we have a hazard pointer on the parent page, as we do for regular splits.

Otherwise, a reverse split could race with eviction of the internal page.


agorrod:

PR #2272, opened 2015-10-29
WT-2182: when internal pages grow large enough, split them into their parents

keithbostic:


keithbostic:
@agorrod, @michaelcahill: with 2d937df and 71d3672, this branch appears to run for me (both with the problematic CONFIG and a more general smoke/stress test). I'm going to leave things running for awhile, but we should talk about next steps today, assuming it runs in Alex's tests and nothing new shows up.

agorrod:
Stress testing uncovered a problem:

```
t: read_row: row 34505: found in Berkeley DB, not found in WiredTiger
#  RUN PARAMETERS
abort=0
auto_throttle=1
firstfit=0
bitcnt=8
bloom=1
bloom_bit_count=26
bloom_hash_count=24
bloom_oldest=0
cache=1
checkpoints=1
checksum=uncompressed
chunk_size=4
compaction=0
compression=lz4
data_extend=0
data_source=table
delete_pct=0
dictionary=0
encryption=rotn-7
evict_max=2
file_type=row-store
backups=0
huffman_key=0
huffman_value=0
insert_pct=12
internal_key_truncation=1
internal_page_max=13
isolation=snapshot
key_gap=1
key_max=93
key_min=16
leak_memory=0
leaf_page_max=14
logging=0
logging_compression=zlib
logging_archive=1
logging_prealloc=1
long_running_txn=0
lsm_worker_threads=3
merge_max=18
mmap=1
ops=100000
prefix_compression=0
prefix_compression_min=0
repeat_data_pct=27
reverse=0
rows=100000
runs=100
salvage=1
split_pct=75
statistics=0
statistics_server=0
threads=1
timer=20
value_max=3018
value_min=2
verify=1
wiredtiger_config=
write_pct=11
```

It didn't reproduce for me after ~15 runs.

keithbostic:
@agorrod, I saw this failure twice over-night, so it's real.

agorrod:
@keithbostic I've been staring at this today, and don't have a definite answer. I have a few observations though. Firstly the configuration isn't doing very much splitting at all, I've tweaked the config so that it splits more often:

```
#  RUN PARAMETERS
abort=0
auto_throttle=1
firstfit=0
bitcnt=8
bloom=1
bloom_bit_count=26
bloom_hash_count=24
bloom_oldest=0
cache=1
checkpoints=1
checksum=uncompressed
chunk_size=4
compaction=0
compression=none
data_extend=0
data_source=table
delete_pct=0
dictionary=0
encryption=rotn-7
evict_max=2
file_type=row-store
backups=0
huffman_key=0
huffman_value=0
insert_pct=12
internal_key_truncation=1
internal_page_max=13
isolation=snapshot
key_gap=1
key_max=93
key_min=16
leak_memory=0
leaf_page_max=12
logging=0
logging_compression=none
logging_archive=1
logging_prealloc=1
long_running_txn=0
lsm_worker_threads=3
merge_max=18
mmap=1
ops=30000
prefix_compression=0
prefix_compression_min=0
repeat_data_pct=27
reverse=0
rows=100000
runs=100
salvage=1
split_pct=75
statistics=0
statistics_server=0
threads=1
timer=100
value_max=3018
value_min=2
verify=1
wiredtiger_config=
write_pct=11
```

Specifically a smaller leaf page size. Splits also seem to happen early in the run, so I've dialled down the number of operations. While I was doing this I disabled the verify and salvage parts of test/format, and decreased the split_deepen_min_child to make splits more common, and turned the die into an assert so I get a core dump (not super useful - since it's too late by then).

```
--- a/test/format/ops.c
+++ b/test/format/ops.c
@@ -1253,6 +1253,7 @@ notfound_chk(const char *f, int wt_ret, int bdb_notfound, uint64_t keyno)
                        fprintf(stderr, " row %" PRIu64 ":", keyno);
                fprintf(stderr,
                    " found in Berkeley DB, not found in WiredTiger\n");
+               __wt_abort(NULL);
                die(0, NULL);
        }
        return (0);
diff --git a/test/format/t.c b/test/format/t.c
index 603706e..ced5bc0 100644
--- a/test/format/t.c
+++ b/test/format/t.c
@@ -227,6 +227,7 @@ main(int argc, char *argv[])
                                        wts_stats();

                                                        /* Verify */
+                               if (0)
                                wts_verify("post-ops verify");
                        }

@@ -241,12 +242,13 @@ main(int argc, char *argv[])
                 * If single-threaded, we can dump and compare the WiredTiger
                 * and Berkeley DB data sets.
                 */
-               if (SINGLETHREADED)
+               if (0 && SINGLETHREADED)
                        wts_dump("standard", 1);

                /*
                 * Salvage testing.
                 */
+               if (0)
                wts_salvage();

                /* Overwrite the progress line with a completion line. */
diff --git a/test/format/wts.c b/test/format/wts.c
index 23823c2..8b60fc8 100644
--- a/test/format/wts.c
+++ b/test/format/wts.c
@@ -304,7 +304,8 @@ wts_create(void)
        p += snprintf(p, REMAIN(p, end),
            "key_format=%s,"
            "allocation_size=512,%s"
-           "internal_page_max=%d,leaf_page_max=%d",
+           "internal_page_max=%d,leaf_page_max=%d"
+           ",split_deepen_min_child=300,",
            (g.type == ROW) ? "u" : "r",
            g.c_firstfit ? "block_allocation=first," : "",
            maxintlpage, maxleafpage);
```

I've also added a bit of debugging code to ``bt_split.c`` and ``row_srch.c``, some of the more useful changes:

```
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -240,6 +240,7 @@ __cursor_row_search(

        WT_WITH_PAGE_INDEX(session,
            ret = __wt_row_search(session, &cbt->iface.key, leaf, cbt, insert));
+       WT_ASSERT(session, ret != WT_RESTART);
        return (ret);
 }

diff --git a/src/btree/bt_split.c b/src/btree/bt_split.c
index a160a4c..25fc6da 100644
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -385,6 +385,7 @@ __split_root(WT_SESSION_IMPL *session, WT_PAGE *root)
        /* The root page will be marked dirty, make sure that will succeed. */
        WT_RET(__wt_page_modify_init(session, root));

+       fprintf(stderr, "Splitting the root\n");
        /*
         * Our caller is holding the root page locked to single-thread splits,
         * which means we can safely look at the page's index without setting a
@@ -1269,7 +1270,9 @@ __split_internal_unlock(WT_SESSION_IMPL *session, WT_PAGE *parent, bool hazard)
        if (hazard)
                ret = __wt_hazard_clear(session, parent);

+       __wt_sleep(0, 10000);
        WT_TRET(__wt_fair_unlock(session, &parent->page_lock));
+       __wt_sleep(0, 10000);
        return (ret);
 }

@@ -1315,9 +1318,11 @@ __split_internal_should_split(WT_SESSION_IMPL *session, WT_REF *ref)
         * splitting into parent pages can become large enough to result
         * in slow operations.
         */
-       if (!__wt_ref_is_root(ref) &&
-           pindex->entries > btree->split_deepen_min_child)
+       if (/*!__wt_ref_is_root(ref) &&*/
+           pindex->entries > btree->split_deepen_min_child) {
+               fprintf(stderr, "Splitting\n");
                return (true);
+       }

        return (false);
 }
```

With all of the above I can reproduce the failure about once every 5 minutes, running 10 test/formats in parallel:

```
for i in {1..10}; do ./t -c c2181 -h RUNDIR.$i & done
```

Interestingly the test we put in for a race condition with updated internal page indexes was looking for append workloads, and this is doing a point search. I wonder if we need to look for that race in another code path as well.

michaelcahill:
Thanks @agorrod, I've been running a similar config.  One change I made was to `wts_load`, to only bulk load every 10th row -- my thinking was that leaving space in between would make splits more likely.

I have also started seeing this failure mode:

```
nextprev: prev key mismatch:
	bdb-key {0000068655.00/opqr}
	 wt-key {0000078501.07/opqr}
```

Often the keys are close, but occasionally they differ by a lot (as above).  I had hoped this was a race between search and split, but this suggests there might be more going on...

michaelcahill:
Here's a faster reproducer with `test/checkpointer`.  Apply this diff to induce lots of splitting: 

```diff
diff --git a/test/checkpoint/workers.c b/test/checkpoint/workers.c
--- a/test/checkpoint/workers.c
+++ b/test/checkpoint/workers.c
@@ -39,12 +39,15 @@ static int
 create_table(WT_SESSION *session, COOKIE *cookie)
 {
 	int ret;
-	char *p, *end, config[128];
+	char *p, *end, config[1024];

 	p = config;
 	end = config + sizeof(config);
 	p += snprintf(p, (size_t)(end - p),
 	    "key_format=%s,value_format=S", cookie->type == COL ? "r" : "q");
+	if (cookie->type == ROW)
+		p += snprintf(p, (size_t)(end - p),
+		    ",allocation_size=1KB,leaf_page_max=1KB,internal_page_max=1KB,memory_page_max=32KB");
 	if (cookie->type == LSM)
 		(void)snprintf(p, (size_t)(end - p), ",type=lsm");

@@ -129,10 +132,10 @@ static inline int
 worker_op(WT_CURSOR *cursor, uint64_t keyno, u_int new_val)
 {
 	int ret;
-	char valuebuf[64];
+	char valuebuf[1024];

 	cursor->set_key(cursor, keyno);
-	(void)snprintf(valuebuf, sizeof(valuebuf), "%037u", new_val);
+	(void)snprintf(valuebuf, sizeof(valuebuf), "%037u%0200u", new_val, new_val);
 	cursor->set_value(cursor, valuebuf);
 	if ((ret = cursor->insert(cursor)) != 0) {
 		if (ret == WT_ROLLBACK)
```

Then run:

```sh
./t -t r -n 1000000 -k 5000000 -C cache_size=10MB
```

In seconds, you'll see something like:

```
Key/value mismatch: 2742091/000000000000000000000000000000001078300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010783 from a ROW table is not 2395580/000000000000000000000000000000000395100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003951 from a ROW table
t: 1st cursor didn't find 1st key: WT_NOTFOUND: item not found
t: 1st cursor didn't find 1st key: WT_NOTFOUND: item not found
t: verify_checkpoint - mismatching data: Bad address
Ran workers for: 0.485841 seconds
```

keithbostic:
@agorrod, @michaelcahill, two places I think we do not need to test for races are:

* at the bottom of `__wt_tree_walk`, where we're descending the tree during a walk,
* when we have an exact match in `__wt_row_search` (that is, `(cmp == 0)`), we use the last index slot on the page with an exactly matching key, rather than testing `(srch_key.data > item)`.

If anybody disagrees, let's talk.

Another possibility: we're currently doing a full tree search in the case of hitting the race, but I don't think that's absolutely necessary, I suspect we could restart the search at the parent page. The only thing that made me nervous about it is the fact that the parent page could itself have split up the tree when we restart, and thinking about that race makes my head hurt. I don't plan to make this change, but it's worth remembering if we see lots of restarts in some workload.

agorrod:
For the record - my test that was failing reliably yesterday has passed 1000 iterations of test/format without issue now.

agorrod:
@keithbostic This branch is looking good against develop now, but I'm consistently seeing a bug when running a wtperf workload. The workload is:

```
conn_config="cache_size=200MB"
table_config="type=file,allocation_size=1KB,leaf_page_max=1KB,internal_page_max=1KB,memory_page_max=32KB"
icount=1000000
random_range=50000000
report_interval=5
run_time=240
populate_threads=1
threads=((count=4,inserts=1),(count=8,reads=1))
```

The symptom is:

```
1292067 reads, 504852 inserts, 0 updates, 0 truncates, 0 checkpoints in 5 secs (190 total secs)
1268917 reads, 498003 inserts, 0 updates, 0 truncates, 0 checkpoints in 5 secs (195 total secs)
[1447299967:379453][8261:0x7f46857fa700], file:test.wt, WT_CURSOR.search: the 1 and 3 keys on page at [write-check] are incorrectly sorted
read failed for: 0000000000025037461, range: 51000000 Error: WT_ERROR: non-specific WiredTiger error
Worker thread(s) exited without finishing. Error: WT_ERROR: non-specific WiredTiger error
[1447299971:40839][8261:0x7f4698199700], file:test.wt, lookaside table: the 1 and 3 keys on page at [write-check] are incorrectly sorted
[1447299971:40894][8261:0x7f4698199700], file:test.wt, lookaside table: Final close of file:test.wt failed: WT_ERROR: non-specific WiredTiger error
[1447299971:49291][8261:0x7f4698199700], metadata-ckpt: Connection has open file handles: test.wt
[1447299971:49359][8261:0x7f4698199700], metadata-ckpt: cache server: exiting with 13724818 pages in memory and 13724748 pages evicted
[1447299971:49379][8261:0x7f4698199700], metadata-ckpt: cache server: exiting with 1040260 bytes in memory
[1447299971:49393][8261:0x7f4698199700], metadata-ckpt: cache server: exiting with 986340 bytes dirty and 49 pages dirty
Error closing connection to WT_TEST Error: WT_ERROR: non-specific WiredTiger error
```

I caught a failure in a debugger, and it shows the following call stack:

```
Thread 1 (Thread 0x7fe2e0736700 (LWP 10785)):
#0  0x00007fe2e076ba98 in raise () from /lib64/libc.so.6
#1  0x00007fe2e076d72a in abort () from /lib64/libc.so.6
#2  0x00000000004c254d in __wt_abort (session=session@entry=0x243b460) at ../src/os_posix/os_abort.c:25
#3  0x000000000048d2cc in __wt_verify_dsk_image (session=session@entry=0x243b460,
    tag=tag@entry=0x514710 "[write-check]", dsk=0x7fe2a0000b50, size=<optimized out>,
    empty_page_ok=empty_page_ok@entry=false) at ../src/btree/bt_vrfy_dsk.c:66
#4  0x000000000048e44f in __wt_verify_dsk (session=session@entry=0x243b460,
    tag=tag@entry=0x514710 "[write-check]", buf=<optimized out>) at ../src/btree/bt_vrfy_dsk.c:183
#5  0x0000000000477457 in __wt_bt_write (session=session@entry=0x243b460, buf=buf@entry=0x7fe2a00008d8,
    addr=addr@entry=0x7fe2e0735a10 "\001\001\364e\201\344k\305`\261",
    addr_sizep=addr_sizep@entry=0x7fe2e0735990, checkpoint=checkpoint@entry=false, compressed=false)
    at ../src/btree/bt_io.c:223
#6  0x000000000043f79f in __rec_split_write (session=session@entry=0x243b460, r=r@entry=0x7fe2a00008c0,
    bnd=bnd@entry=0x7fe2a00221e0, buf=buf@entry=0x7fe2a00008d8, last_block=last_block@entry=false)
    at ../src/reconcile/rec_write.c:3299
#7  0x00000000004408c8 in __rec_split (session=session@entry=0x243b460, r=r@entry=0x7fe2a00008c0,
    next_len=<optimized out>) at ../src/reconcile/rec_write.c:2426
#8  0x0000000000446f6d in __rec_row_leaf_insert (session=session@entry=0x243b460,
    r=r@entry=0x7fe2a00008c0, ins=0x7fe2d9e25d30) at ../src/reconcile/rec_write.c:5251
#9  0x0000000000449d39 in __rec_row_leaf (salvage=0x0, page=0x7fe2b4378330, r=0x7fe2a00008c0,
    session=0x243b460) at ../src/reconcile/rec_write.c:5186
#10 __wt_reconcile (session=session@entry=0x243b460, ref=ref@entry=0x7fe2cc337210,
    salvage=salvage@entry=0x0, flags=1) at ../src/reconcile/rec_write.c:416
#11 0x000000000042291a in __evict_review (closing=false, inmem_splitp=<synthetic pointer>,
    ref=0x7fe2cc337210, session=0x243b460) at ../src/evict/evict_page.c:477
#12 __wt_evict (session=session@entry=0x243b460, ref=0x7fe2cc337210, closing=closing@entry=false)
    at ../src/evict/evict_page.c:79
#13 0x0000000000421720 in __evict_page (is_server=true, session=0x243b460) at ../src/evict/evict_lru.c:1467
#14 __evict_lru_pages (is_server=<optimized out>, session=<optimized out>) at ../src/evict/evict_lru.c:818
#15 __evict_server_work (session=0x243b460) at ../src/evict/evict_lru.c:940
#16 __evict_pass (session=0x243b460) at ../src/evict/evict_lru.c:590
#17 __evict_server (arg=0x243b460) at ../src/evict/evict_lru.c:172
```

There is one other thread that may be interesting:

```
Thread 10 (Thread 0x7fe2def33700 (LWP 10788)):
#0  0x00007fe2e081cf97 in sched_yield () from /lib64/libc.so.6
#1  0x0000000000437245 in __wt_yield () at ../src/os_posix/os_yield.c:18
#2  0x0000000000495cac in __wt_row_ikey (session=session@entry=0x243bd60, cell_offset=cell_offset@entry=0,
    key=key@entry=0x7fe2d01fa268, size=16, ref=ref@entry=0x7fe2d185d140) at ../src/btree/row_key.c:520
#3  0x0000000000483eba in __wt_multi_to_ref (session=session@entry=0x243bd60, page=0x7fe2d829ecc0,
    multi=<optimized out>, refp=<optimized out>, incrp=incrp@entry=0x7fe2def328c8)
    at ../src/btree/bt_split.c:1594
#4  0x0000000000484d46 in __split_multi (closing=false, ref=0x7fe2d805d3c0, session=0x243bd60)
    at ../src/btree/bt_split.c:1937
#5  __wt_split_multi (session=session@entry=0x243bd60, ref=ref@entry=0x7fe2d805d3c0,
    closing=closing@entry=0) at ../src/btree/bt_split.c:1984
#6  0x00000000004231f8 in __evict_page_dirty_update (closing=false, ref=0x7fe2d805d3c0, session=0x243bd60)
    at ../src/evict/evict_page.c:290
#7  __wt_evict (session=session@entry=0x243bd60, ref=ref@entry=0x7fe2d805d3c0, closing=closing@entry=false)
    at ../src/evict/evict_page.c:123
#8  0x000000000047b177 in __wt_page_release_evict (ref=0x7fe2d805d3c0, session=0x243bd60)
    at ../src/include/btree.i:1164
#9  __wt_page_in_func (session=session@entry=0x243bd60, ref=ref@entry=0x7fe2d805d3c0, flags=flags@entry=0,
    file=file@entry=0x516000 "../src/btree/row_srch.c", line=line@entry=330) at ../src/btree/bt_read.c:536
#10 0x0000000000498d3d in __wt_page_swap_func (flags=0, file=0x516000 "../src/btree/row_srch.c", line=330,
    want=0x7fe2d805d3c0, held=0x7fe2c827b5d0, session=0x243bd60) at ../src/include/btree.i:1260
#11 __wt_row_search (session=session@entry=0x243bd60, srch_key=srch_key@entry=0x7fe2d000ace8,
    leaf=leaf@entry=0x0, cbt=cbt@entry=0x7fe2d000ac10, insert=insert@entry=true)
    at ../src/btree/row_srch.c:330
#12 0x00000000004e4457 in __cursor_row_search (leaf=0x0, insert=true, cbt=0x7fe2d000ac10,
    session=0x243bd60) at ../src/btree/bt_cursor.c:241
#13 __wt_btcur_insert (cbt=cbt@entry=0x7fe2d000ac10) at ../src/btree/bt_cursor.c:534
#14 0x00000000004a793f in __curfile_insert (cursor=0x7fe2d000ac10) at ../src/cursor/cur_file.c:245
#15 0x000000000040b1f4 in worker (arg=0x24c0730) at ../../../bench/wtperf/wtperf.c:552
#16 0x00007fe2e0afe555 in start_thread () from /lib64/libpthread.so.0
#17 0x00007fe2e0838f3d in clone () from /lib64/libc.so.6
```
You need a diagnostic build to detect this problem - though I came across it via verify - so if you don't have a diagnostic build, this bug makes it to disk. I'll keep digging and see what I can find this afternoon.

keithbostic:
@agorrod: I finally saw this one (I almost gave up, I didn't set it for around a 200 runs).

I don't have much information yet, but this slightly different CONFIG (no readers, smaller icount):
```
conn_config="cache_size=200MB"
table_config="type=file,allocation_size=1KB,leaf_page_max=1KB,internal_page_max=
1KB,memory_page_max=32KB"
icount=10000
random_range=50000000
report_interval=5
run_time=240
populate_threads=1
threads=((count=8,inserts=1))
```
plus this diff:
```
diff --git a/src/btree/bt_split.c b/src/btree/bt_split.c
index 8614903..cd6176a 100644
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -1016,6 +1016,7 @@ __split_internal(WT_SESSION_IMPL *session, WT_PAGE *parent, WT_PAGE *page)
         * this point unless there's a panic.
         */
        complete = true;
+       __wt_sleep(1, 0);
 
        /* Confirm the page's index hasn't moved, then update it. */
        WT_ASSERT(session, WT_INTL_INDEX_GET_SAFE(page) == pindex);
```
makes it fire relatively quickly for me.

Once again, there's an out-of-order key/value pair on a row-store leaf page.

agorrod:
@keithbostic @michaelcahill and I have figured out the cause of the failure. We are still pondering what the solution is.

The issue is related to ``__wt_lex_compare_skip``, the trouble is that we can hold on to an invalid ``skiphigh`` when descending the tree. The reason we see that now and didn't before is because the key range on a page can now be different to what the parent page thinks it is. The scenario is:

A ``__split_parent_climb`` comes through and splits a page that has key range A-F, creates a bunch of new internal pages (A-B, C-D, E-F), and replaces the index in the parent.

A search insert comes through the new index in the parent page (so the key range for the is A-B) into the retained page (with a key range of A-F). Since skiphigh is set to the range between A-B, and our page has A-F, the value is invalid.

To test out the theory, the following change causes the test case to pass reliably for me:

```
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -329,10 +333,12 @@ descend:  /*
                 */
                if ((ret = __wt_page_swap(session, current, descent, 0)) == 0) {
                        current = descent;
+                       skiphigh = 0;
                        continue;
                }
```

In order to detect this sort of problem in the future I think it's worthwhile adding an assertion to ``__wt_lex_compare_skip`` like:

```
WT_ASSERT(session, memcmp(srch_key->data, item->data, *matchp) == 0);
```

michaelcahill:
Thanks for the writeup @agorrod.  I'd phrase the situation slightly differently: we have a solution, and with that change applied, we no longer see out-of-order keys.  We don't yet know the performance impact of comparing more key prefixes after switching pages.

I think a slight refinement of the change is to set `skiphigh = 0` around line 204 (after checking `(page->type != WT_PAGE_ROW_INT)`) -- that keeps the skips going into the leaf page, which can't have a larger key range than we expect when we step into it.

keithbostic:
Nice catch!

I've pushed my version of this fix into the branch so it gets tested over the weekend, it's 50c9bc9. Can one of you please take a close look to make sure I didn't get it wrong? I've added some large comments, too.

Basically, I agree we only need to set `skiphigh` to 0, and we only need to do it when starting the search of a new internal page, when not looking at a short-key or using a collator, so we can tuck the reset into the second part of the if/else clause that wraps the binary search of the internal page.

keithbostic:
@agorrod, @michaelcahill: stress failure [9143](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress/9143/) is real, it's pretty easy to reproduce, too.
```
#  RUN PARAMETERS
abort=0
auto_throttle=1
firstfit=0
bitcnt=2
bloom=1
bloom_bit_count=36
bloom_hash_count=19
bloom_oldest=0
cache=7
checkpoints=1
checksum=uncompressed
chunk_size=8
compaction=0
compression=zlib
data_extend=0
data_source=table
delete_pct=27
dictionary=0
encryption=none
evict_max=5
file_type=row-store
backups=0
huffman_key=0
huffman_value=0
insert_pct=44
internal_key_truncation=1
internal_page_max=10
isolation=read-uncommitted
key_gap=5
key_max=111
key_min=19
leak_memory=0
leaf_page_max=9
logging=1
logging_compression=bzip
logging_archive=1
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=4
merge_max=7
mmap=0
ops=100000
prefix_compression=0
prefix_compression_min=5
repeat_data_pct=66
reverse=0
rows=100000
runs=100
salvage=1
split_pct=43
statistics=0
statistics_server=0
threads=7
timer=20
value_max=2155
value_min=15
verify=1
wiredtiger_config=
write_pct=13
```

agorrod:
For the record I've been running the test/format configurations that caused problems over the weekend and haven't reproduced any failures. Barring any new failures tonight, I think we are ready to enter final review phase on this branch.

src/btree/row_srch.c
```diff
@@ -298,25 +319,41 @@ restart:	page = current->page;
...
-			break;
-		case WT_RESTART:
+			continue;
+		}
+		if (ret == WT_RESTART) {
 			skiphigh = skiplow = 0;
-			goto restart;
-		default:
-			return (ret);
+			goto restart_page;
```
agorrod: @keithbostic I'm not convinced it's good enough to restart on the same page here. Could you convince me? The particular scenario I'm cautious about is if the ``__wt_page_swap`` call triggers a forced eviction, which in turn triggers a split deepen (or a cascade of them all the way up to the root, and then a deepen).

src/btree/col_srch.c
```diff
@@ -90,15 +103,13 @@ restart:	page = current->page;
...
+		if ((ret = __wt_page_swap(session, current, descent, 0)) == 0) {
 			current = descent;
-			break;
-		case WT_RESTART:
-			goto restart;
-		default:
-			return (ret);
+			continue;
 		}
+		if (ret == WT_RESTART)
```
agorrod: We discussed the intricacies of the row search loop on the phone today. I presume the same reasoning applies to the column store case, but it would be good if any comment updates are reflected here as well.

The swap/restart logic is complex enough that it would be nice if column and row store could share it. I know this is a performance path, so I'm not pushing hard for that.

agorrod:
@keithbostic @michaelcahill I'm happy with this branch (LGTM). There is more code duplication in ``bt_split.c`` than I think is ideal, but I don't see a neat way to refactor.

src/evict/evict_page.c
```diff
@@ -179,9 +179,17 @@ __evict_delete_ref(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 		 * something is busy, be sure that the page still ends up
 		 * marked deleted.
 		 */
-		if (ndeleted > pindex->entries / 10 && pindex->entries > 1 &&
-		    (ret = __wt_split_reverse(session, ref)) != EBUSY)
-			return (ret);
+		if (ndeleted > pindex->entries / 10 && pindex->entries > 1) {
+			if ((ret = __wt_split_reverse(session, ref)) == 0)
+				return (0);
```
michaelcahill: @keithbostic, I think this change has introduced a bug that will leave child refs locked.

Unless I am missing something, when we hit the `result_entries == 0` case in `__split_parent`, we will give up on the split but return 0.  The cleanup path in `__split_parent` used to flip the child ref to `WT_REF_DELETED` but this branch has removed that code, and I don't think the code here can know whether the ref was unlocked.

We could have that case return `EBUSY`, but then we would have to suppress the error message in that case because it is an expected condition.  One way to do that would be to pull the `__wt_err` call from `__split_parent` up into callers other than `__wt_split_reverse` (which I think just means `__split_insert` since the other caller has its own error message already.

michaelcahill:
Other than my comment on reverse splits, this lgtm.

P.S.  It's annoying that github won't display the `bt_split.c` changes.

keithbostic:
@agorrod, agreed on the code duplication (the reason this is branch is named "-2" is because I tried a couple of times to do bt_split.c as a set of merged split functions). It's something to keep in mind as we modify this code in the future.

src/evict/evict_page.c
keithbostic: @michaelcahill, thanks! I think you're correct.

Pulling up the error message means fixing `__split_multi` as well as `__split_reverse`, and I'm hesitant to do that, I'd prefer `__split_parent` own the problem of deciding if the split was successful or not. We could push setting `WT_REF_DELETED` back down into `__split_parent` again, but that's ugly, too.

I pushed a change to have that case return `EBUSY` (and, as you noted, suppressing the error message because it's an expected condition), that change is 846125e.

Two other things occurred to me:

* Could we set the WT_REF we're deleting to `WT_REF_DELETED` before calling `__split_parent`? If so, then we don't have to handle the `EBUSY` error return at all. I've pushed a branch (#2309) to show you what I mean.
* Could we continue the internal page reverse split, even if the result would be an empty page, by leaving one slot in the WT_REF_DELETED state?  I've pushed a branch (#2310) to show you what I mean.

michaelcahill:
Thanks, @keithbostic -- lgtm, I'll merge!

PR #2273, opened 2015-10-29
WT-2184 Fix log scan bug when final record has many trailing zeros

sueloverso:
@michaelcahill Please review these changes that implement the fixes we discussed last night.  I am in the process of writing a test program similar to the failing case to put in the tree but that is not complete.  I have a similar diff that fixes it in the 2.6.1 tree they're using also.  We can chat about it tonight.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1220,11 +1220,13 @@ __log_has_hole(WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t offset, bool *hole)
 	    remainder -= bufsz, off += (wt_off_t)bufsz) {
 		rdlen = WT_MIN(bufsz, remainder);
 		WT_ERR(__wt_read(session, fh, off, rdlen, buf));
-		if (memcmp(buf, zerobuf, rdlen) != 0)
+		if (memcmp(buf, zerobuf, rdlen) != 0) {
 			*hole = true;
+			break;
+		}
 	}
-err:
-	if (buf != NULL)
+
+err:	if (buf != NULL)
 		__wt_free(session, buf);
 	if (zerobuf != NULL)
 		__wt_free(session, zerobuf);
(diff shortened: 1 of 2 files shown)


src/log/log.c
```diff
@@ -1173,85 +1173,56 @@ __wt_log_close(WT_SESSION_IMPL *session)
...
-	    off -= (wt_off_t)bufsz) {
-		WT_ERR(__wt_read(session, fh, off, bufsz, buf));
-		if (memcmp(buf, zerobuf, bufsz) != 0)
-			break;
+	for (off = offset; remainder > 0;
+	    remainder -= bufsz, off += (wt_off_t)bufsz) {
+		rdlen = WT_MIN(bufsz, remainder);
+		WT_ERR(__wt_read(session, fh, off, rdlen, buf));
+		if (memcmp(buf, zerobuf, rdlen) != 0)
+			*hole = true;
```
michaelcahill: Might as well `break` here?

src/log/log.c
```diff
@@ -1592,16 +1563,25 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
...
+		 * When recovery finds the end of the log, truncate the file
+		 * and remove any later log files that may exist.
 		 */
 		if (reclen == 0) {
-			/* This LSN is the end. */
-			eol = true;
-			break;
+			WT_ERR(__log_has_hole(
+			    session, log_fh, rd_lsn.offset, &eol));
+			if (eol == true)
```
michaelcahill: style nit, should just be `if (eol)` -- I think lint will complain.

michaelcahill:
Thanks @sueloverso -- I made a couple of minor comments, then pushed fixes.  LGTM, I'll merge once testing is done.

PR #2275, opened 2015-10-30
WT-2170 Application metadata cursors expect read-uncommitted semantics.

michaelcahill:


michaelcahill:
@daveh86 does this make sense to you, and is it sufficient to fix the MongoDB test failures?

daveh86:
@michaelcahill, as I understand things, yes that commit should resolve the issues we have been seeing. I will patch down and test on my local shortly.

daveh86:
This resolve the issues I was seeing in MongoDB. 

The change itself LGTM.

PR #2276, opened 2015-10-30
WT-2127 Back out some of a change to split internal pages sooner.

agorrod:
We aren't doing an awesome job of balancing trees built via append
workloads, especially when splitting the root. Avoid splitting the
root until it grows large.

The work in WT-2182 should help with our tree shape.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -211,7 +211,8 @@ __split_should_deepen(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * splitting into parent pages can become large enough to result
 	 * in slow operations.
 	 */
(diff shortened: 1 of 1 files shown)


agorrod:
This helps the workload referenced in WT-2127. After making this change I see a tree with 25 million entries grow to 7 levels deep instead of over 100.

This isn't ideal - the root page only splits once it has over 75 thousand entries.

agorrod:
For completeness, I think we should merge this change but keep WT-2127 open. We could do a better job of arranging pages when splitting.

michaelcahill:
OK, thanks @agorrod.  I'll merge.

PR #2278, opened 2015-10-30
SERVER-21027 Don't leave empty internal pages in the tree

michaelcahill:


michaelcahill:
@agorrod, this branch has the changes we discussed today, plus several extra fixes.

In particular, 78f504e was responsible for the assertion failures in eviction -- we could see a transaction ID from a previous reconciliation.

PR #2279, opened 2015-10-30
WT-2191: in memory disk image no longer the same as saved updates

keithbostic:
@agorrod, @michaelcahill, I think this is ready for review and merge.

956cc66 is the fix, 0818fc3 is a general naming change.


michaelcahill:
Thanks, @keithbostic!  lgtm, I'll merge.  Fingers crossed this is the last one...

keithbostic:
@michaelcahill, it wasn't sufficient, I've pushed another fix into develop, 713e521, fingers crossed!

It's a one-line change, so I updated develop trying to get past this set of failures, but please review it!

PR #2280, opened 2015-10-30
WT-2187 Add flag for flushing a slot

sueloverso:
so that we locally handle write_lsn rather than waiting for a worker thread, similar to sync.

@michaelcahill Please review this change for the performance issue reported with write-no-sync.  With this change, I see the current logging code perform the same (one populate thread) or 50% faster (8 threads).


michaelcahill:

PR #2281, opened 2015-11-02
WT-2192 Fix the logic around checking whether internal page is evictable.

agorrod:
There is a race between checking whether all children are in state
WT_REF_DISK and whether the internal page is modified.


michaelcahill:
Thanks, @agorrod, lgtm -- I'll merge once testing is done.

PR #2283, opened 2015-11-02
WT-2194: Java close callbacks should handle cursors that Java code did not open.

ddanderson:
This can now happen, as WiredTigerLAS.wt is opened internally.

### DIFF
--- a/lang/java/wiredtiger.i
+++ b/lang/java/wiredtiger.i
@@ -395,15 +395,20 @@ javaCloseHandler(WT_EVENT_HANDLER *handler, WT_SESSION *session,
 	WT_CURSOR *cursor)
 {
 	int ret;
+	JAVA_CALLBACK *jcb;
 
 	WT_UNUSED(handler);
 
-	if (cursor != NULL)
-		ret = cursorCloseHandler(NULL, session, (JAVA_CALLBACK *)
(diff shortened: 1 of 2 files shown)


ddanderson:
@agorrod, can you review?

One thing to consider is whether the close callback should ever see cursors (like for WiredTigerLAS.wt) that were not explicitly opened by the application.  At the moment, "probably" only the Java/Python glue code uses the close callbacks so it doesn't matter much - I don't know if that will continue to be true?  So another solution would be to tag the WiredTigerLAS.wt cursor so it doesn't get sent to the callback.

agorrod:
Thanks @ddanderson I think your solution is valuable, but ideally we wouldn't be making a callback for LAS cursors either. At the moment LAS cursors are special - so I think it would be good to exclude them from the callback as well.

That should be a matter of adding a clause to the cursor close loop in ``__session_close``, skipping any cursor with a URI that matches the ``WT_LAS_URI`` string.

ddanderson:
@agorrod, I just pushed the change you described.  Tested it independently of the other fix to make sure it too would fix the problem.  Can you review?

agorrod:
Thanks @ddanderson lgtm, I'll merge.

PR #2284, opened 2015-11-02
WT-2184 recovery test

sueloverso:
@michaelcahill Please review this new test based off the repro case sent in for WT-2184.  I have confirmed it fails without the fix with the expected detection of missing records and succeeds in the current develop/branch.


test/recovery/recovery.c
```diff
@@ -0,0 +1,235 @@
...
+	__wt_random_init(&rnd);
+	/* Sleep for the configured amount of time before killing the child. */
+	printf("Parent: sleep %" PRIu32 " seconds, then kill child\n", timeout);
+	sleep(timeout);
+
+	/*
+	 * !!! It should be plenty long enough to make sure more than one
+	 * log file exists.  If wanted, that check would be added here.
+	 */
+	kill(pid, SIGKILL);
```
agorrod: On my machine it takes over 10 seconds for this kill command to complete (above the sleep timeout). Do you think we could speed that up? Maybe throttling the worker would get the signal to complete faster.

Also - using sub processes and signals means that this test is going to be difficult to port to Windows. I don't have a good alternative - since we want to force an unclean shutdown.

test/recovery/recovery.c
```diff
@@ -0,0 +1,235 @@
...
+	 * The data in the buffer is already set to random content.
+	 */
+	data.data = buf;
+	for (i = 0;; ++i) {
+		data.size = __wt_random(&rnd) % MAX_VAL;
+		cursor->set_key(cursor, i);
+		cursor->set_value(cursor, &data);
+		if ((ret = cursor->insert(cursor)) != 0)
+			testutil_die(ret, "WT_CURSOR.insert");
+		if (write(fd, &i, sizeof(i)) != sizeof(i))
```
agorrod: How would you feel about switching to a line-based format? It might make it easier to debug problems if the backup content file is human readable.

test/recovery/recovery.c
```diff
@@ -0,0 +1,235 @@
...
+	if ((ret =
+	    session->open_cursor(session, uri, NULL, NULL, &cursor)) != 0)
+		testutil_die(ret, "WT_SESSION.open_cursor: %s", uri);
+
+	/*
+	 * Keep a separate file with the records we wrote for checking.
+	 */
+	unlink(RECORDS_FILE);
+
+	if ((fd = open(RECORDS_FILE, O_CREAT | O_WRONLY, 0777)) == -1)
```
agorrod: Should we open this file unbuffered? Is it possible that content will be lost when the child is killed?

test/recovery/recovery.c
```diff
@@ -0,0 +1,235 @@
...
+	/*
+	 * !!! If we wanted to take a copy of the directory before recovery,
+	 * this is the place to do it.
+	 */
+	chdir(home);
+	if ((ret = wiredtiger_open(NULL, NULL, ENV_CONFIG_REC, &conn)) != 0)
+		testutil_die(ret, "wiredtiger_open");
+	if ((ret = conn->open_session(conn, NULL, NULL, &session)) != 0)
+		testutil_die(ret, "WT_CONNECTION:open_session");
+	if ((ret = session->log_flush(session, "sync=on")) != 0)
```
agorrod: I don't understand why we need to do a log_flush call here - could you add a comment?

test/recovery/recovery.c
sueloverso: Are you thinking of `setvbuf` with the `_IONBF` type?  That is only for C library file handles, not direct OS file descriptors.  When we write(2) to this file it goes directly to the file system.
The test only checks in one direction, that anything in this file exists in the table because we're using write-no-sync, it must protect against process crash.  It definitely is possible for something to be in the table but not yet in this file.

test/recovery/recovery.c
sueloverso: That is a good idea.  I'll do that.

test/recovery/recovery.c
sueloverso: Good idea, I'll add a throttle to the child process.

test/recovery/recovery.c
sueloverso: That was a leftover from something earlier.  I'll remove.

sueloverso:
@agorrod Please take one more look at this.  I did end up using file handles (unbuffered) as it was cleaner and easier to reformat the saved data that way.

agorrod:
Thanks @sueloverso the code looks OK to me now. It still takes about 45 seconds to run for me, which is longer than I'd like, since it will extend every "make check". I couldn't see a simple way to reduce that runtime.

PR #2285, opened 2015-11-03
WT-2195 Fix a hang after giving up on a reverse split.

michaelcahill:


agorrod:

PR #2286, opened 2015-11-03
WT-2196 Fix size-only statistics when there are LSM tables.

agorrod:
It used to generate an error message when attempting to open a
statistics=(size) cursor on an LSM tree.


michaelcahill:
Thanks, @agorrod, if we end up with enough callers of `__wt_filesize_name` that want an error message, it might be worth providing a wrapper function to output the message, but given the few callers we have now, this lgtm.  I'll merge...

PR #2288, opened 2015-11-03
WT-2196 Fix error handling in size only statistics.

agorrod:
Error codes from failed calls to __wt_filesize_name should not be propagated.

Also add a new test case to test fast stats.


agorrod:
@keithbostic Would you mind taking a look at this branch? I'm open to suggestions if you would prefer a different error handling solution in ``schema_stat.c``.

keithbostic:
@agorrod, lgtm.

keithbostic:
@agorrod, we could do better if we just called `__wt_exist` instead of `__wt_filename`?

(No, I guess that doesn't work, drop could race with the exist call.)

agorrod:
>  I guess that doesn't work, drop could race with the exist call

Yup - that was my chain of thought.

PR #2290, opened 2015-11-04
WT-2219: Enhancements to in-memory testing

keithbostic:
@agorrod, here are some changes to make format run with an in-memory configuration, for your review/consideration.

I didn't find any problems that are issues for MongoDB, so this is just a set of cleanups.

This change would turn on format runs with in-memory configurations (1 in 20 runs); that may be too frequent. There's also an issue of sizing the cache for in-memory configurations, I set it to 100MB per thread in add1199, but I have no sense of whether that's reasonable or not.

I also turned off LSM for in-memory configurations in 9da5926, I don't think it makes sense to create in-memory LSM objects. I think I got all of the LSM trigger paths (including option re-configuration), but I might have missed some.

One issue is we don't support overflow items in in-memory configurations as they're written to the file separately from the pages. I think it's sufficient to increase the maximum size of a key/value in the object, so that's the change I made.

There are some additional reconciliation configuration options we could turn off: for example, the work we do calculating checksums so we can compare blocks with previously "written" blocks, calculating a split-percentage, building dictionaries, raw-compression (well, any kind of compression configuration other than Huffman encoding), those sorts of things.


michaelcahill:
Thanks, @keithbostic.

First up, this belongs in a new ticket: WT-2178 is already resolved and shipped in MongoDB so adding more changes now will cause confusion.

I deliberately avoided trying to make test/format work with in-memory, since it's only notion of correctness is that files verify.  I think all this buys us is some code coverage: do you think there's more?

As far as bulk loads into column stores go, we should change bulk load to not assume "append" -- that must already cause problems for dump/load.  I'll comment in WT-2198.

Thanks for disabling LSM, that is in the tracking doc as a todo.  Likewise the config for overflow items in test/format: that's what MongoDB does to avoid overflow items regardless of whether in-memory is configured.

Turning off checksums makes sense -- I think we can just leave this to application code.  Everything to do with making disk images smaller (e.g., dictionary compression) should apply to in-memory configurations (we build page images in memory, might as well make them as compact as possible).  Block compression should never be called so I don't think we need to do anything special to override / disable it.

keithbostic:
Sorry, didn't realize adding new commit messages to an existing ticket would mess things up -- we can add a new ticket once you decide if you want to take these changes.

I think the change buys us more than just code coverage: format is currently the best tool we have for stress testing (it's generally how we find races and core-dumps), and I think it's worth hammering on the in-memory tables occasionally.

src/conn/conn_sweep.c
```diff
@@ -322,27 +322,31 @@ err:		WT_PANIC_MSG(session, ret, "handle sweep server error");
...
 {
 	WT_CONFIG_ITEM cval;
 	WT_CONNECTION_IMPL *conn;
 
 	conn = S2C(session);
 
+	/*
+	 * Sweeping file handles isn't useful if configured in-memory; because
+	 * the default is to configure sweep, turn it off if running in-memory.
+	 */
```
michaelcahill: I don't follow this -- an application can create and drop tables with `in_memory` configured: why wouldn't we want to clean up obsolete handles?  We don't want to close ordinary handles (hence the previous change forcing `sweep_idle_time=0`), but I think we want sweep to be running.

michaelcahill:
The `test/format` parts of this change mostly seem okay.  I have a question about the sweep changes: they don't look right to me.  Also, the changes to bulk load should probably wait on WT-2198 -- that should avoid the need for any special code around bulk loads for in-memory configurations.

agorrod:
@keithbostic I've made some changes in this branch, and I'd be happy for it to be merged now. Would you mind doing a review and merging if you are happy?

I created a new JIRA ticket, since some of these changes are outside of test/format - and I thought that was worth referencing in the pull request.

I pulled the sweep changes - I'm pretty sure they were wrong. We can resurrect them in a different pull request if you'd like them back.

We have WT-2198 to discuss bulk load with column store tables. I had a quick look at making the change discussed in the ticket - but it's not simple due to the ordering of that code.

This pull request still encompases WT-2208 - since it still has test/format changes.

keithbostic:
@agorrod, the problem with removing the sweep changes is it requires applications set `file_manager=(close_idle_time=0)` to make the in-memory configuration work, that is, the default file_manager configuration won't work.

I've pushed a change to make that work for your review, otherwise I'm ready to merge this branch.

keithbostic:
@agorrod, I'm still seeing occasional (although rare), cache-too-full failures. We could crank up the cache size more, or not run the in-memory configuration by default?

agorrod:
> not run the in-memory configuration by default?

I'd be inclined to start there. If we enhance test/format could handle out-of-memory error returns I'd be more comfortable turning it on occasionally. I'll open a ticket so we don't forget.

PR #2291, opened 2015-11-05
WT-2200 Change WiredTiger caching strategy on Windows

markbenvenuto:
This change splits out `direct_io` into two options on Windows. `direct_io` now controls OS buffering while `write_through` controls disk buffering. This allows consumers like MongoDB to have fine grain control over how they write to disk.

I updated the documentation to reflect the new options.

Lastly, I now define `HAVE_POSIX_MEMALIGN` since TCMalloc provides it.

**Note:** I have aligned the python and text to 80 columns. I was not sure if it was supposed to be 80 or some other limit.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -167,6 +167,7 @@ if useTcmalloc:
     if conf.CheckCHeader('gperftools/tcmalloc.h'):
         wtlibs.append("libtcmalloc_minimal")
         conf.env.Append(CPPDEFINES=['HAVE_LIBTCMALLOC'])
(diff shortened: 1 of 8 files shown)


agorrod:
@markbenvenuto There was one piece missing - which I've added.

@michaelcahill I'm happy with the changes now, but it would be good if you can review the API level changes before we merge.

michaelcahill:
Thanks, @markbenvenuto, merging.  These changes will be in the next drop to MongoDB master.

agorrod:
@michaelcahill This is scheduled for back port, but I believe we backed out of using it in MongoDB 3.2. Should we still do the backport?

PR #2292, opened 2015-11-05
WT-2182: detect internal page split races.

keithbostic:
@agorrod, this makes the race I was seeing (append-only workload, column-store), much less likely to hit, but it still occasionally hits.

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -179,9 +179,14 @@ __evict_delete_ref(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 		 * something is busy, be sure that the page still ends up
 		 * marked deleted.
 		 */
-		if (ndeleted > pindex->entries / 10 && pindex->entries > 1 &&
-		    (ret = __wt_split_reverse(session, ref)) != EBUSY)
-			return (ret);
+		if (ndeleted > pindex->entries / 10 && pindex->entries > 1) {
+			if ((ret = __wt_split_reverse(session, ref)) == 0)
+				return (ret);
+			WT_RET_BUSY_OK(ret);
+
+			/* If a reverse split fails, unlock the child. */
+			WT_ASSERT(session, ref->state == WT_REF_LOCKED);
+		}
 	}
 
 	WT_PUBLISH(ref->state, WT_REF_DELETED);
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -389,9 +389,6 @@ __split_root(WT_SESSION_IMPL *session, WT_PAGE *root)
 	 */
 	pindex = WT_INTL_INDEX_GET_SAFE(root);
 
-	/* Acquire a new split generation. */
-	split_gen = __wt_atomic_addv64(&S2C(session)->split_gen, 1);
-
 	/*
 	 * A prepending/appending workload will repeatedly deepen parts of the
 	 * tree that aren't changing, and appending workloads are not uncommon.
@@ -626,6 +623,7 @@ __split_root(WT_SESSION_IMPL *session, WT_PAGE *root)
 	 * fails, we don't roll back that change, because threads may already
 	 * be using the new index.
 	 */
+	split_gen = __wt_atomic_addv64(&S2C(session)->split_gen, 1);
 	size = sizeof(WT_PAGE_INDEX) + pindex->entries * sizeof(WT_REF *);
 	WT_TRET(__split_safe_free(session, split_gen, false, pindex, size));
 	root_decr += size;
@@ -686,9 +684,6 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref, WT_REF **ref_new,
 	 */
 	pindex = WT_INTL_INDEX_GET_SAFE(parent);
 	parent_entries = pindex->entries;
-
-	/* Acquire a new split generation. */
-	split_gen = __wt_atomic_addv64(&S2C(session)->split_gen, 1);
 
 	/*
 	 * Remove any refs to deleted pages while we are splitting, we have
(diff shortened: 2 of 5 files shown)


keithbostic:
Here's the CONFIG that I see fail with this branch; still no clue why.
```
#  RUN PARAMETERS
abort=0
auto_throttle=1
firstfit=0
bitcnt=3
bloom=1
bloom_bit_count=8
bloom_hash_count=14
bloom_oldest=0
cache=100
checkpoints=1
checksum=off
chunk_size=6
compaction=0
compression=none
data_extend=0
data_source=file
delete_pct=40
dictionary=0
encryption=none
evict_max=0
file_type=variable-length column-store
backups=0
huffman_key=1
huffman_value=0
insert_pct=30
internal_key_truncation=1
internal_page_max=9
isolation=read-uncommitted
key_gap=6
key_max=32
key_min=20
leak_memory=0
leaf_page_max=9
logging=0
logging_compression=zlib
logging_archive=1
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=3
merge_max=4
mmap=1
ops=1280907
prefix_compression=1
prefix_compression_min=0
repeat_data_pct=56
reverse=0
rows=1000000
runs=1000
salvage=0
split_pct=53
statistics=0
statistics_server=0
threads=15
timer=360
value_max=64
value_min=16
verify=0
wiredtiger_config=
write_pct=50
```

keithbostic:
I'm also seeing a couple of other failures with a simple smoke CONFIG:
```
cache=40
rows=100000
runs=1000
timer=4
threads=15
compression=none
logging_compression=none
```
I've seen memory underflow:
```
  5: file, variable-length column-store (320 seconds)                        
[1446997019:161872][13510:0x7f6ae9ea8700], t, file:wt.wt, WT_CURSOR.update: WT_PAGE.memory_footprint underflow: decrementing 86848
[1446997026:327899][13510:0x7f6af3ecd7c0], t, file:wt.wt, WT_SESSION.verify: WT_CACHE.bytes_inmem underflow: decrementing 102447
[1446997120:782438][13510:0x7f6af3ecd7c0], t, metadata-ckpt: cache server: exiting with 15599 bytes in memory
```
and I've seen address-sanitizer failures:
```
==13439==ERROR: AddressSanitizer: heap-use-after-free on address 0x604001616724 at pc 0x0000006a6041 bp 0x7fdf5d33a910 sp 0x7fdf5d33a908
READ of size 4 at 0x604001616724 thread T489
    #0 0x6a6040 in __wt_page_in_func /home/ec2-user/work/wiredtiger/src/btree/bt_read.c:450:11
    #1 0x6fb581 in __wt_page_swap_func /home/ec2-user/work/wiredtiger/./src/include/btree.i:1260:8
    #2 0x6fb581 in __wt_row_search /home/ec2-user/work/wiredtiger/src/btree/row_srch.c:329
    #3 0x7aa529 in __cursor_row_search /home/ec2-user/work/wiredtiger/src/btree/bt_cursor.c:241:2
    #4 0x7aa529 in __wt_btcur_search /home/ec2-user/work/wiredtiger/src/btree/bt_cursor.c:327
    #5 0x72335a in __curfile_search /home/ec2-user/work/wiredtiger/src/cursor/cur_file.c:200:2
```

src/evict/evict_page.c
```diff
@@ -179,9 +179,14 @@ __evict_delete_ref(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 		 * something is busy, be sure that the page still ends up
 		 * marked deleted.
 		 */
-		if (ndeleted > pindex->entries / 10 && pindex->entries > 1 &&
-		    (ret = __wt_split_reverse(session, ref)) != EBUSY)
-			return (ret);
+		if (ndeleted > pindex->entries / 10 && pindex->entries > 1) {
+			if ((ret = __wt_split_reverse(session, ref)) == 0)
+				return (ret);
+			WT_RET_BUSY_OK(ret);
```
agorrod: @keithbostic I don't see why this is necessary. If the reverse split failed with EBUSY, we still want to switch the ref back to ``WT_REF_DELETED``, don't we? Is it possible that we never switched it to locked - in that case I'd be inclined to update the ``WT_ASSERT`` to be ``WT_ASSERT(session, ret == EBUSY || ref->state == WT_REF_LOCKED)``, and let ``EBUSY`` fall through (see following).

src/evict/evict_page.c
```diff
@@ -179,9 +179,14 @@ __evict_delete_ref(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
...
+
+			/* If a reverse split fails, unlock the child. */
+			WT_ASSERT(session, ref->state == WT_REF_LOCKED);
```
agorrod: This code path is now eating the error return from ``__wt_split_reverse`` - I think the return statement should be updated to ``return (ret)``

agorrod:
@keithbostic I see what you are saying about resolving the race, and I agree that it should be enough to fix the race I described in JIRA.

I can also reproduce a crash when running on this branch. I haven't made any progress in tracking it down today.

I'd be inclined to merge this branch (modulo the error handling change I suggested), so we are working from a single split-up-the-tree branch. We need this fix, so might as well keep our lives as simple as possible.

src/evict/evict_page.c
keithbostic: It looks to me like we call `__wt_split_reverse` and immediately return if that's successful, then return if ret is anything other than EBUSY, then fall through to reset the lock. Maybe I'm misreading something, but I think we reset the lock if we get EBUSY here. Let's talk this over tonight -- I can change the `return (ret)` to `return (0)` for clarity, too.

keithbostic:
Thanks, @agorrod, I've done the merge. Let's try and brainstorm with @michaelcahill tonight about this failure, I've been chasing it for awhile, and I'm worried there's something fundamental I'm not seeing.

PR #2293, opened 2015-11-06
WT-2217: change WT_CURSOR.insert to clear "set" key/value on return

keithbostic:
Wording changes to clarify when applications can discard memory passed to set-key/set-value.

@michaelcahill, some minor cursor wording changes to clarify when application-memory is used.

One suggestion that came out of the discussion was we could change `WT_CURSOR.insert` to unset the key/value, so not only would the cursor not be positioned after insert returned, it would also not have a set key/value.

I think it's unlikely for an insert cursor to be intentionally re-used without also doing a fresh set-key/set-value call, so I don't think the change would impact real applications and clearing the key/value would mean we'd get a clean error message instead of dropping core.

If you like that idea, let me know and I'll open a ticket and get it done.

### DIFF
--- a/examples/c/ex_scope.c
+++ b/examples/c/ex_scope.c
@@ -46,15 +46,16 @@ static int
 cursor_scope_ops(WT_CURSOR *cursor)
 {
 	struct {
+		const char *op;
 		const char *key;
 		const char *value;
 		int (*apply)(WT_CURSOR *);
 	} *op, ops[] = {
-		{ "key1", "value1", cursor->insert, },
-		{ "key1", "value2", cursor->update, },
-		{ "key1", "value2", cursor->search, },
-		{ "key1", "value2", cursor->remove, },
-		{ NULL, NULL, NULL }
+		{ "insert", "key1", "value1", cursor->insert, },
+		{ "update", "key1", "value2", cursor->update, },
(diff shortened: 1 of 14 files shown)


michaelcahill:
@keithbostic, I think that clearing the key/value after `WT_CURSOR::insert` makes more sense than trying to explain this corner case.  AFAIK, there is no code that depends on being able to do `set_key, insert, search` or similar, but it should be obvious quickly if I'm wrong.

keithbostic:
@michaelcahill, most of the changes for WT-2217 are here. The tests I had to change to support the new semantics were ex_scope and test_autoclose.py (see 6bdd66e). There's still one test that's not passing: test_schema05, and I think it's going to be a messy fix. If you like where this is going, let me know and I'll figure out test_schema05.

michaelcahill:
@keithbostic, I futzed with the Python tests some because they looked odd to me.  It turns out there is an underlying issue with the Python API, but I'll open a separate ticket for that.

I also fixed the table code to keep the primary cursor "positioned" during index updates, which was the cause of the `schema05` failures.  I'm now happy with this branch if you are.

keithbostic:
@michaelcahill, thanks for the fix -- I don't know what I was thinking when I did that table code!

The rest looks good, I'll merge.

PR #2295, opened 2015-11-06
WT-2202: release of the "current" page is wrong.

keithbostic:
We're releasing the "current" page if (leaf != NULL), and that's wrong, we should only release the current page if no leaf-page WT_REF was specified.  The test could be (leaf == NULL), but I think (leaf != current) is a little more obvious.

@agorrod, @michaelcahill, would one of you please do a review? Thanks!

### DIFF
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -487,7 +487,13 @@ leaf_match:	cbt->compare = 0;
 
 	return (0);
 
-err:	if (leaf == NULL)
+err:	/*
+	 * Release the current page if the search started at the root. If the
+	 * search didn't start at the root we should never have gone looking
+	 * beyond the start page.
+	 */
+	WT_ASSERT(session, leaf == NULL || leaf == current);
+	if (leaf == NULL)
 		WT_TRET(__wt_page_release(session, current, 0));
 	return (ret);
 }
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -487,7 +487,7 @@ leaf_match:	cbt->compare = 0;
 
 	return (0);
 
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic I buy that this change is correct, but do you understand why the bug isn't causing things to break everywhere (or anywhere)?

agorrod:
> The test could be (leaf == NULL), but I think (leaf != current) is a little more obvious.

I disagree with this. It makes more sense to choose the behavior based on the input parameter, since that is presumably how the caller will decide.

If we error out from ``__wt_row_search`` after a call with a non-null ``leaf`` parameter, that has moved on from leaf, won't that cause other problems? i.e: we'll release the ``leaf`` ref twice, but never release the ref the search is sitting on?

keithbostic:
> I buy that this change is correct, but do you understand why the bug isn't causing things to break everywhere (or anywhere)?

It's code that's only called on error in a function that never sees an error.

keithbostic:
>> The test could be (leaf == NULL), but I think (leaf != current) is a little more obvious.

> I disagree with this. It makes more sense to choose the behavior based on the input parameter, since that is presumably how the caller will decide.

OK, fine with me, I've pushed the change.

> If we error out from __wt_row_search after a call with a non-null leaf parameter, that has moved on from leaf, won't that cause other problems? i.e: we'll release the leaf ref twice, but never release the ref the search is sitting on?

If leaf is non-null, we don't move on, the search is constrained to the leaf page, and the caller of the search function is responsible for discarding the page.

agorrod:
Thanks @keithbostic lgtm. I added a comment and an assertion. I'll merge.

PR #2297, opened 2015-11-09
WT-2206: change cache operations from flags to an enum.

keithbostic:
@agorrod, would you please review & merge if you agree?


agorrod:

PR #2298, opened 2015-11-10
WT-2182 Simplify the split deepen logic.

agorrod:
Don't save children at the current depth when deepening from the root.
It hurts the shape of trees we generate.


agorrod:
@keithbostic This change removes the special code that attempts to create better trees for append workloads.

I've been running your program that populates a table by appending little items, and creates large trees by setting a small page size. The following are tree shapes generated by inserting 100 million items. The reported counts are only internal pages, and it's the in-memory pages - so the shape may be different if we've evicted internal pages (we shouldn't have).

Using the new split-deepen code without the changes on this branch:
```
count at level 1 is        1
count at level 2 is     1502
count at level 3 is    15800
count at level 4 is    10859
count at level 5 is    13126
count at level 6 is    12700
count at level 7 is     3582
count at level 8 is    10032
count at level 9 is    16637
count at level 10 is     7299
count at level 11 is    13948
count at level 12 is     5839
count at level 13 is     8784
count at level 14 is    12458
count at level 15 is      266
count at level 16 is        0
```

Using this branch I see:
```
count at level 1 is        1
count at level 2 is     1390
count at level 3 is   141775
count at level 4 is        3
count at level 5 is        0
```

Using this branch, and a ``split_deepen_min_child=1000`` setting:
```
count at level 1 is        1
count at level 2 is       10
count at level 3 is     1117
count at level 4 is   118536
count at level 5 is      361
count at level 6 is        0
```

agorrod:
@keithbostic I've run some more tests, and see this branch showing consistently better shaped trees. Could you review and merge if you are happy?

For the record, results are as follows:

```
conn_config="cache_size=200MB"
table_config="type=file,allocation_size=1KB,leaf_page_max=16KB,internal_page_max=1KB,memory_page_max=64k"
icount=40000000
report_interval=5
populate_threads=5
```

That is an append-mostly populate workload that splits a lot (but not unreasonably often), generates the following tree shape with wt-2182-split-up-the-tree-2:

```
Root: [4980882432-4980883456, 1024, 2603503541] row-store internal
Internal page tree-depth (total 12239):
	001: 1
	002: 10
	003: 279
	004: 3116
	005: 3122
	006: 2809
	007: 2902
Leaf page tree-depth (total 306896):
	004: 5257
	005: 78089
	006: 77966
	007: 71698
	008: 73886
```

And the following on this branch:

```
Root: [4981060608-4981061632, 1024, 2540822023] row-store internal
Internal page tree-depth (total 12414):
	001: 1
	002: 16
	003: 460
	004: 11937
Leaf page tree-depth (total 306897):
	005: 306897
```

My testing shows the runtime between these two branches is very similar - 29.14 vs 29.31.

I also ran a more general insert and query workload and see equivalent performance on both branches. The configuration file for that run is:

```
conn_config="cache_size=200MB"
table_config="type=file,allocation_size=1KB,leaf_page_max=1KB,internal_page_max=1KB,memory_page_max=32KB"
icount=1000000
random_range=50000000
report_interval=5
run_time=120
populate_threads=1
threads=((count=4,inserts=1),(count=8,reads=1))
```

keithbostic:
@agorrod, I'm going to go ahead and merge your branch.

I did a final review & cleaned up some additional code we no longer need -- would you please do a brief review of bd6ed94, b1db77d and e224ee1?

agorrod:
Thanks @keithbostic those changes lgtm.

agorrod:
This change was made as part of the split-up-the-tree changes. It shouldn't be back ported.

PR #2299, opened 2015-11-10
Change tree shape dump code to generate json format.

agorrod:
This change only converts some of the debug dump code.
It also adds calls into the code from a couple of places to aid in
tracing.


agorrod:
@keithbostic I've been using the content of this branch to introspect tree shapes. It generates a json dump that captures the shape of the tree. Once I have that, I've got a grapher that turns the json into a html page that shows the shape of the tree.

The process I use to generate the graph is:
Run a workload in WiredTiger, that generates a bunch of files called ``shapeX`` where X is the generation number, and a final ``shape_close`` generated on connection close.

Once you have a json file you'd like to visualise startup a local webserver from the directory with the visualizer html:
```
$ ls
graph_shape.html	shape_close
$ python -m SimpleHTTPServer
```

Then from your browser navigate to:
``
http://localhost:8000/graph_shape.html?source=shape_close
``

You can change the final "shape_close" part of the url to select a different file. It is OK to have sub-directories in that part of the url.

If you ran a big workload, the graph can take a while to display. I've also set the page size limits very large - since there is a lot of information it tends to get overly squashed if displayed smaller.

The content of graph_shape.html is:

```
<!DOCTYPE html>
<meta charset="utf-8">
<style>

.node circle {
  fill: #fff;
  stroke: steelblue;
  stroke-width: 1.5px;
}

.node {
  font: 10px sans-serif;
}

.link {
  fill: none;
  stroke: #ccc;
  stroke-width: 1.5px;
}

</style>
<body>
<script src="//d3js.org/d3.v3.min.js"></script>
<script>

var filename = location.search.split('source=')[1]
var width = 2000,
    height = 8000;

var tree = d3.layout.tree()
    .size([height, width - 160]);

var diagonal = d3.svg.diagonal()
    .projection(function(d) { return [d.y, d.x]; });

var svg = d3.select("body").append("svg")
    .attr("width", width)
    .attr("height", height)
  .append("g")
    .attr("transform", "translate(40,0)");

d3.json(filename, function(error, json) {
  if (error) throw error;

  var nodes = tree.nodes(json),
      links = tree.links(nodes);

  var link = svg.selectAll("path.link")
      .data(links)
    .enter().append("path")
      .attr("class", "link")
      .attr("d", diagonal);

  var node = svg.selectAll("g.node")
      .data(nodes)
    .enter().append("g")
      .attr("class", "node")
      .attr("transform", function(d) { return "translate(" + d.y + "," + d.x + ")"; })

  node.append("circle")
      .attr("r", 4.5);

  node.append("text")
      .attr("dx", function(d) { return d.children ? -8 : 8; })
      .attr("dy", 3)
      .attr("text-anchor", function(d) { return d.children ? "end" : "start"; })
      .text(function(d) { return d.name; });
});

d3.select(self.frameElement).style("height", height + "px");

</script>
```

agorrod:
An example of a tree shape graph I get from running an append workload for 100 million items is:

screen shot 2015-11-10 at 4 01 13 pm

It's hard to scale it down enough to get a reasonable picture, just wanted to give you and idea of the information we can see.

src/btree/bt_debug.c
```diff
@@ -446,24 +470,45 @@ __debug_tree_shape_info(WT_PAGE *page)
...
+		first = true;
+		count = 0;
 		WT_INTL_FOREACH_BEGIN(session, page, ref) {
-			if (ref->state == WT_REF_MEM)
+			/*
+			 * Only print out every 10th entry, should still show
+			 * the shape of the tree, without overwhelming.
+			 */
+			if (ref->state == WT_REF_MEM) {
+				log_next = logged && count++ % 1 == 0;
```
markbenvenuto: Is this supposed to be `mod 10` instead of `mod 1`?

agorrod: Sort of. I was playing around with limiting how much output is generated, but it meant I was missing important parts of the tree.

I should have been clearer that this branch is only intended for my diagnostic needs at the moment. It'll need a lot more work if we decide to pull it in to the WiredTiger tree.

agorrod:
This was diagnostic code, not ready for prime time. I'm going to close the pull request but leave the branch around for now.

PR #2300, opened 2015-11-10
WT-2207 Track whenever a session has a handle exclusive.

michaelcahill:
Previously, this was only tracked for newly-opened handles, but the change for
SERVER-19954 relies on always detecting when a session has a handle exclusive,
regardless of whether it was the first to open the handle.


agorrod:
The test failure is:

```
[1447197657:829794][9139:0x7fbc3e02f740], test_bulk02.test_bulkload_backup.test_bulk_backup(same.named.file), file:data, WT_SESSION.checkpoint: ../src/txn/txn_ckpt.c, 1211: bm != NULL
```

agorrod:

PR #2301, opened 2015-11-11
WT-2199 Fix transaction sync inconsistency.

sueloverso:
@michaelcahill Please review this change.  It actually fixes two issues.  The first is to add a flag to separate the `none` method as `WT_LOG_FLUSH` from the enable/disable of syncing.  The second is a bug fix where if the user sent in `sync=off` in `begin_transaction` it didn't actually turn it off for `fsync`.  I added a test to verify all the different settings.


sueloverso:
@michaelcahill - The pull-request-tester is failing.  @agorrod reminded me of the Jenkins WIREDTIGER_CONFIG settings which includes `transaction_sync`.  The test runs fine when run by hand.  This test is testing `transaction_sync` and sets up the params to `wiredtiger_open`.  But the environment variable is the highest priority config item processed in `wiredtiger_open` and overrides the test settings.  I'm going to think about options.  I suspect removing it from the environment setting will slow down the test suite.

agorrod:
@sueloverso and I discussed overriding the environment variable in this particular test using the Python ``os.environ`` functionality. That works but is an ugly solution.

michaelcahill:
@sueloverso, I'll follow up in the ticket with a longer comment.  The immediate fix I suggest here is a `use_environment` setting to `wiredtiger_open` that defaults `true` and controls whether `WIREDTIGER_CONFIG` and `WIREDTIGER_HOME` are checked for configuration.

I also support @agorrod's suggestion that Jenkins not set `WIREDTIGER_CONFIG` -- I think we could get the performance benefits with `libeatmydata` or similar without any application-visible impact.

sueloverso:
Sure, I'll add in a `use_environment` config setting tomorrow into this PR.  I believe pull-request-tester is already using `libeatmydata`.  That is a clean solution.

michaelcahill:
Thanks, @sueloverso, the underlying functionality is now merged, ready for the test to be updated to use it...

sueloverso:
@agorrod and @michaelcahill the test is now passing the Jenkins job with the new config setting.  Can one of you please review the changes and the test?  I believe it is ready to merge.

agorrod:

PR #2303, opened 2015-11-12
WT-2135 Fix log_only setting for backup cursor. Fix initialization.

sueloverso:
@keithbostic  Please review this change to cur_backup.c for the two issues that were listed in the ticket.  In particular I moved the initialization of cb->list_next earlier next to where we initialize cb->list.


michaelcahill:
@sueloverso, thanks, I took a look -- this lgtm.  I'll merge...

PR #2304, opened 2015-11-14
WT-2182: fixes for splitting up the tree.

keithbostic:
@michaelcahill, this branch is for your review.

I stared at the code that fixes up the moved WT_REF structures today, and found what I think are 3 problems:

* setting WT_PAGE_MODIFY.mod_split_txn as the page is created, see 7dded31,
* only fixing up moved WT_REF's that are in the WT_REF_MEM state when we look at them, see 9464fe3,
* allowing newly created internal pages to split before we fix up the WT_REFs we moved there, see 9464fe3.

I might be wrong on any or all of these, of course, but I hoped coding up an approach to fixing them would help me fully understand the problems and give you something concrete to look at. I put big honkin' comments all over the place.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -353,35 +353,50 @@ __split_ref_move(WT_SESSION_IMPL *session, WT_PAGE *from_home,
 }
 
 /*
- * __split_child_no_evict --
- *	Ensure the newly created child isn't evicted for now.
+ * __split_child_block_evict_and_split --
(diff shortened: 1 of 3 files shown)


agorrod:
@michaelcahill The use of ``UINT64_MAX`` is gone in the most recent version of this branch.

I agree with your comment regards not allocating an ID, but I believe it's an optimization. I'm going to go ahead and merge this branch, so we can get it into testing. I'll transfer the comment to split up the tree 2.

PR #2305, opened 2015-11-15
WT-2210: raw compression fails if row-store recovery precedes column-store recovery

keithbostic:
If we allocate raw structures for a row-store and then re-use them for a column-store, we might not have allocated the record-number array, r->raw_recnos. Allocate the record-number array regardless of the type of object we're reconciling.

@agorrod, for your review.


agorrod:
@keithbostic This change seems strange to me. Is it "normal" to re-use a reconciliation structure for both row and column store tables? Or is it something we'd only do in recovery?

keithbostic:
@agorrod, the reconciliation structures are per session, so it's not insane to see the same session handle (especially an application session handle), reconciling both row- and column-store pages.

I've pushed a new version of the comment leading into that code clause.

agorrod:

PR #2306, opened 2015-11-16
WT-2220: time_t cleanup.

keithbostic:
@agorrod, I just noticed we were misusing time_t in a few places, for your review.

### DIFF
--- a/src/include/connection.h
+++ b/src/include/connection.h
@@ -365,13 +365,13 @@ struct __wt_connection_impl {
 
 	WT_SESSION_IMPL *meta_ckpt_session;/* Metadata checkpoint session */
 
-	WT_SESSION_IMPL *sweep_session;	/* Handle sweep session */
-	wt_thread_t	 sweep_tid;	/* Handle sweep thread */
-	int		 sweep_tid_set;	/* Handle sweep thread set */
-	WT_CONDVAR	*sweep_cond;	/* Handle sweep wait mutex */
-	u_int		 sweep_idle_time;/* Handle sweep idle time */
-	u_int		 sweep_interval;/* Handle sweep interval */
-	u_int		 sweep_handles_min;/* Handle sweep minimum open */
+	WT_SESSION_IMPL *sweep_session;	   /* Handle sweep session */
+	wt_thread_t	 sweep_tid;	   /* Handle sweep thread */
+	int		 sweep_tid_set;	   /* Handle sweep thread set */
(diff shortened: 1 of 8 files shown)


src/include/misc.i
```diff
@@ -30,6 +30,22 @@ __wt_strdup(WT_SESSION_IMPL *session, const char *str, void *retp)
 }
 
 /*
+ * __wt_seconds --
```
agorrod: Is this used in a performance path? I preferred it being bundled in with the other timer functions.

src/include/connection.h
```diff
@@ -369,8 +369,8 @@ struct __wt_connection_impl {
 	wt_thread_t	 sweep_tid;	/* Handle sweep thread */
 	int		 sweep_tid_set;	/* Handle sweep thread set */
 	WT_CONDVAR	*sweep_cond;	/* Handle sweep wait mutex */
-	time_t		 sweep_idle_time;/* Handle sweep idle time */
-	time_t		 sweep_interval;/* Handle sweep interval */
+	u_int		 sweep_idle_time;/* Handle sweep idle time */
```
agorrod: Would you mind if this was defined size? Maybe a ``uint64_t``

agorrod:
@keithbostic This generally looks good to me modulo a few nits/questions. I think we should also watch the Windows build - some googling indicated that difftime works the same on Windows, but worth keeping an eye on.

src/include/misc.i
keithbostic: Then we have to have two copies, that's why I pulled out.

A bunch of the stuff in misc.i isn't really performance functions, it's more like convenience functions than anything else. We could make it a define, actually...

Anyway, if you decide it's better, I'll put two copies back into the os_xxx time source files.

src/include/connection.h
keithbostic: We're inconsistent about the types we use to store config values, but as it's coming from a config value, it might as well be an int64_t, that matches the underlying config size?

keithbostic:
@agorrod, @markbenvenuto: I agree, we should watch the Windows build -- my Googling agrees with yours, difftime is generally available on Windows, but there's always some risk.

src/include/connection.h
agorrod: Sure - ``int64_t`` works for me.

keithbostic: I'm actually curious why you want a defined size? Why does a fixed-size matter here?

keithbostic:
@agorrod, pushed, let me know if you want any other changes.

agorrod:

PR #2307, opened 2015-11-16
WT-2218 Add truncate stats

sueloverso:
@agorrod I have added both a connection and data source statistic for truncate calls.  Can you please review?  You suggested a statistic for pages deleted.  We already have the data source stat `rec_page_delete` that gets incremented in a regular fashion with truncate calls when I run the `mongodb-oplog.wtperf` configuration.  Is there more information that you feel is necessary?

Also, in this change I put the `truncate` stat in with the cursor stats even though it comes through the session handle.  It  felt more logical there.  Let me know if you disagree.


agorrod:
> You suggested a statistic for pages deleted. We already have the data source...

It would be good to add a connection wide statistic. I think it would also be interesting to split out deletes for truncate, if that's not too complex, though I had a look at the code and it wasn't obvious to me how to make that work.

Generally what I'd like to know is:
* How many truncates have been done.
* How much data have those truncates removed.

If you think the existing ``rec_page_delete`` is as close to the second goal as is practical without adding overhead, then I'm happy with the shape of this change.

> I put the truncate stat in with the cursor stats even though it comes through the session handle.

That's a tough one for me. It's in the public API on the session handle, but truncates are definitely correlated with a particular handle - and cursor stats is how we present that distinction. I think tying it to cursors makes sense.

agorrod:

PR #2308, opened 2015-11-17
WT-2220 Split WT_TIMEDIFF macro into unit specific macros.

agorrod:
We regularly divide the results, encapsulate that better.


agorrod:
@keithbostic What do you think of this change? I'd originally wanted to switch the new ``difftime`` calls to ``WT_TIMEDIFF``, but they operate on ``time_t`` not ``struct timespec``. It's not simple to change the sweep code to use timespec so I left it alone.

bench/wtperf/wtperf.c
```diff
@@ -653,7 +653,7 @@ op_err:			if (ret == WT_ROLLBACK && ops_per_txn != 0) {
 					goto err;
 				}
 				++trk->latency_ops;
-				usecs = ns_to_us(WT_TIMEDIFF(stop, start));
+				usecs = ns_to_us(WT_TIMEDIFF_NS(stop, start));
```
sueloverso: I think this should also have its own WT_TIMEDIFF_US.

keithbostic:
@agorrod, I think this is a good idea, ready to merge when you are.

agorrod:
@keithbostic 

> should we have WT_TEN_THOUSAND and WT_HUNDRED_THOUSAND

I prefer ``10000`` to a string, and I also prefer it to ``10 * WT_THOUSAND``. The reason I like ``WT_MILLION`` is that I find it difficult to parse all the zeroes to figure out what the number is.

I'm going to merge - we can keep pushing on this change in another branch if we want.

PR #2309, opened 2015-11-17
WT-2182: update the empty WT_REF before trying the reverse split.

keithbostic:
@michaelcahill, for your review/consideration: reverts 846125e by setting the WT_REF state to WT_REF_DELETED before trying the reverse split.

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -151,13 +151,14 @@ done:	if (((inmem_split && ret == 0) || (forced_eviction && ret == EBUSY)) &&
 static int
 __evict_delete_ref(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 {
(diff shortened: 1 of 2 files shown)


keithbostic:
autobuild: retest this please

src/evict/evict_page.c
```diff
@@ -159,6 +158,8 @@ __evict_delete_ref(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 	if (__wt_ref_is_root(ref))
 		return (0);
 
+	WT_PUBLISH(ref->state, WT_REF_DELETED);
+
```
michaelcahill: What prevents the parent page being evicted now?

keithbostic: And, that kills that idea -- nothing obvious, we're done.

PR #2310, opened 2015-11-17
WT-2182: continue the reverse split even if the resulting internal page is empty 

keithbostic:
@michaelcahill, for your review/consideration: I can't think of a reason we can't leave one of the internal page's WT_REF slots in the WT_REF_DELETED state, and continue with the reverse split. I arbitrarily picked the first slot.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -740,17 +740,19 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref, WT_REF **ref_new,
 	/*
 	 * The final entry count consists of the original count, plus any new
 	 * pages, less any WT_REFs we're removing.
+	 *
+	 * We can't leave an empty internal page: if there are no remaining
+	 * entries on the parent, leave the first element in place and mark
+	 * the page to be evicted soon.
 	 */
 	result_entries = (parent_entries + new_entries) - deleted_entries;
-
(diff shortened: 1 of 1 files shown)


keithbostic:
autobuild: retest this please

src/btree/bt_split.c
```diff
@@ -743,19 +743,21 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref, WT_REF **ref_new,
...
 	if (ref->state != WT_REF_SPLIT)
 		--result_entries;
-
-	/*
-	 * If there are no remaining entries on the parent, give up, we can't
-	 * leave an empty internal page. Mark it to be evicted soon and clean
-	 * up the references that have changed state.
-	 */
 	if (result_entries == 0) {
+		pindex->index[0]->state = WT_REF_DELETED;
```
michaelcahill: This was my first attempt at dealing with issue.  One problem is that we need to know whether `pindex->index[0] == ref`, because in that case, we will override the state we're setting here (at line 810).

keithbostic: Right; and if there's only a single entry, it's even trickier. I think we could make it work, but I don't see it's a big deal (mostly because we'll only fail for this reason once, it won't happen repeatedly), so I'll close this branch as well.

PR #2314, opened 2015-11-18
WT-2224 Track which deleted refs are discarded by a split.

michaelcahill:


michaelcahill:
@keithbostic, can you please review this one?  I'm open to better ideas / cleaner implementation, but I think we need to do something along these lines to be sure that only the thread that wins the race to atomically flip the state of refs to WT_REF_SPLIT goes on to free any obsolete contents.

Also worth considering: are you confident that a reader can't have a cached pointer to `page_del` when this code runs?

```c
 880                         /*
 881                          * The page_del structure can be freed immediately: it
 882                          * is only read when the ref state is WT_REF_DELETED.
 883                          * The size of the structure wasn't added to the parent,
 884                          * don't decrement.
 885                          */
 886                         if (next_ref->page_del != NULL) {
 887                                 __wt_free(session,
 888                                     next_ref->page_del->update_list);
 889                                 __wt_free(session, next_ref->page_del);
 890                         }
```

Maybe they should be switched to `__wt_safe_free`?

keithbostic:
@michaelcahill, I've looked at this one, and I'm missing the failure you're seeing.

To restate the obvious, the page is locked before `__split_parent` is called so only one thread walks the page's index, flipping the `WT_REF` state to `WT_REF_SPLIT`: how did two threads attempt to do the same thing to the same `WT_REF`? That should only happen sequentially, and I don't see how a thread could flip the `WT_REF` to `WT_REF_SPLIT`, free the memory (and safe-free the `WT_REF`), and then another thread would see the state as `WT_REF_DELETED`? And `WT_REF_SPLIT` is only set in `__split_parent`.

I agree it looks like that's what's happening, based on the log files, but I don't see how it's possible.

I've also looked at `page_del`, and I don't see any races. Here's the list of places we touch `page_del`:
* `__wt_delete_page`: `WT_REF` state is `WT_REF_LOCKED`,
* `__wt_delete_page_rollback`: this code is aborting the page-delete transaction, so the transaction can't be globally visible, so `__split_parent` won't have set the `WT_REF` state to `WT_REF_SPLIT`,
* `__wt_delete_page_skip`: `WT_REF` state is `WT_REF_LOCKED`,
* `__wt_delete_page_instantiate`: `WT_REF` state is `WT_REF_LOCKED`,
* `__wt_free_ref`: the `WT_REF` is being freed,
* `__split_parent`: the `WT_REF` state is `WT_REF_SPLIT`,
* `__rec_child_modify`: the `WT_REF` state is `WT_REF_LOCKED`.

That said, I think we should change the free calls to be safe-free, out of sheer, raging paranoia.

One other issue: what do you think of switching the deleted `WT_REF` states to `WT_REF_LOCKED` instead of `WT_REF_SPLIT`. That way, threads trying to read them will do the right thing and wait (in case there's a failure in the split), rather than immediately returning `WT_RESTART`. I can't think of a specific failure that fixes, but it's more consistent with how we're protecting those slots in other places.

Let's talk this over when you come online today and push the changes. I'll prep a couple of branches based on my notes so we can take them or discard them easily at that time.

michaelcahill:
@keithbostic, as discussed, I'm planning to merge this branch and #2315 out of paranoia.  I think some simplification is possible in future, and using `WT_REF_LOCKED` during the split seems like a good idea now that we track the refs we're discarding.  That said, I don't want to try to make those changes today, so let's go with what we have.

PR #2315, opened 2015-11-18
WT-2224: use the split-safe-free function to free the WT_REF's page-delete information

keithbostic:


michaelcahill:
Thanks, @keithbostic, while I agree with your analysis and this probably isn't necessary today, I think some paranoia is warranted here.

PR #2316, opened 2015-11-18
WT-2215 Use 32-bit LSN file and offsets.  Set LSNs as a 64-bit value.

sueloverso:
@michaelcahill Please review this set of changes to make LSNs a single 64 bit value that we can atomically update.  I created macros to set and get the file and offset part of that value.

### DIFF
--- a/src/include/log.h
+++ b/src/include/log.h
@@ -5,6 +5,13 @@
  *
  * See the file LICENSE for redistribution information.
  */
+/*
+ * WT_LSN --
+ *	A log sequence number, representing a position in the transaction log.
+ */
+struct __wt_lsn {
+	uint64_t file_offset;
+};
 
 #define	WT_LOG_FILENAME	"WiredTigerLog"		/* Log file name */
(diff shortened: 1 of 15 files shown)


src/include/log.h
```diff
@@ -5,6 +5,13 @@
  *
  * See the file LICENSE for redistribution information.
  */
+/*
+ * WT_LSN --
+ *	A log sequence number, representing a position in the transaction log.
+ */
+struct __wt_lsn {
+	uint64_t file_offset;
+};
```
michaelcahill: @sueloverso, my big question here is why we can't have `WT_LSN` as:

```c
union {
    uint64_t file_offset;
    struct {
        uint32_t file;
        uint32_t offset;
    } x;
};
```

I understand that we need to consider endianness in order to compare LSNs with a single 64-bit comparison (which seems like a big win), but we could address that in this one place with `#ifdef WORDS_BIGENDIAN` and avoid lots of macros and all of the bit-shifting.

Is there some reason why that isn't possible?

michaelcahill:
Closing in favor of #2461.

PR #2317, opened 2015-11-19
WT-1315 Fix some leaks with join cursors.

michaelcahill:


michaelcahill:
@agorrod, can you please take a look if/when this passes tests?

agorrod:

PR #2321, opened 2015-11-19
WT-2227 Resolve lint errors in join implementation

ddanderson:
@keithbostic, can you review and check with lint?  Thanks!

### DIFF
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -653,9 +653,9 @@ __session_join(WT_SESSION *wt_session, WT_CURSOR *join_cursor,
 	WT_INDEX *idx;
 	WT_SESSION_IMPL *session;
 	WT_TABLE *table;
-	uint8_t flags;
-	uint32_t bloom_bit_count, bloom_hash_count, range;
 	uint64_t count;
+	uint32_t bloom_bit_count, bloom_hash_count;
+	uint8_t flags, range;
 
 	count = 0;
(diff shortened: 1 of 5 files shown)


keithbostic:
@ddanderson, one new one popped up:
```
< #... join_join(session, cjoin, idx, ref_cursor, flags, range, count, bloom_bit_
<           range, count, bloom_bit_count, bloom_hash_count));
< src/session/session_api.c  738  Info 734: Loss of precision (arg. no. 6) (32
<     bits to 8 bits)
```

src/session/session_api.c
```diff
@@ -653,8 +653,9 @@ __session_join(WT_SESSION *wt_session, WT_CURSOR *join_cursor,
 	WT_INDEX *idx;
 	WT_SESSION_IMPL *session;
 	WT_TABLE *table;
-	uint64_t bloom_bit_count, bloom_hash_count, count;
-	uint32_t flags, range;
+	uint8_t flags;
+	uint32_t bloom_bit_count, bloom_hash_count, range;
+	uint64_t count;
```
sueloverso: @ddanderson swap `flags` and `count` lines so that they're in decreasing size order.

ddanderson:
@keithbostic, @sueloverso, I think this should resolve both issues.

ddanderson:
@keithbostic, thanks, I think this is ready.

keithbostic:
Thanks Don, this looks great.

PR #2322, opened 2015-11-19
WT-2229: failed compression back-off

keithbostic:
@agorrod, @sueloverso, @michaelcahill, for your review/consideration.


sueloverso:
@keithbostic  This looks like a good start.  Can you generalize it so that it can be used for log record compression too?  If the data block doesn't compress well, then it is probably likely that log records recording the insert of that data won't compress well either.

keithbostic:
@sueloverso, I don't see a way to generalize it for log record compression, there's no shared information here that I can see, and the supporting functions are pretty small. Might as well duplicate the code?

michaelcahill:
Given the results we've seen in testing and my belief that these workloads aren't common, we've agreed to close this.

PR #2323, opened 2015-11-20
SERVER-21568 Fix a use after-free

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - the changes make sense. I'll merge.

@keithbostic This code is conceptually difficult - I'd appreciate if you can look through the changes as well.

PR #2325, opened 2015-11-20
WT-2231: check the leaf page's parent keys before doing the full binary search

keithbostic:

I've only done the code for row-store, if you decide we want this change I'll do the column-store version.


keithbostic:
autobuild: retest this please

keithbostic:
I just added code in 26f57cf and 3a291f0 to update the WT_REF structure's page-index hint during splits and when reading in new internal pages. If we don't take the rest of this change, we should cherry-pick 26f57cf and possibly 3a291f0.

src/btree/row_srch.c
```diff
@@ -179,8 +229,28 @@ __wt_row_search(WT_SESSION_IMPL *session,
...
-	/* We may only be searching a single leaf page, not the full tree. */
+	/*
+	 * We may be searching only a single leaf page, not the full tree. If
+	 * the page links to a parent (in other words, if it's a cursor with
+	 * a pinned page, not a page be re-instantiated in memory by the split
+	 * code), check the page's parent keys before doing the full search,
+	 * it's faster when the cursor is being re-positioned.
+	 */
 	if (leaf != NULL) {
+		if (leaf->home) {
```
michaelcahill: style nit, should be `leaf->home != NULL`

michaelcahill:
OK, @keithbostic, this makes sense to me.  Merging...

PR #2326, opened 2015-11-22
WT-2198: rework column store bulk-load.

keithbostic:
@agorrod, @michaelcahill, here are the changes for WT-2198, for your review/consideration.

@michaelcahill, I didn't change the `wt load` command to use a bulk cursor. Since it can't work on anything other than simple tables, I doubt `wt load` will be a performance issue for anyone. Let me know if you want that work done, I could certainly add a `bulk` flag to the command.


keithbostic:
@michaelcahill, in the bulk cursor code we save a copy of the insert key (to detect out-of-order row-store inserts), and a copy of the insert value (to detect duplicate column-store values for RLE). We could speed that up by swapping the contents of the WT_CURSOR's WT_ITEM buffer with the local buffer instead of copying the key or value into the local buffer.

I was thinking of making that change, but then I realized it could cause the application to see incorrect data for a WT_CURSOR.get_{key,value} call. It's unlikely an application would do that, but it's possible, of course.

Is there any way to make a buffer swap work?

agorrod:
autobuild: retest this please

src/cursor/cur_bulk.c
```diff
@@ -208,13 +289,12 @@ __curbulk_insert_row_skip_check(WT_CURSOR *cursor)
...
 	 */
 	CURSOR_API_CALL(cursor, session, insert, btree);
+	WT_STAT_FAST_DATA_INCR(session, cursor_insert_bulk);
 
-	WT_CURSOR_NEEDKEY(cursor);
-	WT_CURSOR_NEEDVALUE(cursor);
-
-	WT_ERR(__wt_bulk_insert_row(session, cbulk));
+	WT_CURSOR_CHECKKEY(cursor);
+	WT_CURSOR_CHECKVALUE(cursor);
```
agorrod: Why did you decide to bump the statistic before the error checking?

src/include/cursor.h
```diff
@@ -211,33 +211,30 @@ struct __wt_cursor_btree {
 struct __wt_cursor_bulk {
 	WT_CURSOR_BTREE cbt;
 
-	WT_REF	*ref;			/* The leaf page */
-	WT_PAGE *leaf;
+	bool	 first_insert;		/* First insert, else check order  */
```
agorrod: Can you call these three out in a block saying that they are only used in the column store case?

src/include/cursor.h
```diff
@@ -211,33 +211,30 @@ struct __wt_cursor_btree {
 struct __wt_cursor_bulk {
 	WT_CURSOR_BTREE cbt;
 
-	WT_REF	*ref;			/* The leaf page */
-	WT_PAGE *leaf;
```
agorrod: Was there a reason why you moved these lower in the structure? I quite liked that the shared portions of the bulk cursor were at the top. I don't care much, just wondering.

agorrod:
> I didn't change the wt load command to use a bulk cursor. Since it can't work on anything other than simple tables

I've created a different JIRA ticket to track that work WT-2271.

src/cursor/cur_bulk.c
keithbostic: @agorrod, I've been migrating to this style over the past few months. I don't believe it matters whether we increment statistics before doing error checking (it's just as valid to count the total calls as to count the total "successful" calls), and it usually simplifies code paths at the end of the function.

In other words, get the simple stuff (that can't fail), out of the way as part of the function setup, and there will be fewer complications later on.

In this particular case, it makes less difference than in other cases I've seen, but I get to skip a `WT_ERR` call and I think the code is easier to read.

src/include/cursor.h
keithbostic: @agorrod:

> Was there a reason why you moved these lower in the structure?

The `ref`, `leaf` and `reconcile` fields are all part of reconciliation's support for bulk-load, so I put them in one place.

src/include/cursor.h
keithbostic: @agorrod:

> Can you call these three out in a block saying that they are only used in the column store case?

The `first_insert` field is both row- and column-store; I'll move `first_insert` next to the `WT_ITEM last` field (which is also shared), and call the remaining two out in a block as only column-store support. I can move the reconciliation information to the end, too.

keithbostic:
@agorrod: 

>> I didn't change the wt load command to use a bulk cursor. Since it can't work on anything other than simple tables

> I've created a different JIRA ticket to track that work WT-2271.

Thanks. To be clear, I'd be OK ignoring that work, I doubt it's worth doing.

agorrod:

PR #2327, opened 2015-11-23
SERVER-21585 Clean up LAS based on the current count of inserted records

michaelcahill:


agorrod:
Thanks @michaelcahill this change lgtm - it's definitely better for removes from the LAS to be able to keep up with inserts. @keithbostic Should review and merge if he is happy with the approach.

keithbostic:
@michaelcahill, this looks OK to me. I pushed one bug fix, please review 7c7259a.

PR #2328, opened 2015-11-23
SERVER-21585 Don't use the lookaside file until the cache is stuck full.

agorrod:
Also only write into the lookaside file from internal (aka eviction) threads, so application threads don't pay the performance cost.

It is possible that the cache stuck check is too restrictive, but using only the eviction aggressive flag can result in writing a lot of content to the lookaside file when pages can still be evicted.

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -473,7 +473,8 @@ __evict_review(
 			LF_SET(WT_EVICT_IN_MEMORY | WT_EVICT_UPDATE_RESTORE);
 		else if (page->read_gen == WT_READGEN_OLDEST)
 			LF_SET(WT_EVICT_UPDATE_RESTORE);
-		else if (__wt_eviction_aggressive(session))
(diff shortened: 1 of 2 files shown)


agorrod:
@keithbostic What do you think of this change? I think we probably want the check for whether to use the LAS table to be somewhere in the middle of the current aggressive and stuck flags. This definitely shows better characteristics for the SERVER-21585 workloads I was testing on today.

agorrod:
> shows better characteristics

To be more concrete. Without this change I see the LAS file grow to ~300MB in less than 10 minutes. With this change I don't see the LAS file used in over 2 hours.

keithbostic:
@agorrod, I agree with you, we should turn on the LAS table somewhere in the middle between the current aggressive and stuck flags.

I disagree with the change to only write lookaside records from the eviction threads. If the cache is stuck, I don't think an application thread is getting real work done anyway, so if it's doing eviction work it should configure the lookaside table.

I'll go ahead and merge this, if it helps with SERVER-21585.

Cc: @michaelcahill

PR #2330, opened 2015-11-24
SERVER-21027 Reverse split if there are many deleted pages (3.0)

agorrod:
Many manual merge decisions - worth reviewing if looking for problems.


agorrod:
I thought the hang was WT-2185, but I tested the fix for that locally and it didn't resolve the problem. I need to dig deeper.

michaelcahill:
Thanks, @agorrod, lgtm.  I'll merge, let's get it tested in our branch.

PR #2331, opened 2015-11-24
WT-2237 Have threads publish unique transaction IDs so that updates always become visible immediately on commit

michaelcahill:
Previously, WiredTiger avoided locking in the transaction subsystem by having
threads race to allocate transaction IDs. Only one thread wins the race, but
losing threads could temporarily leave a transaction ID published in the global
state table.  If the winning thread completed its transaction quickly enough,
its changes may not become visible immediately to readers (until the losing
threads have woken up and allocated a new ID).

This change separates ID allocation from the notion of a shared "current"
value.  Every thread gets a unique transaction ID, and spins to update the
current value.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -327,22 +327,22 @@ __wt_txn_id_alloc(WT_SESSION_IMPL *session, bool publish)
 	txn_global = &S2C(session)->txn_global;
 
 	/*
-         * Allocating transaction IDs involves several steps.
-         * 
-         * Firstly, we do an atomic increment to allocate a unique ID.  The
-         * field we increment is not used anywhere else.
-         *
-         * Then we optionally publish the allocated ID into the global
-         * transaction table.  It is critical that this becomes visible before
-         * the global current value moves past our ID, or some concurrent
-         * reader could get a snapshot that makes our changes visible before we
-         * commit.
-         *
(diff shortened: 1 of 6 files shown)


src/include/txn.i
```diff
@@ -332,6 +315,55 @@ __wt_txn_idle_cache_check(WT_SESSION_IMPL *session)
...
+	id = __wt_atomic_addv64(&S2C(session)->txn_global.alloc, 1) - 1;
+
+	if (publish) {
+		session->txn.id = id;
+		WT_PUBLISH(WT_SESSION_TXN_STATE(session)->id, id);
+	}
+
+	while (txn_global->current != id ||
+	    !__wt_atomic_casv64(&txn_global->current, id, id + 1))
+		__wt_yield();
```
agorrod: Is yield the right thing to do here? Don't we want to be aggressive about updating the current ID?

agorrod:
Thanks @michaelcahill lgtm - I'm happy with this change. I had one question - but it's about performance not correctness.

I'll go ahead and merge - if we see performance problems we can have a discussion about backoff.

src/include/txn.i
michaelcahill: We could certainly try spinning with `WT_PAUSE` for a while before yielding.

PR #2332, opened 2015-11-24
WT-2216: remove minor test in the row-store search loop.

keithbostic:
@michaelcahill, this is the change for WT-2216, I'm just getting it into the list of branches for consideration when the tree opens up.


michaelcahill:
Thanks, @keithbostic (and @agorrod and @fedorova).  Let's merge this one and close the ticket.  We can follow up if any of the odd observations translate into real performance differences in tests.

PR #2333, opened 2015-11-24
WT-2237 Avoid yields if we race allocating transaction IDs.

michaelcahill:
It had a major perf impact on parallel populate workloads including YCSB.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -323,6 +323,7 @@ __wt_txn_id_alloc(WT_SESSION_IMPL *session, bool publish)
 {
 	WT_TXN_GLOBAL *txn_global;
 	uint64_t id;
+	u_int i;
 
 	txn_global = &S2C(session)->txn_global;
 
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, can you please review this one and merge if you think it is an improvement?  We'll want YCSB results to know for sure whether it fixes some / all of the perf impact of the fix.

src/include/txn.i
```diff
@@ -353,14 +354,17 @@ __wt_txn_id_alloc(WT_SESSION_IMPL *session, bool publish)
...
-	    !__wt_atomic_casv64(&txn_global->current, id, id + 1))
-		__wt_yield();
+	for (i = 0; txn_global->current != id; i++)
+		if (i < 100)
+			WT_PAUSE();
+		else
+			__wt_yield();
 
-	 return (id);
+	WT_PUBLISH(txn_global->current, id + 1);
```
michaelcahill: The reasoning for the change (apart from spinning) is that by the time we get here, we know `current == id`, so we don't need an atomic compare and swap.

keithbostic:
@michaelcahill, this makes sense to me, merged.

Reading the description of this one, I was wondering if there might be a way to detect the potential failure before the commit returns, rather than doing the work upfront?

PR #2334, opened 2015-11-24
SERVER-21641  Resolve Coverity analysis warnings in join code.

ddanderson:
In __wt_config_gets_def() there is a conditional reference of cfg[2].  Presumably this means that cfg strings are de facto required to have at least 3 entries.  If so, would never manifest as a runtime bug, but it's easy enough to satisfy Coverity.


ddanderson:
Oops, compared against the wrong base (master).  Closed.

PR #2335, opened 2015-11-24
WT-2234: Coverity analysis warnings 

ddanderson:
In __wt_config_gets_def() there is a conditional reference of cfg[2]. Presumably this means that cfg strings are de facto required to have at least 3 entries. If so, would never manifest as a runtime bug, but it's easy enough to satisfy Coverity.

### DIFF
--- a/ext/extractors/csv/csv_extractor.c
+++ b/ext/extractors/csv/csv_extractor.c
@@ -97,8 +97,10 @@ csv_extract(WT_EXTRACTOR *extractor, WT_SESSION *session,
 		strncpy(copy, p, len);
 		copy[len] = '\0';
 		if (csv_extractor->format_isnum) {
-			if ((val = atoi(copy)) < 0)
+			if ((val = atoi(copy)) < 0) {
+				free(copy);
 				return (EINVAL);
+			}
 			result_cursor->set_key(result_cursor, val);
 		} else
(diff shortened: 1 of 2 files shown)


ddanderson:
@keithbostic, can you review?  Is there a way to run it against Coverity?  My commit comment is the only explanation I could come up with on what the complaint means.  Thanks!

keithbostic:
@ddanderson, sorry, didn't realize you didn't have access to the Coverity information.

The `cfg` complaint is because Coverity thinks this code in `config.c`:
```
if (cfg == NULL || cfg[0] == NULL || cfg[1] == NULL)
        return (0);
if (cfg[2] == NULL) {
        WT_RET_NOTFOUND_OK(
            __wt_config_getones(session, cfg[1], key, value));
        return (0);
}
```
is a problem because it looks at `cfg[2]`, even though we know that `cfg[1] == NULL` if there's no `cfg[2]`. I can't figure out a way to shut it up, so we just ignore those complaints.

That warning is referenced in both WT-2234 and SERVER-21641.

There's an additional complaint in WT-2234, we could leak memory on an error return out of `csv_extract()`, and that one is real.

I've reverted your commit c9907c6 and fixed the resource leak in bc1301a, please take a quick look when you have a second.

ddanderson:
Thanks, @keithbostic, that all looks good.

PR #2336, opened 2015-11-25
SERVER-21619 Don't do internal page splits after a tree is marked DEAD.

agorrod:
It leads to problems where eviction attempts to write back to a file after the block manager is already closed.

### DIFF
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -77,7 +77,7 @@ __wt_delete_page(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
 		}
 
 		(void)__wt_atomic_addv32(&S2BT(session)->evict_busy, 1);
-		ret = __wt_evict(session, ref, 0);
+		ret = __wt_evict(session, ref, false);
 		(void)__wt_atomic_subv32(&S2BT(session)->evict_busy, 1);
 		WT_RET_BUSY_OK(ret);
 	}
--- a/src/btree/bt_discard.c
+++ b/src/btree/bt_discard.c
@@ -50,15 +50,20 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	page = *pagep;
 	*pagep = NULL;
 
+	if (F_ISSET(session->dhandle, WT_DHANDLE_DEAD) &&
+	    __wt_page_is_modified(page))
+		__wt_page_modify_clear(session, page);
+
 	/*
-	 * We should never discard ...
+	 * We should never discard:
+	 * - a dirty page,
+	 * - a page queued for eviction, or
+	 * - a locked page.
 	 */
-	WT_ASSERT(		/* ... a dirty page */
-	    session, !__wt_page_is_modified(page));
-	WT_ASSERT(		/* ... a page queued for LRU eviction */
-	    session, !F_ISSET_ATOMIC(page, WT_PAGE_EVICT_LRU));
-	WT_ASSERT(		/* ... a locked page */
-	    session, !__wt_fair_islocked(session, &page->page_lock));
+	WT_ASSERT(session, !__wt_page_is_modified(page) ||
+	    F_ISSET(session->dhandle, WT_DHANDLE_DEAD));
+	WT_ASSERT(session, !F_ISSET_ATOMIC(page, WT_PAGE_EVICT_LRU));
+	WT_ASSERT(session, !__wt_fair_islocked(session, &page->page_lock));
 
 #ifdef HAVE_DIAGNOSTIC
 	{
--- a/src/btree/bt_discard.c
+++ b/src/btree/bt_discard.c
@@ -60,8 +60,7 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	 * - a page queued for eviction, or
 	 * - a locked page.
 	 */
-	WT_ASSERT(session, !__wt_page_is_modified(page) ||
-	    F_ISSET(session->dhandle, WT_DHANDLE_DEAD));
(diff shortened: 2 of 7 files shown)


agorrod:
@daveh86 Could you review this change, and try the patch out on your reproducer?

agorrod:
@michaelcahill Could you do a review of this change please?

daveh86:
I've run this version of the patch 4 times and was not able to reproduce. Looks like this fixes the problem.

Code changes LGTM

keithbostic:
@agorrod, why we can't call `__wt_evict_page_clean_update` without first calling `__evict_review`? There's code in `__wt_evict_file` that makes me think that's possible:
```
case WT_SYNC_DISCARD:
        /*
         * Dead handles may reference dirty pages; clean the
         * page, both to keep statistics correct, and to let
         * the page-discard function assert no dirty page is
         * ever discarded.
         */
        if (F_ISSET(session->dhandle, WT_DHANDLE_DEAD))

        WT_ASSERT(session,
            F_ISSET(session->dhandle, WT_DHANDLE_DEAD) ||
            __wt_page_can_evict(session, ref, false, NULL));
        __wt_evict_page_clean_update(session, ref, 1);
        break;
```
I'm also not clear on the error path, I'd like to understand how/when we're getting into the insert code after dropping the file, and what change introduced the problem.

agorrod:
> why we can't call __wt_evict_page_clean_update without first calling __evict_review

Evict review is going to stop us evicting internal pages with children, which I believe is necessary, or at least makes me a lot less nervous.

> I'd like to understand how/when we're getting into the insert code after dropping the file,

It is a split that is the problem, the call stack where we are setting the page modified flag on a dead tree is:

```
#2  0x0000000001ba7717 in __wt_abort (session=session@entry=0x2dd9500) at src/third_party/wiredtiger/src/os_posix/os_abort.c:25
#3  0x0000000001bdf233 in __wt_assert (session=session@entry=0x2dd9500, error=error@entry=0,
    file_name=file_name@entry=0x1d28e38 "src/third_party/wiredtiger/src/include/btree.i", line_number=line_number@entry=333, fmt=fmt@entry=0x1c60626 "%s")
    at src/third_party/wiredtiger/src/support/err.c:470
#4  0x0000000001b3cc5d in __wt_page_only_modify_set (page=0x12c30da0, session=0x2dd9500) at src/third_party/wiredtiger/src/include/btree.i:333
#5  __wt_page_modify_set (page=0x12c30da0, session=0x2dd9500) at src/third_party/wiredtiger/src/include/btree.i:426
#6  __split_parent (session=session@entry=0x2dd9500, ref=0x0, ref@entry=0x133c2f00, ref_new=ref_new@entry=0x0, new_entries=new_entries@entry=0,
    parent_incr=296, parent_incr@entry=0, exclusive=exclusive@entry=false, discard=true) at src/third_party/wiredtiger/src/btree/bt_split.c:918
#7  0x0000000001b3f4a2 in __wt_split_reverse (session=session@entry=0x2dd9500, ref=ref@entry=0x133c2f00)
    at src/third_party/wiredtiger/src/btree/bt_split.c:2016
#8  0x0000000001b8c709 in __evict_delete_ref (session=session@entry=0x2dd9500, ref=ref@entry=0x133c2f00, closing=closing@entry=false)
    at src/third_party/wiredtiger/src/evict/evict_page.c:184
#9  0x0000000001b8c8ff in __wt_evict_page_clean_update (session=0x2dd9500, ref=0x133c2f00, closing=<optimized out>)
    at src/third_party/wiredtiger/src/evict/evict_page.c:224
#10 0x0000000001b8d058 in __wt_evict (session=session@entry=0x2dd9500, ref=ref@entry=0x133c2f00, closing=closing@entry=false)
    at src/third_party/wiredtiger/src/evict/evict_page.c:115
#11 0x0000000001b897fa in __evict_page (session=session@entry=0x2dd9500, is_server=is_server@entry=false)
    at src/third_party/wiredtiger/src/evict/evict_lru.c:1467
#12 0x0000000001b89aea in __evict_lru_pages (is_server=<optimized out>, session=<optimized out>) at src/third_party/wiredtiger/src/evict/evict_lru.c:818
#13 __evict_worker (arg=0x29cd6a0) at src/third_party/wiredtiger/src/evict/evict_lru.c:434
```

I'm not entirely sure that the problem here is dirtying a page in a dead tree, but there are problems with doing splits in a file that we are closing or that is dead. I'd originally hoped to reuse the ``closing`` flag that already gets passed in to ``__wt_evict``, but doing that skips the call to ``__wt_evict_list_clear_page`` in ``__evict_review``, which is necessary - since eviction is walking the file, so may have an eviction point set (another reason why calling ``__evict_review`` is necessary).

keithbostic:
Thanks @agorrod, this looks good to me.

keithbostic:
@agorrod, in `__evict_page()`, there's this code:
```
/*
 * If we are evicting in a dead tree, don't write dirty pages.
 *
 * Force pages clean to keep statistics correct and to let the
 * page-discard function assert that no dirty pages are ever
 * discarded.
 */
if (F_ISSET(btree->dhandle, WT_DHANDLE_DEAD))
```
immediately before calling `__wt_evict()`. Maybe the code is no longer necessary with your changes?

michaelcahill:
How is that safe?  We don't have exclusive access, e.g., the eviction server could be walking the tree, there could be leaf pages underneath, etc.

michaelcahill:
Scratch that, misunderstood...

agorrod:
> Maybe the code is no longer necessary with your changes?

I'll shuffle the code so we don't need that test any more, we still want to mark the page clean for the reasons stated in the comment.

agorrod:

PR #2337, opened 2015-11-25
WT-2241 Use a lock to protect transaction ID allocation.

michaelcahill:
We still need to make sure that transaction IDs are published in the state
table before the current ID is incremented so that snapshot reads don't see
uncommitted updates.  However, a lock simplifies the code and performs better
in testing than the initial fix.

### DIFF
--- a/src/include/txn.h
+++ b/src/include/txn.h
@@ -70,7 +70,7 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_txn_state {
 };
 
 struct __wt_txn_global {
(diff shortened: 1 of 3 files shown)


sueloverso:
@michaelcahill is this ready for review and merge?

sueloverso:
@michaelcahill, assuming this still fixes Mathias' problem, this LGTM.  I ran this branch against the wtperf tests from yesterday and see equal performance.

michaelcahill:
Thanks, @sueloverso -- this has been tested against a variety of MongoDB workloads including the original repro from SERVER-21275.

PR #2338, opened 2015-11-25
WT-2242 dead tree eviction

keithbostic:
@agorrod, @michaelcahill: some changes to prioritize the eviction of dead trees.

This tree is only very lightly tested (especially since we don't have tests that do lots of drops).

I think it's pretty reasonable except for the loop in `__evict_walk`: I didn't want to hold the walk lock too long, so I cycled around the dead tree, without switching out the current data-handle, and that's messy, but I didn't see anything cleaner without restructuring.


src/conn/conn_dhandle.c
```diff
@@ -134,7 +134,17 @@ __conn_dhandle_mark_dead(WT_SESSION_IMPL *session)
...
+
+	/* The handle is dead. */
 	F_SET(session->dhandle, WT_DHANDLE_DEAD);
+
+	/*
+	 * Reset the tree's eviction priority, and the tree is evictable by
+	 * definition.
+	 */
+	__wt_evict_priority_clear(session);
+	F_CLR(S2BT(session), WT_BTREE_NO_EVICTION);
```
agorrod: Resetting the no eviction flag is only necessary when ``evict_reset`` is returned from exclusive on. Otherwise we'll start considering pages in eviction when other code assumes the eviction server won't interfere.

src/evict/evict_lru.c
```diff
@@ -27,24 +27,31 @@ static int  __evict_server_work(WT_SESSION_IMPL *);
...
+
 	/* Never prioritize empty slots. */
 	if (entry->ref == NULL)
 		return (UINT64_MAX);
 
 	page = entry->ref->page;
 
+	/* Any page from a dead tree is a great choice. */
+	if (F_ISSET(btree->dhandle, WT_DHANDLE_DEAD))
+		return (WT_READGEN_OLDEST);
```
agorrod: I don't know if I'm that enthusiastic. How about limiting this to leaf pages? It's wasted effort if there are live children.

src/evict/evict_lru.c
```diff
@@ -1575,6 +1586,26 @@ __wt_cache_eviction_worker(WT_SESSION_IMPL *session, bool busy, u_int pct_full)
...
+	S2BT(session)->evict_priority = v;
+}
+
+/*
+ * __wt_evict_priority_clear --
+ *	Clear a tree's eviction priority.
+ */
+void
+__wt_evict_priority_clear(WT_SESSION_IMPL *session)
+{
```
agorrod: What does clearing the priority mean? Presumably if we have a priority system, every handle should start with a given priority. I think some defined values here would make the priority more obvious. Something like:

```
#define WT_BTREE_EVICT_PRIORITY_DEFAULT 5
#define WT_BTREE_EVICT_PRIORITY_MAX 10
#define WT_BTREE_EVICT_PRIORITY_MIN 0
```

The any call you have to __wt_evict_priority_clear would become ``__wt_evict_priority_set(WT_BTREE_EVICT_PRIORITY_DEFAULT)``.

src/evict/evict_lru.c
```diff
@@ -1012,7 +1019,7 @@ retry:	while (slot < max_entries && ret == 0) {
 				cache->evict_file_next = NULL;
 			else
 				dhandle = TAILQ_FIRST(&conn->dhqh);
-		} else {
+		} else if (!F_ISSET(dhandle, WT_DHANDLE_DEAD) || dead_next) {
```
agorrod: Is it possible that focusing on a single file will starve out evictions from other files, specifically meaning eviction doesn't see pages marked as ``WT_READGEN_OLDEST`` for longer?

src/conn/conn_dhandle.c
keithbostic: I'm not following this one: if `WT_BTREE_NO_EVICTION` is set on the file does nothing at all and returns `evict_reset=false`, which means a dead tree with `WT_BTREE_NO_EVICTION` set won't be considered for eviction by the eviction server threads?

src/evict/evict_lru.c
keithbostic: We could reset the read-generation to `WT_READGEN_OLDEST` and then do the usual internal-page skew?

src/evict/evict_lru.c
keithbostic: The only "priority" we have at the moment is `(1<<20)`, set for bloom filters and internal pages.

I agree we could use something better, but that may be outside of the scope of this change; all this change does is re-implement what we have, with slightly better information hiding, and resetting the eviction priority to 0 for dead trees. (That's necessary because a dead tree with a non-zero eviction priority won't be considered for eviction unless we get "aggressive.)

src/evict/evict_lru.c
keithbostic: Isn't that OK? Isn't a page marked `WT_READGEN_OLDEST` potentially more useful a page from a dead tree? (Unless `WT_READGEN_OLDEST` is code for "forcible eviction"?)

agorrod: > Unless WT_READGEN_OLDEST is code for "forcible eviction"?

That is what I was worried about. We use the ``__wt_page_evict_soon`` to try to get a page evicted before it causes problems - I'm concerned focusing eviction too heavily on a single file will limit how well that works.

src/evict/evict_lru.c
agorrod: My vote would be to add a better defined scheme. I hadn't seen that the btree priority factors in to the per-page read generation. Given that - the numbers in my scheme aren't the right ones to choose.

keithbostic: > My vote would be to add a better defined scheme.

I agree, and that's probably worth a separate ticket and some general design/discussion.

src/evict/evict_lru.c
keithbostic: @agorrod, on second thought, I've changed this to be limited to leaf pages. (And this is another reason `__evict_walk` can't focus exclusively on dead trees, too.)

src/evict/evict_lru.c
keithbostic: > We use the __wt_page_evict_soon to try to get a page evicted before it causes problems - I'm concerned focusing eviction too heavily on a single file will limit how well that works.

@agorrod, I agree, that's a bigger change than I want to make and I've reverted this change.

src/evict/evict_lru.c
```diff
@@ -1174,15 +1185,17 @@ __evict_walk_file(WT_SESSION_IMPL *session, u_int *slotp)
...
-	walk_flags = WT_READ_CACHE | WT_READ_NO_EVICT |
-	    WT_READ_NO_GEN | WT_READ_NO_WAIT;
+	start = cache->evict_queue + *slotp;
+	end = start + WT_EVICT_WALK_PER_FILE;
+	if (F_ISSET(session->dhandle, WT_DHANDLE_DEAD) ||
+	    end > cache->evict_queue + cache->evict_slots)
+		end = cache->evict_queue + cache->evict_slots;
 
+	walk_flags =
+	    WT_READ_CACHE | WT_READ_NO_EVICT | WT_READ_NO_GEN | WT_READ_NO_WAIT;
```
keithbostic: @agorrod, since `evict_walk()` is no longer staying in the dead file, I made a slightly different change that means we'll take up to `WT_EVICT_MAX_PER_FILE` eviction candidates from dead trees instead of `WT_EVICT_WALK_PER_FILE` . Does that make sense to you?

src/conn/conn_dhandle.c
agorrod: I think there is a bug in ``__wt_evict_file_exclusive_on`` - it should clear the ``WT_BTREE_NO_EVICTION`` flag before returning an error. Otherwise the current behavior is expected - sometimes we want to clear ``WT_BTREE_NO_EVICTION`` so that eviction keeps looking at the file and sometimes we are planning to flush the file ourselves and need eviction to be locked out.

keithbostic: I agree, and I've created JIRA ticket WT-2259 (pull request #2359) to fix it.

agorrod: I think my last comment was unclear. Apart from the bug (thanks for pushing a fix), I believe the change is still incorrect. i.e: The following was intended to be a separate statement:

the current behavior is expected - sometimes we want to clear WT_BTREE_NO_EVICTION so that eviction keeps looking at the file and sometimes we are planning to flush the file ourselves and need eviction to be locked out.

keithbostic: @agorrod
> the current behavior is expected - sometimes we want to clear WT_BTREE_NO_EVICTION so that eviction keeps looking at the file and sometimes we are planning to flush the file ourselves and need eviction to be locked out.

First, I'm sorry, but I still don't understand your point.

There are four cases where we use`__wt_evict_file_exclusive_on/__wt_evict_file_exclusive_off` pairs:

*  `__sync_file: WT_SYNC_CHECKPOINT`: doesn't want to lock out eviction, just doing an on/off pair to drain the queue as part of an eviction state change,
* `__conn_dhandle_mark_dead`: doesn't want to lock out eviction, doing an off/off pair to drain the queue as part of an eviction state change,
* `__sweep_expire_one`: does want to lock out eviction,
* `__wt_evict_file`: does want to lock out eviction

The code looks correct to me, those functions call `__wt_evict_file_exclusive_on`, then call `__wt_evict_file_exclusive_off` to clear `WT_BTREE_NO_EVICTION` once the exclusive operation is complete.

Second, I don't think this branch changes the previous behavior, so any problem you're seeing is probably in develop as well.

keithbostic:
autobuild: retest this please

src/conn/conn_dhandle.c
```diff
@@ -134,7 +134,17 @@ __conn_dhandle_mark_dead(WT_SESSION_IMPL *session)
...
+
 	if (evict_reset)
 		__wt_evict_file_exclusive_off(session);
```
michaelcahill: I don't follow all of the above, but maybe we just make the call to `__wt_evict_file_exclusive_off` here unconditional on `evict_reset`, along with a comment.  That seems better than clearing `NO_EVICTION` a couple of lines above, then conditionally doing it again in this call.

michaelcahill:
This branch needs a merge with develop.

I guess after that we should merge -- I think part of the reason this has taken so long is that I'm not aware of any evidence that eviction in dead trees is a problem.  Dead trees only stay in the system until sweep kicks them out, which is usually fast.  For these changes to have an impact on applications, we'd need some case where sweep is struggling to keep up, and I don't know of any workloads where sweep is the bottleneck.

src/conn/conn_dhandle.c
keithbostic: @michaelcahill 
> but maybe we just make the call to __wt_evict_file_exclusive_off here unconditional on evict_reset, along with a comment.

This code was reshuffled as part of your changes to ef9927e, where you moved the `__wt_evict_file_exclusive_on/off` calls, so the two actions are no longer adjacent.

keithbostic:
@michaelcahill 

I've merged the trees, and once it passes tests, I'll merge it in.

I don't feel strongly about these changes, they're just clean up and a little better layering in a couple of places. I do think they're worth taking and will make it less likely we'll get in trouble in the future.

PR #2340, opened 2015-11-26
WT-2244 - Trigger in-memory splits sooner.

michaelcahill:
Specifically, do an in-memory split when we hit 80% of memory_page_max.


michaelcahill:
These changes make a big impact to stalls during checkpoints (SEVER-21652).  We seem to generally be able to make progress with far less interruption from checkpoints.

Perf tests here: https://evergreen.mongodb.com/version/5657e3163ff122395f00000a_0 (though some of the results are difficult to interpret).

michaelcahill:
This looks *much* better in testing: eviction working much less hard to keep cache usage under control...

PR #2341, opened 2015-11-27
SERVER-21553 Enable fast-path truncate after splits.

michaelcahill:
A truncate operation attempts to mark leaf pages deleted without reading them
into cache.  One of the conditions that has to be met for that fast-path
truncate of pages is that the leaf page not contain overflow items (or we would
need to read it in order to delete the overflow items).

The "no overflow" flag was not being preseved across internal page splits, so
recent changes to splits were defeating fast-path truncation.

Add tracking of the "no overflow" flag for in-memory page addresses so
fast-path truncates work after internal pages are split.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -352,6 +352,7 @@ __split_ref_move(WT_SESSION_IMPL *session, WT_PAGE *from_home,
 			break;
 		WT_ILLEGAL_VALUE(session);
 		}
+		ref->addr = addr;
 	}
 
 	/* And finally, copy the WT_REF pointer itself. */
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -340,9 +340,18 @@ __split_ref_move(WT_SESSION_IMPL *session, WT_PAGE *from_home,
 			return (ret);
 		}
 		addr->size = (uint8_t)unpack.size;
-		addr->type =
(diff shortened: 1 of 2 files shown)


michaelcahill:

@keithbostic, likewise, does this look right to you?  What we were seeing was *no* fast path truncation in oplog workloads...

src/btree/bt_split.c
```diff
@@ -340,9 +340,18 @@ __split_ref_move(WT_SESSION_IMPL *session, WT_PAGE *from_home,
 			return (ret);
 		}
 		addr->size = (uint8_t)unpack.size;
-		addr->type =
-		    unpack.raw == WT_CELL_ADDR_INT ? WT_ADDR_INT : WT_ADDR_LEAF;
-		ref->addr = addr;
```
daveh86: I think this line is need? I tested against the truncate workload for WT-2240 and it started segfaulting. Adding this line back seems to resolve.

michaelcahill: Oh, right -- yes, that was a typo, only meant to touch `addr->type`.  I'll revert that line...

agorrod:
Thanks @michaelcahill. This lgtm and makes fast path truncates work much better in the workload I've been running (see https://jira.mongodb.org/browse/WT-2240).

keithbostic:
@michaelcahill, this looks right to me as well.

PR #2343, opened 2015-11-28
WT-2247: variable-length column-store in-memory page splits.

keithbostic:
@michaelcahill, @agorrod, this change supports in-memory page splits for variable-length column-store trees.

I branched off WT-2244 (#2340) because there was stuff there that would collide, but that was just for convenience, if WT-2244 goes in before we open up develop more generally, I'll rebase the pull request.


michaelcahill:
Replaced by #2345, against develop.

PR #2344, opened 2015-11-28
WT-2248: WT_SESSION.close is updating WT_CONNECTION_IMPL.default_session

keithbostic:
@michaelcahill, can you do a quick sanity check on this one?

I think it's safe (I'm using the same trick as in `__conn_close`), but wanted to be sure.

The repro is watching `WT_SESSION.name` change as we work our way through `__wt_connection_close`.


michaelcahill:
Thanks @keithbostic, lgtm.  I'd noticed the `session.name` oddness but hadn't tracked it down.

PR #2345, opened 2015-11-30
WT-2247, WT-2312: variable-length column-store in-memory page splits

michaelcahill:
Replaces #2343.


keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill, this is a real failure, there's a problem with column-store splits that needs addressing.

If you append rows 100-1000 to a page, and split that page in-memory, the split code moves record 1000 to a new page, creating a new top-level WT_REF with starting record 1000 (and a new page with the WT_INSERT structure for record 1000). The problem is if record 999 isn't yet transactionally visible.

Imagine record 998 is transactionally visible, 998 is not visible, 1000 is visible. In the reconciliation code we write record 998, then a deleted record for 999 and then record 1000. If we've moved record 1000, we don't know to write record 999 on that page, and when we verify the tree we find a missing record (the last record we've seen in the tree is 998, and suddenly there's an internal page referencing record 1000).

We could:

1. Change the split page's append list by adding a WT_INSERT entry causing any missing records to be written. This is going to be messy, it's hard to figure out what records might be missing, especially since the missing records change based on the current transaction state.
2. Add a new value to the split page's modification information that makes the page responsible for an additional chunk of the column-store name space when it's written, and look for that value when reconciling the page.
3. Don't split column-store pages in-memory unless there's no gap in the name space for the record we're about to move: the next-to-last entry on the split page's append list has to be the record number immediately before the one we're moving, and it has to be globally visible. It's a bit painful to get to the next-to-last entry on a skiplist, and we're going to spend time waiting for that entry to be globally visible...

Absent a better idea, I guess I'll go with door number 2. I wanted to review this with you, in case there are other issues in this area I'm not thinking about, or there's a better idea.

All of this, of course, implies deleted records can magically appear at the end of the column-store tree when transactions are aborted, but that's already the case.

Thoughts?

Here's the CONFIG:
```
#  RUN PARAMETERS
abort=0
auto_throttle=1
backups=0
bitcnt=2
bloom=1
bloom_bit_count=50
bloom_hash_count=9
bloom_oldest=0
cache=32
checkpoints=1
checksum=uncompressed
chunk_size=6
compaction=0
compression=none
data_extend=0
data_source=table
delete_pct=0
dictionary=0
encryption=rotn-7
evict_max=1
file_type=var
firstfit=0
huffman_key=1
huffman_value=1
in_memory=0
insert_pct=30
internal_key_truncation=1
internal_page_max=11
isolation=read-committed
key_gap=8
key_max=57
key_min=18
leaf_page_max=14
leak_memory=0
logging=0
logging_archive=1
logging_compression=none
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=4
merge_max=9
mmap=1
ops=100000
prefix_compression=1
prefix_compression_min=7
repeat_data_pct=0
reverse=0
rows=10000
runs=100
salvage=1
split_pct=77
statistics=0
statistics_server=0
threads=4
timer=20
value_max=2260
value_min=0
verify=1
wiredtiger_config=
write_pct=23
```

michaelcahill:
@keithbostic, option 2 makes sense to me.

To make sure I understand, column store pages currently either have a fixed key range when the page is created, or are the last page in the tree with an append list.  Now that the last page can transition to not being at the end by still having an append list, we need to know the key range it finishes at.

One question: can we infer the last record number during reconciliation by looking in the parent page at `pindex[slot + 1]`?

keithbostic:
@michaelcahill:

> To make sure I understand, column store pages currently either have a fixed key range when the page is created, or are the last page in the tree with an append list. Now that the last page can transition to not being at the end by still having an append list, we need to know the key range it finishes at.

Yes.

> One question: can we infer the last record number during reconciliation by looking in the parent page at pindex[slot + 1]?

I wondered about that, too, but splits/eviction scared me, I don't think we're guaranteed the page we're reconciling isn't the last entry in its parent's page-index, in which case we'd have to climb the tree and find the "next" page in the tree. We could do that, but it felt messy. Let me know if I'm over-complicating this, I'd much rather take the value from the "next" page than maintain a separate value in the page-modify structure.

keithbostic:
Current state of the world: as of b8e992b, the CONFIG listed above passes for both fixed- and variable-length objects, but I think there's still a lurking problem, I sometimes see a failure, but only after several hundred runs.

src/btree/bt_split.c
```diff
@@ -1643,9 +1645,11 @@ __split_insert(WT_SESSION_IMPL *session, WT_REF *ref)
 	F_SET_ATOMIC(page, WT_PAGE_SPLIT_INSERT);
 
 	/* Find the last item on the page. */
-	ins_head = page->pg_row_entries == 0 ?
+	ins_head = type == WT_PAGE_ROW_LEAF ?
```
agorrod: I really dislike nested ternery statements. Can this be untangled?

keithbostic: > I really dislike nested ternery statements. Can this be untangled?

Sure, done.

agorrod:
@keithbostic Do you have any thoughts about how we might be able to track down the remaining failure? Can we construct a test case that is going to be more likely to trigger a failure? Would it help to put in more aggressive error checking code in diagnostic mode?

keithbostic:
@agorrod, no ideas yet -- the symptom is an internal page with the wrong starting record number (which is the bug I thought the last set of changes fixed). Anyway, I'll get back to this one and chase it down next chance I get.

keithbostic:
@agorrod, found/fixed the problem in column-store splits, commit is 70ef762.

It's a problem in the develop branch, it just wasn't ever triggering without column-store splits.

keithbostic:
New failure:
```
#  RUN PARAMETERS
abort=0
auto_throttle=1
backups=0
bitcnt=5
bloom=1
bloom_bit_count=57
bloom_hash_count=29
bloom_oldest=0
cache=250
checkpoints=0
checksum=uncompressed
chunk_size=7
compaction=0
compression=none
data_extend=0
data_source=table
delete_pct=17
dictionary=1
direct_io=0
encryption=none
evict_max=3
firstfit=0
huffman_key=0
huffman_value=0
in_memory=1
insert_pct=39
internal_key_truncation=1
internal_page_max=17
isolation=snapshot
key_gap=6
key_max=83
key_min=28
leaf_page_max=17
leak_memory=0
logging=0
logging_archive=0
logging_compression=zlib
logging_prealloc=1
long_running_txn=0
lsm_worker_threads=4
merge_max=16
mmap=1
ops=100000
prefix_compression=1
prefix_compression_min=8
repeat_data_pct=18
reverse=0
rows=100000
runs=100
salvage=0
split_pct=83
statistics=0
statistics_server=1
threads=1
timer=20
value_max=128
value_min=15
verify=0
wiredtiger_config=
write_pct=35
```

keithbostic:
@agorrod, found/fixed the problem in fixed-length column-store splits, commit is fe9ec20.

keithbostic:
@agorrod, @michaelcahill: this branch has now passed a few stress runs, and I think it's appropriate to consider it for merging again.

I also believe this branch needs relatively careful review: there are significant changes in the split code, the cursor-prev code, and the reconciliation code. There aren't many changes by line count, but they're in sensitive places.

src/btree/bt_curprev.c
```diff
@@ -156,27 +166,26 @@ __cursor_fix_append_prev(WT_CURSOR_BTREE *cbt, bool newpage)
...
+		 * If not, check if there are any records on the page. If there
+		 * aren't, then we're in the magic zone, keep going until we get
+		 * to a record number matching the first record on the page.
 		 */
-		if (cbt->ins != NULL &&
-		    cbt->recno <= WT_INSERT_RECNO(cbt->ins))
-			WT_RET(__cursor_skip_prev(cbt));
 		if (cbt->ins == NULL &&
-		    (cbt->recno == 1 || __col_fix_last_recno(page) != 0))
+		    (cbt->recno == page->pg_fix_recno ||
```
michaelcahill: The change looks right to me.  Why haven't test/format prev tests hit this?

src/btree/bt_split.c
```diff
@@ -1720,49 +1721,89 @@ __split_insert(WT_SESSION_IMPL *session, WT_REF *ref)
...
-	parent_incr += sizeof(WT_REF) + sizeof(WT_IKEY) + key->size;
-	__wt_scr_free(session, &key);
+		child->key.recno = ref->key.recno;
+
+	/*
+	 * Don't copy any deleted page state: we may be splitting a page that
+	 * was instantiated after a truncate and that history should not be
+	 * carried onto these new child pages.
+	 */
+	child->page_del = NULL;
```
michaelcahill: How could this be set?  We start with zeroed memory and only set the fields we want.

I understand wanting the comment, maybe it can go near line 1716?

src/btree/bt_split.c
```diff
@@ -1786,15 +1827,19 @@ __split_insert(WT_SESSION_IMPL *session, WT_REF *ref)
...
 	 * can be ignored.)
 	 */
-	right->pg_row_ins[0]->head[0] =
-	    right->pg_row_ins[0]->tail[0] = moved_ins;
+	if (type == WT_PAGE_ROW_LEAF)
+		right->pg_row_ins[0]->head[0] =
+		    right->pg_row_ins[0]->tail[0] = moved_ins;
+	else
+		right->modify->mod_append[0]->head[0] =
+		    right->modify->mod_append[0]->tail[0] = moved_ins;
```
michaelcahill: This makes my head hurt.  Can we set a `WT_INSERT_HEAD` local variable, so this becomes something like:

```c
inshead = (type == WT_PAGE_ROW_LEAF) ?
    right->pg_row_ins[0] : right->modify->mod_append[0];
inshead->head[0] = inshead->tail[0] = moved_ins;
```

test/format/config.h
```diff
@@ -54,7 +54,7 @@ typedef struct {
 	uint32_t	maxrand;	/* Maximum value randomly chosen */
 	uint32_t	maxset;		/* Maximum value explicitly set */
 	uint32_t	*v;		/* Value for this run */
-	char		**vstr;		/* Value for string options */
+	const char	**vstr;		/* Value for string options */
```
michaelcahill: Is there some benefit to this?  We allocate the memory, so we can write to it -- the only effect I see of this change is that we need to cast to `(void *)` in order to free...

test/format/util.c
```diff
@@ -50,7 +50,7 @@ dmalloc(size_t len)
  * dstrdup --
  *	Call strdup, dying on failure.
  */
-static char *
+void *
 dstrdup(const char *str)
```
michaelcahill: Why not return `char *`?

michaelcahill:
@keithbostic, I've made some comments but basically this lgtm.  It would be good to get it merged tomorrow if possible.

src/btree/bt_curprev.c
keithbostic: @michaelcahill 
> The change looks right to me. Why haven't test/format prev tests hit this?

Before we started splitting column-store pages, the first page (recno == 1), was the only place this could ever happen, so the simple test was sufficient.

src/btree/bt_split.c
keithbostic: @michaelcahill 
> How could this be set? We start with zeroed memory and only set the fields we want.
I understand wanting the comment, maybe it can go near line 1716?

Agreed, fixed.

src/btree/bt_split.c
keithbostic: @michaelcahill 

> Can we set a WT_INSERT_HEAD local variable,

Sure, done.

test/format/config.h
keithbostic: @michaelcahill 

> Is there some benefit to this? We allocate the memory, so we can write to it -- the only effect I see of this change is that we need to cast to (void *) in order to free...

Agreed, fixed.

test/format/util.c
keithbostic: @michaelcahill 

> Why not return char *?

Agreed, fixed.

PR #2346, opened 2015-11-30
WT-2249 Keep eviction stuck until cache usage is under 100%.

michaelcahill:
Fix a real bug: STUCK is a cache flag, not a cache->state flag.

Don't try to do eviction when dumping the cache, and only kill processes if no
pages have been evicted since the STUCK flag was first set.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -159,7 +159,8 @@ __evict_server(void *arg)
 	WT_DECL_RET;
 	WT_SESSION_IMPL *session;
 #ifdef HAVE_DIAGNOSTIC
-	struct timespec now, stuck_ts = { 0, 0 };
+	struct timespec now, stuck_ts;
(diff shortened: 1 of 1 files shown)


src/evict/evict_lru.c
```diff
@@ -1460,10 +1469,6 @@ __evict_page(WT_SESSION_IMPL *session, bool is_server)
 
 	WT_RET(ret);
 
-	cache = S2C(session)->cache;
-	if (F_ISSET(cache, WT_CACHE_STUCK))
-		F_CLR(cache, WT_CACHE_STUCK);
-
 	return (ret);
```
keithbostic: @michaelcahill, we can probably remove the `WT_RET` call.

michaelcahill: Agreed, thanks @keithbostic.

PR #2347, opened 2015-11-30
WT-2250 Minor fix.  Use SET instead of INCRV for stat.

sueloverso:
@keithbostic Here's the minor change for the logging stat.  Please review and merge when develop is open again.


keithbostic:
lgtm, it's so small a change, I just merged it.

PR #2348, opened 2015-11-30
WT-2142 test cleanup

ddanderson:
WT-2142.  At the end of a test run, all open connections should be closed to prevent cascading errors. Clean up and (mostly) standardize how connection configuration is added to and modified.


ddanderson:
@agorrod or @michaelcahill, this is ready for review.  Thanks!

test/suite/test_async01.py
```diff
@@ -132,18 +132,9 @@ class test_async01(wttest.WiredTigerTestCase, suite_subprocess):
         ('table-row', dict(tablekind='row',uri='table')),
     ])
 
-    # Overrides WiredTigerTestCase so that we can configure
-    # async operations.
-    def setUpConnectionOpen(self, dir):
-        self.home = dir
-        conn_params = \
-                'create,error_prefix="%s: ",' % self.shortid() + \
+    conn_config = lambda self, dir: \
```
agorrod: @ddanderson I really like the simplifications this change provides, but I don't like Python lambda functions (mostly because I encounter them rarely enough that I need to go read documentation). Is there a different solution to setting up the conn_config?

ddanderson:
@agorrod, Sorry for the delay (and thanks for the ping).  I've eliminated the lambda expressions.  Can you take another look?

agorrod:
Thanks @ddanderson lgtm.

PR #2351, opened 2015-12-01
WT-2553: prioritize WT_READGEN_OLDEST pages for eviction.

keithbostic:


michaelcahill:
Thanks, @keithbostic, lgtm and helps with SERVER-21553.

PR #2352, opened 2015-12-02
SERVER-21553 Free blocks during reverse splits.

michaelcahill:


src/evict/evict_page.c
```diff
@@ -248,7 +248,7 @@ __evict_page_dirty_update(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 	switch (mod->rec_result) {
 	case WT_PM_REC_EMPTY:				/* Page is empty */
 		/* Discard the parent's address. */
-		__wt_ref_free_addr(session, ref);
+		__wt_ref_addr_free(session, ref);
```
michaelcahill: Should this be `__wt_ref_block_free`?  If not, how do we end up here with `ref->addr != NULL`?

src/evict/evict_page.c
```diff
@@ -303,7 +303,7 @@ __evict_page_dirty_update(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 			return (EBUSY);
 
 		/* Discard the parent's address. */
-		__wt_ref_free_addr(session, ref);
+		__wt_ref_addr_free(session, ref);
```
michaelcahill: Same as above: should this be `__wt_ref_block_free`?

daveh86:
I've run this against the workload in WT-2240, and it has made a very large improvement.

File size is stable with that workload at around 250MB after 5 mins of running. Without patch file-size is upwards of 4GB.

src/reconcile/rec_write.c
```diff
@@ -1390,13 +1371,9 @@ __rec_child_deleted(WT_SESSION_IMPL *session,
...
-	if (ref->addr != NULL &&
-	    (page_del == NULL ||
-	    __wt_txn_visible_all(session, page_del->txnid))) {
-		WT_RET(__wt_ref_info(session, ref, &addr, &addr_size, NULL));
-		WT_RET(__rec_block_free(session, addr, addr_size));
-		__wt_ref_free_addr(session, ref);
-	}
+	if (page_del == NULL ||
+	    __wt_txn_visible_all(session, page_del->txnid))
+		WT_RET(__wt_ref_block_free(session, ref));
```
keithbostic: @michaelcahill, I'd put the test for `ref->addr != NULL` back in: if you don't the comment above it no longer makes sense, and this code parallels the code immediately below, which tests `ref->addr` again. Else: we rewrite the comment to explain how `ref->addr` is cleared by `__wt_ref_block_free`?

michaelcahill: Sure, done.

src/evict/evict_page.c
keithbostic: I agree, this isn't necessary.

src/evict/evict_page.c
keithbostic: I think you're correct (nice catch!).

The code that frees the block is:
```
static int
__rec_write_wrapup(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
{
        ...
        switch (mod->rec_result) {
        case 0: /*
                 * The page has never been reconciled before, free the original
                 * address blocks (if any).  The "if any" is for empty trees
                 * created when a new tree is opened or previously deleted pages
                 * instantiated in memory.
                 *
                 * The exception is root pages are never tracked or free'd, they
                 * are checkpoints, and must be explicitly dropped.
                 */
                if (__wt_ref_is_root(ref))
                        break;
                WT_RET(__wt_ref_block_free(session, ref));
                break;
```
and it should happen on the first reconciliation, from which point on, `ref->addr` should stay NULL. I can't think of any reason the path to `WT_PM_REC_MULTIBLOCK` is any different from the two cases you noted, and it doesn't free `ref->addr`, which lends weight to the argument.

It's a tricky change, though: for now, we could assert at the top of `__evict_page_dirty_update` that `ref->addr == NULL`?

keithbostic:
@michaelcahill, looks good to me.

src/evict/evict_page.c
michaelcahill: Thanks, @keithbostic -- that's what I thought (that `ref->addr` should always be `NULL`), otherwise we could be leaking blocks there (which we clearly aren't, or verify would find them).

I've made the change you suggested: assert `ref->addr == NULL` at the top, remove these calls.  Nothing has fallen over yet.

agorrod:

PR #2353, opened 2015-12-02
WT-2553 Evict pages left behind by in-memory splits.

michaelcahill:
Include pages over the in-memory splits threshold when in the "would block" phase of eviction.


agorrod:
Thanks @michaelcahill This lgtm - evicting dirty pages as they are split makes sense to me.

PR #2354, opened 2015-12-02
SERVER-21553 3.0 backport

agorrod:
Also includes fixes for:
SERVER-21568 - use after free
WT-2251 - addr structure leak

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -1030,6 +1030,16 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref,
 #endif
 
 	/*
+	 * Page-delete information is only read when the WT_REF state is
+	 * WT_REF_DELETED.  The page-delete memory wasn't added to the
+	 * parent's footprint, ignore it here.
+	 */
+	if (ref->page_del != NULL) {
+		__wt_free(session, ref->page_del->update_list);
+		__wt_free(session, ref->page_del);
+	}
+
+	/*
 	 * Reset the page's original WT_REF field to split.  Threads cursoring
 	 * through the tree were blocked because that WT_REF state was set to
 	 * locked.  This update changes the locked state to split, unblocking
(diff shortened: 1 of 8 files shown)


agorrod:
I don't think the latest failure is a result of this patch. test/fops dropped core with the following error:

``
Program terminated with signal 11, Segmentation fault.
#0  __session_discard_dhandle (session=0x1236540, dhandle_cache=0x0) at ../src/session/session_dhandle.c:50
50		bucket = dhandle_cache->dhandle->name_hash % WT_HASH_ARRAY_SIZE;
Missing separate debuginfos, use: debuginfo-install glibc-2.17-55.140.amzn1.x86_64
(gdb) where
#0  __session_discard_dhandle (session=0x1236540, dhandle_cache=0x0) at ../src/session/session_dhandle.c:50
#1  0x00007fac0466607d in __wt_session_get_btree (session=0x1236540, uri=0x4024ba "file:wt", checkpoint=0x0, cfg=0x7fac0239ee70, flags=264)
    at ../src/session/session_dhandle.c:425
#2  0x00007fac046666aa in __wt_session_get_btree_ckpt (session=0x1236540, uri=0x4024ba "file:wt", cfg=0x7fac0239ee70, flags=264)
    at ../src/session/session_dhandle.c:277
#3  0x00007fac0460bbe3 in __wt_curfile_open (session=0x1236540, uri=<optimized out>, owner=0x0, cfg=0x7fac0239ee70, cursorp=0x7fac0239ee88)
    at ../src/cursor/cur_file.c:516
#4  0x00007fac04664456 in __wt_open_cursor (session=0x1236540, uri=<optimized out>, owner=0x0, cfg=0x7fac0239ee70, cursorp=0x7fac0239ee88)
    at ../src/session/session_api.c:274
#5  0x00007fac04664860 in __session_open_cursor (wt_session=0x1236540, uri=0x4024ba "file:wt", to_dup=0x0, config=<optimized out>,
    cursorp=0x7fac0239eed8) at ../src/session/session_api.c:350
#6  0x000000000040109d in obj_bulk () at ../../../test/fops/file.c:49
#7  0x0000000000401851 in fop (arg=<optimized out>) at ../../../test/fops/fops.c:135
#8  0x00007fac03f6bdf3 in start_thread () from /lib64/libpthread.so.0
``

It isn't easy to reproduce.

agorrod:
autobuild: retest this please

src/btree/bt_split.c
```diff
@@ -1030,6 +1030,16 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref,
 #endif
 
 	/*
+	 * Page-delete information is only read when the WT_REF state is
+	 * WT_REF_DELETED.  The page-delete memory wasn't added to the
+	 * parent's footprint, ignore it here.
+	 */
+	if (ref->page_del != NULL) {
```
agorrod: @michaelcahill This is the only part of the back port I'm unsure about. On develop this free and the subsequent ``WT_PUBLISH`` are wrapped in an ``if (discard)`` which would be equivalent to ``if (closing)`` on this branch.

I know that's because split_parent is used in different ways on develop, so I *think* this is the right approach, but wanted to call it out to make sure you agree.

michaelcahill: Thanks, @agorrod.  I think this is correct.  We are replacing or discarding `ref`.  Either way, we should free `page_del` if there is one still attached.

michaelcahill:
I agree about that test failure: I think it's unrelated.  This LGTM, merging.

PR #2355, opened 2015-12-02
WT-2113 Retry updating the oldest id if we race.

sueloverso:
@michaelcahill Please review and we should chat about this fix for the EBUSY return during truncate01.  This fix passes all of my repro cases and I have confirmed that it hit the race during those runs.


sueloverso:
I briefly considered adding a stat to count these races, but decided it wasn't all that interesting.  Happy to add one if you want.

michaelcahill:
Thanks, @sueloverso, I've stared at this for a while and it lgtm, so I'll merge.

I'd like to revisit this code at some point: I think it can be simplified, and I also think we should investigate a RW lock instead of `scan_count`.

PR #2356, opened 2015-12-02
WT-2257 Fixes when given multiple thread workload configs.

sueloverso:
@daveh86 or @agorrod please review this fix for the issues regarding multiple "thread" workload configs.


daveh86:
This LGTM

agorrod:
@sueloverso This change seems reasonable to me. Could you update the documentation as well, since I find it confusing. The comment in ``wtperf_opt.i`` currently states:

``multiple workload configurations may be specified``

Which I find confusing - I'm not sure if it was intended to mean thread groups, or workload command lines.

PR #2357, opened 2015-12-02
WT-2256 Fix interval timer for wtperf throttling.

sueloverso:
@daveh86 or @agorrod Please review this change that fixes throttling.  We need to set the timer to the post-sleep time.  This is what was causing the nearly 2x number of ops.


daveh86:
This LGTM

agorrod:
Thanks @sueloverso lgtm.

PR #2358, opened 2015-12-02
WT-2258: ignore pre-load on direct I/O files.

keithbostic:
@michaelcahill, @agorrod, this is a bug fix to ignore pre-loading on any file configured for direct I/O.


agorrod:

PR #2359, opened 2015-12-03
WT-2259: fix __wt_evict_file_exclusive_on function error handling

keithbostic:
@agorrod, I agree with your comment, here's the change for your review.


agorrod:
Thanks @keithbostic lgtm - I'll merge

PR #2360, opened 2015-12-04
WT-2265 - increase mongodb-oplog wtperf workload throttle to compensate for fix

daveh86:
@agorrod, can you have a quick look at this change.

It fixes the regression we have seen in the WTPERF truncate workload in jenkins noticed by @sueloverso in https://jira.mongodb.org/browse/WT-2245

### DIFF
--- a/bench/wtperf/runners/mongodb-oplog.wtperf
+++ b/bench/wtperf/runners/mongodb-oplog.wtperf
@@ -8,4 +8,4 @@ run_time=500
 populate_threads=1
 # Setup three threads to insert into the oplog
 # Setup one thread to be doing truncates from the oplog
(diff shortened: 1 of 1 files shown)


daveh86:
The following are the list of wtperf runners that have a throttle value in them:
- long-txn-btree.wtperf
- long-txn-lsm.wtperf
- mongodb-large-oplog.wtperf
- mongodb-oplog.wtperf
- mongodb-small-oplog.wtperf
- multi-btree-long.wtperf
- truncate-btree-workload.wtperf

Of these 7 only one is run in our regular perf scripts and is the subject of this Pull. The others are not used regularly in our suites so I suspect can remain as they are, given we are not using them as a benchmark.

The two long-txn one workloads are deliberately running far lower numbers for their throttle to make a reader txn that lasts for 20 seconds, so I think these would be better left as is.

agorrod:

PR #2361, opened 2015-12-04
WT-2260 Avoid adding internal pages to the eviction queue.

agorrod:
We added code to make eviction fairer when there are lots of files
open, but it had the effect that when there are only a few files
open eviction could add lots of internal pages to the eviction
queue, since it focused on a small area of the tree.


agorrod:
@michaelcahill Would you mind reviewing this change?

Could you please let me know if you'd like to hold off merging for a while. I'm interested to see if it alters the Jenkins performance jobs, so I'll update those to point to this branch if it won't be merged soon.

michaelcahill:
Thanks, @agorrod, the performance results look good, I'll merge.

PR #2363, opened 2015-12-04
WT-2262 Have random sampling walk the tree so it isn't biased in skewed trees.

michaelcahill:


keithbostic:
@agorrod, @michaelcahill, I've added a new configuration, `next_random_sample_percent ` that specifies a percentage of the tree to skip over before retrieving each new record, and reworked the code so we use the original algorithm of selecting from the entire tree on each retrieval if that new configuration isn't specified.

Alex, I think this one is ready for merge after your final review.

michaelcahill:
Sorry we didn't talk this through in detail, I expected the parameter to be called "sample_size" and to be a count of the number of samples that the application will perform (which might be 1000, or 0.1%).

keithbostic:
@agorrod, @michaelcahill: I've pushed the changes Michael wanted.

dist/api_data.py
```diff
@@ -814,21 +814,19 @@ def __cmp__(self, other):
...
+        use the WT_CURSOR::insert method to insert rows.  When
+        bulk-loading, rows must be loaded in sorted order.  The value
+        is usually a true/false flag; when bulk-loading fixed-length
+        column store objects, the special value \c bitmap allows chunks
+        of a memory resident bitmap to be loaded directly into a file
+        by passing a \c WT_ITEM to WT_CURSOR::set_value where the \c
+        size field indicates the number of records in the bitmap (as
+        specified by the object's \c value_format configuration).
+        Bulk-loaded bitmap values must end on a byte boundary relative
+        to the bit count (except for the last set of values loaded)'''),
```
agorrod: Is this just line wrapping changes?

agorrod:
@keithbostic I'm happy with this change, I made a couple of mostly style based changes. Merge if you are happy with them, or I'll be happy to talk them through in your evening.

src/btree/bt_cursor.c
```diff
@@ -844,11 +850,63 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
...
+		 * entirely wrong, as it depends on how many pages are in this
+		 * particular checkpoint, how large the leaf and internal pages
+		 * really are, and other factors). Then, divide that value by
+		 * the configured sample size and increment the final result to
+		 * make sure tiny files don't leave us with a skip value of 0.
+		 *
+		 * !!!
+		 * Ideally, the number would be prime to avoid restart issues.
+		 */
+		if (cbt->next_random_sample_size != 0)
```
agorrod: @keithbostic I changed your version because I preferred the idea of duplicating a condition check than using a goto.

dist/api_data.py
keithbostic: > Is this just line wrapping changes?

No; the bulk cursor docs said "... and cursors configured for bulk-load only support the WT_CURSOR::insert and WT_CURSOR::close methods.", which isn't true (they also support reconfigure and reset).

Rather than try and enumerate the supported methods, I changed it to mention WT_CURSOR::insert as the "primary" method applications will use.

I noticed because the "next_random" wording had the same problem.

keithbostic:
@agorrod, I think this one is done, at least as far as develop goes, I'm merging it.

Do we have any sample unbalanced trees around we can use to test this fix in 3.0?

PR #2364, opened 2015-12-04
WT-2264 Don't have write leaves chase inserts indefinitely,

michaelcahill:
Cut off inserts that are more recent than when write leaves starts.

This is just the first step.  The next part is to stop checkpoints stalling.


agorrod:
@michaelcahill This change looks good to me. Should I merge, or would you like to add work to fix stalling checkpoints to this branch as well?

michaelcahill:
Let's keep this branch open for now and I will reproduce the checkpoint stalls then see if we can come up with a fix.

michaelcahill:
@agorrod, I tried to reproduce the checkpoint dropouts with the tip of MongoDB master and this branch, using the workload in SERVER-21553.  These changes don't seem to cause any problems for that workload, so I am inclined to merge them into WiredTiger develop and see where we end up.

Can you please review?

agorrod:

PR #2366, opened 2015-12-04
WT-2060: minor statistics cleanup.

keithbostic:


keithbostic:
@agorrod, @michaelcahill, this simplifies some of the existing statistics code, and with it I plan to close WT-2060.

I can't think of any way to "fix" the second item in WT-2060 JIRA ticket: in summary, we have a bunch of statistics we don't maintain in the statistics structures, we maintain them somewhere else, and then simply set them in the statistics fields whenever the application makes a statistics call.

The only idea I had would be to add a new `dist/stat_data.py`field, "single", that implies:

1. there won't be any increment/decrement calls in the source code to update that statistic,
2. when we update that statistic, we'll simply set slot 0, we won't clear the other slots,
3. we won't do any aggregation when we read that statistic, we'll simply read slot 0,

But, "single" would just be a convention so it's not clear doing that is all that much better than what we have now. Maybe it would be a little less fragile?

Let me know if you like adding "single", or have a better idea, otherwise ready for review/merge.

agorrod:

PR #2368, opened 2015-12-04
WT-2246: faster column appends

keithbostic:
Ready for review/merge.


src/btree/col_srch.c
```diff
@@ -120,7 +120,17 @@ restart_page:	page = current->page;
...
-	cbt->compare = 0;
+
+	/* 
+	 * Don't bother searching if the caller is appending a new record where
+	 * we'll allocate the record number; we're not going to find a match by
+	 * definition, and we figure out the record number and position when we
+	 * do the work.
+	 */
+	if (recno == UINT64_MAX) {
+		cbt->compare = -1;
```
agorrod: The code below used to setup cbt->slot and cbt->recno in this case. I'm not confident that isn't necessary - can you reassure me?

keithbostic:
@agorrod, I've reworked this one so we don't use UINT64_MAX to flag a `WT_CURSOR.append` search, and I think that means the handling of `cbt->slot` and `cbt->recno` is returned to its previous state, that is, any legitimate search (including a search for record number UINT64_MAX) will set `cbt->slot` and `cbt->recno` as before.

I also wrote some basic CRUD tests of record number UINT64_MAX, which required a couple of bug fixes.

michaelcahill:

PR #2369, opened 2015-12-05
WT-2262: Have random sampling walk the tree so it isn't biased in skewed trees.

keithbostic:
@michaelcahill, some changes for your consideration.


keithbostic:
@michaelcahill, there are a couple of additional questions I have about this code (whichever version you end up using, the questions apply to both).

1. Using `file-size / allocation-size` to estimate the page count.

    I was thinking maybe we could get closer with `btree->bm->block.live.ckpt_size`, the byte size of the last checkpoint? That might be 0, of course, and we could fall back to the file size in that case.

    Also, the allocation-size is likely to be relatively small to the real page size. I was thinking we could use the return of `__wt_split_page_size()`, which calculates the split-page size, that is, the target for page splitting.

2. Calling `__wt_btcur_prev()` if `__wt_btcur_next()` fails.

    It's been a few years, but I'm pretty sure there was a reason we don't do that in the search-near code. There's commit dff10b7 where we switched from calling `__wt_btcur_prev` to calling `__wt_btcur_last`, but that didn't work out, so there's commit a34f1e6, where we switched to repeating the search, we didn't switch back to calling `__wt_btcur_prev`. I vaguely recall it was something about finding intermediate records during the backward walk that weren't there during the forward walk? Or maybe it was just slow when walking over deleted-but-instantiated column-store records? Anyway, I wanted to raise the point, in case there's something there to worry about.

src/btree/bt_cursor.c
```diff
@@ -842,49 +840,69 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
...
+		 *
+		 * Pages read for data sampling aren't "useful"; don't update
+		 * the read generation of pages already in memory, and if a page
+		 * is read, set its generation to a low value so it is evicted
+		 * quickly.
+		 */
+		for (skip = cbt->rand_leaf_skip; cbt->ref == NULL || skip > 0;)
+			WT_ERR(__wt_tree_walk(session, &cbt->ref, &skip,
+			    WT_READ_NO_GEN | WT_READ_SKIP_INTL |
+			    WT_READ_SKIP_LEAF | WT_READ_WONT_NEED));
```
michaelcahill: @keithbostic, this makes no sense to me as it stands -- skipping both leaf and internal pages should return nothing, but you've made it skip some number of leaf pages instead.  If we're going to push this logic into `__wt_walk_tree`, the flag should be called something other than `WT_READ_SKIP_LEAF` -- maybe `WT_READ_SAMPLE`?

src/btree/bt_walk.c
```diff
@@ -304,10 +325,25 @@ __wt_tree_walk(WT_SESSION_IMPL *session,
...
+			 * the next leaf page we can; be cautious around the
+			 * page decrement: if for some reason we end up not
+			 * taking this particular page, we can take the next
+			 * one, and, there are additional tests/decrements when
+			 * we're about to return a leaf page.
+			 */
+			if (LF_ISSET(WT_READ_SKIP_LEAF)) {
+				WT_ERR(__ref_is_leaf(session, ref, &isleaf));
+				if (isleaf && *walkcntp > 0) {
+					--*walkcntp;
```
michaelcahill: I'm not a fan of in/out parameters if we can avoid it.  A separate "skip_count" (input) parameter would be better.

michaelcahill:
@keithbostic, about using `file size / allocation size` to estimate the count: yes, it's an overestimate, and I think it has to be.  If we choose anything larger than `allocation size` as the denominator and the file is made up of `allocation size` blocks, we won't sample the whole file and the bug won't be fixed.  We're only walking internal pages so I don't think overestimating is that big a deal.

The more interesting question to me is the API -- I'm expecting to add another parameter to set the sample size (which I assumed to be 100 for the calculation).  If that isn't set (i.e., we don't know a sample size), then I think we should keep the existing behavior.

> Calling __wt_btcur_prev() if __wt_btcur_next() fails.

I didn't change that, it could certainly be slow if there are lots of tombstones, but `next_random` is not used as frequently as `search_near` and I don't know of any issue with that code so I am inclined to leave it alone unless we have a workload that demonstrates a problem.  If we do anything, I'd probably be inclined to call `search_near` with whatever key we land on.

src/btree/bt_cursor.c
keithbostic: @michaelcahill: my inclination is to not rename as `WT_READ_SAMPLE`, if at a future date we want to skip leaves for other reasons, that will end up being hard to work with.

We could add a parameter that set the number of leaves to skip (and it would be nice it if had an out-of-band value that caused all leaves to be skipped).

src/btree/bt_walk.c
keithbostic: OK: it would be nice to have an out-of-band value if we want to skip all of the leaf pages.

src/btree/bt_cursor.c
agorrod: @keithbostic re: ``WT_READ_SKIP_LEAF`` when I see ``skip`` in any page related code I immediately think of skip lists, not of skipping some proportion of pages during a walk. I'd prefer if this flag had a more alarming name.

keithbostic: @agorrod, `WT_READ_SKIP_LEAF` matches `WT_READ_SKIP_INTL`, though.

We could use `WT_READ_IGNORE_LEAF` and `WT_READ_IGNORE_INTL`?

I honestly don't care a lot, I'm OK with whatever you and @michaelcahill prefer.

michaelcahill:
@keithbostic, I'm not sure what's going on with this pull request (it looks like things went bad after develop was merged here but not the base branch).  Can you please extract the patch and apply it to the WT-2262 branch?

Let's follow up there on what `__wt_tree_walk` should look like: I don't think we got to the end of that discussion.

keithbostic:
@michaelcahill, git got sad because I merged develop, not the base branch. I merged the base branch and it's all OK, AFAICT.

I pushed the change to use a new variable instead of overloading `walkcntp`, everybody seemed in favor of that. I also made a minor change so you can skip all the leaf pages in the tree by not passing in a count of leaf pages to skip.

I agree it's a kluge `__wt_tree_walk` takes a count of leaf pages to skip, obviously I thought it was better than splitting the work between `__wt_btcur_next_random` and `__wt_tree_walk`. That said, I'm good to go with whatever you decide: you can probably cherry-pick a59e9dc, 56a82ac and 5f0820c, those are all fixes to pass the test suite before I changed the change's structure.

@agorrod suggested renaming `WT_READ_SKIP_LEAF` (and, presumably `WT_READ_SKIP_INTL`),  I suggested `WT_READ_IGNORE_XXX`. Fine by me, either way.

michaelcahill:
@keithbostic, I'm going to merge this so we're down to one branch.

I think callers should either set `WT_READ_SKIP_LEAF` *or* pass in a non-zero sample count.  Doing both makes no sense to me, because I think of the 'SKIP' flag as meaning "never return leaf pages".

keithbostic:
> I think callers should either set WT_READ_SKIP_LEAF or pass in a non-zero sample count. Doing both makes no sense to me, because I think of the 'SKIP' flag as meaning "never return leaf pages".

I really disagree: it makes more sense to me to have a "skip leaf pages" flag that's modified by a count, than to have two different ways to skip leaf pages, especially given that in the world with two different ways to skip leaf pages, we'd have both flags and arguments that configure types of pages the tree-walk skips, not just flags alone.

If you're sure, just let me know and I'll make the change.

michaelcahill:
I'm sure: `SKIP_|NTL` returns no internal pages.  If we have a `SKIP_LEAF`, it should skip all leaf pages.  Setting a non-zero sample count should determine whether we are sampling leaf pages.

PR #2370, opened 2015-12-07
Cut WiredTiger 2.7.0 release

agorrod:


michaelcahill:
Thanks, @agorrod, I think we're good to go.  I'll merge, then we can cut the final release from develop.

PR #2371, opened 2015-12-07
WT-2235. Changes to wt printlog to add -x option and remove -p.

ddanderson:
wt printlog -x produces extra "-hex" fields when appropriate. For example,
when printing a "row_put" entry, in addition to "key" and "value" (printed
as JSON strings with Unicode escapes), there will also be "key-hex" and
"key-value" entries in the JSON dictionary, containing a hex dump for each
of these fields.

For consistency with existing WT outputs, all hex output is in lower case.
One visible change is that our Unicode escapes in JSON, also used for
'wt dump -j', now produce hex characters in lower case ('a'-'f'). Before,
'wt dump -j' produced upper case hex (using 'A'-'F').

As part of testing, I took outputs from test_txn08 and validated them using jsonlint.com .


ddanderson:
@sueloverso, can you review please?

src/cursor/cur_json.c
```diff
@@ -8,6 +8,8 @@
 
 #include "wt_internal.h"
 
+static const u_char hex[] = "0123456789abcdef";
+
```
sueloverso: This is now the 7th place in the code we declare `static const ... hex`.  They're not all the same either, some are `static const char` and some are `static const u_char`.  It would be nice to have a single one.

src/docs/command-line.dox
```diff
@@ -260,7 +260,7 @@ Display the database log.
 The \c printlog command outputs the database log.
 
 @subsection util_printlog_synopsis Synopsis
-<code>wt [-rVv] [-C config] [-E secretkey ] [-h directory] printlog [-p] [-f output]</code>
+<code>wt [-rVv] [-C config] [-E secretkey ] [-h directory] printlog [-x] [-f output]</code>
```
sueloverso: Not directly related to your change, but I just noticed that the `-rVv` should be `-RVv`.  Can you fix for all the places in there it is wrong please?

sueloverso:
Thanks Don.  LGTM.  I made a couple minor comments.

ddanderson:
@sueloverso this is ready. I separated out the sharing of the hex array (in 3 places) into its own commit. I don't think we can portably share data items from WT shared library to applications.

PR #2372, opened 2015-12-08
WT-2229 Add compression skipping to log compression.

sueloverso:
@keithbostic Here are changes to generalize the tracking functions and add it into logging for your consideration.


keithbostic:
@sue, lgtm, merged.

PR #2375, opened 2015-12-09
WT-2177 Add an optional random seed to WiredTiger rand implementation

michaelcahill:
From @fedorova.

Added a function to optionally initialize a seed, used by __wt_random(), from a random number. Added a call to this function to wtperf. Please take a look.


agorrod:
Thanks @fedorova lgtm - I'll merge.

PR #2376, opened 2015-12-09
WT-2275: core dump when "wt dump" fails.

keithbostic:
@michaelcahill, @agorrod: two problems. First, we weren't closing a cursor in a table-cursor open error path (commit 2cc1025), and second, we were looking at `WT_CURSOR.uri` in a cursor close path and it hadn't yet been initialized (commit 5196a17).

Please check the table-cursor open change carefully, I'm not that comfortable with the table cursor code, I might have gotten it wrong.

Alex, commit 5196a17 is because `WT_CURSOR.uri` is the last thing we fill in when initializing a cursor, and we were using it inside WiredTiger cursor close code. What would you think of disallowing the use of `WT_CURSOR.uri` inside the WiredTiger library, and only allow using `WT_CURSOR.internal_uri`?

### DIFF
--- a/src/cursor/cur_table.c
+++ b/src/cursor/cur_table.c
@@ -968,8 +968,11 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 	WT_ERR(__wt_strdup(session, tmp->data, &ctable->cfg[1]));
 
 	if (0) {
-err:		WT_TRET(__curtable_close(cursor));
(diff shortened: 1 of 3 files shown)


michaelcahill:

Note that I don't think this fixes the underlying issue in WT-2275 (unreadable file when crashing during/after truncate), so the ticket should stay open for now.

PR #2378, opened 2015-12-10
WT-2269 - Add ability to write config dump

daveh86:


daveh86:

bench/wtperf/wtperf.c
```diff
@@ -2308,6 +2313,31 @@ main(int argc, char *argv[])
...
+	}
+
+	snprintf(path, req_len + 14, "%s/CONFIGBACKUP", cfg->home);
+	if ((fp = fopen(path, "w")) == NULL) {
+		lprintf(cfg, errno, 0, "%s", path);
+		goto err;
+	}
+
+	/* Print the config dump */
+	fp = fopen(path, "w+");
```
sueloverso: The file is already open.  You can remove this 2nd call to fopen.

bench/wtperf/wtperf.c
```diff
@@ -2308,6 +2313,31 @@ main(int argc, char *argv[])
...
+	while (!TAILQ_EMPTY(&cfg->config_head)) {
+		config_line = TAILQ_FIRST(&cfg->config_head);
+		TAILQ_REMOVE(&cfg->config_head, config_line, c);
+		fprintf(fp, "%s\n", config_line->string);
+		free(config_line->string);
+		free(config_line);
+	}
+	free(path);
+	fclose(fp);
```
sueloverso: I'd prefer if all of this code was in its own function, {{config_print}} and called, similar to {{config_sanity()}} rather than mainlined in {{main()}}.  I suggest {{config_print}} because that is the name of the function used in test/format.

bench/wtperf/wtperf.c
```diff
@@ -2308,6 +2313,31 @@ main(int argc, char *argv[])
...
+
 	/* Display the configuration. */
 	if (cfg->verbose > 1)
 		config_print(cfg);
```
sueloverso: Oops!  Looks like we already have {{config_print}}.  Can we combine the functionality in there or avoid similar code?

daveh86:
@agorrod and @sueloverso, would you mind taking another look over this?

bench/wtperf/wtperf.c
```diff
@@ -2308,6 +2311,10 @@ main(int argc, char *argv[])
 	if ((ret = config_sanity(cfg)) != 0)
 		goto err;
 
+	/* Save a copy of the config. */
```
agorrod: I'd prefer for this to be "write" rather than "save"

bench/wtperf/config.c
```diff
@@ -653,6 +666,43 @@ config_sanity(CONFIG *cfg)
...
+	char *path;
+
+	/* Backup the config */
+	req_len = strlen(cfg->home) + 100;
+	if ((path = calloc(req_len, 1)) == NULL) {
+		(void)enomem(cfg);
+		return;
+	}
+
+	snprintf(path, req_len + 14, "%s/CONFIGBACKUP", cfg->home);
```
agorrod: test/format calls the equivalent file ``CONFIG``, how about calling this one ``CONFIG.wtperf``

bench/wtperf/wtperf.c
```diff
@@ -2308,6 +2311,10 @@ main(int argc, char *argv[])
...
+	if (cfg->verbose > 0)
```
agorrod: I'd write this out regardless since it's going to a file not to stdout.

bench/wtperf/config.c
```diff
@@ -653,6 +666,43 @@ config_sanity(CONFIG *cfg)
...
+		lprintf(cfg, errno, 0, "%s", path);
+		return;
+	}
+
+	/* Print the config dump */
+	while (!TAILQ_EMPTY(&cfg->config_head)) {
+		config_line = TAILQ_FIRST(&cfg->config_head);
+		TAILQ_REMOVE(&cfg->config_head, config_line, c);
+		fprintf(fp, "%s\n", config_line->string);
+		free(config_line->string);
```
agorrod: There is a shared function called ``config_free`` that is a better place for the cleanup. That way it will be cleaned up even if we don't take this code path, and is more consistent with the rest of the config structure.

dist/s_string.ok
```diff
@@ -52,6 +52,7 @@ CKPT
 CMP
 CONCAT
 CONFIG
+CONFIGBACKUP
```
sueloverso: This change can go away now that the file name changed.

bench/wtperf/config.c
```diff
@@ -653,6 +674,45 @@ config_sanity(CONFIG *cfg)
...
+	}
+err:
+	free(path);
+	if (fp != NULL)
+		(void)fclose(fp);
+}
+
+/*
  * config_print --
  *	Print out the configuration in verbose mode.
```
sueloverso: @daveh86 and @agorrod which `config_print` is not part of these changes, it is related for consideration.  That function is only called when verbose > 1 and is an textual style printing of the config.  However, the function isn't really complete AFAICT as there are lots of other config options not printed here.  (I've personally never used it so I don't know if anyone does or how useful they find it.)  Does the new code to write out the config make this obsolete?  If it is useful, it should probably be brought up to date.

bench/wtperf/config.c
```diff
@@ -571,14 +579,27 @@ config_opt_line(CONFIG *cfg, const char *optstr)
...
 	}
 
+	/* Append the current line to our copy of the config. */
+	if ((string_copy = calloc(len + 1, 1)) == NULL) {
+		return (enomem(cfg));
+	}
+	config_line = calloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
+	strncpy(string_copy, optstr, len);
+	config_line->string = string_copy;
+	TAILQ_INSERT_TAIL(&cfg->config_head, config_line, c);
```
sueloverso: Please add a comment that the config is simply in the order the user had in the wtperf config file and then any override configs on the command line via `-o`.  There is no sorting or reordering so overriding lines will not replace nor appear near the original ones.

sueloverso:
@daveh86 and @agorrod it occurs to me that while this is a big step forward in recreating and knowing the config file that ran.  One omission I thought of, otherwise this is nearly ready:

This only writes out the lines that exist explicitly in the config files or command line.  If something takes the default, and we change the default at some point, then rerunning may behave differently.  Admittedly we very rarely change defaults.  But it is something to be aware of when saving.  I don't now how easy it would be to have a function to create a line with all the defaults at the head of the list, then the config file overrides, then command line overrides so that it is complete (similar to the way test/format dumps a complete CONFIG).

bench/wtperf/config.c
daveh86: I don't personally believe that this makes the output obsolete. 

I find the output useful when running tests as I can see the config options and startup params on screen just prior to the test starting.

Updating it seems worthwhile for a second ticket.

daveh86:
@agorrod and @sueloverso would you mind having a further look at this one?

I've added the dedupe logic.

bench/wtperf/config.c
```diff
@@ -571,14 +579,32 @@ config_opt_line(CONFIG *cfg, const char *optstr)
 {
 	WT_CONFIG_ITEM k, v;
 	WT_CONFIG_PARSER *scan;
+	CONFIG_QUEUE_ENTRY *config_line;
```
sueloverso: Please list this first in the local vars, alphabetically.

bench/wtperf/config.c
```diff
@@ -571,14 +579,32 @@ config_opt_line(CONFIG *cfg, const char *optstr)
...
 
+	/*
+	 * Append the current line to our copy of the config. The config is
+	 * stored in the order it is processed, so added options will be after
+	 * any parsed from the original config.
+	 */
+
+	if ((string_copy = calloc(len + 1, 1)) == NULL) {
+		return (enomem(cfg));
+	}
```
sueloverso: Remove {}

bench/wtperf/config.c
```diff
@@ -652,6 +678,72 @@ config_sanity(CONFIG *cfg)
 	return (0);
 }
 
+void
+config_queue_dedupe(CONFIG *cfg)
```
sueloverso: 1.  The function should have a comment.
2.  Please call it `config_consolidate` as `dedupe` is not as descriptive.

bench/wtperf/config.c
```diff
@@ -652,6 +678,72 @@ config_sanity(CONFIG *cfg)
...
+{
+	CONFIG_QUEUE_ENTRY *conf_line, *test_line, *tmp;
+	char *string_key;
+
+	conf_line = TAILQ_FIRST(&cfg->config_head);
+	while (conf_line != NULL) {
+		string_key = strchr(conf_line->string, '=');
+		tmp = test_line = TAILQ_NEXT(conf_line, c);
+		while (test_line != NULL) {
```
sueloverso: This loop and how it works is worthy of a good long comment.

bench/wtperf/config.c
```diff
@@ -652,6 +678,72 @@ config_sanity(CONFIG *cfg)
...
+			if (strncmp(conf_line->string, test_line->string,
+			    string_key - conf_line->string + 1) == 0) {
```
sueloverso: Can you verify that this works correctly when we have a config string that is a proper substring of another in all directions?  The only ones I know of at the moment is `max_latency` and `max_latency_fatal` and `min_throughput` and `min_throughput_fatal`.  That is, make sure having multiple ones of the shorter string and the longer string are consolidated properly.  I'm pretty sure the `+1` in the length forces it to match the `=` too.  That is a good comment to add.

bench/wtperf/config.c
```diff
@@ -652,6 +678,72 @@ config_sanity(CONFIG *cfg)
...
+
+	snprintf(path, req_len + 14, "%s/CONFIG.wtperf", cfg->home);
+	if ((fp = fopen(path, "w")) == NULL) {
+		lprintf(cfg, errno, 0, "%s", path);
+		goto err;
+	}
+
+	/* Print the config dump */
+	fprintf(fp, "# Warning. This config can include defaults.\n");
+	fprintf(fp, "# This make cause differences in behaviour.\n");
```
sueloverso: How about:
# Warning: This config includes unwritten, implicit configuration defaults.
# Changes to those values may cause differences in behavior.

daveh86:
Here is the test config I have been using.
```
conn_config="cache_size=6GB,checkpoint=(wait=100000),statistics=(fast,clear),statistics_log=(wait=5)"
table_config="type=file"
icount=1000000
report_interval=5
run_time=90
populate_threads=1
table_count=1
table_count_idle=200
table_count_idle=10
threads=((count=2,reads=1,ops_per_txn=100),(count=3,inserts=1),(count=1,truncate=1,truncate_pct=10,truncate_count=1000000))
```

And here is the result:
```
# Warning. This config includes unwritten, implicit configuration defaults.
# Changes to those values may cause differences in behavior.
icount=1000000
report_interval=5
run_time=90
populate_threads=1
table_count=1
table_count_idle=10
threads=((count=2,reads=1,ops_per_txn=100),(count=3,inserts=1),(count=1,truncate=1,truncate_pct=10,truncate_count=1000000))
verbose=2
conn_config="create,cache_size=6GB,checkpoint=(wait=100000),statistics=(fast,clear),statistics_log=(wait=5),"
table_config="key_format=S,value_format=S,type=lsm,exclusive=true,allocation_size=4kb,internal_page_max=64kb,leaf_page_max=4kb,split_pct=100,type=file,"
```

Note how the `conn_config` line is different, but only one is printed. Additionally, note how only one `table_count_idle` is included, if we were over-matching we would not get both returned as we would be comparing `table_count_idle=2` and `table_count_idle=1`. Along with this we are not under-scanning as we get both the `table_count` and `table_count_idle` variables.

daveh86:
@sueloverso, would you mind taking another look over this? I hope this is the final version

sueloverso:
Thanks @daveh86.  Merging.

PR #2380, opened 2015-12-10
WT-2282 Remove last_running_moved from verbose conditional.

sueloverso:
@michaelcahill Here's a simple fix but you should review it.  I removed the ``last_running_moved`` part of the conditional.  Or is the "pinned" verbose message more subtle than that and it needs to be moved or copied up into the other part of the new "want to update" part?


keithbostic:
@sueloverso, I'm going to go ahead and pull this into develop (Coverity is complaining, too), and this can't break anything

PR #2381, opened 2015-12-10
WT-2283 Add yield to avoid tight looping.

sueloverso:
@michaelcahill This fixes the hang I was seeing in my test/format run.  It hung 100% of the time without this fix and has passed/completed several iterations with it.  However it may be a band-aid like fix as the {{scan_count}} usage is taking the place of read/write lock.


agorrod:
@sueloverso I know test/format isn't a good tool for measuring this, but could you try the same configuration with the update oldest from before the fix for WT-2113?

sueloverso:
@agorrod with this fix, this branch runs at about 90-95% of the speed of the same configuration from before the fix for WT-2113.  Without the fix, test/format never finished for me.

PR #2382, opened 2015-12-10
WT-2099 Account for overflow update structures

sueloverso:
that get appended to the update list in page->memory_footprint.

@michaelcahill Please review the fix we discussed last night.  I have run this fix with the failing configuration and made sure we took that path but did not get underflow.


michaelcahill:

PR #2383, opened 2015-12-10
WT-2231: check the leaf page's parent keys before doing the full binary search 

keithbostic:
This is a replay of the changes for the previous WT-2231 branch (see pull request #2325) for the individual commit details. There's an additional bug fix over and above those commits, fixing a clang sanitizer error.

In summary: check the leaf page's parent keys before doing the full binary search of the leaf page, avoiding the binary search entirely when the cursor is being re-positioned, at the cost of two additional searches when the cursor is not being re-positioned. Additionally, do some work to improve WT_REF page hints.


keithbostic:
@michaelcahill, for some reason, the clang assertion doesn't fire for me, but I think I figured out what it was from the stack. It might be worth a couple of runs on the machine where the failure happened, if that's easy to do.

keithbostic:
@michaelcahill, @agorrod: please ignore this branch for now -- there's a bug I'm going to try and fix by refactoring the code.

src/btree/row_srch.c
```diff
@@ -132,6 +132,67 @@ __wt_search_insert(
...
+	/*
+	 * Check if the search key is less than the parent's starting key for
+	 * this page.
+	 */
+	__wt_ref_key(leaf->home, leaf, &item->data, &item->size);
+	WT_RET(__wt_compare(session, collator, srch_key, item, &cmp));
+	if (cmp < 0) {
+		cbt->compare = 1;		/* page keys > search key */
+		return (0);
+	}
```
michaelcahill: I don't feel strongly, but could we just return `WT_NOTFOUND` instead to indicate that the key is out of range?

It seems odd that we're carefully setting `cbt->compare` differently for the two cases, but the calling code only cares about `cbt->compare != 0`...

keithbostic: I agree -- for some reason I've been thinking a fine-grained understanding of why we failed might be useful in the future, but I doubt that's really true. Might as well just return a boolean (since the column-store version of this function can't fail, so it simplifies it a lot).

keithbostic: > I don't feel strongly, but could we just return WT_NOTFOUND instead to indicate that the key is out of range?

On second thought... I remember why I did it this way. In short, the callers of the search function look at `cbt->compare` to decide if the search was successful, so I have to set it to something that's non-zero so the callers know the search failed. Setting it to some incorrect value is scary, so I took the trouble to set it to the "correct" non-zero value.

keithbostic:
@agorrod, @michaelcahill, I think this branch is ready for final review/merge.

src/btree/col_srch.c
```diff
@@ -9,6 +9,53 @@
...
+	 *
+	 * !!!
+	 * Check that "indx + 1" is a valid page-index entry first, because it
+	 * also checks that "indx" is a valid page-index entry, and we have to
+	 * do that latter check before looking at the indx slot of the array
+	 * for a match to leaf (in other words, our page hint might be wrong).
+	 */
+	WT_INTL_INDEX_GET(session, leaf->home, pindex);
+	indx = leaf->pindex_hint;
+	if (indx + 1 < pindex->entries && pindex->index[indx] == leaf)
```
agorrod: Should the test in the first clause be ``<=``? I thought the ``pindex_hint`` field was zero-based.

agorrod:
@keithbostic I had one question (the consequence of which is we could bail out of the fast path early, I believe so it's not correctness), apart from that this change looks good to me.

src/btree/col_srch.c
keithbostic: > Should the test in the first clause be <=? I thought the pindex_hint field was zero-based.

I think it's correct as written. The idea is we're checking two fields, `indx` and `indx + 1`, and both of them have to be less than `pindex->entries`. Instead of doing two tests, I'm doing one test, for the larger value.

It is zero-based, so if we're going to look at slots 3 and 4, then `pindex->entries` must be at least 5.

Please let me know if this doesn't make sense, and we'll talk it over.

keithbostic:
Thanks, @agorrod, I'll go ahead and merge this then.

src/btree/col_srch.c
agorrod: Thanks, I see.

PR #2384, opened 2015-12-11
WT-2265: Correct PPC64 barriers

markbenvenuto:
Use the cheaper `lwsync` for read and write barriers on PPC64. `lwsync` only allows
store#load re-ordering.

See Book 2, section 4.4.3 in Power ISA 2.07B manual for more details. In figure 5, the manual says that `mbll` and `mbss` map to `lwsync`.

Test: Builds on POWER8 ppc64le, and runs examples.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -144,6 +144,7 @@ INIT
 INITIALIZER
 INMEM
 INTL
+ISA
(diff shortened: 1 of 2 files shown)


keithbostic:
@markbenvenuto, lgtm, merged.

PR #2385, opened 2015-12-11
WT-2285: configure should set BUFFER_ALIGNMENT_DEFAULT to 4kb on linux

markbenvenuto:
A variable name changed as part of ecc674b8f5fc58c1df9a1c6b957491d73a27603d.


keithbostic:
@markbenvenuto, lgtm, merged.

PR #2387, opened 2015-12-11
WT-2276 decode checkpoints via 'wt list -c' and via standalone Python utility.

keithbostic:
@ddanderson, some suggestions for how this change might be structured inside WiredTiger, feel free to discard it if you don't like it better.

Entirely untested (but it builds & lints cleanly).


ddanderson:
@keithbostic, thanks, that's a nice improvement.

PR #2388, opened 2015-12-14
SERVER-21833: fix compaction

keithbostic:
@agorrod, this fixes a problem that's effectively turning off compaction. @sueloverso has a test she's cleaning up for inclusion in the tree, obviously, this one needs to go in immediately.

We can cherry-pick ad33ace if we don't want to merge the whole thing into the MongoDB 3.2 branch.


sueloverso:
@keithbostic I have cleaned up the test.  I'll pull this branch and run it against the new version.

sueloverso:
@keithbostic I tested the new unit test and it passes on this branch.  Please review the test now.  In particular:
1.  The sizes I compare/assert are pretty broad.  Maybe they can be tightened.
2.  Maybe it should not be `test_bug016.py` but `test_compact##.py`.

keithbostic:
autobuild: retest this please

sueloverso:
@keithbostic when I run the test, it fails about 50% of the time both on my Mac and AWS.  The statistics show that the file size is not changing on those runs.  Is there some alternate path or scheduling dependency during those runs?

keithbostic:
@sueloverso, another fix for compaction, we weren't compacting in the case of a reconciliation creating a multi-block replacement.

sueloverso:
@keithbostic I'm going to need to stare at the library changes a bit more to understand them.  They look good on the surface.  Does this explain why it was sometimes failing or is this change fallout from adding more variations to the test?

keithbostic:
It was sometimes failing because whenever eviction kicked in, if the right pages were written, then the test would work/fail. If you increase the cache size so eviction never kicks in, then it behaves consistently.

keithbostic:
These changes  are because, when you increase the cache size, you can end up with multi-block reconciliation of internal pages: this test appends internal pages to the end of the file, so for the test to work, those pages have to be compacted, and this fix is required if the cache size is large.

PR #2391, opened 2015-12-14
WT-2289: fix for the fast key check

keithbostic:
This is a fix for the code added in WT-2231 (#2383).


agorrod:
Thanks @keithbostic lgtm.

PR #2240, opened 2015-10-02
WT-2149 WT-2156 Allow eviction workers to restart.

michaelcahill:

PR #2395, opened 2015-12-16
WT-2256 - Update WTPERF throttling implementation

daveh86:
Currently used for discussion. Not ready for merge


SConstruct
```diff
@@ -459,6 +459,7 @@ t = env.Program("wtperf", [
     "bench/wtperf/track.c",
     "bench/wtperf/wtperf.c",
     "bench/wtperf/wtperf_truncate.c",
+    "bench/wtperf/wtperf_throttle.c",
```
sueloverso: Move up one line to alphabetize.

bench/wtperf/Makefile.am
```diff
@@ -5,7 +5,8 @@ LDADD = $(top_builddir)/libwiredtiger.la -lm
 noinst_PROGRAMS = wtperf
 wtperf_LDFLAGS = -static
 wtperf_SOURCES =\
-	config.c misc.c track.c wtperf.c wtperf_truncate.c wtperf.h wtperf_opt.i
+	config.c misc.c track.c wtperf.c wtperf.h \
+	wtperf_opt.i wtperf_truncate.c wtperf_throttle.c
```
sueloverso: Alphabetize here too.

bench/wtperf/wtperf.c
```diff
@@ -434,10 +433,9 @@ worker(void *arg)
 		}
 	}
 	/* Setup the timer for throttling. */
-	if (thread->workload->throttle != 0 &&
-	    (ret = __wt_epoch(NULL, &interval)) != 0) {
-		lprintf(cfg, ret, 0, "Get time call failed");
-		goto err;
+	if (thread->workload->throttle != 0) {
+		if ((setup_throttle(thread) != 0))
+			goto err;
```
sueloverso: Please simplify to:
```
if (thread->workload->throttle != 0 && setup_throttle(thread) != 0)
    goto err;
```

bench/wtperf/wtperf.c
```diff
@@ -681,13 +679,10 @@ op_err:			if (ret == WT_ROLLBACK && ops_per_txn != 0) {
 			op = thread->workload->ops;
 
 		/*
-		 * Check throttling periodically to avoid taking too
-		 * many time samples.
+		 * Decrement throttle tickets and check if we need to reload
```
sueloverso: End sentence with a period.  Or maybe you meant to type more in this comment?

bench/wtperf/wtperf.h
```diff
@@ -127,6 +127,15 @@ struct __truncate_queue_entry {
...
 typedef struct __truncate_queue_entry TRUNCATE_QUEUE_ENTRY;
 
+/* Steering for the throttle configuration */
+struct __throttle_config {
+	struct timespec last_increment;
+	uint64_t ticket_queue;
+	uint64_t tickets_per_increment;
+	uint64_t usecs_increment;
+};
+typedef struct __throttle_config THROTTLE_CONFIG;
```
sueloverso: Usually we declare them like:
```
typedef struct {
...
} THROTTLE_CONFIG;
```
(also fix the `TRUNCATE_QUEUE_ENTRY` above it)

bench/wtperf/wtperf.h
```diff
@@ -261,7 +271,7 @@ struct __config_thread {		/* Per-thread structure */
 	TRACK truncate;			/* Truncate operations */
 	TRACK truncate_sleep;		/* Truncate sleep operations */
 	TRUNCATE_CONFIG trunc_cfg;	/* Truncate configuration */
-
+	THROTTLE_CONFIG throttle_cfg;	/* Throttle configuration */
```
sueloverso: I'd leave in the blank line, but move both of them above the tracking section.  The structure has blank lines between unrelated fields.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+ * OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include "wtperf.h"
+
+/*
+ * Put the initial config together for running a throttled workload.
+ */
+int
+setup_throttle(CONFIG_THREAD *thread) {
```
sueloverso: Opening { goes on its own line.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+	THROTTLE_CONFIG *throttle_cfg;
```
sueloverso: Add a blank line after local vars.

bench/wtperf/wtperf.h
```diff
@@ -127,6 +127,15 @@ struct __truncate_queue_entry {
 };
 typedef struct __truncate_queue_entry TRUNCATE_QUEUE_ENTRY;
 
+/* Steering for the throttle configuration */
+struct __throttle_config {
+	struct timespec last_increment;
+	uint64_t ticket_queue;
```
sueloverso: Can you call it something other than `_queue`?  Maybe `_count` since that is also what is used in the workload.  To me that means a TAILQ-style implementation.  Also I find `ticket` confusing.  Can you call it `operations` or `ops` instead?  I suggest `ops_count` and `ops_per_increment`.

bench/wtperf/wtperf.h
```diff
@@ -127,6 +127,15 @@ struct __truncate_queue_entry {
...
+	uint64_t tickets_per_increment;
+	uint64_t usecs_increment;
```
sueloverso: Please add a comment on each line describing the field.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+	throttle_cfg = &thread->throttle_cfg;
+
+	/* If the interval is very small, we do laps of 1 */
+	if (thread->workload->throttle < THROTTLE_OPS) {
```
sueloverso: Assign the fields in the same order in each of the conditionals to make it clear each one is setting the same things.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+
+/*
+ * Put the initial config together for running a throttled workload.
+ */
+int
+setup_throttle(CONFIG_THREAD *thread) {
+	THROTTLE_CONFIG *throttle_cfg;
+	throttle_cfg = &thread->throttle_cfg;
+
+	/* If the interval is very small, we do laps of 1 */
```
sueloverso: The comment here should be inside its piece of the conditional.  I'd like a big comment here describing the scenarios below and the approach for each one.  Here's my understanding having gone through a few concrete examples:
- When we throttle to a very small number of operations per second, we do one operation with a longer sub-second period between operations.
- When we throttle to a medium number, we do 100 ops per period and adjust the period accordingly.
- When we throttle to a large number of ops per second, we do a small number of ops much more frequently per second.

I think we don't need the third case and can consolidate to two.

bench/wtperf/wtperf.h
```diff
@@ -190,6 +199,7 @@ struct __config {			/* Configuration structure */
 #define	ELEMENTS(a)	(sizeof(a) / sizeof(a[0]))
 
 #define	THROTTLE_OPS	100
+#define	THROTTLE_INTVL	(USEC_PER_SEC / THROTTLE_OPS)
```
sueloverso: The names of these macros does not make sense to me with their usage.  Both are used in various ways for both operation counts and time-related calculations.  I'd rather see more, clearer macros that may use the same values as others.  However, I think I may have come to the conclusion that `THROTTLE_INTVL` is not needed.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+			USEC_PER_SEC / thread->workload->throttle;
+	} else if (thread->workload->throttle < THROTTLE_INTVL) {
+		throttle_cfg->usecs_increment =
+		    USEC_PER_SEC / thread->workload->throttle * THROTTLE_OPS;
+		throttle_cfg->tickets_per_increment = THROTTLE_OPS;
+	} else {
+		/* If the interval is too large, we do more tickets per batch */
+		throttle_cfg->usecs_increment = THROTTLE_OPS;
+		throttle_cfg->tickets_per_increment =
+		    thread->workload->throttle / THROTTLE_INTVL;
```
sueloverso: AFAICT, the code doesn't match the comment.  With 100000, usecs gets set to 100 and ticket ends up as 10.  I think that what we want to do here is reduce to two situations.  The first is the "very small" case.  The other is keeping the number of operations constant, and adjusting the usecs value.  So:
```
if (thread->workload->throttle < THROTTLE_OPS) {
    /* For very small throttle values, do one operation less frequently. */
    throttle_cfg->tickets_per_increment = 1;
    throttle_cfg->usecs_increment =
        USEC_PER_SEC / thread->workload->throttle;
else {
    /* Do THROTTLE_OPS each time, and adjust the number of times we do them, per second */
    throttle_cfg->tickets_per_increment = THROTTLE_OPS;
    throttle_cfg->usecs_increment =
        USEC_PER_SEC * THROTTLE_OPS  / thread->workload->throttle;
}
```
Hmm, this falls down if the throttle is set to more than 100million.  If we set THROTTLE_OPS to 1000, then we have a window of 1 billion ops per second.  We're probably safe there.  In any case, I think the choices are doing fixing the number of ops per period and adjusting the period, or fixing the period and adjusting the number of ops.  I like adjusting the period because it spreads out the scheduling and sleeping.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+	    throttle_cfg->tickets_per_increment, throttle_cfg->usecs_increment);
+	printf("this means we are performing %llu ops per second\n",
+	    (USEC_PER_SEC / throttle_cfg->usecs_increment) * throttle_cfg->tickets_per_increment); 
+	return (0);
+}
+
+/*
+ * TODO: Spread the stalls out, so we don't flood at the start of each
+ * second and then pause. Doing this every 10th of a second is probably enough
+ */
```
sueloverso: The comment is no longer valid.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+int
+worker_throttle(CONFIG_THREAD *thread) {
+	THROTTLE_CONFIG *throttle_cfg;
+	struct timespec now;
+	uint64_t usecs_delta;
+
+	throttle_cfg = &thread->throttle_cfg;
+
+	if (throttle_cfg->ticket_queue != 0)
+		return (0);
```
sueloverso: This is a per-thread item and we only call `worker_throttle` when `ticket_queue` is 0 so this should be unnecessary.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+
+	WT_RET(__wt_epoch(NULL, &now));
+
+	/*
+	 * If we did enough operations in the current interval, sleep for
+	 * the rest of the interval. Then add more tickets to the queue.
+	 */
+	usecs_delta = WT_TIMEDIFF_US(now, throttle_cfg->last_increment);
+	if (usecs_delta < throttle_cfg->usecs_increment - THROTTLE_OPS) {
```
sueloverso: I don't understand the use of `THROTTLE_OPS` here and how it relates to time calculations.

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,108 @@
...
+		(void)usleep(
+		    (useconds_t)(throttle_cfg->usecs_increment - usecs_delta));
+		throttle_cfg->ticket_queue =
+		     throttle_cfg->tickets_per_increment;
+	} else
+		throttle_cfg->ticket_queue =
+		     (usecs_delta / throttle_cfg->usecs_increment) *
+		     throttle_cfg->tickets_per_increment;
```
sueloverso: Do you want to cast to float for the multiplication?  Unless `usecs_delta` is 2X or more of `usecs_increment` this won't change the value because the integer division will give us 1.

A comment would hep make it clear what adjustments you're making in the else clause.
As an example, say throttle is 5000, giving us `tickets_per_increment` of 100 and `usecs_increment` of 20000.  If it takes 10% longer, or `usecs_delta` is 22000 what do we want in `ticket_queue`?  Even with float of 1.1 in this example, the calculation would make `ticket_queue` larger which seems wrong.  If it took longer than our increment to do 100, it will take even longer to do 110 ops.  Did you mean to divide:
```
(uint64_t) (float)throttle_cfg->tickets_per_increment / (float)(usecs_delta / throttle_cfg->usecs_increment)
```
so that 100/1.1 would give us a new `ticket_queue` of 90?  I'm not clear what value you want there.

bench/wtperf/wtperf_throttle.c
daveh86: I'd done some experimenting along these lines initially. The problem I found was that the timing becomes less accurate when we are doing very few operations per run through the loop. 

For example, if I cut down and merge the two bottom clauses I get the following output when I attempt to do 1000 reads per second over 2 threads and 247600 inserts per second over 3 threads.
```
9294 reads, 3368119 inserts, 0 updates, 34 truncates, 0 checkpoints in 5 secs (5 total secs)
9317 reads, 3371651 inserts, 0 updates, 34 truncates, 0 checkpoints in 5 secs (10 total secs)
9330 reads, 3377288 inserts, 0 updates, 33 truncates, 0 checkpoints in 5 secs (15 total secs)
```

Whereas with the three clauses I get the following with the same configuration as above.
```
10000 reads, 3677246 inserts, 0 updates, 37 truncates, 0 checkpoints in 5 secs (5 total secs)
10000 reads, 3661618 inserts, 0 updates, 37 truncates, 0 checkpoints in 5 secs (10 total secs)
10000 reads, 3650866 inserts, 0 updates, 36 truncates, 0 checkpoints in 5 secs (15 total secs)
10000 reads, 3649161 inserts, 0 updates, 37 truncates, 0 checkpoints in 5 secs (20 total secs)
10000 reads, 3592797 inserts, 0 updates, 36 truncates, 0 checkpoints in 5 secs (25 total secs)
```
I had initially not had the initial very small values clause at the top, but found that there was also another hiccough at these very low values, so I added a third clause.

daveh86:
Still have one large thing that Sue suggested to think about. Hopefully I can get to it tomorrow.

bench/wtperf/wtperf_throttle.c
daveh86: The logic here is basically - we ran long with our last batch towards throttle so lets over-fill the ticket queue up to where we /should/ be and see if we can catch up a little. 

The aim in doing this is to be a little flexible in cases where we see some variance in how long it takes to complete a set of operations. If for example, throughput drops while there is a checkpoint going on, we can then catch up and do a few more operations after it finishes. 

The multiplication was intended as as if we have taken 2 increments of time to perform our last batch, we should allocate two increments worth of tickets. So, in the example above, we will allocate `110` tickets in the hope that in the `20000us` we have, we will do all `110` operations, and thus "catch up" with the required throughput per second.

The catch is that we never want to do "too many" operations so we have the blow-off valve for overfilling on the line below.

michaelcahill:
@daveh86, I'm doing a pass through pull requests.  What state is this one in: is it ready for review or is there more work to do?

daveh86:
@michaelcahill, I've looked over this again and downloaded to re-test. I think it is good to go.

bench/wtperf/wtperf.c
```diff
@@ -503,11 +502,8 @@ worker(void *arg)
 		}
 	}
 	/* Setup the timer for throttling. */
-	if (thread->workload->throttle != 0 &&
-	    (ret = __wt_epoch(NULL, &interval)) != 0) {
-		lprintf(cfg, ret, 0, "Get time call failed");
-		goto err;
-	}
+	if (thread->workload->throttle != 0 && setup_throttle(thread) != 0)
+			goto err;
```
michaelcahill: whitespace

bench/wtperf/wtperf_throttle.c
```diff
@@ -0,0 +1,117 @@
...
+		    (float)throttle_cfg->ops_per_increment;
+
+	/* Don't over-fill the queue */
+	throttle_cfg->ops_count =
+	    WT_MIN(throttle_cfg->ops_count, thread->workload->throttle);
+
+	/*
+	 * After sleeping, set the interval to the current time.
+	 */
+	WT_RET(__wt_epoch(NULL, &throttle_cfg->last_increment));
```
michaelcahill: Can we avoid the second call to `__wt_epoch` in the fast path (where we aren't sleeping) and just set `throttle_cfg->last_increment = now`?

bench/wtperf/wtperf.c
michaelcahill: Also, normal style is to propagate out all errors.  In other words, change this to:

```c
	if (thread->workload->throttle != 0 &&
	    (ret = setup_throttle(thread)) != 0)
		goto err;
```

michaelcahill:
@daveh86 before this change can be merged, please `git rebase -i` and fold all of the changes into one, then change the commit message to reference WT-2267.  We can't add commits to WT-2256 because it is already resolved.

michaelcahill:
@sueloverso, you put a lot of time into reviewing this change.  Are you happy with where it has ended up?

michaelcahill:
Closed in favour of #2240.

PR #2396, opened 2015-12-16
WT-2290: don't reuse previous multi-block write blocks when compacting

keithbostic:
@agorrod, @sue, ready for review/merge.

### DIFF
--- a/src/include/session.h
+++ b/src/include/session.h
@@ -137,8 +137,6 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_session_impl {
 	void	*reconcile;		/* Reconciliation support */
 	int	(*reconcile_cleanup)(WT_SESSION_IMPL *);
 
-	bool compaction;		/* Compaction did some work */
-
 	uint32_t flags;
 
 	/*
--- a/src/include/session.h
+++ b/src/include/session.h
@@ -74,7 +74,10 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_session_impl {
 	TAILQ_HEAD(__cursors, __wt_cursor) cursors;
 
 	WT_CURSOR_BACKUP *bkp_cursor;	/* Hot backup cursor */
-	WT_COMPACT	 *compact;	/* Compact state */
+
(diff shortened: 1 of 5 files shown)


test/suite/test_compact02.py
```diff
@@ -115,11 +115,9 @@ def test_compact02(self):
...
         self.pr('Removed total ' + str((count * 9666) / mb) + 'MB')
 
-        # 4. Reopen the connection to force the file to disk (if we leave dirty
-        # blocks in the cache, it can affect how compact works depending on how
-        # and when those blocks are written, because they're being written best
-        # fit, not first-fit, as compaction does).
-        self.reopen_conn()
+        # 4. Checkpoint to make the deleted records available for reuse, so
+        # compaction can be successful.
+        self.session.checkpoint()
```
sueloverso: I don't think this is necessary.  The test runs without it (I ran 50 iterations on the AWS box that was failing reliably).  The code in `compact_file` calls checkpoint before starting and on each pass already.  I also think not having it is more realistic for what a user is going to do.  Did you see a failure without it?

src/include/session.h
```diff
@@ -74,7 +74,10 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_session_impl {
 	TAILQ_HEAD(__cursors, __wt_cursor) cursors;
 
 	WT_CURSOR_BACKUP *bkp_cursor;	/* Hot backup cursor */
-	WT_COMPACT	 *compact;	/* Compact state */
+
+	WT_COMPACT	 *compact;	/* Compaction information */
+	enum { WT_COMPACT_NONE=0,
+	    WT_COMPACT_RUNNING, WT_COMPACT_SUCCESS } compact_state;
```
sueloverso: Do we still need `session->compaction`?  If not, I don't see its removal here.

sueloverso:
@keithbostic This LGTM.  I had a couple minor comments/questions.

test/suite/test_compact02.py
keithbostic: I agree, this probably isn't necessary, I'll pull it out, as well as the subsequent stat call.

src/include/session.h
keithbostic: Thanks! Change pushed.

PR #2397, opened 2015-12-17
WT-2291: Update random walk in a skip list to improve it's randomness.

agorrod:
It is currently failing. The test is currently targeted at small in-memory trees (i.e: the skiplist random code).

### DIFF
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -627,10 +627,13 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
 	WT_INSERT *ins, **start, **stop;
 	WT_INSERT_HEAD *ins_head;
 	WT_PAGE *page;
-	uint32_t entries;
+	uint32_t choice, entries, i;
 	int level;
 
 	page = cbt->ref->page;
+
+	start = stop = NULL;		/* [-Wconditional-uninitialized] */
+	entries = 0;			/* [-Wconditional-uninitialized] */
 
 	/* If the page has disk-based entries, select from them. */
 	if (page->pg_row_entries != 0) {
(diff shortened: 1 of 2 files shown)


agorrod:
@keithbostic I didn't get back to this issue today, so nothing to add after our call this morning.

keithbostic:
@agorrod, I took another run at this one, for your reviewing pleasure.

With this set of changes, the current version of the test script passes for me, consistently staying at about 50%:
```
% pyt test_cursor_random02
differentKeys: 1 of 1, 100%
differentKeys: 132 of 250, 52%
differentKeys: 269 of 500, 53%
differentKeys: 2414 of 5000, 48%
differentKeys: 5040 of 10000, 50%
differentKeys: 24182 of 50000, 48%
differentKeys: 45369 of 100000, 45%
```
I adapted your C test program as well to confirm the results, and with it I see roughly the same results:
```
% ./t -r 10
start run... 10: 0.00
6 hits of 10 records, 60%
	maximum hits: 2, 2, 1, 1, 1, 1, 0, 0, 0, 0
% ./t -r 100
start run... 100: 0.00
61 hits of 100 records, 61%
	maximum hits: 4, 4, 4, 4, 3, 3, 3, 3, 2, 2
% ./t -r 1000
start run... 1000: 0.00
507 hits of 1000 records, 50%
	maximum hits: 10, 9, 9, 8, 7, 7, 7, 7, 7, 6
% ./t -r 10000
start run... 10000: 0.06
5034 hits of 10000 records, 50%
	maximum hits: 20, 17, 15, 15, 15, 13, 12, 12, 12, 12
% ./t -r 100000
start run... 100000: 6.70
47540 hits of 100000 records, 47%
	maximum hits: 30, 29, 26, 25, 25, 24, 24, 24, 24, 23
18.375u 0.009s 0:18.39 99.8%	1209+203k 0+84io 0pf+0w
% ./t -r 1000000
start run... 1000000: 823.89
412193 hits of 1000000 records, 41%
	maximum hits: 177, 175, 169, 160, 113, 90, 81, 80, 73, 72
```
What I find encouraging is the number of requests go up by a factor of 10 but the number of repeated hits increases linearly (the "maximum hits" line is the maximum number of repeated hits in a single bucket).

Maybe you'll find a bug in the code that will get our numbers up higher?

I'm a little disappointed in the results, to be honest: my testing has convinced me our skiplist code isn't producing good/uniform distributions. The 1/4 probability we use to determine the depth doesn't lead to uniform distributions, at least, not at the higher levels of the skiplist. In other words, the better our performance on this test, the worse our results.

keithbostic:
autobuild: retest this please

agorrod:
Thanks @keithbostic I dialled up the number of random next operations on the test to be 20x the number of records inserted, and the distribution looks better:

differentKeys: 1 of 1, 100%
differentKeys: 250 of 250, 100%
differentKeys: 499 of 500, 99%
differentKeys: 4872 of 5000, 97%
differentKeys: 9836 of 10000, 98%
differentKeys: 48903 of 50000, 97%
Keys: 95508 of 100000, 95%

That coverage is good. Interestingly if I look at the distribution of which offsets are chose more regularly I see a bias towards the right hand side of the tree. The current code gives special consideration for the case where we choose the end of a range but there is more content to the right a level down, but it doesn't do a similar step for the range to the left.

I also ran an experiment with our random number generator, to see what the best result we could expect are. That is generating a distribution across a range of 5000 entries with 5000 samples I see:

``Saw 3168 of 5000 entries across 5000 samples``

So hitting ~50% with this random code seems reasonable.

agorrod:
@keithbostic I reviewed this change and it LGTM. I took a run at improving this without finding any improvements. Please go ahead and merge if you are happy.

PR #2398, opened 2015-12-17
WT-2291 experimental random_next skip list change.

agorrod:
Not currently working well, but an attempt to alter the control flow so we don't need to handle the edge cases as often, which will hopefully avoid adding as much artificial skew.


agorrod:
This has been replaced by a different implementation.

PR #2402, opened 2015-12-19
WT-2297: Fix off-by-one error in Huffman config file parsing

GeertBosch:
Line-numbers are one-based, so reading a Huffman configuration file with 256 lines would result in an complaint that the file had more than 255 lines.


wtbuild:

keithbostic:
Thanks, @GeertBosch, merged (with a related fix to the error message).

PR #2403, opened 2015-12-20
WT-2299: upper-level WiredTiger code is  reaching into the block manager

keithbostic:
@agorrod, just a minor cleanup for your consideration/review.


agorrod:
Thanks @keithbostic lgtm.

PR #2404, opened 2015-12-21
WT-2307 debugging branch

keithbostic:
Debugging branch, never intended to be merged.

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -431,6 +431,112 @@ __wt_btcur_iterate_setup(WT_CURSOR_BTREE *cbt)
 }
 
 /*
+ * __wt_last_op_name --
+ *	Return the last operation name.
+ */
+static const char *
+__wt_last_op_name(int v)
+{
+	switch (v) {
+	case WT_LASTOP_NONE: return ("none");
+	case WT_LASTOP_NEXT: return ("next");
+	case WT_LASTOP_PREV: return ("prev");
+	case WT_LASTOP_RESET: return ("reset");
+	case WT_LASTOP_SEARCH: return ("search");
+	case WT_LASTOP_SEARCH_NEAR: return ("search-near");
+	case WT_LASTOP_INSERT: return ("insert");
+	case WT_LASTOP_TRUNCATE: return ("truncate");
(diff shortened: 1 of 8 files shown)


src/btree/bt_curnext.c
```diff
@@ -431,6 +431,147 @@ __wt_btcur_iterate_setup(WT_CURSOR_BTREE *cbt)
...
+	    __wt_last_op_name(cbt->last_op[13]),
+	    __wt_last_op_name(cbt->last_op[14]),
+	    __wt_last_op_name(cbt->last_op[15]),
+	    __wt_last_op_name(cbt->last_op[16]),
+	    __wt_last_op_name(cbt->last_op[17]),
+	    __wt_last_op_name(cbt->last_op[18]),
+	    __wt_last_op_name(cbt->last_op[19]));
+
+	for (i = cbt->next_track - 1; i >= 0; --i)
+		__wt_errx(session,
```
RedBeard0531: Assuming that cursors are calloc'ed (or otherwize zero initialized) does it make sense to skip printing in these loops if .locate == 0? Or make __wt_track_location_name have a special case for 0?

keithbostic: They get set to 0 if the cursor doesn't ref a page -- I don't anticipate that happening, but it's a sanity check if it does.

keithbostic:
@agorrod, @sueloverso, @daveh86: This branch was a debugging branch, and is no longer interesting. However, I would like to understand why key-out-of-order tests are flagging some of the standard tests (that's why it's not passing checks):
```
ERROR: test_cursor05.test_cursor05.test_without_colgroups
ERROR: test_cursor05.test_cursor05.test_with_colgroups
ERROR: test_txn01.test_read_committed_default.test_read_committed_default
```
Anyway, if somebody gets to it before I do, that'd be great, otherwise I'll get back to it.

Alex's branch #2407 fails in similar ways (it has the same checking code).

sueloverso:
@keithbostic I can reproduce the error on these debug branches, but it is fixed now with the merge of #2415.

keithbostic:
Deleting this diagnostic pull request for now, #2415 will hopefully resolve the problem.

PR #2406, opened 2015-12-21
WT-2302: change test/format to check cursor movement key order.

keithbostic:
@agorrod, for your consideration/review.


keithbostic:
Deleting this diagnostic pull request for now, #2415 will hopefully resolve the problem.

Instead of doing this check in test/format, we're going to add DIAGNOSTIC checks to the WiredTiger library itself.

PR #2408, opened 2015-12-22
WT-2304 - Fix memory leaking issue in config dump when there are multiple databases

daveh86:
This inits a new config object, thus being always empty and not running afoul of the config_free function.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -54,6 +54,7 @@ static void config_opt_usage(void);
 int
 config_assign(CONFIG *dest, const CONFIG *src)
 {
+	CONFIG_QUEUE_ENTRY *conf_line, *tmp_line;
 	size_t i, len;
 	char *newstr, **pstr;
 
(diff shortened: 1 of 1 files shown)


daveh86:
@agorrod, this should fix WT-2304. Can we discuss if there are better ways or merge if you think this is suitable?

agorrod:
@daveh86 I think each copy of the config should have the structure filled in. The multi-database code assumes that it's dealing with a fully populated CONFIG structure. I know we don't currently use that configuration stack for anything other than re-writing it, but that may change in the future.

sueloverso:
When looking at this today, it surprised me that we kept the config lines allocated in memory after writing the file, until the end of the test.  What if, instead of copying it to each database, we freed it after dumping the config. I'm not sure if we need just one copy in WT_TEST or a copy in each database subdir.

agorrod:
>  it surprised me that we kept the config lines allocated in memory after writing the file, until the end of the test

I'm inclined to keep the configuration allocation and deallocation all together, rather than splitting it up in different locations. A concern I had with freeing the memory after dumping the configuration is how we would know to free or copy the content if for some reason wtperf didn't go through the config dump phase.

bench/wtperf/config.c
```diff
@@ -96,6 +97,18 @@ config_assign(CONFIG *dest, const CONFIG *src)
 		}
 
 	TAILQ_INIT(&dest->stone_head);
+	TAILQ_INIT(&dest->config_head);
+
+	/* Clone the config string information into the new cfg object */
+	conf_line = TAILQ_FIRST(&src->config_head);
+	while (conf_line != NULL) {
```
agorrod: Could you use ``TAILQ_FOREACH`` instead of a while loop?

bench/wtperf/config.c
```diff
@@ -96,6 +97,18 @@ config_assign(CONFIG *dest, const CONFIG *src)
...
+		len = strlen(conf_line->string);
+		tmp_line = calloc(sizeof(CONFIG_QUEUE_ENTRY),1);
```
agorrod: You should check for an allocation failure and return an error if it did.

agorrod:
@daveh86 Could you try running this via valgrind for the ``evict-btree.wtperf``, ``shared-cache-stress.wtperf`` and ``multi-btree.wtperf`` workloads?

daveh86:
@agorrod, I've checked those three workloads with valgrind and no new leaks/errors have shown up.

Only thing that showed a non-zero value was the following at end of test
```
==21681==    still reachable: 72,704 bytes in 1 blocks
```

However, this appears to be pre-existing as I was able to get the same result pre-patch.

I think this means valgrind testing was a success.

agorrod:

PR #2412, opened 2015-12-23
WT-2307 workaround. Reposition and retry next when we detect out of order.

agorrod:
This improves the situation with the wtperf reproducer, but I still
see wtperf reporting out of order keys during traversal occasionally.


agorrod:
@keithbostic I was hoping this would be a viable workaround for the bug described in WT-2307, but I'm still seeing (less regular) out of order keys reported in wtperf.

Would you mind taking a look? If you think this is a reasonable approach (and can spot what I missed), it would be interesting to create a version of this change against the mongodb-3.2 branch for testing.

agorrod:
> but I'm still seeing (less regular) out of order keys reported in wtperf

Got it, it wasn't retrying if it saw the error twice in a row.

keithbostic:
This pull request is hopefully not needed, currently using #2415 instead.

PR #2413, opened 2015-12-24
WT-2307 workaround for cursor next bug.

agorrod:
Detect when a cursor next returned an invalid result and retry
the next operation. Horrible since it adds a comparison into each
cursor next, but fixes the symptom until we can fix the underlying bug.


keithbostic:
This pull request is hopefully not needed, currently using #2415 instead.

PR #2414, opened 2015-12-24
WT-2305 - Fix coverity issues

daveh86:
Fixes for some of the recently found coverity issues.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -736,8 +736,8 @@ config_to_file(CONFIG *cfg)
 	fp = NULL;
 
 	/* Backup the config */
-	req_len = strlen(cfg->home) + 14;
-	if ((path = calloc(req_len + 1, 1)) == NULL) {
+	req_len = strlen(cfg->home) + strlen("/CONFIG.wtperf") + 1;
+	if ((path = calloc(req_len, 1)) == NULL) {
 		(void)enomem(cfg);
 		goto err;
 	}
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -601,7 +601,9 @@ config_opt_line(CONFIG *cfg, const char *optstr)
 		return (enomem(cfg));
 
 	strncpy(string_copy, optstr, len);
-	config_line = calloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
+	if ((config_line = calloc(sizeof(CONFIG_QUEUE_ENTRY), 1)) == NULL)
+		return (enomem(cfg));
+
 	config_line->string = string_copy;
 	TAILQ_INSERT_TAIL(&cfg->config_head, config_line, c);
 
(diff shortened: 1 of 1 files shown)


bench/wtperf/config.c
```diff
@@ -734,13 +736,13 @@ config_to_file(CONFIG *cfg)
...
-	req_len = strlen(cfg->home) + 100;
-	if ((path = calloc(req_len, 1)) == NULL) {
+	req_len = strlen(cfg->home) + 14;
+	if ((path = calloc(req_len + 1, 1)) == NULL) {
 		(void)enomem(cfg);
 		goto err;
 	}
 
-	snprintf(path, req_len + 14, "%s/CONFIG.wtperf", cfg->home);
+	snprintf(path, req_len, "%s/CONFIG.wtperf", cfg->home);
```
keithbostic: @daveh86, a safer way to do this is:
```
req_len = strlen(cfg->home) + strlen("/CONFIG.wtperf") + 1;
```
that way you don't get caught if the name changes, or you just miscounted the characters (which I invariably do).

On modern compilers, it's just as fast as calling `sizeof`, too, the compiler figures out it's a `strlen` of a constant string in most cases (and we don't care here, anyway).

keithbostic:
@daveh86, @agorrod:

There's a lot of (inconsistent) memory allocation error handling in `wtperf`: my suggestion would be to add a function that takes a config argument and calls calloc/malloc, and just exits if the calloc/malloc fails (logging the error if the config isn't NULL). Then, do a pass to replace every call to calloc/malloc in wtperf with a call to the new function.

If wtperf can't allocate memory, we're going to eventually exit anyway, it's not like we're trying recover and continue the run, right? And that change will simplify an awful lot of code.

keithbostic:
@daveh86, here's what test/format does:
```
/*
 * dmalloc --
 *      Call malloc, dying on failure.
 */
void *
dmalloc(size_t len)
{
        void *p;
        
        if ((p = malloc(len)) == NULL)
                die(errno, "malloc");
        return (p);
}
                
/*
 * dstrdup --
 *      Call strdup, dying on failure.
 */
static char *
dstrdup(const char *str)
{
        char *p;
        
        if ((p = strdup(str)) == NULL)
                die(errno, "strdup");
        return (p);
}
```
It would also be good to replace any occurrences of this pattern:
```
len = strlen(*pstr) + 1;
if ((newstr = malloc(len)) == NULL)
        return (enomem(src));
strncpy(newstr, *pstr, len);
```
with the simpler call to strdup:
```
if ((newstr = strdup(*pstr)) == NULL)
        return (enomem(src));
```

even if you don't switch to the dmalloc/dstrdup calls.

In general, when you see a call to `strncpy` in code, it's almost always wrong -- if there's a terminating nul byte, `strcpy` is better, and if there's no terminating nul byte, `memmove` is better. (The `strncpy` function can also leave you with a non-nul-byte terminated string, the call is just strange and mostly useless.)

daveh86:
@keithbostic, thanks for the feedback. Really like the idea of splitting the allocations out into their own functions, I've forked off WT-2326 for this change and will make a patch for this soon.

agorrod:
@daveh86 Should I go ahead and merge this branch, or will the changes for WT-2326 replace these changes?

daveh86:
@agorrod, these changes are independent of those in WT-2326. Its probably best to get this merged first as is, then fix everything with WT-2326. I think that makes for a cleaner series of changes, if that makes sense.

agorrod:
Sounds good. I'll go ahead and merge.

PR #2416, opened 2015-12-28
WT-2307 Use single enum for split error handling phases.

sueloverso:
@agorrod and @keithbostic here's a pass at creating a single enum.  I did leave the long comment in both functions with the more thorough description.  Please review and merge if you like it.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -15,6 +15,23 @@
 } while (0)
 
 /*
+ * A note on error handling: this function first allocates/initializes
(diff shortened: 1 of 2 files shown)


sueloverso:
@keithbostic I moved the enum into `bt_split.c`.  Although I added the use of it in `split_parent` I did not make additional changes to that function to use the new `_step1` and `_step2` functions to minimize changes.  Maybe for this change we don't care about minimizing the changes?  Anyway - for your review.

keithbostic:
@sueloverso, I did the merge after a couple of minor changes.

PR #2417, opened 2015-12-28
WT-2307: diagnostic checks for cursors returning keys out-of-order

keithbostic:


keithbostic:
@agorrod, @sueloverso, the changes we discussed, ready for review/merge.

I'd like to have caught an out-of-order search/search-near followed by a cursor movement, but I couldn't think of an easy way to make that work, it ended up being invasive and ugly, so I didn't keep that work.

I also increased the amount of cursor motion done by test/format.

I've verified this branch catches the test case for WT-2307 when applied to an old version of develop.

test/format/ops.c
```diff
@@ -504,7 +504,7 @@ skip_insert:			if (col_update(tinfo,
 		 */
 		if (!insert) {
 			dir = (int)mmrand(&tinfo->rnd, 0, 1);
-			for (np = 0; np < mmrand(&tinfo->rnd, 1, 8); ++np) {
+			for (np = 0; np < mmrand(&tinfo->rnd, 1, 30); ++np) {
```
sueloverso: Were you able to reproduce the issue when run with the old code with this change?

src/btree/bt_curnext.c
```diff
@@ -430,6 +438,104 @@ __wt_btcur_iterate_setup(WT_CURSOR_BTREE *cbt)
...
+
+	WT_ERR(__wt_scr_alloc(session, 512, &a));
+	WT_ERR(__wt_buf_set_printable(
+	    session, a, cbt->lastkey->data, cbt->lastkey->size));
+
+	WT_ERR(__wt_scr_alloc(session, 512, &b));
+	WT_ERR(__wt_buf_set_printable(session, b, key->data, key->size));
+
+	WT_PANIC_ERR(session, EINVAL,
+	    "WT_CURSOR.%s out-of-order returns: returned key %.*s then "
```
sueloverso: My only suggestion is to make these two error messages different so that we know immediately which type they're using.

test/format/ops.c
keithbostic: > Were you able to reproduce the issue when run with the old code with this change?

I originally made changes in test/format to try and catch the WT-2307 failure (see WT-2302, #2406), but discarded those changes in favor doing the same checks inside the library. Those original changes never detected the bug.

If I apply these changes to a version of the develop branch without the WT-2307 fixes and run the wtperf split-stress workload, these changes detect the bug.

src/btree/bt_curnext.c
keithbostic: @sueloverso and I agreed this information is probably available from the key, so we'll leave the current messages unchanged.

sueloverso:
Thanks @keithbostic.  LGTM.

PR #2419, opened 2016-01-02
WT-2314: better error handling in the page-swap function and callers.

keithbostic:
@agorrod, @sueloverso: ready for review/merge.


keithbostic:
One comment: having two booleans side-by-side in the argument list is a bad idea, we'll get them confused, but I can't think of anything better.

src/include/btree.i
```diff
@@ -1281,7 +1281,7 @@ __wt_page_release(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
  */
 static inline int
 __wt_page_swap_func(WT_SESSION_IMPL *session, WT_REF *held,
-    WT_REF *want, uint32_t flags
+    WT_REF *want, bool notfound_ok, bool restart_ok, uint32_t flags
```
sueloverso: Although not a whole lot cleaner, you could add two more flags instead of adding two boolean args.  `WT_READ_NOTFOUND_OK` and `WT_READ_RESTART_OK`.  There are a few other flags that are defined and used in other places, but not used within `wt_page_in_func` as precedent.  It makes the callers more readable at a glance with:
```
flags |= WT_READ_NOTFOUND_OK;
WT_RET(__wt_page_swap_func(...));
```
rather than having to figure out which arg is which every time.

src/include/btree.i
```diff
@@ -1310,20 +1310,40 @@ __wt_page_swap_func(WT_SESSION_IMPL *session, WT_REF *held,
 #endif
 	    );
 
-	/* Expected failures: page not found or restart. */
-	if (ret == WT_NOTFOUND || ret == WT_RESTART)
-		return (ret);
+	/*
+	 * Expected failures: page not found or restart. Our callers list the
+	 * errors they're expecting to handle.
```
sueloverso: I think the comment should be clear that the caller is only expecting to handle errors from the `wt_page_in_func` function and making assumptions based on that.  It took me a bit to understand the later code to distinguish these two cases and the user not handling the same error from one of the release functions.

sueloverso: Actually, to me, this makes my earlier suggestion of using `WT_READ_RESTART_OK` even more readable because it implies the error is coming from the read/page-in function.

agorrod:
I'm with @sueloverso on this one. I like the change of making error handling explicit, but I'm not a fan of the boolean variables - I'd prefer flags.

keithbostic:
@sueloverso, @agorrod, I agree, that does make things better.

Change pushed.

sueloverso:
Thanks!  LGTM.

PR #2421, opened 2016-01-05
WT-2275 Convert a table URI truncate into a range truncate.

sueloverso:
@michaelcahill Please review this branch.  The ticket revealed a bug keeping files and metadata in sync during an abort in a truncate.  @keithbostic suggested we could consider removing support for URI truncate in favor of drop and create.  In thinking about that, I came up with this solution instead.

If we consider that a truncate is emptying the content of a table, it avoids the metadata/file interactions to translate a URI based truncate into a range truncate with a start cursor on the first item.  It also removes a lot of code.


michaelcahill:
@sueloverso, thanks, this is an interesting idea...

On the negative side, this version of truncate will have to walk internal pages in order to mark leaf pages deleted, and the filesystem space will not be reclaimed immediately.

On the positive side, this fixes the window where the metadata can be inconsistent with a file, is one less operation that needs exclusive access to the file, makes truncate immediately transactional (so it can safely be rolled back), means that `WT_SESSION::truncate` has consistent rules about when it is safe to call (it will never fail because of an open cursor in the file).

I'm a fan.  This is definitely preferable to drop and create, at least until those operations can be bundled into a transaction.  Let me think a little more about documentation / upgrading, but I think this is the way to go.

michaelcahill:
@sueloverso, I'd like to go in this direction but I have some concerns:

1. This will turn a truncate of an LSM tree into inserting tombstones for every record.  I don't think that is acceptable long-term, but OTOH I bet there are similar bugs to WT-2275 with the existing LSM tree truncate.  I suggest we make the slow path correct and open a ticket to figure out how to make LSM truncate fast again later?

2. This will need upgrading documentation.  In particular, the old truncate needed exclusive access to the data source but this version doesn't, and the new version can fail with `WT_ROLLBACK` but I don't think we would ever have done that before.  Performance may differ in hard-to-predict ways, and disk space will not become available immediately.  Can you please cover this in the docs?

sueloverso:
@michaelcahill Please review the doc changes.  I could not find anywhere that stated the URI version of `session->truncate` required exclusive access to the table.  If I missed it, let me know where it is and I'll edit that too.  I also could not find any python unit test that calls truncate (URI or range) on an LSM tree.

michaelcahill:
Closed in favor of #2454.

PR #2422, opened 2016-01-09
WT-2316: WT_CURSOR.prev out-of-order returns, fix a split race.

keithbostic:
WT-2316: WT_CURSOR.prev out-of-order returns, fix a split race.


src/btree/bt_walk.c
```diff
@@ -132,6 +131,43 @@ __page_ascend(WT_SESSION_IMPL *session,
...
+		 * slot on the page would be in the wrong part of the namespace.
+		 * The test isn't necessary for a "next" cursor because we do
+		 * right-hand splits on internal pages and the first part of the
+		 * page's namespace won't have changed as part of the split.
+		 * Rather than testing the direction boolean, do the test the
+		 * "previous" cursor movement requires, it's not wrong for the
+		 * "next" cursor movement.
+		 */
+		if (pindex->index[*slotp]->home == ref->page)
+			break;
```
michaelcahill: @keithbostic, this seems related to `__wt_split_intl_race` -- could we save the `pindex` from the ascend to reuse the same test?

keithbostic:
@michaelcahill, I pushed the change to do cursor key order checking after search/search-near, in 70c8b78.

In short, I turn on testing on successful return from search/search-near, and turn off testing when `__cursor_func_init` is called with `reenter` set. Can you please confirm that approach sounds reasonable?

src/btree/bt_walk.c
keithbostic: @michaelcahill 

> this seems related to __wt_split_intl_race -- could we save the pindex from the ascend to reuse the same test?

I don't think so; here's what I'm thinking at the moment.

To restate, there are two tests we have when descending the tree:

* In column-store and row-store search, test if the parent's page-index used to descend has changed since we used it; this test catches if we already descended into the wrong page (we searched the parent before it's updated and searched the child after it's updated).

    In this test, what we care about is if we descended into the wrong namespace, specifically, we searched a page for a namespace it no longer has.

* In cursor walk, test if the last WT_REF in an internal page points to the internal page; this test catches if we are descending into a WT_REF no longer associated with a page.

    In this test, what we care about is if we're about to descend into a namespace that's moved to a different parent.

Using the walk test in the search code: the WT_REF into which we're descending can be correct (the problem WT_REF is not a WT_REF moved by the split, it's one that remains on the original page after the split).

Using the search test in the walk code: the parent (ascend) page-index doesn't help us because what we need is the combination of a page-index and a WT_REF that are in sync. We could descend into the wrong WT_REF without detecting it if we only tested if the parent page-index was changed, because that doesn't tell us if the WT_REF we're about to descend into is in sync with that page-index.

Maybe there's a test that works for both cases, but I'm struggling to find it.

Can we talk this over, maybe with @agorrod?

Would it be useful to draw pictures of the 3 cases we have, and if so, where would those pictures go? (I could try ASCII art in the functions.)

keithbostic:
@michaelcahill, @agorrod, I've added a bunch of text to each of the race cases, trying to explain what each is intended to fix, in f1977bb.

Please let me know if I can make it better.

agorrod:
Thanks @keithbostic This change lgtm now, and the comments make sense.

@michaelcahill wanted to check that different refs aren't going to cause problems before this is merged.

michaelcahill:
Thanks, @keithbostic -- this lgtm.  I convinced myself that `ref->home` is a reasonable thing to check, and the search+next/prev checking looks right to me.  I'll merge...

PR #2423, opened 2016-01-10
SERVER-22064: Coverity, function return value not checked for error

keithbostic:
Coverity analysis defect 77699: Unchecked return value, problem
introduced in SERVER-21619 change, commit 354c031.

Instead of calling __wt_evict_page_clean_update() when discarding pages,
call __wt_ref_out() directly, __wt_evict_page_clean_update() doesn't do
any useful additional work.

This allows __wt_evict_page_clean_update() to be static in evict_page.c,
rename to __evict_page_clean_update().


keithbostic:

michaelcahill:
Thanks @keithbostic -- lgtm.  This used to not be possible because discards could give up part way through, but now that they just throw everything away, there's no point going through eviction.

I'll merge.

PR #2424, opened 2016-01-10
WT-2309: Add yields and/or sleeps in #DIAGNOSTIC mode.

keithbostic:
WT-2309: Add yields and/or sleeps in #DIAGNOSTIC mode.

Add a new function __wt_diagnostic_yield() that yields (and optionally sleeps), add it to the split code to encourage races.

### DIFF
--- a/src/include/misc.i
+++ b/src/include/misc.i
@@ -70,3 +70,19 @@ __wt_verbose(WT_SESSION_IMPL *session, int flag, const char *fmt, ...)
 	return (0);
 #endif
 }
+
+/*
+ * __wt_diagnostic_yield --
+ * 	Yield, and optionally pause, in diagnostic mode.
+ */
+static inline void
+__wt_diagnostic_yield(uint64_t seconds, uint64_t micro_seconds)
+{
+#ifdef HAVE_DIAGNOSTIC
+	if (seconds == 0 && micro_seconds == 0)
+		__wt_yield();
+	else
+		__wt_sleep(seconds, micro_seconds);
+#else
+#endif
+}
--- a/src/include/misc.i
+++ b/src/include/misc.i
@@ -83,6 +83,5 @@ __wt_diagnostic_yield(uint64_t seconds, uint64_t micro_seconds)
 		__wt_yield();
 	else
 		__wt_sleep(seconds, micro_seconds);
-#else
 #endif
 }
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -650,12 +650,16 @@ __split_root(WT_SESSION_IMPL *session, WT_PAGE *root)
 	/* Prepare the WT_REFs for the move. */
 	__split_ref_step1(session, alloc_index, false);
 
+	__wt_diagnostic_yield(0, 100000);	/* Encourage a race. */
+
 	/*
 	 * Confirm the root page's index hasn't moved, then update it, which
 	 * makes the split visible to threads descending the tree.
 	 */
 	WT_ASSERT(session, WT_INTL_INDEX_GET_SAFE(root) == pindex);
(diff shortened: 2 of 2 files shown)


keithbostic:

agorrod:
Thanks for getting this started @keithbostic I'd prefer to add a new configure option to enable/disable this functionality. My reasoning is that I'd like to be able to aggressively pause when this configuration is enabled, which will cause a significant performance drop. I like that diagnostic currently introduces less than an order of magnitude performance penalty, so it's practical to run a majority of our test code with diagnostic enabled.

keithbostic:
@agorrod, absent measurement proving otherwise, I doubt this change will change diagnostic run performance significantly.

We already do slow things when DIAGNOSTIC is enabled (for example, full decompression and page verification of every raw compression reconciliation result), I don't anticipate this pushing us over an edge.

src/include/misc.i
```diff
@@ -70,3 +70,19 @@ __wt_verbose(WT_SESSION_IMPL *session, int flag, const char *fmt, ...)
 	return (0);
 #endif
 }
+
+/*
+ * __wt_diagnostic_yield --
```
sueloverso: I'd also like to have a diagnostic function to introduce aborts or errors in some code paths.  We could generalize this to support those uses too or add a couple more similar functions.

src/include/misc.i
```diff
@@ -70,3 +70,19 @@ __wt_verbose(WT_SESSION_IMPL *session, int flag, const char *fmt, ...)
...
+ */
+static inline void
+__wt_diagnostic_yield(uint64_t seconds, uint64_t micro_seconds)
+{
+#ifdef HAVE_DIAGNOSTIC
+	if (seconds == 0 && micro_seconds == 0)
+		__wt_yield();
+	else
+		__wt_sleep(seconds, micro_seconds);
+#else
```
sueloverso: The `#else` isn't necessary.

src/include/misc.i
keithbostic: @sueloverso
> I'd also like to have a diagnostic function to introduce aborts or errors in some code paths.

That could be a separate ticket? I'd like to get this pushed and into the stress tests.

src/include/misc.i
keithbostic: @sueloverso
> The #else isn't necessary.

Thanks, fixed.

src/include/misc.i
sueloverso: Sure.  I created https://jira.mongodb.org/browse/WT-2324 for that.

sueloverso:
@keithbostic I appreciate you wanting to get this into stress testing ASAP, I think its current form may only be interim and warrants more discussion.  I have pulled this branch to add calls to this in a few places to force WT-2313 sooner.  That made me wonder if we want to add a call to `wt_diagnostic_yield` in `WT_ASSERT` itself.  If many asserts are testing for correct operations/state and race conditions, adding a yield in will encourage those races.  The other thought I had was that this really has the feel of verbose where we want to add it liberally to many places but selectively turn it on.

agorrod:
My reading is that this change causes a thread to pause by 0.2 seconds when doing a root split, or 0.3 seconds for an internal split. I'm concerned that slowing down splits by that much in all diagnostic builds will alter how much work we are getting done and the shape of trees we generate compared to non-diagnostic builds. If we want to add a change like this into all diagnostic builds soon I'd be inclined to decrease the pause time by an order of magnitude.

keithbostic:
OK.

I have my own changes I use for stress testing splits (and that branch has longer sleeps in it than this, anyway, so it isn't going away). This change is too simplistic to keep around if it's not going to be merged immediately, I'll discard the branch.

PR #2427, opened 2016-01-10
WT-2320 Only check copyright as part of cutting a release.

michaelcahill:
Builds on another change that was pushed directly to develop branch: 00c2116d630416dbc871b539b5e234c14248f5b6

### DIFF
--- a/dist/s_copyright
+++ b/dist/s_copyright
@@ -1,5 +1,8 @@
 #! /bin/sh
 
+# Only run when building a release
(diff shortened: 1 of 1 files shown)


michaelcahill:
@agorrod, I screwed up by pushing a change to develop, but I think this branch does what we want (note that `s_release` sets the environment variable and `s_docs` already checks it).  Can you please review?

agorrod:

PR #2430, opened 2016-01-12
WT-2326 - Add dcalloc, dmalloc and dstrdup to wtperf 

daveh86:
and replace all existing malloc, calloc and strncpy calls

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -62,9 +62,7 @@ config_assign(CONFIG *dest, const CONFIG *src)
 	memcpy(dest, src, sizeof(CONFIG));
 
 	if (src->uris != NULL) {
-		dest->uris = calloc(src->table_count, sizeof(char *));
-		if (dest->uris == NULL)
-			return (enomem(dest));
+		dest->uris = dcalloc(src->table_count, sizeof(char *));
 		for (i = 0; i < src->table_count; i++)
 			dest->uris[i] = strdup(src->uris[i]);
 	}
@@ -75,9 +73,7 @@ config_assign(CONFIG *dest, const CONFIG *src)
 	if (src->base_uri != NULL)
 		dest->base_uri = strdup(src->base_uri);
 	if (src->workload != NULL) {
-		dest->workload = calloc(WORKLOAD_MAX, sizeof(WORKLOAD));
-		if (dest->workload == NULL)
-			return (enomem(dest));
+		dest->workload = dcalloc(WORKLOAD_MAX, sizeof(WORKLOAD));
 		memcpy(dest->workload,
 		    src->workload, WORKLOAD_MAX * sizeof(WORKLOAD));
 	}
@@ -89,9 +85,8 @@ config_assign(CONFIG *dest, const CONFIG *src)
 			    ((u_char *)dest + config_opts[i].offset);
 			if (*pstr != NULL) {
 				len = strlen(*pstr) + 1;
-				if ((newstr = malloc(len)) == NULL)
-					return (enomem(src));
-				strncpy(newstr, *pstr, len);
+				newstr = dmalloc(len);
+				newstr = dstrdup(*pstr);
 				*pstr = newstr;
 			}
 		}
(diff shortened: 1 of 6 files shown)


daveh86:

This change is based off the suggestion @keithbostic made that we should change how wtperf uses malloc. I took a little liberty and added a dcalloc, as that seems to be the main source of allocations within wtperf.

bench/wtperf/config.c
```diff
@@ -89,9 +85,8 @@ config_assign(CONFIG *dest, const CONFIG *src)
 			    ((u_char *)dest + config_opts[i].offset);
 			if (*pstr != NULL) {
 				len = strlen(*pstr) + 1;
-				if ((newstr = malloc(len)) == NULL)
-					return (enomem(src));
-				strncpy(newstr, *pstr, len);
+				newstr = dmalloc(len);
+				newstr = dstrdup(*pstr);
```
keithbostic: I don't think the `dmalloc` call is needed.

keithbostic: Or the `len =`?

bench/wtperf/config.c
```diff
@@ -102,11 +97,9 @@ config_assign(CONFIG *dest, const CONFIG *src)
...
 	TAILQ_FOREACH(conf_line, &src->config_head, c) {
 		len = strlen(conf_line->string);
-		if ((tmp_line = calloc(sizeof(CONFIG_QUEUE_ENTRY), 1)) == NULL)
-			return (enomem(src));
-		if ((tmp_line->string = calloc(len + 1, 1)) == NULL)
-			return (enomem(src));
-		strncpy(tmp_line->string, conf_line->string, len);
+		tmp_line = dcalloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
+		tmp_line->string = dcalloc(len + 1, 1);
+		tmp_line->string = dstrdup(tmp_line->string);
```
keithbostic: Again, probably don't need the `len =` or the `dcalloc` call?

bench/wtperf/config.c
```diff
@@ -610,11 +595,10 @@ config_opt_line(CONFIG *cfg, const char *optstr)
...
 	 */
-	if ((string_copy = calloc(len + 1, 1)) == NULL)
-		return (enomem(cfg));
+	string_copy = dcalloc(len + 1, 1);
 
-	strncpy(string_copy, optstr, len);
-	config_line = calloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
+	string_copy = dstrdup(optstr);
+	config_line = dcalloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
 	config_line->string = string_copy;
```
keithbostic: Another example, I think this would be:
```
config_line = dcalloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
config_line->string = dstrdup(optstr);
```
and `string_copy` isn't needed.

keithbostic:
@daveh86, I think there are some places you cut-and-pasted too quickly, but I think this is the right direction.

daveh86:
Thanks for this @keithbostic. I saw a couple of other places where similar changes could be made but held off to make this a literal "replace" those calls, which looks like a bad decision in hindsight.

daveh86:
@keithbostic, would you be able to take another look at this. Please excuse the mess I made of the commits, I rebased instead of doing a merge.

agorrod:
@daveh86 Have you run this branch using valgrind to check for memory leaks using a number of different configuration files?

daveh86:
@agorrod, got this done today and have found a series of leaks.

These are now resolved.

keithbostic:
@daveh86, I took a look, this is much cleaner, thanks!

I did some additional minor cleanup, can you please quickly review f2699e8, 26b8aaf, 2b35a62, d2ee578 and 65c59d7?

There's really no point to almost all of the error-handling in wtperf, it's not like the application is trying to keep running after an error. I suspect `lprintf` could probably call `die`, which would remove 95% of it, but that's probably a change for another day.

PR #2431, opened 2016-01-12
WT-2295 When populating an index, detect if already exists to avoid full scan

ddanderson:
This passes the submitted test case.


ddanderson:

michaelcahill:
@ddanderson, I have looked through this and there are more changes than I'd expected.

It seems like this test is closely connected with `exclusive=true` (i.e., an exclusive create failing is the signal to skip populating).  Is there a path to using the existing logic?

ddanderson:
Thanks, @michaelcahill.  I've created pull request #2442 with that approach.

PR #2432, opened 2016-01-12
WT-2328: schema drop does direct unlink, change to a block manager interface

keithbostic:
WT-2328: schema drop does direct unlink, change to a block manager interface

@agorrod, @michaelcahill: a small restructuring to put the block manager into the drop-file path.

If/when we have multiple block managers, this will change, but for now I think it's correct.


agorrod:

PR #2433, opened 2016-01-13
WT-2313 Wrap clearing of WT_DHANDLE_OPEN with eviction exclusive calls.

sueloverso:
@keithbostic Please review and run this on  your machine with `verbose=(temporary)` turned on.  If you see the `clear_open...` message and it successfully completes then it took the original code path.  This is not ready to merge until the debug verbose message is removed.

### DIFF
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -125,41 +125,13 @@ __wt_conn_dhandle_find(
 static int
 __conn_dhandle_mark_dead(WT_SESSION_IMPL *session)
 {
-	bool evict_reset;
-
 	/*
 	 * Handle forced discard (e.g., when dropping a file).
 	 *
 	 * We need exclusive access to the file -- disable ordinary
 	 * eviction and drain any blocks already queued.
 	 */
-	WT_RET(__wt_evict_file_exclusive_on(session, &evict_reset));
 	F_SET(session->dhandle, WT_DHANDLE_DEAD);
-	if (evict_reset)
-		__wt_evict_file_exclusive_off(session);
-	return (0);
-}
-
-/*
- * __conn_dhandle_clear_open --
- *	Clear a data handle's open flag .
- */
-static int
-__conn_dhandle_clear_open(WT_SESSION_IMPL *session)
-{
-	bool evict_reset;
-
-	/*
-	 * We need exclusive access to the file -- disable ordinary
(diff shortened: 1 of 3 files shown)


sueloverso:
Note, I did consider generalizing at a few different levels but decided to go with the simpler fix initially.

sueloverso:
I will note that this change does not fix the problem.  More debugging.

src/conn/conn_dhandle.c
```diff
@@ -125,18 +125,13 @@ __wt_conn_dhandle_find(
 static int
 __conn_dhandle_mark_dead(WT_SESSION_IMPL *session)
```
michaelcahill: We don't need a separate function just to set a flag -- I guess I should merge it?

michaelcahill:
@sueloverso, can you please run the test on this branch and let me know if it fires?  If it runs overnight without firing, I think we can call this one done.  Thanks!

src/conn/conn_dhandle.c
sueloverso: Correct, same with the new {{conn_dhandle_clear_open}} they only wrapped the flag manipulation with the exclusive on/off calls.  So those both can go away and the F_SET/F_CLR move back to {{conn_btree_sync_and_close}}

sueloverso:
More to debug as it hit on the 2nd iteration:
```
bengal:format2 115 >> ./t -q
t: process 14360
   1: file, row-store (117 seconds)
[1452719110:691468][14360:0x7fbc785f1700], t, eviction-server: src/evict/evict_lru.c, 1130: F_ISSET(dhandle, WT_DHANDLE_OPEN) && !F_ISSET(btree, WT_BTREE_NO_EVICTION) && dhandle->session_inuse > 0
[1452719110:691505][14360:0x7fbc785f1700], t, eviction-server: aborting WiredTiger library
Abort (core dumped)
```
Both dhandle and btree flags are 0.

michaelcahill:
@sueloverso, thanks, sorry I was slow but I think I finally understood this one -- the eviction server can race with the `NO_EVICTION` flag, which is why it checks it again after getting a mutex.  If we only save handles at that point, I think it should fix the race.

I'm still not reproducing this readily, so apologies for wasting your time.  Can you kick it off again?

keithbostic:
I've got 300 runs on bengal without a failure, I believe this is fixed.

PR #2434, opened 2016-01-14
WT-2321: race between eviction and worker threads on the eviction queue

keithbostic:
WT-2321: race between eviction and worker threads on the eviction queue

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -860,7 +860,7 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
 	WT_DECL_RET;
 	WT_EVICT_ENTRY *evict;
 	uint64_t cutoff;
-	uint32_t candidates, entries, i;
+	uint32_t base, candidates, entries, i;
 
 	cache = S2C(session)->cache;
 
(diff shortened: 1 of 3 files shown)


src/evict/evict_lru.c
```diff
@@ -1385,9 +1385,15 @@ __evict_get_ref(
...
+	/*
+	 * Get the next page queued for eviction: check both the WT_REF and the
+	 * page's flag: they're not set atomically, we can race with the server
+	 * setting them. Check the page's flag last, it's the element set last
+	 * and atomically.
+	 */
 	while ((evict = cache->evict_current) != NULL &&
-	    evict < cache->evict_queue + candidates && evict->ref != NULL) {
+	    evict < cache->evict_queue + candidates && evict->ref != NULL &&
+	    F_ISSET_ATOMIC(evict->ref->page, WT_PAGE_EVICT_LRU)) {
```
agorrod: This makes me a bit nervous. Will a page stay on the eviction queue forever if it is added without the ``WT_PAGE_EVICT_LRU`` flag set? i.e: I'd rather be able to assert that the flag is always set, rather than use it as a condition to break out of here.

keithbostic: > Will a page stay on the eviction queue forever if it is added without the WT_PAGE_EVICT_LRU flag set?

Yes, but adding a page to the queue includes setting the flag, by definition (setting the flag embodies the "publish" of the other fields).

You could spin on the flag instead of skipping buffers where the flag isn't yet set, but I like that less.

agorrod: I'm also surprised there is a race here - we add entries to the evict_queue linked list starting after the highest available candidate (see where ``slot = cache->evict_entries;`` in ``__evict_walk``). A candidate that is added to the queue in ``__evict_init_candidate`` should not be available to ``__evict_get_ref`` until after ``__evict_lru_walk`` updates the ``cache->evict_candidates`` field.

keithbostic:
@agorrod, the eviction and candidate counts could overlap, see 20b7f43 for the fix.

src/evict/evict_lru.c
```diff
@@ -922,13 +922,19 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
...
-		    i++, evict++)
+	/*
+	 * Clear anything more than the maximum tracked number of entries (but
+	 * don't clear past the number of candidates for eviction). This sets
+	 * the base of where new entries are added to the array and checking
+	 * against the candidate count avoids operating in the same space as
+	 * threads removing entries from the array.
+	 */
+	base = WT_MAX(cache->evict_candidates, WT_EVICT_WALK_BASE);
+	if (cache->evict_entries > base) {
```
agorrod: Nice catch.

src/evict/evict_lru.c
agorrod: I'd prefer to switch this from a condition to an assertion inside the loop. If you are happy to make that change I'd be happy to merge.

keithbostic: I don't like that change: I don't want to fail in the field if we introduce this bug again, and it's a reasonable check to make (because setting the flag is the atomic operation that sets the other fields). If you're unconvinced, you should make the change you want and merge, I don't think there's any need for additional review.

agorrod:
@michaelcahill We should talk about what the best approach is for the remaining issue in this pull request - @keithbostic and I have different opinions.

michaelcahill:
Simplified as discussed, this lgtm, merging...

PR #2435, opened 2016-01-14
WT-2331 Require that reference cursors that are joined be positioned

ddanderson:
Additionally add tests to verify that cursors that merely have a key set, or
that have failed a positioning API (search), not be usable in a join.  Give
a better error message in all these cases.

Add a check in __wt_struct_repack so that an empty or shorter than expected
input array give an error, and not be interpreted as unbounded size.


ddanderson:
@agorrod, any chance you can review?  Thanks!

agorrod:

PR #2436, opened 2016-01-14
WT-60: big endian support

keithbostic:
WT-60: big endian support

@agorrod, here's something that should give you an idea of where I'm heading. I figure this one will be contentious enough it's better to get something out for comment quickly.

I've smoke-tested this, and think it handles the block manager code; I haven't looked at the logging code at all.


src/support/hash_city.c
```diff
@@ -86,33 +86,60 @@ static uint32_t UNALIGNED_LOAD32(const char *p) {
...
 #define	bswap_32(x) OSSwapInt32(x)
 #define	bswap_64(x) OSSwapInt64(x)
 
-#elif defined(__sun)
+#elif defined(__sun) || defined(sun)
 
 #include <sys/byteorder.h>
-#define	bswap_32 BSWAP_32
-#define	bswap_64 BSWAP_64
+#define	bswap_32(x) BSWAP_32(x)
```
agorrod: I'm surprised about these changes - I'd thought this would get very simple with ``endian.h``. What am I missing?

agorrod:
@keithbostic The shape of these changes look good to me. I'm happy to talk through details if you'd like otherwise I think it's heading in the right direction.

src/support/hash_city.c
keithbostic: This could be made simple with `endian.h` -- I went this way because I was hesitant to mess with whatever Google thought their portability layer should look like. Maybe there's a reason they prefer these functions to the compiler built-ins? Or maybe not, of course. Anyway, I just don't know, so I took the easy way of using what they're shipping.

keithbostic:
@sueloverso, can you please take a fast look at this branch in terms of what the logging code will need?

michaelcahill:
@keithbostic, likewise, this is looking like the right direction to me.

Will you add code to swap `WT_PAGE_HEADER`?  Are there other structures outside of logging that you're aware of that need to be swapped?

michaelcahill:
One more thing while I think of it.

Presumably we have to disallow `mmap` on big endian machines, which will have some implications for LSM (not necessarily bad, just different to what we've done so far).  Are there other limitations or caveats around big endian support that you are aware of?

keithbostic:
@michaelcahill
> Will you add code to swap WT_PAGE_HEADER?

Yes; I deliberately left that one alone in the first pass, it's going to be messier than the ones I've already done.

Just to confirm: our goal is absolutely identical files on all machines? In other words, we want the raw files to be portable between machines?

> Are there other structures outside of logging that you're aware of that need to be swapped?

I wouldn't bet much on being right, but no, I can't think of anything else, other than logging and the page headers.

keithbostic:
@michaelcahill 

> Presumably we have to disallow mmap on big endian machines, which will have some implications for LSM (not necessarily bad, just different to what we've done so far). Are there other limitations or caveats around big endian support that you are aware of?

I hadn't thought of mmap, actually!

Based on about 10 seconds of thought, I bet we could make mmap work, there isn't much stuff we take from the underlying page image that's byte-swapped, and it's pretty concentrated in a few areas. Do you think it's worth doing that?

agorrod:
> I bet we could make mmap work... Do you think it's worth doing that?

My vote is "no" at least until we are given a reason.

sueloverso:
Clearly logging will need similar changes.  The file header is written in {{log_file_header}} and read in {{log_openfile}}.  The record header is written in {{log_write_internal}} and read in {{wt_log_scan}}.

michaelcahill:
> Just to confirm: our goal is absolutely identical files on all machines? In other words, we want the raw files to be portable between machines?

Yes, and that will be what the test harness has to confirm.

keithbostic:
@michaelcahill, are we expecting the base configuration file to be portable?

Something I just ran across: if the base config file has `buffer_alignment=512` in it, that won't port to a system without a `posix_memalign` call.

We could ignore `buffer_alignment` in the base configuration file (or not put it there to start with).

If we want the base configuration file to be portable, that expands our testing problems.

michaelcahill:
@keithbostic, first up, MongoDB disables the basecfg file, so this is not an issue for MongoDB.

I expect the basecfg file to be portable in the sense that it can be read on every platform.  That should be straightforward because it is a text file.

I don't expect every basecfg configuration setting to be correct for every platform.  A simple example (and the original motivation for basecfg) is absolute paths to extension modules.

We should document that when taking a backup or otherwise copying a database directory from one system to another, the original database configuration stored in the basecfg file should be checked for validity on the new system.  To avoid portability issues, either (a) choose the configuration carefully when the database is first created, (b) don't copy the basecfg file across systems, (c) override it with settings in `WiredTiger.config` or (d, least desirable) edit it by hand.

michaelcahill:
P.S. It's also worth noting that the upgrade tests have special handling of basecfg for the same reason (that basecfg files created by version N+1 may not be meaningful to version N).

keithbostic:
@agorrod, @michaelcahill, @sueloverso: I think this one is working.

I still have some additional smoke-testing to do, but once that's done I'll be ready to merge.

I only lightly tested compression (either mainline or logging). As expected, none of the compression engines (I checked snappy, lz4 and zlib) write byte-independent files, so it's hard to know if something is broken because we're getting it wrong or the the underlying compression engine can't parse its data.

Sue, would you please review the changes I made in the logging code? I'm particularly concerned about changing `__log_fill` to use the WT_ITEM size instead of the log record size.

Alex, Michael: can we talk over what "real" testing we'd like to do?

keithbostic:
@agorrod, @michaelcahill, I can now move a variety of databases to/from big/little-endian systems.

Here's the script that builds a list of databases to test (please let me know if you can think of anything else to include).
```
#! /bin/sh

d=LE
i=0

rm -f $d.tgz
rm -rf $d && mkdir $d

arg1="-1 -c . -q rebalance=0 salvage=0 verify=0"
arg2="data_source=table ops=100000 rows=10000 threads=4"

args="$arg1 $arg2 compression=none logging=0"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=fix
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row data_source=lsm
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=var

args="$arg1 $arg2 compression=none logging=0 huffman_key=1 huffman_value=1"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

args="$arg1 $arg2 compression=none logging=0 reverse=1"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

args="$arg1 $arg2 compression=none logging=0 checksum=on"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

args="$arg1 $arg2 compression=none logging=0 encryption=rotn"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

args="$arg1 $arg2 compression=lz4 logging=0"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row
args="$arg1 $arg2 compression=snappy logging=0"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row
args="$arg1 $arg2 compression=zlib logging=0"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

args="$arg1 $arg2 logging=1 logging_prealloc=1"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

args="$arg1 $arg2 compression=lz4 logging=1 logging_compression=lz4"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row
args="$arg1 $arg2 compression=snappy logging=1 logging_compression=snappy"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row
args="$arg1 $arg2 compression=zlib logging=1 logging_compression=zlib"
i=$(($i + 1)); ./t -h $d/RUNDIR.$i $args file_type=row

tar cf - $d | gzip > $d.tgz
```
Here's the test script I've been using:
```
#! /bin/sh

rm -rf BE && cat BE.tgz | gunzip | tar xf -

run()
{
	echo "$1 ..."

	wt -h $1 list -v > /dev/null
	wt -h $1 verify table:wt
	wt -h $1 dump table:wt > /dev/null

	if test -f "$1/*Log*"; then
		wt -h $1 printlog > /dev/null
	fi
}

if test "$#" -eq 0; then
	for i in BE/*; do
		run $i
	done
else
	run BE/RUNDIR.$1
fi
```

keithbostic:
@agorrod, @michaelcahill: if you're happy with this branch, I'm ready to merge it.

src/log/log.c
```diff
@@ -602,13 +600,24 @@ __log_file_header(
...
+	 * big-endian format, swap it into place in a separate step.
 	 */
 	logrec->len = log->allocsize;
 	logrec->checksum = 0;
+	__wt_log_record_byteswap(logrec);
 	logrec->checksum = __wt_cksum(logrec, log->allocsize);
+#ifdef WORDS_BIGENDIAN
+	logrec->checksum = __wt_bswap32(logrec->checksum);
+#endif
+	buf->size = log->allocsize;
```
sueloverso: Can you move this up to the code where we allocate and initialize `buf`?   Visually, I missed this line when I first read this (both diff and real code).

keithbostic: @sueloverso 
> Can you move this up to the code where we allocate and initialize buf?

Sure, will do.

sueloverso:
I'm good to go with this when @agorrod and @michaelcahill are ready.

michaelcahill:
Thanks, @keithbostic, lgtm.  I'll merge.  There is WT-2337 for automating the testing.

PR #2437, opened 2016-01-15
WT-2326 - Add back a nullbyte that was incorrectly removed from wtperf

daveh86:


daveh86:
@agorrod, It appears that I introduced a bug earlier in WT-2326. Can you review this?

daveh86:
@keithbostic, understand the reluctance here as I had misunderstood the code myself first pass through - thus my going back and replacing this line of code. This is currently causing the perf-stress failures seen [here](http://build.wiredtiger.com:8080/job/wiredtiger-perf-stress/350/console). Search and you can see the line `invalid compression configuration: snappy"` is the cause.

This line is a truncation of the string to remove the last quotation mark. Consider the string entry 
`compressor="snappy"`, the duplicated portion of this string is `snappy"`. We need to remove that last quotation mark, thus adding the nullbyte at v->len (rather than v->len +1)

daveh86:
Added a comment to make it clear that this is a truncation.

agorrod:
@daveh86 A better way to handle this would be to use ``strndup``, explicitly adding a ``NULL`` terminator after a call to ``strdup`` is a danger sign that there is an issue in the code, even if it's reasonable.

daveh86:
Thanks @agorrod, thats good to know. I've switched this patch around accordingly.

bench/wtperf/config.c
```diff
@@ -442,7 +442,11 @@ config_opt(CONFIG *cfg, WT_CONFIG_ITEM *k, WT_CONFIG_ITEM *v)
 		}
 		strp = (char **)valueloc;
 		free(*strp);
-		newstr = dstrdup(v->str);
+		/*
+		 * We duplicate the string to len rather than len+1 as we want
+		 * to truncate the trailing quotation mark.
+		 */
+		newstr = strndup(v->str,  v->len);
```
agorrod: The use of ``strndup`` should follow the same pattern as ``strdup`` - i.e: create a wrapper function that handles errors.

agorrod:
lgtm - I'll merge.

PR #2438, opened 2016-01-17
WT-2335: crash in config_check_search with invalid configuration string

keithbostic:
WT-2335: crash in config_check_search with invalid configuration string

@michaelcahill, would you please do the review?


agorrod:
@keithbostic I'd find this one easier to verify if we had a test case. Is it possible to test this failure mode with a Python test case?

michaelcahill:

PR #2439, opened 2016-01-18
WT-1801: Add a directory sync after rollback of a WT_SESSION::rename operation

keithbostic:
WT-1801: Add a directory sync after rollback of a WT_SESSION::rename operation

@michaelcahill, @sueloverso, for your review/consideration.


michaelcahill:
Thanks, @keithbostic -- this is definitely better than what we're doing now for renames.  LGTM, I'll merge...

PR #2441, opened 2016-01-19
WT-1517: schema format edge cases

keithbostic:
WT-1517: schema format edge cases

@michaelcahill, a branch for the fixes for WT-1517.

### DIFF
--- a/lang/python/wiredtiger/packing.py
+++ b/lang/python/wiredtiger/packing.py
@@ -70,6 +70,8 @@ def __unpack_iter_fmt(fmt):
             size = (size * 10) + int(char)
             havesize = 1
         else:
+            if not havesize:
+                size = 1
             yield offset, havesize, size, char
             size = 0
             havesize = 0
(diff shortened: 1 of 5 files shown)


keithbostic:
@michaelcahill, @ddanderson:

I think these changes fix the WiredTiger library, but I'm suspicious there's a bug in the Python packing/unpacking code, and I have no clue how to start hunting for that.

I've commented out a test in test_config06 that demonstrates it: basically, if I open a table with `key_format=s` (so, the default 's' format, should be the same as '1s'), then manipulate a Python cursor with `key="ABCDEFG", cursor[key[:1]]`, it fails, while an explicit `key_format=1s` works. Looking inside the library, a string is being passed in as the key, not a single byte.

I wrote the obvious C test program using `key_format=s`, and it works just fine, with or without the leading digit.

Given all the Python packing code, maybe the 's' format can't be tested in the test suite, anyway?

keithbostic:
autobuild: retest this please

michaelcahill:
Thanks, @keithbostic -- I fixed the Python test, added a trivial C test and simplified where we set `size = 1`.  This lgtm, I'll merge.

PR #2442, opened 2016-01-19
WT-2295 On index create, do not populate existing indices.

ddanderson:
These commits fix the test case for populating indices attached to WT-2295.
Also fixes exclusive create for "index:", and disallows calling create on an existing index with a different configuration string.

Added explicit tests for exclusive create and recreating index with a different configuration.  It's probably not possible in python to create a test that detects that existing indices are not repopulated, at least not without help in C.

### DIFF
--- a/src/schema/schema_create.c
+++ b/src/schema/schema_create.c
@@ -595,8 +595,11 @@ __create_table(WT_SESSION_IMPL *session,
 		return (EINVAL);
 
 	if ((ret = __wt_schema_get_table(session,
-	    tablename, strlen(tablename), false, &table)) == 0)
+	    tablename, strlen(tablename), false, &table)) == 0) {
+		if (exclusive)
+			WT_ERR(EEXIST);
 		exists = true;
+	}
 	WT_RET_NOTFOUND_OK(ret);
 
 	WT_ERR(__wt_config_gets(session, cfg, "colgroups", &cval));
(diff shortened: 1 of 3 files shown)


ddanderson:
@michaelcahill, can you review this, please?  This follows your suggestion from #2431, and also prevents recreating an index with a different configuration.

src/schema/schema_create.c
```diff
@@ -403,6 +403,14 @@ __create_index(WT_SESSION_IMPL *session,
 		WT_RET_MSG(session, EINVAL, "Invalid index name, "
 		    "should be <table name>:<index name>: %s", name);
 
+	/* Check if the file already exists. */
+	if ((ret = __wt_metadata_search(session, name, &origconf)) !=
+	    WT_NOTFOUND) {
+		if (exclusive)
+			WT_TRET(EEXIST);
+		exists = true;
+	}
```
michaelcahill: I think we need a `WT_ERR(ret);` here -- otherwise we're assuming `origconf` is valid below...

michaelcahill:
Thanks, @ddanderson, this looks much better and also fixes exclusive index creates.  I made one comment: I'll fix the error handling later tonight if you don't beat me to it.  Then this one lgtm.

michaelcahill:
Thanks, @ddanderson, I took the opportunity to review handling of exclusive for tables and column groups and tried to get them as close as possible.  Note that I didn't touch file creates because the metadata doesn't stay constant after creation (checkpoint information is added).

lgtm now, I'll merge.

PR #2443, opened 2016-01-19
WT-2332 Add logging guarantee assertions.

sueloverso:
@michaelcahill Please review this change.  Although 2.7 and develop don't have the original bug, I believe adding in the assertions from this ticket are worthwhile.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1970,10 +1970,13 @@ err:
 	    myslot.slot != NULL)
 		ret = myslot.slot->slot_error;
 
-	if (LF_ISSET(WT_LOG_FLUSH))
-		WT_ASSERT(session, __wt_log_cmp(&log->write_lsn, &lsn) >= 0);
-	if (LF_ISSET(WT_LOG_FSYNC))
-		WT_ASSERT(session, __wt_log_cmp(&log->sync_lsn, &lsn) >= 0);
+	/*
+	 * If one of the sync flags is set, assert the proper LSN has moved to match.
+	 */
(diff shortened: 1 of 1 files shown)


src/log/log.c
```diff
@@ -1970,6 +1970,10 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
 	    myslot.slot != NULL)
 		ret = myslot.slot->slot_error;
 
+	if (LF_ISSET(WT_LOG_FLUSH))
+		WT_ASSERT(session, __wt_log_cmp(&log->write_lsn, &lsn) >= 0);
+	if (LF_ISSET(WT_LOG_FSYNC))
+		WT_ASSERT(session, __wt_log_cmp(&log->sync_lsn, &lsn) >= 0);
```
agorrod: @sueloverso I'm uncomfortable with ``if`` statements that only contain an assertion. I know it works with how we define ``WT_ASSERT`` in non-diagnostic builds, but it looks odd to me. Could these be refactored like:

``
WT_ASSERT(session, !LF_ISSET(WT_LOG_FLUSH) || __wt_log_cmp(&log->write_lsn, &lsn) >= 0);
``

To avoid the ``if`` statement? I know that makes the logic less clear (probably to the point where a comment is warranted). This is just a suggestion - if you prefer it as-is this is good to merge.

sueloverso: I like that idea and have pushed that fix.  Once it says it can be merged I'll do so.

PR #2444, opened 2016-01-19
WT-2296 Yield instead of sleep if the wrlsn thread is finding work.

sueloverso:
@michaelcahill Please review this change that eliminates the performance hit in the 2.7 release for write-no-sync mode.  In fact, this code, in write-no-sync, now runs at 3x the (fixed from WT-2332) 2.6 code (on Linux, it runs equal on my Mac).  This may have a good impact on in-memory log configuration too because the change is in the `wrlsn_server` thread.  This yield code existed in 2.6 and earlier logging code and changed during the rewrite.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -511,7 +511,7 @@ typedef struct {
  *	write_lsn in LSN order after the buffer is written to the log file.
  */
 int
-__wt_log_wrlsn(WT_SESSION_IMPL *session)
+__wt_log_wrlsn(WT_SESSION_IMPL *session, int *yield)
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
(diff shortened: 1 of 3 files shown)


src/conn/conn_log.c
```diff
@@ -660,22 +662,30 @@ __log_wrlsn_server(void *arg)
...
 		/*
 		 * Write out any log record buffers.
 		 */
-		WT_ERR(__wt_log_wrlsn(session));
-		WT_ERR(__wt_cond_wait(session, conn->log_wrlsn_cond, 10000));
+		WT_ERR(__wt_log_wrlsn(session, &yield));
+		/*
+		 * If __wt_log_wrlsn did work we want to yield instead of sleep.
+		 */
+		if (yield++ < 1000)
```
agorrod: I don't understand why the count is here. Is there a reason to limit the number of times we yield to 1000?

sueloverso: First it restores the original code that was in the earlier logging code.  There are a lot of places that yield for 1000 loops and then sleep.  In this case, we've had complaints about worker threads consuming too much CPU.  I might be inclined to also restore the original 100000 `wt_cond_wait` timeout time as well.

agorrod:
@sueloverso I'm happy with this change now - should I go ahead and merge?

PR #2445, opened 2016-01-19
WT-2338 Skip pre-allocated log files if backup cursor is open.

sueloverso:
@michaelcahill Please review this fix for WT-2338 which is from BF-1399 - skip using pre-allocated log files if a backup is in progress.  This does not include a test for forcing that specific scenario.  I can add one if you want.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -694,12 +694,12 @@ __log_server(void *arg)
 	WT_LOG *log;
 	WT_SESSION_IMPL *session;
 	int freq_per_sec;
-	bool signalled;
+	bool locked, signalled;
 
 	session = arg;
 	conn = S2C(session);
 	log = conn->log;
-	signalled = false;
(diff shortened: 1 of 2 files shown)


sueloverso:
@michaelcahill Don't merge this yet.  There is another path that needs thought and fixing.  In addition to skipping using pre-allocated files, there is a path where we rename the tmp file to the pre-allocated file after the header is written and synced.  We should talk about this path.

sueloverso:
@agorrod or @michaelcahill I believe this is ready to merge now.

agorrod:
@sueloverso I added a comment. I'm happy for this change to go in (lgtm). Would it be possible to test this with a Python test?

sueloverso:
I agree a test would be helpful.  I created WT-2343 for a more general test and adding assertions in rename and remove.  I think the test would be better under that ticket so we can get the BF ticket resolved.

PR #2446, opened 2016-01-20
WT-2287: WT_SESSION.rebalance

keithbostic:
WT-2287: WT_SESSION.rebalance


agorrod:
Thanks @keithbostic lgtm - I'll merge once it finishes testing.

PR #2447, opened 2016-01-20
WT-2275 Restore call to LSM tree truncate.

sueloverso:
@michaelcahill In thinking harder about the LSM truncate and looking at the code, I restored the call to `wt_lsm_tree_truncate`.  The LSM code doesn't do truncate and file manipulation like the others.  While there may be bugs, the LSM code is very different where the truncate sets up a new chunk and moves all the previous chunks to the `old_chunks` list and then does its processing on those chunks.  So if there are bugs in the metadata consistency, it is unlike the original truncate code for files and it exists for any LSM table when it removes chunks from the `old_chunks` array.

There is precedent for checking the `lsm:` URI in `session_api.c` and `session_compact.c` also.

### DIFF
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -836,6 +836,13 @@ __session_truncate(WT_SESSION *wt_session,
 				    "target after the log: URI prefix.");
 			ret = __wt_log_truncate_files(session, start, cfg);
 			goto done;
+		} else if (WT_PREFIX_MATCH(uri, "lsm:")) {
+			/*
+			 * LSM tree truncate needs to handle truncating each
+			 * chunk on its own.
+			 */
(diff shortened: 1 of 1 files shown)


michaelcahill:
@sueloverso, thanks for this -- it would clearly help the case where someone truncates an "lsm:foo" source directly.

Unfortunately, I think there's still a problem with truncating a "table:foo" or "index:foo" of type LSM -- in that case, we'll still map that to cursor truncates.  I think we need more of the original `__wt_schema_truncate` machinery to make this work, where we only map `__truncate_file` onto a range truncate.

michaelcahill:
Closed in favor of #2454.

PR #2448, opened 2016-01-21
WT-2339: format post-rebalance verify failure

keithbostic:
WT-2339: format post-rebalance verify failure


agorrod:
Thanks @keithbostic lgtm. I'll merge once testing has finished.

PR #2450, opened 2016-01-21
WT-2333 Add a "no_wait" mode to WT_SESSION::drop.

michaelcahill:


michaelcahill:
The ugliest part of this change is that `WT_WITH_LOCK` now touches `ret` -- ideally there would be a better solution.

I haven't yet figured out how to test this: obviously I can pass in `lock_wait=false` to `WT_SESSION::drop` and make sure it is accepted, but forcing it to fail is harder.

dist/api_data.py
```diff
@@ -771,6 +771,9 @@ def __cmp__(self, other):
     Config('remove_files', 'true', r'''
         should the underlying files be removed?''',
         type='boolean'),
+    Config('lock_wait', 'true', r'''
+        wait for locks, if false will return \c EBUSY instead of blocking''',
```
agorrod: This implies that it won't return EBUSY if you don't specify ``lock_wait``, I don't think that's true (just much less likely)

agorrod:
@michaelcahill I agree re: the macro structure being worse after this change, but I don't see a better solution.

This lgtm modulo clarifying the comment in the api.

dist/api_data.py
michaelcahill: You're right, we certainly can return `EBUSY` without this change.  I'll try to clarify the wording...

michaelcahill:
Thanks, @agorrod, I'll merge to get this into testing.

PR #2451, opened 2016-01-21
WT-2342 Add support for background schema operations in wtperf

agorrod:


agorrod:
@michaelcahill This branch has changes to wtperf that should allow you to test out fixes for WT-2264. It includes a wtperf configuration file that regularly fails with a create taking over 5 seconds. I suspect the exact configuration required will vary depending on the machine being used to run the test.

michaelcahill:
Closing in favor of #2452.

PR #2452, opened 2016-01-21
WT-2342 Add background create/drop to wtperf.

agorrod:
Works on idle tables. Gives us a way to ensure that schema operations don't get blocked during high throughput workloads.

### DIFF
--- a/bench/wtperf/idle_table_cycle.c
+++ b/bench/wtperf/idle_table_cycle.c
@@ -1,0 +1,177 @@
+/*-
+ * Public Domain 2014-2016 MongoDB, Inc.
+ * Public Domain 2008-2014 WiredTiger, Inc.
+ *
+ * This is free and unencumbered software released into the public domain.
+ *
+ * Anyone is free to copy, modify, publish, use, compile, sell, or
+ * distribute this software, either in source code form or as a compiled
+ * binary, for any purpose, commercial or non-commercial, and by any
+ * means.
+ *
+ * In jurisdictions that recognize copyright laws, the author or authors
+ * of this software dedicate any and all copyright interest in the
+ * software to the public domain. We make this dedication for the benefit
+ * of the public at large and to the detriment of our heirs and
+ * successors. We intend this dedication to be an overt act of
+ * relinquishment in perpetuity of all present and future rights to this
+ * software under copyright law.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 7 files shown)


agorrod:
@michaelcahill Could you do a review for me please?

bench/wtperf/wtperf.h
```diff
@@ -187,6 +187,9 @@ struct __config {			/* Configuration structure */
 	volatile int stop;		/* notify threads to stop */
 	volatile int in_warmup;		/* Running warmup phase */
 
+	pthread_t idle_table_cycle_thread; /* Thread ID of idle cycle thread */
+	volatile int idle_cycle_run;	/* Signal idle cycle thread */
```
michaelcahill: Minor point, but it looks like this could be a `bool`?

bench/wtperf/idle_table_cycle.c
```diff
@@ -0,0 +1,177 @@
...
+			return (NULL);
+		start = stop;
+
+		/* Drop the table. */
+		if ((ret = session->drop(session, uri, NULL)) != 0) {
+			lprintf(cfg, ret, 0,
+			     "Table drop failed in cycle_idle_tables.");
+			cfg->error = ret;
+			return (NULL);
+		}
```
michaelcahill: Is `EBUSY` from drop reasonable here, given that there are checkpoints?

bench/wtperf/runners/checkpoint_stress_schema_ops.wtperf
```diff
@@ -0,0 +1,19 @@
...
+# Setup a cache size large enough that checkpoints can take a long time.
+conn_config="cache_size=8GB,log=(enabled=false),checkpoint=(wait=30)"
+table_config="leaf_page_max=4k,internal_page_max=16k,type=file"
+icount=10000000
+table_count_idle=100
+# Turn on create/drop of idle tables, and error if a single operation takes
+# more than 5 seconds.
+idle_table_cycle=5
+populate_threads=5
+checkpoint_threads=0
```
michaelcahill: For my benefit: when do you choose a wtperf checkpoint thread vs a WT library checkpoint thread?

michaelcahill:
@agorrod, I've made a few comments, but this LGTM.

bench/wtperf/runners/checkpoint_stress_schema_ops.wtperf
agorrod: The wtperf checkpoint functionality pre-dates the WiredTiger library checkpoint thread. The only time the wtperf functionality is necessary is if you want multiple threads checkpointing, which is more interesting for stress testing than benchmarking. I'd be happy to remove the functionality from wtperf.

sueloverso: The other use is for the monitoring output.  It makes note of when checkpoints are running so that they can be noted in the monitor file and correlated to that performance.  I don't think we want to lose that.

PR #2453, opened 2016-01-21
SERVER-22069 Better seeding of WT's random number generator

keithbostic:
SERVER-22069: Better seeding of WT's random number generator

### DIFF
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -1376,8 +1376,15 @@ __open_session(WT_CONNECTION_IMPL *conn,
 
 	WT_ERR(__wt_cond_alloc(session, "session", false, &session_ret->cond));
 
-	if (WT_SESSION_FIRST_USE(session_ret))
-		__wt_random_init(&session_ret->rnd);
+	/*
+	 * The first use of a session initializes its PRNG. Use the refreshed
+	 * value from the connection's default session so sessions don't start
+	 * in lock step.
+	 */
+	if (WT_SESSION_FIRST_USE(session_ret)) {
+		(void)__wt_random(&conn->default_session->rnd);
+		session_ret->rnd = conn->default_session->rnd;
+	}
 
 	__wt_event_handler_set(session_ret,
 	    event_handler == NULL ? session->event_handler : event_handler);
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -1382,7 +1382,8 @@ __open_session(WT_CONNECTION_IMPL *conn,
 	 * in lock step.
 	 */
 	if (WT_SESSION_FIRST_USE(session_ret)) {
-		(void)__wt_random(&conn->default_session->rnd);
+		for (i = 0; i < 100; ++i)
+			(void)__wt_random(&conn->default_session->rnd);
 		session_ret->rnd = conn->default_session->rnd;
 	}
 


michaelcahill:
My understanding is that with this change, each session will still generate the same sequence, but they will now start one step apart.  Is that correct?

If so, why does that help: won't session N+1 generate the same sequence as session N apart from the first and last keys?

agorrod:
@keithbostic Is there some way we could gain confidence that using a random seed in each session is reasonable instead of always starting with the same seed? I know we have verified the current seed value, but using a single seed every time doesn't feel very random to me.

keithbostic:
@michaelcahill 
> My understanding is that with this change, each session will still generate the same sequence, but they will now start one step apart. Is that correct?

Yes, that's correct.

> If so, why does that help: won't session N+1 generate the same sequence as session N apart from the first and last keys?

I can just tell you that, experimentally, it introduces enough variation that we no longer get lockstep from the sessions.

I initially did a version that both stepped further down the sequence and tried to randomize how far down the sequence it would step; that version worked, too, but no better than this change, which is obviously simpler.

I believe this is mostly a "benchmark" problem: the way to get this failure is to create new sessions (so they have the same initial random number state), and perform identical operations in a newly created tree. If the sessions have been used for other purposes, or we're not working with a single page's in-memory skiplist, I don't believe the problem will reproduce. However, it's not too surprising a user would do something like that when experimenting with a feature, and be surprised when the results are less than random.

keithbostic:
@agorrod 
> Is there some way we could gain confidence that using a random seed in each session is reasonable instead of always starting with the same seed? I know we have verified the current seed value, but using a single seed every time doesn't feel very random to me.

The only way I know of to get confidence is to run the random seed in the algorithm for an appropriate period of time (however long you plan to use the seed), looking for cycles and measuring the uniformity of the bits.

I think there are two choices: either use a known good seed and let it run for as long a period as possible), or use potentially bad seeds and don't let them run for a long period so a bad seed won't affect the system for very long.

We currently do the former, we have a known good seed with a long period that produces reasonably uniform bits we can use for mod operations (my primary concern with our PRNG is selecting skip list depths).

I believe it's better to continue to do the former, for two reasons:

First, the latter means we have to figure out some fast way to generate random seeds (for example, using the system clock, is known to be a bad idea, because it has an obvious pattern). There are system sources we could use to create a key (Intel chips have assembly instructions, or FreeBSD's /dev/random, or Windows' CryptGenRandom, every system has something), but I have no clue how fast or good they are, there's portability and measuring work to be done there. We could probably generate a table of 10K seeds, and test them, but again, there's work there.

Second, the cost of a bad seed is real for us: if a session inserts 100K key/value pairs, all with a skiplist depth insert of 1, that's bad. With an untested seed, that's always a possibility.

keithbostic:
@michaelcahill, I did some additional testing with larger initial insert skiplists, and moving further down the sequence of random numbers during each session create performs better than moving a single slot. I've updated the change to reflect that.

michaelcahill:
@keithbostic, thanks for the update.  Are you using the test case from SERVER-22069?  If not, can we add your test case to the tree somewhere?

keithbostic:
I'm using the test case from SERVER-22069.

@agorrod wrote `test_cursor_random02.py` that checks for randomness when searching insert lists, we could probably adapt it to open multiple session handles, if you want.

cswanson310:
Sorry to jump in here so late (I meant to do so earlier), but if I understand this patch correctly, it would alleviate the problem to a certain extent, but I don't think it addresses the fundamental problem. If you increase the number of iterations in my test case from 10 to 1000, then we'll probably still see a skewed distribution.

As a simplified example to illustrate the point, assume the following things:
* We have 10 sessions open
* Random walks are the only things happening on these sessions, and we repeatedly ask for new random records
* Each random walk will only ask for one random number from the PRNG (probably would ask for a new one at each level of the tree, but this assumption simplifies things, so bear with me)
* There are no writes are going on

It's my understanding that the second session created will start with a PRNG state which has advanced 100 more times than the first session's. If this is the case, then the 101st record returned by the first session will be the same as the 1st record returned by the second session. That means that this document has a skewed probability of being selected via a random walk.

More broadly, if we ask for enough random walks, the 1001st random walk on the first session is guaranteed to return a document that has already been seen by another session (or will be seen if enough are requested). So if we just kept asking for random records from each session, we would see that record 10 times.

Is my understanding correct?

keithbostic:
> If you increase the number of iterations in my test case from 10 to 1000, then we'll probably still see a skewed distribution.

I think that's probably correct, however, the tree shape and amount of serialization and isolation required to trigger the problem makes me doubt it's worth fixing.

Regardless, @michaelcahill has decided to take a different approach from this branch, so we're going in a different direction; I expect the solution he chooses will be a better overall fix.

PR #2454, opened 2016-01-21
WT-2275 Convert a file URI truncate into a range truncate.

sueloverso:
@michaelcahill Here is a new branch that only converts only the file URI truncate into a range truncate and leaves LSM and tables alone.  If this is the way to go, the other two branches/pull requests can be closed and removed.


michaelcahill:
Thanks, @sueloverso -- as discussed, this LGTM now.  It would be great to avoid the schema and checkpoint locks entirely, but let's leave that as an exercise for a future reader.  I'll merge.

PR #2455, opened 2016-01-22
WT-2130 Don't round the split_pct to an allocation size.

agorrod:
The value is passed in via the API, and rounding makes it likely we
won't split as the user expects. Especially if the split percentage
is high, when we'll likely round up to 100%.

### DIFF
--- a/src/include/misc.h
+++ b/src/include/misc.h
@@ -49,7 +49,7 @@
 	((((uintmax_t)(n)) + ((v) - 1)) & ~(((uintmax_t)(v)) - 1))
 
 #define	WT_ALIGN_NEAREST(n, v)						\
-	((((uintmax_t)(n)) + ((v) / 2 + 1)) & ~(((uintmax_t)(v)) - 1))
+	((((uintmax_t)(n)) + ((v) / 2)) & ~(((uintmax_t)(v)) - 1))
 
 /* Min, max. */
 #define	WT_MIN(a, b)	((a) < (b) ? (a) : (b))
--- a/src/include/misc.h
+++ b/src/include/misc.h
@@ -48,6 +48,9 @@
 #define	WT_ALIGN(n, v)							\
 	((((uintmax_t)(n)) + ((v) - 1)) & ~(((uintmax_t)(v)) - 1))
 
+#define	WT_ALIGN_NEAREST(n, v)						\
+	((((uintmax_t)(n)) + ((v) / 2 + 1)) & ~(((uintmax_t)(v)) - 1))
+
 /* Min, max. */
 #define	WT_MIN(a, b)	((a) < (b) ? (a) : (b))
 #define	WT_MAX(a, b)	((a) < (b) ? (b) : (a))
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1889,7 +1889,15 @@ __wt_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
 	 * we don't waste space when we write).
 	 */
 	a = maxpagesize;			/* Don't overflow. */
-	split_size = (uint32_t)((a * (u_int)btree->split_pct) / 100);
+	split_size = (uint32_t)WT_ALIGN_NEAREST(
+	    (a * (u_int)btree->split_pct) / 100, btree->allocsize);
+
+	/*
+	 * Respect the configured split percentage if the calculated split
+	 * size is either zero or a full page.
+	 */
+	if (split_size == 0 || split_size == maxpagesize)
+		split_size = (uint32_t)((a * (u_int)btree->split_pct) / 100);
 
 	return (split_size);
 }
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1889,8 +1889,7 @@ __wt_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
 	 * we don't waste space when we write).
 	 */
 	a = maxpagesize;			/* Don't overflow. */
-	split_size = (uint32_t)
-	    WT_ALIGN((a * (u_int)btree->split_pct) / 100, btree->allocsize);
+	split_size = (uint32_t)((a * (u_int)btree->split_pct) / 100);
 
 	/*
 	 * If the result of that calculation is the same as the allocation unit
@@ -3286,6 +3285,16 @@ supd_check_complete:
 		}
 	}
 
+	bnd->entries = r->entries;
+	/* Output a verbose message if we create a page without many entries */
+	if (WT_VERBOSE_ISSET(session, WT_VERB_SPLIT) && r->entries < 6)
+		WT_ERR(__wt_verbose(session, WT_VERB_SPLIT,
+		    "Reconciling page with memory footprint %" PRIu64
+		    ", %" PRIu32 " th page has"
+		    " %" PRIu32 " entries, %s, split state: %d\n",
+		    r->page->memory_footprint, r->bnd_next, r->entries,
+		    F_ISSET(r, WT_EVICTING) ? "evict" : "checkpoint",
+		    r->bnd_state));
 	WT_ERR(__wt_bt_write(session,
 	    buf, addr, &addr_size, false, bnd->already_compressed));
 	WT_ERR(__wt_strndup(session, addr, addr_size, &bnd->addr.addr));
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1894,7 +1894,11 @@ __wt_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
 
 	/*
 	 * Respect the configured split percentage if the calculated split
-	 * size is either zero or a full page.
+	 * size is either zero or a full page. The user has either configured
+	 * an allocation size that matches the page size, or a split
+	 * percentage that is close to zero or one hundred. Rounding is going
+	 * to provide a worse outcome than having a split point that doesn't
+	 * fall on an allocation size boundary in those cases.
 	 */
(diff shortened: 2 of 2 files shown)


agorrod:
@keithbostic You mentioned in the JIRA ticket that you don't think this is the right approach. Could you create a workload for me where it causes a behavior regression?

keithbostic:
@agorrod, I have an idea I think might give both of us what we want.

What if we deprecate `split_pct` in favor of a new value, `split_size`, and require it be a multiple of the allocation size less than the maximum page size?

That way, I get the maximally full blocks I want and you get clarity.

MongoDB probably sets `split_size=28k`.

keithbostic:
@agorrod 
> Could you create a workload for me where it causes a behavior regression?

We'll see the wasted space if we increase the allocation size:
```
import wiredtiger, wttest
class f(wttest.WiredTigerTestCase):
    def test_f(self):
        uri = 'file:f'
        nentries = 100000 
        self.session.create(uri,
            'key_format=S,value_format=S,' +
            'allocation_size=32k,internal_page_max=64k,leaf_page_max=64k')
        cursor = self.session.open_cursor(uri, None, None)
        for i in range(0, nentries):
            cursor['key' + str(i)] = 'value' + str(i)
  
if __name__ == '__main__':
    wttest.run()
```
In the develop branch, the created file is 2.22MB, and in this branch, the created file is 2.9MB, 30% larger.

michaelcahill:
@keithbostic, it seems like we could address your concern with documentation by pointing out that choosing `split_pct` equal to a block boundary can help create more compact files.

If a developer genuinely wants 32KB pages with a 90% split (e.g., because they know something about how their data compresses, or want to make sure there is never less than 3.2KB of data per 4KB block), why shouldn't we allow that?

agorrod:
I think I'll put together a little design document for this, and outline:
* How the ``split_pct`` is calculated from the configured value.
* What the ``split_pct`` currently does inside WiredTiger.
* Suggest a short term, back portable fix for current use case.
* Explore some use cases and potential medium term fixes that will make ``split_pct`` simpler to tune.

keithbostic:
@michaelcahill

Having applications select `split_pct` equal to a block boundary is difficult (as @agorrod said), targeting 28K with a 32K max leaf page size means selecting a split percentage of 87.5%.

I think it's impossible for an application to be that precise about their split percentages: the application can't predict how WT will layout the data on the page, and that layout can change without notice. Compression, of course, makes the problem much worse.

It seems to me if an application writes 90% of a 32KB chunk, it's creating a 32KB page with some empty space, so when it gets read back in, it can grow to 32KB without further splits. But if I split to 28KB instead, I get the same effect on subsequent reads/updates without wasting any space, and I'm struggling to imagine a workload where the application could lay out its data such that the additional ~1KB of data written on the initial write would justify the wasted 3KB of space.

agorrod:
@keithbostic and @michaelcahill would you mind reviewing this please?

src/include/misc.h
```diff
@@ -48,6 +48,9 @@
 #define	WT_ALIGN(n, v)							\
 	((((uintmax_t)(n)) + ((v) - 1)) & ~(((uintmax_t)(v)) - 1))
 
+#define	WT_ALIGN_NEAREST(n, v)						\
+	((((uintmax_t)(n)) + ((v) / 2 + 1)) & ~(((uintmax_t)(v)) - 1))
+
```
keithbostic: @agorrod, I can't imagine it has any practical significance, but shouldn't this be:
```
((((uintmax_t)(n)) + ((v) / 2)) & ~(((uintmax_t)(v)) - 1))
```

Hurts my head to think about it, so here's the test program:
```
% cat t.c
#include <inttypes.h>
#include <stdio.h>
#include <stdlib.h>

#define	WT_ALIGN_NEAREST_ORIG(n, v)					\
    ((((uintmax_t)(n)) + ((v) / 2 + 1)) & ~(((uintmax_t)(v)) - 1))
#define	WT_ALIGN_NEAREST_FIX(n, v)					\
    ((((uintmax_t)(n)) + ((v) / 2)) & ~(((uintmax_t)(v)) - 1))

int main(int argc, char **argv)
{
	int a;

	a = atoi(argv[1]);
	printf("%d -> orig %d, fix %d\n", a,
	    (int)WT_ALIGN_NEAREST_ORIG(a, 1024),
	    (int)WT_ALIGN_NEAREST_FIX(a, 1024));
	return 0;
}
% cc t.c
% ./a.out 510
510 -> orig 0, fix 0
% ./a.out 511
511 -> orig 1024, fix 0
% ./a.out 512
512 -> orig 1024, fix 1024
% ./a.out 513
513 -> orig 1024, fix 1024
```

src/reconcile/rec_write.c
```diff
@@ -1889,15 +1889,14 @@ __wt_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
...
 	/*
-	 * If the result of that calculation is the same as the allocation unit
-	 * (that happens if the maximum size is the same size as an allocation
-	 * unit, use a percentage of the maximum page size).
+	 * Respect the configured split percentage if the calculated split
+	 * size is either zero or a full page.
 	 */
-	if (split_size == btree->allocsize)
+	if (split_size == 0 || split_size == maxpagesize)
 		split_size = (uint32_t)((a * (u_int)btree->split_pct) / 100);
```
keithbostic: I thought we intended to allow applications to set the split percentage to 100%?

Also, I'm a little worried about the original code here: it seems to me if you set both the allocation size and the maximum page size to the minimum possible allocation size, you can end up with a calculated split_size that's less than an allocation size? 

What do you think of something like this, that allows 100%, and bounds the calculated split size more simply?

```
uint32_t
__wt_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
{
        uintmax_t a;
        uint32_t split_size;

        /*
         * The application may not be modifying records or inserting anywhere
         * but the end, allow maximum page-size splits.
         */
        if (btree->split_pct == 100)
                return (maxpagesize);

        /*
         * Ideally, the split page size is some percentage of the maximum page
         * size rounded to an allocation unit (round to an allocation unit so
         * we don't waste space when we write).
         */
        a = maxpagesize;                        /* Don't overflow. */
        split_size = (uint32_t)WT_ALIGN_NEAREST(
            (a * (u_int)btree->split_pct) / 100, btree->allocsize);

        /*
         * Bound the calculated split size at an allocation size minimum and  
         * a maximum page size minus an allocation size, maximum.
         */
        if (split_size == maxpagesize)
                split_size -= btree->allocsize;
        if (split_size < btree->allocsize)
                split_size = btree->allocsize;

        return (split_size);
}
```

src/reconcile/rec_write.c
```diff
@@ -3286,6 +3285,17 @@ __rec_split_write(WT_SESSION_IMPL *session,
...
+	/* Output a verbose message if we create a page without many entries */
+	if (WT_VERBOSE_ISSET(session, WT_VERB_SPLIT) && r->entries < 6)
+		WT_ERR(__wt_verbose(session, WT_VERB_SPLIT,
+		    "Reconciliation creating a page with %" PRIu32
+		    " entries, memory footprint %" PRIu64
+		    ", page count %" PRIu32 ", %s, split state: %d\n",
+		    r->entries, r->page->memory_footprint, r->bnd_next,
+		    F_ISSET(r, WT_EVICTING) ? "evict" : "checkpoint",
+		    r->bnd_state));
+
```
keithbostic: I'm not sure about this verbose message: if it's been useful to you, I'm good with it, but if it's not been sufficiently helpful, I'd remove it (or ifdef it out). The reason is if you have a small page size (like a lot of format runs do), most created pages will have less than 6 entries, which means all we'll see is these messages.

Alternatively, maybe we could do something like:
```
(r->split_size / r->entries > 512)
```
but I don't know if that's sufficient for what you were trying to debug.

I'm also curious about the memory footprint and eviction/checkpoint distinction. When you were debugging this one, how did those end up being useful? (I have no objection to them being there, just curious why they were useful to you.)

src/reconcile/rec_write.c
michaelcahill: The original code exists for a customer with 4KB allocation size and 4KB maximum page size.  They need the split percentage to apply and not be rounded to an allocation size.

src/include/misc.h
agorrod: Nice catch. Fixed.

agorrod:
@keithbostic I believe @michaelcahill answered your questions. Do you have outstanding concerns?

keithbostic:
@agorrod, yes, I see @michaelcahill's reply, I agree, I'm good to merge. (Maybe it's worth enhancing the comments to flag those two edge cases?)

src/reconcile/rec_write.c
agorrod: @keithbostic I realized I didn't reply to your question here. You are right that the entry count isn't an ideal metric, but it was a good filter for the workload I was using, and I'm happy for that to evolve over time.
Do you run test/format with verbose split turned on very often? I've thought a couple of times that it would be nice if verbose messages had a unique identifier, to make it simple to filter out just the messages you are interested in keeping.

keithbostic: @agorrod, I don't think I've run with `WT_VERB_SPLIT` in years -- I used it long ago when debugging the original split code, and it's been pretty useless since then.

Instead of creating another level of naming by adding unique identifiers to verbose messages, why not create a new category of verbose messages? It doesn't cost us any performance to have more verbose message types, and the current implementation supports around 40 more verbose message types without any change.

There's also `WT_VERB_TEMPORARY`: you could leave the verbose message in the source using verb-temporary, but ifdef'd out, and it's trivial to turn it on whenever you want it?

PR #2456, opened 2016-01-22
WT-2346 Don't hold the schema lock during checkpoint I/O.

michaelcahill:
This required removing the special case code for named checkpoints of
trees that are in the process of being bulk loaded.


michaelcahill:
autobuild: please retest this

michaelcahill:
autobuild: retest this please

dist/flags.py
```diff
@@ -113,6 +113,7 @@
         'SESSION_INTERNAL',
         'SESSION_LOCKED_CHECKPOINT',
         'SESSION_LOCKED_HANDLE_LIST',
+        'SESSION_LOCKED_METADATA',
```
michaelcahill: Note that the metadata lock is closely related to the turtle lock -- I think they can be combined.

agorrod:
@michaelcahill Do you mind if I move stress testing over to this branch for a while?

michaelcahill:
Thanks, @agorrod, sounds good to me.

I would like to see this code survive a lot of testing, both ours and crash/recovery testing, before I'd want to inflict it on anyone.  OTOH, I've never seen slow creates or drops in the new wtperf workload on this branch and it fails reliably for me on develop...

@keithbostic, can you please take a look and we can talk through the implications for named checkpoints?  After looking at this, I think there are existing rough edges around checkpoints and exclusive operations (e.g., bulk loads can fail to start if they race with a checkpoint), but I don't think this change makes them worse.

test/suite/test_bulk02.py
```diff
@@ -72,11 +71,8 @@ def test_bulkload_checkpoint(self):
         # In the case of named checkpoints, verify they're still there,
         # reflecting an empty file.
         if self.ckpt_type == 'named':
-            cursor = self.session.open_cursor(
-                self.uri, None, 'checkpoint=myckpt')
-            self.assertEquals(cursor.next(), wiredtiger.WT_NOTFOUND)
-            cursor.close()
-
+            self.assertRaises(wiredtiger.WiredTigerError,
+                lambda: self.session.open_cursor(self.uri, None, 'checkpoint=myckpt'))
```
michaelcahill: To be clear, as far as I know, this is the only functional change -- named checkpoints will skip files that are in the process of being bulk-loaded when the checkpoint starts.

OTOH, it's possible to create new tables while a checkpoint is in progress, which is a big win for some workloads.

agorrod:
autobuild: retest this please

src/txn/txn_ckpt.c
```diff
@@ -249,23 +243,18 @@ __wt_checkpoint_list(WT_SESSION_IMPL *session, const char *cfg[])
...
-		    &session->ckpt_handle[session->ckpt_handle_next++].name);
+	if ((ret = __wt_session_get_btree(session, name, NULL, NULL, 0)) != 0)
+		return (ret == EBUSY ? 0 : ret);
 
-	return (ret);
+	/* Make sure there is space for the next entry. */
+	WT_RET(__wt_realloc_def(session, &session->ckpt_handle_allocated,
+	    session->ckpt_handle_next + 1, &session->ckpt_handle));
+	session->ckpt_handle[session->ckpt_handle_next++] = session->dhandle;
+	return (0);
```
keithbostic: @michaelcahill 
To be clear, this isn't a change, but shouldn't we do the allocation before calling `__wt_session_get_btree`, that is, is it safe to leave the result of `__wt_session_get_btree` in the session handle?

keithbostic:
@michaelcahill, this looks good to me, but I would like to talk this over when you have a few minutes.

One thought: now we're not holding the schema lock, maybe it's reasonable to flush/sync files as we checkpoint them, and then we could release the handles immediately, instead of holding them until the end? (We could even order that process, if we did the checkpoint/release of files that could potentially be bulk-loaded first, those operations would be less likely to block -- not really arguing for that, but throwing it out there.)

A slightly different step on that path could be that `__wt_checkpoint` could schedule an asynchronous flush of the file?

src/txn/txn_ckpt.c
michaelcahill: Agreed, I'll reverse the order.

src/txn/txn_ckpt.c
```diff
@@ -543,16 +520,24 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	 * Recovery relies on the checkpoint LSN in the metadata only being
 	 * updated by full checkpoints so only checkpoint the metadata for
 	 * full or non-logged checkpoints.
+	 *
+	 * XXX lots of overlap with __wt_meta_track_off.
```
michaelcahill: This comment probably shouldn't be merged despite being true.

src/txn/txn_ckpt.c
```diff
@@ -411,20 +400,7 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
 
-	/*
-	 * The underlying flush routine scheduled an asynchronous flush
-	 * after writing the leaf pages, but in order to minimize I/O
-	 * while holding the schema lock, do a flush and wait for the
-	 * completion. Do it after flushing the pages to give the
-	 * asynchronous flush as much time as possible before we wait.
-	 */
-	if (F_ISSET(conn, WT_CONN_CKPT_SYNC))
-		WT_ERR(__checkpoint_apply(session, cfg, __wt_checkpoint_sync));
```
agorrod: Is this change orthogonal to the rest of this change? I don't think it's altering correctness of checkpoints, but I think we added the sync for a particular workload (maybe iibench?)

agorrod:
@michaelcahill I think a mention in upgrading documentation would be worthwhile. I had one other question, and @keithbostic has an outstanding question as well.

Once those minor concerns are addressed I'm happy for this to be merged.

michaelcahill:
This branch has been reviewed and has run the crash tests for enough loops that we have some level of confidence that it doesn't break anything.  Merging...

PR #2458, opened 2016-01-22
WT-2229 Remove compression backoff stats

sueloverso:
@daveh86 Here's a branch of the compression-backoff change that removes those stats per the suggestion from @keithbostic .  I created a pull request but I don't expect to merge this unless it is shown to be the cause of the performance regression.


michaelcahill:
Given the results we've seen in testing and my belief that these workloads aren't common, we've agreed to close this.

PR #2460, opened 2016-01-25
WT-2308 Fix for joins on indices that use extractors and 'u' at end of format

ddanderson:
Passes the submitted test case.

### DIFF
--- a/src/packing/pack_impl.c
+++ b/src/packing/pack_impl.c
@@ -164,7 +164,7 @@ __wt_struct_repack(WT_SESSION_IMPL *session, const char *infmt,
 	 * The former case has the size embedded before the item, the latter
 	 * does not.
 	 */
-	if ((len = strlen(outfmt)) > 1 && outfmt[len - 1] == 'u' &&
+	if ((len = strlen(outfmt)) > 0 && outfmt[len - 1] == 'u' &&
 	    strlen(infmt) > len && infmt[len - 1] == 'U') {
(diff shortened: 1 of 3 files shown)


ddanderson:
@michaelcahill, can you review this please?

ddanderson:
@michaelcahill, please hold off on any review until I can get to the bottom of the latest test case from Denis in WT-2308.

src/packing/pack_impl.c
```diff
@@ -164,7 +164,7 @@ __wt_struct_repack(WT_SESSION_IMPL *session, const char *infmt,
 	 * The former case has the size embedded before the item, the latter
 	 * does not.
 	 */
-	if ((len = strlen(outfmt)) > 1 && outfmt[len - 1] == 'u' &&
+	if ((len = strlen(outfmt)) > 0 && outfmt[len - 1] == 'u' &&
```
michaelcahill: @ddanderson, I know you're still bug hunting, but this part of the change confuses me.  Won't it turn ordinary "u" into "U"?  In other words, the common case of a `WT_ITEM` will need to be encoded with a size prefix.

michaelcahill:
@ddanderson, just doing a review, and the first changes for WT-2308 have already been merged into MongoDB master.  Given that, we'll need a new ticket for the remaining fix to avoid confusion the release management process.

ddanderson:
See #2487, that has a comprehensive fix for all tests.

PR #2461, opened 2016-01-25
WT-2215 Use 32-bit LSN file and offsets. Set LSNs as a 64-bit value.

sueloverso:
@michaelcahill Here is a new branch that implements the LSN changes as a union, getting rid of the macros from the earlier version.  This is off develop and replaces the other branch.  Please review.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -180,10 +180,10 @@ __log_archive_once(WT_SESSION_IMPL *session, uint32_t backup_file)
 	 * disk and the checkpoint LSN.
 	 */
 	if (backup_file != 0)
-		min_lognum = WT_MIN(log->ckpt_lsn.lsn_file, backup_file);
+		min_lognum = WT_MIN(log->ckpt_lsn.l.file, backup_file);
 	else
 		min_lognum = WT_MIN(
-		    log->ckpt_lsn.lsn_file, log->sync_lsn.lsn_file);
+		    log->ckpt_lsn.l.file, log->sync_lsn.l.file);
 	WT_RET(__wt_verbose(session, WT_VERB_LOG,
 	    "log_archive: archive to log number %" PRIu32, min_lognum));
 
@@ -317,7 +317,7 @@ __wt_log_truncate_files(
 	backup_file = 0;
 	if (cursor != NULL)
 		backup_file = WT_CURSOR_BACKUP_ID(cursor);
-	WT_ASSERT(session, backup_file <= log->alloc_lsn.lsn_file);
+	WT_ASSERT(session, backup_file <= log->alloc_lsn.l.file);
 	WT_RET(__wt_verbose(session, WT_VERB_LOG,
 	    "log_truncate_files: Archive once up to %" PRIu32,
 	    backup_file));
@@ -367,7 +367,7 @@ __log_file_server(void *arg)
 			 * could see mismatched settings.  If we do, yield
 			 * until it is set.  This should rarely happen.
 			 */
-			while (log->log_close_lsn.lsn_file < filenum)
+			while (log->log_close_lsn.l.file < filenum)
 				__wt_yield();
 
 			if (__wt_log_cmp(
@@ -399,9 +399,9 @@ __log_file_server(void *arg)
 				 * zeroed space.
 				 */
 				WT_ERR(__wt_ftruncate(session,
-				    close_fh, close_end_lsn.lsn_offset));
+				    close_fh, close_end_lsn.l.offset));
 				WT_SET_LSN(&close_end_lsn,
-				    close_end_lsn.lsn_file + 1, 0);
+				    close_end_lsn.l.file + 1, 0);
 				__wt_spin_lock(session, &log->log_sync_lock);
 				locked = true;
 				WT_ERR(__wt_close(session, &close_fh));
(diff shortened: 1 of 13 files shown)


src/include/log.h
```diff
@@ -6,6 +6,27 @@
...
+#else
+			uint32_t offset;
+			uint32_t file;
+#endif
+		} l;
+		uint64_t file_offset;
+	} u;
+};
+#define	lsn_file	u.l.file
+#define	lsn_offset	u.l.offset
```
michaelcahill: @sueloverso, not a big deal, but if you wanted to get rid of these macros, you could have:

```c
union __wt_lsn {
	struct {
#ifdef	WORDS_BIGENDIAN
		uint32_t file;
		uint32_t offset;
#else
		uint32_t offset;
		uint32_t file;
#endif
	} l;
	uint64_t file_offset;
};
```

Then `s/lsn_file/l.file/g` and `s/lsn_offset/l.offset/g`.  I don't feel strongly: I'm not a fan of macros to access struct/union fields, but we already have a bunch.

src/include/log.h
```diff
@@ -13,32 +34,43 @@
 /* Logging subsystem declarations. */
 #define	WT_LOG_ALIGN			128
 
+/*
+ * Atomically set the two components of the LSN.
+ */
+#define	lsn_file		u.l.file
+#define	lsn_offset		u.l.offset
```
michaelcahill: Repeated macros?

src/include/log.h
```diff
@@ -13,32 +34,43 @@
...
+#define	WT_SET_LSN(l, f, o) (l)->u.file_offset = (((uint64_t)(f) << 32) + (o))
```
michaelcahill: Could this be?

```c
do {
    (l)->lsn_file = (f);
    (l)->lsn_offset = (o);
} while (0)
```

src/include/log.h
```diff
@@ -13,32 +34,43 @@
...
-	(l)->file = UINT32_MAX;						\
-	(l)->offset = INT64_MAX;					\
+	WT_SET_LSN((l), UINT32_MAX, INT32_MAX);			\
 } while (0)
 
 #define	WT_ZERO_LSN(l)	do {						\
-	(l)->file = 0;							\
-	(l)->offset = 0;						\
+	WT_SET_LSN((l), 0, 0);						\
 } while (0)
```
michaelcahill: I don't think any of these need `do ... while(0)` loops any more.

src/include/log.h
```diff
@@ -13,32 +34,43 @@
 /* Logging subsystem declarations. */
 #define	WT_LOG_ALIGN			128
 
+/*
+ * Atomically set the two components of the LSN.
+ */
+#define	lsn_file		u.l.file
+#define	lsn_offset		u.l.offset
+#define	WT_SET_LSN(l, f, o) (l)->u.file_offset = (((uint64_t)(f) << 32) + (o))
+#define	WT_SET_LSN_FILE(l, f)	WT_SET_LSN((l), (f), (l)->lsn_offset)
```
michaelcahill: or just `(l)->lsn_file = (f)`?

src/include/log.h
```diff
@@ -13,32 +34,43 @@
...
+#define	WT_SET_LSN_OFFSET(l, o)	WT_SET_LSN((l), (l)->lsn_file, (o))
```
michaelcahill: or `(l)->lsn_offset = (o)`?

michaelcahill:
@sueloverso, I've made a few comments on the remaining macros, but this lgtm.

src/include/log.h
sueloverso: Yeah, I like that.  I'll push a new version with this.

src/include/log.h
sueloverso: The whole point is to set the LSN as a single 64-bit value rather than two assignments so that reading an LSN never sees something half-assigned.

michaelcahill: Ack, just trying to avoid mixing bit shifts with union field accesses.  We could do it here (by assigning to the fields of a local variable then using its `file_offset`) but it's not worth it.

sueloverso:
@michaelcahill I took most of your suggestions.  New code is pushed.

michaelcahill:
Thanks, @sueloverso -- lgtm, I'll merge.

src/conn/conn_log.c
```diff
@@ -440,9 +440,9 @@ __log_file_server(void *arg)
 				 * this worker thread process that older file
 				 * immediately.
 				 */
-				if ((log->sync_lsn.file <
-				    log->bg_sync_lsn.file) ||
-				    (log->sync_lsn.file < min_lsn.file))
+				if ((log->sync_lsn.l.file <
+				    log->bg_sync_lsn.l.file) ||
+				    (log->sync_lsn.l.file < min_lsn.l.file))
 					continue;
```
baiwfg2, line 446: Hi. I'm researching on wiredtiger source recently. May I ask a question ? After the `continue`, If I assert `log->sync_lsn.l.file == log->bg_sync_lsn.l.file` and `log->sync_lsn.l.file == min_lsn.l.file`, is that right ? Because sync lsn should not exceed write lsn. (`sync` occurs after `write`).

lukech: Hi @baiwfg2, this pull request had been merged for a really long time. If you want better attention of your question, I'd suggest raising a topic in the [wiredtiger-users](https://groups.google.com/g/wiredtiger-users) google group.

PR #2462, opened 2016-01-25
WT-2348: xargs -P isn't portable

keithbostic:
WT-2348: xargs -P isn't portable


agorrod:

PR #2464, opened 2016-01-28
WT-2356 Fix recovery bug in log_scan advancing.

sueloverso:
If we find a partial record, that is the end of log.
Add test program.

@michaelcahill Please review this change.  The fix itself is small, in log.c.  Most of this is the test program and changing the directory structure under test/recovery.

The test program passes with the fix and fails without it.


michaelcahill:
@sueloverso, the change to `log.c` looks good, but we had discussed another issue where we skip over partial log records at the end of a file if they are smaller than the allocation size.  Is that case handled by this change?

Would you mind if I put the two tests back into one directory?

sueloverso:
@michaelcahill I pushed a new version with the small partial at the end.  It was more involved than the other fix.  Please give it careful review.

michaelcahill:

PR #2465, opened 2016-01-28
WT-2363 WT-2352 Remove support for bzip2 and allow build and test without requiring lz4 

keithbostic:
WT-2363 Remove support for bzip2 compression
WT-2352 Allow build and test without requiring lz4.

@agorrod, here's the changes that drop bzip2, and change test/format to only select from built-in compression engines.


test/format/config.c
```diff
@@ -256,24 +256,25 @@ config_compression(const char *conf_name)
 		switch (mmrand(NULL, 1, 20)) {
 		case 1: case 2: case 3: case 4:		/* 20% no compression */
```
agorrod: Would you mind shuffling this around so the no-compression case becomes a default and update the "20% no compression" comment to reflect that the proportion varies depending on what compression options are available?

keithbostic: @agorrod 
> Would you mind shuffling this around so the no-compression case becomes a default and update the "20% no compression" comment to reflect that the proportion varies depending on what compression options are available?
 
Sure, done.

agorrod:
Thanks @keithbostic lgtm - I'll get this merged.

PR #2466, opened 2016-01-29
WT-2366 Extend wtperf to support updates that grow records.

agorrod:
Add a new wtperf stress configuration file that appends new records and grows existing records.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -254,6 +254,12 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
 					goto err;
 				continue;
 			}
+			if (STRING_MATCH("update_grow", k.str, k.len) ||
+			    STRING_MATCH("updates_grow", k.str, k.len)) {
+				if ((workp->update_grow = v.val) < 0)
+					goto err;
+				continue;
+			}
 			if (STRING_MATCH("update", k.str, k.len) ||
 			    STRING_MATCH("updates", k.str, k.len)) {
 				if ((workp->update = v.val) < 0)
@@ -293,7 +299,8 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
 		if (ret != 0)
 			goto err;
 		if (workp->insert == 0 && workp->read == 0 &&
-		    workp->update == 0 && workp->truncate == 0)
+		    workp->update == 0 && workp->update_grow == 0 &&
+		    workp->truncate == 0)
 			goto err;
 		/* Why run with truncate if we don't want any truncation. */
 		if (workp->truncate != 0 &&
@@ -306,7 +313,8 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
 		if (workp->truncate != 0 && workp->threads > 1)
 			goto err;
 		if (workp->truncate != 0 &&
-		    (workp->insert > 0 || workp->read > 0 || workp->update > 0))
+		    (workp->insert > 0 || workp->read > 0 ||
+		    workp->update > 0 || workp->update_grow > 0))
 			goto err;
 		cfg->workers_cnt += (u_int)workp->threads;
 	}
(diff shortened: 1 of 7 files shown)


agorrod:
FWIW: Even the short running configuration file I added in this pull request demonstrates poor page layouts at the end of the run, and the pattern is simple to decode, it is:

{{
Page type, Number of entries, Page size
L 1552 32749
L 2 82
L 2 82
L 2 82
L 1552 32742
L 2 82
L 2 82
L 2 82
L 1552 32739
L 2 82
L 2 82
L 2 82
}}

i.e: We are reconciling pages just larger than a page_size regularly, and since the split percentage is being rounded to 100%, carving off a tiny amount and putting it onto a new page.

Running with the changes in https://github.com/wiredtiger/wiredtiger/pull/2455 to not align the split percentage (so this configuration doesn't setup a 100% split percentage) helps in this short run. It will be interesting to see how much it helps over a longer time period.

sueloverso:
@agorrod while the code here is fine, it isn't quite what I'd like to see merged into the tree.  First, it would be good if it could support both growing and contracting.  I suspect there may be interesting things to see in the tree if updates make things smaller each time too.  Second, it would be good if we could avoid adding another operation type, but somehow specified it as part of the existing update workload.  Third, maybe the grow_delta should be part of the workload spec instead of a global.  You could have a workload expanding by different rates, and another contracting that would need to be specified separately.  I'll try to think about a concrete configuration to suggest.

agorrod:
@sueloverso That is a reasonable request. I partially pushed ahead with this approach because I think it'd be nice if it were easier to add different operation types to wtperf. Demonstrating that it's currently hard doesn't help anyone.

There is also already an ``insert_rmw`` option that is global in wtperf. It would make sense to add that as a sub-option to workload groups at the same time as adding growing update support.

> it would be good if it could support both growing and contracting

I'm struggling to see how to confluence those into a single configuration setting though. Maybe if they are options to a workload the result could be achieved by doing something like:

```
threads=((count=1,inserts=1,throttle=8300),(count=2,updates=1,update_grow_delta=20),(count=2,updates=1,update_shrink_delta=20))
```

If we extended wtperf option parsing to support signed numbers we could potentially allow negative values.

agorrod:
The tl;dr for my previous comment is: Don't merge this branch yet.

sueloverso:
> I'm struggling to see how to confluence those into a single configuration setting though.

My thoughts were headed in the direction of the negative delta.  That makes sense from the user perspective.
```
threads=((count=2,updates=1,delta=3),(count=2,updates=1,delta=-1))
```

agorrod:
> My thoughts were headed in the direction of the negative delta. That makes sense from the user perspective.
> 
> threads=((count=2,updates=1,delta=3),(count=2,updates=1,delta=-1))

OK, that seems reasonable. I've got a couple of resolvable concerns:
* It's the first use of signed values in wtperf configurations
* The structure inside threads isn't as well defined as other options in wtperf, and adding more content there increases the complexity
* I think we need to be careful with naming. What would you expect to happen with:

```threads=((count=2,updates=1,inserts=1,delta=3)```

While I'm thinking about the functionality it's obvious that delta only applies to updates, I'd like to make that clear when choosing a name. Especially with the thought of supporting different operation types in the future.

sueloverso:
( @agorrod I think your last comment may have been mangled? It seemed like you were going to put in an example of some kind)
I agree adding more increases the complexity, but it does make sense as it is describing parts of the workload.  For `delta`, I think if a user defined it on anything other than an `update` workload (or a multiple operation workload that has updates as part of it) it should either be an error, or a no-op (with a warning).

The other thing is that you need `value_sz`, `value_max` and `value_min`.  This also opens up the question about what a contracting workload should do when it reaches the min.  Should it reset the value to the default or maximum and then contract again?  Should it just stay at the minimum?

agorrod:
Closed in favor of #2538

PR #2467, opened 2016-01-29
WT-2368: row-store can pass garbage keys to collator functions

keithbostic:
WT-2368: row-store can pass garbage keys to collator functions

@ddanderson, this is the fix for WT-2368.


michaelcahill:
FTR, this reverts WT-2216.  LGTM.  I'll merge.

PR #2470, opened 2016-01-29
WT-2123: Don't clear allocated memory if not required

keithbostic:
WT-2123: Don't clear allocated memory if not required

I don't think we're going to get anywhere with this ticket as a complete set of changes, it's too hard to figure out which allocations need memory cleared, and which don't. To allow us to make progress, I added the `__wt_malloc` API (and made one obvious change to use it, where we duplicate strings). I figure we can merge this change, close the JIRA ticket, and then iteratively change code to use `__wt_malloc` where it makes sense.

Second, the one place I think we should change now is buffers we allocate for reading/writing the file. Those are big allocations and at the moment they're cleared, and they don't need to be. The problem is figuring out which allocations don't need to be cleared -- as far as I can tell, we never need to clear memory allocated for a WT_ITEM via the buffer functions, and so I drove that distinction down into the `os_alloc.c` code. It's worse layering than we had before, but I think it's worth doing to avoid the buffer clearing.


src/support/scratch.c
```diff
@@ -40,14 +40,8 @@ __wt_buf_grow_worker(WT_SESSION_IMPL *session, WT_ITEM *buf, size_t size)
...
-	if (size > buf->memsize) {
-		if (F_ISSET(buf, WT_ITEM_ALIGNED))
-			WT_RET(__wt_realloc_aligned(
-			    session, &buf->memsize, size, &buf->mem));
-		else
-			WT_RET(__wt_realloc(
-			    session, &buf->memsize, size, &buf->mem));
-	}
+	if (size > buf->memsize)
+		WT_RET(__wt_realloc_item(session, buf, size));
```
michaelcahill: What's the motivation for this change?

We set the `WT_ITEM_ALIGNED` flag to indicate whether alignment is required: this is now ignoring that flag.  Can we do a better job of only aligning allocations that could be used for read/write operations?

keithbostic: > What's the motivation for this change?

To add a `__wt_malloc` function so we can iterate forward on not clearing memory that doesn't need to be cleared.

I also changed the `__wt_strndup` function to use `__wt_malloc` (because it was easy), and changed read/write operations to not clear their buffers (because I think that's likely to have the largest impact on performance).

The latter change wasn't as simple as I'd like -- the simplest solution I found was based on the facts that I/O is always done into `WT_ITEM` structures, and I didn't see any place we ever need to clear `WT_ITEM` structure memory. So, I pushed a `WT_ITEM` reallocation API down into the system-dependent allocation code and that path optionally aligns buffers and never clears memory. That means we won't clear scratch buffers or I/O buffers, which seems about right.

> We set the WT_ITEM_ALIGNED flag to indicate whether alignment is required: this is now ignoring that flag.

This change moves the `WT_ITEM_ALIGNED` flag test  to the `__wt_realloc_item` function.

> Can we do a better job of only aligning allocations that could be used for read/write operations?

I didn't investigate that, I left our current approach unchanged.

michaelcahill:
@keithbostic, sorry I misunderstood this change on first reading.

My query was really about having `WT_ITEM` logic in the `os_posix` layer: could the flag check go back into `support/scratch.c` (which is no longer a great name, but that's a different story).  That would mean exposing the two flavors of `realloc` from `os_alloc.c`, but I think the layering will be clearer.

keithbostic:
The flag check could go back into the higher-level code, but I think the result is two new booleans in the allocation API, one that clears the allocated memory and one that aligns the allocated memory.

And name the new function something else so we don't have to fix calls to `__wt_realloc` everywhere:

```
__wt_realloc_detail(WT_SESSION_IMPL *session,
    size_t *bytes_allocated_ret, size_t bytes_to_allocate,
    bool align_memory, bool clear_memory,
    void *retp)
```

That gives us finer-grained control in the future, which might be a good thing.

Am I understanding you correctly, and is that your preference?

michaelcahill:
@keithbostic, sorry for the delay.

My feeling here is that we want to head in the direction of making our calls match the usual C semantics.  So eventually, we'd have `__wt_realloc` *not* zero by default, and have some other call, like `__wt_recalloc` for cases that require memory to be cleared.

AFAIK, `__wt_realloc_aligned` never needs to clear memory -- in other words, we don't need the case of `align_memory = true && clear_memory = true` -- is that correct?

If so, I'd be inclined to export:

`__wt_realloc` -- current behavior
`__wt_realloc_aligned` -- current behavior
`__wt_realloc_noclear` -- ordinary `realloc` semantics.

At some future point, we can have a flag day where we swap over so the `noclear` call becomes the default.  Does that make sense?

keithbostic:
@michaelcahill, I've made the changes and I think this is ready for merge.

agorrod:

PR #2471, opened 2016-01-31
WT-2371: parent split cannot access the page after page-index swap

keithbostic:

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -875,16 +875,24 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref, WT_REF **ref_new,
 	/* The split is complete and correct, ignore benign errors. */
 	complete = WT_ERR_IGNORE;
 
-	WT_ERR(__wt_verbose(session, WT_VERB_SPLIT,
-	    "%p: %s %s" "split into parent %p, %" PRIu32 " -> %" PRIu32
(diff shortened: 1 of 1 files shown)


src/btree/bt_split.c
```diff
@@ -875,16 +875,24 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref, WT_REF **ref_new,
...
+		WT_ERR(__wt_verbose(session, WT_VERB_SPLIT,
+		    "%p: reverse split into parent %p, %" PRIu32 " -> %" PRIu32
+		    " (-%" PRIu32 ")",
+		    ref->page, parent, parent_entries, result_entries,
+		    parent_entries - result_entries));
+	else
+		WT_ERR(__wt_verbose(session, WT_VERB_SPLIT,
+		    "%p: split into parent %p, %" PRIu32 " -> %" PRIu32
+		    " (+%" PRIu32 ")",
+		    ref->page, parent, parent_entries, result_entries,
```
agorrod: Won't ``ref->page`` always be NULL?

michaelcahill:

src/btree/bt_split.c
keithbostic: Yes, it will -- do you think it's worth fixing?

agorrod: Nope.

PR #2472, opened 2016-02-01
WT-2372 - Fix windows inline define, changes for wt_internal inclusion

daveh86:


daveh86:

src/include/msvc.h
```diff
@@ -11,7 +11,9 @@
 #error "Only x64 is supported with MSVC"
 #endif
 
+#if !defined(__cplusplus) || !defined(inline)
```
agorrod: Do you need both of these? If so I'd recommend adding a comment to explain why.

ext/compressors/snappy/snappy_compress.c
```diff
@@ -26,13 +26,13 @@
  * OTHER DEALINGS IN THE SOFTWARE.
  */
 
+#include <wt_internal.h>
```
agorrod: Why is this moved above the other includes? Generally it's better to include system headers first, so I don't really like this change.

agorrod:
Thanks @daveh86 lgtm, I'll merge.

PR #2473, opened 2016-02-02
WT-2375 Add collator tests to test suite using a demonstration collator

ddanderson:
@michaelcahill, this is ready for review.


ddanderson:
@michaelcahill, ignore my previous comment about ENOMEM, I reworked this and encountered a couple different issues.  This pull request is one (not completely satisfactory) solution to those issues.

Some context on the change to packing.i: on one callback to the collator we are given a one byte buffer with "0x80", and the following byte is null.  Now the collator uses these and a "ii" format to unpack (this was the best I could do - it doesn't know yet whether it was given a primary key, so does it have "ii" or "i"? We decode "ii", and if wrong, hope for an error return). Without the proposed change to packing.i, on the second time through the loop in __wt_struct_unpackv, __unpack_read() gets called with a length of 0.  A zero length is allowed in __unpack_read and its brethren (I assume 0 means infinite size), the null byte is decoded and since it's supposed to be "i", it returns EINVAL.  Okay, we overflowed the input buffer, but  we gave the caller an error.  However the collator is also called with a one byte buffer with "0x80", and the following byte is "0x80".  Using the same path, the buffer is overflowed, and the second "i" finds a valid int.  But the assertion in __wt_struct_unpackv fires and we're dead.  Which is all the say, bad and variable things happen if we don't do bounds checking along the way -- it appears that WT_SIZE_CHECK_UNPACK calls in __unpack_read is supposed to do that check, but having maxlen 0 as a special (infinite?) case let this slip through the cracks.

Secondly, in the revint_compare function, I don't love the multiple calls to unpack, but I don't have a better solution -- unless the collator got a pair of flags telling it what to expect.  Not sure the caller to the collator knows, either.  Or maybe the collator receives an opaque pointer to the index table that we can hand to the unpacker so that it does the right thing.

Finally, this simple example works with index key format "i" and primary key format "i".  But what about when both of them are "u"?  (And the difference in how "u" and "uu" are packed will really require the collator to know *exactly* what it is handed).  I think we add in a padding byte in indices to avoid this sort of oddity - do we need to expose that to the collator?  The opaque pointer to the index table could handle that as well - it could be hidden in the WT_COLLATOR struct.

ext/collators/revint/revint_collator.c
```diff
@@ -0,0 +1,148 @@
...
+	 * primary key. The reason is that if the application calls
+	 * WT_CURSOR::search on a index key for which there are more than
+	 * one value, the search key will not yet have a primary key.  We
+	 * want to position the cursor at the 'first' matching index key so
+	 * that repeated calls to WT_CURSOR::next will see them all.
+	 *
+	 * To keep this code simple, we do not reverse the ordering
+	 * when comparing primary keys.
+	 */
+	if ((ret = wtapi->struct_unpack(wtapi, session,
```
agorrod: @ddanderson Did you consider using the packing API here instead? i.e using the `WT_PACK_STREAM` structure via the `wiredtiger_unpack_start` etc APIs.

ddanderson: @agorrod, yes I considered the packing APIs, it's possible it would work, but the ext libraries are not linked against the wiredtiger library.  At least it fails to link libwiredtiger_revint_collator.so on Linux because of missing external symbols.

michaelcahill: Understood.  We've been talking this over: I think the best solution we have at the moment is to add the `WT_PACK_STREAM` functions to the extension API.

Obviously `struct_unpack`

src/include/packing.i
```diff
@@ -677,6 +677,8 @@ __wt_struct_unpackv(WT_SESSION_IMPL *session,
 
 	if (fmt[0] != '\0' && fmt[1] == '\0') {
 		pv.type = fmt[0];
+		if (p >= end && pv.type != 'u')
+			return (ENOMEM);
 		if ((ret = __unpack_read(session, &pv, &p, size)) == 0)
 			WT_UNPACK_PUT(session, pv, ap);
 		return (0);
```
michaelcahill: would `return (ret)` here have the same effect?

src/include/packing.i
```diff
@@ -684,6 +686,8 @@ __wt_struct_unpackv(WT_SESSION_IMPL *session,
 
 	WT_RET(__pack_init(session, &pack, fmt));
 	while ((ret = __pack_next(&pack, &pv)) == 0) {
+		if (p >= end && pv.type != 'u')
+			return (ENOMEM);
 		WT_RET(__unpack_read(session, &pv, &p, (size_t)(end - p)));
```
michaelcahill: Similarly, doesn't `__unpack_read` detect that there aren't enough bytes?

ddanderson: @michaelcahill, in this case, __unpack_read does not detect that there aren't enough bytes when `end == p`. In that case, a zero size is passed to `__unpack_read`, and `WT_SIZE_CHECK_UNPACK` is called with second arg being 0.  That is allowed, presumably a zero size means no checks.

I'm up against my vacation (leaving tomorrow AM), probably won't get to work on this until I return. Nobody is waiting on this pull request AFAIK.

michaelcahill:
@agorrod, I have merged `develop` into this branch including your changes in WT-2404.  Do you have followup changes to @ddanderson's test code that could now be applied here?

agorrod:
@michaelcahill Thanks for the ping - I have pushed the updated collator. Would you mind doing a final review and merge if you are happy.

michaelcahill:
Thanks @agorrod and @ddanderson, lgtm.  I'll merge.

PR #2474, opened 2016-02-02
WT-2374 If a backup file exists, remove/recreate metadata and turtle.

sueloverso:
@michaelcahill Please review this change for managing the backup, metadata and turtle files if a copy has been done.

The one thing that isn't fixed that should be is the comment/code disparity in `wt_checkpoint_close`.


sueloverso:
@michaelcahill The test that is failing, txn04, fails because it runs `wt backup -t uri`.  When run with a target, the backup directory ends up with both a backup file and a turtle file.  In the old code, it skips/removes the backup file in favor of the turtle file.  In the new code, we remove the turtle file and regenerate from the backup file.  

I don't see any way to distinguish these two cases in `wt_turtle_init`.

However, the test currently expects that a committed operation is not there.  That seems wrong to me now.  I added the exception check in the test years ago when it was first added.  I'm not sure why we would expect that?  I do not remember, do you?  The difference is the LSN in the turtle file.  In the old code the LSN is up to the committed operation and causes recovery to skip the committed operation.  In the new code, we remove the turtle file and recreate it and therefore it doesn't have an LSN and we do a full recovery.

sueloverso:
autobuild: retest this please

michaelcahill:
@sueloverso, I made the change to have tests expect the message.  This now LGTM.  If you are happy with my changes, please merge.

PR #2475, opened 2016-02-03
WT-2377 - change wtperfs dstrndup to not call strndup

daveh86:
as it is not supported on MSVC.

Looks like my regex that was to replace spaces with tabs as part of a copy/paste found a formatting nit, I suspect that it can be left as is.

### DIFF
--- a/bench/wtperf/wtperf.h
+++ b/bench/wtperf/wtperf.h
@@ -403,7 +403,8 @@ dstrdup(const char *str)
 
 /*
  * dstrndup --
- *      Call strndup, dying on failure.
+ *      Call emulating strndup, dying on failure. Don't use actual strndup here
+ *	as it is not supported within MSVC.
  */
 static inline char *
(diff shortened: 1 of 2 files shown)


bench/wtperf/wtperf.h
```diff
@@ -409,10 +409,11 @@ static inline char *
 dstrndup(const char *str, const size_t len)
 {
 	char *p;
+	p = dcalloc(len + 1, 1);
```
agorrod: Could you add a comment explaining why we don't call strndup - I can see my self changing this back in the future.

daveh86:
@agorrod, can you have a look at this branch which will allow us to compile wtperf on Windows?

I have fixes both for both issues with dstrndup and adds the idle_table_cycle.c file to the SConstruct file.

agorrod:
Thanks @daveh86 lgtm, I'll merge.

PR #2476, opened 2016-02-03
WT-2378 Always wait for the handle list lock.

michaelcahill:
That is, the "lock_wait=false" configuration to WT_SESSION::drop does
not apply to the handle list lock.  The reason is code (specifically in
LSM) that acquires the handle list lock to get an LSM tree, does some
work, then acquires the handle list lock again when discarding the tree.
If the first acquire happens to succeed and the second acquire fails,
the tree is left in the system with its reference count bumped.

### DIFF
--- a/src/include/schema.h
+++ b/src/include/schema.h
@@ -79,27 +79,37 @@ struct __wt_table {
 #define	WT_COLGROUPS(t)	WT_MAX((t)->ncolgroups, 1)
 
 /*
- * WT_WITH_LOCK --
- *	Acquire a lock, perform an operation, drop the lock.
+ * WT_WITH_LOCK_WAIT --
+ *	Wait for a lock, perform an operation, drop the lock.
  */
-#define	WT_WITH_LOCK(session, ret, lock, flag, op) do {			\
-	ret = 0;							\
+#define	WT_WITH_LOCK_WAIT(session, lock, flag, op) do {			\
 	if (F_ISSET(session, (flag))) {					\
 		op;							\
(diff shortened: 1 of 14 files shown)


michaelcahill:
@agorrod, I had a simpler version of this change that kept an unused `ret` parameter to `WT_WITH_HANDLE_LIST_LOCK`.  That kept the various macros the same shape, but that doesn't seem like a good idea long term.  Can you please take a look?

src/include/schema.h
```diff
@@ -79,27 +79,37 @@ struct __wt_table {
...
  * WT_WITH_LOCK --
  *	Acquire a lock, perform an operation, drop the lock.
  */
 #define	WT_WITH_LOCK(session, ret, lock, flag, op) do {			\
 	ret = 0;							\
-	if (F_ISSET(session, (flag))) {					\
-		op;							\
-	} else if (F_ISSET(session, WT_SESSION_LOCK_NO_WAIT)) {		\
+	if (!F_ISSET(session, (flag)) &&				\
+	    F_ISSET(session, WT_SESSION_LOCK_NO_WAIT)) {		\
```
michaelcahill: FTR, this is structured this way to allow for reuse of the `WT_WITH_LOCK_WAIT` macro, rather than repeating that logic.

agorrod:
Thanks @michaelcahill This change lgtm - I'll merge. As we discussed in person, I'm a bit nervous that the ``lock_wait`` functionality is more subtle than it appears on the surface, so we will want to be cautious before adding it to APIs other than drop.

PR #2478, opened 2016-02-03
WT-2361: column-store starting record number error

keithbostic:
WT-2361: column-store starting record number error


keithbostic:
NB: Not yet ready to be merged.

keithbostic:
@michaelcahill, @agorrod, I'm stuck on this one, I'd like to talk it over with you.

As far as I can tell, the failure doesn't have anything to do with the addition of column-store splits (other than it permits it to be triggered), I think it's a checkpoint problem, or we're writing corrupt trees for some unknown reason.

The failure mode is (almost) always something like this:
```
The parent page:
        ....
        recno 18136
        memory 0x804b51020 [5180416-5180928, 512, 1679098416]
        recno 18153
        memory 0x8034227c0 [5180928-5181440, 512, 3230426940]
        recno 18155
        disk [65536-66048, 512, 2645476425]

The recno 18153 page:
column-store variable-length leaf page, recno 18153, entries 3, encrypted, generation 22247
        value/short: len 30, rle: 1
        {0000018153/LMNOPQRSTUVWXYZABCD}
        value/short: len 21, rle: 1
        {0000018154/LMNOPQRSTU}
        value/short: len 21, rle: 1
        {0000018155/LMNOPQRSTU}

The recno 18155 page  
column-store variable-length leaf page, recno 18155, entries 1, encrypted, generation 22261
        value/short: len 21, rle: 1
        {0000018155/LMNOPQRSTU}
```

Notice record 18155 appears in both leaf pages. Verify is failing because it found recno 18155 in the parent page, and recno 18155 is not greater than the last record in the last leaf page it read.

* If I turn off checkpoints in format, the bug doesn't trigger.
* The leaf page that fails is a page that's been insert-split (in this case, there's been an insert-split where recno 18155 was pushed to a new page).
* The page that fails is always the last leaf page in the table.
* Other records have been entered into the tree after the split, but they don't show up in the tree, so, for example, record 20000 might have been inserted in this run, but it doesn't appear in the tree.

keithbostic:
A different failure mode:
```
        append 20578
        value {DUPLICATEV/LMNOPQRSTUVWXYZAB}
        append 20579
        value {0000020579/LMNOPQRST}
        append 20580
        value {0000020580/LMNOPQRSTUVWXYZABCDEFGHIJKLMNOPQRSTUVWXYZABCDEFGHIJKLM
        append 20582
        value {0000020582/LMNOPQRSTUVWXYZABCDE}
        append 20583
        value {0000020583/LMNOPQRSTUVWXYZ}
```
We managed to split a page at record number 20581, and then appended new records into the page; notice this snapshot, from the page's reconciliation, record number 20581 is missing.

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill, a few more commits in the directions we discussed.

Please take a look at fc1c983 and 4f54a73:

fc1c983 is largely wordsmithing and comment clean up in reconciliation, but I removed some cases from `__rec_child_modify()` that I believe are no longer needed.

4f54a73 is the change to prevent internal page splits during the last pass of checkpoint. With this change it appears like it's significantly less likely we'll see a failure in this CONFIG (I'm getting more test runs before failure), but it doesn't eliminate the failure altogether, there's still something wrong.

keithbostic:
@michaelcahill, I noticed the `WT_CKPT_PREPARE` value for `WT_BTREE.checkpointing` is never actually used? I'm guessing it's deliberate, but wanted to call it out to you.

michaelcahill:
@keithbostic, `WT_CKPT_PREPARE` is used: it is set at `bt_sync.c:116` and checked everywhere we have `btree->checkpointing != WT_CKPT_OFF`.

michaelcahill:
@keithbostic, I merged `develop` with the wildly optimistic hope that the fixes from WT-2367 might help...

keithbostic:
> `WT_CKPT_PREPARE` is used: it is set at `bt_sync.c:116` and checked everywhere we have `btree->checkpointing != WT_CKPT_OFF`.

What I was trying to say is that `WT_CKPT_PREPARE` is functionally equivalent to `WT_CKPT_RUNNING`: I think the only place we use them other than testing `checkpointing != WT_CKPT_OFF` is in a debugging assert in the split code.

keithbostic:
> I merged develop with the wildly optimistic hope that the fixes from WT-2367 might help...

Good thought, but no chance. :)

keithbostic:
@michaelcahill, @agorrod, I believe the last problem I've seen with this CONFIG is fixed in 3afbc37, we were skipping part of the split page pair during checkpoint.

I'm still running tests, but so far it looks good.

Notes:

* we set `first_dirty_txn` 3 times in the split code, maybe we need a function that does that, or different structuring?
* In 0f24204 I moved reconciliation initialization inside of the page lock: I don't see any specific bugs at the moment, but I'm concerned reading (and updating!) page transactional information before acquiring that lock could potentially collide with insert splits. There's not a lot of work done inside reconciliation initialization, but if anyone is concerned about that, we could split reconciliation initialization into two parts, before and after acquiring the lock.

Assuming no failures today, I think this one is ready to merge.

Backports for MongoDB:

* 5c22973, ab34a66 and 8b8614a (it's all one fix, I had some problems typing that day). It would also be pretty easy to create a smaller fix than I ended up with,
* 4f54a73, 3511f76 and 86b3d2f (it's all one fix, I reverted most of the first commit based on Michael's review),
* 3afbc37 (and maybe 6cc6b78, which is just code shuffling for clarity), and
* b96f95c

There are additional fixes in here, but I don't think any of them are relevant to MongoDB.

michaelcahill:
@keithbostic, thanks, lgtm -- I'll merge.

> we set first_dirty_txn 3 times in the split code, maybe we need a function that does that, or different structuring?

I *think* it's sufficient to set `right->modify->first_dirty_txn = page->modify->first_dirty_txn` for insert splits, but let's leave that for another change.  I wouldn't mind settting `first_dirty_txn` in `__split_parent` if that doesn't work out.

> In 0f24204 I moved reconciliation initialization inside of the page lock: I don't see any specific bugs at the moment, but I'm concerned reading (and updating!) page transactional information before acquiring that lock could potentially collide with insert splits. There's not a lot of work done inside reconciliation initialization, but if anyone is concerned about that, we could split reconciliation initialization into two parts, before and after acquiring the lock.

That's okay with me: I'd be surprised if this is measurable, but if it shows up in perf tests, let's revisit.

PR #2483, opened 2016-02-07
WT-2367: WT_CURSOR.next out-of-order returns failure

keithbostic:
@michaelcahill, @agorrod for your review/consideration.

### DIFF
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -9,18 +9,17 @@
 #include "wt_internal.h"
 
 /*
- * __wt_search_insert_append --
+ * __search_insert_append --
  *	Fast append search of a row-store insert list, creating a skiplist stack
  * as we go.
  */
 static inline int
-__wt_search_insert_append(WT_SESSION_IMPL *session,
-    WT_CURSOR_BTREE *cbt, WT_ITEM *srch_key, bool *donep)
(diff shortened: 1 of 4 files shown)


src/btree/row_srch.c
```diff
@@ -590,16 +583,16 @@ leaf_match:	cbt->compare = 0;
...
 		cbt->slot = WT_ROW_SLOT(page, page->pg_row_d + (base - 1));
 
-		cbt->ins_head = WT_ROW_INSERT_SLOT(page, cbt->slot);
+		ins_head = WT_ROW_INSERT_SLOT(page, cbt->slot);
 	}
 
 	/* If there's no insert list, we're done. */
-	if (WT_SKIP_FIRST(cbt->ins_head) == NULL)
+	if (WT_SKIP_FIRST(ins_head) == NULL)
 		return (0);
```
keithbostic: For the record, this line is the actual bug fix: if we race with row-modify such that we've created and swapped into place the page's insert list (but not put any items on the list), the iteration setup functions uses the non-NULL state of
the `cbt->ins_head` field to decide where to start the iteration, and returns the same key/value pair twice.

michaelcahill:

I spent some time trying to push this further, to get rid of `cbt->ins_head` entirely and find the list using the slot when it is needed.  I didn't get that change finished for all formats, but if we decide to do that, it's a big enough change in addition to this that it should get its own ticket.

PR #2484, opened 2016-02-09
WT-2387 Fix the error string we test for next_random on column stores

michaelcahill:


agorrod:

PR #2486, opened 2016-02-09
WT-2374 Fix inconsistent comment.

sueloverso:


michaelcahill:
@sueloverso, I've reviewed the code and I agree with your change.  The point of the "meta tracking" code in checkpoints is to gather up a set of handle locks and release them all when the checkpoint is done.  With bulk loads, there is only a single handle and it is already locked so the meta tracking code isn't needed.

I'll merge.

PR #2488, opened 2016-02-10
WT-2390 Fix OS X warning.

michaelcahill:
```
src/conn/conn_log.c:636:16: error: implicit conversion loses integer precision: 'wt_off_t' (aka 'long long') to 'uint32_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
         slot->slot_last_offset;
```


michaelcahill:
@sueloverso, FYI, I'm pretty sure this was the intent.  I thought about asserting that the offset is in range, but ended up just fixing the warning that has been breaking the OS X build.

sueloverso:
Thanks. LGTM

PR #2489, opened 2016-02-10
WT-2280 CRC32 for POWER 8 

markbenvenuto:
I have integrated this library https://github.com/antonblanchard/crc32-vpmsum that takes advantage of the new CRC32 SIMD instructions (called vpmsum*) available on POWER8. This library is dual licensed GPLv2+ and Apache License 2.

**Changes**
First commit is just the code from the original repo with an update to license.dox. Let me know if there is another file for me to update for licensing.

Second commit is the build integration, and fixups for `s_all`. The code builds on PPC64LE, and also x86 because of the defines.

Finally, I do not do a runtime check for POWER8 at the moment, let me know if you would like one added. FYI, POWER8 is the minimum platform for PPC64LE on Linux.

**Test Results**
When I run the `medium-btree` test in wtperf, I get the following results:

| Test | Software CRC32 | Hardware CRC32 |
| --- | --- | --- |
| Load | 35.09 sec | 24.29 sec |
| Run | 254972 ops/sec | 440983 ops/sec | 
Test Machine: Bare Metal, POWER 8, 10 Core, 8 SMT, 256 GB RAM, Ramdisk used to back WT files


michaelcahill:

It generally looks okay to me, but the licenses definitely need careful review, and there might be a cleaner way to hook this into WT than having Power-specific code in `src/support`.

keithbostic:
@markbenvenuto, I think this one is ready to go. If you'll confirm I didn't break anything, I'll do the merge.

Thanks!

keithbostic:
@michaelcahill, I thought about creating `src/thirdparty/power8` (which I don't like much), and `src/{crc_x86,crc_power8}/` which I like a bit better than the current approach and matches our per-OS directory structure.

I don't feel strongly, to be honest.

Preferences?

The licenses are fine; the Power8 code ships under GNU "V2 or later", so it matches the default WiredTiger license.

markbenvenuto:
@keithbostic I can confirm that the code still compiles on my ppc64le machine. Thanks

PR #2490, opened 2016-02-11
WT-2173: test/format cache stuck full 

keithbostic:
WT-2173: test/format cache stuck full

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1369,10 +1369,12 @@ __evict_get_ref(
     WT_SESSION_IMPL *session, bool is_server, WT_BTREE **btreep, WT_REF **refp)
 {
 	WT_CACHE *cache;
+	WT_CONNECTION_IMPL *conn;
 	WT_EVICT_ENTRY *evict;
 	uint32_t candidates;
 
-	cache = S2C(session)->cache;
+	conn = S2C(session);
+	cache = conn->cache;
 	*btreep = NULL;
 	*refp = NULL;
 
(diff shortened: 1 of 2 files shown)


keithbostic:
@agorrod, @michaelcahill:
 
WT-2173 is a heavily-threaded configuration (31 threads), relatively small cache (83MB, so 2MB per-working thread), and it's write-heavy (90% writes/inserts).

Leaf pages aren't too small, 128K, but the maximum in-memory leaf page size is a little more than 6MB. In practice, we drive up the size of the in-memory pages to about 12MB before we manage to evict them, so roughly 8 pages fit into memory at any one time. The configuration is for larger values (4KB), so updates will be large.

In other words, 30 threads, writing relatively large records into 8 relatively large pages.

Additionally there's only a single eviction server thread, there aren't any worker threads.
 
This is a bad configuration, of course, there's not enough cache for this many threads randomly updating the same chunks of data.
 
Eviction is usually failing because a thread is holding a hazard reference, or the page has a not-yet-globally visible change.

I've put two changes into this branch, only one of which is important.
 
First, db58fef: if there aren't any eviction work threads, or the cache is stuck, the server doesn't ignore half of the page candidates. I think this increases the amount of time spent gathering more pages for no particular gain, especially when the cache is stuck. I made this change when I was trying to figure out why the application threads couldn't evict pages but the eviction server could... which leads me to the second change.

Second, 98d083d reverts part of 5eb3df7. 5eb3df7 was part of the work for [SERVER-21585](https://jira.mongodb.org/browse/SERVER-21585), restricting the lookaside table to server threads. The problem I'm seeing is the eviction server finds a candidate, one of the worker threads grabs it, and the eviction attempt fails because that thread can't write the lookaside table. The eviction attempts that succeed are the ones where the eviction server attempts the eviction, but with 30 worker threads, it's relatively rare the eviction gets to the candidate page first.

I've reviewed SERVER-21585 and I don't see any discussion about this part of 5eb3df7, so I suspect that's something you two added based on your testing?

Regardless, I think there's an argument to make that if the cache is stuck, application threads are already paying a performance cost, and we should maximize their chances of succeeding at eviction.

With that change, this workload no longer gets stuck.

michaelcahill:
Thanks for digging into this one, @keithbostic -- I agree that once we're stuck, we might as well try as hard as we can to evict any candidates we can find.  I'll merge.

PR #2491, opened 2016-02-11
WT-2382: Problem with custom collator for 'u' format with join cursor

keithbostic:
Commit a6c183b removed the only caller of `__wt_struct_unpack_size`.


keithbostic:
@ddanderson, did we want to leave `__wt_struct_unpack_size` in place? It's currently unused.

agorrod:
lgtm

PR #2493, opened 2016-02-12
WT-2381: dump utility discards table config

keithbostic:
@sueloverso, I ended up waiting for a bunch more tests to run today, so I went a little further on this one.

I think what's going on here is that simple tables are a special case: there's no `colgroup` entry in a simple table's metadata entry, and we're not finding the table's underlying file information.

This works for complex tables because their column-groups are listed in the table's metadata entry: even though we will write the same (incorrect) information for the complex table that we write for the simple table, it will be overridden by the correct information stored for the specific column-groups.

I did a "fix" by special-casing simple tables in the dump code, but that could be completely wrong. I don't have a handle on how this "ought" to behave, it's just the only path to victory I saw.

I thought a little bit about testing: the only thing that came to mind was changing test_dump to compare the metadata before and after the dump and re-load (dump the `WiredTiger.wt` file before and after the dump/re-load and assert it's the same). Obviously, you have to strip all of the checkpoint information and maybe some other stuff before that will work -- I didn't go far down that path. We'd need some way to parse the dump output in Python, too. There's a Python package to do that kind of parsing ([pyparsing](http://pythonhosted.org//pyparsing/), or [here](http://stackoverflow.com/questions/1651487/python-parsing-bracketed-blocks)), but we'd have to make sure it's installed to use it?

Anyway, hope this is useful to you, just toss the branch if it's not.


michaelcahill:
autobuild: retest this please

sueloverso:
@keithbostic I added a new unit test.  It fails 3 out of 5 scenarios.  The good news is that your change on this branch does help - on develop it fails 4 out of 5 scenarios.

keithbostic:
And it removes the need for `metadata:create`.  It does mean we'd return "special" stuff out of the metadata file, but I don't think we care. Good catch, I'll take a run at it.

keithbostic:
@michaelcahill, I've gone ahead and pushed the change you suggested.

As far as I'm concerned, this one is ready for merge.

michaelcahill:

PR #2495, opened 2016-02-15
WT-2397: Cursor traversal from end of the tree skips records.

keithbostic:
Positioning a cursor at the end of the tree can race with page splits.
During the initial descent of the tree in the cursor.prev case, do the
same tests for races we do for a search past the end of the tree.

### DIFF
--- a/src/btree/bt_walk.c
+++ b/src/btree/bt_walk.c
@@ -89,11 +89,11 @@ __ref_is_leaf(WT_REF *ref)
 }
 
 /*
- * __page_ascend --
+ * __ref_ascend --
  *	Ascend the tree one level.
  */
-static void
-__page_ascend(WT_SESSION_IMPL *session,
+static inline void
+__ref_ascend(WT_SESSION_IMPL *session,
     WT_REF **refp, WT_PAGE_INDEX **pindexp, uint32_t *slotp)
 {
 	WT_REF *parent_ref, *ref;
@@ -163,11 +163,11 @@ __page_ascend(WT_SESSION_IMPL *session,
 }
 
 /*
- * __page_descend_prev --
+ * __ref_descend_prev --
  *	Descend the tree one level, during a previous-cursor walk.
  */
-static void
-__page_descend_prev(
+static inline void
+__ref_descend_prev(
     WT_SESSION_IMPL *session, WT_REF *ref, WT_PAGE_INDEX **pindexp)
 {
 	WT_PAGE_INDEX *pindex;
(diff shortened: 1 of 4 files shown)


keithbostic:
@agorrod, @michaelcahill: a different approach that works for awhile, but eventually races (using Adam's test program).

agorrod:
@keithbostic This branch passes my local testing, and testing using the test case that was used to report the problem.

src/btree/bt_walk.c
```diff
@@ -163,12 +163,12 @@ __page_ascend(WT_SESSION_IMPL *session,
...
 /*
- * __page_descend --
- *	Descend the tree one level.
+ * __page_descend_prev --
+ *	Descend the tree one level, during a previous-cursor walk.
  */
 static void
-__page_descend(WT_SESSION_IMPL *session,
-    WT_PAGE *page, WT_PAGE_INDEX **pindexp, uint32_t *slotp, bool prev)
+__page_descend_prev(
```
agorrod: For consistency this function should change name to `__ref_descend_prev` due to the parameter change.

src/btree/bt_walk.c
```diff
@@ -233,21 +230,41 @@ __page_descend(WT_SESSION_IMPL *session,
...
 	}
 	*pindexp = pindex;
 }
 
 /*
+ * __page_initial_descent_prev --
+ *	Descend the tree one level, when setting up the initial cursor position
+ * for a previous-cursor walk.
+ */
+static bool
```
agorrod: I'd mark this as inline.

The "initial" in the name makes me think we are descending from the root page, but I see that it matches the local variable in the caller. I don't love it but don't have a good alternative suggestion.

agorrod:
@keithbostic I've made a couple of minor comments. Apart from those I'm happy with this patch.

src/btree/bt_walk.c
keithbostic: Agreed, I'll make a consistency pass.

src/btree/bt_walk.c
keithbostic: Will do. Yes, the reason I chose the name was to match the local variable in the caller, open to anything better, I'm not excited by the choice either.

keithbostic:
@michaelcahill, can you please review this change?

Alex and I are happy with it, but I'd appreciate another set of eyes.

michaelcahill:
@keithbostic, thanks, I've reviewed and I'm happy with this change.  I'll merge.

PR #2496, opened 2016-02-16
WT-2396 Grab the checkpoint lock around drops and renames.

michaelcahill:
Since database-wide checkpoints have references to all handles in the
system, drops are generally not possible during a checkpoint regardless
of this change.  This prevents a deadlock where checkpoint waits on the
handle list lock after it has some handles locked exclusive.

A better longer-term solution would be to have checkpoint gather all
handles (including locking old checkpoints) at the beginning so it
doesn't have to acquire the handle list lock after locking handles.

### DIFF
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -597,9 +597,10 @@ __session_rename(WT_SESSION *wt_session,
 	WT_ERR(__wt_str_name_check(session, uri));
 	WT_ERR(__wt_str_name_check(session, newuri));
 
-	WT_WITH_SCHEMA_LOCK(session, ret,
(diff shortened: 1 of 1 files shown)


agorrod:
For the record we made a similar change for salvage, upgrade and verify here: https://jira.mongodb.org/browse/WT-1404

agorrod:
@michaelcahill I'm happy with this change. There is a genuine conflict between checkpoints and drops at the moment, so this seems like a reasonable solution. Though from an application point of view I don't think it's ideal, since checkpoints are orthogonal to other operations, so it can be difficult to predict when they will get EBUSY.

PR #2497, opened 2016-02-16
WT-2399 Add initial implementation for cursor traversal test application

agorrod:
Based loosely on the test/checkpoint implementation. Hopefully we can
refactor, so code is shared between the two via test_util.i.

### DIFF
--- a/build_posix/Make.subdirs
+++ b/build_posix/Make.subdirs
@@ -26,6 +26,7 @@ lang/python PYTHON
 # Make the tests
 test/bloom
 test/checkpoint
+test/cursor_order
 test/fops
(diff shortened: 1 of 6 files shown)


agorrod:
This isn't ready for merging yet.

agorrod:
@michaelcahill Could you please review this once it passes the pull request tester? The test fails without the fix for WT-2397.

I've started structuring the code so that it can be expanded to cover more cursor traversal patterns, I expect the structure to evolve over time.

michaelcahill:
Thanks, @agorrod, let's get this in.  I'll merge...

PR #2498, opened 2016-02-17
SERVER-22676 Don't check duplicated create calls for matching configurations

agorrod:
We added the check in WT-2295 as part of another fix. We have never
checked before and it makes maintaining backward compatability more
complex in the future for limited gain.


agorrod:
@michaelcahill Here is a patch that backs out the additional error checking for create calls on an existing table. Would you please review?

michaelcahill:

PR #2501, opened 2016-02-18
WT-2405: test utility error handling.

keithbostic:
Add a check routine to make simple error handling easier.

@agorrod, @daveh86, here's the example branch.

If you like this, I thought we could add a corresponding macro to `test_util.i`.


agorrod:
@keithbostic I like it, I was a bit concerned about wrapping function calls I'm likely to step through in a debugger in a macro, but that works fine as long as the macro stays simple.

I think we should take this branch, and add similar functionality to `test_util.i`

daveh86:
@keithbostic, I agree with @agorrod, this is a good thing to have and that adding to test_util.i is a good direction. I think there is a good argument for someone (me?) taking a bigger look at how we can unify and standardise much of the testing going forward, although I think that is a discussion for another day.

sueloverso:
Before we merge this branch, can we add it to all the other programs in test/*?  If no one gets to it today, I can whip it up tomorrow.  I'd prefer that if it be named `testutil_check` if it is in testutil.i.

keithbostic:
@sueloverso, fbm, if you don't get to it, I'm happy to do the pass as well.

FWIW, I thought we might might want two other flavors of that macro: one that outputs `errno` instead of the functions return (so we could use it for C library calls), and maybe one that took more complex error function signatures so we could use the macro in places where there were legitimate reasons to output additional information in the error message.

I wasn't convinced about either of those additions, though. I was going to try and get a feel for if they'd be useful as I reviewed calls I couldn't convert, and I don't want some fantastical "perfect" design to get in the way of moving things forward in a useful way.

Agreed on `testutil_check`.

agorrod:
I've pushed some changes into this branch, including moving the macro into the shared test utilities and adding it's usage to test/bloom.

@keithbostic I also converted the test/format die function to use the version provided by the utilities.

I also think it's worth converting all of our tests to use this pattern.

> one that took more complex error function signatures so we could use the macro in places where there were legitimate reasons to output additional information in the error message.

I'm cautious about this - I'm concerned that more sophisticated error handling macros would make debugging more difficult. I think that means that I'm in agreement - lets get this simple version done, then consider extending it.

sueloverso:
My feeling on a more complex macro is that, at least to start, I am not a fan.  The macro is great for all the WT calls that are part of running but we expect to succeed and don't want to think about too much.  These are test programs we write and if there is special error handling or expected errors I prefer they be right there in my face.  The fact they are explicit code and not using a macro will make them noteworthy.

keithbostic:
@agorrod, @sueloverso, I agree about special error handling, all I'm proposing is a way to associate additional error output with calls "we expect to succeed and don't want to think about too much".

Take a look at #2508 and you'll see what I mean. If you don't like, just discard that branch.

sueloverso:
Thanks for clearing that up @keithbostic.  I agree with the change you made and merged it here.

agorrod:
Thanks @keithbostic lgtm, I'll merge. We can complete the translation in the other tests later.

PR #2502, opened 2016-02-18
WT-2406: cursor_order lint, minor bug fixes

keithbostic:


agorrod:
Thanks @keithbostic

PR #2503, opened 2016-02-18
WT-2407: recovery lint

keithbostic:


sueloverso:
Thanks @keithbostic . LGTM.

PR #2504, opened 2016-02-18
WT-2395 Use allocation_size from metadata.

sueloverso:
If we find a file that needs to be created, use its metadata config
in case it sets its own allocation size.

@michaelcahill or @agorrod please review this change.  I have confirmed it fixes the issue with the ticket and the test passes and continues.  Basically it allows it to create the LSM chunk with the correct allocation size.


michaelcahill:
Thanks, @sueloverso -- this lgtm.  Nice catch!  I'll merge.

PR #2505, opened 2016-02-19
WT-2411 Drop the checkpoint lock when LSM is draining its queue.

michaelcahill:


michaelcahill:
This is a trivial fix for test hangs, I'm going to merge.

PR #2506, opened 2016-02-19
WT-2349 Fix Jenkins memory failures

sueloverso:
@keithbostic If you're around please review a small diff to fix a few of the Jenkins failures from the readonly merge.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -136,9 +136,9 @@ config_free(CONFIG *cfg)
 	cleanup_truncate_config(cfg);
 	free(cfg->base_uri);
 	free(cfg->ckptthreads);
-	free((char *)cfg->partial_config);
+	free(cfg->partial_config);
 	free(cfg->popthreads);
-	free((char *)cfg->reopen_config);
+	free(cfg->reopen_config);
 	free(cfg->workers);
 	free(cfg->workload);
 }
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -134,9 +134,11 @@ config_free(CONFIG *cfg)
 	}
 
 	cleanup_truncate_config(cfg);
-	free(cfg->ckptthreads);
-	free(cfg->popthreads);
 	free(cfg->base_uri);
+	free(cfg->ckptthreads);
+	free((char *)cfg->partial_config);
+	free(cfg->popthreads);
+	free((char *)cfg->reopen_config);
 	free(cfg->workers);
 	free(cfg->workload);
 }
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -1533,8 +1533,8 @@ close_reopen(CONFIG *cfg)
 		lprintf(cfg, ret, 0, "Closing the connection failed");
 		return (ret);
 	}
-	if ((ret = wiredtiger_open(cfg->home,
-	    NULL, (const char *)cfg->reopen_config, &cfg->conn)) != 0) {
+	if ((ret = wiredtiger_open(
+	    cfg->home, NULL, cfg->reopen_config, &cfg->conn)) != 0) {
 		lprintf(cfg, ret, 0, "Re-opening the connection failed");
 		return (ret);
 	}
@@ -1810,8 +1810,8 @@ create_tables(CONFIG *cfg)
 
 	for (i = 0; i < cfg->table_count; i++) {
 		if (cfg->log_partial && i > 0) {
-			if (((ret = session->create(session, cfg->uris[i],
-			    (const char *)cfg->partial_config)) != 0)) {
+			if (((ret = session->create(session,
+			    cfg->uris[i], cfg->partial_config)) != 0)) {
 				lprintf(cfg, ret, 0,
 				    "Error creating table %s", cfg->uris[i]);
 				return (ret);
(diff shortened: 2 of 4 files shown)


sueloverso:
@keithbostic I believe this is ready for review now.

bench/wtperf/config.c
```diff
@@ -134,9 +134,11 @@ config_free(CONFIG *cfg)
 	}
 
 	cleanup_truncate_config(cfg);
+	free(cfg->base_uri);
 	free(cfg->ckptthreads);
+	free((char *)cfg->partial_config);
 	free(cfg->popthreads);
-	free(cfg->base_uri);
+	free((char *)cfg->reopen_config);
 	free(cfg->workers);
```
keithbostic: @sueloverso, why do we need to cast the arguments to free?

src/conn/conn_api.c
```diff
@@ -2199,6 +2201,8 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
 	__wt_scr_free(session, &i2);
 	__wt_scr_free(session, &i3);
 
+	if (free_merge_cfg)
+		__wt_free(session, merge_cfg);
```
keithbostic: @sueloverso, if we initialize `merge_cfg` to NULL, I don't think we need the additional boolean (and we don't have to test before "calling" `__wt_free`, either.

sueloverso: Yes, if the connection is not readonly, then `merge_cfg` is the final merged config that ends up in `conn->cfg` and that needs to stay for the lifetime of the connection.  For readonly, it is part 1 and the readonly config overrides are added in and we merge again.  In that case, `conn->cfg` is that new merged  config and we need to free `merge_cfg`.

bench/wtperf/config.c
sueloverso: Without it, `free()` complains:
```
../../../bench/wtperf/config.c:139: warning: passing argument 1 of ‘free’ discards qualifiers from pointer target type
```

src/conn/conn_api.c
keithbostic: Then I'd suggest:
```
diff --git a/src/conn/conn_api.c b/src/conn/conn_api.c
index 25db237..5d5244e 100644
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1887,7 +1887,7 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
        WT_DECL_RET;
        const WT_NAME_FLAG *ft;
        WT_SESSION_IMPL *session;
-       bool config_base_set, free_merge_cfg;
+       bool config_base_set;
        const char *enc_cfg[] = { NULL, NULL }, *merge_cfg;
        char version[64];
 
@@ -1899,7 +1899,7 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
 
        conn = NULL;
        session = NULL;
-       free_merge_cfg = false;
+       merge_cfg = NULL;
 
        WT_RET(__wt_library_init());
 
@@ -2025,10 +2025,11 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
                 * read-only are tested in their individual locations later.
                 */
                __conn_config_readonly(cfg);
-               free_merge_cfg = true;
                WT_ERR(__wt_config_merge(session, cfg, NULL, &conn->cfg));
-       } else
+       } else {
                conn->cfg = merge_cfg;
+               merge_cfg = NULL;
+       }
 
        /*
         * Configuration ...
@@ -2201,8 +2202,8 @@ err:      /* Discard the scratch buffers. */
        __wt_scr_free(session, &i2);
        __wt_scr_free(session, &i3);
 
-       if (free_merge_cfg)
-               __wt_free(session, merge_cfg);
+       __wt_free(session, merge_cfg);
+
        /*
         * We may have allocated scratch memory when using the dummy session or
         * the subsequently created real session, and we don't want to tie down
```

Basically, to "give away" the memory, clear the original, and there's never more than a single reference.

I don't like it when two variables have to stay in sync, it always makes me nervous that code movement in the future will break things.

bench/wtperf/config.c
keithbostic: If you have to cast it, then I'd cast it to `(void *)`, because that's the type free takes.

But, this variable isn't really a const, we're casting it in other places so we can modify it:
```
        cfg->reopen_config = dcalloc(req_len, 1);
        if (cfg->readonly)
                snprintf((char *)cfg->reopen_config, req_len, "%s%s",
                    (char *)cfg->conn_config, READONLY_CONFIG);
        else
                snprintf((char *)cfg->reopen_config, req_len, "%s",
                    (char *)cfg->conn_config);
```
That cast of `cfg->conn_config` looks pretty odd, too.

src/conn/conn_api.c
sueloverso: Done and pushed.

bench/wtperf/wtperf.c
```diff
@@ -1810,8 +1810,8 @@ create_tables(CONFIG *cfg)
 
 	for (i = 0; i < cfg->table_count; i++) {
 		if (cfg->log_partial && i > 0) {
-			if (((ret = session->create(session,
-			    cfg->uris[i], cfg->partial_config)) != 0)) {
+			if (((ret = session->create(session, cfg->uris[i],
+			    (const char *)cfg->partial_config)) != 0)) {
```
keithbostic: @sueloverso, I'm not following the need for the `(const char *)` cast here, or above?

PR #2507, opened 2016-02-19
WT-2410: Casting function pointers to different types

keithbostic:

### DIFF
--- a/src/block/block_mgr.c
+++ b/src/block/block_mgr.c
@@ -69,6 +69,21 @@ __bm_checkpoint(WT_BM *bm,
 }
 
 /*
+ * __bm_checkpoint_readonly --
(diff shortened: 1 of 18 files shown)


sueloverso:
@keithbostic There is some sadness here with the explosion of new functions.  What about consolidation of functions that take an identical signature?  That is the direction I started in with the session notsup functions since there were 2 primary signatures.

keithbostic:
@sueloverso, I thought about consolidating the functions based on signature, and decided against it, but I don't feel strongly.

If you feel strongly I'd make that change.

keithbostic:
The branch with consolidated functions based on signature is #2510.

agorrod:

PR #2508, opened 2016-02-19
WT-2405: test utility error handling.

keithbostic:
Add testutil_checkfmt, a macro that supports additional error output.


sueloverso:
Thanks Keith - I do like this.  LGTM

PR #2510, opened 2016-02-19
WT-2410: Casting function pointers to different types

keithbostic:
@agorrod, @sueloverso, here's the other approach.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -930,6 +930,7 @@ setkv
 setstr
 setv
 sfence
+sig
 sii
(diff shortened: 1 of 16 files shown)


agorrod:
@keithbostic I prefer the original approach (https://github.com/wiredtiger/wiredtiger/pull/2507), though I'm glad you coded this up. We could move to auto generating the stubs if maintenance/code duplication become concerns.

michaelcahill:
Agreed, this looks like a maintenance headache to me.  I'll close...

PR #2511, opened 2016-02-21
WT-2376: Modules should compile without including wt_internal.h header file

keithbostic:
@agorrod, for your review consideration, the changes to not include `wt_internal.h` in the compression modules.


agorrod:
Thanks @keithbostic lgtm, I'll merge.

PR #2512, opened 2016-02-21
WT-2107: Add example code including an event handler

keithbostic:

### DIFF
--- a/examples/c/Makefile.am
+++ b/examples/c/Makefile.am
@@ -12,6 +12,7 @@ noinst_PROGRAMS = \
 	ex_cursor \
 	ex_data_source \
 	ex_encrypt \
+	ex_event_handler \
 	ex_extending \
 	ex_extractor \
 	ex_hello \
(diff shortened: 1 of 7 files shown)


agorrod:
@keithbostic I've made some changes in here, would you mind reviewing the new version?

I pulled the example out of ex_all.c, since that file has grown over 1000 lines I think it is losing it's utility as an example.

I switched the name of the handler functions from `wiredtiger_handle_` to `handle_wiredtiger_` to make it clear they aren't part of the WiredTiger namespace.

Apart from that it was general shuffling to suit my preference, so feel free to back out my changes if you think they make the example/documentation less clear.

keithbostic:
@agorrod, this looks great to me, thanks!

I agree with moving code out of ex_all.c, but we don't list ex_all.c in the documentation as "an example program". If it's not an example program, I don't worry about how big it is or if it makes sense, it's just a place to put random code snippets for the documentation.

PR #2514, opened 2016-02-22
SERVER-22554 Fix a reference counting bug in dhandles.

agorrod:
The code has diverged a lot between 3.0 and the latest develop, and
a backport included half of a reference counting change. This
patches up the reference counting in 3.0. The problem being solved
is:
There are two phases to opening a handle in a session for the first
time. First retrieve or open the handle from the connection cache,
then add it to the session cache. The code was structured to split
those into two separate phases (one while holding a lock the other
after the lock had been released). The reference count needs to be
bumped while the lock is being held, the session cache doesn't need
to be updated while holding the lock.

This change does both while holding the lock to keep reference count
tracking sane and correct.

### DIFF
--- a/src/session/session_dhandle.c
+++ b/src/session/session_dhandle.c
@@ -360,10 +360,32 @@ __session_dhandle_sweep(WT_SESSION_IMPL *session)
  */
 static int
 __session_find_shared_dhandle(WT_SESSION_IMPL *session,
-    const char *uri, const char *checkpoint, uint32_t flags)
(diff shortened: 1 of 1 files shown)


michaelcahill:
autobuild: retest this please

agorrod:

michaelcahill:
Thanks, @agorrod -- lgtm.  Let's discuss the merge in person to make sure it will appear correctly in the next drop...

PR #2515, opened 2016-02-22
WT-2394 - Have pages split as part of compact checkpoints use first-fit

daveh86:
Change to make compact significantly more robust. Resolves one very major cause of the short cutting of compact runs. I was able to run 150x of the compact test with 0 failures, with both the 128KB and 64KB block sizes.


daveh86:
@agorrod, would you mind taking a look at this? I'm not 100% if this is the optimal approach, and if not can we discuss?

keithbostic:
First, @daveh86, thank you for figuring this one out -- I shouldn't have just assumed it was the block sizes leading to failure.

Second, I think the best solution is probably to move compaction start/stop up into the session layer so that all operations done during the compaction process are done first-fit.

Can you and @agorrod take a look at #2518, and see what you think?

There is one problem this change makes a lot worse: by changing first-fit configuration to span logical operations (like file compaction and checkpoints), we open ourselves up to races with other compactions, or the file being dropped, for example. We can close the window a lot, but I don't see anything that would cleanly handle the file being dropped inside the loop where we're compacting the file and doing checkpoints, for example.

michaelcahill:
@keithbostic I've put code into #2525 that I think should deal with the handle configuration / locking / cleanup issues.  Please take a look.

keithbostic:
@michaelcahill, @daveh86, I'm happy to merge this one when you're ready.

michaelcahill:
Thanks @daveh86 and @keithbostic, lgtm.  I'll merge.

PR #2516, opened 2016-02-22
WT-2422: multiple definitions of custom die function

keithbostic:


agorrod:
Thanks @keithbostic lgtm. I'd hoped to avoid tests needing to define the function if they didn't want to use it, but it's better for the code to be correct ;)

PR #2519, opened 2016-02-22
SERVER-22784: Coverity analysis defect 77722: Unused value

keithbostic:

Please look at this one fairly closely, it's a minor change, but to the walk code.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -437,6 +437,7 @@ cfkos
 change's
 changelog
 chdir
+checkfmt
 checkpointed
 checkpointer
 checkpointing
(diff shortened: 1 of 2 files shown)


agorrod:
@keithbostic lgtm, though it took a while to convince myself - I think that means `__tree_walk_internal` is too complex. @michaelcahill another set of eyes won't hurt before merging this one.

michaelcahill:
Yes, lgtm -- I'll merge.

FWIW, I think this particular Coverity defect is more noise than signal, so if we get a lot more like this, let's talk.

keithbostic:
OK. We don't see a lot of this kind of Coverity report.

(But I do I try and fix these things when I can, to keep down the noise. Otherwise we get a refreshed complaint every time the code moves.)

PR #2520, opened 2016-02-22
WT-2419 - Fix missing __func__ defintion under MSVC

daveh86:


daveh86:
@keithbostic, can you have a look at this when you get a moment? This is the patch discussed on WT-2419

keithbostic:
lgtm, merging.

PR #2521, opened 2016-02-23
WT-2423 Cleanup on error when opening a session handle.

agorrod:
It's a bug, but found by inspection. Vanishingly unlikely.

### DIFF
--- a/src/session/session_dhandle.c
+++ b/src/session/session_dhandle.c
@@ -11,31 +11,6 @@
 static int __session_dhandle_sweep(WT_SESSION_IMPL *);
 
 /*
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic Could you review this change please?

The change doesn't diff well, might be best looking at the change standalone. The error handling in `__session_add_dhandle` is unusual now, go for it if you think you can make it cleaner.  I toyed with changing `__session_discard_handle` so that it could cleanup when the handle hadn't yet been added to the cache, but that was at least as ugly as this.

keithbostic:
@agorrod, there's an alternate approach in #2527.

PR #2523, opened 2016-02-23
WT-2409 - Have LSM checkpoints wait for locks.

daveh86:
Having the checkpoints wait for a lock instead of looping round and retrying later had a positive performance impact  - to the tune of 10%.

### DIFF
--- a/src/lsm/lsm_work_unit.c
+++ b/src/lsm/lsm_work_unit.c
@@ -341,13 +341,11 @@ __wt_lsm_checkpoint_chunk(WT_SESSION_IMPL *session,
 	 * time, and our checkpoint operation should be very quick.
 	 */
 	WT_ERR(__wt_meta_track_on(session));
-	F_SET(session, WT_SESSION_LOCK_NO_WAIT);
 	WT_WITH_CHECKPOINT_LOCK(session, ret,
 	    WT_WITH_SCHEMA_LOCK(session, ret,
 		ret = __wt_schema_worker(
 		session, chunk->uri, __wt_checkpoint, NULL, NULL, 0)));
 	WT_TRET(__wt_meta_track_off(session, false, ret != 0));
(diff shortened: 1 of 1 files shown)


daveh86:
@agorrod, you are likely looking at this anyway, so can you please review?

src/lsm/lsm_work_unit.c
```diff
@@ -341,20 +341,13 @@ __wt_lsm_checkpoint_chunk(WT_SESSION_IMPL *session,
...
 	WT_ERR(__wt_meta_track_on(session));
-	F_SET(session, WT_SESSION_LOCK_NO_WAIT);
 	WT_WITH_CHECKPOINT_LOCK(session, ret,
 	    WT_WITH_SCHEMA_LOCK(session, ret,
 		ret = __wt_schema_worker(
 		session, chunk->uri, __wt_checkpoint, NULL, NULL, 0)));
 	WT_TRET(__wt_meta_track_off(session, false, ret != 0));
-	F_CLR(session, WT_SESSION_LOCK_NO_WAIT);
-	if (ret != 0) {
-		if (ret == EBUSY) {
```
agorrod: @daveh86 This change makes me nervous, what it means is that the LSM worker is now going to sit waiting for a possibly long running transaction to finish. Since LSM workers are shared by all LSM trees in a database there may be other work it can do in the mean time that is being starved out because we are waiting here. Could you try a different change, which is to still set the `WT_SESSION_LOCK_NO_WAIT` flag like we used to, but in the error handling add a new work unit to the queue so the checkpoint operation is tried again. Something like:

``
WT_ERR(__wt_lsm_manager_push_entry(
                    session, WT_LSM_WORK_FLUSH, 0, lsm_tree));
``

agorrod:
Thanks @daveh86 lgtm - I'll merge. Could you keep an eye on Jenkins for the next couple of days to ensure this has (only) the desired effect.

PR #2524, opened 2016-02-23
WT-2420 Gather backup metadata in a single pass.

michaelcahill:
Previously, we did one pass to gather entries for the backup file, and a
separate pass to gather handles.  Since WT-2395, those two passes could
be out-of-sync, leading to files missing from a backup.


michaelcahill:
@sueloverso, can you please review?  If you could also kick off the test that was failing occasionally on this branch, that would be great.  Thanks!

sueloverso:
Thanks @michaelcahill.  LGTM.  I have started the test with both CONFIGs that failed.  I will merge after they have both run a while.

sueloverso:
Ok, each CONFIG has run about 1000 iterations successfully.  While I'll let them keep running until my script ends (at 10000 iterations), I will merge this.

PR #2525, opened 2016-02-23
WT-2394 Gather handles for compact at the beginning.

michaelcahill:
This allows us to set the "first fit" flag once and clear it consistently.  It also prevents drop or other exclusive operation during compact without reentering the schema lock each time.


michaelcahill:
@daveh86, @keithbostic, for your consideration.

keithbostic:
lgtm, I'll merge.

PR #2527, opened 2016-02-23
WT-2423 Cleanup on error when opening a session handle.

keithbostic:
@agorrod, I tried a different approach, for your review/consideration.

### DIFF
--- a/src/session/session_dhandle.c
+++ b/src/session/session_dhandle.c
@@ -11,6 +11,28 @@
 static int __session_dhandle_sweep(WT_SESSION_IMPL *);
 
 /*
+ * __session_add_dhandle --
+ *	Add a handle to the session's cache.
+ */
+static int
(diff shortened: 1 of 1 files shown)


src/session/session_dhandle.c
```diff
@@ -463,14 +447,23 @@ __session_get_dhandle(
 		return (0);
 	}
 
+	/* Sweep the handle list to remove any dead handles. */
+	WT_RET(__session_dhandle_sweep(session));
```
agorrod: My concern with this change is that we could kick the handle we are about to access out of cache if the sweep happens before the find. That won't break anything, but it will mean there is more work (and locking) in this path.

agorrod:
Thanks @keithbostic lgtm. My question about the session cache wasn't real, since we've checked the session cache before the call to sweep.

PR #2529, opened 2016-02-24
WT-2427 Have wtperf use builtin compression when configured.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'll merge once it passes testing.

PR #2530, opened 2016-02-25
WT-2428 Add a "json" mode to statistics logging.

michaelcahill:
This generates statistics in a format that is compatible with MongoDB
so that we can share tools for analysis.


dist/stat_data.py
```diff
@@ -315,7 +315,7 @@ def __init__(self, name, desc, flags=''):
     YieldStat('page_sleep', 'page acquire time sleeping (usecs)'),
 ]
 
-connection_stats = sorted(connection_stats, key=attrgetter('name'))
+connection_stats = sorted(connection_stats, key=attrgetter('desc'))
```
michaelcahill: These changes resulted in most of the line movement -- I needed the statistics logically grouped rather than being returned in order of the internal names we give them.  I think it makes more sense anyway: our names aren't chosen consistently to group stats together but the type prefixes in the description are.

dist/stat.py
```diff
@@ -98,11 +98,11 @@ def print_defines():
 f.close()
 compare_srcfile(tmp_file, '../src/include/wiredtiger.in')
 
-def print_func(name, handle, list):
+def print_func(name, handle, statlist):
     '''Print the structures/functions for the stat.c file.'''
     f.write('\n')
     f.write('static const char * const __stats_' + name + '_desc[] = {\n')
-    for l in list:
+    for l in statlist:
```
michaelcahill: Just a Python style nit: `list` is a builtin.

dist/stat.py
```diff
@@ -143,7 +143,7 @@ def print_func(name, handle, list):
 __wt_stat_''' + name + '_clear_single(WT_' + name.upper() + '''_STATS *stats)
 {
 ''')
-    for l in sorted(list):
+    for l in statlist:
```
michaelcahill: We used to sort all of the lists again here, into a different order.  I've changed the input to just sort them once and generate everything consistently from that list.

michaelcahill:
I don't like that setting `json` changes the timestamp default, but (a) changing the default for real would break existing tools, (b) having to remember the timestamp format that MongoDB uses every time we want to monitor stats would be a pain and (c) it turns out to be difficult to pass in timestamp formats because they need quoting and `wtperf`'s config file parsing doesn't make that easy.

michaelcahill:
All of that said, this is the payoff (from a `wtperf` workload):

wt-timeseries

agorrod:
Thanks @michaelcahill - being able to generate JSON format statistics is very cool. I'll merge.

PR #2531, opened 2016-02-25
WT-2429 Add a statistic to track eviction aggressive mode.

agorrod:


agorrod:
@michaelcahill Would you mind taking a look at this branch please? I thought about making the statistic more general, so it wouldn't need to change if we make aggressive a scale in the future rather than an on/off switch, but figured something that matches reality is better than potential future proofing.

michaelcahill:
Did you consider just exporting `cache->state` as a stat?  That would give more information than just aggressive on/off, but is potentially harder to interpret?

agorrod:
> Did you consider just exporting cache->state as a stat?

I thought about that and the primary down side was how hard it would be to interpret the data. The other downside is that the other state I think is related is the cache stuck state - but that's stored in a separate field (cache->flags). It's a trade off between available information, self describing statistics and code complexity.

michaelcahill:
Fair enough, I agree that trying to map flags onto a statistic is fraught.

One other idea: what about exporting the `loop` value in the main eviction server loop?  That should hover near zero when eviction is making progress and ramp up to 100 (at which point we get "stuck") when eviction is struggling.

michaelcahill:
@agorrod, as discussed in person, having this stat is definitely an improvement, and it has the virtue of being easy to explain and interpret.  Let's merge.

PR #2532, opened 2016-02-25
SERVER-22831 Queue more leaf pages than internal pages for eviction.

michaelcahill:
Unless we get aggressive, putting internal pages on the LRU queue is
counterproductive.  That is particularly true for workloads that
transition to read-only, where some tables are not being queried.  In
that case, all of the leaf pages are evicted, and eviction wastes a lot
of effort walking and queuing internal pages that are not actually
evicted.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1267,7 +1267,8 @@ __evict_walk_file(WT_SESSION_IMPL *session, u_int *slotp)
 
 		/* Pages we no longer need (clean or dirty), are found money. */
 		if (__wt_page_is_empty(page) ||
-		    F_ISSET(session->dhandle, WT_DHANDLE_DEAD))
+		    F_ISSET(session->dhandle, WT_DHANDLE_DEAD) ||
+		    page->read_gen == WT_READGEN_OLDEST)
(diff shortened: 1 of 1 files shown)


src/evict/evict_lru.c
```diff
@@ -1267,7 +1267,8 @@ __evict_walk_file(WT_SESSION_IMPL *session, u_int *slotp)
 
 		/* Pages we no longer need (clean or dirty), are found money. */
 		if (__wt_page_is_empty(page) ||
-		    F_ISSET(session->dhandle, WT_DHANDLE_DEAD))
+		    F_ISSET(session->dhandle, WT_DHANDLE_DEAD) ||
+		    page->read_gen == WT_READGEN_OLDEST)
```
michaelcahill: This is really a separate change: I just noticed while in the area that we could skip (particularly internal) pages that had been marked with `__wt_page_evict_soon`.

agorrod:
@michaelcahill I like this change - lets see how the Jenkins performance jobs go with it before merging.

agorrod:
lgtm, and perf testing didn't show anything surprising. I'll merge.

PR #2533, opened 2016-02-26
WT-2430 Ensure that statistics cursor is always positioned at the first 'set' of statistics.

ddanderson:
This fix is relevant for getting statistics on join cursors.  This fixes the user submitted test case, and also passes a new test case, which is derived from the user submitted one.


ddanderson:

keithbostic:
lgtm, merged.

PR #2534, opened 2016-02-26
WT-2433 Allow statistics logging in readonly mode.

sueloverso:
@michaelcahill Here are changes for allowing statistics logging even in read-only mode.  This is helpful for debugging the performance differences.  Please review.


michaelcahill:
@sueloverso, thanks, lgtm -- I'll merge.

PR #2536, opened 2016-02-29
WT-2434 Fix a race between forced drops and sweep.

michaelcahill:
During a forced drop, trees are marked dead but remain in cache
temporarily.  The sweep thread is aggressive about trying to discard
dead trees from cache.

If sweep found a tree marked dead before the drop operation had
completed, they could race setting / clearing the "no eviction" flag on
the tree, allowing eviction to run concurrent with sweep's discard.
This could cause a segfault where pages are discarded underneath
eviction.

The fix is to move setting / clearing the "no eviction" flag inside the
spinlock that protects against concurrent close operations.  That way,
even if sweep does find a tree before drop has finished closing it,
sweep will block on the close lock before attempting to discard it.
While in the area, tighten up other checks relating to the "no
eviction" flag.

### DIFF
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -159,7 +159,11 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
 	 */
 	__wt_spin_lock(session, &dhandle->close_lock);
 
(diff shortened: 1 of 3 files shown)


keithbostic:
lgtm, merged.

PR #2537, opened 2016-02-29
WT-2435: __wt_evict_file_exclusive_on/off cleanups

keithbostic:

This was the simplest approach I saw for reference counting requests for eviction disable/enable on a file. A few comments:

* I moved the `WT_CACHE.evict_walk_lock` to `WT_BTREE.evict_lock` in the btree handle, so now there's a cache `evict_lock` and a per-handle `evict_lock`. The reason is because reference counting eviction disable/enable requests requires we hold the no-eviction lock a lot longer than before, and I didn't want file close to stall eviction. We could avoid this part of the change by doing a try-lock call on the global lock, but this seemed better to me.

* I left the `WT_BTREE_NO_EVICTION` flag in place. If you like this change, we might want to take a run at removing that flag in favor of just a reference count (instead of initially setting the flag, we'd increment and never decrement the reference count). Where that gets tricky is we explicitly set `WT_BTREE_NO_EVICTION` in a few places and we'd need be sure there's no way the reference count could go wrong. See `__wt_btree_evictable` for one of the places I'm thinking of.

* I left the `__evict_request_walk_clear` code alone. If there's a per-btree eviction lock, we could use that instead of the current signalling mechanism, that is, the eviction server could try to acquire the per-btree eviction lock instead of checking the `WT_SESSION_CLEAR_EVICT_WALK` flag, and the exclusive code could then clear the current eviction point if it's holding that lock. Not sure if that's a bad idea or not.

* A few miscellaneous fixes: 8ed209b, 5a1f23d, 1257c17, I'd like to cherry-pick these out if you don't like the overall direction of this change.

Anyway, if you don't like this approach feel free to discard it, I was looking at Michael's comment in BF-1841 and wanted to see how deep the rabbit hole went.

### DIFF
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -164,6 +164,7 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 
 	/* Destroy locks. */
 	WT_TRET(__wt_rwlock_destroy(session, &btree->ovfl_lock));
+	__wt_spin_destroy(session, &btree->evict_lock);
 	__wt_spin_destroy(session, &btree->flush_lock);
 
 	/* Free allocated memory. */
@@ -350,7 +351,8 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
 	/* Initialize locks. */
 	WT_RET(__wt_rwlock_alloc(
 	    session, &btree->ovfl_lock, "btree overflow lock"));
-	WT_RET(__wt_spin_init(session, &btree->flush_lock, "btree flush lock"));
+	WT_RET(__wt_spin_init(session, &btree->evict_lock, "btree evict"));
+	WT_RET(__wt_spin_init(session, &btree->flush_lock, "btree flush"));
 
 	btree->checkpointing = WT_CKPT_OFF;		/* Not checkpointing */
 	btree->modified = 0;				/* Clean */
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -147,6 +147,7 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 
 	btree = S2BT(session);
 
+	/* Close the block manager. */
 	if ((bm = btree->bm) != NULL) {
 		/* Unload the checkpoint, unless it's a special command. */
 		if (!F_ISSET(btree,
@@ -162,10 +163,7 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 	/* Close the Huffman tree. */
 	__wt_btree_huffman_close(session);
 
-	/* Destroy locks. */
 	WT_TRET(__wt_rwlock_destroy(session, &btree->ovfl_lock));
-	__wt_spin_destroy(session, &btree->evict_lock);
-	__wt_spin_destroy(session, &btree->flush_lock);
 
 	/* Free allocated memory. */
 	__wt_free(session, btree->key_format);
@@ -181,6 +179,17 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 	btree->kencryptor = NULL;
 
 	btree->bulk_load_ok = false;
+
+	/* Reset eviction information. */
+	__wt_spin_destroy(session, &btree->evict_lock);
+	btree->evict_ref = NULL;
+	btree->evict_walk_period = 0;
+	btree->evict_walk_skips = 0;
+	btree->evict_disabled = 0;
+	btree->evict_busy = 0;
+	F_CLR(btree, WT_BTREE_NO_EVICTION);
+
+	__wt_spin_destroy(session, &btree->flush_lock);
 
 	F_CLR(btree, WT_BTREE_SPECIAL_FLAGS);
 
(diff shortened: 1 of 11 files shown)


michaelcahill:
@keithbostic, can we just get rid of `WT_BTREE_NO_EVICTION` and check `btree->evict_disabled != 0` instead?  I think it would be simpler to use `WT_BTREE_IN_MEMORY` if we want to keep a tree in cache long-term (i.e., in `__wt_btree_evictable`), and to always take hazard pointers so we don't have that extra complexity.

src/btree/bt_handle.c
```diff
@@ -181,6 +180,17 @@ __wt_btree_close(WT_SESSION_IMPL *session)
...
 	btree->bulk_load_ok = false;
 
+	/* Reset eviction information. */
+	__wt_spin_destroy(session, &btree->evict_lock);
+	btree->evict_ref = NULL;
+	btree->evict_walk_period = 0;
+	btree->evict_walk_skips = 0;
+	btree->evict_disabled = 0;
+	btree->evict_busy = 0;
+	F_CLR(btree, WT_BTREE_NO_EVICTION);
```
michaelcahill: This change defeats a lot of what currently makes interactions between sweep and eviction complicated, which is trying to allow ordinary eviction of trees after they are marked dead but before sweep has gotten around to discarding them.

That's okay with me: I don't know of any cases where sweep is a bottleneck, and if we find one we could explicitly wake sweep up after marking a tree dead.  I just think if we're going to make that change we should do it consciously and check the places in sweep and eviction that currently try to deal with "open-but-dead" trees.

src/btree/bt_handle.c
```diff
@@ -181,6 +180,17 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 
 	btree->bulk_load_ok = false;
 
+	/* Reset eviction information. */
+	__wt_spin_destroy(session, &btree->evict_lock);
```
michaelcahill: I'm confused about how this works: eviction can pick up a file that has been marked dead but still has a tree in cache.  With these changes, won't that try to acquire this spin lock and fail?

src/include/btree.h
```diff
@@ -129,10 +129,12 @@ struct __wt_btree {
 	uint64_t rec_max_txn;		/* Maximum txn seen (clean trees) */
 	uint64_t write_gen;		/* Write generation */
 
-	WT_REF  *evict_ref;		/* Eviction thread's location */
-	uint64_t evict_priority;	/* Relative priority of cached pages */
-	u_int    evict_walk_period;	/* Skip this many LRU walks */
-	u_int    evict_walk_skips;	/* Number of walks skipped */
+	WT_SPINLOCK evict_lock;		/* Eviction lock */
```
michaelcahill: Can we use `dhandle->close_lock

keithbostic:
@michaelcahill:

> can we just get rid of WT_BTREE_NO_EVICTION and check btree->evict_disabled != 0 instead? I think it would be simpler to use WT_BTREE_IN_MEMORY if we want to keep a tree in cache long-term (i.e., in __wt_btree_evictable), and to always take hazard pointers so we don't have that extra complexity.

That is the next step in this, I agree, but it's complicated: both lsm and btree set/clear `WT_BTREE_NO_EVICTION` on the handle as part of setting/clearing bulk-load configuration, and it wasn't clear to me they only do that a single time (so a simple reference count may not map directly to the set/clear of the flag).

I don't think it makes things much simpler to always take hazard pointers, but it probably does make things simpler to always review trees in memory for eviction, and set `WT_BTREE_NO_EVICTION` in fewer cases.

Have you decided you want to proceed in this direction? If so, I'll see about getting rid of `WT_BTREE_NO_EVICTION`.

keithbostic:
@michaelcahill, there was a comment you made I can no longer find, not sure what's up with that.

You said:
```
> @@ -181,6 +180,17 @@ __wt_btree_close(WT_SESSION_IMPL *session)
>  
>  	btree->bulk_load_ok = false;
>  
> +	/* Reset eviction information. */
> +	__wt_spin_destroy(session, &btree->evict_lock);
> +	btree->evict_ref = NULL;
> +	btree->evict_walk_period = 0;
> +	btree->evict_walk_skips = 0;
> +	btree->evict_disabled = 0;
> +	btree->evict_busy = 0;
> +	F_CLR(btree, WT_BTREE_NO_EVICTION);

This change defeats a lot of what currently makes interactions between sweep and eviction
complicated, which is trying to allow ordinary eviction of trees after they are marked dead
but before sweep has gotten around to discarding them.

That's okay with me: I don't know of any cases where sweep is a bottleneck, and if we find one
we could explicitly wake sweep up after marking a tree dead. I just think if we're going to make
that change we should do it consciously and check the places in sweep and eviction that
currently try to deal with "open-but-dead" trees.
```
I think you're correct, and I will back this change out.

Here's where I get confused: we have "btree close", and then we do additional work using the btree structure, which I don't like much.

src/btree/bt_handle.c
keithbostic: Yes, this change has to go.

src/include/btree.h
keithbostic: Let me take a look at that, I don't disagree.

michaelcahill:
Thanks, @keithbostic, I am still not sure about the new `btree->evict_lock` but the structural change here is a definite improvement.  I'll merge and we'll watch tests over the next few days.

PR #2538, opened 2016-03-01
WT-2366 Extend wtperf to support changing record sizes in update ops

agorrod:
Via a new update_delta setting in the thread group which takes either
a signed number of "rand" as an argument. Defaults to 0 (disabled).

There is also a new value_sz_max configuration option to limit how
large values can grow.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -266,10 +266,9 @@ config_threads(CONFIG *cfg, const char *config, size_t len)
 				if ((workp->truncate = v.val) != 1)
 					goto err;
 				/* There can only be one Truncate thread. */
-				if (cfg->has_truncate != 0) {
+				if (F_ISSET(cfg, CFG_TRUNCATE))
 					goto err;
-				}
-				cfg->has_truncate = 1;
+				F_SET(cfg, CFG_TRUNCATE);
 				continue;
 			}
 			if (STRING_MATCH("truncate_pct", k.str, k.len)) {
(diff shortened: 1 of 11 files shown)


agorrod:
@sueloverso Would you please review this change? I believe it's close to what we talked about in the previous pull request.

sueloverso:
I tried with a shrinking delta and it doesn't work with that.  I think it needs a value_sz_min for that too and two more configs: a shrink config and a mixed grow/shrink config.  I'll try to get some of that in there today.

agorrod:
> I tried with a shrinking delta and it doesn't work with that.

It worked in my testing, if I setup a smaller config:

```
conn_config="cache_size=2GB,checkpoint=(wait=60)"
table_config="type=file,leaf_page_max=32k,leaf_value_max=128k,split_pct=90"
icount=200000
report_interval=5
run_time=240
populate_threads=1
# Continue inserting new records.
threads=((count=4,update=1,update_delta=-100))
value_sz=20
```

Then check the content of the table at the end I see that all records are down to 1 character (which is how it's currently implemented.

> two more configs: a shrink config and a mixed grow/shrink config

I'm not averse to switching from allowing a negative delta to having an explicit shrink config. Adding both shrink and grow configs is a way to keep a stable volume of data. My hesitation is in the API complexity more than the implementation complexity; that is I'm hesitant to add so many configuration knobs here. I was thinking of allowing a `rand_mixed` and `rand_neg` option to the `update_delta` setting, but figured that could happen later when we see a case where we'd like to test that functionality.

> I think it needs a value_sz_min for that too

I can see the utility in that setting, but I was waiting for use cases before implementing it.

sueloverso:
Hmm, it didn't seem to work when I tried it.  I added a min and a shrinking job.  I have some ideas for mixed runs but didn't get to creating and testing those yet.  Take a look at the commits I made and see what you think.

sueloverso:
Also sorry I was not clear.  I did not mean that I wanted to add grow/shrink config options to wtperf itself, I meant adding in a shrink configuration job and mixed configuration jobs.  Not changing the wtperf API.

agorrod:
Thanks @sueloverso they are great improvements. Sorry my previous comment overlapped with work you were already doing. I'm happy for this to be merged now if you are.

sueloverso:
@agorrod I'll merge after I commit the new jobs I want to add for mixed deltas.  Thanks for the review.

PR #2540, opened 2016-03-01
WT-2437 Adjust expected error messages to support Windows.

sueloverso:
Look for WT-specific portion of the message or detect Posix/Windows.
@michaelcahill Here is a change to fix the test failures on Windows.

I modified the tests to look for non-Posix portions of error messages.  It does bring up again whether we need a translation layer.

### DIFF
--- a/test/suite/test_backup05.py
+++ b/test/suite/test_backup05.py
@@ -44,14 +44,6 @@ class test_backup05(wttest.WiredTigerTestCase, suite_subprocess):
     create_params = 'key_format=i,value_format=i'
     freq = 5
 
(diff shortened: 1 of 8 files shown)


agorrod:
Thanks @sueloverso I've switched the Jenkins Windows build job to this branch (http://build.wiredtiger.com:8080/job/wiredtiger-basic-windows-tester/), once it passes I'd be happy to merge.

Long term we probably need a read-only test artifact solution that spans Windows and POSIX systems, but I'm inclined to keep that change separate.

agorrod:
Thanks @sueloverso The Jenkins tests have passed so I'll merge.

PR #2542, opened 2016-03-01
WT-2384 Fix cursor joins with compare=le or lt in first index

ddanderson:
For cursor joins, joining the first index with only compare=le/lt (without a gt/ge) requires us to create another cursor internally to get results in order.  Previously we weren't returning any results.

This fixes the given test case.

### DIFF
--- a/src/cursor/cur_join.c
+++ b/src/cursor/cur_join.c
@@ -8,6 +8,10 @@
 
 #include "wt_internal.h"
 
+static int
+__curjoin_insert_endpoint(WT_SESSION_IMPL *, WT_CURSOR_JOIN_ENTRY *, u_int,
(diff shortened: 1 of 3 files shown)


ddanderson:
@michaelcahill, can you review this fix?

michaelcahill:
@ddanderson, this lgtm.  To make sure I understand, we want to iterate through the first index in ascending order, so to satisfy a "le/lt" constraint we need another cursor that walks the first index from the beginning and stops at the specified point.  Is that right?

Regardless, I'll merge.

ddanderson:
@michaelcahill, that's exactly right.  The extra cursor is all about iterating in order.  Thanks.

PR #2543, opened 2016-03-02
Wt 2440 Fix POWER8 CRC32

markbenvenuto:
Per the PPC64 ABI, v20-v31 are non-volatile registers. See http://refspecs.linuxfoundation.org/ELF/ppc64/PPC-elf64abi.html#REG
for calling convention.

In experiments under GDB, we see that after crc32_vpmsum is called that
registers v20-v31 are modified. We discovered this issue after
integrating the crc32_vpmsum code into our C++ code base which targeted
POWER8 for code generation, see
https://jira.mongodb.org/browse/SERVER-22773 for repro.

### DIFF
--- a/src/support/power8/crc32.S
+++ b/src/support/power8/crc32.S
@@ -90,6 +90,31 @@ FUNC_START(__crc32_vpmsum)
 	std	r26,-48(r1)
 	std	r25,-56(r1)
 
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @markbenvenuto lgtm - I'll merge once it passes testing.

PR #2544, opened 2016-03-02
WT-2436 For cursor joins with lt/le, fix how Bloom filters are populated

ddanderson:
Fixes the submitted test case.


ddanderson:
@michaelcahill, can you please review?

michaelcahill:
@ddanderson, I had to stare a little to remember that `WT_CURJOIN_END_LE` sets `WT_CURJOIN_END_LT` and so is handled by the same check, but given that, this lgtm.  I'll merge.

PR #2546, opened 2016-03-02
WT-2322 Do not permit read-uncommitted with bloom filters

ddanderson:
Since it might be hard to predict the results of having read-uncommitted with bloom filters, we shouldn't permit it.  This pull request includes a couple of small fixes and a test case for read uncommitted with/without bloom filters.

### DIFF
--- a/src/cursor/cur_join.c
+++ b/src/cursor/cur_join.c
@@ -434,8 +434,7 @@ __curjoin_init_iter(WT_SESSION_IMPL *session, WT_CURSOR_JOIN *cjoin)
 		       if (session->txn.isolation == WT_ISO_READ_UNCOMMITTED)
 			       WT_RET_MSG(session, EINVAL,
 				   "join cursors with Bloom filters cannot be "
-				   "used with WT_ISO_READ_UNCOMMITTED "
-				   "isolation");
+				   "used with read-uncommitted isolation");
 			if (je->bloom == NULL) {
 				/*
 				 * Look for compatible filters to be shared,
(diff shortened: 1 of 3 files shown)


ddanderson:
@agorrod, is this something you can review?

agorrod:
Thanks @ddanderson I'll take a look today.

src/cursor/cur_join.c
```diff
@@ -431,6 +431,11 @@ __curjoin_init_iter(WT_SESSION_IMPL *session, WT_CURSOR_JOIN *cjoin)
 			F_SET(cjoin, WT_CURJOIN_SKIP_FIRST_LEFT);
 
 		if (F_ISSET(je, WT_CURJOIN_ENTRY_BLOOM)) {
+		       if (session->txn.isolation == WT_ISO_READ_UNCOMMITTED)
+			       WT_RET_MSG(session, EINVAL,
+				   "join cursors with Bloom filters cannot be "
+				   "used with WT_ISO_READ_UNCOMMITTED "
```
agorrod: The defined symbol is internal, it'd be better to use the configuration string as per our API doc, which is ``read-uncommitted`` - [see](http://source.wiredtiger.com/develop/struct_w_t___c_o_n_n_e_c_t_i_o_n.html#afc75c44ff4497627c59d9b6aaa64c9d8)

agorrod:
Thanks @ddanderson I have one comment about an error message, apart from that this lgtm.

PR #2547, opened 2016-03-03
WT-2381 Rewrite LSM metadata to fix dump / load.

michaelcahill:
Include all of the WT_SESSION::create config in the ordinary LSM metadata so it is merged correctly into the dump header.  Provide an upgrade path for LSM metadata in the old format.

** Backwards bracking change for LSM: ** once metadata is upgraded to the new format, LSM trees cannot be opened with older versions of WiredTiger.


michaelcahill:
@sueloverso, can you please take a look?  I couldn't see a less invasive way to make this work: LSM metadata is currently a mess.  That said, we haven't had a downgrade-breaking change for a long time, so I'm very open to discussing alternatives or ways to minimize the impact.

I've tested manually by creating an LSM tree with the old version, making sure it can be dumped with this version, then doing a `wt write` with this version, making sure it can still be dumped and that the metadata is as expected.  If you have any ideas about how to test that (e.g., as part of upgrade tests), let's talk.

agorrod:

PR #2550, opened 2016-03-04
WT-2318 Add implementation for auto adjusting condition variables.

agorrod:
Where the amount of time spent waiting varies depending on whether work is being done between timeouts.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -468,7 +468,7 @@ __log_file_server(void *arg)
 				locked = false;
 				__wt_spin_unlock(session, &log->log_sync_lock);
 			} else {
-				WT_ERR(__wt_cond_signal(
+				WT_ERR(__wt_cond_auto_signal(
 				    session, conn->log_wrlsn_cond));
 				/*
 				 * We do not want to wait potentially a second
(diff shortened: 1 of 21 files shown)


agorrod:
@sueloverso I've written up the wrapper functions we talked about this morning. Would you mind taking a look at switching some of the condition wait usage in the logging code to this new mechanism?

It might be interesting to track some statistics about how often different code paths are being taken as well.

src/conn/conn_log.c
```diff
@@ -793,8 +817,8 @@ __log_server(void *arg)
 		}
 
 		/* Wait until the next event. */
-		WT_ERR(__wt_cond_wait_signal(session, conn->log_cond,
-		    WT_MILLION / WT_FORCE_PER_SECOND, &signalled));
+		WT_ERR(__wt_cond_auto_wait_signal(session, conn->log_cond,
+		    did_work || signalled, &signalled));
```
agorrod: I don't think you need signalled in the condition - the auto conditionals clear the timer automatically when woken by a signal.

agorrod:
autobuild: retest this please

sueloverso:
@agorrod I wrote a test program to look for high % CPU usage.  The only problem is that the value is quite variable across different machines.  I set the max to 10.0 percent after some experiementation.  On my Mac it was 0.0, on my AWS box, 8.3 and on the pull-request-tester machine it was 3.0.  But 10.0 is lower than the current-develop value when I run it on my Mac (and as high as 42% on my AWS box!!).  So the test fails everywhere on develop and passes everywhere, for now, with this branch.  We can revisit if the turns out to be problematic.

agorrod:
Thanks @sueloverso I'm happy for this to be merged.

PR #2552, opened 2016-03-04
WT-2404 Add streaming pack and unpack APIs to the extension API.

agorrod:
Will allow us to deprecate the incomplete pack/unpack there now.

Not yet ready for review. This is a work in progress. Need to add tests, and figure out what to do with the existing pack/unpack API.


src/btree/row_srch.c
```diff
@@ -614,14 +620,7 @@ leaf_match:	cbt->compare = 0;
...
 
-err:	/*
-	 * Release the current page if the search started at the root. If the
-	 * search didn't start at the root we should never have gone looking
-	 * beyond the start page.
-	 */
-	WT_ASSERT(session, leaf == NULL || leaf == current);
-	if (leaf == NULL)
-		WT_TRET(__wt_page_release(session, current, 0));
+err:	WT_TRET(__wt_page_release(session, current, 0));
```
agorrod: @keithbostic I found this when testing out a custom collator. When the collator returned an error, I was ending up with a hazard pointer error (the page was being released twice). This change simplifies the logic around release - by ensuring that only one place references a page at once. The original change went in https://github.com/wiredtiger/wiredtiger/commit/f43b07032ce1cfa0e844a4365082b0e001e0a83d and in WT-2202

agorrod:
@michaelcahill Would you mind reviewing this branch please? I have tested against the pull request for WT-2375 and it works to simplify the unpacking code there.

src/btree/row_srch.c
keithbostic: @agorrod, I think we still need this code (although the test does change):
```
err:    /*
         * Release the current page if the search started at the root. If the
         * search didn't start at the root we should never have gone looking
         * beyond the start page.
         */
        if (leaf == NULL)
                WT_TRET(__wt_page_release(session, current, 0));
```
otherwise, in the case of searching just a leaf page, we'll release the leaf page on error, which we didn't use to do.

michaelcahill: @keithbostic, I stared at this with @agorrod and couldn't figure out why that test would be needed.

If `leaf == NULL`, then we've found `current` in an ordinary search and it should be released.

If `leaf != NULL`, then we'll jump to `leaf_only` and set `current = NULL`, making `__wt_page_release` a noop.

Is there some other path through here (e.g., via restart) that we need to think through?  That condition has always confused me...

keithbostic: @agorrod, @michaelcahill: I missed the fact that we went through leaf-only, clearing "current" -- I agree, the test is no longer needed. Sorry, Alex!

michaelcahill:

PR #2553, opened 2016-03-06
WT-2391: De-prioritize eviction from indexes

keithbostic:
@agorrod, @michaelcahill: I spent a bunch of time chasing what I thought was the problem in WT-2391, but it turned out to be a red herring. Anyway, along the way I accumulated some changes that I think are a minor improvement over the current page read-generation code.

I cleaned them up and created this branch.

If you like this better than the current code, great, if you don't, then just discard the branch. As I said, it's a minor set of cleanups at most, and if you don't like them that's fine with me.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -486,6 +486,15 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
 				WT_RET(__wt_cache_eviction_check(
 				    session, 1, NULL));
 			WT_RET(__page_read(session, ref));
+
+			/*
+			 * If configured to not trash the cache, leave the page
+			 * generation unset, we'll set it before returning to
+			 * the oldest read generation, so the page is forcibly
+			 * evicted as soon as possible. We don't do that set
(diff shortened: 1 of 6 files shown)


agorrod:
@keithbostic these changes look fine to me - and are a definite improvement. I don't see how they are related specifically to reducing the amount of eviction being done on indexes.

I can see how improving our oldest read generation tracking would improve some particular cases where indexes are evicted, but I was hoping WT-2391 could cover a broader discussion of how we can prioritise indexes in cache more generally. Once we understand which part of this change improves the workload in SERVER-22423, and verified that the Jenkins [perf](http://build.wiredtiger.com:8080/job/wiredtiger-perf/) jobs aren't any slower I'm happy for this branch to be merged. I'll also open a new Jira ticket where we can discuss more general solutions to reduce the likelihood of evicting indexes from cache.

agorrod:
@keithbostic I switched the Jenkins performance jobs over to this branch (and back), and didn't see any regressions.

keithbostic:
@agorrod, I went ahead and merged this branch, and agree we'll take the design into the new JIRA ticket.

PR #2556, opened 2016-03-07
WT-2451 Allow eviction of WiredTiger metadata.

michaelcahill:
Switch from making eviction impossible to merely unlikely.  This allows
metadata pages to split and for eviction to discard deleted data.
Without this change, the trivial test added in test_schema07.py hangs
because the metadata fills the cache.

### DIFF
--- a/src/meta/meta_table.c
+++ b/src/meta/meta_table.c
@@ -67,18 +67,16 @@ __wt_metadata_cursor_open(
 	btree = ((WT_CURSOR_BTREE *)(*cursorp))->btree;
 
 	/* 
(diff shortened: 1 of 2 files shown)


michaelcahill:

My only real concern with this change is that there is some path somewhere that could deadlock because it is holding the schema lock, goes to update the metadata and is forced to wait for eviction.  I believe that is taken care of by `__wt_session_can_wait`, but would appreciate your thoughts.

keithbostic:
@michaelcahill, I agree, this change looks safe to me.

I didn't merge it because I wasn't sure you were ready, but merging is fine with me.

PR #2558, opened 2016-03-07
WT-2418: test_rebalance failing with EBUSY

keithbostic:
Fix a test cut-and-paste error, we're testing WT_SESSION.rebalance here,
not WT_SESSION.drop.


agorrod:
Thanks @keithbostic lgtm (nice catch). I'll merge.

PR #2559, opened 2016-03-09
WT-2456 Update CRC32 Power8 Code

markbenvenuto:
Update CRC32 Power8 after the PR for the non-volatile registers was accepted. The new code is more efficient at saving and restoring registers.


keithbostic:
Thanks, @markbenvenuto, merged.

PR #2560, opened 2016-03-09
WT-2454: checkpoint_sync=false does not prevent flushes/sync to disk

keithbostic:
Ready for review/merge.


michaelcahill:

PR #2562, opened 2016-03-10
WT-2461 Adjust sweep timeout values.

sueloverso:
@keithbostic Can you please review this minor change to the timeout value `sweep01`?  It was failing on the PPC machine and this fixes it there.


sueloverso:
Never mind.  I just merged it.

PR #2565, opened 2016-03-11
WT-2465: Coverity 1352899: Dereference before null check

keithbostic:
@daveh86, for your review/consideration.


agorrod:

PR #2571, opened 2016-03-11
WT-2471: review WiredTiger "int" printf formats

keithbostic:


bench/wtperf/wtperf.c
```diff
@@ -1668,7 +1668,7 @@ execute_workload(CONFIG *cfg)
 	for (threads = cfg->workers, i = 0,
 	    workp = cfg->workload; i < cfg->workload_cnt; ++i, ++workp) {
 		lprintf(cfg, 0, 1,
-		    "Starting workload #%d: %" PRId64 " threads, inserts=%"
+		    "Starting workload #%u: %" PRId64 " threads, inserts=%"
```
sueloverso: Should the %d to %u really be to % PRIu32?  That would be more consistent and there is no need for `i` to be `u_int` in that function.  It certainly could be uint32_t.

src/block/block_open.c
```diff
@@ -369,7 +369,7 @@ __desc_read(WT_SESSION_IMPL *session, WT_BLOCK *block)
 		WT_ERR_MSG(session, WT_ERROR,
 		    "unsupported WiredTiger file version: this build only "
 		    "supports major/minor versions up to %d/%d, and the file "
```
sueloverso: Should those also be fixed?  Even if not necessary, it would be more consistent.

src/lsm/lsm_manager.c
```diff
@@ -433,8 +433,10 @@ __lsm_manager_run_server(WT_SESSION_IMPL *session)
 				    session, WT_LSM_WORK_BLOOM, 0, lsm_tree));
 				WT_ERR(__wt_verbose(session,
 				    WT_VERB_LSM_MANAGER,
-				    "MGR %s: queue %d mod %d nchunks %d"
-				    " flags 0x%x aggressive %d pushms %" PRIu64
+				    "MGR %s: queue %" PRIu32 " mod %d "
```
sueloverso: Why is `lsm_tree->modified` an int instead of a bool?  That seems to have been missed in that bool pass a while ago.

src/lsm/lsm_worker.c
```diff
@@ -20,7 +20,7 @@ int
 __wt_lsm_worker_start(WT_SESSION_IMPL *session, WT_LSM_WORKER_ARGS *args)
 {
 	WT_RET(__wt_verbose(session, WT_VERB_LSM_MANAGER,
-	    "Start LSM worker %d type 0x%x", args->id, args->type));
+	    "Start LSM worker %u type %#" PRIx32, args->id, args->type));
```
sueloverso: Again not sure why `wt_lsm_worker_args->id` is a `u_int` instead of a sized `uint32_t` or `uint64_t`.  I guess it is beyond the scope of this ticket, but should we revisit those too?  In for a penny in for a pound...

bench/wtperf/wtperf.c
keithbostic: Actually, `cfg->workload_cnt` is a `u_int`, so I think this change is correct, and `i ` should stay a `u_int`. (Or they both should change?)

src/block/block_open.c
keithbostic: I think this is correct as written, those are printing defined constants, WT_BLOCK_MAJOR_VERSION and WT_BLOCK_MINOR_VERSION.

src/lsm/lsm_manager.c
keithbostic: Agreed, I've pushed that change.

src/lsm/lsm_worker.c
keithbostic: It's probably safe, hard to imagine any number of slots that wouldn't fit into a u_int.

PR #2577, opened 2016-03-14
WT-2460 Fix a flags test in LAS code to use mask not is-set.

agorrod:
It was attempting to save which flags were set by checking F_ISSET (which returns a boolean). Should have been using WT_MASK.

This was leading to the WT_SESSION_NO_EVICTION flag being cleared unexpectedly and WT_SESSION_CAN_WAIT being set as well.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -205,7 +205,7 @@ __wt_las_cursor(
 	 * useful more than once.
 	 */
 	*session_flags =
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, please also take a look at this one.  I'm going to merge because I think it should fix the `WT_ROLLBACK` from checkpoint we've been seeing since enabling eviction of metadata.

@agorrod, I tried to find some way to get the compiler to detect this without any luck.  We're assigning a truth value to a `uint32_t`, but since that can't lose information, neither gcc or clang seem to have checks for it.

PR #2578, opened 2016-03-14
WT-2477 Checks for a 64-bit build without polluting config.h files.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'll merge once tests pass.

PR #2583, opened 2016-03-14
WT-2484: Coverity 1345809: unchecked return value

keithbostic:
Explicitly discard the sscanf return value, hopefully that will quiet
Coverity.


keithbostic:
autobuild: retest this please

PR #2587, opened 2016-03-15
WT-2457 Wait for LSM work units to drain when getting exclusive access.

agorrod:
Otherwise we can return EBUSY unexpectedly. Fixed several outstanding
bugs along the way.

This change isn't ready for review yet - it needs further testing and a round of cleanup.


agorrod:
@michaelcahill Would you mind reviewing this change please?

michaelcahill:
Thanks, @agorrod -- I made a couple of cosmetic changes, lgtm if my commit passes.

PR #2590, opened 2016-03-15
WT-2485 Adjust lsm_tree->merge_min when we read the metadata.

sueloverso:
@agorrod or @michaelcahill here is one potential fix and an added assertion.  This fix has run over 300 iterations on both CONFIGs.

### DIFF
--- a/src/lsm/lsm_meta.c
+++ b/src/lsm/lsm_meta.c
@@ -349,6 +349,13 @@ __lsm_meta_read_v1(
 	 * Ignore any other values: the metadata entry might have been
 	 * created by a future release, with unknown options.
 	 */
+	/*
+	 * If the default merge_min was not overridden, calculate it now.  We
+	 * do this here so that trees created before merge_min was added get a
+	 * sane value.
+	 */
+	if (lsm_tree->merge_min < 2)
+		lsm_tree->merge_min = WT_MAX(2, lsm_tree->merge_max / 2);
 
 err:	__wt_scr_free(session, &buf);
 	return (ret);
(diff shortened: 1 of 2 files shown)


src/lsm/lsm_meta.c
```diff
@@ -349,6 +349,13 @@ __lsm_meta_read_v1(
 	 * Ignore any other values: the metadata entry might have been
 	 * created by a future release, with unknown options.
 	 */
+	/*
+	 * If the default merge_min was not overridden, calculate it now.  We
+	 * do this here so that trees created before merge_min was added get a
```
agorrod: I think this change is correct, but I don't think the comment is accurate. The default merge_min is 0 - because we want the default to be half of merge_max (i.e: needing to do this calculation is expected). We could go in a couple of different directions here:
* Update the comment and leave the behavior as is.
* Change the merge_min default to be 7 (half the default merge_max).
* Ensure that a merge_min is added to the configuration in ``__wt_lsm_meta_write`` so it will always be present.

I'm inclined to go with the first option. The second option avoids this same error happening in the future, but leads to an API change.

sueloverso: The comment and code is copied directly from `lsm_meta_read_v0`.  It may be worth refactoring some of the code so we have that code once.

sueloverso:
@agorrod Please look at the diff now.  I refactored to make a single place where we adjust merge_min.

michaelcahill:
Thanks, @sueloverso -- @agorrod and I reviewed and this looks good to us.  I'll merge.

PR #2592, opened 2016-03-16
WT-2490: search_near() returns wrong key for column-store

keithbostic:

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -467,7 +467,6 @@ comparep
 compat
 concat
 cond
-condvar
 conf
 confchk
 config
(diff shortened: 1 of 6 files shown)


dist/s_string.ok
```diff
@@ -467,6 +467,7 @@ comparep
 compat
 concat
 cond
+condvar
```
sueloverso: This should be unnecessary.

sueloverso: I removed this string's usage in the manydbs test yesterday.

michaelcahill:
Thanks, @keithbostic -- this one looks good to me.  Avoiding that special case of `slot == UINT32_MAX` makes it easier to follow.  I'll merge.

PR #2594, opened 2016-03-16
WT-2494: review calls to __wt_free, plus minor bug in an error path.

keithbostic:


sueloverso:
Thanks @keithbostic.  LGTM.  I'll merge when checks complete.

PR #2595, opened 2016-03-16
WT-2481 Set memory_page_max after we read chunk_max.

sueloverso:
@agorrod Please review this fix for correctly setting memory_page_max in LSM.  Note that the performance issue won't be fixed until WT-2485 is fixed.


agorrod:

PR #2596, opened 2016-03-16
WT-2414 Avoid main table reads and extractions for the first index

ddanderson:
When iterating cursor joins, take advantage of the fact that we have the index key for the first index available.  We don't need to extract it again.


ddanderson:
@michaelcahill, can you review this please?

michaelcahill:
Thanks, @ddanderson, lgtm.  I'll merge.

PR #2598, opened 2016-03-16
WT-2426 Lock all handles for checkpoints in one pass.

michaelcahill:
In particular, gather all locks (including exclusive locks on old
checkpoints) under the schema / handle list lock so we never reenter
those locks once we have a handle exclusive.

This extends the duration that we hold exclusive locks on checkpoints
to include the write leaves phase.  If that turns out to be a problem,
we could further split out write leaves into a separate pass (that
releases all handles at the end).


agorrod:
@michaelcahill This one looks good to me. I've looked at it both as a diff and walked through the new code paths and both seem reasonable. Should I merge and get this into stress testing.

michaelcahill:
As discussed, let's take this one a little slowly and see whether we can get `develop` stable overnight.  We can decide tomorrow how much risk we're prepared to take on...

sueloverso:
I will restart running on AWS with this branch with the two CONFIGs from WT-2420

agorrod:
autobuild: retest this please

PR #2602, opened 2016-03-18
WT-2496 Fix a race between checkpoint and starting a hot backup.

michaelcahill:
In particular, a checkpoint could mark and lock an old checkpoint for
deletion before a hotbackup starts.  The hotbackup could then write
metadata referencing that old backup which the checkpoint goes on to
delete.

The fix is crude: don't allow a hotbackup to start while a checkpoint
is in progress.


agorrod:

PR #2603, opened 2016-03-18

keithbostic:
@agorrod, @markbenvenuto, @sueloverso, @michaelcahill:

Here's first cut at creating a pluggable OS layer, for your review/consideration.

I'm happy to talk about this or respond to review comments, whichever works best. There are a lot of moving parts, and I tried a bunch of different approaches, and this is the one I thought worked best.

Some notes:

* `WT_VERB_FILEOPS` was left broken. I don't feel strongly about it: I can fix it, remove it entirely, or limit it to tracking open/close calls only, whatever people want.

* The WiredTiger library no longer uses FILE * objects; file descriptor and stream functions now use the same open and close functions, and both are represented by a WT_FH handle. Where the code previously used stdout or stderr, I'm faking it with illegal pointer values. We could create real WT_FH handles for stdout/stderr, but I didn't see a compelling reason, happy to do that if people prefer it.

* The sync/async versions of `__wt_fsync` have been merged, there's now a boolean flag that set wait/no-wait on a flush.

* The transaction log's print-log function no longer supports specified output handles, output is always to stdout (which the wt utility redirects into another file on demand). @sueloverso, please flag if this is a problem.

* @michaelcahill and I talked about giving each `WT_FH` a set of methods (so, for example, `__wt_fprintf(session, fh, ...)` becomes `WT_FH.printf(session, ...)`. I didn't do that in this commit, because it's going to touch everything, I'd rather do it in a separate commit. Let me know if you like/dislike that change.

Remaining tasks:

* Windows hasn't been touched yet, there's a reasonable chunk of work there.

* We're still "writing" the 4KB description blocks from the block manager in the in-memory case, and I haven't written the code to detect/complain about writing more than trivial amounts when running in-memory.

* `wiredtiger_open` still isn't checking `WIREDTIGER_CONFIG` for in-memory and readonly options.


keithbostic:
@michaelcahill, it's actually tricky to avoid writing the initial description block into created files.

In `__create_file`, we create the file (including the description block), then write the metadata, then open the file/checkpoint, which reads the description block, setting things like the block allocation size.

How strongly do you feel about losing an allocation size chunk per created file?

keithbostic:
Review request: one place I'm worried I've messed up is in how the various flavors of sync are structured. I'd appreciate it if someone reviewed the sync paths to make sure they're unchanged.

src/os_posix/os_map.c
```diff
@@ -57,10 +59,16 @@ __wt_mmap_preload(WT_SESSION_IMPL *session, const void *p, size_t size)
 {
 #ifdef HAVE_POSIX_MADVISE
 	/* Linux requires the address be aligned to a 4KB boundary. */
-	WT_CONNECTION_IMPL *conn = S2C(session);
-	WT_BM *bm = S2BT(session)->bm;
+	WT_CONNECTION_IMPL *conn;
+	WT_BM *bm;
```
sueloverso: nit - alphabetize.

src/os_posix/os_open.c
```diff
@@ -30,180 +84,77 @@ __open_directory(WT_SESSION_IMPL *session, char *path, int *fd)
  */
 int
 __wt_open(WT_SESSION_IMPL *session,
-    const char *name, bool ok_create, bool exclusive, int dio_type, WT_FH **fhp)
+    const char *name, int dio_type, u_int flags, WT_FH **fhp)
```
sueloverso: I like that you got rid of multiple bool args and replaced it with flags.  I would've made it a uint32_t myself.

sueloverso: Actually, the FLD_* macros assumes flags are uint32_t.

src/os_posix/os_map.c
keithbostic: Thanks, fixed.

src/os_posix/os_open.c
keithbostic: @sueloverso, I don't see a reason to go either fixed-size or not-fixed-size, really -- what is your thinking here?

sueloverso: Consistency, pretty much.  If you look at all the structures and function definitions with `flags` in them, they're nearly all `uint32_t` so to see one not like the others will make me pause and wonder why it is special/inconsistent, in the future.

src/os_posix/os_inmemory.c
```diff
@@ -0,0 +1,463 @@
+/*-
```
sueloverso: The file seems misplaced to me.  It doesn't appear as though there is much in this file that is directly posix related and this file should be above the OS layer.  Maybe in support/?

keithbostic: It's providing OS-level functionality, though, for example, "does this file exist"?

I've tried to resist wholesale file movements until the Windows port is done, but we do need to do something.

There's going to be general OS/library functions, file-handle API functions, file-system API functions, and those 3 categories come in at least 3 flavors: Windows, POSIX, in-memory, and of course, some files are shared between categories.

Support is starting to be a little bit of a dumping ground (for example, the PowerPC checksum code is there, because the POSIX code was there, and then it was #ifdef'd for the x86).

So far I haven't thought of anything I like, better ideas welcomed.

michaelcahill:
As far as file naming goes, I think we should have a `fs` directory at the top level, so we have `fs/fs_inmemory.c`.

keithbostic:
@agorrod, @michaelcahill: I've been chasing a problem today that I need a little help on.

In short, if I change the OS layer to back read-only data objects with read-only file descriptors, LSM fails.

Why I think this matters is a little obscure.

First, we handle direct IO configuration based on "readonly", it's this code here:
```
/* Open the underlying file handle. */
WT_ERR(__wt_open(session, filename, false, false,
    readonly ? WT_FILE_TYPE_CHECKPOINT : WT_FILE_TYPE_DATA,
    &block->fh));
```
which eventually ends up in the open code:
```
#ifdef O_DIRECT
        if (dio_type && FLD_ISSET(conn->direct_io, dio_type)) {
                f |= O_DIRECT;
                direct_io = true;
        }
#endif
```
In other words, if `direct_io=(checkpoint)` is configured, and a data file is opened read-only, we'll configure `O_DIRECT` on the handle. If we then writing the file, that's not quite what that direct I/O
configuration is documented to mean.

Second, when we open a file read-only, we don't truncate it, which means if we open a file in read-only mode, then open for read/write, there's a truncate call done on an already open file. It's obviously not causing a problem at the moment, but I think it's potentially dangerous.

Third, when we open a file read-only, we potentially map into memory, and I don't like writing to file descriptors backed by mapped regions. Again, not a problem at the moment, but I think it's potentially dangerous.

Back to LSM.

Here's the stack when LSM calls block-manager with readonly set (it's read-only because we're opening a checkpoint):
```
(gdb) where
#0  __wt_block_open (session=0x80304f040, filename=0x8030465a5 "xxx-000002.lsm", 
    cfg=0x803046560, forced_salvage=false, readonly=true, allocsize=4096, 
    blockp=0x803047220) at src/block/block_open.c:167
#1  0x000000080085cf30 in __wt_block_manager_open (session=0x80304f040, 
    filename=0x8030465a5 "xxx-000002.lsm", cfg=0x803046560, 
    forced_salvage=false, readonly=true, allocsize=4096, bmp=0x803088700)
    at src/block/block_mgr.c:585
#2  0x0000000800883e46 in __wt_btree_open (session=0x80304f040, 
    op_cfg=0x7fffffffe350) at src/btree/bt_handle.c:71
#3  0x00000008008ddd5a in __wt_conn_btree_open (session=0x80304f040, 
    cfg=0x7fffffffe350, flags=0) at src/conn/conn_dhandle.c:323
#4  0x00000008009ad86f in __wt_session_get_btree (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", 
    checkpoint=0x8030465c0 "WiredTigerCheckpoint.1", cfg=0x7fffffffe350, flags=0)
    at src/session/session_dhandle.c:528
#5  0x00000008009ad6ea in __wt_session_get_btree (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", 
    checkpoint=0x8030465c0 "WiredTigerCheckpoint.1", cfg=0x7fffffffe350, flags=0)
    at src/session/session_dhandle.c:519
#6  0x00000008009ad1e2 in __wt_session_get_btree_ckpt (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", cfg=0x7fffffffe350, flags=0)
    at src/session/session_dhandle.c:339
#7  0x00000008008f8bf2 in __wt_curfile_open (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", owner=0x80307a380, 
    cfg=0x7fffffffe350, cursorp=0x8030dce70) at src/cursor/cur_file.c:552
#8  0x000000080099d68d in __session_open_cursor_int (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", owner=0x80307a380, other=0x0, 
    cfg=0x7fffffffe350, cursorp=0x8030dce70) at src/session/session_api.c:328
#9  0x000000080099d1f2 in __wt_open_cursor (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", owner=0x80307a380, 
    cfg=0x7fffffffe350, cursorp=0x8030dce70) at src/session/session_api.c:381
#10 0x0000000800938ac0 in __clsm_open_cursors (clsm=0x80307a380, update=true, 
    start_chunk=0, start_id=0) at src/lsm/lsm_cursor.c:601
#11 0x000000080093d782 in __clsm_enter (clsm=0x80307a380, reset=false, 
    update=true) at src/lsm/lsm_cursor.c:257
#12 0x000000080093c0b0 in __clsm_insert (cursor=0x80307a380)
    at src/lsm/lsm_cursor.c:1372
#13 0x00000008008f3c04 in __curdump_insert (cursor=0x803029480)
    at src/cursor/cur_dump.c:313
#14 0x00000000004076ac in insert (cursor=0x803029480, name=0x803041070 "lsm:xxx")
    at src/utilities/util_load.c:544
#15 0x00000000004064d7 in load_dump (session=0x80304f040)
    at src/utilities/util_load.c:144
#16 0x0000000000406252 in util_load (session=0x80304f040, argc=0, 
    argv=0x7fffffffe920) at src/utilities/util_load.c:84
#17 0x000000000040a7bc in main (argc=3, argv=0x7fffffffe908)
    at src/utilities/util_main.c:228
(gdb) frame 2
#2  0x0000000800883e46 in __wt_btree_open (session=0x80304f040, 
    op_cfg=0x7fffffffe350) at src/btree/bt_handle.c:71
71		WT_ERR(__wt_block_manager_open(session, filename, dhandle->cfg,
(gdb) p dhandle->checkpoint
$7 = 0x803046580 "WiredTigerCheckpoint.1"
```
Subsequently, a checkpoint happens, and the file gets opened again, but this time it's not in readonly mode, and we end up trying to truncate the file. Here's the stack for that open:
```
(gdb) where
#0  __wt_block_open (session=0x80304f040, filename=0x803046045 "xxx-000002.lsm", 
    cfg=0x8030464c0, forced_salvage=false, readonly=false, allocsize=4096, 
    blockp=0x803047860) at src/block/block_open.c:169
#1  0x000000080085cf30 in __wt_block_manager_open (session=0x80304f040, 
    filename=0x803046045 "xxx-000002.lsm", cfg=0x8030464c0, 
    forced_salvage=false, readonly=false, allocsize=4096, bmp=0x803088900)
    at src/block/block_mgr.c:585
#2  0x0000000800883e46 in __wt_btree_open (session=0x80304f040, 
    op_cfg=0x7fffffffe530) at src/btree/bt_handle.c:71
#3  0x00000008008ddd5a in __wt_conn_btree_open (session=0x80304f040, 
    cfg=0x7fffffffe530, flags=0) at src/conn/conn_dhandle.c:323
#4  0x00000008009ad86f in __wt_session_get_btree (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", checkpoint=0x0, cfg=0x7fffffffe530, 
    flags=0) at src/session/session_dhandle.c:528
#5  0x00000008009ad1e2 in __wt_session_get_btree_ckpt (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", cfg=0x7fffffffe530, flags=0)
    at src/session/session_dhandle.c:339
#6  0x000000080099c3d0 in __wt_schema_worker (session=0x80304f040, 
    uri=0x8030463c0 "file:xxx-000002.lsm", 
    file_func=0x8009c01b0 <__wt_checkpoint_get_handles>, name_func=0, 
    cfg=0x7fffffffe530, open_flags=0) at src/schema/schema_worker.c:58
#7  0x000000080094f0b4 in __wt_lsm_tree_worker (session=0x80304f040, 
    uri=0x803007c00 "lsm:xxx", 
    file_func=0x8009c01b0 <__wt_checkpoint_get_handles>, name_func=0, 
    cfg=0x7fffffffe530, open_flags=0) at src/lsm/lsm_tree.c:1350
#8  0x000000080099c6d4 in __wt_schema_worker (session=0x80304f040, 
    uri=0x803007c00 "lsm:xxx", 
    file_func=0x8009c01b0 <__wt_checkpoint_get_handles>, name_func=0, 
    cfg=0x7fffffffe530, open_flags=0) at src/schema/schema_worker.c:81
#9  0x00000008009c4a73 in __checkpoint_apply_all (session=0x80304f040, 
    cfg=0x7fffffffe530, op=0x8009c01b0 <__wt_checkpoint_get_handles>, fullp=0x0)
    at src/txn/txn_ckpt.c:134
#10 0x00000008009c22d1 in __txn_checkpoint (session=0x80304f040, 
    cfg=0x7fffffffe530) at src/txn/txn_ckpt.c:394
#11 0x00000008009c1235 in __wt_txn_checkpoint (session=0x80304f040, 
    cfg=0x7fffffffe530) at src/txn/txn_ckpt.c:667
#12 0x00000008009a9823 in __session_checkpoint (wt_session=0x80304f040, 
    config=0x8030d1240 "target=(\"lsm:xxx\")") at src/session/session_api.c:1441
#13 0x000000000040b242 in util_flush (session=0x80304f040, 
    uri=0x803041070 "lsm:xxx") at src/utilities/util_misc.c:144
#14 0x000000000040655b in load_dump (session=0x80304f040)
    at src/utilities/util_load.c:157
#15 0x0000000000406252 in util_load (session=0x80304f040, argc=0, 
    argv=0x7fffffffe920) at src/utilities/util_load.c:84
#16 0x000000000040a7bc in main (argc=3, argv=0x7fffffffe908)
    at src/utilities/util_main.c:228
```
If I assert we're using a read/write file descriptor when we're about to truncate (which will cause the truncate to fail):
```
#6  0x00000008008642e6 in __wt_ftruncate (session=0x80304e040, fh=0x803041400,
    len=106496) at misc.i:321
#7  0x00000008008641f5 in __wt_block_truncate (session=0x80304e040,
    fh=0x803041400, len=106496) at src/block/block_write.c:30
#8  0x00000008008552fa in __wt_block_checkpoint_load (session=0x80304e040,
    block=0x80309f800,
    addr=0x8032d1560 "\001\227\201?Rj?\206\230\201??:??\200\200\200\200\200\200?
\001\177??\001_?", addr_size=29, root_addr=0x7fffffffddf0 "\227\201?Rj?\206",
    root_addr_sizep=0x7fffffffdef8, checkpoint=false)
    at src/block/block_ckpt.c:146
#9  0x000000080085d50d in __bm_checkpoint_load (bm=0x803046780,
    session=0x80304e040,
    addr=0x8032d1560 "\001\227\201?Rj?\206\230\201??:??\200\200\200\200\200\200?
\001\177??\001_?", addr_size=29, root_addr=0x7fffffffddf0 "\227\201?Rj?\206",
    root_addr_sizep=0x7fffffffdef8, checkpoint=false)
    at src/block/block_mgr.c:97
#10 0x0000000800883ee2 in __wt_btree_open (session=0x80304e040,
    op_cfg=0x7fffffffe590) at src/btree/bt_handle.c:100
#11 0x00000008008ddd5a in __wt_conn_btree_open (session=0x80304e040,
    cfg=0x7fffffffe590, flags=0) at src/conn/conn_dhandle.c:323
#12 0x00000008009ad86f in __wt_session_get_btree (session=0x80304e040,
    uri=0x8030453c0 "file:test_dump-000002.lsm", checkpoint=0x0,
    cfg=0x7fffffffe590, flags=0) at src/session/session_dhandle.c:528
#13 0x00000008009ad1e2 in __wt_session_get_btree_ckpt (session=0x80304e040,
    uri=0x8030453c0 "file:test_dump-000002.lsm", cfg=0x7fffffffe590, flags=0)
    at src/session/session_dhandle.c:339
#14 0x000000080099c3d0 in __wt_schema_worker (session=0x80304e040,
    uri=0x8030453c0 "file:test_dump-000002.lsm",
    file_func=0x8009c01b0 <__wt_checkpoint_get_handles>, name_func=0,
    cfg=0x7fffffffe590, open_flags=0) at src/schema/schema_worker.c:58
#15 0x000000080094f0b4 in __wt_lsm_tree_worker (session=0x80304e040,
    uri=0x803007c00 "lsm:test_dump",
    file_func=0x8009c01b0 <__wt_checkpoint_get_handles>, name_func=0,
    cfg=0x7fffffffe590, open_flags=0) at src/lsm/lsm_tree.c:1350
#16 0x000000080099c6d4 in __wt_schema_worker (session=0x80304e040,
    uri=0x803007c00 "lsm:test_dump",
    file_func=0x8009c01b0 <__wt_checkpoint_get_handles>, name_func=0,
    cfg=0x7fffffffe590, open_flags=0) at src/schema/schema_worker.c:81
#17 0x00000008009c4a73 in __checkpoint_apply_all (session=0x80304e040,
    cfg=0x7fffffffe590, op=0x8009c01b0 <__wt_checkpoint_get_handles>,
    fullp=0x0) at src/txn/txn_ckpt.c:134
#18 0x00000008009c22d1 in __txn_checkpoint (session=0x80304e040,
    cfg=0x7fffffffe590) at src/txn/txn_ckpt.c:394
#19 0x00000008009c1235 in __wt_txn_checkpoint (session=0x80304e040,
    cfg=0x7fffffffe590) at src/txn/txn_ckpt.c:667
#20 0x00000008009a9823 in __session_checkpoint (wt_session=0x80304e040,
    config=0x803252080 "target=(\"lsm:test_dump\")")
    at src/session/session_api.c:1441
#21 0x000000000040b242 in util_flush (session=0x80304e040,
    uri=0x803018210 "lsm:test_dump") at src/utilities/util_misc.c:144
#22 0x000000000040655b in load_dump (session=0x80304e040)
    at src/utilities/util_load.c:157
#23 0x0000000000406252 in util_load (session=0x80304e040, argc=0,
    argv=0x7fffffffe988) at src/utilities/util_load.c:84
#24 0x000000000040a7bc in main (argc=3, argv=0x7fffffffe970)
    at src/utilities/util_main.c:228
(gdb) frame 6
#6  0x00000008008642e6 in __wt_ftruncate (session=0x80304e040, fh=0x803041400,
    len=106496) at misc.i:321
321             WT_ASSERT(session, !fh->ro);
(gdb) p fh->ro
$4 = true
```
Can one of you please help me understand what's going on here?

keithbostic:
@sueloverso, would you please review a167c84?

I thought this change would improve things more than it did, but I eventually decided I liked it enough to keep it. Please let me know if you dislike moving in the direction of more general file types, and I'll revert it.

keithbostic:
@sueloverso, in `__log_has_hole`, we retrieve the file size from `WT_FH.size`. The callers of that function call `__wt_filesize()` to set a local variable with the file size -- is there any reason we couldn't pass that local variable into `__log_has_hole`?

Something like this:
```
diff --git a/src/log/log.c b/src/log/log.c
index ab8e2a4..1132b54 100644
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1204,18 +1204,18 @@ __wt_log_close(WT_SESSION_IMPL *session)
  *     file is zeroes.
  */
 static int
-__log_has_hole(WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t offset, bool *hole)
+__log_has_hole(WT_SESSION_IMPL *session,
+    WT_FH *fh, wt_off_t log_size, wt_off_t offset, bool *hole)
 {
        WT_CONNECTION_IMPL *conn;
        WT_DECL_RET;
        WT_LOG *log;
-       wt_off_t log_size, off, remainder;
+       wt_off_t off, remainder;
        size_t bufsz, rdlen;
        char *buf, *zerobuf;
 
        conn = S2C(session);
        log = conn->log;
-       log_size = fh->size;
        remainder = log_size - offset;
        *hole = false;
 
@@ -1559,7 +1559,8 @@ advance:
                                 * See if there is anything non-zero at the
                                 * end of this log file.
                                 */
-                               WT_ERR(__log_has_hole(session, log_fh,
+                               WT_ERR(__log_has_hole(
+                                   session, log_fh, log_size,
                                    rd_lsn.l.offset, &partial_record));
                        /*
                         * If we read the last record, go to the next file.
@@ -1623,7 +1624,7 @@ advance:
                 */
                if (reclen == 0) {
                        WT_ERR(__log_has_hole(
-                           session, log_fh, rd_lsn.l.offset, &eol));
+                           session, log_fh, log_size, rd_lsn.l.offset, &eol));
                        if (eol)
                                /* Found a hole. This LSN is the end. */
                                break;
```

keithbostic:
@sueloverso, I pushed a change to `log.c` in 87f40ed, passing the log size to the `__log_has_hole` function.

Here's the link: https://github.com/wiredtiger/wiredtiger/commit/87f40ed3a1bcbb163c9a3c2b7f6189ad3a6a493e#diff-764f5432bdeaa6ffb0296d56d50ca6ce

It looks safe to me, but it needs review. If I've gotten this change wrong, I'd like to fix it by calling `__wt_filesize` as necessary (although that may be a problem given it's a system call).

sueloverso:
@keithbostic The change to pass in `log_size` to `log_has_hole` LGTM.

keithbostic:
@michaelcahill, @agorrod, @sueloverso: I'm getting down to a short-list of issues on this one.

Here's my list:

- File renaming. I haven't thought of anything clever, and I agree with Michael we should have a top-level `fs` directory. So, I propose we create `fs_posix` and `fs_win` to match the `os_posix` and `os_win` directories, and move the file system code there. (There are about 15 files so we could do something like `fs/posix_dir.c` and `fs/win_dir.c` and categorize the files inside of a top-level `fs` directory, but I suspect that plan will become annoying eventually.)

- Windows testing: basic tests run on my spawn host, but more complicated ones fail. The errors indicate shell failure, and there's no core (??). I'm suspicious there's something systemic going on, I'll talk to Alex or David about this tonight.

- LSM opens "read-only" handles, then writes them (if this is real, can move to a new ticket).

- I haven't made the memory-map functions pluggable (move to a new ticket).

- There's no API to allow applications to set the file system methods (move to a "pluggable filesystems" ticket).

- There are still a couple fields in the WT_FH structure shared between the block-manager and the OS code (ignore for now).

- The Btree page discard code calls the OS code to unmap memory chunks (ignore for now).

sueloverso:
> File renaming. I haven't thought of anything clever, and I agree with Michael we should have a top-level fs directory.
> So, I propose we create fs_posix and fs_win to match the os_posix and os_win directories, and move the file system code there.
> (There are about 15 files so we could do something like fs/posix_dir.c and fs/win_dir.c and categorize the files inside of a top-level fs directory, but I suspect that plan will become annoying eventually.)

This appears to be the only real issue remaining.  It seems that something in this is going to feel out of place since we have three file system "types", two of which are backed with OS-specific calls and one that is not.  Keith proposes we have something like `fs/fs_inmemory.c, fs/fs_posix.c, fs/fs_win.c`.  But then the file system OS calls are out of place with all other OS calls.  I'd prefer all OS-specific code remain in the `os_posix` and `os_win` directories like it is today.  I'd propose we do have a `fs/fs_inmemory.c` file that would be out of place with actual OS versions.  Rename `os_posix/os_posix.c` to `os_posix/os_fs.c` (and `os_win/os_fs.c`) to make a clear connection it is the file system related calls.  `os_posix/os_open.c` could become `fs/fs_handle.c` or `fs/fs_open.c` since it does not contain any OS code.

keithbostic:
> Keith proposes we have something like fs/fs_inmemory.c, fs/fs_posix.c, fs/fs_win.c.

Of the two possibilities I noted, that's the one I like least.

My preference is to either leave things alone (with some file name changes, for example, `os_posix/os_open.c` renamed to `os_posix/fs_handle.c`), or to create `fs_posix` and `fs_win`, where `fs_posix` has the default FS implementation, and `fs_win` selectively replaces the files where Windows has its own implementation.

sueloverso:
Here's a summary of where the main file naming issues are:
Two files, `os_posix/os_inmemory.c` and `os_posix/os_open.c` do not contain any actual OS-specific code.  `os_inmemory.c` contains the underlying in-memory implementation for in-memory files.  The OS equivalents of that for actual OS-specific files are `os_posix/os_posix.c` and `os_win/os_win.c`.  `os_open.c` contains higher level `fh` handling, searching the `conn` lists, etc and entry points for `wt_open` and `wt_close`.  Where should the first two files (`os_inmemory.c and os_open.c`) reside and what should they be called?  Should we rename `os_posix.c` and `os_win.c` based on the answer to the first question.

src/block/block_map.c
```diff
@@ -58,12 +50,15 @@ __wt_block_map(
 
 	/*
 	 * Map the file into memory.
-	 * Ignore errors, we'll read the file through the cache if map fails.
+	 * Ignore not-supported errors, we'll read the file through the cache
+	 * if map fails.
 	 */
-	(void)__wt_mmap(session, block->fh, mapp, maplenp, mappingcookie);
+	ret = block->fh->fh_map(session, block->fh, mapp, maplenp);
+	if (ret == ENOTSUP)
```
agorrod: I don't believe checking ENOTSUP is safe across platforms. We don't translate error codes from OS calls to be POSIX compliant.

michaelcahill:
I opened WT-2518 to track the LSM checkpoint / truncate issue.

On the arrangement of source files, after discussing with @keithbostic, he thinks of `os_posix` as the "default", whereas @sueloverso and I think of it as one instance of an OS layer.  I would prefer `src/os_common` with common files (i.e., everything from os_posix that appears in `build_win/filelist.win`).

I think `os_posix/os_open.c` should be renamed `os_common/os_fhandle.c`.

We can revisit whether creating a top-level `src/fs` makes sense as part PM-480.

keithbostic:
@sueloverso, would you please briefly review 6b72f0a?

Thanks!

sueloverso:
@keithbostic The setvbuf change LGTM.  Thanks.

src/block/block_map.c
keithbostic: @agorrod 
> I don't believe checking ENOTSUP is safe across platforms. We don't translate error codes from OS calls to be POSIX compliant.

In these calls, I'm checking for specific returns from the WiredTiger Windows code, not OS calls, for example, the Windows map-preload code looks like this:
```
/*
 * __wt_win_map_preload --
 *      Cause a section of a memory map to be faulted in.
 */
int
__wt_win_map_preload(
    WT_SESSION_IMPL *session, WT_FH *fh, const void *p, size_t size)
{
        WT_UNUSED(session);
        WT_UNUSED(fh);
        WT_UNUSED(p);
        WT_UNUSED(size);

        return (ENOTSUP);
}
```
I think that's safe, unless the POSIX and Windows error namespace overlaps and a different error from a Windows call could look like `ENOTSUP`.

keithbostic:
@michaelcahill, @sueloverso: I did the rename, here's the list to make it easy to review:
```
os_posix/os_abort.c -> os_common/os_abort.c
os_posix/os_alloc.c -> os_common/os_alloc.c
os_posix/os_getline.c -> os_common/os_getline.c
os_posix/os_getopt.c -> os_common/os_getopt.c
os_posix/os_init.c -> os_common/os_init.c
os_posix/os_inmemory.c -> os_common/os_fs_inmemory.c
os_posix/os_open.c -> os_common/os_fhandle.c
os_posix/os_posix.c -> os_posix/os_fs_posix.c
os_posix/os_setvbuf.c -> os_common/os_setvbuf.c
os_posix/os_stdio.c -> os_common/os_fs_stdio.c
os_posix/os_strtouq.c -> os_common/os_strtouq.c
os_win/os_win.c -> os_win/os_fs_win.c
support/filename.c -> os_common/filename.c
```
Let me know if I missed anything.

sueloverso:
The rename looks good to me.  I was going to suggest that we didn't need the suffix on `os_posix/os_fs_posix.c` (and Windows too) and just make it `os_posix/os_fs.c` since it is clearly already posix.  But I don't mind given the other `os_fs_*` files.  It is consistent with that.

keithbostic:
@sueloverso 
>  I was going to suggest that we didn't need the suffix on os_posix/os_fs_posix.c (and Windows too) and just make it os_posix/os_fs.c since it is clearly already posix.

I think you're right, fixed.

keithbostic:
@michaelcahill, @markbenvenuto, @sueloverso, @agorrod:

I just pushed a pretty scary change into the WT-2330 branch, it's 3dff36f.

I explained what's going on in the commit message, but in short, I'm scared that a hot database backup on Windows can cause WiredTiger to fail to open one of the files being copied (for example, `WiredTiger.turtle`).

I don't know in detail how hot backups work in WiredTiger, so I'm looking for comments as to whether or not this is a real problem, and/or ideas for solutions.

src/os_win/os_errno.c
```diff
@@ -63,14 +63,33 @@ int
 __wt_errno(void)
 {
 	/*
+	 * Check for 0:
+	 * It's easy to introduce a problem by calling the wrong error function,
+	 * for example, this function when the MSVC function set the C runtime
+	 * error value. Handle gracefully and always return an error.
+	 */
+	return (errno == 0 ? WT_ERROR : errno);
```
agorrod: @keithbostic This change to Windows error handling bothers me. The semantic for Windows calls is to check for failure via checking the return, and then call ``GetLastError`` once or check ``errno``.

I can see cases where our code doesn't adhere to that policy (e.g: truncate), but I think we should fix the code rather than resort to returning a generic WT_ERROR. Did you make this change to solve a particular problem?

keithbostic: @agorrod 
> This change to Windows error handling bothers me.

I think you're missing part of this change. In the Windows code, after a POSIX or ANSI C error, we call `__wt_errno()`, after a Windows error we call `__wt_win32_errno()`, so it looks like this:

```
if (filehandle_secondary == INVALID_HANDLE_VALUE)
    WT_RET_MSG(session,
        __wt_win32_errno(), "%s: handle-open: CreateFileA: secondary", name);

if (fclose(fh->fp) != 0)  
    WT_RET_MSG(session,
        __wt_errno(), "%s: handle-close: fclose", fh->name);
```

There's an additional change, to always return an error value, and that's because it's easy to make a mistake and call the wrong error function (the Windows documentation isn't clear in some cases, too).

If the wrong error function gets called, we could return "success" out of the middle of a function when we should have returned an error. And, of course, these are error paths, so we're less likely to test them before a release, and if a call fails in the field we could be surprised.

By always returning an error code out of those functions, we ensure correct behavior in the caller.

agorrod: > There's an additional change, to always return an error value

I wasn't clear - that is the change I was concerned with. Maybe we could add the assertions back in to both error handlers (and clear the errno in ``__wt_errno``), so we can detect if something has gone wrong in diagnostic builds.

keithbostic: I wouldn't object to adding the assertion back in HAVE_DIAGNOSTIC builds, if you want to do that.

I wouldn't try to clear `errno`; enough systems play games with the declaration of `errno` I doubt it's useful (or will even necessarily work).

src/block/block_open.c
```diff
@@ -194,41 +196,23 @@ __wt_block_open(WT_SESSION_IMPL *session,
 	/* Configuration: optional OS buffer cache maximum size. */
 	WT_ERR(__wt_config_gets(session, cfg, "os_cache_max", &cval));
 	block->os_cache_max = (size_t)cval.val;
-#ifdef HAVE_POSIX_FADVISE
```
agorrod: Why did you decide to drop this error checking? If we are going to start silently ignoring this, we should probably update our documentation to reflect that this is an advisory setting.

agorrod:
@keithbostic I have two minor questions here. Neither of them are critical, so I'm inclined to merge the branch and create new JIRA tickets if those questions lead to further work.

src/conn/conn_open.c
```diff
@@ -182,6 +167,13 @@ __wt_connection_close(WT_CONNECTION_IMPL *conn)
 		WT_TRET(__wt_dlclose(session, dlh));
 	}
 
+	/* Close the lock file, opening up the database to other connections. */
+	if (conn->lock_fh != NULL)
```
keithbostic: @agorrod
> Are you sure it's safe to unlock the connection before we have finished closing? I could believe that it's safe after all other handles are closed and we ensure they are never opened again.

I think it's safe.

Previously, we:
1. Walked any remaining open files and closed them (skipping the lock file)
2. Closed some other subsystems (cache, txn, extensions)
3. Cleaned up the sessions
4. Closed the lock file

Now, we:
1. Close some other subsystems (cache, txn, extensions)
2. Close the lock file
3. Walk any remaining open files and close them
4. Clean up the sessions.

So, I'm closing the lock file earlier, but after everything has been shut down, I think that's a neutral change.

I'm closing "any remaining open files" later, after everything has been shut down. I think that's a good change, I'd rather close all the subsystems first, before checking for left-open handles.

src/block/block_open.c
keithbostic: @agorrod

> Why did you decide to drop this error checking? If we are going to start silently ignoring this, we should probably update our documentation to reflect that this is an advisory setting.
 
The change was in 9240bb3. At the time, I wrote:

> Remove block manager handle open tests of cache-max and dirty-cache-max
configuration conflicts with specifying direct I/O, lacking posix_fadvise
and lacking sync_file_range. We can't know in the block manager what the
underlying layer can do, and also, the old code was wrong, direct I/O is
set on a file type basis, so testing against a specific handle open isn't
correct.

In short, with pluggable file systems, we don't know until we see the ENOTSUP whether or not any particular call is going to work.

We could pretty easily output an message the first time one of these calls fails with ENOTSUP, if that's better.

PR #2604, opened 2016-03-19

keithbostic:
@sueloverso, @michaelcahill, for your review/consideration.

The idea here is to check the initial inputs we have (the wiredtiger_open config string and the WIREDTIGER_CONFIG environment variable), for either read-only or in-memory configurations, before looking at any of the on-disk information.

Also, there's this code:
```
        /* Ignore the base_config file if config_base_set is false. */
        if (config_base_set || F_ISSET(conn, WT_CONN_READONLY))
                WT_ERR(
                    __conn_config_file(session, WT_BASECONFIG, false, cfg, i1));
```

I didn't touch it, but it looks wrong to me -- why are we ignoring the base configuration file if we're configured read-only?

### DIFF
--- a/build_win/filelist.win
+++ b/build_win/filelist.win
@@ -109,6 +109,7 @@ src/os_posix/os_init.c
 src/os_posix/os_inmemory.c
 src/os_posix/os_mtx_rw.c
 src/os_posix/os_posix.c
+src/os_posix/os_stdio.c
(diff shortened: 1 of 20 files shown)


keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

PR #2606, opened 2016-03-21
WT-2501 LSM drop can violate access assumptions.

agorrod:
Add a test case demonstrating the problem.


agorrod:
Thanks for updating the branch @michaelcahill. The new test case is passing now, so I'm going to merge.

PR #2607, opened 2016-03-21
WT-2502 Free local allocation.

sueloverso:
@michaelcahill Please review.


agorrod:
Thanks @sueloverso - I pushed a tiny change: our free clears the pointer already so we don't need to do it again. lgtm - I'll merge.

PR #2608, opened 2016-03-22
WT-2506 Reorder to fix uninitialized variable warning.

michaelcahill:


agorrod:

PR #2609, opened 2016-03-22
WT-2503 Rename "wait" to avoid shadowing system call.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'll merge

PR #2612, opened 2016-03-27

keithbostic:

I was reviewing OS-specific code in the library, and noticed the wt backup code had some Windows specific code to flush the copied file to disk. However, I don't think that's sufficient, on Linux, at least, we have to flush the enclosing directory to disk as well.

I think the simplest solution is to create a WiredTiger function to do the copy safely, WiredTiger already knows how to do the magic.


sueloverso:
Thanks @keithbostic.  LGTM.

PR #2613, opened 2016-03-28
WT-2453 Add multiple eviction queues.

sueloverso:
Not yet ready for merge.


sueloverso:
@michaelcahill This is now ready to review.

src/evict/evict_lru.c
```diff
@@ -993,14 +1015,21 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
...
+			evict_queue->evict_candidates = candidates;
 		}
 	}
 
-	cache->evict_current = cache->evict_queue;
-	__wt_spin_unlock(session, &cache->evict_lock);
+	__wt_spin_unlock(session, &evict_queue->evict_lock);
+	/*
+	 * Now we can set the next queue.
+	 */
```
agorrod: I think it would be interesting to add some statistics here. One that tracks that we switched the queue. Possibly also information about the state of the queues (e.g: if the one being switched out is empty) - though there is a tradeoff between accuracy and doing more work while holding the ``evict_queue_lock``.

sueloverso: We do already have that, in `evict_walk` where the server thread notes whether the existing queue is empty when it starts filling the other one.  That stat is done at the beginning, without holding any locks.  I will move that code to here.  It will only add a few lines to the lock code (which is still much, much less than the code in `evict_get_ref`).

agorrod:
Thanks @sueloverso lgtm - and the performance benefits are compelling. I'll go ahead and merge. Lets watch the Jenkins performance plots for fallout.

PR #2614, opened 2016-03-29
WT-2517: wtperf uses setvbuf in a way that isn't supported on Windows

keithbostic:
@agorrod, @sueloverso, for your review/consideration.

### DIFF
--- a/bench/wtperf/misc.c
+++ b/bench/wtperf/misc.c
@@ -54,7 +54,7 @@ setup_log_file(CONFIG *cfg)
 		return (ret);
 
 	/* Use line buffering for the log file. */
-	(void)setvbuf(cfg->logf, NULL, _IOLBF, 32);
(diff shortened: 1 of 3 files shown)


bench/wtperf/misc.c
```diff
@@ -54,7 +54,7 @@ setup_log_file(CONFIG *cfg)
 		return (ret);
 
 	/* Use line buffering for the log file. */
-	(void)setvbuf(cfg->logf, NULL, _IOLBF, 32);
+	(void)setvbuf(cfg->logf, NULL, _IOLBF, 1024);
```
sueloverso: According to the MSDN doc @agorrod noted in the ticket, only the sizes of 0 were a problem.  But having them all the same buffer size is goodl

sueloverso:
Thanks @keithbostic.  LGTM.  I'll merge when checks complete.

sueloverso:
FYI it appears bt_debug.c also has one.

PR #2615, opened 2016-03-30
WT-2512 - Change how wtperf throttle values per thread are calculated

daveh86:


daveh86:
@agorrod , would you mind taking a look at this?

bench/wtperf/wtperf_throttle.c
```diff
@@ -104,10 +104,9 @@ worker_throttle(CONFIG_THREAD *thread)
 		 */
 		WT_RET(__wt_epoch(NULL, &throttle_cfg->last_increment));
 	} else {
-		throttle_cfg->ops_count =
-		    (uint64_t) (float)(usecs_delta /
-		    throttle_cfg->usecs_increment) *
-		    (float)throttle_cfg->ops_per_increment;
+		throttle_cfg->ops_count = usecs_delta *
+		    throttle_cfg->ops_per_increment /
```
agorrod: @daveh86 Would you mind adding a set of brackets around this expression so that it's clear which order the multiplication and division happens. I know, but I have to think about it.

agorrod:
Thanks @daveh86 lgtm - I'll merge

PR #2616, opened 2016-03-31
WT-2522: Incorrect format code in message

keithbostic:

### DIFF
--- a/src/os_win/os_fs.c
+++ b/src/os_win/os_fs.c
@@ -361,8 +361,8 @@ __win_handle_read(
 		if (!ReadFile(fh->filehandle, addr, chunk, &nr, &overlapped))
 			WT_RET_MSG(session,
 			    nr == 0 ? WT_ERROR : __wt_getlasterror(),
-			    "%s: handle-read: ReadFile: failed to read %"
-			    WT_SIZET_FMT " bytes at offset %" PRIuMAX,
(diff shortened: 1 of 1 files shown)


src/os_win/os_fs.c
```diff
@@ -361,8 +361,8 @@ __win_handle_read(
 		if (!ReadFile(fh->filehandle, addr, chunk, &nr, &overlapped))
 			WT_RET_MSG(session,
 			    nr == 0 ? WT_ERROR : __wt_getlasterror(),
-			    "%s: handle-read: ReadFile: failed to read %"
-			    WT_SIZET_FMT " bytes at offset %" PRIuMAX,
+			    "%s: handle-read: ReadFile: failed to read %lu "
```
agorrod: @keithbostic I think ``%lu`` expects a 64 bit value, whereas a ``DWORD`` is 32 bits - am I missing something?

keithbostic: @agorrod, I believe a `DWORD` is typedef'd as an `unsigned long` on Windows, and both are 4B.

https://msdn.microsoft.com/en-us/library/aa383751(VS.85).aspx

agorrod: D'oh. Thanks @keithbostic I misread the description in the ticket - I'd thought a capital "i" was an "l".

agorrod:
Thanks @keithbostic lgtm

PR #2620, opened 2016-04-01
WT-2330 Fix GCC 4.7 compiler warnings.

agorrod:


agorrod:
@daveh86 Would you mind taking a look at this change and merging it if it passes tests?

PR #2622, opened 2016-04-01
WT-2520: WT_SESSION::verify should not alter tables

keithbostic:


agorrod:
@keithbostic I'd like for this change to include some testing to ensure that the file is unchanged after verify, probably both to our python test suite and test/format.

keithbostic:
@agorrod, I took a quick run at this, but it's harder than it sounds.

I was thinking we could take a checksum of the file before and after the verify, but the file doesn't get written from cache to disk until the `WT_SESSION.verify` call forces it as part of opening a btree verify handle, so that's relatively difficult to do.

Any suggestions?

agorrod:
Thanks @keithbostic. This change lgtm.

PR #2623, opened 2016-04-01
WT-2518: LSM checkpoint handle acquisition optimization

keithbostic:


keithbostic:
@michaelcahill, sorry for the noise -- I pushed a change into this branch wasn't supposed to be there, so I reverted it. The only interesting change here is 7d5732c.

agorrod:
@keithbostic This change makes sense to me, but I'm not seeing how the described bug is triggered by the code being changed.

If the symptom were "opens a readonly handle read-write", I could understand this fix. I wouldn't think that any write would ever be done on the file (i.e: a new checkpoint should only be created in the file if there are changes since the last checkpoint).

keithbostic:
@agorrod, I've changed the title so it makes more sense.

We still want to merge this, even after merging WT-2526 (which gets rid of the notion of read-only handles), because it means less work during LSM checkpoints.

agorrod:
Thanks @keithbostic lgtm.

PR #2624, opened 2016-04-01
WT-2526: mixing and matching readonly and read/write handles

keithbostic:
Diagnostic code to detect mixing and matching readonly and read/write handles, currently fails the test suite.


keithbostic:
Discussed with @agorrod and @michaelcahilll; we're going to remove the read-only distinction in the WiredTiger OS layer, this branch is no longer needed.

PR #2625, opened 2016-04-01
WT-2527: OS X compile error, missing POSIX_FADV_WILLNEED #define

fedorova:
Not having this check was causing a compilation on OS X to fail.


wtbuild:

keithbostic:
Thanks, @fedorova, I'll merge a change in a few minutes.

PR #2627, opened 2016-04-02
WT-2450: salvage releases pages, then explicitly evicts them.

keithbostic:
@agorrod, ready for review.


agorrod:
@keithbostic Why did you decide to go with this option rather than setting ``WT_READ_WONT_NEED`` as described in the ticket? That would be consistent with ``compact``.

keithbostic:
To close the loop; we discussed this off-line, we're going to turn off eviction (instead of using `WT_READ_WONT_NEED`), it's a simpler solution.

agorrod:

PR #2628, opened 2016-04-04
WT-2529 Be less aggressive asserting in readonly connections

agorrod:
Since assertions call fsync on the standard I/O channel regardless of whether the connection is open read only.


agorrod:
@michaelcahill Could you review this for me please?

michaelcahill:

PR #2629, opened 2016-04-04
WT-2531: in-memory tables are wasting space in truncation

keithbostic:
@agorrod, for your review.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -135,8 +135,11 @@ __wt_block_checkpoint_load(WT_SESSION_IMPL *session, WT_BLOCK *block,
 	 * that was done when the checkpoint was first written (re-writing the
 	 * checkpoint might possibly make it relevant here, but it's unlikely
 	 * enough I don't bother).
+	 *
+	 * If in-memory, we don't read or write the object, and the truncate
+	 * will unnecessarily allocate buffer space.
 	 */
-	if (!checkpoint) {
+	if (!checkpoint && !F_ISSET(S2C(session), WT_CONN_IN_MEMORY)) {
 		/*
 		 * The truncate might fail if there's a file mapping (if there's
 		 * an open checkpoint on the file), that's OK.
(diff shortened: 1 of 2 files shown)


sueloverso:
@keithbostic Please include the new python unit test in the test suite.  Otherwise LGTM.

keithbostic:
@agorrod, here's what's going on.

We changed the block manager to not write a descriptor block for in-memory data objects in order to save memory (the descriptor block was a 32KB chunk of memory per "file" that we'd never use again), and that's likely the source of the current basic1.js test problem, because as far as the OS-layer knows, nothing has been written to the file so it returns that the file is zero-length.

While looking at this, I noticed we were allocating the in-memory buffer even though we were never writing it, and that turned out to be a bug in the in-memory truncation support, we weren't extending the file when a truncate is past the current end of file.

Commit 659c977 fixes the bug in truncate, and my guess is that will make basic1.js run again. But, that means wasting a big chunk of memory per file.

Commit af4e3f1 changes the block manager to not truncate in-memory files, and my guess is that will make basic1.js fail again.

I'm not sure what's the correct fix.

We could always return a non-zero size (4KB?) for an in-memory data object?

At a higher-level, maybe the secondaries shouldn't be looking at on-disk sizes to decide if a copy needs to be made? Or maybe they should copy any object that exists, regardless of the size?

keithbostic:
@agorrod, another idea:

We could make data objects special in the OS layer, and track the current file size for them without doing any memory allocation or reading/writing. That would mean slightly less special-case code in the block manager, and maybe a couple of other places, at the cost of  a few tests in the OS-layer in-memory code, and then the file size would always be non-zero (although the size wouldn't ever reflect reality, of course, since the upper-level btree code never reads/writes in-memory objects).

sueloverso:
I added some commentary in the ticket itself digging into the repl code a bit.

agorrod:
Thanks @keithbostic - this fix looks good to me. You are right that it doesn't solve the test failure from MongoDB - I'm going to push a change for that in another branch.

PR #2631, opened 2016-04-04
SERVER-23504: Coverity analysis defect 98177: Resource leak

keithbostic:
Don't leak the variable value if the call to __wt_fprintf fails.

@agorrod, for your review.


agorrod:

PR #2632, opened 2016-04-04
WT-2526: mixing and matching readonly and read/write handles

keithbostic:
@agorrod, here are the changes to default to opening file handles read/write (unless the database is configured readonly).

For your review.


michaelcahill:

PR #2633, opened 2016-04-05
WT-2533 Don't let in-memory tables return a zero size.

agorrod:
Returning a zero size breaks MongoDB replication. Return a non-zero
size for now until SERVER-23526 is resolved.


agorrod:
@michaelcahill Could you review this change for me please? It's a workaround for an issue in MongoDB - I've created SERVER-23526 to track that issue.

michaelcahill:
Thanks, @agorrod -- I hope we can fix this in MongoDB eventually (so that a storageSize of zero has no special meaning), but in the meantime this lgtm.

daveh86:
WFIW, I've run this through through testing and the MongDB replication test suite is now passing.

PR #2635, opened 2016-04-05
WT-2453 Add ability for eviction workers to populate eviction queues.

sueloverso:
Not for review/merge.  Just making a PR for easy viewing of the changes.  When I ran this with the workload it was half the performance of the wt-2453 branch.  I'm planning to profile to see where the time is spent.


fedorova:
@sueloverso: Sue, if the profiling numbers are not revealing anything interesting, let me know if you'd like me do the timing instrumentation and figure out where the time is being spent in these two branches.

sueloverso:
autobuild: retest this please

PR #2636, opened 2016-04-07
WT-2505: Review clang analyzer warnings

keithbostic:


agorrod:
Thanks @keithbostic lgtm.

PR #2639, opened 2016-04-08
WT-2540: Separate stream and file handle methods

keithbostic:
@agorrod, @michaelcahill, for your consideration.


michaelcahill:
Let's wait to see where WT-2539 ends up before deciding about this branch.

keithbostic:
@michaelcahill, did you consider merging this and branch #2632 (the pull request for WT-2526), before trying the work in WT-2539? I think that will give you a better/cleaner place from which to start and avoids some future merging.

michaelcahill:
OK, @keithbostic, I'll merge.

PR #2640, opened 2016-04-10
WT-2535: Extend test/format to test for transactions reading their writes

keithbostic:
Change test format to verify snapshot isolation operations when terminating the transaction.


keithbostic:
@agorrod, @michaelcahill: this is a change to track test/format operations done at snapshot isolation and then verifying them as part of committing or rolling back a transaction.

I've had this on my cleanup list for a long time. I added code to do basic value checking as part of debugging long-running transactions and the look-aside table, but I never got around to doing the more general change -- WT-2535 pushed me to get it done.

If you think this change is worthwhile (in addition to Alex's change in WT-2535, of course), I'd suggest reviewing the `ops()` function, that's where all the code that handles transaction commit and rollback lives. Also, I rewrote the pre-thread operations loop as part of these changes, it was rolling back transactions that it shouldn't have (for example, when records weren't found because they'd been deleted).

I've been running the code this weekend, and it seems to work, with one exception (see the next comment).

It would be interesting to know if this would have caught the PPC problem you folks chased last week?

keithbostic:
I'm not sure yet, but I think this is finding a bug in our fixed-length column store support.

Running the appended CONFIG errors in 30 seconds or less, with something like:
```
snap_check: 5443 search: expected {0x77}, found {0x17}
```

What makes me think this might be real:
* the error is returning the wrong value (it's not a 0-value row being confused with `WT_NOTFOUND`),
* there aren't any deletes (although there are inserts),
* and both reducing the number of threads or turning off checkpoints makes the problem go away.

I've chased this a bit today without any luck, but I'll take another run at it tomorrow.

Here's the CONFIG:
```
#  RUN PARAMETERS
abort=0
auto_throttle=1
backups=0
bitcnt=7
bloom=1
bloom_bit_count=47
bloom_hash_count=26
bloom_oldest=0
cache=51
checkpoints=1
checksum=uncompressed
chunk_size=2
compaction=0
compression=none
data_extend=0
data_source=table
delete_pct=0
dictionary=0
direct_io=0
encryption=none
evict_max=5
firstfit=0
huffman_key=0
huffman_value=1
in_memory=0
insert_pct=20
internal_key_truncation=1
internal_page_max=15
isolation=snapshot
key_gap=10
key_max=20
key_min=20
leaf_page_max=10
leak_memory=0
logging=1
logging_archive=1
logging_compression=none
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=4
merge_max=10
mmap=1
ops=1000000
prefix_compression=1
prefix_compression_min=7
quiet=0
repeat_data_pct=69
reverse=0
rows=10000
runs=1000
rebalance=1
salvage=1
split_pct=61
statistics=0
statistics_server=0
threads=20
timer=20
transaction-frequency=19
value_max=3321
value_min=15
verify=1
wiredtiger_config=
write_pct=12
```

agorrod:
Thanks @keithbostic the shape of this change looks reasonable to me. I'm going to hold off merging for now until we have a better understanding of the new test failure.

keithbostic:
@agorrod, @michaelcahill, with WT-2542 applied, this branch runs.

I think it's ready for review/merge, if you decide this is a reasonable test to include long-term.

It would interesting to know if it would have detected the PPC problem you were chasing, if that's an easy test to run.

agorrod:

PR #2645, opened 2016-04-11
WT-2537 Check for an old and new uninitialized LSN for recovery.

sueloverso:
@michaelcahill Please review this change for the reported bug when running recovery with a database that contained an old INIT_LSN in the metadata.


michaelcahill:

PR #2646, opened 2016-04-12
WT-2539 Make streams a separate handle type from files.

michaelcahill:
Initial commit, doesn't yet handle buffering.


michaelcahill:
@keithbostic, I should be able to finish this off tomorrow and get it ready for review.  If you take a look in the meantime, let me know what you think.  I prefer having a separate type for streaming, and it will make the public API for pluggable filesystems simpler.

michaelcahill:
@keithbostic, this is now ready for review.  Can you please take a look?

keithbostic:
@michaelcahill, this looks good to me, I'll merge. I didn't expect to like this (as you know), but I was wrong, this is quite a nice change!

PR #2647, opened 2016-04-12
WT-2542: fixed-length column store reconciliation overwrites original values

keithbostic:
@agorrod, @michaelcahill, this is the one I've been chasing in WT-2535.

The problem is fixed-length column-store  reconciliation was updating the original page any time a transaction was globally visible, and that gets snapshot isolation wrong because earlier readers in the system subsequently see the wrong value.

The only interesting commit is 93059b7, ready for review/merge.

### DIFF
--- a/src/btree/bt_debug.c
+++ b/src/btree/bt_debug.c
@@ -952,8 +952,7 @@ __debug_update(WT_DBG *ds, WT_UPDATE *upd, bool hexbyte)
 			__dmsg(ds, "\tvalue {deleted}\n");
 		else if (hexbyte) {
 			__dmsg(ds, "\t{");
-			__debug_hex_byte(ds,
-			    ((uint8_t *)WT_UPDATE_DATA(upd))[0]);
+			__debug_hex_byte(ds, *(uint8_t *)WT_UPDATE_DATA(upd));
 			__dmsg(ds, "}\n");
 		} else
 			__debug_item(ds,
--- a/src/btree/bt_ret.c
+++ b/src/btree/bt_ret.c
@@ -46,7 +46,7 @@ __wt_kv_return(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd)
 		}
 
 		/* Take the value from the original page. */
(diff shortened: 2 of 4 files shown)


keithbostic:
Here's the CONFIG.
```
#  RUN PARAMETERS
abort=0
auto_throttle=1
backups=0
bitcnt=7
bloom=1
bloom_bit_count=47
bloom_hash_count=26
bloom_oldest=0
cache=51
checkpoints=1
checksum=uncompressed
chunk_size=2
compaction=0
compression=none
data_extend=0
data_source=table
delete_pct=0
dictionary=0
direct_io=0
encryption=none
evict_max=5
firstfit=0
huffman_key=0
huffman_value=1
in_memory=0
insert_pct=20
internal_key_truncation=1
internal_page_max=15
isolation=snapshot
key_gap=10
key_max=20
key_min=20
leaf_page_max=10
leak_memory=0
logging=1
logging_archive=1
logging_compression=none
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=4
merge_max=10
mmap=1
ops=1000000
prefix_compression=1
prefix_compression_min=7
quiet=0
repeat_data_pct=69
reverse=0
rows=10000
runs=1000
rebalance=1
salvage=1
split_pct=61
statistics=0
statistics_server=0
threads=20
timer=20
transaction-frequency=19
value_max=3321
value_min=15
verify=1
wiredtiger_config=
write_pct=12
```

agorrod:

michaelcahill:
@keithbostic, to make sure I'm understanding this, I don't think the issue here was globally visible updates.  Checkpoints will write updates as of their transaction snapshot, and before this fix, that would overwrite the original page image.

If there was a globally visible update, readers should have seen it.  The problem here was when there was *no* globally visible update, readers would fall back to reading from the page image, which could be changed underneath them by a checkpoint.  Is that right or am I misunderstanding?

keithbostic:
@michaelcahill, you're correct. When I said "globally visible updates", I meant to refer to the updates overwriting the original page image, and those aren't necessarily globally visible.

PR #2649, opened 2016-04-13
SERVER-23622 Allow eviction to consider pages from files being checkpointed

agorrod:
Otherwise eviction will never queue clean pages from files while they
are being checkpointed. This causes problems in some MongoDB workloads.
In particular workloads where the oplog is generating most of the
work in a checkpoint, and is the major source of read pressure
(i.e: clean pages) in the cache.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1143,11 +1143,10 @@ retry:	while (slot < max_entries && ret == 0) {
 			continue;
 
 		/*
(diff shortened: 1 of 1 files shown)


michaelcahill:
@agorrod, where did this change end up -- have we seen any workloads benefit from it?

agorrod:
I liked it, but it doesn't help for the workload I was looking at. I'll close the branch for now.

PR #2648, opened 2016-04-12
WT-2453 Evict server decides whether or not to evict.

fedorova:
@sueloverso: take a look at this change. The evict server probabilistically decides whether or not to evict based on the evict queue empty / not empty statistics. This gives me up to 40% performance improvements relative to wt-2453 in my tests. I have not yet had a chance to investigate whether we can do any better: wanted to run this by you first. Do you see any performance regressions from this change in your tests?

PR #2650, opened 2016-04-13
WT-2546 Evict server decides whether or not to evict

michaelcahill:
See initial discussion in #2648.

### DIFF
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -2180,9 +2180,6 @@ main(int argc, char *argv[])
 	config_opts = NULL;
 	cc_buf = tc_buf = user_tconfig = NULL;
 
-	/* XXX mjc: kludge to always enable statistics. */
-	user_cconfig = dstrdup("statistics=(fast)");
-
 	/* Setup the default configuration values. */
 	cfg = &_cfg;
 	memset(cfg, 0, sizeof(*cfg));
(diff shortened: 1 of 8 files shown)


fedorova:
@michaelcahill, your changes look good to me and do not affect my numbers.

fedorova:
@sueloverso, could you please take a look at this change? Jenkins numbers are back up where they used to be, and this also fixes the regression I was seeing with 8 threads.

sueloverso:
FYI, I did switch Jenkins back to the develop branch.  I'll move it back now and we'll let the perf tests run overnight and tomorrow on this branch.

michaelcahill:
Sorry it took me a while to notice this, but don't the Jenkins jobs run with the default settings for statistics (i.e., off)?  If so, using statistics for the calculation isn't going to work...

I've changed the `wiredtiger-perf-evict` job to turn statistics on to see whether that makes a difference.  If so, I think we'll want a different long-term approach that doesn't rely on statistics.

sueloverso:
@michaelcahill that was a good catch!  The plots are looking good this morning.  I'll run on my machine and create a change to calculate without statistics.

sueloverso:
@fedorova  and @michaelcahill I believe this is ready for review and merging.  I removed Michael's temporary stat change in wtperf and made the percent threshold a macro.  Thanks Sasha for running all of your tests too and coming up with the computation that works well for all the thread/core values.  The plots are looking great now.

michaelcahill:
@sueloverso and @fedorova, this lgtm and the performance results so far look good.  I'll merge.

PR #2653, opened 2016-04-14
WT-2103: add incremental backup testing to format

keithbostic:
General work on backup testing in format, and specifically to add support for incremental backups.


keithbostic:
@michaelcahill, @agorrod, @sueloverso:

This branch mostly runs, but I've seen several failures that look something like this:
```
t: wt.wt: handle-read: pread: failed to read 7680 bytes at offset 32403456: WT_ERROR: non-specific WiredTiger error
```
What I believe is happening is a hot backup starts, then a checkpoint starts and checkpoints a file. As part of that checkpoint, the file is truncated and so the hot backup hits the end of the file before that's expected. (Imagine a workload that extends the file but then removes the data so the extended blocks are no longer necessary: as part of a checkpoint, the block manager checks if the last chunk of the file's "available" list is at the end of the file, in which case it discards that chunk and truncates the file.)

I suspect the same thing can happen if we "unload" the live checkpoint, the file will be truncated for similar reasons.

Here's one of the stacks, where we're truncating a file and the `WT_CONNECTION.hot_backup` flag is true.
```
#4  0x000000000055d3a7 in __wt_abort (session=0x80345a100)
    at src/os_common/os_abort.c:22
#5  0x00000000004a6cfc in __wt_assert (session=0x80345a100, error=0,
    file_name=0x5ac005 "src/block/block_write.c", line_number=19,
    fmt=0x5a6e59 "%s") at src/support/err.c:481
#6  0x000000000056e374 in __wt_block_truncate (session=0x80345a100,
    block=0x8034bc800, len=209746944) at src/block/block_write.c:19
#7  0x00000000005945cb in __wt_block_extlist_truncate (session=0x80345a100,
    block=0x8034bc800, el=0x8034bc9f0) at src/block/block_ext.c:1372
#8  0x000000000058e7b6 in __ckpt_process (session=0x80345a100,
    block=0x8034bc800, ckptbase=0x809399400) at src/block/block_ckpt.c:621
#9  0x000000000058dedd in __wt_block_checkpoint (session=0x80345a100,
    block=0x8034bc800, buf=0x80905a218, ckptbase=0x809399400, data_cksum=true)
    at src/block/block_ckpt.c:269
#10 0x000000000056a005 in __bm_checkpoint (bm=0x8034ae3c0,
    session=0x80345a100, buf=0x80905a218, ckptbase=0x809399400,
    data_cksum=true) at src/block/block_mgr.c:67
#11 0x00000000004d81c6 in __wt_bt_write (session=0x80345a100, buf=0x80905a218,
    addr=0x0, addr_sizep=0x0, checkpoint=true, compressed=false)
    at src/btree/bt_io.c:360
#12 0x0000000000476cb8 in __rec_write_wrapup (session=0x80345a100,
    r=0x80905a200, page=0x809042150) at src/reconcile/rec_write.c:5630
#13 0x00000000004718b8 in __wt_reconcile (session=0x80345a100,
    ref=0x7fffdd9e9818, salvage=0x0, flags=0) at src/reconcile/rec_write.c:421
#14 0x0000000000477a6b in __rec_root_write (session=0x80345a100,
    page=0x8034e1b00, flags=0) at src/reconcile/rec_write.c:675
#15 0x0000000000471f83 in __wt_reconcile (session=0x80345a100,
    ref=0x80349f6c8, salvage=0x0, flags=0) at src/reconcile/rec_write.c:463
#16 0x00000000004fa4ad in __sync_file (session=0x80345a100,
    syncop=WT_SYNC_CHECKPOINT) at src/btree/bt_sync.c:184
#17 0x00000000004f9f50 in __wt_cache_op (session=0x80345a100,
    op=WT_SYNC_CHECKPOINT) at src/btree/bt_sync.c:293
#18 0x00000000004b918a in __checkpoint_tree (session=0x80345a100,
    is_checkpoint=true, cfg=0x7fffdd9e9d50) at src/txn/txn_ckpt.c:1125
#19 0x00000000004bac74 in __checkpoint_tree_helper (session=0x80345a100,
    cfg=0x7fffdd9e9d50) at src/txn/txn_ckpt.c:1205
#20 0x00000000004ba8c6 in __checkpoint_apply (session=0x80345a100,
    cfg=0x7fffdd9e9d50, op=0x4bac50 <__checkpoint_tree_helper>)
    at src/txn/txn_ckpt.c:186
#21 0x00000000004b81dc in __txn_checkpoint (session=0x80345a100,
    cfg=0x7fffdd9e9d50) at src/txn/txn_ckpt.c:507
#22 0x00000000004b6be5 in __wt_txn_checkpoint (session=0x80345a100,
    cfg=0x7fffdd9e9d50) at src/txn/txn_ckpt.c:668
#23 0x000000000049fb43 in __session_checkpoint (wt_session=0x80345a100,
    config=0x0) at src/session/session_api.c:1441
```

We could ignore the error if the file unexpectedly ends during a hot backup copy, I don't think it affects consistency. But, that's hard for applications, especially when they're using system utilities to do the backup, and further, this doesn't fail all the time, so we're creating a situation where this could fail for the first time in the field.

We could completely turn off file truncation in the block manager whenever hot backup is running, but there are some "innocent" users of truncation, like salvage, and that makes me hesitate. So, what I think I'll try instead is having hot-backup turn off file truncation on the specific files the backup cursor is going to return to the application.

Thoughts?

keithbostic:
@sueloverso, another failure in this branch (I've now seen it twice). There's definitely some chance this is my mistake, I haven't dug into it yet.
```
[1460652988:395700][31935:0x80340e000], t, file:wt.wt, txn-recover: RUNDIR/BACKUP/WiredTigerLog.0000000003: handle-open: open: No such file or directory
[1460652988:395791][31935:0x80340e000], t, file:wt.wt, txn-recover: Recovery failed: No such file or directory
```
Here's the CONFIG from the last time:
```
data_source=table
isolation=snapshot
logging=1
timer=20
backups=1
ops=0
rows=10000
runs=1000
```

agorrod:
> We could completely turn off file truncation in the block manager whenever hot backup is running, but there are some "innocent" users of truncation, like salvage, and that makes me hesitate.

I don't mind either approach. I'd be inclined to say that running salvage and backup in parallel isn't "innocent". I think it'd be reasonable to disable truncate from any file while a backup is in progress.

keithbostic:
FTR, @michaelcahill wants to go with the simpler approach, turn off all file truncation if a backup cursor is open.

keithbostic:
@sueloverso, would you please review my wording changes in 8876677 and b182925?

keithbostic:
@sueloverso, I'm still seeing recovery errors from this branch, would you have time to take a look?

The failure happens reasonably quickly in my test runs, and after the failure, the `BACKUP_COPY` directory reproduces it.

The error looks something like this:
```
t, file:wt.wt, txn-recover: RUNDIR/BACKUP/WiredTigerLog.0000000003: handle-open: open: No such file or directory
t, file:wt.wt, txn-recover: Recovery failed: No such file or directory
```
Here's the CONFIG:
```
backups=1
data_source=table
logging=1
ops=0
rows=10000
runs=1000
timer=20
```
The error stack is:
```
#0  __wt_err (session=0x803057340, error=2, 
    fmt=0x8009dbcd9 "%s: handle-open: open") at src/support/err.c:319
#1  0x0000000800960a23 in __posix_handle_open (session=0x803057340, 
    fh=0x803046600, name=0x805e2c040 "X/WiredTigerLog.0000000003", 
    file_type=8, flags=0) at src/os_posix/os_fs.c:681
#2  0x000000080095b8d2 in __wt_open (session=0x803057340, 
    name=0x805aef000 "WiredTigerLog.0000000003", file_type=8, flags=0, 
    fhp=0x7fffffffe240) at src/os_common/os_fhandle.c:238
#3  0x000000080092cfdb in __log_openfile (session=0x803057340, 
    ok_create=false, fhp=0x7fffffffe240, 
    file_prefix=0x8009d7b70 "WiredTigerLog", id=3) at src/log/log.c:676
#4  0x000000080092f551 in __wt_log_scan (session=0x803057340, lsnp=0x0, 
    flags=9, func=0x8009cd240 <__txn_log_recover>, cookie=0x7fffffffe2e8)
    at src/log/log.c:1593
#5  0x00000008009cc9af in __wt_txn_recover (session=0x803057340)
    at src/txn/txn_recover.c:533
#6  0x00000008008e4319 in __wt_connection_workers (session=0x803057000, 
    cfg=0x7fffffffe410) at src/conn/conn_open.c:237
```

It's quite possible I messed up somewhere in the incremental backup code, but if so, I don't see it.

src/docs/backup.dox
```diff
@@ -98,29 +104,31 @@ and removing log files from the original database home:
 
 1. Perform a full backup of the database (as described above).
 
-2. Perform a full database checkpoint.
```
sueloverso: If the user does not do a checkpoint then the log files cannot be archived.  So opening the backup cursor on the log files will continue to accumulate more and more files and the user will be copying them multiple times.  That was the reason that step is in there.

keithbostic: @sueloverso, thanks, I've added a note to that effect.

sueloverso:
FTR, the ENOENT error above was because log archiving was turned on and checkpoint was running at "random" times.  So the BACKUP directories ended up with a non-contiguous set of log files.

keithbostic:
Thank you for the fix, @sueloverso.

I've pushed all of my changes, I think this is ready for your review & merge.

src/block/block_write.c
```diff
@@ -15,6 +15,23 @@
...
+	 * backup (for example, doing incremental backups, which only copies log
+	 * files, or targeted backups, stops all truncation). We may want a more
+	 * targeted solution at some point.
+	 */
+	if (S2C(session)->hot_backup)
+		return (EBUSY);
+
+	/*
+	 * Additionally, the truncate might fail if there's a file mapping (if
+	 * there's an open checkpoint on the file), return EBUSY in that case.
```
sueloverso: This comment is misleading.  We don't return EBUSY explicitly in this function after this comment.

keithbostic: I've pushed a fix.

sueloverso:
Thanks @keithbostic .  This LGTM now.  I see you committed a fix for my one comment.  I'm good for you to merge this once the PR tester finishes.

PR #2655, opened 2016-04-15
WT-2554 Add initial C test framework.

agorrod:
With a few sample applications. Need to do further work to cleanup
the same applications and share as much code/style as possible.


agorrod:
Not ready for review - just a placeholder for now.

sueloverso:
@agorrod I took a look at this today.  While the test util part is looking fine, one part I want to bring up now before getting too far down the road is the naming and structure of the actual tests.  There is no description in them describing what the underlying problem is/was and what they're testing for.  I would prefer to list the ticket number in a comment and have the program name describe the issue such as test_manyrecords.c or test_wt1965_manyrecords.c (just an example, I don't know what wt1965 was actually testing).  If it is just named for the ticket, I have to go find that and maybe dig through a lot of comments to figure out what it is testing.

agorrod:
@sueloverso I've renamed the tests - it is a good point.

@keithbostic Would you mind taking a look at the content in this pull request? I deliberately started with the existing test cases to ensure the harness captured useful abstractions. The result was that the initial test cases started out a long way from our code standards (I guess that's an apology for there being a bunch of inconsistencies in the three test cases). I'd like to chat to you about the direction this is heading in, and how we can get there expediently.

sueloverso:
@agorrod Thank you for the file renaming.  I think it is in the right direction but not far enough.  Since I'd expect these test cases to be fairly random in what they're testing, I'd like text/comment in the code (enforced by some part of s_all) describing the bug and briefly what the test is doing.  I'm not sure where the test code came from originally, but they definitely should conform to all the coding standards.

Many directories currently existing in `test` have evolved from tickets and test cases.  Do we retroactively move them to put them under this C test umbrella too?

agorrod:
@keithbostic I pulled in most changes from your branch here. I liked the getopt change, but ran into a linker error when I tried to make the change locally:

```
make[2]: Entering directory '/home/alexg/work/wiredtiger/build_posix/test/csuite'
  CCLD     test_wt1965_col_efficiency
../../test/utility/.libs/libtest_util.a(parse_opts.o): In function `testutil_parse_opts':
/home/alexg/work/wiredtiger/build_posix/test/utility/../../../test/utility/parse_opts.c:59: undefined reference to `__wt_optarg'
```

I have more to do here before I'm ready to ask for a review. I want to add comments to the test cases, and construct a ``dist/s_new_ctest``, that will generate a stub C test program given a test name.

agorrod:
@sueloverso I've pushed this closer to being ready for merge. Would you mind doing a full review. I'm happy for you to push changes directly into the branch.

test/utility/parse_opts.c
```diff
@@ -0,0 +1,133 @@
...
+	opts->verbose = false;
+
+	if ((opts->progname = strrchr(argv[0], '/')) == NULL)
+		opts->progname = argv[0];
+	else
+		++opts->progname;
+
+	while ((ch =
+	    __wt_getopt(opts->progname,
+		argc, argv, "A:h:n:o:pR:T:t:vW:")) != EOF)
```
sueloverso: There are a number of options used by other tests like `manydbs`, `readonly` and `recovery` that are not covered in this set.  We should try to change existing tests to use this infrastructure too.

test/utility/thread.c
```diff
@@ -0,0 +1,194 @@
...
+	opts->running = false;
+
+	return (NULL);
+}
+
+/*
+ * Repeatedly walk backwards through the records in a table.
+ */
+void *
+thread_prev(void *arg)
```
sueloverso: I wanted to add a `thread_next` function to go along with this.  But it doesn't appear we use this function anywhere.  So I didn't want to add another function that isn't called.  Maybe we should have a test that both repeatedly walks backwards and another thread repeatedly walks forward through a table.

test/utility/parse_opts.c
```diff
@@ -0,0 +1,133 @@
...
+			    "[-A append thread count] "
+			    "[-h home] "
+			    "[-n record count] "
+			    "[-o op count] "
+			    "[-p] "
+			    "[-R read thread count] "
+			    "[-T thread count] "
+			    "[-t c|f|r table type] "
+			    "[-v] "
+			    "[-W write thread count] ",
```
sueloverso: The difference between A,R,W and T is not clear to me.  What does T imply or do?

test/utility/test_util.h
```diff
@@ -0,0 +1,119 @@
...
+		TABLE_ROW=3	/* Row-store */
+	} table_type;
+	bool	 preserve;
+	bool	 verbose;
+	uint64_t     nrecords;
+	uint64_t     nops;
+	uint64_t     nthreads;
+	uint64_t     n_append_threads;
+	uint64_t     n_read_threads;
+	uint64_t     n_write_threads;
```
sueloverso: All of these should have comments.  Do we anticipate ever needing a thread indicator for creates, updates, deletes, truncates?  Having a specialized field for each, plus a general nthreads may be confusing later.

test/utility/thread.c
```diff
@@ -0,0 +1,194 @@
...
+
+	return (NULL);
+}
+
+/*
+ * Append to a table in a "racy" fashion - that is attempt to insert the
+ * same record another thread is likely to also be inserting.
+ */
+void *
+thread_insert_race(void *arg)
```
sueloverso: Is this too specialized for the utility directory?  It is only used in one test and feels like it should be part of that test (at least until we have a second use for it).

sueloverso:
@agorrod I have reviewed this and have a few questions but otherwise it is good to go for a start.  However anticipating changing existing tests in other directories makes me think that options/args are going to be an ongoing bit of a pain point and deciding what should be part of an individual test versus added to the general utility code will require thought.

test/utility/parse_opts.c
agorrod: This is a demonstration of different tests needing different options. The ``-T`` option is used by test cases that only have a single thread type that doesn't fit neatly into append/read/write single operation type thread.

sueloverso: That is going to be hard to think about far into the future.  I think you mentioned at the meeting, we really probably need a config language, similar to wtperf, instead of command-line args.  Something where we can add special need options going forward.

daveh86:
This is a test comment. Does this trigger the build?

agorrod:
Thanks @markbenvenuto I really appreciate your help.

agorrod:
@sueloverso Could you do a (hopefully) final round of review on this branch? I believe it's ready for initial commit now.

sueloverso:
autobuild: retest this please

PR #2656, opened 2016-04-15
WT-2223 WT-2553 in_memory documentation and stress testing

keithbostic:
WT-2223: in-memory stress testing, WT-2553: document in-memory configuration

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -722,8 +722,8 @@ wiredtiger_open = wiredtiger_open_common + [
         \c create option''',
         type='boolean'),
     Config('in_memory', 'false', r'''
-        keep data in-memory only, minimize disk I/O''',
-        type='boolean', undoc=True),
+        keep data in-memory only. See @ref in_memory for more information''',
+        type='boolean'),
     Config('use_environment', 'true', r'''
         use the \c WIREDTIGER_CONFIG and \c WIREDTIGER_HOME environment
         variables if the process is not running with special privileges.
--- a/dist/api_err.py
+++ b/dist/api_err.py
@@ -54,8 +54,10 @@ errors = [
     Error('WT_CACHE_FULL', -31807,
         'operation would overflow cache', '''
         This error is only generated when wiredtiger_open is configured
-        to run in-memory, and an insert or update operation requires more
-        than the configured cache size to complete.'''),
+        to run in-memory, and an insert or update operation requires
+        more than the configured cache size to complete. The operation
+        may be retried; if a transaction is in progress, it should be
+        rolled back and the operation retried in a new transaction.'''),
     Error('WT_PERM_DENIED', -31808,
         'permission denied (internal)', undoc=True),
 ]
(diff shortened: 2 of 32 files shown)


keithbostic:
@agorrod, @michaelcahill, for your consideration, with these changes test/format is happy to run in-memory.

Fair disclosure: revision de811ee is probably not OK with you. That commit changes `WT_CACHE_FULL` to not be fatal, and allows the transaction to proceed after that error, even attempting other insert/update calls (which may or may not fail). That feels right to me: it lets the user decide how to resolve the failure without losing their work, and there's no reason in our implementation the transaction cannot continue. (And, to be honest, I just hate the idea of any error other than `WT_ROLLBACK` meaning a transaction rollback is required.)

If you choose to require that `WT_CACHE_FULL` be fatal, let me know and I'll fix up the branch.

michaelcahill:
@keithbostic, as predicted, I don't think de811ee is okay -- for example, if we're updating a table with indices, we could get `WT_CACHE_FULL` part way through and there's no way for the application to know what was done and what wasn't.  If we had subtransactions, so each operation could do a partial rollback, we could allow a transaction to continue after `WT_CACHE_FULL`, but that's a much bigger change.

keithbostic:
@michaelcahill, @agorrod: of course you're right, indices hadn't crossed my mind.

OK, easy enough change -- I'll push a revised branch tomorrow unless someone beats me to it.

keithbostic:
@agorrod, Alex, I think this one is ready to be reviewed/merged, with this change, test/format runs in-memory. (One thing, we might want to tune it better so we don't create unreasonably small caches for in-memory runs.)

Up front, let me apologize for mixing-and-matching WT-2223 and WT-2553 in this branch; I kind of backed into it without thinking, and I'll make sure it never happens again.

agorrod:
@keithbostic To complete the loop here - we talked about changing how test/format does testing of in-memory so that it:
* Doesn't configure such a huge cache.
* Can be chosen randomly sometimes.
* Handles or avoids filling the cache during the bulk load phase+

+ It would be valuable to understand what happens when a WT_CACHE_FULL is returned during a bulk load - since MongoDB could potentially encounter that situation.

keithbostic:
@agorrod, here are the changes we discussed.

It was messier than anticipated because format was creating huge (80K byte) keys relatively frequently, and so in-memory configurations were blowing out the in-memory cache all the time. I think I've fixed that, but that change deserves some review, and is almost certainly going to ripple out into other workloads.

The list:

* We no longer configure a huge cache. We configure a cache of at least 2x the base "data size" required by the initial load, and if the load fails (which it occasionally does), we continue with the key/value pairs we were were able to load, and set the insert percentage low and the delete percentage higher so the run can actually get work done instead of just constantly failing updates/inserts.

* We randomly configure an in-memory run 5% of the time.

* We don't have to worry about bulk loads, in-memory configuration turns off bulk loads.

michaelcahill:
Thanks @keithbostic, this lgtm and I'm keen to see it in regular testing.  I'll merge.

PR #2659, opened 2016-04-17
WT-2558 WT_PAGE structure reorganization

keithbostic:
@agorrod, @michaelcahill, for your consideration, I thought it likely reducing the size of the `WT_PAGE` structure to 64B could help our malloc fragmentation. (The `WT_PAGE_MODIFY` structure remains at 104B.)

This one needs relatively careful review -- I'm happy to do that myself, this week, if you want the change.

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -86,10 +86,10 @@ __cursor_fix_next(WT_CURSOR_BTREE *cbt, bool newpage)
 
 	/* Initialize for each new page. */
 	if (newpage) {
-		cbt->last_standard_recno = __col_fix_last_recno(page);
+		cbt->last_standard_recno = __col_fix_last_recno(cbt->ref);
 		if (cbt->last_standard_recno == 0)
 			return (WT_NOTFOUND);
-		__cursor_set_recno(cbt, page->pg_fix_recno);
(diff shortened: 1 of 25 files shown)


agorrod:
@keithbostic Could you demonstrate the benefit in a workload? Particularly one with reasonable page sizes. I'd have thought that the size of the disk image copy we keep in the ``WT_PAGE`` structure would overwhelm the saving.

keithbostic:
@agorrod, in terms of size, I would expect the disk image copy to overwhelm the savings, no question. 

I talked with Martin this morning, and a 72B allocation will be done from an 80B bucket, so there's even less waste than I anticipated. Additionally, tcmalloc's allocation speed isn't affected by the size of the allocation (at least for small allocations).

Given that, I think the only remaining question is if it's a worthwhile cleanup to not maintain the page's record number in two places (we track the page's record number in both the `WT_REF` and the `WT_PAGE`, this change removes it from `WT_PAGE`). I like that change (it's better not to have two versions of the information that have to be kept in sync), but I won't fight for it.

Feel free to kill this branch without further discussion.

michaelcahill:
@keithbostic, thanks -- I'll merge.

FTR, I don't love having `pageref` and `ref` in the same function: I would have changed the latter to `child` or similar, but that's going to touch more code.

PR #2660, opened 2016-04-18
WT-2560 Use a rwlock to protect transaction state, don't spin.

michaelcahill:


michaelcahill:
The thing that makes this change more invasive than ideal is that the transaction calls couldn't fail previously (they returned `void`).  Now that they use ordinary rwlocks, they need to deal with the lock calls failing, which has knock-on effects on other code.

michaelcahill:
@agorrod, this looks pretty good in testing.  Can you please review?

michaelcahill:
@agorrod belay that, there were failures in stress testing overnight, leading to WT-2561.

michaelcahill:
This is currently waiting on WT-2566 (adding barriers to locks).

michaelcahill:
OK, @agorrod, I think this one is finally ready for review.  Can you please take a look?

agorrod:
Thanks @michaelcahill I like this change - I'll go ahead and merge.

PR #2661, opened 2016-04-18
WT-2548 Cap the amount of data handed to raw compression.

michaelcahill:
Previously, we had special case code for zlib to bound the compression
ratio to an arbitrary factor.  This is driven by synthetic workloads
where data can be compressed on disk to 1% or less of the in-memory
size.  If WiredTiger allows such on-disk pages to grow, they can end up
larger than the maximum page size permitted in memory and either
immediately force eviction or in extreme cases grow larger than the
total cache size.

This change bounds the amount of data handed to raw compression based on
the maximum page size in memory before attempting in-memory splits.

### DIFF
--- a/ext/compressors/zlib/zlib_compress.c
+++ b/ext/compressors/zlib/zlib_compress.c
@@ -307,17 +307,9 @@ zlib_compress_raw(WT_COMPRESSOR *compressor, WT_SESSION *session,
 		/*
 		 * If there's more compression to do, save a snapshot and keep
 		 * going, otherwise, use the current compression.
-		 *
-		 * Don't let the compression ratio become insanely good (which
-		 * can happen with synthetic workloads).  Once we hit a limit,
-		 * stop so the in-memory size of pages isn't hugely larger than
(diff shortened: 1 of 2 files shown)


michaelcahill:
@keithbostic, this one is all yours -- the ticket has context, but the summary is that lz4 raw compression was causing `test/format` to hang.

We had a hack in zlib, this replaces it with something more sane in reconciliation (rather than copying the same hack into the lz4 compressor).  In configurations with reasonable cache and page sizes, this will permit much larger compression ratios than before, but in the constrained cases `test/format` exercises it should avoid hangs.

michaelcahill:
btw, I'm not sure why we started to see this recently: it's possible that upgrading lz4 on test machines is leading to better compression ratios.  I saw 100x compression(!) with the data generated by the workload in the ticket.

keithbostic:
@michaelcahill, this makes sense to me, but isn't this still too large?

Wouldn't it make sense to use something like 80% (or even less) of `splitmempage`, so a couple of updates don't immediately push us past the boundary?

agorrod:

PR #2662, opened 2016-04-18
WT-2549 Fix for recno keys in cursor joins

ddanderson:
In a join cursor, a key that passes the conditions is in 'raw' format, and when the join cursor is not raw, it must be converted to 64bit in the recno case.

This occurs only for non-raw join cursors, which is why it wasn't detected by the test suite (Python is always in raw mode).  The case that fails is ex_schema, which does not produce any keys from the join when it fails.


ddanderson:
@michaelcahill, this is ready for review.

michaelcahill:
Thanks, @ddanderson, lgtm.  I'll merge.

PR #2664, opened 2016-04-18
WT-2559 Open a local log file handle for sync.

sueloverso:
The one in the log structure may change and point to a different log
file than the one we expect.  Open our own to the correct file.

@keithbostic Please review this change.


keithbostic:
lgtm, merged.

PR #2665, opened 2016-04-18
WT-2504 Remove READONLY conditional for base config.

sueloverso:
It was a remnant from early development.

@keithbostic Can you review this small change?


keithbostic:
lgtm, merged.

PR #2668, opened 2016-04-19
WT-2557 Need to truncate logs before closing backup cursor.

sueloverso:
@keithbostic Please review this change for the incremental log truncating log files.  The truncate/archive needs to be done before closing the backup cursor.  I should have caught this when I reviewed it yesterday.  It is the cause of the heap-use-after-free sanitizer failure.


keithbostic:
lgtm, merged.

PR #2669, opened 2016-04-19
WT-2565: item 3573 on page at [write-check] is a corrupted cell

keithbostic:


keithbostic:
@michaelcahill, I didn't track it down, but by inspection, I think the problem is we have to copy any rows that aren't "taken" down in the page, and if we exit the original loop early, we won't have slots referencing those rows, so we'll end up with corruption (or lost rows).

I you don't like this approach, maybe a better solution would be to change your code to not exit the loop, but instead to save a "maximum usable slot" value and pass that value to the underlying compression routine instead of the real "slots" value, minimizing the slots the compression engine will consider.

That's a little nicer because it gives us more fine-grained control of the image size (it's closer to half of `WT_BTREE.maxmempage`, where my change won't be, mine will be whatever chunk crosses that boundary), but I didn't think of it first.

michaelcahill:
@keithbostic, I'm missing something here: when this version sets `image_too_large` true, it still passes all of the data to the raw compression callback.  How does that solve the original problem?

michaelcahill:
Maybe the previous "fix" didn't work the way I thought it did...

What I would like to have happen here is that we cap the amount of data supplied to raw compression to the maximum size.  Any subsequent slots that reconciliation has accumulated should be past the end of the slot array we pass to the callback.  Does that make sense?

michaelcahill:
TL;DR -- I like your suggested "better solution" ;-)

keithbostic:
@michaelcahill 
> I'm missing something here: when this version sets image_too_large true, it still passes all of the data to the raw compression callback. How does that solve the original problem?

From the tl;dr, I'm guessing you've figured this out, but as soon as the total data available to the compression engine is larger than `WT_BTREE.maxmempage`, we'll set the `no more data` flag so the engine quits asking for more data. Unless that's the first chunk of data (which I don't think can happen), the original problem is fixed.

keithbostic:
See #2672 for an alternate approach to this branch.

michaelcahill:
replaced by #2672, closing.

PR #2672, opened 2016-04-20
WT-2565: limit the amount of data passed to raw compression

keithbostic:
See #2669 for discussion, this is an alternate approach to that branch.


michaelcahill:

PR #2673, opened 2016-04-20
WT-2567 Check if logging is enabled for truncate.

sueloverso:
@keithbostic Please review this change.  I was able to repro the original segfault on my AWS box.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -316,11 +316,12 @@ __wt_log_truncate_files(
 	conn = S2C(session);
 	if (!FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))
 		return (0);
-	log = conn->log;
 	if (F_ISSET(conn, WT_CONN_SERVER_RUN) &&
 	    FLD_ISSET(conn->log_flags, WT_CONN_LOG_ARCHIVE))
 		WT_RET_MSG(session, EINVAL,
 		    "Attempt to archive manually while a server is running");
+
+	log = conn->log;
 
 	backup_file = 0;
 	if (cursor != NULL)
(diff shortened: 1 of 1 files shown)


src/conn/conn_log.c
```diff
@@ -314,6 +314,8 @@ __wt_log_truncate_files(
 
 	WT_UNUSED(cfg);
 	conn = S2C(session);
+	if (!FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))
+		return (0);
```
keithbostic: IMHO, I'd suggest re-ordering these lines and tossing in some whitespace, the initializaiton of "log" in-between two tests of connection flags seems messy:
```
WT_UNUSED(cfg);

conn = S2C(session);
if (!FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))
        return (0);
if (F_ISSET(conn, WT_CONN_SERVER_RUN) && 
    FLD_ISSET(conn->log_flags, WT_CONN_LOG_ARCHIVE))
        WT_RET_MSG(session, EINVAL,
            "Attempt to archive manually while a server is running");

log = conn->log;
```

keithbostic:
lgtm, some whitespace suggestions, but change seems fine.

Any idea why we never saw this before?

src/conn/conn_log.c
michaelcahill: Thanks @keithbostic, I've made the style change.

I suspect we're seeing this now because I added diagnostic yields into `__wt_writelock`: that should help shake out some additional races (e.g., on open or close in this case).

sueloverso:
I suspect we never saw this before because all incremental backup testing (i.e. ex_backup.c) always has logging enabled.  test/format and incremental backup with logging off is a new combination.

PR #2674, opened 2016-04-20
WT-2568 Add a backward compatible constructor for PackInputStream.

ddanderson:
This fixes java's PackTest (and potentially other code) that may have been affected by changing a constructor in WT-2550.

Aside: Apparently our pull request tester doesn't build java, and I didn't see it because I didn't rebuild from scratch when testing.


michaelcahill:
Thanks @ddanderson, I did wonder whether we had other code relying on that constructor.  LGTM, I'll merge.

PR #2676, opened 2016-04-21
WT-2544 Fix eviction stats when clear is used

fedorova:
…ed properly within the library, so reported statistics were garbage. I think these statistics update macros were accidentally removed when we change the code to keep track of server, worker and app evicts internally in the cache, not to rely on statistics that may or may not be turned on by the user. Users may still want to see these statistics, so they need to be updated appropriately.


fedorova:
That code is not working. I was seeing totally inconsistent statistics. The sum of pages evicted by application threads, server and eviction workers did not add up to the total number of pages evicted, by a huge margin. The 2544 branch, which updated the statistics in the same way as in my proposed fix did not have this problem. My statistics are configured with the "clear" flag, so they are periodically reset. I can take a look to understand why.

fedorova:
@michaelcahill : I figured out why this code is not working for me. I am running with a "clear" flag in the stats configuration, because I like to see how stats change over time. I know that @agorrod also likes to use this setting; in fact I use it because he recommended it. The code that updates statistics, __wt_cache_stats_update, sets statistics to equal to the values kept in the cache struct: server_evicts, worker_evicts, app_evicts. These values are accumulated over time and are not reset when we reset these statistics. That is why they always appear cumulative, even if I configure to clear statistics periodically. Other stats are cleared correctly, so I am seeing inconsistent numbers. I'll fix my branch to not read the cumulative values in __wt_cache_stats_update, retest and commit.

fedorova:
@michaelcahill : the other stat fields that are set in __wt_cache_stats_update are okay, because they are never meant to be cleared: the __wt_stat_connection_clear_single code skips them when clearing statistics.

fedorova:
@michaelcahill, wanted to check with you if my explanation of the bug makes sense.

sueloverso:
@fedorova  I will look into this today

sueloverso:
@michaelcahill I've updated to the latest develop.  This LGTM.  I don't see any way around this change given the use with `clear`.  If you have any ideas let us know.  Otherwise this is good to merge.

michaelcahill:
Thanks @fedorova and @sueloverso, this does make sense to me -- I no longer use "clear" because the tool we use to visualize stats takes care of calculating rates for stats that are "events / second".

I'll merge.

PR #2677, opened 2016-04-21
WT-2560 Spin on transaction locks.

michaelcahill:
The default backoff sleeps cause severe performance regressions due to
contention between readers trying to scan.


agorrod:

PR #2679, opened 2016-04-21
WT-2572: don't select an in-memory format run if incompatible options configured

keithbostic:
@sueloverso, for your review.

### DIFF
--- a/test/format/config.c
+++ b/test/format/config.c
@@ -35,6 +35,7 @@ static void	   config_encryption(void);
 static const char *config_file_type(u_int);
 static CONFIG	  *config_find(const char *, size_t);
 static void	   config_in_memory(void);
+static void	   config_in_memory_check(void);
 static int	   config_is_perm(const char *);
 static void	   config_isolation(void);
 static void	   config_lrt(void);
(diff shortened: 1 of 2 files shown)


sueloverso:
@keithbostic That looks good to me for the recovery test.  Feel free to merge when testing completes.

test/format/config.c
```diff
@@ -331,11 +328,43 @@ config_encryption(void)
...
+ *	Periodically set up an in-memory configuration.
  */
 static void
 config_in_memory(void)
 {
+	/*
+	 * Configure in-memory before configuring anything else, in-memory has
+	 * many related requirements. Don't configure in-memory if there's any
+	 * incompatible configurations, so we don't have to configure in-memory
+	 * every time we configure something like LSM, that's too painful.
```
sueloverso: The comment mentions LSM, but where do we check for LSM?  The others look good.  I think we want to add "compact" to the list.

sueloverso: Actually, compact seems to be a no-op, not an error.

keithbostic: @sueloverso, thanks, you're right, and I've added the additional check.

sueloverso:
Thanks! LGTM once testing completes.

PR #2683, opened 2016-04-21
WT-2577: core dump discarding non-existent addresses

keithbostic:
Ready for review/merge.


michaelcahill:
Thanks, @keithbostic -- nice catch, this lgtm.  I'll merge.

PR #2684, opened 2016-04-22
 WT-2574 Ensure test/format doesn't leak memory

agorrod:
It can if an option is specified in the configuration file and
the command line.


agorrod:
autobuild: retest this please

michaelcahill:

PR #2686, opened 2016-04-22
WT-2553 Fix implicit size_t conversion in test/format.

michaelcahill:
test/format/config.c:415:16: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'uint32_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]


michaelcahill:
Another one I'll self-approve.

PR #2687, opened 2016-04-22
WT-2576: variable-length column-store out-of-order return

keithbostic:
Ready for review/merge.


sueloverso:
Thanks @keithbostic.  LGTM.  I will merge.

PR #2688, opened 2016-04-22
WT-2579: in-memory configurations break debugging support

keithbostic:
@michaelcahill, @agorrod: I want to give up trying to run the debugging functions through the pluggable I/O functions, it's too hard.

This also removes the need for the `WT_STREAM_LINE_BUFFER` flag.


agorrod:
@keithbostic Will this change mean we can't compile with diagnostic on Windows any more?

keithbostic:
@agorrod, I don't think so: it's just `fopen` and `fprintf` calls, which work on Windows as far as I know, and we've had them in the code until relatively recently.

agorrod:
lgtm - I'll merge.

PR #2681, opened 2016-04-21
WT-2571 Join cleanup

ddanderson:
This is a followup for WT-2360.  Renaming and reordering functions, some reorganization of test cases to use scenarios.

PR #2689, opened 2016-04-22
WT-2571 Join code clean up

ddanderson:
This is a followup for WT-2360. Renaming and reordering functions (big diffs!), and some reorganization of test cases to use scenarios.  (Use this pull request instead of #2681)


ddanderson:
@michaelcahill, some code clean up.  Everything in cur_join.c is essentially moving functions around.

michaelcahill:

PR #2690, opened 2016-04-22
WT-2574: format doesn't free all allocated configure memory

keithbostic:
Rework format configuration to use config_single() more, that way we
discard allocated memory when in-memory configurations turn off previous
configurations.


sueloverso:
This LGTM, although I don't fully understand test/format's structure.

PR #2691, opened 2016-04-22
WT-2576 Skip tiny pages if we can.

sueloverso:
@keithbostic Please review this eviction change to skip tiny pages.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -648,7 +648,7 @@ __evict_pass(WT_SESSION_IMPL *session)
 		 */
 		if (pages_evicted == cache->pages_evict) {
 			WT_STAT_FAST_CONN_INCR(session,
-					       cache_eviction_server_slept);
+			    cache_eviction_server_slept);
 			/*
 			 * Back off if we aren't making progress: walks hold
 			 * the handle list lock, which blocks other operations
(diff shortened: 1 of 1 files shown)


sueloverso:
@michaelcahill Please review.  Do you think using allocsize is adequate?  I chose it because leaf_page_max needs to be a multiple of allocation_size, but can be as small as 512b.  So I wanted to avoid a hard-coded value and that seemed reasonable.  Is that too small?

src/evict/evict_lru.c
```diff
@@ -1389,6 +1389,11 @@ __evict_walk_file(WT_SESSION_IMPL *session, uint32_t queue_index, u_int *slotp)
 		    page->memory_footprint < btree->splitmempage)
 			continue;
 
+		/* Skip tiny pages unless we're aggressive.  */
+		if (!FLD_ISSET(cache->state, WT_EVICT_PASS_AGGRESSIVE) &&
+		    page->memory_footprint < btree->allocsize)
+			continue;
+
```
michaelcahill: @sueloverso, I would have checked this condition instead:

```c
/* Skip tiny pages when running in-memory.  */
if (F_ISSET(conn, WT_CONN_IN_MEMORY) &&
    page->memory_footprint < WT_THOUSAND)
```

michaelcahill:
Decided to take a more general look at eviction for in-memory workloads in WT-2594.

PR #2694, opened 2016-04-23
WT-2584: don't use periods in error messages

keithbostic:
Periods followed by colons looks bad, remove periods from error messages.


agorrod:

PR #2696, opened 2016-04-25
WT-2583 Incremental backup fix

sueloverso:
@keithbostic Here's a fix for the incremental backup failure problem that leaves the source directory in a bad state if we fail while the incremental backup cursor is open.  I have tested this with the repro case you posted and the verify command succeeds.  Please review.


sueloverso:
@keithbostic Please review these changes for backup.

michaelcahill:
Thanks, @sueloverso -- as discussed, needing the temporary file here isn't pretty, but this looks right to me.  I'll merge.

PR #2697, opened 2016-04-27
WT-2446 Add stat for number of calls to get a page.

sueloverso:
@keithbostic Please review this change to add a stat to count the number of calls to get a page.  With this, and "pages read into cache" the user can compute the cache miss ratio, and therefore, the cache hit ratio.  I also added one at the dsrc level too.


agorrod:
@sueloverso Do know whether this has a performance consequence?

sueloverso:
The two runs of evict-btree.wtperf did not show any performance impact at all.  But that is not a large dataset.

sueloverso:
@agorrod I ran my full eviction set of tests and there was no performance impact.

keithbostic:
lgtm, merged.

PR #2699, opened 2016-04-28
WT-2479 Refactored json dump to eliminate duplicate code.

ddanderson:
This also fixes a malloc leak in dump_table_config_complex().

Compared JSON output from all jsondump runs in test/suite, all identical modulo a difference arising from using a `metadata:create` cursor previously, now we use a (regular) `metadata:`.   The differences will be tracked in WT-2591.  For JSON dump the difference appears to be benign: 1) the test cases run normally 2) the previous values dumped for key_format/value_format of column groups appeared to be wrong, with this commit they are right 3) the key_format/value_format for a column group is likely to be ignored on input (it's normally not specified), since it can be derived from the formats given for the `table:`.

### DIFF
--- a/src/utilities/util_dump.c
+++ b/src/utilities/util_dump.c
@@ -463,7 +463,7 @@ dump_prefix(WT_SESSION *session, bool hex, bool json)
 
 	if (printf(
 	    "WiredTiger Dump (WiredTiger Version %d.%d.%d)\n",
-		vmajor, vminor, vpatch) < 0 ||
+	    vmajor, vminor, vpatch) < 0 ||
 	    printf("Format=%s\n", hex ? "hex" : "print") < 0 ||
 	    printf("Header\n") < 0)
 		return (util_err(session, EIO, NULL));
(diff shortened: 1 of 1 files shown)


ddanderson:
retest this please

ddanderson:
@keithbostic, this is ready for review.  I know this isn't quite the way you suggested to fix this, but I think this may be a more minimal approach.  What do you think?

michaelcahill:
lgtm, @ddanderson -- @keithbostic should also review but I think this is fine to merge.

@keithbostic, just a note to please edit the commit message so the ticket number only appears once.

keithbostic:
@ddanderson, this looks good to me as well (I pushed a few trivial whitespace changes).

However, most of the problems in WT-2381 were in LSM, and there were changes to `test/suite/test_dump.py` to test LSM tables.

Should we add similar tests to `test_jsondump01.py`, or would it be simpler to merge that JSON testing into `test_dump.py`?

ddanderson:
@keithbostic, I opened WT-2597 to track the need for better tests, thanks!
I'll merge.

PR #2700, opened 2016-04-28
WT-2586: Remove ex_config.c until config cursors are supported

keithbostic:
Ready for review/merge.


michaelcahill:
@keithbostic, there's no point having `ex_config.c` atm, we should just remove it from the tree until WT-1133 is done.

keithbostic:
@michaelcahill, I'm pretty sure `ex_config.c` doesn't have anything to do with "config:", other than it opened the "config:" URI long ago -- that's kind of why I got confused.

From the top of the file: `This is an example demonstrating how to configure various database and table properties.`

It's currently the home of at least one documentation snippet, too.

michaelcahill:
@keithbostic, I'd forgotten about the cache size snippet -- that should be in `ex_all.c`.  The rest of the code was supposed to demonstrate querying configuration, hence why it opens a config cursor.

keithbostic:
@michaelcahill, OK, done.

PR #2701, opened 2016-04-28
WT-2343 Add test to make sure we cannot rename/drop when a backup cursor is open.

sueloverso:
@keithbostic Please review this change to a backup test that verifies we cannot change the file names via rename or drop while a backup cursor is open.


sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

michaelcahill:
Thanks @sueloverso, lgtm -- I'll merge.

PR #2703, opened 2016-04-29
WT-2586 Fix Windows build after removal of ex_config

agorrod:


agorrod:
I'm self reviewing here - this fixes the Windows build.

PR #2704, opened 2016-04-29
WT-2595 Fix compiler warning in packing test.

agorrod:
While here switch from using asserts to test util functions.


agorrod:
@keithbostic Would you mind doing a review of this small test change?

keithbostic:
lgtm, merged.

PR #2705, opened 2016-04-29
WT-2592 Fix joins for the non-recno, non-raw case

ddanderson:
This fixes a rather fundamental problem.  Any joins written in C that do not use recno were not seeing results.  Internally we are using raw keys, but our final 'main' cursor may not be in raw mode, so we may not be able to retrieve the values using the key we have.

Unfortunately, the bulk of the test cases for join are in Python, which use RAW mode, so unaffected.  Our C example uses recno, which took a different path.

### DIFF
--- a/src/cursor/cur_join.c
+++ b/src/cursor/cur_join.c
@@ -1113,7 +1113,11 @@ __curjoin_next(WT_CURSOR *cursor)
 		c->set_key(c, iter->curkey);
 		if (!F_ISSET(cursor, WT_CURSTD_RAW))
 			F_CLR(c, WT_CURSTD_RAW);
-		WT_ERR(c->search(c));
+
+		/* A failed search is not expected, don't return WT_NOTFOUND. */
+		if ((ret = c->search(c)) == WT_NOTFOUND)
+			ret = WT_ERROR;
+		WT_ERR(ret);
 		F_SET(cursor, WT_CURSTD_KEY_INT | WT_CURSTD_VALUE_INT);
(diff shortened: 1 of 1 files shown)


ddanderson:
@michaelcahill, can you review please?

src/cursor/cur_join.c
```diff
@@ -1104,19 +1103,21 @@ __curjoin_next(WT_CURSOR *cursor)
...
-			WT_ERR(__wt_vunpack_uint(&p, iter->curkey->size,
-			    &cjoin->iface.recno));
-			c->set_key(c, cjoin->iface.recno);
-		} else
-			c->set_key(c, iter->curkey);
-		WT_ERR(c->search(c));
+		F_SET(c, WT_CURSTD_RAW);
+		c->set_key(c, iter->curkey);
+		if (!F_ISSET(cursor, WT_CURSTD_RAW))
+			F_CLR(c, WT_CURSTD_RAW);
```
michaelcahill: @ddanderson, we have `__wt_cursor_set_raw_key` that should take care of these 4 lines for you:
```c
    __wt_cursor_set_raw_key(c, iter->curkey);
```

michaelcahill:
@ddanderson, it sounds like we should we add a non-recno join test in C?

Otherwise, this looks fine to me with a minor simplification.

ddanderson:
@michaelcahill, thanks for the simplification!  I added WT-2605 to track the need for better tests.  That should be easier once Alex has a C test framework (WT-2554).

Merging now.

PR #2706, opened 2016-04-29
SERVER-23661: $sample takes disproportionately long time on newly created collection

keithbostic:
@michaelcahill, calling `__wt_page_evict_soon` works well, in my tests it brings $sample of newly created collections down to about the same time as a re-opened collection.

I ended up with a pretty stupid "should we evict this page" test of "greater than 5MB". That performed as well as anything else I tried, and it didn't seem unreasonable to use the page's memory footprint for this purpose, once we know the page's footprint reflects just the one big skiplist.

Ready for review/merge.

### DIFF
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -634,6 +634,7 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
 	WT_INSERT *ins, **start, **stop;
 	WT_INSERT_HEAD *ins_head;
 	WT_PAGE *page;
+	uint64_t samples;
 	uint32_t choice, entries, i;
 	int level;
 
@@ -688,7 +689,7 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
 	 * Step down the skip list levels, selecting a random chunk of the name
 	 * space at each level.
 	 */
-	while (level > 0) {
+	for (samples = entries; level > 0; samples += entries) {
 		/*
 		 * There are (entries) or (entries + 1) chunks of the name space
 		 * considered at each level. They are: between start and the 1st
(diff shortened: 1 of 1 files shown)


src/btree/row_srch.c
```diff
@@ -765,6 +765,15 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
 	cbt->ins_head = ins_head;
 	cbt->compare = 0;
 
+	/*
+	 * MongoDB $sample can be significantly slower than expected if the
+	 * collection is newly created and the content is in a in a single
+	 * skiplist. If searching a relatively large skiplist, schedule it
+	 * for eviction to split up the skiplist.
+	 */
+	if (page->memory_footprint > 5 * WT_MEGABYTE)
```
agorrod: @keithbostic Can you explain why the regular ``__evict_force_check`` check from ``__wt_page_in_func`` wasn't catching this? Did you consider using ``memory_page_max`` rather than a magic number?

keithbostic: The performance problem appears in pages with big skiplists but not yet large enough to be forcibly evicted, that is, pages with a big footprint but less than `memory_page_max`.

I considered using `memory_page_max`, but decided `memory_page_max / 2` was more likely to go wrong than `5MB`: this really is "what memory footprint implies a skiplist too large to walk 100K times in the next couple of seconds?"

keithbostic:
@agorrod, commit df7f734's JIRA ticket number is wrong, it should be SERVER-23661, not SERVER-23611.

I fixed the branch title, but I'm guessing that's not enough, let me know if I should re-create the branch.

agorrod:
@keithbostic FWIW: I don't mind about the branch name being wrong - it's fine to keep working in this pull request.

We talked this morning about switching the trigger from being based on the page size, to being based on the amount of work being done in the ``__wt_row_random_leaf`` function to get a single result (i.e: a proxy for the size of the skip list).

keithbostic:
> FWIW: I don't mind about the branch name being wrong 

Just to be clear, the commit log message has the wrong ticket number as well.

keithbostic:
@agorrod, I've pushed a change that doesn't evict  the page unless there are at least 5K entries in the page. In my test runs, we see a lot of small samples (~100), but periodically we see a large sample count, and so as soon as we see that large count, we'll evict/split the page.

I left in the test for a 5MB page, I don't think we want to evict small pages any more than we want to evict pages with too-few entries on the skiplist. That said, I don't think it matters a lot, so if you disagree, feel free to remove that part of the test.

agorrod:
Thanks @keithbostic I did take out the page size check. I'll merge.

PR #2707, opened 2016-04-30
WT-2599: split out the checksum code from the support directory

keithbostic:
@michaelcahill, @agorrod, for review/merge.


keithbostic:
I verified the patch builds on the PPC, so this should be ready to go.

michaelcahill:
This is an odd collection of changes, but they look okay to me.  I'll merge.

PR #2708, opened 2016-05-01
WT-2600: clean up test program #includes

keithbostic:
@agorrod, this is just some minor changes to make lint happier with the test programs.

I'm sure it's going to break on some machine/compiler configuration, but I'll test it where I can explicitly and shepherd it through if it breaks.


agorrod:
Thanks @keithbostic lgtm.

PR #2709, opened 2016-05-01
WT-2598: in-memory FS needs fast lookup on file names

keithbostic:
@agorrod, we talked about the changes to support hash buckets on in-memory filesystem names, that's in 347fc73.

Additionally, 1a932f2 changes our hash bucket implementations to track the bucket ID instead of the name hash. We don't resize the hash queues so there's never a need to recompute the bucket ID. I doubt it will speed anything up, but it's wasted effort.


agorrod:
@keithbostic Would you mind splitting this into two different pull requests? I know the changes are somewhat related, but I've always thought that the hash lookups use the name to give us the option of dynamic hash buckets in the future. If we remove that I'd like an easy way to find the original version - I'm particularly thinking we may run into this with some of the 100k+ table workloads we've encountered in MongoDB.

keithbostic:
Sure, done.

sueloverso:
LGTM.  Thanks.

PR #2710, opened 2016-05-02
WT-2602 In LSM, use the chunk size to set maximum page size in memory.

michaelcahill:


michaelcahill:
@agorrod, with the `test/format` config in the ticket, this keeps uncompressed page images under 3.5MB.  Previously they were growing to 9MB.  This is still large, but should not be able to fill the cache with pinned pages.

Can you please review?

agorrod:

PR #2711, opened 2016-05-02
WT-2603 Track the hash bucket ID for hash queues, not the name hash value

keithbostic:

As I said before, I don't care too much if we take this or discard it.

The most valuable piece is probably skipping the metadata name comparison?

FTR, as far as dynamically resizing the hash table, I don't think having the original hash changes the problem much, it only means we don't have to recalculate the hash during the resize.

### DIFF
--- a/src/block/block_open.c
+++ b/src/block/block_open.c
@@ -102,11 +102,9 @@ __block_destroy(WT_SESSION_IMPL *session, WT_BLOCK *block)
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
-	uint64_t bucket;
 
 	conn = S2C(session);
-	bucket = block->name_hash % WT_HASH_ARRAY_SIZE;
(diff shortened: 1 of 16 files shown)


sueloverso:
FTR, the changes themselves LGTM.  I'm ambivalent on the change itself.  I agree about the metadata name change too.

If we find the hash table needs resizing, I suspect we'd just pick a new size.  Dynamic resizing sounds like overkill to me at the moment, envisioning future issues.  But I could be wrong.

michaelcahill:
I have no love for the current check for the metadata, so I'd take that part of this branch.

That said, having the hash value seems more useful than the bucket: we could skip most calls to `strcmp` by first checking that the hash values match.  We also have the option of having the same handle appear in differently-sized hash tables (e.g., if we decided the session cache should be a different size to the connection cache).

keithbostic:
Close without merging; moving the metadata check change to WT-2674.

PR #2712, opened 2016-05-03
WT-2554 Add initial C test framework.

keithbostic:
@agorrod, @sueloverso: here's something to show you the direction I'd take on this one.

Changes:

* every test program has its own directory (multiple standalone programs in a directory means naming collisions/confusion for no gain that I can think of),
* there's a script dist/s_csuite that creates Makefiles for the test programs, and complains if they're not listed in the build_posix/Make.subdirs file, so people don't forget to add them there,
* split the `test_util.i` file into `test_util.h`, `misc.c` and `parse_opts.c`, now that we have a utility library, there's no point to having 20 copies of the `testutil_die` function per program.


keithbostic:
autobuild: retest this please

sueloverso:
@keithbostic most of these changes are fine with me.  But I'm going to continue to beat the naming drum.  ``csuite/wt1965/main.c`` is meaningless to me.  The test itself (currently) has no comments whatsoever.  So, what does that actually test?  No idea.  Looking at WT-1965, it has about 15-20 comments, and the ticket title is about high CPU usage.  Still no idea.  There are going to be tickets that evolve in their discussion and some test will test a single item in some long stream of comments.  I don't want to guess nor dig through the ticket to find that.  So I suggest:
* calling the source file something meaningful.  There's no reason it needs to be ``main.c``.
* enforce some kind of test description in every test in there.  Something enforced by a script in ``s_all``.  (Include existing tests, such as cursor_order, readonly, recovery, etc)
* do we bring existing tests into line?  For instance, ``test/recovery/`` has two unrelated tests but both dealt with bugs about recovery originating from tickets.  Other subdirs in ``test`` are similar, such as ``cursor_order, readonly, manydbs, ...``  Those don't reference their original tickets although the commits do (for now).
* basically, I want to know what the test specifically tests (most important) and where it came from (i.e. ticket) (less important) without having to search or dig anywhere but the test itself.

keithbostic:
@sueloverso 

I agree, an explanatory README file in every test directory would be wonderful. However, once there's a README file, I see no reason for additional file naming requirements.

sueloverso:
A README file would be acceptable.  I'd like it to be enforced by some script run by ``s_all`` and perhaps required to have a minimum size (maybe 100 bytes) to make sure we get a minimal description.  I'd also be okay with more structure to it, if others want that.

michaelcahill:
My main comment here is that I would strongly prefer a single `Makefile.am` in the top-level directory.  That's because (a) changes to `Make.subdirs` in the top-level build requires re-running configure, and (b) I want these tests to be run in parallel during `make check`.

keithbostic:
Discard this branch, there's nothing we're taking.

michaelcahill:
To be clear, we could organize the source files into a directory per test and still have a single `Makefile.am` in the top level directory.  I haven't reviewed the rest of the changes here, but I wasn't suggesting it all be thrown away.

PR #2714, opened 2016-05-03
WT-2611 Modify wtperf to handle escaped quotes.

ddanderson:
This simple solution allows backquotes to get into the WT config parsing, where they are accepted but not transformed (see the JIRA comments), and then unescape them when we get them back from the parser.  

Question for reviewer: it may be preferred to put the unescape code in the config parsing within WiredTiger.  If we go that route, we probably need to allocate a buffer when needed (probably hanging off of WT_CONFIG) to hold the unescaped text.

Also wanted to make sure I wasn't missing some other way of escaping quotes before I went for a more involved solution.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -47,6 +47,53 @@ static void config_opt_usage(void);
 	(strncmp(str, bytes, len) == 0 && (str)[(len)] == '\0')
 
 /*
+ * config_unescape --
+ *	Modify a string in place, replacing any backslash escape sequences.
+ *	The modified string is always shorter.
+ */
+int
+config_unescape(char *orig)
+{
+	char ch, *dst, *s;
+
+	for (dst = s = orig; *s != '\0';) {
+		if ((ch = *s++) == '\\') {
+			ch = *s++;
+			switch (ch) {
+			case 'b':
+				*dst++ = '\b';
+				break;
+			case 'f':
+				*dst++ = '\f';
+				break;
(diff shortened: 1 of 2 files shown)


ddanderson:
@sueloverso, can you try this out?  If it works, there are ways to refine it.  Thanks!

sueloverso:
Thanks @ddanderson.  I tried with a single backslash to escape a double-quote within a wtperf config file.  It is working as expected!  That is really helpful.  If you want to refine it, go ahead.

sueloverso:
@ddanderson You mentioned wanting to refine this change possibly.  Is this PR ready for review and merge or is it still in-progress?

michaelcahill:
@sueloverso, this seems okay to me: can you please do the review?

sueloverso:
@ddanderson and @michaelcahill I did review and I am good with the current code.  Don mentioned wanting to refine the change.  I am happy to get this merged or to wait for that change.

ddanderson:
@michaelcahill, my thought was that a better solution might be to put the unescape code in the config parsing within WiredTiger. Then anyone using config parsing gets the benefit. If we go that route, we probably need to allocate a buffer when needed (probably hanging off of WT_CONFIG) to hold the unescaped text.

Is it worth it? Should we merge this one in the meantime?

sueloverso:
While that may be something to consider, I think we should merge this in the meantime.  If we want to move it into the library we can make the necessary changes here too at that time.

PR #2717, opened 2016-05-04
WT-2613 - Add WT_UNUSED to a variable to fix Windows compilation.

daveh86:


daveh86:
@agorrod, mind having a look?

agorrod:
Thanks @daveh86 lgtm

PR #2718, opened 2016-05-04
WT-2617: Bug in pluggable file system example

keithbostic:
@agorrod, the only thing I can find here is the file system example didn't separate the concepts of "buffer data size " and "buffer memory size", so the read of the file was somehow getting confused. I don't see the exact path that would cause it, but I think there's a potential problem there.

If you merge 461c63d, I think we can close the JIRA ticket until it appears again.


agorrod:
Thanks @keithbostic These changes look good to me. I'm not convinced they will fix the failure - I suspect the failure is due to a race between read/write that isn't protected by the example file system implementation.

I'm going to hold off merging this branch for now, in case I can think of some reasonable way to add locking into the example code without introducing a platform problem.

keithbostic:
@agorrod, that makes sense it might be a race.

Why not add calls to `__wt_writelock` in the methods?

PR #2719, opened 2016-05-04
WT-2616: Fix a deadlock with in-memory size lookups.

keithbostic:
Eliminate a race where the handle could be removed between the search
and the size check.

### DIFF
--- a/src/os_common/os_fs_inmemory.c
+++ b/src/os_common/os_fs_inmemory.c
@@ -80,6 +80,22 @@ __im_handle_remove(WT_SESSION_IMPL *session,
 }
 
 /*
+ * __im_handle_size --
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, I looked at the race issue yesterday and didn't think it could happen -- we're already relying on the upper layer for most concurrency control in there (e.g., the in-memory reference count is set to 1 on open, never incremented, then decremented on close).  This change is fine by me, but I think the race already prevented by locking / reference counting in the `WT_FH` layer.

keithbostic:
@michaelcahill, I agree the race can't happen today (at least, I didn't see a path where it could happen).

However, I know of no policy that would prevent a future code change from creating a race, which is why I made the change.

PR #2720, opened 2016-05-04
WT-2615: Enabling checkpoints in test/format leads to reduced concurrency

keithbostic:
Ready for review.


agorrod:

PR #2721, opened 2016-05-05
WT-2621 Fix missing parentheses on call to __wt_errno

daveh86:


daveh86:
@agorrod, can you review this? Should allow alt-compile in Win64

agorrod:

PR #2723, opened 2016-05-05
WT-2633 Don't allow eviction of the metadata during checkpoints in MongoDB 3.0

keithbostic:
Avoid a path to the checkpoint thread performing eviction.


keithbostic:
@michaelcahill, here's the change I'd recommend for BF-2313.

I was never able to reproduce the error, but I think this change is low-risk and a likely fix.

I didn't add `WT_SESSION_NO_EVICTION`, instead, I added a check for `WT_SESSION_NO_CACHE_CHECK` in the page-release eviction path.

I think that's a smaller change more inline with the way things worked in 3.0 (compare with a similar check in `__wt_page_in_func`, [linked here](https://github.com/wiredtiger/wiredtiger/blob/mongodb-3.0/src/btree/bt_page.c#L135-L152)).

FTR, this change isn't needed in MongoDB 3.2.

agorrod:

PR #2725, opened 2016-05-06
WT-2624 Fix example program build with MSVC 2013

daveh86:


daveh86:
@agorrod, mind having a look at this?

examples/c/ex_file_system.c
```diff
@@ -31,6 +31,12 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#ifndef _WIN32
+#include <unistd.h>
```
agorrod: Why did you add ``unistd.h`` here? Is there a missing symbol without it?

daveh86: Added so that it uses the same logic as the other example files.

I don't think its needed in this case but wanted to keep the examples looking similar.

Can remove if need be.

agorrod: @daveh86 Yes - please make that change. Maybe copy the clause from ``ex_scope.c``

agorrod:

PR #2726, opened 2016-05-06
WT-2626 - Fix MSVC 2015 snprintf redefinition

daveh86:


agorrod:
@daveh86 This change looks reasonable to me, but can you explain why ``ex_log.c`` doesn't suffer from the same problem?

daveh86:
ex_log isn't compiled as part of "check" on Windows, as are several of the other tests. We have made this change in a few different ways throughout the various examples, it may be worth doing something to make all the examples consistent.

agorrod:

PR #2727, opened 2016-05-06
WT-2627: Coverity complaints

keithbostic:
Ready for review/merge.

Dear reviewer: there's a complaint Coverity has where I can't spot the error, and I don't see why Coverity would get it wrong. If you have the time to take a look at it, it's [Coverity complaint 1355027](https://scan6.coverity.com/reports.htm#v27510/p10746). I've simplified the code in 224a88d to try and clarify the issue.

### DIFF
--- a/src/config/config_collapse.c
+++ b/src/config/config_collapse.c
@@ -38,6 +38,8 @@ __wt_config_collapse(
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 
+	*config_ret = NULL;
+
 	WT_RET(__wt_scr_alloc(session, 0, &tmp));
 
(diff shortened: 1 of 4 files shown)


sueloverso:
@keithbostic I studied the Coverity issue and its details.  I don't see it either.  While I see all the red lines describing it, following the stack all the way back up I don't see a path that Coverity thinks will give a non-zero return from `metadata_search` while allocating the memory.

daveh86:
@keithbostic, changes LGTM. I've also had a look at the coverity issue in question and agree with Sue.
I can't see a path where there would be space allocated into the value variable in question that also results in a non-zero return.

PR #2728, opened 2016-05-07
WT-2628: reconciliation can return without unlocking the page lock

keithbostic:
Read for review.


agorrod:

PR #2730, opened 2016-05-12
WT-2630 Rename pluggable filesystem methods to avoid reserved names.

michaelcahill:
Use "fstr" consistently as the variable name and prefix for WT_FSTREAM.


daveh86:
retest this please

daveh86:
retest this please

daveh86:
retest this please

michaelcahill:
@keithbostic, can you please review this one?

keithbostic:
Lgtm.

You didn't rename the `WT_FSTREAM.close` and `WT_FILE_HANDLE.close` methods, and close is potentially an issue. If that was deliberate, I won't argue (if `close` were ever to become a problem, we're in trouble, changing `WT_CURSOR.close`, or `WT_CURSOR.remove`, would be painful).

michaelcahill:
@keithbostic, thanks for the review.  I did think about `close`, but as you note, we're already relying on that working in `WT_CURSOR`.  I'll merge...

PR #2731, opened 2016-05-16
WT-2610 Reduce hazard pointer array size.

sueloverso:
@michaelcahill Here are the changes.  I committed the changes to wtperf to add `session_count_idle` as a separate commit in case we don't want to keep that.  Please review.


daveh86:
retest this please

sueloverso:
@agorrod If you want to, please review these changes.  We can consider the wtperf changes separately from the hazard array changes.

sueloverso:
@agorrod I separately committed the couple of changes we discussed.  I believe this is now ready for review.

src/btree/bt_discard.c
```diff
@@ -63,30 +63,6 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
...
-		__wt_sleep(0, 10000);
-	}
-	if (hp != NULL)
-		__wt_errx(session,
-		    "discarded page has hazard pointer: (%p: %s, line %d)",
-		    hp->page, hp->file, hp->line);
-	WT_ASSERT(session, hp == NULL);
-	}
-#endif
-
```
michaelcahill: @sueloverso, what was the thinking with this change?  This check has caught real bugs in the past: is it necessary to get rid of it?

sueloverso: @keithbostic and I discussed this a bit on the bus ride last week, questioning if it was still useful.  This diagnostic assertion was added almost 3 years ago and hasn't found anything in a while and the hazard code seems stable.  It certainly impacts diag runs.  But, it is under diagnostic mode so it doesn't impact end users.  I can go either way on this.

agorrod:
Thanks @sueloverso lgtm - I'll go ahead and merge. Let's watch the perf tests for improvements.

PR #2732, opened 2016-05-17
WT-2635: Test/format on 3.0 attempts to use bzip compression

keithbostic:
Change test/format to not choose bzip compression.


agorrod:

PR #2733, opened 2016-05-17
WT-2637: The file-extension documentation doesn't cover not-supported cases

keithbostic:
The file-extension documentation doesn't cover not-supported cases.

Public domain the POSIX file system implementation, it's a natural starting point for anyone writing their own implementation.


michaelcahill:

PR #2736, opened 2016-05-19
WT-2646 - Allow drop without waiting for checkpoints

daveh86:
…ithout the checkpoint lock

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -787,15 +787,19 @@ methods = {
 ]),
 
 'WT_SESSION.drop' : Method([
+    Config('checkpoint_wait', 'true', r'''
+        do not wait for the chheckpoint lock, if \c checkpoint_wait=false,
+        fail if this lock is not available immediately''',
+        type='boolean'),
     Config('force', 'false', r'''
         return success if the object does not exist''',
-        type='boolean'),
-    Config('remove_files', 'true', r'''
-        should the underlying files be removed?''',
         type='boolean'),
     Config('lock_wait', 'true', r'''
         wait for locks, if \c lock_wait=false, fail if any required locks are
         not available immediately''',
+        type='boolean'),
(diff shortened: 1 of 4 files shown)


dist/api_data.py
```diff
@@ -787,16 +787,20 @@ def __cmp__(self, other):
 ]),
 
 'WT_SESSION.drop' : Method([
+    Config('checkpoint_wait', 'true', r'''
+        do not wait for the chheckpoint lock, if \c checkpoint_wait=false,
```
sueloverso: typo

src/session/session_api.c
```diff
@@ -722,18 +722,29 @@ __wt_session_drop(WT_SESSION_IMPL *session, const char *uri, const char *cfg[])
...
 
 	if (!lock_wait)
 		F_SET(session, WT_SESSION_LOCK_NO_WAIT);
 
-	WT_WITH_CHECKPOINT_LOCK(session, ret,
-	    WT_WITH_SCHEMA_LOCK(session, ret,
-		WT_WITH_TABLE_LOCK(session, ret,
-		    ret = __wt_schema_drop(session, uri, cfg))));
+	/*
+	 * The checkpoint lock is here only as a correctness measure, we acquire
```
sueloverso: The comment is confusing to me.  If the lock is to avoid a spurious EBUSY, then it is not necessary for correctness.  Maybe "The checkpoint lock only is needed to avoid a spurious EBUSY error return."

michaelcahill:
@daveh86, can you please set `undoc=True` on both of those options?  This feels like something we should step back and reconsider in light of WT-1598 and I don't want to be tied to supporting these as public options.

daveh86:
@agorrod , mind taking a final look at this?

agorrod:

PR #2737, opened 2016-05-20
WT-2651: Coverity 1355591 resource leak

keithbostic:
@ddanderson, 20e0a57 is a new resource leak.

@agorrod, the rest is cleanup.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -497,8 +497,10 @@ config_opt(CONFIG *cfg, WT_CONFIG_ITEM *k, WT_CONFIG_ITEM *v)
 			free(*strp);
 			begin = &newstr[newlen - 1 - v->len];
 		}
-		if ((ret = config_unescape(begin)) != 0)
+		if ((ret = config_unescape(begin)) != 0) {
+			free(newstr);
 			return (ret);
+		}
 		*strp = newstr;
 		break;
 	case STRING_TYPE:
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -141,7 +141,7 @@ config_assign(CONFIG *dest, const CONFIG *src)
 
 	/* Clone the config string information into the new cfg object */
 	TAILQ_FOREACH(conf_line, &src->config_head, c) {
-		tmp_line = dcalloc(sizeof(CONFIG_QUEUE_ENTRY), 1);
+		tmp_line = dmalloc(sizeof(CONFIG_QUEUE_ENTRY));
 		tmp_line->string = dstrdup(conf_line->string);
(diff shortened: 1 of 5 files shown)


bench/wtperf/wtperf.c
```diff
@@ -1858,7 +1858,7 @@ create_uris(CONFIG *cfg)
 	base_uri_len = strlen(cfg->base_uri);
 	cfg->uris = dcalloc(cfg->table_count, sizeof(char *));
 	for (i = 0; i < cfg->table_count; i++) {
-		uri = cfg->uris[i] = dcalloc(base_uri_len + 5, 1);
+		uri = cfg->uris[i] = dmalloc(base_uri_len + 5);
 		/*
```
ddanderson: @keithbostic, this change to `dealloc` does not look safe. I think the memcpy code following depends on a null byte at the end of uri.

bench/wtperf/config.c
```diff
@@ -558,7 +560,7 @@ config_opt_file(CONFIG *cfg, const char *filename)
 		goto err;
 	}
 	buf_size = (size_t)sb.st_size;
-	file_buf = dcalloc(buf_size + 2, 1);
+	file_buf = dmalloc(buf_size + 2);
```
ddanderson: @keithbostic, it looks like this could be one less byte: `dmalloc(buf_size + 1)`

bench/wtperf/wtperf.c
```diff
@@ -1955,9 +1955,9 @@ start_all_runs(CONFIG *cfg)
 
 	home_len = strlen(cfg->home);
 	cmd_len = (home_len * 2) + 30; /* Add some slop. */
-	cmd_buf = dcalloc(cmd_len, 1);
+	cmd_buf = dmalloc(cmd_len);
 	for (i = 0; i < cfg->database_count; i++) {
-		next_cfg = dcalloc(1, sizeof(CONFIG));
+		next_cfg = dmalloc(sizeof(CONFIG));
```
ddanderson: @keithbostic, is this safe? Looks like `next_cfg` as used in `config_assign` needs to be cleared, else `config_free` will access `config_head` and other fields and be very confused.

keithbostic:
@ddanderson, you're correct, I stupidly didn't check for that. I've reverted that commit, it's not worth it.

ddanderson:
@keithbostic, looks good now.  I'll merge when the check has completed.

ddanderson:
retest this please

keithbostic:
@ddanderson, @agorrod, I've reworked wtperf so it compiles on Windows.

Don, would you please do the review?

ddanderson:
@keithbostic, LGTM.

PR #2739, opened 2016-05-23
WT-2629 Don't make stacks executable in assembly source.

michaelcahill:


michaelcahill:
Note that there is also a request to only include the PPC-specific files in PPC builds, I propose fixing the problem first (with the one-line change here), then figuring out how to deal with conditional source lists across automake plus two scons builds separately.

michaelcahill:
@keithbostic, can you please review?  This is literally inserting what GCC inserts into generated assembly code, as I understand it.

keithbostic:
This looks correct to me. Feel free to point me at the "only include the PPC-specific files in PPC builds" ticket if you want, I know where most of that stuff lives.

PR #2741, opened 2016-05-23
WT-2653: the custom file-system example should show device configuration

keithbostic:
@agorrod, one of the groups looking at the pluggable file system wanted to know how to pass local configuration information into their initialization function, so I added that to the example. I don't think we show it anywhere else, and it's pretty hard to figure out from the documentation.

While looking at that, I dropped core, and that's commit d687210. We could avoid this change by changing `__wt_ext_strerror()` to use the underlying session method instead of the connection's default, but that's even messier than what I'm doing here. I don't like that change much, hoping you or @michaelcahill have a better one.


michaelcahill:

PR #2742, opened 2016-05-24
WT-2629 Make the stack non-executable with GCC only.

michaelcahill:


michaelcahill:
@agorrod, can you please review?  This is blocking testing, so I'm keen to get it in.

agorrod:
@michaelcahill lgtm - I'll go ahead and merge.

PR #2743, opened 2016-05-24
WT-2656 - Fix GCC47 compiler complaints

daveh86:


daveh86:
Fixes the build complaints in http://build.wiredtiger.com:8080/job/wiredtiger-gcc47-warnings/4297

test/csuite/wt1965_col_efficiency/main.c
```diff
@@ -55,8 +55,7 @@ static void
 	TEST_OPTS *opts;
 	WT_CURSOR *cursor, *idx_cursor;
 	WT_SESSION *session;
-	uint64_t *obj_data, thr_idx, ts = g_ts;
-	int i, o, r;
+	uint64_t i, o, *obj_data, r, thr_idx, ts = g_ts;
```
agorrod: @daveh86 While you are changing this code could you untangle it please? Specifically:

* It's generally nicer not to have pointers and non-pointers in the same initialization line.
* Most of our code doesn't do initialization in the declaration; ``ts = g_ts`` should be split onto a different line.
* It's not clear what at least ``o`` and ``r`` mean from their names - it would be nice to give them more meaningful names.

agorrod:

PR #2744, opened 2016-05-24
WT-2659: csuite tests, assorted lint and cleanup.

keithbostic:
@sueloverso, @agorrod, I pointed lint at the new csuite tests, just some minor cleanup.

I added a new macro, `testutil_assert`, to make it easy to assert stuff and die if it's not true.


keithbostic:
One other thing, `TEST_OPTS.{conn_config,table_config}` are never set, but I assumed that was deliberate.

sueloverso:
@keithbostic LGTM.  Thanks.

PR #2747, opened 2016-05-25
WT-2660 Fix a hang between eviction and connection close

sueloverso:
@michaelcahill Please review this change.  I added a condition to have the app thread helping with eviction return if it notices we're closing.  You'll see by my first commit (reverted in the second) that I tried to encourage the situation, but have so far been unable to reproduce the original hang.

One consideration for this change is whether we could hit one of the conditionals in `conn_cache_destroy` but I think it is okay.  I just want someone else to think about it for a minute too.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -434,10 +434,6 @@ __wt_evict_destroy(WT_SESSION_IMPL *session)
 
 		conn->evict_session = NULL;
 	}
-	__wt_spin_lock(session, &cache->evict_queue_lock);
-	cache->evict_current = NULL;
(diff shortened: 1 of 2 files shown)


agorrod:
autobuild: test this please

michaelcahill:
@sueloverso, any reason not to put this inside `__wt_eviction_needed` (along with a comment explaining it)?

I think what we're doing here is trying to make sure that even if the cache is wedged, `WT_CONNECTION::close` can complete.  This seems like a reasonable solution to me.

sueloverso:
Thanks @michaelcahill .  I have fixed that and will merge when the tester completes.

PR #2748, opened 2016-05-25
WT-2662 For spell check, strip out double quote literals, they confuse aspell

ddanderson:


ddanderson:

keithbostic:
lgtm, merged.

PR #2749, opened 2016-05-25
WT-2268 WT-2597 JSON load/dump Unicode fixes

ddanderson:
This version of JSON load/dump is incompatible with older versions.  The issue is that older JSON dump tried to dump 's' and 'S' strings as Unicode.  If the byte sequence ... 0x03 0xC0 ... appeared, then this is a valid Unicode escape sequence and we showed '...\u03c0...'.  All fine if the string is Unicode, but if not (e.g. ... 0xFF 0xFE ...), there is no Unicode sequence that we can emit that will map back to these bytes when we reload. It seemed to me the simplest way to solve this was to treat every character in a string as a raw byte (that is, not interpreted as Unicode), but emit a unicode value between '\u0000' and '\u00ff'.  If the byte is printable, we show the printable character in the string.  Byte arrays (schema type 'u') are dumped using the same rules, except that we don't never should the printable character, always \u0000 to \u00ff.

Since this breaks compatibility (old JSON dumps may translate to different bytes in a new JSON loader, and vice versa), I added a version string and a check. A new JSON dump loaded by old code will also give an error (it won't expect to see the version string). I didn't make any attempt to try to load old dumps via some byte translation; it may be possible, but there may be ambiguities. I figured anyone really interested in moving data via dump/load would not use JSON to do it. The important thing is to catch mistakes in misusage.

Note: This also fixes WT-2597, I've added lots of new tests, including LSM.

### DIFF
--- a/src/cursor/cur_dump.c
+++ b/src/cursor/cur_dump.c
@@ -155,7 +155,9 @@ __curdump_set_key(WT_CURSOR *cursor, ...)
 	WT_SESSION_IMPL *session;
 	uint64_t recno;
 	va_list ap;
+	const uint8_t *up;
 	const char *p;
+	bool json;
 
 	cdump = (WT_CURSOR_DUMP *)cursor;
 	child = cdump->child;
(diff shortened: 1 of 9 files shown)


ddanderson:
@michaelcahill, can you review?

michaelcahill:
Thanks, @ddanderson, lgtm -- I'll merge.  At some point, I'd like to talk about the 'j', 'J' and 'K' formats to see if there's any way to avoid JSON-specific logic in the packing code...

PR #2753, opened 2016-05-27
WT-2551 Use tcmalloc overhead to adjust available cache.

michaelcahill:


michaelcahill:
Note that this is not ready for review / merge.

In particular, it needs more robust build support for detecting whether tcmalloc is enabled across autoconf and scons.  It also introduces significant stalls into some workloads (that currently grow the heap over the cache size).

This is at best only one part of a solution to the general problem of allocator overhead / fragmentation.

agorrod:
We've decided not to go down this path - it is too unpredictable from an application point of view.

PR #2755, opened 2016-05-29
WT-2671 dump more information about the file layout in verify debug mode

keithbostic:
WT-2671: dump more information about the file layout in verify debug mode

Dump the checkpoint final and correct sizes when loading checkpoints in verify dump_XXX mode.

Dump the checkpoint alloc, discard and avail lists when loading checkpoints in verify dump_XXX mode.

Fix a bug, convert the count of tree pages from uint32_t to uint64_t, there's no reason page counts can't overflow 4B.

Fix a bug, reset the internal/leaf page counts at the end of every checkpoint, we were doing cumulative counts.

Fix a bug, where we were using "%03zu" for a size_t type, fix the dist/s_style script to catch that case.

Add a divider line between checkpoints to make it easier to read the output.

Add a new __wt_buf_set_size() function that displays byte values in human-readable format.

Move __wt_buf_set_printable() from btree/bt_misc.c to support/scratch.c, it's a general-purpose function.

Rename verify debug option from "dump_shape" to "dump_layout", we're dumping more than the tree shape now.


keithbostic:
retest this please

michaelcahill:
@keithbostic I don't know if you wanted a review, but this lgtm.

PR #2756, opened 2016-05-30
SERVER-24306 Do log file prealloc every second.

michaelcahill:
With the switch to auto-tuning condition variables, logging threads
need to track elapsed time to determine when to do heavyweight
operations such as log file pre-allocation and archiving.


agorrod:
@michaelcahill would you like a review of this one, or are you waiting to talk to @sueloverso?

michaelcahill:
This doesn't fix the problem, replacing with #2761.

PR #2757, opened 2016-05-30

keithbostic:

I ended up with 3 macros instead of 2, because I couldn't think of a way to handle `rename` in the same macro as the rest of the errno-style calls.

I don't see anything much simpler than this that captures all of the possibilities, maybe you or @agorrod can.


src/include/os.h
```diff
@@ -6,18 +6,16 @@
...
-			break;						\
-		}							\
-		switch ((ret) = __wt_errno()) {				\
-		case 0:							\
-			/* The call failed but didn't set errno. */	\
-			(ret) = WT_ERROR;				\
-			break;						\
+/*
+ * System call macro to retry system/library calls on "expected" errors.
+ * Three versions: one of which looks for a -1 return value on error and
```
sueloverso: I'd prefer to see this comment formatted as a numbered or bulleted list of the three versions rather than a dense paragraph.

keithbostic: > I'd prefer to see this comment formatted as a numbered or bulleted list of the three versions rather than a dense paragraph.

Sure, pushed.

michaelcahill:
closing because we took a newer version.

PR #2758, opened 2016-05-30
WT-2658 Only include PPC-specific files in PPC builds

keithbostic:
@michaelcahill, this isn't tested on a PPC yet, but I think it's mostly there, you might want to do a quick skim to make sure it's what you had in mind.


keithbostic:
@michaelcahill, @agorrod, I think this one is ready for review/merge.

build_posix/Make.subdirs
```diff
@@ -20,8 +21,8 @@ ext/test/kvs_bdb HAVE_BERKELEY_DB
 api/leveldb LEVELDB
 bench/wtperf
 examples/c
-lang/java JAVA
 examples/java JAVA
+lang/java JAVA
```
michaelcahill: The ordering here matters: we can't build the Java examples until the Java language support is built.

dist/s_win
```diff
@@ -67,7 +67,7 @@ win_filelist()
 	echo 'src/os_win/os_thread.c'
 	echo 'src/os_win/os_time.c'
 	echo 'src/os_win/os_vsnprintf.c'
-	echo 'src/os_win/os_yield.c') < filelist | sort > $t
+	echo 'src/os_win/os_yield.c') | sort > $t
```
michaelcahill: How do you feel about marking the POSIX-only files in `dist/filelist` with a condition of `POSIX` and including the Windows-specific files with a condition of `WINDOWS` so we don't need `build_win/filelist.win`?

dist/dist.py
```diff
@@ -2,21 +2,18 @@
...
-def source_files(skip_includes=False):
-    if not skip_includes:
-        for line in glob.iglob('../src/include/*.[hi]'):
-            yield line
+def source_files():
     file_re = re.compile(r'^\w')
     for line in open('filelist', 'r'):
         if file_re.match(line):
-            yield os.path.join('..', line.rstrip())
+            yield os.path.join('..', line.split(None, 1)[0])
```
michaelcahill: I think these can just be `list.split()[0]` -- no need to pass anything to split?

build_posix/Make.subdirs
keithbostic: Sorry; I've reverted the change.

dist/dist.py
```diff
@@ -2,21 +2,18 @@
 
 # source_files --
 #    Return a list of the WiredTiger source file names.
-def source_files(skip_includes=False):
-    if not skip_includes:
-        for line in glob.iglob('../src/include/*.[hi]'):
-            yield line
```
michaelcahill: It looks like we only use this in `style.py` -- it still needs to pick up the header / inline files, I think.

michaelcahill:
Thanks, @keithbostic -- I've made a few minor comments, but this lgtm.

dist/s_win
keithbostic: > How do you feel about marking the POSIX-only files in `dist/filelist` with a condition of `POSIX` and including the Windows-specific files with a condition of `WINDOWS` so we don't need `build_win/filelist.win`?

I'd vote no, because it would complicate the dist shell scripts, they'd now have to select/not-select lines from `dist/filelist` based on the POSIX/WINDOWS conditions.

Is there a way to avoid that additional complexity?

dist/dist.py
keithbostic: > I think these can just be `list.split()[0]` -- no need to pass anything to split?

Thanks, fixed.

dist/dist.py
keithbostic: Sorry, you're right -- fixed.

keithbostic:
@michaelcahill, I've pushed fixes for your review comments, I think this one is ready to merge.

keithbostic:
@michaelcahill, @markbenvenuto: merging the Windows and POSIX file lists turned out to be easier than I expected, d1cbb5c is a set of changes for your consideration. (Feel free to revert it if you don't like this change for any reason.)

Obviously, this requires minor changes to [issue 71130001](https://mongodbcr.appspot.com/71130001/).

markbenvenuto:
@keithbostic `env['TARGET_ARCH'] == 'ppc64le'` will always return false because `TARGET_ARCH` is something the MongoDB Scons build system sets, but I am happy with his change regardless. It is eloquent change.

I verified it works on the Windows build.

sueloverso:
Let me know when this is ready for me to test on my Mac.  I'm still hand-editing all my other branches.

keithbostic:
@markbenvenuto, thank you (BTW, you don't need to test Windows independently any more, David fixed things so we test Windows builds as part of every branch submit).

@michaelcahill, I guess I'd be inclined to simply set POSIX_HOST and POWERPC_HOST to false in the WiredTiger SConstruct file, then, the same way I set POSIX_HOST and WINDOWS_HOST for automake.

keithbostic:
@michaelcahill, I updated this change to fix SConstruct based on @markbenvenuto's review. I think this should be ready to merge into develop (but still requires a similar SConstruct change for MongoDB before a drop).

sueloverso:
@keithbostic This branch builds on my Mac.

michaelcahill:
Thanks @keithbostic, lgtm and I've merged the changes for SERVER-24334 so this shouldn't break any automated tests.  I'll merge.

PR #2760, opened 2016-05-31
WT-2673 Stop automatically increasing memory page max

agorrod:


agorrod:

src/btree/bt_handle.c
```diff
@@ -689,21 +689,21 @@ __btree_page_sizes(WT_SESSION_IMPL *session)
...
+	btree->maxmempage = cval.val;
 	if (!F_ISSET(conn, WT_CONN_CACHE_POOL)) {
-		if ((cache_size = conn->cache_size) > 0)
+		if ((cache_size = conn->cache_size) > 0) {
+			btree->maxmempage =
+			    WT_MIN(btree->maxmempage, cache_size / 10);
+			/* Use a lower bound of a single disk leaf page */
 			btree->maxmempage =
-			    WT_MIN(btree->maxmempage, cache_size / 4);
+			    WT_MAX(btree->maxmempage, btree->maxleafpage);
```
michaelcahill: @agorrod, I don't disagree with the change, but shouldn't we apply the `maxleafpage` lower bound outside this condition about cache size?

michaelcahill:
Other than the ordering issue I mentioned, this change lgtm.

agorrod:
Thanks @michaelcahill I've made the change you suggested. I'm going to go ahead and merge.

src/btree/bt_handle.c
```diff
@@ -689,22 +689,20 @@ __btree_page_sizes(WT_SESSION_IMPL *session)
 		    "size (%" PRIu32 "B)", btree->allocsize);
 
 	/*
-	 * When a page is forced to split, we want at least 50 entries on its
-	 * parent.
-	 *
 	 * Don't let pages grow larger than a quarter of the cache, with too-
```
dmitri-shubin: It seems to be one tenth now, not quarter.

agorrod: Thanks - fixed.

PR #2761, opened 2016-05-31
SERVER-24306 Fix stall in log_flush switching to new files.

michaelcahill:


michaelcahill:
@sueloverso, can you please review?

I'm open to better solutions (maybe we can make `write_lsn` switch into the new file somehow?), but the workload in the ticket demonstrates long pauses without this change.  This is now a 3.2.7 blocker.

Thanks!

sueloverso:
Using the `log_close_lsn` is the right value to use to wait for the flush.  It represents the end of the previous log at that point.  We could potentially rename it since it has expanded beyond its original use.  I pushed a minor KNF I noticed as well.  I will merge.

PR #2764, opened 2016-06-01
WT-2677 Fix JSON output so only printable ASCII is produced.

ddanderson:
The root cause is a sign extension error that shows up on systems
that have a signed 'char'.


ddanderson:

ddanderson:
BTW, this change has been tested on Solaris for the failing test.

keithbostic:
@ddanderson, lgtm, I've merged this one.

PR #2765, opened 2016-06-01

keithbostic:
@michaelcahill, I tried coding up your suggestion, and I agree it's better than #2757.

I think we should review how we're using SYSCALL_RETRY, though. We're calling it for things that aren't worth retrying (gettimeofday), and we're not calling it for things that might reasonably need to be retried (pread, pwrite). I could see using retry for open(2) calls, but I don't think it has much value outside of that, since pread/pwrite clearly don't need it.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -442,6 +442,7 @@ bzDecompressInit
 bzalloc
 bzfree
 bzip
+call's
 calloc
 cas
 catfmt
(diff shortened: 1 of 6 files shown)


michaelcahill:
Thanks, @keithbostic -- this lgtm.  I agree that it makes sense to split `WT_SYSCALL_RETRY` into two parts: one that gets the best error code for a given call and another than retries in the cases where that makes sense.

I'd forgotten we retry for cases like `ENOSPC`, so it may not be as simple as saying `pwrite` should never retry.

PR #2766, opened 2016-06-01
SERVER-23659 Improve log error message at startup.

sueloverso:
@michaelcahill Please review this change to improve the error message if there's an error reading a log record at startup.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1090,6 +1090,7 @@ __wt_log_open(WT_SESSION_IMPL *session)
 	logcount = 0;
 	lastlog = 0;
 	firstlog = UINT32_MAX;
+	F_SET(log, WT_LOG_NOT_VERIFIED);
 
 	/*
 	 * Open up a file handle to the log directory if we haven't.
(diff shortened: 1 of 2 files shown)


src/log/log.c
```diff
@@ -1744,6 +1745,22 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
...
+		__wt_errx(session,
+		    "WiredTiger is unable to read the recovery log.");
+		__wt_errx(session, "This may be due to the log"
+		    " files being encrypted, being from an older"
+		    " version or due to corruption on disk");
+		__wt_errx(session, "You should confirm that you have"
+		    " opened the database with the correct options including"
+		    " all encryption and compression options");
+	}
+	F_CLR(log, WT_LOG_NOT_VERIFIED);
```
michaelcahill: Thanks, @sueloverso -- this lgtm.  It isn't an issue for now because `\_\_wt_log_scan` isn't in a performance path, but consider only clearing the flag if it is set, otherwise all callers will unconditionally update a shared cache line.

PR #2768, opened 2016-06-02
WT-2646 Add checkpoint_wait configuration option to drop (#2736)

daveh86:
* Default checkpoint_wait is true. This change is useful because it means concurrent create/drop calls don't generate EBUSY returns.
* Mark lock_wait and checkpoint_wait as undoc

(cherry picked from commit 4b48ad6fb787a8b8662b5eacad6840d0b3ae9fb5)

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -787,15 +787,19 @@ methods = {
 ]),
 
 'WT_SESSION.drop' : Method([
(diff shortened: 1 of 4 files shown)


daveh86:
@agorrod, can you take a look at this backport?

agorrod:

PR #2769, opened 2016-06-02

keithbostic:
@michaelcahill, in #2765  you said:

> I agree that it makes sense to split WT_SYSCALL_RETRY into two parts: one that gets the best error code for a given call and another than retries in the cases where that makes sense.

This is that change; I left retry in for calls that allocate file descriptors (opendir, open), and that write the file system and were already retrying (fallocate, fcntl(F_FULLFSYNC), fsync, ftruncate, sync_file_range), but didn't add retry semantics to calls that write the file system but weren't already retrying (pread, pwrite).

The only borderline case where I potentially changed things was that I don't retry file-descriptor close calls: technically, they can write the filesystem, but the fsync should have already been done (if we care about the durability of the file).

You also said:

> I'd forgotten we retry for cases like ENOSPC, so it may not be as simple as saying pwrite should never retry.

It doesn't seem unreasonable to me for pread/pwrite to retry, particularly on EIO, and in the case of pwrite, on ENOSPC. Should I make that change?

### DIFF
--- a/src/os_posix/os_thread.c
+++ b/src/os_posix/os_thread.c
@@ -19,7 +19,7 @@ __wt_thread_create(WT_SESSION_IMPL *session,
 	WT_DECL_RET;
 
 	/* Spawn a new thread of control. */
-	WT_SYSCALL(pthread_create(tidret, NULL, func, arg), ret);
+	WT_SYSCALL_RETRY(pthread_create(tidret, NULL, func, arg), ret);
 	if (ret == 0)
 		return (0);
 	WT_RET_MSG(session, ret, "pthread_create");
--- a/src/os_posix/os_thread.c
+++ b/src/os_posix/os_thread.c
@@ -19,7 +19,7 @@ __wt_thread_create(WT_SESSION_IMPL *session,
 	WT_DECL_RET;
 
 	/* Spawn a new thread of control. */
-	WT_SYSCALL_RETRY(pthread_create(tidret, NULL, func, arg), ret);
+	WT_SYSCALL(pthread_create(tidret, NULL, func, arg), ret);
 	if (ret == 0)
 		return (0);
 	WT_RET_MSG(session, ret, "pthread_create");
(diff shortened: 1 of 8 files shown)


keithbostic:
One other question: the only purpose of WT_SYSCALL at this point is to do the error mapping for calls where we don't know what they might return (for example, posix_madvise), and of course, WT_SYSCALL_RETRY uses it, because that way WT_SYSCALL_RETRY works for any system call we want to retry.

But, we use WT_SYSCALL in lots of places (for example, clock_gettime, gettimeofday), where we know what the error returns are going to be, there's no need for any error mapping, and, of course, we don't use WT_SYSCALL for error handling in lots of places because (I'm guessing), we never wanted to retry those calls (for example, dlopen, dlsym and dlclose in os_dlopen.c), so we inlined the error handling.

We could (1) rewrite all of the calls to use WT_SYSCALL, we could (2) rewrite most current users of WT_SYSCALL to inline error handling unless there's actually a question about what the error return might be, or (3) leave it all alone.

I don't feel strongly; I don't think (1) improves things, (2) makes error handling more consistent and we'd have some clarity about what we're really trying to do the next time this question comes up, (3) skips making a low-value change.

Thoughts?

michaelcahill:
@keithbostic, if we're going to do anything, I think it should be (1).  Spreading around this logic in more places and making strong assumptions that all calls conform to POSIX seems dangerous to me.

In the specific case of {{dlopen}}, error handling is only documented via {{dlerror}}, which we should keep using when generating error messages.  I'm not sure `WT_SYSCALL` would buy us anything because the only call that returns an integer is {{dlclose}}, and it doesn't set {{errno}}.

src/os_posix/os_thread.c
```diff
@@ -19,7 +19,7 @@ __wt_thread_create(WT_SESSION_IMPL *session,
 	WT_DECL_RET;
 
 	/* Spawn a new thread of control. */
-	WT_SYSCALL_RETRY(pthread_create(tidret, NULL, func, arg), ret);
```
michaelcahill: It looks like `pthread_create` should retry to deal with `EAGAIN`.

michaelcahill:
With the one nit mentioned above, this lgtm.

src/os_posix/os_thread.c
keithbostic: > It looks like pthread_create should retry to deal with EAGAIN.
 
Agreed, I'll push that fix and merge.

keithbostic:
@michaelcahill, I see your point.

The one thing left in this chain is if pread/pwrite should retry?

PR #2770, opened 2016-06-02
WT-2664 Have callers of evict_exclusive_on clear the dhandle directly.

sueloverso:
@michaelcahill I did the work for clear walks on a sub-branch for you to review.  We can either keep or toss these changesw.  The good news is that this branch is on-par with develop for all the eviction performance combinations we've been testing - generally within 1%.  I never see clear-walk related functions in profiler output.  But this cleans up code and gets rid of a few functions.


src/evict/evict_lru.c
```diff
@@ -594,12 +594,8 @@ __evict_pass(WT_SESSION_IMPL *session)
 		 * If there is a request to clear eviction walks, do that now,
 		 * before checking if the cache is full.
 		 */
-		if (F_ISSET(cache, WT_CACHE_CLEAR_WALKS)) {
-			F_CLR(cache, WT_CACHE_CLEAR_WALKS);
-			WT_RET(__evict_clear_walks(session));
-			WT_RET(__wt_cond_signal(
-			    session, cache->evict_waiter_cond));
-		}
+		if (F_ISSET(cache, WT_CACHE_CLEAR_WALKS))
```
michaelcahill: Can't we get rid of this flag as well?

sueloverso: It is used now to cause the code to know a thread wants to clear its walks so the thread doing the populate can stop as soon as possible.  Otherwise a thread may way for an entire populate and evict pass to complete.  That seems like it could cause stalls (although I haven't measured that).

src/evict/evict_lru.c
```diff
@@ -841,10 +780,15 @@ __wt_evict_file_exclusive_on(WT_SESSION_IMPL *session)
...
 	 */
 	F_SET(btree, WT_BTREE_NO_EVICTION);
+	F_SET(cache, WT_CACHE_CLEAR_WALKS);
 	WT_FULL_BARRIER();
 
 	/* Clear any existing LRU eviction walk for the file. */
-	WT_ERR(__evict_request_clear_walk(session));
+	__wt_spin_lock(session, &cache->evict_pass_lock);
+	ret = __evict_clear_walk(session);
+	__wt_spin_unlock(session, &cache->evict_pass_lock);
```
michaelcahill: Can you please make a `WT_WITH_EVICT_PASS_LOCK` macro that uses `WT_WITH_LOCK` to acquire `evict_pass_lock` and set a session flag?  We should be able to assert in `__evict_clear_walk` that we are holding the lock and at the moment we can't.

src/evict/evict_lru.c
michaelcahill: OK, for now let's rename it to something like `WT_CACHE_PASS_INTERRUPT` or `STOP` or similar.

src/evict/evict_lru.c
michaelcahill: If it's too much work to restructure those macros to support trylock semantics, that's fine -- we can merge this and do that work separately.  I'm a little wary of repeating the "op" 3 times in `WT_WITH_LOCK`.

michaelcahill:
@sueloverso, can you please point the perf branch here, otherwise I'm happy for this to be merged.  It's certainly good to get rid of all of that inter-thread communication code.

PR #2771, opened 2016-06-02
WT-2677 Fix JSON output so only printable ASCII is produced.

ddanderson:
(Reopening this pull request - @keithbostic saw a compilation error using gcc 4.7, I didn't see that error)

The root cause is a sign extension error that shows up on systems
that have a signed 'char'.

### DIFF
--- a/dist/log.py
+++ b/dist/log.py
@@ -178,7 +178,7 @@ __wt_logop_read(WT_SESSION_IMPL *session,
 }
 
 static size_t
-__logrec_json_unpack_str(char *dest, size_t destlen, const char *src,
+__logrec_json_unpack_str(char *dest, size_t destlen, const u_char *src,
     size_t srclen)
 {
 \tsize_t total;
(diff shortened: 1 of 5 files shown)


ddanderson:
@keithbostic, for the record, here's my gcc line using Amazon Linux with gcc 4.7:

```gcc -DHAVE_CONFIG_H -I. -I./api/leveldb -I./src/include -D_GNU_SOURCE -m64 -MT src/cursor/cur_json.lo -MD -MP -MF src/cursor/.deps/cur_json.Tpo -c src/cursor/cur_json.c  -fPIC -DPIC -o src/cursor/.libs/cur_json.o```

keithbostic:
@ddanderson, my guess is you'll need to add `-Wall`at least.

michaelcahill:
Isn't the problem here that `isprint` is undefined for values outside `[-1, 127]`?

keithbostic:
> Isn't the problem here that isprint is undefined for values outside [-1, 127]?

My understanding is `isprint` is defined for anything that fits into an `unsigned char` or is equal to EOF, but those are not simple terms. The compiler is converting the `signed char` to an `int` before passing it to `isprint`, of course, which is how we're going negative, but I don't know why the `char` is > 127 at that point.

ddanderson:
>  but I don't know why the char is > 127 at that point.
The test case is contrived to put char values from 0x01 to 0xFF into the string.

@keithbostic, the latest changes pass tests for compilers I've tried.  I didn't have -Wsign-conversion on before.  Can you review again please?

keithbostic:
lgtm, I'll merge.

PR #2772, opened 2016-06-02
WT-2589 Clear the cursor statistics for LAS when 'clear' is set.

sueloverso:
@keithbostic Please review this change.  I tried a handful of wtperf configs with statistics turned on but didn't find any that set the LAS stats.  If you have a workload, let me know.


sueloverso:
I got a test/format workload that shows the problem on develop and that this fixes it.

PR #2773, opened 2016-06-03
WT-2682 add option to configure WiredTiger with strict compiler flags

keithbostic:
@agorrod, let me know what you think -- currently, it just supports gcc 4.7, but the framework is there.


michaelcahill:
Can we check each possible warning to see if the compiler supports it, as per http://www.gnu.org/software/autoconf-archive/ax_cflags_warn_all.html (and MongoDB's scons)?

Also, clang has the handle `-Weverything` which turns on *all* warnings.  We can build with something like this: `CFLAGS=-Weverything Wno-conversion -Wno-padded -Wno-zero-length-array -Wno-cast-align -Wno-cast-qual -Wno-packed`

keithbostic:
> Can we check each possible warning to see if the compiler supports it

I may be missing something, but as soon as I turn on any gcc standard (for example, `-std=c11`), the compile just breaks, with complaints about FreeBSD system headers and so on.

> clang has the parameter

Agreed, that's my plan.

keithbostic:
@michaelcahill, @agorrod, I've moved the list of compiler warnings from `configure.ac.in` into a separate file, `aclocal/strict.m4`. I would rather we simply loaded them from some flat-text file somewhere as Alex suggested, but I don't see a simple way to make that work, and I don't want a new syntax to configure, I want `--enable-strict` to just work.

With this set of changes, I get clean compiles on gcc 4.7, gcc 5.3, clang 3.4 and clang 3.8 (all on my FreeBSD box).

keithbostic:
This branch builds cleanly on OS X as well, `Apple LLVM version 7.3.0 (clang-703.0.31)`.

keithbostic:
At some point, we might want to have a conversation about `-Wno-cast-align`: we could fix that, but it would require taking `const` more seriously than we do.

agorrod:
@sueloverso Would you mind trying out the build with your tool chain? Specifically try building with `../configure --enable-strict` on the machines you do builds on?

agorrod:
I should have been more concrete - if this change works for @sueloverso I'm happy for it to be merged. I've done a review of the changes and they LGTM.

keithbostic:
What should we do about clang's `-Wcast-qual`? I think we can:
* turn it off for now,
* Change `WT_PAGE.dsk` to non-const and fix the warnings,
* Drive the `WT_PAGE.dsk` const work through the system (but that isn't quite possible).

sueloverso:
When I run on my Mac, I get:
```
make  all-recursive
cc1: error: unrecognized command line option "-Wdouble-promotion"
cc1: error: unrecognized command line option "-Wjump-misses-init"
cc1: error: unrecognized command line option "-Wmissing-parameter-type"
cc1: error: unrecognized command line option "-Wsign-conversion"
make[2]: *** [reverse_collator.lo] Error 1
```

sueloverso:
While the compile works (as in `--enable-strict` doesn't produce command line errors) on PPC, it does generate a code error message:
```
../src/checksum/checksum.c:1111:1: error: '__wt_cksum_sw' defined but not used [-Werror=unused-function]
 __wt_cksum_sw(const void *chunk, size_t len)
 ^
cc1: all warnings being treated as errors
```

keithbostic:
@sueloverso, what compiler are you using on OS X, what's the output of `cc --version`?

sueloverso:
```
evict 1918 >> cc --version
Apple clang version 4.1 (tags/Apple/clang-421.11.66) (based on LLVM 3.1svn)
Target: x86_64-apple-darwin12.4.0
Thread model: posix
```

keithbostic:
@agorrod, @sueloverso, that's a 3-year-old release of OS X and a 4-year-old release of LLVM, absent a reason, I won't bother fixing --enable-strict to handle it.

(That said, it's just another line in the `configure.ac.in` file and another clause in `build_posix/aclocal/strict.m4`, it wouldn't be hard to support that compiler.)

sueloverso:
Removing those three lines from `aclocal/strict.m4` on my Mac allows it to proceed to build.  It quickly fails with this warning/error:
```
In file included from ../src/include/wt_internal.h:340,
                 from ../src/async/async_api.c:9:
../src/include/btmem.h:817: warning: packed attribute is unnecessary for ‘recno’
../src/include/btmem.h:818: warning: packed attribute is unnecessary for ‘rle’
../src/include/btmem.h:819: warning: packed attribute is unnecessary for ‘indx’
../src/include/btmem.h:885: warning: packed attribute is unnecessary for ‘txnid’
../src/include/btmem.h:887: warning: packed attribute is unnecessary for ‘next’
../src/include/btmem.h:897: warning: packed attribute is unnecessary for ‘size’
```

keithbostic:
@sueloverso, I was about to go fix this branch to build on your system, when I realized, you're building with the gcc flags, not the clang flags.

I don't see how that's possible, it seems to me this code from build_posix/configure.ac.in should be matching on "clang" from your compiler's output for the `--version` flag, I don't see how it's getting to the gcc flags.
```
# If enable-strict is configured, turn on as much error checking as we can for
# this compiler. Intended for developers, and only works for gcc/clang, but it
# fills a need.
if test "$wt_cv_enable_strict" = "yes"; then
        wt_cv_cc_version="`$CC --version | sed -eq`"
        case "$wt_cv_cc_version" in
        *clang*)
                AM_CLANG_WARNINGS;;
        *gcc*)
                AM_GCC_WARNINGS;;
        *)
                AC_MSG_ERROR(
                    [--enable-strict does not support "$wt_cv_cc_version".]);;
        esac
```
Can we please sync up tomorrow, and figure this out?

sueloverso:
@keithbostic Ah, the `r` script is doing this: `{$CC:-gcc}` and I don't have `CC` set in my environment, so it is taking the gcc.  My gcc is 4.2.1.

keithbostic:
>  My gcc is 4.2.1.

GCC 4.2.1 was released in 2007, maybe switch to clang on your laptop?

build_posix/aclocal/options.m4
```diff
@@ -215,6 +215,18 @@ pthread_adaptive|pthreads_adaptive)
...
 AC_MSG_RESULT($with_spinlock)
 
+AH_TEMPLATE(HAVE_STRICT, [Enable strict compiler checking.])
+AC_MSG_CHECKING(if --enable-strict option specified)
+AC_ARG_ENABLE(strict,
+	[AS_HELP_STRING([--enable-strict],
+	    [Enable strict compiler checking.])], r=$enableval, r=no)
+case "$r" in
+no)	wt_cv_enable_strict=no;;
+*)	AC_DEFINE(HAVE_STRICT)
```
michaelcahill: Can we do without a macro?  AFAICT, it isn't currently used and I'd be unhappy with having any code that we add just to fix the strict build.

keithbostic: @michaelcahill:
> Can we do without a macro?

Yes, there's no need for one, I just cut-and-pasted without thinking it through. I've removed it and pushed the change.

sueloverso:
@keithbostic AFAICT, I'm getting `gcc` from `configure.ac.in:33`.  I have a script that does `env CC="clang"...` and that seems to be working, in that the build is failing quickly with warning/errors.  On `develop` my clang build spits out tons of warnings, but ultimately builds.

keithbostic:
@agorrod, @michaelcahill, --enable-strict now builds cleanly on Sue's machine (so, we're supporting an older OS X variant of clang), but we didn't make her GCC 4.2 work.

I think this one is ready to merge if you are happy with it.

agorrod:
@keithbostic This change lgtm - I'm going to merge and watch for fallout.

PR #2775, opened 2016-06-03
WT-2676 Don't use key size in column store in-memory splits.

michaelcahill:


michaelcahill:
@keithbostic, can you please review this one?

keithbostic:
lgtm, merged.

PR #2779, opened 2016-06-06
WT-2686 Report an error any time we fail to scan the log.

michaelcahill:
Don't use a flag on the log structure: we may not have one if logging
is disabled, but we still scan log files if they exist.

### DIFF
--- a/src/include/log.h
+++ b/src/include/log.h
@@ -256,9 +256,6 @@ struct __wt_log {
 #ifdef HAVE_DIAGNOSTIC
 	uint64_t	 write_calls;		/* Calls to log_write */
 #endif
-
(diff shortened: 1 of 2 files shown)


michaelcahill:
@sueloverso, can you please take this one?

It fixes the core dump, I guess the question is whether it could result in the message appearing more than once, or in some other situation where the original version would not.  I think it's okay, because we'll fail very quickly if we can't read the log, but please take a look.

michaelcahill:
@sueloverso, I'm going to merge because tests are now failing.  I'd still appreciate a review when you get a chance.

Thanks!

sueloverso:
I have reviewed this change and agree.  While the longer error message would be printed for the first read attempt of every call to `log_scan` I don't think we'll ever call it a second time if the first time returns an error.

PR #2780, opened 2016-06-06
WT-2682 fix --enable-strict warnings when building on PPC.

keithbostic:
@agorrod, @sueloverso, this should make --enable-strict compile cleanly on the PPC.

### DIFF
--- a/src/checksum/checksum.c
+++ b/src/checksum/checksum.c
@@ -1103,6 +1103,7 @@ static const uint32_t g_crc_slicing[8][256] = {
 #endif
 };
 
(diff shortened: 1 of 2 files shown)


keithbostic:
@michaelcahill, the CPP macro in checksum/checksum.c is getting a little convoluted since we broke out the power8 code. Should we split out the x86 code into its own directory, or is it not worth the bother?

keithbostic:
autobuild retest this please

michaelcahill:
@keithbostic, I don't feel strongly -- probably not worth it because I don't expect ongoing development of the checksum code.

PR #2781, opened 2016-06-06
WT-2685 Fix clear walk hazard pointer failure.

sueloverso:
Add missing WT_WITH_PASS_LOCK with call to clear walks on shutdown.
Add an assertion in evict_clear_walk to confirm the lock is held.


sueloverso:
@michaelcahill or @agorrod Please review this change.  The real fix is that the `WT_WITH_PASS_LOCK` was missing from the final call to clear walks on shutdown.  I also added an assertion in clear walks that the lock is held which meant I had to set the flag in `evict_thread_run` because it uses the `walk_session` at a much lower level.

michaelcahill:

PR #2782, opened 2016-06-06
WT-2689 Fix heap-use-after-free on cursor error path.

sueloverso:
@agorrod Please review this.  This fixes the bug reported.  However, I went through the error path code of all `cur_*.c` for `open_cursor` and there is a lot of variation.  Some just free, some just call `wt_cursor_close`, some set `*cursorp = NULL` and some do not.  I'm not convinced in `cur_table.c` that the conditional for `*cursorp` is needed at all.

### DIFF
--- a/src/cursor/cur_table.c
+++ b/src/cursor/cur_table.c
@@ -972,7 +972,8 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 
 	if (0) {
 err:		if (*cursorp != NULL) {
-			WT_TRET(__wt_cursor_close(*cursorp));
+			if (*cursorp != cursor)
+				WT_TRET(__wt_cursor_close(*cursorp));
 			*cursorp = NULL;
 		}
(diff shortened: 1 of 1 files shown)


src/cursor/cur_table.c
```diff
@@ -972,7 +972,8 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 
 	if (0) {
 err:		if (*cursorp != NULL) {
-			WT_TRET(__wt_cursor_close(*cursorp));
+			if (*cursorp != cursor)
+				WT_TRET(__wt_cursor_close(*cursorp));
```
agorrod: @sueloverso The cleanup path is certainly convoluted. Is it true that the only time we'd expect this close call to happen is when opening a dump cursor?

sueloverso: Yes, the only time on any call to `wt_cursor_init` that they're not equal is when opening a dump cursor.

agorrod:

PR #2783, opened 2016-06-07
WT-2691 remove/disallow 'char' declarations without 'unsigned'.

ddanderson:
Also added casts so that -Wsign-conversion can be used on clang/OSX.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -54,9 +54,9 @@ static void config_opt_usage(void);
 static int
 config_unescape(char *orig)
 {
-	char ch, *dst, *s;
+	u_char *dst, *s, ch;
 
-	for (dst = s = orig; *s != '\0';) {
+	for (dst = s = (u_char *)orig; *s != '\0';) {
 		if ((ch = *s++) == '\\') {
 			ch = *s++;
 			switch (ch) {
--- a/dist/s_style
+++ b/dist/s_style
@@ -138,6 +138,14 @@ else
 		}
 	fi
 
+	# Plain char declarations.
+	egrep '\<char[[:space:]]+[[:alnum:]_]+[,;)]' $f |
+	    sed -e '/unsigned char/d' > $t
+	test -s $t && {
+	    echo "$f: type "char" declaration, undefined if signed"
+	    cat $t
+	}
+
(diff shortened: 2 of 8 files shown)


ddanderson:
@keithbostic, this was slightly more disruptive than I'd hoped. I think it's worth it, what do you think?

I modified your egrep slightly, it should detect char arguments, but made no attempt to detect char x[10].

keithbostic:
Thanks, @ddanderson.

I agree it's more disruptive than expected, but given the difficulty with detecting these errors, I'm inclined toward taking this change.

@michaelcahill, @agorrod, what do you think?

I tried futzing with gcc's `-fsigned-char` and `-funsigned-char` flags to see if I could get better error reporting, but I didn't find anything that helped. That said, we could change `--enable-strict` to set the `-fsigned-char` flag on gcc, it seems to me running tests with `-fsigned-char` should enable us to see these failures sooner?

michaelcahill:
@keithbostic, @ddanderson -- I tried similar compiler flag combinations on GCC and clang and couldn't find anything that generated a warning for the original bug.

To be clear, what I believe is problematic is the conversion from ambiguous `char` to `int`, where values greater than unsigned 127 may or may not map to negative `int`s.  The issue is that from the compiler's point of view, there is no possible loss of information, since regardless of whether `char` is signed, `int` can hold the full range of possible values.

At the moment, the rule we are enforcing here seems somewhat arbitrary to me: as noted in the ticket, it's still easy to call `isprint` with a signed char and get undefined behavior.  If the proposal was to not allow bare calls to the functions in `ctype.h` and instead provide inline wrappers that take `u_char` args, I'd have more confidence that we're really avoiding the issue in future.

keithbostic:
I guess I agree with the wrapper approach -- I don't have anything better to suggest.

There are roughly 20 "portable" ctype functions, I can't think of anything better than checking for all of them in `dist/s_style`.

It's tempting to disallow including `<ctype.h>` in any source files except the one wrapping the ctype functions we're actually using, but that means we don't inline the functions any more, which might hit performance?

ddanderson:
I'm going to close this pull request, as I've implemented the wrapper approach, see #2818.

PR #2784, opened 2016-06-07
WT-2319 Add stats to measure count and duration of the sync calls made for log and checkpoint modules

sulabhM:


wtbuild:

agorrod:
Autobuild: test this please

sulabhM:
@daveh86  Can you review the changes I made?

daveh86:
@sulabhM, this LGTM

dist/stat_data.py
```diff
@@ -285,6 +287,10 @@ def __init__(self, name, desc, flags=''):
     TxnStat('txn_snapshots_created', 'number of named snapshots created'),
     TxnStat('txn_snapshots_dropped', 'number of named snapshots dropped'),
     TxnStat('txn_sync', 'transaction sync calls'),
+    TxnStat('txn_fsync_pre', 'transaction fsync calls before creating a transaction id'),
```
agorrod: These should be added in the list in alphabetical order. I'm also inclined to rename them so they sort alphabetically how you want, i.e change `txn_fsync_duration_pre` to be `txn_fsync_pre_duration` similarly for post.

Minor phrasing change, I'd like to switch from:
``calls before creating a transaction id``
To:
``calls for checkpoint before allocating the transaction ID``

Similarly for the other descriptions as well.

src/log/log.c
```diff
@@ -95,11 +95,14 @@ __wt_log_background(WT_SESSION_IMPL *session, WT_LSN *lsn)
...
 __wt_log_force_sync(WT_SESSION_IMPL *session, WT_LSN *min_lsn)
 {
+	struct timespec fsync_start, fsync_stop;
 	WT_DECL_RET;
 	WT_FH *log_fh;
 	WT_LOG *log;
+	uint64_t fsync_duration_usecs;
 
 	log = S2C(session)->log;
+	fsync_duration_usecs = 0;
```
agorrod: We generally don't initialize variables unless we need to. Also variable initialization blocks are usually alphabeticalized (i.e: the order would be swapped here, so `log` followed `fsync..`

dist/stat_data.py
```diff
@@ -285,6 +287,10 @@ def __init__(self, name, desc, flags=''):
     TxnStat('txn_snapshots_created', 'number of named snapshots created'),
     TxnStat('txn_snapshots_dropped', 'number of named snapshots dropped'),
     TxnStat('txn_sync', 'transaction sync calls'),
+    TxnStat('txn_fsync_post', 'transaction fsync calls for checkpoint after allocating the transaction ID'),
```
agorrod: The whole block of statistics should be moved to just below `txn_fail_cache`

src/txn/txn_ckpt.c
```diff
@@ -351,6 +351,8 @@ static int
 __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 {
 	struct timespec start, stop, verb_timer;
+	struct timespec fsync_start, fsync_stop;
+	uint64_t fsync_duration_usecs;
```
agorrod: The `uint64_t` line belongs below after the `u_int` declaration

src/txn/txn_ckpt.c
```diff
@@ -351,6 +351,8 @@ static int
 __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 {
 	struct timespec start, stop, verb_timer;
+	struct timespec fsync_start, fsync_stop;
```
agorrod: Generally declarations should be in alphabetical order as well - i.e: this line would be better one line above.

agorrod:
Thanks @sulabhM lgtm - I'll merge.

PR #2785, opened 2016-06-07
WT-2682 Re-order AC_PROG_CC in the configure.ac.in script.

keithbostic:


sueloverso:
@keithbostic I can now build cleanly with clang on my Mac with this branch.  LGTM.

PR #2786, opened 2016-06-07
WT-2652 Remove unnecessary wt_ftruncate call.

sueloverso:
Also make file length the actual maximum set, not the maximum plus the header record.  Fix s_string.ok.

@michaelcahill or @agorrod Please review these changes for minor fixes to the logging code.  The fallocate and ftruncate calls were not obeying the log file maximum.  The manual zero-fill does.


michaelcahill:

PR #2789, opened 2016-06-08
WT-2514 Fix logging to not use an empty path name.

sueloverso:
@keithbostic Please review this change to set the default log path to '.' rather than an empty string.  I also added a section to the config test for it.


keithbostic:
lgtm, merged.

PR #2791, opened 2016-06-08
WT-2697 dump statistics for simple tables

keithbostic:


keithbostic:
This is the wrong way to make this change, and we're not going to do anything better now.

PR #2793, opened 2016-06-09
WT-2541 Add stats showing active readers and writers

daveh86:


daveh86:
@agorrod , mind taking a look at this tomorrow?

src/os_posix/os_fs.c
```diff
@@ -341,16 +342,19 @@ __posix_file_read(WT_FILE_HANDLE *file_handle,
 	    len >= S2C(session)->buffer_alignment &&
 	    len % S2C(session)->buffer_alignment == 0));
 
+	WT_STAT_FAST_CONN_INCR(session, block_active_read);
 	/* Break reads larger than 1GB into 1GB chunks. */
 	for (addr = buf; len > 0; addr += nr, len -= (size_t)nr, offset += nr) {
 		chunk = WT_MIN(len, WT_GIGABYTE);
+		WT_STAT_FAST_CONN_DECR(session, block_active_write);
```
sueloverso: This looks like it is a leftover.  This function is read, and that is the write stat.

sueloverso:
@daveh86 I took a look at this.  I have two general comments:
* I think the stats should go in ConnStat because they're not just block manager, but also logging and any other file we might read/write.  They seem to go along with `ConnStat('read_io',...)`, etc.  Doing that, I'd rename them to sort with the `read_io` and `write_io` stats since they're related.
* Can we just add the stat INCR/DECR in `wt_read` and `wt_write` rather than in the OS-specific layers?  All calls must come through there already.  If they're in the OS-specific code we lose the information when calling into a pluggable FS from the user.  Knowing threads are stuck in a user-provided file system call would be useful if they have a bug.

daveh86:
@sueloverso , thanks for the feedback.

I've addressed your comments, mind taking another look?

sueloverso:
@daveh86 This LGTM.  Thanks - it is much cleaner and smaller now too.  Github is showing merge conflicts, probably from the other branch I just merged.  Once that is cleaned up I'm good to go with this.

dist/stat_data.py
```diff
@@ -113,12 +113,15 @@ def __init__(self, name, desc, flags=''):
     ConnStat('cond_auto_wait_reset', 'auto adjusting condition resets'),
     ConnStat('cond_wait', 'pthread mutex condition wait calls'),
     ConnStat('file_open', 'files currently open', 'no_clear,no_scale'),
+    ConnStat('fsync_active', 'active filesystem fsync calls','no_clear,no_scale'),
```
sueloverso: Thanks!  Looking at the code, we probably want to be consistent and add in a total fsync counter, similar to `read_io` and `write_io`.  Thinking more generally, those 6 stats, plus `file_open` seem to make a critical mass to have a new section of stats, `FSStat` for file system stats.  @agorrod  is that worthwhile?  Are there other stats on the horizon that might fit in with file system ops for keeping track of things with pluggable file systems?

src/include/os_fhandle.i
```diff
@@ -92,14 +100,19 @@ static inline int
 __wt_read(
     WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t offset, size_t len, void *buf)
 {
+	WT_DECL_RET;
 	WT_RET(__wt_verbose(session, WT_VERB_HANDLEOPS,
 	    "%s: handle-read: %" WT_SIZET_FMT " at %" PRIuMAX,
 	    fh->handle->name, len, (uintmax_t)offset));
 
 	WT_STAT_FAST_CONN_INCR(session, read_io);
+	WT_STAT_FAST_CONN_INCR(session, read_active);
```
sueloverso: Minor nit: alphabetize these two lines please.

src/include/os_fhandle.i
```diff
@@ -149,7 +163,11 @@ __wt_write(WT_SESSION_IMPL *session,
 	    fh->handle->name, len, (uintmax_t)offset));
 
 	WT_STAT_FAST_CONN_INCR(session, write_io);
+	WT_STAT_FAST_CONN_INCR(session, write_active);
```
sueloverso: Same here - nit alphabetize.

sueloverso:
@daveh86 and @agorrod This LGTM.  I had a minor nit and one broader question for both of you.  The broad question would require larger changes but now would be the time to do them if we want them.

michaelcahill:
If we are going to add a new section to the statistics, I'd suggest we think of these new statistics as "thread states" rather than filesystem-specific.  What I'm looking for with this change is to see over time where threads are stalled in the system.

The initial focus is on I/O, but we could add consider adding counters for total threads in the library, threads in eviction, threads blocked on various resources, etc.  I'd prefer them to all be grouped together rather than having thread counts connected to specific subsystems.

daveh86:
I've added a commit to this pull that changes the location of the stats. I like the idea of grouping these kinds of statistics together as Michael has suggested, as thats often how I would be using them - to determine what the state of a given thread (or series of threads) is.

dist/stat_data.py
```diff
@@ -85,6 +85,10 @@ class TxnStat(Stat):
     prefix = 'transaction'
     def __init__(self, name, desc, flags=''):
         Stat.__init__(self, name, TxnStat.prefix, desc, flags)
+class ThreadState(Stat):
```
sueloverso: Alphabetizes before TxnState.

sueloverso:
Thanks @daveh86.  I pushed the one minor change and will merge once the tester completes.

sueloverso:
I thought of two general questions that should be addressed before merging.  @agorrod and @daveh86:
* We're not supposed to change the text of existing statistics.  This moves/changes the prefix for `read_io` and `write_io`.  Is that a problem?  What code needs to know about that?  Timeseries comes to mind.
* Should `ThreadState` be added to any of the groups in `stat_data.py`?  This is used for groupings in the `wtstats.py` output.  Maybe `system`?  Anywhere `ConnStat` is?

daveh86:
Thanks @sueloverso, I've made some changes to address your concerns. The new stats are now the only members of the ThreadState stat group which will save any backwards compatibility problems. I've also added the new stats to the evict and system groups as those made the most sense for these stats.

Finally, there is a new addition to these changes following from a discussion with @michaelcahill  and @agorrod surrounding using atomics for these stats.

dist/stat_data.py
```diff
@@ -97,10 +101,20 @@ def __init__(self, name, desc, flags=''):
 ##########################################
 groups = {}
 groups['cursor'] = [CursorStat.prefix, SessionStat.prefix]
-groups['evict'] = [CacheStat.prefix, ConnStat.prefix, BlockStat.prefix]
+groups['evict'] = [
+    CacheStat.prefix,
+    ConnStat.prefix,
+    BlockStat.prefix,
```
sueloverso: nit - Please alphabetize.

src/include/os_fhandle.i
```diff
@@ -92,14 +101,19 @@ static inline int
 __wt_read(
     WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t offset, size_t len, void *buf)
 {
+	WT_DECL_RET;
```
sueloverso: nit - add a blank line after the declaration

src/include/os_fhandle.i
```diff
@@ -140,6 +154,7 @@ static inline int
 __wt_write(WT_SESSION_IMPL *session,
     WT_FH *fh, wt_off_t offset, size_t len, const void *buf)
 {
+	WT_DECL_RET;
```
sueloverso: same here

sueloverso:
Thanks @daveh86.  LGTM.  I pushed a change for a couple minor nits and then I'll merge.

PR #2794, opened 2016-06-10
WT-2696 Wait if we find an unbuffered flag without the size set yet.

sueloverso:
@michaelcahill Not ready to merge as I don't have the test committed yet, but this is the code fix.


sueloverso:
@michaelcahill This is ready for review.  I replaced the existing test with the new version and confirmed it still detects the original bug when the test was written on an old source tree, plus it detects the current bug and verifies the fix.

michaelcahill:

PR #2796, opened 2016-06-10
WT-2698 Change flag into atomically manipulated field to avoid deadlock.

sueloverso:
@michaelcahill or @agorrod This is ready for review.  I converted the flag into an atomically set counter.  I have not been able to reproduce the original hang running the recover.sh script before the fix however, so I haven't been able to verify the fix of the deadlock.


michaelcahill:
Thanks, @sueloverso, lgtm -- I'll merge.

PR #2797, opened 2016-06-14
WT-2667 Split evergreen compile-posix into separate compile and test …

daveh86:
…stages

### DIFF
--- a/test/mciproject.yml
+++ b/test/mciproject.yml
@@ -45,7 +45,7 @@ tasks:
             set -o verbose
 
             ./reconf
-            ${configure_env_vars|} ../configure --enable-diagnostic --enable-python --enable-zlib
+            ${configure_env_vars|} ../configure --enable-diagnostic --enable-python --enable-zlib --enable-strict
             ${make_command|make} ${smp_command|}
             ${make_command|make} check
       - command: archive.targz_pack
--- a/test/mciproject.yml
+++ b/test/mciproject.yml
@@ -8,6 +8,18 @@ functions:
     command: git.get_project
     params:
       directory: wiredtiger
+  "fetch artifacts" : &fetch_artifacts
+    - command: s3.get
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        remote_file: wiredtiger/${build_variant}/${revision}/artifacts/${build_id}.tgz
+        local_file: wiredtiger.tgz
+        bucket: build_external
+    - command: shell.exec
+      params:
+        script: |
+          tar -xvf wiredtiger.tgz
 
 pre:
   - command: shell.exec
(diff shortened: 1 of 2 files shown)


daveh86:
@agorrod, mind taking a look at this?

test/mciproject.yml
```diff
@@ -21,25 +33,53 @@ post:
...
             set -o errexit
             set -o verbose
 
-            ./build_posix/reconf
-            ${configure_env_vars|} ./configure --enable-diagnostic --enable-python --enable-zlib
+            ./reconf
+            ${configure_env_vars|} ../configure --enable-diagnostic --enable-python --enable-zlib
             ${make_command|make} ${smp_command|}
             ${make_command|make} check
+      - command: shell.exec
```
markbenvenuto: use `archive.targz_pack` instead, see https://github.com/mongodb/mongo/blob/master/etc/evergreen.yml for examples.

test/mciproject.yml
```diff
@@ -8,6 +8,18 @@ functions:
...
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        remote_file: wiredtiger/${build_variant}/${revision}/artifacts/${build_id}.tgz
+        local_file: wiredtiger.tgz
+        bucket: build_external
+    - command: shell.exec
+      params:
+        script: |
+          tar -xvf wiredtiger.tgz
```
markbenvenuto: Don't you need `z`?

daveh86: Tar wierdness, the {{z}} is implicit. However, you're right and I should be explicit.

test/mciproject.yml
daveh86: I tried that, but the wildcard didn't work. Gave it a second go this evening with double wildcard and that seems to be the key.

daveh86:
@agorrod, can you have a look at this?

markbenvenuto:
@daveh86 I like the `-j` flag on the tests, nice.

BTW, you can also exclude files (like say `.git`).

LGTM

test/mciproject.yml
```diff
@@ -21,25 +29,55 @@ post:
...
-          working_dir: "wiredtiger"
+          working_dir: "wiredtiger/build_posix"
           script: |
             set -o errexit
             set -o verbose
 
-            ./build_posix/reconf
-            ${configure_env_vars|} ./configure --enable-diagnostic --enable-python --enable-zlib
+            ./reconf
+            ${configure_env_vars|} ../configure --enable-diagnostic --enable-python --enable-zlib
```
agorrod: @daveh86 Can you add the ``--enable-strict`` option to the command line?

agorrod:
Thanks @daveh86 I had one request, apart from that this lgtm - it's a big step forward in our Evergreen testing.

daveh86:
@agorrod, done.

agorrod:

PR #2798, opened 2016-06-14
WT-2702 Block operations when the cache is 100% full.

michaelcahill:


michaelcahill:

I don't think the change itself is contentious, but there is work to do on testing, both a regression test for the original issue and perf tests for this change.

agorrod:
Thanks @michaelcahill I've switched the Jenkins perf jobs over to this branch.

agorrod:
Thanks @michaelcahill. The Jenkins tests look stable, I believe this solution introduces longer maximum latencies (details in the JIRA ticket), but sticking to the cache size is more important.

PR #2799, opened 2016-06-15
WT-2665 Limit allocator fragmentation from the WiredTiger cache

michaelcahill:


michaelcahill:
These changes are not ready for merging.

In particular, a key part of the changes is currently disabled (the new "scrub" eviction mode).  There is a performance regression early in workloads before the first split in a tree.

Also, far more performance testing is required in general.

sueloverso:
@michaelcahill This branch is showing some severe performance issues on a number of tests.  It has a very interesting effect on the grow/shrink stress test.

michaelcahill:
@sueloverso, thanks, some of that is expected.  This branch is not complete, and in particular has a known regression related to "forced eviction" (where a page is growing large).  I didn't expect the `medium-btree` regression: investigating that is on my todo list.

src/include/cache.i
```diff
@@ -113,6 +113,47 @@ __wt_cache_dirty_inuse(WT_CACHE *cache)
...
+ */
+static inline uint64_t
+__wt_cache_bytes_other(WT_CACHE *cache)
+{
+	uint64_t bytes_image, bytes_inmem, bytes_other;
+
+	bytes_image = cache->bytes_image;
+	bytes_inmem = cache->bytes_inmem;
+
+	/* XXX Some kind of race or accounting error. */
```
agorrod: Could this be an assertion instead of a scary comment?

agorrod:
@michaelcahill lgtm % a tiny nit.

michaelcahill:
@agorrod, I can make the comment less scary, but I suspect as assertion here will fire occasionally.  A race is certainly possible -- there could be an incr or decr in between reading the two values, for example.

agorrod:
@michaelcahill This looks good to me now, I'm going to merge and watch for fallout.

PR #2800, opened 2016-06-15
WT-2707 stricter checks in dist/s_label, and add some WT_ERR calls

ddanderson:


ddanderson:

keithbostic:
@ddanderson, lgtm. This was a nice catch -- thanks!

PR #2801, opened 2016-06-15
WT-2650 Allow quotes to be escaped within a configuration string

ddanderson:
__wt_config_next() potentially allocates storage, requiring a call to to clean up WT_CONFIG structs using __wt_config_free().


ddanderson:
This change turned out to be too invasive, and it would take more fiddling to make it really work.

PR #2802, opened 2016-06-16
WT-2622 Clear old position for all random lookups.

michaelcahill:


michaelcahill:
As noted in the ticket, this change fixes the bug for me.

The clue was when I looked at a core dump and saw `cbt->ins != NULL` but `cbt->ref->page` had `page->modify == NULL` (i.e., it had never been modified, so the cursor should not have been pointing at an inserted record.

agorrod:
@michaelcahill This fix looks good to me, but I believe it references the wrong JIRA ticket.

michaelcahill:
Thanks @agorrod -- WT-2622 (one 6 two 2s, not vice versa).

I think we can fix that during the merge: is that okay or should I open a new pull request?

agorrod:

PR #2803, opened 2016-06-16
WT-2706 Fix lost log writes when switching files

sueloverso:
without writing a log record.
@michaelcahill and @agorrod Please review.  This has been running for a few hundred iterations with both the mongoDB reproducer and random-abort and is ready to merge I believe.
* There was an incorrect assertion where we'd read a slot and fail the assertion because the fields of the slot were changing.
* I added some randomization to the random-abort program.  The minimums are the old values but I let it also run longer and with more threads too.
* There is the fix for the log_fh which is the real fix for the data loss.
* I changed the conditional in `wt_log_write` and added a flag instead of looking at the `log_fh`.  I could have gone either way, adding a `flags` field to the log structure or adding in a log state flag into the connection flags instead.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -790,6 +790,7 @@ __log_newfile(WT_SESSION_IMPL *session, bool conn_open, bool *created)
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
+	WT_FH *log_fh;
 	WT_LOG *log;
 	WT_LSN end_lsn;
 	int yield_cnt;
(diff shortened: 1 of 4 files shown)


src/log/log.c
```diff
@@ -862,8 +863,15 @@ __log_newfile(WT_SESSION_IMPL *session, bool conn_open, bool *created)
...
+	/*
+	 * Since the file system clears the output file handle pointer before
+	 * searching the handle list and filling in the new file handle,
+	 * we must pass in a local file handle.  Otherwise there is a wide
+	 * window where another thread could see a NULL log file handle.
+	 */
 	WT_RET(__log_openfile(session,
-	    false, &log->log_fh, WT_LOG_FILENAME, log->fileid));
+	    false, &log_fh, WT_LOG_FILENAME, log->fileid));
+	log->log_fh = log_fh;
```
michaelcahill: I suggest making this `WT_PUBLISH(log->log_fh, log_fh);` to make it really clear that we are updating a shared handle.

sueloverso: Fixed and pushed.

michaelcahill:

PR #2804, opened 2016-06-16
WT-2650 Process string escape characters in WT_CONFIG_PARSER API.

ddanderson:


ddanderson:
This is not ready for merge.  I realized I should back out WT-2611, and test that for proper behavior, since this is supposed to replace that.

michaelcahill:
@ddanderson, if we're going to go down this path, can we put the unescaped string into `WT_CONFIG_ITEM`?

ddanderson:
@michaelcahill, wouldn't putting an allocated string in WT_CONFIG_ITEM mean that structure needs to be initialized and freed? Right now WT_CONFIG_ITEM is lightweight. I suppose we could have a lightweight version of WT_CONFIG_ITEM for internal use, and a heavier one for the API? Or were you thinking of something else?

src/config/config.c
```diff
@@ -558,6 +577,51 @@ __config_process_value(WT_CONFIG *conf, WT_CONFIG_ITEM *value)
 		 */
 		if (value->type == WT_CONFIG_ITEM_NUM && errno == ERANGE)
 			goto range;
+	} else if (punescaped != NULL &&
+	    value->type == WT_CONFIG_ITEM_STRING &&
+	    __config_strnchr(value->str, '\\', value->len) != NULL) {
+		__wt_realloc(conf->session, NULL, value->len, punescaped);
```
agorrod: @ddanderson I've been sitting on this pull request because I'm very uncomfortable adding a memory allocation into this code path, but I haven't put in enough thought to make a concrete alternative suggestion. Can you think of another way around this problem?

ddanderson:
@agorrod, I believe the memory allocation (and hence this capability) only happens when the caller uses WT_CONFIG_PARSER.  So, it wouldn't happen with a wiredtiger_open call.  In addition, even when using WT_CONFIG_PARSER, an allocation only happens when getting an item with a backslash in the string.

Hold off on the review for the moment...  We may want to do unescaping during a wiredtiger_open call, for example, error_prefix may need to have arbitrary characters in it, maybe there may be other cases.  So I'm making the unescape method callable directly.  In the error_prefix case, there's already a strdup, the allocation done by unescape would replace that.  That should be committed soon, I'll ping you.  Thanks!

ddanderson:
@agorrod, I've added pull request #2932 as a branch off this one, as a potential add-on (If you want it, merge that into this branch before merging this one into develop).  They are both ready for review.

michaelcahill:
@ddanderson I've taken more of a look at this one.  I am inclined to either go all the way, and always have strings unescaped in `WT_CONFIG_ITEM` (and add cleanup code to free that memory everywhere), or to otherwise make it really clear to callers that they have to call `\_\_wt_config_unescape`, probably with a new `WT_CONFIG_ITEM.type` like `WT_CONFIG_ITEM_ESCSTRING`.  Maybe both.

I'll take a look at how much work would be involved in adding cleanup code everywhere.  This is also probably a place where scratch buffers make more sense than explicitly allocating and freeing memory.

daveh86:
Trying to trigger a spurious build.

daveh86:
Sorry @ddanderson, I used this PR to test and see if age is a factor in comments triggering PR builds. Turns out it is...

ddanderson:
@daveh86, Interesting discovery!  No problem, I've been meaning to get back to this one anyway.

ddanderson:
Closing this pull request per discussion in https://jira.mongodb.org/browse/WT-2650 .

PR #2805, opened 2016-06-16
WT-2710 WT_FILE_HANDLE_INMEM no longer needs an off field

keithbostic:
Probably a left-over from when stdio was still part of the pluggable API.


agorrod:

PR #2806, opened 2016-06-17
SERVER-24580 Don't have application threads spin to access the evicti…

michaelcahill:
…on queue.

When the cache hits eviction triggers, all application threads can
hammer the eviction queue lock, starving each other and server threads.

Also, noticed with the same workload, the eviction server doesn't need
to force updates to the oldest ID (which can starve the eviction server
thread if there are hundreds of application threads getting snapshots).
It is sufficient to update it lazily.


agorrod:

PR #2807, opened 2016-06-17
WT-2712 Coverity 1356928 and 1356929: ASSERT_SIDE_EFFECT

keithbostic:
@ddanderson, a couple of Coverity complaints in the csuite join tests.


ddanderson:
@keithbostic, thanks for handling this.  LGTM.  I guess I'll steer clear of side effects in testutil_assert calls.

PR #2808, opened 2016-06-17
WT-2714 lint

keithbostic:
@sueloverso, @ddanderson, some lint in the csuite test programs.


sueloverso:
Thanks @keithbostic .

PR #2810, opened 2016-06-17
WT-2715 random abort failed with partial write

sueloverso:
@michaelcahill Please review this change to fix the random-abort recovery test program.  I actually made three substantive changes:
* Fix a bug in counting the total records.
* Added a `-v` option to only run the data verification phase of the test (needed for me to test this fix on the database that failed).
* Fixed the verification code to detect the partially written record.


sueloverso:
autobuild: retest this please

sueloverso:
@michaelcahill Although the fix has been working well for a few days, it occurred to me that a better fix is probably to just set line buffering on the side file rather than no buffering.  The file with the partial line was truncated on a 4k boundary.

sueloverso:
This is ready for review.  I moved to line buffering and removed the check for partial data that I added.  I have restarted the test.

agorrod:

PR #2811, opened 2016-06-18
WT-2709 connection reconfigure segfault in __wt_conn_cache_pool_destroy

keithbostic:


agorrod:
@keithbostic The fix looks good to me, I'd really like a Python test case that covers fuzz testing of the reconfigure API. Is that something you could take a run at?

keithbostic:
@agorrod, sure, will do. I'm going to merge this branch to get rid of the SECURITY-393 ticket, and I'll do that testing in a new ticket.

PR #2812, opened 2016-06-19
WT-2692 Fix race in file system example

keithbostic:


agorrod:
@keithbostic I agree that this branch is the best we can do without significant extra work. I pushed some nits up. If you are happy with those please go ahead and merge.

keithbostic:
@agorrod, I made one further change in 8e4b186, to not require the `WT_FILE_HANDLE.sync` method if the file is readonly. That matches the documentation, and makes sense to me.

I reviewed the rest of the `WT_FILE_HANDLE` methods against the documentation, and they agree.

I'll merge once the tests pass.

PR #2813, opened 2016-06-20
WT-2605 Fix valgrind

daveh86:


daveh86:
@agorrod, this fixes one of the valgrind issues found. Can you review?

agorrod:
@daveh86 I believe this will fix the error Valgrind sees, but I'm not convinced it's the root cause for the failure.

The code being complained about is:

```
 		testutil_check(join_cursor->get_key(join_cursor, &recno));
 		testutil_check(join_cursor->get_value(join_cursor, &country, &year, &population));
 		printf("ID %" PRIu64, recno);
```

If the call to get_key is operating as expected (i.e populating the value when not returning an error), I'd expect ``recno`` to always be initialized to something. I thought Valgrind was clever enough to detect variable initialization via a sub-call.

agorrod:
@ddanderson Would you mind picking this one up today? If the fix from @daveh86 is complete please go ahead and merge.

ddanderson:
@daveh86, lgtm, I'll merge.

PR #2816, opened 2016-06-20
WT-2408 Windows error translation layer

keithbostic:
@agorrod, @michaelcahill, @markbenvenuto: here's an approach to the Windows-to-POSIX/ANSI translation layer.

My goal was to leave as much code alone as possible and I wanted us to be able to continue to mix-and-match code in the Windows port that returns errno, MSVC system errors and WiredTiger errors without any new macros (for example, I didn't want to create `WT_RET_MSG_WIN32`). To make that work, I ended up inlining the "standard" code in a wrapping function that maps any returned Windows errors to a POSIX/ANSI error, and leaves all other returns alone.

I moved the Windows encoding to be past the WiredTiger error range encoding so there's no way they can overlap.

I replaced @sueloverso's readonly mapping with POSIX error checks.

This branch doesn't yet have the changes to export error mapping functionality through the `WT_EXTENSION_API` handle, I wanted to make sure I wasn't making this more complicated than it should be and there's not a simpler/better solution.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -86,6 +86,7 @@ DbEnv
 Decrement
 Decrypt
 DeleteFileA
+EACCES
 EAGAIN
 EB
 EBUSY
@@ -601,6 +602,7 @@ eof
 eq
 equalp
 errhandler
+errmap
 errno
 errx
 esc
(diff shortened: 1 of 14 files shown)


keithbostic:
I expect we're going to throw this branch away, but if we do end up merging it, fix this first:
```
       /* NULL means load from the current binary */
        if (path == NULL) {
                if (GetModuleHandleExA(
                    0, NULL, (HMODULE *)&dlh->handle) == FALSE) {
                        ret = __wt_getlasterror();
                        WT_ERR_MSG(session, ret,
                            "GetModuleHandleEx(%s): %s", path, 0);
                }
```
It's in `os_win/os_dlopen.c`, notice we're passing `0` to a `%s` format string, it's been there since the initial Windows port.

keithbostic:
We're going with #2819, close this pull request.

PR #2817, opened 2016-06-21
WT-2720 Change the return codes in run.py to flag on fails

daveh86:


daveh86:
@agorrod, mind taking a look at this? It's a number of changes to the return codes for the unit-test suite.

agorrod:
@daveh86 nice catch. I see two other suspicious uses of sys.exit in our Python suite:

```
test/suite/wttest.py
554:    sys.exit(not result.wasSuccessful())

test/suite/run.py
321:    sys.exit(not result.wasSuccessful())
```

Did you consider those and decide they were reasonable?

daveh86:
@agorrod, those two seem alright to me.

The issue was that a return of False, meant returning 0. In two cases above, the False would only be the exit value when wasSuccessful() is True, this should be the correct behaviour.

agorrod:
lgtm

PR #2818, opened 2016-06-21
WT-2691 Use wrappers for ctype functions to avoid sign extension errors

ddanderson:
Create __wt_* versions of all ctype functions, and use them whenever wt_internal.h is available.  Add check to prevent direct use of type functions from core source.

This is an alternative to simply banning the use of plain 'char', as in #2783.


ddanderson:
@keithbostic, can you review please?

src/include/ctype.i
```diff
@@ -0,0 +1,139 @@
...
+
+/*
+ * __wt_iscntrl --
+ *	Wrap the ctype function without sign extension.
+ */
+static inline int
+__wt_iscntrl(int c)
+{
+	return iscntrl((unsigned char)c);
+}
```
michaelcahill: Consider omitting unused wrapper until we need them.

src/include/ctype.i
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+#include <ctype.h>
+
+/*
+ * __wt_isalnum --
+ *	Wrap the ctype function without sign extension.
+ */
+static inline int
+__wt_isalnum(int c)
```
michaelcahill: Can our wrappers take a `u_char` argument to avoid the explicit cast?  That seems more likely to generate warnings from invalid code...

I'm not so fussed about the `int` return here, but we could consider returning a `bool`.  In other words, this would become:

```c
static inline bool
__wt_isalnum(u_char c)
{
        return (isalnum(c) != 0);
}
```

ddanderson:
@michaelcahill, thanks for the comments, please merge if this resolves them.

bench/wtperf/config.c
```diff
@@ -599,7 +600,7 @@ config_opt_file(CONFIG *cfg, const char *filename)
 			rtrim = comment;
 
 		/* Skip trailing space again. */
-		while (rtrim > ltrim && isspace(rtrim[-1]))
+		while (rtrim > ltrim && isspace((unsigned char)rtrim[-1]))
```
keithbostic: @ddanderson, I'd suggest using `u_char`.

keithbostic:
@ddanderson, lgtm, merge at will as far as I'm concerned.

ddanderson:
self reviewing this last bit, and merging.

PR #2820, opened 2016-06-21
WT-2704 Fix bug in diag code detecting duration of stuck cache.

sueloverso:
@agorrod or @michaelcahill please review this change to detect the stuck cache, dump it and abort.  I tested it with the failing CONFIG in the ticket and it correctly dumped the cache.


agorrod:

PR #2821, opened 2016-06-21
WT-2722 s_label fix for false positve

ddanderson:
Regular expression meta-character needed to be escaped so that egrep works consistently on all systems.


ddanderson:
@sueloverso, can you review this small change please?

sueloverso:
Thanks @ddanderson

PR #2822, opened 2016-06-22
WT-2704 Don't re-initialize pages evicted on eviction thread start.

agorrod:
We aren't guaranteed that the cache structure is populated yet, and the value is initialized on cache create regardless.


agorrod:
Self reviewing, lgtm.

sueloverso:
Thanks @agorrod

PR #2823, opened 2016-06-22
WT-2605 fix key format for join suite test, which fixes valgrind error

ddanderson:


ddanderson:
@michaelcahill and @daveh86, I think this fixes the valgrind problem.

agorrod:
@ddanderson That fix looks correct to me - thanks.

PR #2826, opened 2016-06-22
WT-2726 Revert arg change to wt_txn_update_oldest.

sueloverso:


michaelcahill:
@sueloverso, thanks for identifying the problem -- I'm hoping #2829 will solve it without re-breaking  the multi-threaded workloads in SERVER-24580.

PR #2827, opened 2016-06-22
WT-2724 Pass in session address, not pointer to it.

sueloverso:
@agorrod This fixes the issue with eviction workers exiting immediately from the `btree-split-stress.wtperf` workload.  The issue does not affect tests like  `evict-btree.wtperf`.  The `btree-split-stress.wtperf` hit the problem because it sets an eviction thread worker minimum so those threads get started in `evict_workers_resize`.  Other tests like `evict-btree.wtperf` do not set an explicit minimum and therefore, threads are created in `evict_pass` as needed.  That path was correct.  Only the resize path was wrong.  Please review.


agorrod:
Autobuild: retest this please

src/evict/evict_lru.c
```diff
@@ -232,6 +232,12 @@ __evict_thread_run(void *arg)
 	WT_ERR(__wt_verbose(
 	    session, WT_VERB_EVICTSERVER, "cache eviction thread exiting"));
 
+	/*
+	 * Eviction threads should only ever stop when the connection is
+	 * closing - check that otherwise fewer eviction worker threads may be
+	 * running than expected.
+	 */
+	WT_ASSERT(session, ret == 0 || F_ISSET(conn, WT_CONN_CLOSING));
```
michaelcahill: What's the case where `ret == 0 && !F_ISSET(conn, WT_CONN_CLOSING)` is okay?

agorrod: I don't know of one - I'll take out the checking for return conditions. I debated shuffling the ``if (0)``, since it's not my preferred style - which is how I ended up checking the return value.

michaelcahill:

PR #2829, opened 2016-06-23
SERVER-24580 Update oldest txn ID with "strict, nowait" flags.

michaelcahill:
The earlier change in SERVER-24580 made the eviction server not force update to the oldest transaction ID.  That was done to avoid a performance collapse with many threads in the library (every run of the eviction server had to wait for all threads to exit the snapshot table and blocked all threads from entering it).

Previously "force" meant both (a) "update no matter how far behind the oldest has fallen" and (b) "wait for locks".  This change splits those apart, calling (a) "strict" and (b) "wait".  It sets strict, no-wait for the eviction server updates to the oldest ID.


src/evict/evict_lru.c
```diff
@@ -631,7 +631,7 @@ __evict_pass(WT_SESSION_IMPL *session)
 		 * of whether the cache is full, to prevent the oldest ID
 		 * falling too far behind.
 		 */
-		WT_RET(__wt_txn_update_oldest(session, false));
+		WT_RET(__wt_txn_update_oldest(session, true, false));
```
michaelcahill: This is the key fix -- the eviction server should update the oldest transaction ID any time it has fallen behind, but not wait for locks on the snapshot table.

src/evict/evict_page.c
```diff
@@ -420,7 +420,7 @@ __evict_review(
 		 * fallen behind current.
 		 */
 		if (modified)
-			WT_RET(__wt_txn_update_oldest(session, false));
+			WT_RET(__wt_txn_update_oldest(session, true, false));
```
michaelcahill: This change is arguable -- before evicting a page, be strict about updating the oldest ID but don't wait for a lock on the snapshot table.  I could go either way on this one based on further perf testing, but I think this change is safe and better than what we were doing before.

michaelcahill:

agorrod:
@michaelcahill the functional changes lgtm. I think it would be more maintainable to use named flags rather than a pair of booleans, but that's personal preference.

agorrod:

PR #2831, opened 2016-06-23
WT-2725 Change how we detect whether a page can be evicted when cache…

daveh86:
… is full

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1247,18 +1247,7 @@ fast:		/* If the page can't be evicted, give up. */
 		if (!__wt_page_can_evict(session, page, true, NULL))
 			continue;
 
-		/*
-		 * If the page is clean but has modifications that appear too
-		 * new to evict, skip it.
-		 *
-		 * Note: take care with ordering: if we detected that the page
-		 * is modified above, we expect mod != NULL.
-		 */
 		mod = page->modify;
-		if (!modified && mod != NULL && !LF_ISSET(
-		    WT_EVICT_PASS_AGGRESSIVE | WT_EVICT_PASS_WOULD_BLOCK) &&
-		    !__wt_txn_visible_all(session, mod->rec_max_txn))
-			continue;
 
 		/*
 		 * If the oldest transaction hasn't changed since the last time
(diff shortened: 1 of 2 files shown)


src/evict/evict_lru.c
```diff
@@ -1273,8 +1262,7 @@ __evict_walk_file(WT_SESSION_IMPL *session, u_int *slotp, uint32_t flags)
 		 * running last time we wrote the page has since rolled back,
 		 * or we can help get the checkpoint completed sooner.
 		 */
-		if (modified && !LF_ISSET(
-		    WT_EVICT_PASS_AGGRESSIVE | WT_EVICT_PASS_WOULD_BLOCK) &&
+		if (modified &&
 		    (mod->disk_snap_min == S2C(session)->txn_global.oldest_id ||
```
michaelcahill: This part of the change should be unrelated.  Basically, this check is an optimization -- don't try to evict pages if we can tell with these quick checks that eviction will fail.

src/include/btree.i
```diff
@@ -1083,6 +1083,13 @@ __wt_page_can_evict(WT_SESSION_IMPL *session,
 	/* Pages that have never been modified can always be evicted. */
 	if (mod == NULL)
 		return (true);
+	 /*
+	  * If the page is clean but has modifications that appear too
+	  * new to evict, skip it.
+	  */
+	if (!__wt_page_is_modified(page) &&
+	    !__wt_txn_visible_all(session, mod->rec_max_txn))
+	        return (false);
```
michaelcahill: If you change this to `if (!__wt_txn_visible_all(session, mod->rec_max_txn)) ...`, is that sufficient to avoid the assertion (without the unrelated change to `evict_lru.c`)?

daveh86:
test this please

src/include/btree.i
daveh86: I've patched these changes in for testing and adding the check for the WT_EVICT_PASS_WOULD_BLOCK allows the test to pass, however adding the WT_EVICT_PASS_AGGRESSIVE flag as an exclusion causes the test to fail.

daveh86:
@agorrod , would you mind taking a look at this?

src/include/btree.i
```diff
@@ -1085,6 +1085,14 @@ __wt_page_can_evict(WT_SESSION_IMPL *session,
 		return (true);
 
 	/*
+	 * If the page is clean but has modifications that appear too
+	 * new to evict, skip it.
+	 */
+	if (!__wt_txn_visible_all(session, mod->rec_max_txn) ||
+	    !__wt_txn_visible_all(session, mod->update_txn))
```
agorrod: @daveh86 I suggested looking for different transaction IDs that might be more suitable here, but I don't think ``mod->update_txn`` is the one we are interested in. The ``update_txn`` is used as an advisory value, and shouldn't be involved in correctness checks (which we are pursuing here).

The expectation in the code is that either:
* The page is clean and ``rec_max_txn`` accurately reflects the ID of the most recent update on the page (which needs to be kept around as long as the oldest reader was started before the update).
* The page is dirty, and the state of the update lists are checked during reconciliation.

Can you reproduce this failure reliably? If so, could you capture the assertion in a debugger? We would want to be able to see the fully populated page-modify structure, so we may need to restructure the code a bit to find the problem before cleaning up content.

agorrod:
Closing. This will be replaced with a different version.

PR #2832, opened 2016-06-24
WT-2729 Focus eviction on the largest trees in cache.

michaelcahill:
Randomize visits to trees that use a tiny fraction of the cache.

### DIFF
--- a/src/conn/conn_cache.c
+++ b/src/conn/conn_cache.c
@@ -235,7 +235,7 @@ __wt_cache_stats_update(WT_SESSION_IMPL *session)
 	 */
 	if (conn->evict_session != NULL)
 		WT_STAT_SET(session, stats, cache_eviction_walks_active,
-		    conn->evict_session->nhazard);
+		    cache->walk_session->nhazard);
 }
 
 /*
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1311,6 +1311,13 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		 * Randomly walk trees with a tiny fraction of the cache in
 		 * case there are so many trees that none of them use enough of
 		 * the cache to be allocated slots.
+		 *
+		 * Map a random number into the range [0..1], and if the result
+		 * is greater than the fraction of the cache used by this tree,
+		 * give up.  In other words, there is a small chance we will
+		 * visit trees that use a small fraction of the cache.  Arrange
+		 * this calculation to avoid overflow (e.g., don't multiply
+		 * anything by UINT32_MAX).
 		 */
 		if (__wt_random(&session->rnd) / (double)UINT32_MAX >
 		    btree_inuse / (double)cache_inuse)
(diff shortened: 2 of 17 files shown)


agorrod:
This introduces a regression on medium-btree wtperf workload - we should investigate before merging.

agorrod:
@michaelcahill Looks like this covers WT-2261 now as well - should we mark that ticket as a duplicate?

src/evict/evict_lru.c
```diff
@@ -1297,11 +1287,41 @@ __evict_walk_file(WT_SESSION_IMPL *session, uint32_t queue_index, u_int *slotp)
...
+	target_pages = (uint32_t)(
+	    (btree_inuse + bytes_per_slot / 2) / bytes_per_slot);
+	if (target_pages == 0) {
+		/*
+		 * Randomly walk trees with a tiny fraction of the cache in
+		 * case there are so many trees that none of them use enough of
+		 * the cache to be allocated slots.
+		 */
+		if (__wt_random(&session->rnd) / (double)UINT32_MAX >
+		    btree_inuse / (double)cache_inuse)
```
agorrod: This condition could use a comment - I'm not sure what the calculation is achieving.

src/evict/evict_lru.c
```diff
@@ -1559,9 +1582,12 @@ __evict_get_ref(
 		candidates /= 2;
 
 	/* Get the next page queued for eviction. */
-	while ((evict = cache->evict_current) != NULL &&
-	    evict < evict_queue->evict_queue + candidates &&
-	    evict->ref != NULL) {
+	for (evict = cache->evict_current;
```
agorrod: Can we be sure that`` evict_current`` is never NULL on entry to this for loop?

agorrod:
@michaelcahill I've done an initial review and the changes look good modulo a few comments.

I switched test-format-stress over to this branch as well.

michaelcahill:
@agorrod, thanks for the review.

The medium-btree regression is now fixed -- the branch is no 10% faster than develop for `medium-btree` and 20% faster for `update-btree`.  The newly-added workload in 10x faster on this branch than develop.

Agreed about WT-2261: I've closed it as a dup.

src/evict/evict_lru.c
michaelcahill: No we can't, but the loop condition checks that `evict` is between the start of the queue and the end, so it will catch `NULL`.

src/evict/evict_lru.c
michaelcahill: I've added some more words: is it clear now?

src/conn/conn_cache.c
```diff
@@ -235,7 +235,7 @@ __wt_cache_stats_update(WT_SESSION_IMPL *session)
 	 */
 	if (conn->evict_session != NULL)
 		WT_STAT_SET(session, stats, cache_eviction_walks_active,
-		    conn->evict_session->nhazard);
+		    cache->walk_session->nhazard);
```
michaelcahill: This is a bug fix for `develop`: this is different in the 3.2 code and it was screwed up as the change moved between branches.

agorrod:
@michaelcahill I'm happy with this branch now - should I go ahead and merge?

PR #2834, opened 2016-06-24
WT-2711 statistics log paths

keithbostic:


keithbostic:
@agorrod, I think this one is ready for your review/merge when the tests complete.

The last set of changes (to disallow `statistics_log=(path=XXX)` in the `WT_CONNECTION.reconfigure` method), turned into a mess for a couple of reasons: first, it's fundamentally messy and second, it took me awhile to remember how all the configuration stuff works, and I started off in the wrong direction. I'd suggest ignoring the individual commits, and review the final diff, I think it will be an easier review that way.

One thing I don't like at the moment: every time we call reconfigure, we start/stop the checkpoint and statistics-log server threads. We don't anticipate applications reconfiguring things that often, but still, it's ugly that reconfiguring statistics-logging starts/stops the checkpoint server. We could fix it, but that requires handling the server running at the same time we're potentially changing the server's `WT_CONNECTION_IMPL` information, and there are obvious races. I left it alone because I don't think there's a significant problem at the moment, but wanted to call it to your attention.

Finally, a few places in our code we'd assumed applications would always bundle their reconfiguration information in a single call, that is (just as an example), doing:
```
conn.reconfigure(conn, "checkpoint=(log_size=1MB,wait=5)");
```
and not:
```
conn.reconfigure(conn, "checkpoint=(log_size=1MB)");
conn.reconfigure(conn, "checkpoint=(wait=5)";
```
I fixed the ones I saw, and I can't think of any way to make that better (or somehow catch those kinds of assumptions in the code), but I think both forms have to work.

cc: @michaelcahill

test/suite/test_reconfig01.py
```diff
@@ -95,19 +95,24 @@ def test_reconfig_checkpoints(self):
...
-        self.conn.reconfigure("statistics_log=(wait=0)")
-        self.conn.reconfigure("statistics_log=(wait=2,sources=[lsm:])")
-        self.conn.reconfigure("statistics_log=(wait=0)")
-        self.conn.reconfigure("statistics_log=(wait=2,timestamp=\"t%b %d\")")
-        self.conn.reconfigure("statistics_log=(wait=0)")
-        self.conn.reconfigure("statistics_log=(wait=2,path=\"wts.%d.%H\")")
-        self.conn.reconfigure("statistics_log=(wait=0)")
-        self.conn.reconfigure(
-             "statistics_log=(wait=2,sources=[lsm:],timestamp=\"%b\")")
+        self.conn.reconfigure("statistics_log=(wait=1,json=true)")
```
agorrod: The old code tested toggling statistics logging on and off, not just resetting different timer intervals. Is there a reason you took that out?

src/os_win/os_path.c
```diff
@@ -16,12 +16,38 @@ bool
...
+	 * -- "C:tempdir\tmp.txt" refers to a file in a subdirectory to the
+	 *    current directory on drive C.
 	 */
-	if (strlen(path) >= 3 && __wt_isalpha(path[0]) && path[1] == ':')
-		path += 2;
-	return (path[0] == '/' || path[0] == '\\');
+	if (path[0] == '\\')
+		return (true);
+
+	if (__wt_isalpha(path[0]) && path[1] == ':')
```
agorrod: This will cause a segfault if the path is less than two characters long. It also looks incorrect to me, before a ``C:tmp.txt`` would return false, now it will return true. It also isn't testing for Windows style path separators ``/`` any more.

I like the comment, but prefer the previous version of the code.

src/conn/conn_stat.c
```diff
@@ -96,24 +113,32 @@ __statlog_config(WT_SESSION_IMPL *session, const char **cfg, bool *runp)
...
-	if (!*runp && !FLD_ISSET(conn->stat_flags, WT_CONN_STAT_ON_CLOSE))
-		return (0);
-
-	/*
-	 * If any statistics logging is done, this must not be a read-only
-	 * connection.
+	 * We don't allow the path to be reconfigured. The reason is MongoDB
+	 * allows admin privileges to reconfigure running WiredTiger instances,
+	 * but admin privileges may be different from the privileges used to
+	 * start MongoDB. There's no strong reason it's useful to reconfigure
```
agorrod: I'd prefer if this comment was generalized away from MongoDB. Maybe something like:

``
We don't allow the log path to be reconfigured due to security concerns. Any application that passes input strings directly to reconfigure would expose themselves to a security hole. The added utility of reconfiguring a statistics log path isn't worth the security risk.
``

agorrod:
@keithbostic This change generally looks good to me. I had some specific comments, but it's looking good.

agorrod:
```
Finally, a few places in our code we'd assumed applications would always bundle their reconfiguration information in a single call, that is (just as an example), doing:

conn.reconfigure(conn, "checkpoint=(log_size=1MB,wait=5)");
and not:

conn.reconfigure(conn, "checkpoint=(log_size=1MB)");
conn.reconfigure(conn, "checkpoint=(wait=5)";
I fixed the ones I saw, and I can't think of any way to make that better (or somehow
 catch those kinds of assumptions in the code), but I think both forms have to work.
```

I don't care too much about this. Technically with JSON it's the last occurrence of an option that overrides previous occurrences. So in your example ``checkpoint=(wait=5)`` would cover all checkpoint configuration options (i.e ``log_size=1MB`` would be ignored). It'd be nice if we handled such options consistently, but I'm not going to lose sleep about it.

agorrod:
```
One thing I don't like at the moment: every time we call reconfigure, we start/stop the checkpoint and statistics-log server threads.
```

It took a number of iterations to get that right. I'd rather leave it be unless it's causing problems. AFAI can tell not being able to restart the server threads is the reason we can't reconfigure most of the logging functionality - which is another vote for leaving it the way it is.

I've heard of one MongoDB user who was regularly reconfiguring their cache size via reconfigure. The solution there was to tell them to stop since it didn't assist their use case.

test/suite/test_reconfig01.py
keithbostic: It was accidental, I didn't mean to do that. You're right, it's better the old way, I've reverted that part of the change.

src/os_win/os_path.c
keithbostic: I don't see the segfault, `__wt_isalpha(path[0])` will be false if the path is less than two characters long, so the test of `path[1]` won't happen. Am I missing something?

I agree with you about `C:tmp.txt`, the previous code was better, I'll revert that part of the change.

src/conn/conn_stat.c
keithbostic: Sure, done.

keithbostic:
@agorrod 

> Technically with JSON it's the last occurrence of an option that overrides previous occurrences. So in your example checkpoint=(wait=5) would cover all checkpoint configuration options (i.e log_size=1MB would be ignored). 

I agree it's the last occurrence of an option that overrides previous occurrences, but in the example, I don't think there's an implicit `log_size` option that somehow overrides a previous occurrence. Can you please point me to something that explains this? I'm still confused.

keithbostic:
@agorrod 
> It took a number of iterations to get that right. I'd rather leave it be unless it's causing problems.

My concern is checkpoints can take a long time, which would make it look like the `reconfigure` method has just hung, while it waits for the checkpoint server to finish.

Anyway, I agree, we can leave it alone until it's a problem.

keithbostic:
@agorrod, I believe I've addressed all of your review comments on this one.

agorrod:
Thanks @keithbostic I'll merge.

PR #2835, opened 2016-06-26
WT-2708 split child-update race with reconciliation/eviction

keithbostic:
@michaelcahill, this one is probably yours for review.


michaelcahill:
Thanks, @keithbostic -- this lgtm.  I've reviewed the other places where `ref->addr` is updated and don't see any where a race looks possible.  In particular, this is the only place I found that updates a shared `ref->addr` without setting `ref->state` to `WT_REF_LOCKED`.  I'll merge.

PR #2837, opened 2016-06-27
WT-2486 remove manydbs/smoke.sh; run only manydbs/t for make check

sulabhM:


sulabhM:
@agorrod : Can you please review this change

agorrod:
Thanks @sulabhM I like this change. Running ``make -j 12 check VERBOSE=1`` using ``libeatmydata`` still takes 9:13 on my primary development machine. I'd love to get that closer to 5 minutes.

Can you see any other changes we could investigate to further reduce the runtime before merging this?

sulabhM:
With these changes following is the reduction in the time taken:

 test/format  - 0:13  - from 0:28
 test/manydbs - 0:40 - from 2:46
 test/recovery - 0:16 - from from 0:33
 test/thread - 0:03 - from 0:40


make check on my laptop takes 4:52 now
with libeatmydata and "make -j 12" it takes 3:00

agorrod:
Thanks @sulabhM lgtm - I'll merge.

PR #2839, opened 2016-06-28
WT-2380 Add option to return an error code when tests fail

daveh86:


daveh86:
@agorrod , mind having a look at this?

PR #2840, opened 2016-06-28
WT-2026 Allow forced eviction to split pages when BTREE_NO_EVICTION set

sulabhM:


agorrod:
@sulabhM I like the changes here - could you take a run at updating and adding comments please?

sulabhM:
Passed pull-request checks on the latest change.
Have started format-stress tests on this branch. Will keep it running overnight to gain confidence in the change.

sulabhM:
@agorrod : Can you please review this change.

agorrod:
@sulabhM lgtm

@keithbostic Would you mind doing a review of this change as well? The intent is to allow force-eviction based in-memory splits to happen while a checkpoint is in the process of setting exclusive on, since we've seen cases where hot pages grow very large during that period. A deliberate consequence of this change is to stop doing reconciliation of pages pulled from the eviction queue, while a checkpoint is waiting for the queue to drain.

keithbostic:
@sulabhM, @agorrod, I've stared at this a bit, it seems like a good change, nice catch Sulabh!

1. It occurred to me we might not need a new flag: I've pushed a branch in #2845 to give you an idea of what I'm thinking. The reason I looked at this is:
    * It occurred to me that in this branch, the change in `__wt_page_in_func` has an unintended side-effect: if there's a page that cannot split in memory, but `WT_BTREE_NO_RECONCILE` is set, I think we'll execute parts of `__evict_force_check` we don't want, for example, `__wt_page_evict_soon`. I don't see problems in doing that, but I don't think it's what we want either.

    * I was thinking the check of `WT_BTREE_NO_RECONCILE`  in `__evict_review` should be replaced with `WT_BTREE_NO_EVICTION`, it's a better flag. In other words, "if we can't split, check if we can't evict", not "if we can't split, check if we're in this special state".

    So, I took a look at using just the `WT_BTREE_NO_EVICTION` flag, checking it in `__evict_force_check` and `__evict_review`, and I think it might be sufficient?

    Sulabh, I don't have your test case, so this is an entirely untested idea.

2. The name `WT_BTREE_NO_RECONCILE` bothers me, the flag is named "no-reconcile", I would read that as "no reconciliations period", not " you can do internal splits even if you can't evict". @agorrod, this is a knock-on from the fact that we named everything "evict", and then added in-memory splits in the evict path. I don't know, we could rename a couple of things "cache" instead of "evict" and make things a little better? Anyway, I'd like a different name, myself, but happy to be voted down on that one. :)

3. As I mentioned above, I think the check of `WT_BTREE_NO_RECONCILE`  in `__evict_review` should be replaced with `WT_BTREE_NO_EVICTION`, it's a better flag and has the same effect.

sulabhM:
@keithbostic : Regarding the naming of the flag, I am inclined to keep it as it is for now, i.e. WT_BTREE_NO_RECONCILE. What we need is a cleanup of the eviction code with a clear terminology setup. I have created a ticket WT-2740 to take care of that. We can come up with a better name for the flag as part of that ticket.

agorrod:
> The name WT_BTREE_NO_RECONCILE bothers me, the flag is named "no-reconcile", I would read that as "no reconciliations period", not " you can do internal splits even if you can't evict".

For completeness, I'll record my opinion here. We have different views of what reconciliation means, I think of reconciliation as the process of getting a page out to disk. So an in-memory split is something different. I think WT_BTREE_NO_EVICTION, means don't let the eviction server do anything with pages from the tree, including an in-memory split or a reconciliation.

You seem to think of reconciliation as transforming a page from one form to another and eviction as the process of writing a page out to disk.

I think both are reasonable but inaccurate models how the system actually works. Let's resolve this in WT-2740 as Sulabh suggests.

agorrod:
Thanks for this change @sulabhM I'm going to go ahead and merge. As discussed we can continue the naming conversation in another ticket.

PR #2841, opened 2016-06-28
WT-2733 Backport fixes for races between eviction and dead handle cleanup

agorrod:
Cherry picks of WT-2313 and WT-2434

### DIFF
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -247,30 +247,6 @@ err:	WT_TRET(__wt_rwlock_destroy(session, &dhandle->rwlock));
 }
 
 /*
- * __conn_dhandle_mark_dead --
- *	Mark a data handle dead.
- */
-static int
-__conn_dhandle_mark_dead(WT_SESSION_IMPL *session)
-{
-	bool evict_reset;
-
-	WT_ASSERT(session, F_ISSET(session, WT_SESSION_LOCKED_HANDLE_LIST));
-
-	/*
-	 * Handle forced discard (e.g., when dropping a file).
-	 *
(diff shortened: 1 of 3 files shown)


agorrod:
@michaelcahill or @keithbostic here is an attempt to backport WT-2313 and WT-2434, would you mind doing a review? I think it's worth comparing the version of ``__wt_evict_file_exclusive_on`` in this branch and the one at the tip of develop - there were some differences that got "sucked in" while I was doing the cherry pick.

src/conn/conn_dhandle.c
```diff
@@ -309,6 +286,13 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
 	__wt_spin_lock(session, &dhandle->close_lock);
 
 	/*
+	 * Ensure we aren't racing with the eviction server; inside the close
+	 * lock so threads won't race setting/clearing the tree's "no eviction"
+	 * flag.
+	 */
+	WT_ERR(__wt_evict_file_exclusive_on(session, &evict_reset));
+
+	/*
```
michaelcahill: I know that in develop we're doing this outside `close_lock`.  Is that just because we can with the new version of `exclusive_on/off`?

michaelcahill:
@agorrod, I had one question about `close_lock`, otherwise this lgtm.

src/conn/conn_dhandle.c
agorrod: There are cases where we race through sync_and_close that are protected by the close lock, so I'd rather be conservative and use the lock to avoid potential races through exclusive on as well.

PR #2842, opened 2016-06-30
WT-2665 Limit allocator fragmentation in WiredTiger

keithbostic:
@agorrod, ready to merge, for your review/consideration.


agorrod:
@keithbostic I've pushed a bunch of other changes into this branch. Would you mind doing a review and merging if you are happy?

I wasn't comfortable returning errors from some of the places you added WT_RET/WT_ERR. In the least they require careful review.

I was never able to reproduce the failure in test-unit-long, but I've run the Jenkins job against this branch enough times to be confident that the changes in this branch fix the bug.

keithbostic:
@agorrod, lgtm, I'll merge.

I understand your concern about returning errors in some of these paths, but I also really hate ignoring errors, and absent strong statements to the contrary, we're certainly going to add those error checks back in the next time the code is modified and someone (probably me!) notices a return code isn't checked.

A little outside the box, but what would you and @michaelcahill think of changing our locking functions to log an error and panic the system on failure, instead of returning an error code?

First, the locking error paths are entirely untested, and second, it would simplify error handling all over the system (for example,  `__wt_page_evict_soon` would now return a void).

It's unclear to me what we should do if `pthread_mutex_lock` fails anyway, that failure reasonably means things have gone entirely pear-shaped and stopping the run is the right thing to do.

PR #2843, opened 2016-07-01
WT-2380 Add option to s_all to return an error code when tests fail

keithbostic:
@daveh86, @agorrod, I replaced part of the 083b69b commit with a more portable construct.

It's fine with me if we want to commit to only running scripts under bash, but I think we should make that decision explicitly, and in that case, we should change the shell script interpreter from `/bin/sh` to `/bin/bash` to enforce it.

FTR, this broke under FreeBSD 10.2.

### DIFF
--- a/dist/s_all
+++ b/dist/s_all
@@ -17,8 +17,6 @@ force=
 reconf=0
 errmode=0
 errfound=0
-# Some tests just shouldn't return an error. We exclude them here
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @keithbostic, I tested on all the platforms I had access to, and didn't know how to get better confidence. I'll ask for your review explicitly next time.

PR #2845, opened 2016-07-01
WT-2026 Allow forced eviction to split pages when BTREE_NO_EVICTION set

keithbostic:
@sulabhM, for your review/consideration.

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -378,11 +378,14 @@ static int
 __evict_review(
     WT_SESSION_IMPL *session, WT_REF *ref, uint32_t *flagsp, bool closing)
 {
+	WT_BTREE *btree;
 	WT_CACHE *cache;
 	WT_DECL_RET;
 	WT_PAGE *page;
 	uint32_t flags;
 	bool modified;
+
+	btree = S2BT(session);
 
 	flags = WT_EVICTING;
 	if (closing)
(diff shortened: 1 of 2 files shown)


src/evict/evict_page.c
```diff
@@ -459,14 +462,22 @@ __evict_review(
...
 		 */
 		if (LF_ISSET(WT_EVICT_INMEM_SPLIT))
 			return (__wt_split_insert(session, ref));
+
+		/*
+		 * If we can't split the page, check if eviction is possible
+		 * for this file (eviction is turned off during some parts of
+		 * checkpoint).
+		 */
+		if (F_ISSET(btree, WT_BTREE_NO_EVICTION))
```
agorrod: @keithbostic The problem with this change is that the current code expects the ``WT_BTREE_NO_EVICTION`` flag to be ignored here when closing a file - @sulabhM is that correct?

sulabhM: @agorrod : That is correct. While closing a file we have WT_BTREE_NO_EVICTION set and we allow the reconciliation to happen. @keithbostic : If we return error from this point if WT_BTREE_NO_EVICTION is set, we will unintentionally fail for file-close case as well.

keithbostic: @sulabhM we have the `closing` flag, is checking it here sufficient?

agorrod:
I'm going to close this branch - I prefer the other solution, and think there is less risk.

PR #2846, opened 2016-07-02
WT-2739 pluggable file systems documentation cleanups

keithbostic:
Document that readonly pluggable file systems may fail attempts to open
files read/write with EACCES.

Document that pluggable file systems should fail attempts to open
non-existent files with ENOENT.


agorrod:
Thanks @keithbostic lgtm.

PR #2847, opened 2016-07-03
WT-2730 column-store incorrectly matches key slots on new pages

keithbostic:


agorrod:
Thanks @keithbostic lgtm - nice catch.

PR #2848, opened 2016-07-04
WT-2730 column-store incorrectly matches key slots on new pages

keithbostic:
@agorrod, I think `WT_CURSOR_BTREE.rip_saved` potentially has the same problem.

I've never seen it happen, this is just by inspection.


agorrod:

PR #2850, opened 2016-07-05
WT-2743 Fixup new I/O thread count statistics.

agorrod:
The increment and decrement were backward, so they never reported positive values.
While I was there make the statistics macros consistent.


agorrod:
@sulabhM Would you mind reviewing this change please? I tested with the wtperf workload from WT-2389

sulabhM:
LGTM

PR #2851, opened 2016-07-05
WT-2744 Detect and ignore partial line.

sueloverso:
@keithbostic Please review this change to the random-abort test to check for a partial line.

### DIFF
--- a/test/recovery/random-abort.c
+++ b/test/recovery/random-abort.c
@@ -91,7 +91,8 @@ thread_run(void *arg)
 	if ((fp = fopen(buf, "w")) == NULL)
 		testutil_die(errno, "fopen");
 	/*
-	 * Set to no buffering.
+	 * Set to line buffering.  But that is advisory only.  We've seen
+	 * cases where the result files end up with partial lines.
 	 */
 	__wt_stream_set_line_buffer(fp);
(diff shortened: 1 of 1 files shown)


test/recovery/random-abort.c
```diff
@@ -317,12 +318,23 @@ main(int argc, char *argv[])
...
+			 * If we're unlucky, the last line may be a partially
+			 * written key at the end that can result in a false
+			 * negative error for a missing record.  Detect it.
+			 */
+			if (last_key != UINT64_MAX && key != last_key + 1) {
+				printf("%s: Ignore partial record %" PRIu64
+				    " last valid key %" PRIu64 "\n",
+				    fname, key, last_key);
+				break;
+			}
```
keithbostic: @sueloverso, why not use `fgets` to retrieve the lines, because then you can explicitly check for a `newline` character? That seems simpler than depending on incrementing keys.

Regardless, if you like this approach better, lgtm, just go ahead and merge it in.

sueloverso: I coded this up and I am going to leave it as-is.  Testing with `fgets` and looking for the newline runs much slower than the `fscanf` and the key check.  Running 3 times on the failed data, the `fscanf` took between 2.1-2.5 seconds per rn.  The `fgets` took 20-33 seconds per run.  This is test code, and that is sufficient reason.

PR #2852, opened 2016-07-05
WT-2728 Only verify the log file header during recovery.

sueloverso:
@agorrod Please review this change.  It is a small piece to reduce an unnecessary read/verify of the log file header on every `log_flush` call.


agorrod:

PR #2853, opened 2016-07-05
WT-2746 track checkpoint I/O separately from eviction I/O

keithbostic:
Separate cache write statistics into checkpoint and non-checkpoint I/O.


src/btree/bt_io.c
```diff
@@ -365,6 +366,12 @@ __wt_bt_write(WT_SESSION_IMPL *session, WT_ITEM *buf,
 	WT_STAT_FAST_DATA_INCR(session, cache_write);
 	WT_STAT_FAST_CONN_INCRV(session, cache_bytes_write, dsk->mem_size);
 	WT_STAT_FAST_DATA_INCRV(session, cache_bytes_write, dsk->mem_size);
+	if (checkpoint_io) {
```
agorrod: @keithbostic This is useful and I definitely want to keep it, but it can be misleading if the application has compression enabled. I know it's a layering violation to figure this out in the block manager, but it's the only place we know the actual disk-block write size (unless it's hidden in the dsk structure) at the moment.

keithbostic: I thought you only wanted the relationship between bytes written for checkpoint vs. bytes written for other purposes. How does having the absolute bytes written to disk instead of bytes pushed out of cache improve that information?

Regardless, we can know the disk-block write size in `bt_io.c` (and in fact, we have to figure it out there, for compressed and encrypted blocks), that's what the `bm->write_size()` call does.

agorrod: > How does having the absolute bytes written to disk instead of bytes pushed out of cache improve that information?

Often I and other people compare byte written numbers to raw disk numbers. Especially since the iostat numbers are available in the MongoDB diagnostic data.

I hadn't realized that ``bm->write_size()`` exists, that seems like a good solution to me. We'll probably miss counting the root page, but that's OK. I'll add a commit to this ticket.

agorrod: @keithbostic on closer inspection, I don't think ``write_size`` is what I was looking for. I've added more detail here: https://jira.mongodb.org/browse/WT-2746?focusedCommentId=1316325&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1316325

keithbostic:
@agorrod, I haven't done anything other than smoke-tests on this branch, but if it works for you, I'm happy with it.

(Note we don't track checkpoint I/O per file in this model, it's a block-manager statistic now.)

src/block/block_write.c
```diff
@@ -235,7 +235,7 @@ __wt_block_write(WT_SESSION_IMPL *session, WT_BLOCK *block,
 int
 __wt_block_write_off(WT_SESSION_IMPL *session, WT_BLOCK *block,
     WT_ITEM *buf, wt_off_t *offsetp, uint32_t *sizep, uint32_t *cksump,
-    bool data_cksum, bool caller_locked)
+    bool data_cksum, bool checkpoint_io, bool caller_locked)
```
sueloverso: Three bool args in a row is beyond my threshold for converting it into a `flags` arg.  (My threshold is two so both `wt_block_write` and `wt_block_write_off` could use flags, IMO.)

keithbostic: I'm sympathetic, but:

* two of the booleans (`data_cksum` & `checkpoint_io`) are shared by the btree layer, and one (`caller_locked`) is private to the block manager layer,
* the block-manager's checkpoint routine takes one of the public booleans (`data_cksum`), but not both of them,
* if there's ever another block manager, any flags would have to be shared between block managers.

We could change the block-manager's checkpoint API take `checkpoint_io` too, but we already know any I/O done for that API is checkpoint related (and the other arguments to the block-manager's checkpoint  API aren't common with the block-manager's write API).

It's a mess, but it's only a mess inside the block manager, so I went with the booleans.

Let's see what @agorrod thinks.

agorrod: My preference is with @sueloverso from a code review, but you are much more familiar with the layering in the code, so I'm happy to go with your decision.

agorrod:
Thanks @keithbostic these changes look good to me, and I've verified that they generate the output I expect. I'm a bit torn as to whether WRITE_LEAVES should be included as checkpoint I/O, but I think you are right and it should be.

PR #2854, opened 2016-07-06
WT-2751 column-store statistics incorrectly calculates the number of entries

keithbostic:
> Also Alex made some changes to delete some keys and then use wt utility to get stats on the database, and the wt utility shows wrong number of key/val pairs.

@sulabhM, @agorrod, this is the fix for the wt utility mistake.


agorrod:

PR #2855, opened 2016-07-06
WT-2752 Fix errors in wtperf config

sueloverso:
@agorrod Please review these changes to the new zipfian workload.  Looking at that config, I believe there are errors and these changes reflect the actual intent.


agorrod:
Thanks @sueloverso lgtm.

PR #2856, opened 2016-07-07
WT-1162 Add a latency specific test

sueloverso:
@agorrod This branch only contains a new wtperf config.  I have added some commands to the Jenkins perf-stress job that are commented out.  Once this config exists, I want to turn that on and create a latency plot.  This is the config from WT-2705 primarily.  I used a warning threshold of a quarter second, which is commonly hit on my AWS box (same instance type as the perf jobs).  Please review.


agorrod:

PR #2857, opened 2016-07-08
WT-2711 Change statistics log configuration options

keithbostic:
If there's no server running, discard any configuration information so
we don't leak memory during reconfiguration.


sueloverso:
LGTM.

PR #2858, opened 2016-07-08
WT-2755 flexelint configuration treats size_t as 4B type

keithbostic:
My flexelint configuration thought size_t was 4B, fix it and clean up the result.


agorrod:

PR #2859, opened 2016-07-09
WT-2693 Check open_cursor error paths for consistent handling

keithbostic:
@sueloverso, I think this one is probably yours -- there's some tricky stuff in here, happy to talk through anything that's confusing.

### DIFF
--- a/src/cursor/cur_backup.c
+++ b/src/cursor/cur_backup.c
@@ -83,8 +83,13 @@ __curbackup_close(WT_CURSOR *cursor)
 	CURSOR_API_CALL(cursor, session, close, NULL);
 
 	/*
-	 * If we got far enough to hold the hot backup lock, then we own all
-	 * of the necessary cleanup (on error or success).
+	 * When starting a hot backup, we serialize hot backup cursors and set
+	 * the connection's hot-backup flag. Once that's done, we set the
+	 * cursor's backup-locker flag, implying the cursor owns all necessary
+	 * cleanup (including removing temporary files), regardless of error or
+	 * success. The cursor's backup-locker flag is never cleared (it's just
+	 * discarded when the cursor is closed), because that cursor will never
+	 * not be responsible for cleanup.
 	 */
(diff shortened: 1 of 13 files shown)


src/cursor/cur_backup.c
```diff
@@ -82,14 +82,20 @@ __curbackup_close(WT_CURSOR *cursor)
 
 	CURSOR_API_CALL(cursor, session, close, NULL);
 
-	WT_TRET(__backup_cleanup_handles(session, cb));
+	/*
+	 * If we got far enough to hold the hot backup lock, then we own all
+	 * of the necessary cleanup (on error or success).
+	 */
```
sueloverso: I'd like a sentence here or where we set the `WT_CURBACKUP_LOCKER` flag that explicitly states that it will never be cleared.  It gets "cleared" implicitly by calling `wt_cursor_close` which frees the cursor.

src/include/cursor.h
```diff
@@ -413,7 +416,9 @@ struct __wt_cursor_log {
 	uint32_t	step_count;	/* Intra-record count */
 	uint32_t	rectype;	/* Record type */
 	uint64_t	txnid;		/* Record txnid */
-	uint32_t	flags;
+
+#define	WT_CURLOG_ARCHIVE_LOCK	0x01	/* Archive lock held */
+	u_int		flags;
```
sueloverso: `flags` are always sized and we have a 64-bit version.  I'd prefer these be/remain `uint32_t`.

sueloverso:
@keithbostic I have gone through this and it LGTM.  I think it is much cleaner and consistent error handling.  I didn't see any holes or issues calling close on the error paths.  I had a minor comment or two.

src/cursor/cur_backup.c
keithbostic: Sure, done & pushed.

src/include/cursor.h
keithbostic: Sure, done & pushed. (I used uint8_t because that's what the other cursors used.)

PR #2860, opened 2016-07-09
WT-2688 configure --enable-python doesn't check for availability of swig

keithbostic:
@sulabhM, @agorrod, swig 3.0 is out.

I upgraded the swig autoconf archive package and added a check for swig 3.0.


keithbostic:
Move these changes to their own JIRA ticket (WT-2756), delete this branch in favor of #2861.

PR #2861, opened 2016-07-09
WT-2756 the autoconf archive package doesn't check for swig 3.0.

keithbostic:
@sulabhM, @agorrod, swig 3.0 is out.

I upgraded the swig autoconf archive package and added a check for swig 3.0.


agorrod:

PR #2862, opened 2016-07-09
WT-2648 cache-line alignment for new ports

keithbostic:
@agorrod, I think this is all we need to do for this ticket.

We might want to consider if we're wasting too much space at some point (the zSeries alignment is 256B, and we align transaction states, for example), but for now I think we should make this change.

### DIFF
--- a/src/include/hardware.h
+++ b/src/include/hardware.h
@@ -45,7 +45,16 @@
 	    &(p)->flags_atomic, __orig, __orig & ~(uint8_t)(mask)));	\
 } while (0)
 
-#define	WT_CACHE_LINE_ALIGNMENT	64	/* Cache line alignment */
+/*
+ * Cache line alignment.
+ */
+#if defined(__PPC64__) || defined(PPC64)
+#define	WT_CACHE_LINE_ALIGNMENT	128
+#elif defined(__s390x__)
+#define	WT_CACHE_LINE_ALIGNMENT	256
+#else
+#define	WT_CACHE_LINE_ALIGNMENT	64
+#endif
 #define	WT_CACHE_LINE_ALIGNMENT_VERIFY(session, a)			\
 	WT_ASSERT(session,						\
 	    WT_PTRDIFF(&(a)[1], &(a)[0]) >= WT_CACHE_LINE_ALIGNMENT &&	\


agorrod:
@keithbostic I agree that this change looks correct. Should we take this opportunity to consider which structures set explicit alignment (there is an [undefined behavior sanitizer](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) mode that could help)? Is alignment always for performance, or is there a correctness motivation? If it is for performance, there could be a significant space penalty associated with a 256 byte alignment - should we measure the trade off?

keithbostic:
I don't believe alignment is ever necessary for correctness, it's only for performance.

The structures we align, the base size, and how many we allocate:
```
WT_LOGSLOT, base size is 120B, we allocate 128 (#define WT_SLOT_POOL)
WT_SESSION_IMPL, base size is 792B, we allocate session-max + 10
WT_SPINLOCK, base size is 24B, we allocate ~40
WT_TXN_STATE, base size is 16B, we allocate session-max + 10
```

The interesting ones are the WT_SESSION_IMPL and WT_TXN_STATE structures, because of the number of sessions, MongoDB currently sets session-max to 20,000.

```
((20010 * 1024) - (20010 * 792)) / 2^20 == 4.427MB of 15.113MB, 23%
((20010 *  256) - (20010 *  16)) / 2^20 == 4.579MB of  0.305MB, 94%
```

So, at 256B alignment, MongoDB wastes 4.4MB on WT_SESSION_IMPL structures and 4.6MB on WT_TXN_STATE structures.

I don't like throwing away memory, but on the other hand, I doubt it's significant.

I asked @michaelcahill awhile back why we didn't merge the WT_TXN_STATE structure into the WT_SESSION_IMPL structure, but I honestly don't remember why he wanted them separate. That would save us about 4MB in this case.

tl;dr: I doubt there's a significant space penalty. Merging the WT_TXN_STATE into the WT_SESSION_IMPL structure would save us 4MB.

michaelcahill:
@keithbostic, thanks, lgtm.  Those number don't look too scary to me: the machines in question typically have hundreds of GB of RAM.  I'll merge this one, we will need a new ticket to fix the undefined behavior issues by actually using the cache alignment during allocations.

PR #2863, opened 2016-07-09
WT-2578 remove write barriers from the TAILQ_INSERT_XXX macros

keithbostic:
@agorrod, @michaelcahill confirmed we shouldn't need write barriers in the TAILQ macros any more, so I've upgraded to the current FreeBSD sys/queue.h file and removed the write-barriers we had in those macros.

I reviewed our TAILQ macro usage and don't see any problems, but Michael suggested the PPC test instance on Jenkins would do a good job of figuring out if we're relying on these barriers anywhere. Maybe we could run this branch overnight on that platform?


michaelcahill:
Thanks @keithbostic, this lgtm and stress tests have run on PPC and zSeries overnight.  I'll merge..

PR #2865, opened 2016-07-11
WT-2711 Remove posix expanded strftime values and use older C89 values

daveh86:


daveh86:
@agorrod , this fixes the windows compilation issue.

daveh86:
@keithbostic would you mind taking a look at this?

I had to switch out the %F and %T as they cause MSVC13 to segfault. I've replaced them with the "unpacked" versions. Are you OK with the straight replacement? Or would you prefer if we over-ride in the os_win code?

keithbostic:
Thank you @daveh86, that looks great to me. I checked the MongoDB sources and they uses the same, older format strings.

PR #2866, opened 2016-07-11
WT-2759 Releasing the hot-backup lock doesn't require the schema lock.

keithbostic:
@agorrod, for your consideration/review.

FTR, here's the original commit: 69bc7d1


agorrod:

PR #2868, opened 2016-07-11
WT-2760 Backup file removal requires ordering, so needs to sync the directory

keithbostic:
@agorrod, @michaelcahill, something occurred to me today, I thought I'd run it by you.

As you know, we have a file-system directory-sync method, provided for Linux filesystems that require the enclosing directory to be synced for durability during file create, remove and rename. Currently, only Linux needs the directory-sync method.

 First, by having a separate method, we're making it significantly harder to write pluggable filesystems. Imagine a pluggable filesystem that has to do something for durability, but it's nothing like directory-sync. That filesystem can either not specify a directory-sync method and make all operations durable (which might be expensive), or it can specify a directory-sync method and somehow match WiredTiger's call of the directory-sync method to a previous filesystem method call. (It would have to somehow match the path names, there's nothing else to use.)

Second, currently we call `__wt_filename()` (or a variant), to create the pathnames we're going to open, remove or rename. We then re-create that pathname in order to call the directory-sync method. Staring at the two chunks of code, I think they're correct (but they're not identical). It would be much cleaner to only create a single path and use it both for the operation and for any subsequent directory flush.

I suggest we get rid of the directory-sync method, and replace it with boolean "durable" flags to the open, remove and rename methods; here's the branch.

### DIFF
--- a/examples/c/ex_file_system.c
+++ b/examples/c/ex_file_system.c
@@ -124,12 +124,10 @@ static int demo_fs_directory_list(WT_FILE_SYSTEM *, WT_SESSION *,
     const char *, const char *, char ***, uint32_t *);
 static int demo_fs_directory_list_free(
     WT_FILE_SYSTEM *, WT_SESSION *, char **, uint32_t);
-static int demo_fs_directory_sync(WT_FILE_SYSTEM *file_system,
-    WT_SESSION *session, const char *directory);
 static int demo_fs_exist(WT_FILE_SYSTEM *, WT_SESSION *, const char *, bool *);
-static int demo_fs_remove(WT_FILE_SYSTEM *, WT_SESSION *, const char *);
+static int demo_fs_remove(WT_FILE_SYSTEM *, WT_SESSION *, const char *, bool);
 static int demo_fs_rename(
-    WT_FILE_SYSTEM *, WT_SESSION *, const char *, const char *);
+    WT_FILE_SYSTEM *, WT_SESSION *, const char *, const char *, bool);
 static int demo_fs_size(
     WT_FILE_SYSTEM *, WT_SESSION *, const char *, wt_off_t *);
 static int demo_fs_terminate(WT_FILE_SYSTEM *, WT_SESSION *);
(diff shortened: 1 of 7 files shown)


src/os_win/os_fs.c
```diff
@@ -36,13 +36,14 @@ __win_fs_exist(WT_FILE_SYSTEM *file_system,
...
-__win_fs_remove(
-    WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session, const char *name)
+__win_fs_remove(WT_FILE_SYSTEM *file_system,
+    WT_SESSION *wt_session, const char *name, bool durable)
 {
 	DWORD windows_error;
 	WT_SESSION_IMPL *session;
 
 	WT_UNUSED(file_system);
+	WT_UNUSED(durable);
```
sueloverso: My only minor nit is to alphabetize these WT_UNUSED calls (this one and the one in rename).

examples/c/ex_file_system.c
```diff
@@ -124,12 +124,10 @@ static int demo_fs_directory_list(WT_FILE_SYSTEM *, WT_SESSION *,
...
 static int demo_fs_directory_list_free(
     WT_FILE_SYSTEM *, WT_SESSION *, char **, uint32_t);
-static int demo_fs_directory_sync(WT_FILE_SYSTEM *file_system,
-    WT_SESSION *session, const char *directory);
 static int demo_fs_exist(WT_FILE_SYSTEM *, WT_SESSION *, const char *, bool *);
-static int demo_fs_remove(WT_FILE_SYSTEM *, WT_SESSION *, const char *);
+static int demo_fs_remove(WT_FILE_SYSTEM *, WT_SESSION *, const char *, bool);
 static int demo_fs_rename(
-    WT_FILE_SYSTEM *, WT_SESSION *, const char *, const char *);
+    WT_FILE_SYSTEM *, WT_SESSION *, const char *, const char *, bool);
```
sueloverso: Can we imagine any other possible conditionals for remove and/or rename?  I.e. instead of a bool maybe we start with some WT_FS_ flags similar to the new open flag.

src/os_win/os_fs.c
keithbostic: It's function argument order instead of alphabetical order.

sueloverso: I guess I've been doing it wrong all this time!

examples/c/ex_file_system.c
keithbostic: I don't have any other conditionals in mind at the moment.

My biggest reason to resist this change is flags would require documentation, and booleans don't.

PR #2869, opened 2016-07-12
SERVER-24971 Don't cache buffers after application eviction.

michaelcahill:


agorrod:
Thanks @michaelcahill this change looks good to me. I've been doing some testing of different allocators in our automated perf testing, so I'd like to hold off merging this change until those have settled.

PR #2871, opened 2016-07-12
WT-2762 Handle 'Y' from wtperf monitor file.

sueloverso:
@keithbostic Please review this small change to handle a Y in the wtperf monitor file.


keithbostic:
LGTM.

PR #2872, opened 2016-07-13
WT-2763 Change test ID's in test_intpack to fix OSX test errors

daveh86:


daveh86:
@agorrod , mind taking a look?

agorrod:
@daveh86 The changes look OK to me here, I'd rather there was a comment and clearer names.

There is also the "check_scenarios" function that checks for duplicate names. It'd be worthwhile expanding the implementation of "check_scenarios" to do case insensitive comparisons, if that is a constraint on OS X.

agorrod:
Thanks @daveh86 lgtm.

PR #2873, opened 2016-07-13
WT-2765 wt dump: indices need to be shown in the dump output

ddanderson:
This fixes a problem where a table with 1 or more indices, and no column groups would not have any dump output lines for the indices.  Added associated tests for dump and JSON dump.


ddanderson:

agorrod:

PR #2875, opened 2016-07-14
WT-2725 Enforce visibility constraints before evicting a page for mongodb-3.0.

agorrod:
Otherwise there is a race between when we add a page to the queue
and when the page is pulled off the queue - where the updates on
the page may have changed, thus violating transactional constraints.


michaelcahill:
Thanks @agorrod, this lgtm and brings the 3.0 logic inline with 3.2+.  I'll merge.

PR #2877, opened 2016-07-14
WT-2738 checkpoint=(name=FOO) configuration

keithbostic:
Remove the checkpoint=(name=FOO) configuration.


agorrod:

PR #2878, opened 2016-07-14
WT-2767 in test/suite/run.py, add -s N option to run an individual scenario

ddanderson:
Several complimentary changes to make life easier for testing:
- Added -s N option to restrict the run to a scenario with a given number or name.
- When showing the name of the test (for example, with -v 2, or when a test fails), if there is a scenario, print as 'fulltestname -s scenario_number' so it can be easily copied for another command run.
- Created a 'testname.txt' file in the test directory with the name of the test (and '-s NNN', if appropriate).
- Fixed the ordering of the symbolic name of a scenario (e.g. 'file.recno.hex'), it was reversed before.

### DIFF
--- a/test/suite/run.py
+++ b/test/suite/run.py
@@ -108,10 +108,10 @@ def restrictScenario(testcases, restrict):
     elif restrict.isdigit():
         s = int(restrict)
         return [t for t in testcases
-                if hasattr(t, 'scenario_number') and t.scenario_number == s]
+            if hasattr(t, 'scenario_number') and t.scenario_number == s]
     else:
         return [t for t in testcases
-                if hasattr(t, 'scenario_name') and t.scenario_name == restrict]
+            if hasattr(t, 'scenario_name') and t.scenario_name == restrict]
 
 def addScenarioTests(tests, loader, testname, scenario):
     loaded = loader.loadTestsFromName(testname)
--- a/test/suite/run.py
+++ b/test/suite/run.py
@@ -87,6 +87,7 @@ Options:\n\
   -j N    | --parallel N         run all tests in parallel using N processes\n\
   -l      | --long               run the entire test suite\n\
   -p      | --preserve           preserve output files in WT_TEST/<testname>\n\
+  -s N    | --scenario N         use scenario N (N can be number or symbolic)\n\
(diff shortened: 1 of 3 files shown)


agorrod:
Thanks @ddanderson this looks like a great change to me.

@daveh86 Could you review the functionality changes here (especially the name ordering), and check our Jenkins jobs to make sure none of them will require follow on changes.

daveh86:
@agorrod , I've had a look and this change makes sense overall, but I have a few questions for @ddanderson. 

Is it possible to restrict the new -s option to only work when a test is specified? I don't see this as a blocker but running the following command seemed to produce an unintuitive result
```LD_LIBRARY_PATH=.libs/ python ../test/suite/run.py -v 2 -j 8 -s 1```

I also had a dive into the various names/numbers of scenario's that this change is keying off, it looks like a very large number of scenarios have no scenario_name or scenario_number set (so cannot be referenced, like test_async02). There are also a number of scenarios that seem to have numeric names (test_backup03 for example) that can cause a logical conflict.

None of these are blockers, but there is potentially some follow up work on out tests to ensure that  all tests have a scenario name and number so that they can be keyed by the -s option in either numeric or symbolic mode.

ddanderson:
@daveh86, I pushed a change to restrict the -s to when a test is specified.

You're right about the number of tests that don't have numbered/named scenarios.  The number_scenarios() function must be called to tag each scenario with a name/number.  That's not always done. Between that function, and check_scenarios() and multiply_scenarios(), it's kind of a confusing mess.  I think I'd like to replace them all with a make_scenarios(), which handles all of these, including the multiplied scenarios case.  Then we can call just one function, and do it every time.  That's a change that will touch pretty much the entire test directory, so I'd like to split that out into a followup JIRA.

sueloverso:
@ddanderson I think you're right.  Certainly I do not really know the difference between all the different scenario functions.  My workflow for a test is to find one that does something similar, copy it and make my specific changes.  So it is generally just an artifact which one gets used based on whatever function the original one used!

ddanderson:
Created WT-2778 to track the scenario function cleanup.

ddanderson:
I'm self reviewing the change to require a test name when -s is used.

PR #2874, opened 2016-07-13
WT-2737 Scrub dirty pages rather than evicting them

keithbostic:
@agorrod, this isn't for merge, I'm just creating a branch for us to talk about.

PR #2881, opened 2016-07-17
WT-2774 minor cleanups/improvements

keithbostic:
@michaelcahill, I realized I had a bunch of random stuff tucked away in #2874 (and that branch is probably never going to be merged), so I pulled those changes out into this branch.

It's mostly noise, but I do want the enhanced `__wt_cache_dump` output (33e1aca) and the rework of your e41c3e5 commit, in c608c00.

(It occurred to me it would be better to clean up the reconciliation structures when reconciliation finishes, rather than in `__wt_evict`, and then it occurred to me we were already doing that, so I went looking. Long-story short, I merged your change -- which would repeat some of the cleanup we were already doing in reconciliation -- into the existing cleanup code in `__wt_reconcile`. Sorry for not noticing that when you first asked me for review.)

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1886,13 +1886,15 @@ __wt_cache_dump(WT_SESSION_IMPL *session, const char *ofile)
 	WT_DATA_HANDLE *dhandle, *saved_dhandle;
 	WT_PAGE *page;
 	WT_REF *next_walk;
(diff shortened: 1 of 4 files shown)


michaelcahill:
autobuild: retest this please

michaelcahill:
@keithbostic, thanks, these change lgtm.  I'll merge.

PR #2882, opened 2016-07-18
WT-2773 search_near in an index should find exact matches.

michaelcahill:
More specifically, if search_near is called with a key that appears in
an index, the cursor should be positioned on the first matching key and
`cmp` should be set to zero.


michaelcahill:
@ddanderson, can you please review?

ddanderson:
Hi @michaelcahill, might be good, as a followup, to have a single shared inline function for search and search_near.  Otherwise, this LGTM.

PR #2884, opened 2016-07-18
WT-2769 Update documentation to reflect correct limits of memory_page_max

sulabhM:


sulabhM:
@agorrod : Please take a look

agorrod:
Thanks @sulabhM. The change looks good to me - could you also add a comment into the source code where the limits are calculated, letting the user that they should update the API documentation if they make changes in the future?

sulabhM:
@agorrod : Sure will update the pull request with the comment added.

sulabhM:
@agorrod : Please take a look.

PR #2885, opened 2016-07-18
WT-2731 Finer adjustment for page size with raw compression

keithbostic:
clang 3.4.1 doesn't like this:
```
rec_write.c:1975:46: error: implicit conversion loses integer precision: 'unsigned long' to
'uint32_t' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
```

Also, minor cleanup in `__wt_evict()`.


sueloverso:
LGTM.  I'll merge once tests complete.

PR #2886, opened 2016-07-19
WT-2779 Add an assertion that in-memory pages don't grow too large.

agorrod:


agorrod:
We don't want this assertion to be so strong. For example it's failing in a Python test that inserts large records and configures a small memory_page_max.

PR #2887, opened 2016-07-19
WT-2403 Add next_random LSM cursor

daveh86:


daveh86:
retest this please

daveh86:
@agorrod , would you mind having a look at this?

agorrod:
@daveh86 Did you try the MongoDB test that triggered this work originally? Do the changes in this branch make that test pass?

daveh86:
I had to remove the code which caused the test to skip on LSM to make the test run, but after that yes!

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,80 @@ err:	__clsm_leave(clsm);
...
+static int
+__clsm_next_random(WT_CURSOR *cursor)
+{
+	WT_CURSOR_LSM *clsm;
+	WT_CURSOR *c;
+	WT_DECL_RET;
+	WT_LSM_CHUNK *chunk;
+	WT_SESSION_IMPL *session;
+	uint64_t checked_docs, i, rand_chunk, rand_doc, total_docs;
+	const char *cfg[3];
```
sulabhM: Should const char be placed before uint64_t wrt alphabetic ordering.

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,80 @@ err:	__clsm_leave(clsm);
...
+	WT_CURSOR_NOVALUE(cursor);
+	WT_ERR(__clsm_enter(clsm, false, false));
+
+	cfg[0] = WT_CONFIG_BASE(session, WT_SESSION_drop);
+	cfg[1] = "next_random=true";
+	cfg[2] = NULL;
+
+retry:	total_docs = 0;
+	/* We need to know roughly how many docs are in the tree */
+	for (i = clsm->lsm_tree->nchunks; i > 0; i--) {
```
sulabhM: Would this be more readable:
for (i = 0; i < clsm->lsm_tree->nchunks; i++) {
	total_docs += clsm->lsm_tree->chunk[i]->count;
}

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,80 @@ err:	__clsm_leave(clsm);
...
+		total_docs += clsm->lsm_tree->chunk[i-1]->count;
+	}
```
sulabhM: Would this be more readable:
{code}
for (i = 0; i < clsm->lsm_tree->nchunks; i++) {
       total_docs += clsm->lsm_tree->chunk[i]->count;
}
{code}

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,80 @@ err:	__clsm_leave(clsm);
...
+
+	/*
+	 * We pick one of the docs at random from the tree to pick a random
+	 * chunk, weighted by the size of the chunks.
+	 */
+	rand_doc = __wt_random(&session->rnd);
```
sulabhM: Can this be:
rand_doc = __wt_random(&session->rnd) % total_docs;

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,80 @@ err:	__clsm_leave(clsm);
...
+	rand_doc = rand_doc % total_docs;
+	checked_docs = rand_chunk = 0;
+	for (i = clsm->lsm_tree->nchunks; i > 0; i--) {
```
sulabhM: Same as before, can use a count up for loop.

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,79 @@ err:	__clsm_leave(clsm);
...
+	uint64_t checked_docs, i, rand_doc, total_docs;
+	const char *cfg[3];
+
+	clsm = (WT_CURSOR_LSM *)cursor;
+
+	CURSOR_API_CALL(cursor, session, next, NULL);
+	WT_CURSOR_NOVALUE(cursor);
+	WT_ERR(__clsm_enter(clsm, false, false));
+
+	cfg[0] = WT_CONFIG_BASE(session, WT_SESSION_drop);
```
agorrod: {{WT_SESSION_drop}} looks out of place here - is it correct? I'd expect setting up this configuration stack to be more like the one in {{\__wt_lsm_merge}} since the configuration string we are adding is always constant.

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,79 @@ err:	__clsm_leave(clsm);
...
+	cfg[1] = "next_random=true";
+	cfg[2] = NULL;
+
+	/* Initialize to the 0th chunk */
+	chunk = clsm->lsm_tree->chunk[0];
+
+	/* We need to know roughly how many docs are in the tree */
+retry:	for (total_docs = i = 0; i < clsm->lsm_tree->nchunks; i++) {
```
agorrod: It's better to avoid goto - can you restructure this so it uses a function call instead?

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,79 @@ err:	__clsm_leave(clsm);
...
+	CURSOR_API_CALL(cursor, session, next, NULL);
+	WT_CURSOR_NOVALUE(cursor);
+	WT_ERR(__clsm_enter(clsm, false, false));
+
+	cfg[0] = WT_CONFIG_BASE(session, WT_SESSION_drop);
+	cfg[1] = "next_random=true";
+	cfg[2] = NULL;
+
+	/* Initialize to the 0th chunk */
+	chunk = clsm->lsm_tree->chunk[0];
```
agorrod: I'm not clear why this initialization is here - the following for loop doesn't seem to require it. It would be better if {{chunk}} had a more descriptive name - AFAI can see it is the randomly selected chunk we are going to search within.

src/lsm/lsm_cursor.c
```diff
@@ -889,6 +889,79 @@ err:	__clsm_leave(clsm);
...
+		total_docs += clsm->lsm_tree->chunk[i]->count;
+	}
+
+	/*
+	 * We pick one of the docs at random from the tree to pick a random
+	 * chunk, weighted by the size of the chunks.
+	 */
+	rand_doc = __wt_random(&session->rnd) % total_docs;
+	for (checked_docs = i = 0; i < clsm->lsm_tree->nchunks; i++) {
+		checked_docs += clsm->lsm_tree->chunk[i]->count;
```
agorrod: This code appears to assume that the LSM tree structure doesn't change - I believe we need to lock the LSM tree to ensure that.

Given that we'll need to hold a lock the code that traverses the LSM tree should be structured in such a way that it is as fast as possible.

src/lsm/lsm_cursor.c
```diff
@@ -1576,6 +1649,14 @@ __wt_clsm_open(WT_SESSION_IMPL *session,
 	 */
 	clsm->dsk_gen = 0;
 
+	/* If the next_random option is set, configure a random cursor */
+	WT_ERR(__wt_config_gets_def(session, cfg, "next_random", 0, &cval));
+	if (cval.val != 0) {
+		__wt_cursor_set_notsup(cursor);
+		cursor->next = __clsm_next_random;
+		cursor->reset = __clsm_reset;
```
agorrod: Replacing the reset function pointer looks redundant to me.

daveh86: You're right. The notsup function makes the reset a noop, this doesn't matter as reset is redundant on a random cursor.

src/lsm/lsm_cursor.c
daveh86: Understood re: locks. I see two methods of approaching this problem:
Loop through twice under the lock
Loop through once, creating a static "copy" of the chunks array as it is "now" and then iterate over that

I'm not sure which would be faster in this case. I lean towards the former but am not sure how to judge this.

src/lsm/lsm_cursor.c
daveh86: Correct, re: Drop. This has to be more like the cfg from {{__wt_lsm_merge}} rather than drop_cfg as cursor will not have a valid session until after CURSOR_API_CALL.

michaelcahill:
Status update: waiting on additional tests to be added.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,104 @@ err:	__clsm_leave(clsm);
...
+	cfg[1] = "next_random=true";
+	cfg[2] = NULL;
+
+	do {
+		WT_ERR(__clsm_random_chunk(session, clsm, &random_chunk));
+		/*
+		 * Open a cursor on the chunk and pick a random key from within.
+		 */
+		WT_ERR(__wt_open_cursor(
+		    session, random_chunk->uri, NULL, cfg, &c));
```
sulabhM: Nothing prevents random_chunk to have been freed by the time you call __wt_open_cursor() on it. Need to look into it more to figure out if that is possible or how to prevent that.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,104 @@ err:	__clsm_leave(clsm);
...
+		ret = c->next(c);
+		/*
+		 * Sometimes we may run on an empty chunk or for some reason be
+		 * unable to get a random doc. Thanks okay and we retry.
```
sulabhM: Typo Thanks .. Thats

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,104 @@ err:	__clsm_leave(clsm);
...
+		WT_ERR(c->close(c));
+
+		/*
+		 * Search back through the tree from the top to resolve any
+		 * updates or tombstones. If we find a tombstone, then pick a
+		 * new random doc and try again. We retry forever as eventually
+		 * we should see a current doc or the tombstones will get
+		 * merged out.
+		 */
+		ret = __clsm_lookup(clsm, &cursor->value);
```
sulabhM: In case __clsm_lookup fails for reasons other than WT_NOTFOUND, we return that error to the user. Even though __clsm_next_random() failed at that point, we end up modifying cursor->key. Ideally we shouldn't modify the passed arguments if our call failed.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,104 @@ err:	__clsm_leave(clsm);
...
+static int
+__clsm_random_chunk(WT_SESSION_IMPL *session,
+    WT_CURSOR_LSM *clsm, WT_LSM_CHUNK **random_chunk)
+{
+	WT_DECL_RET;
+	WT_LSM_TREE *lsm_tree;
+	uint64_t checked_docs, i, rand_doc, total_docs;
+
+	lsm_tree = clsm->lsm_tree;
+	rand_doc = __wt_random(&session->rnd);
```
sulabhM: Since __wt_random returns a uint32_t, doesn't that mean that in a collection with huge number of docs, we skew towards finding a random doc within first set of chunks that combine to have UINT32_MAX of docs. In other words, if docs > UINT32_MAX, won't we always pick one from the beginning set of chunk that meet total docs >= UINT32_MAX criterion.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,104 @@ err:	__clsm_leave(clsm);
...
+
+	for (checked_docs = i = 0; i < clsm->nchunks; i++) {
+		checked_docs += lsm_tree->chunk[i]->count;
+		if (rand_doc <= checked_docs) {
+			*random_chunk = lsm_tree->chunk[i];
+			break;
+		}
+	}
+	__wt_lsm_tree_readunlock(session, lsm_tree);
+	return (ret);
```
sulabhM: Didn't set a value for ret ? May be this function can be a void return

test/csuite/wt2403_lsm_workload/main.c
```diff
@@ -0,0 +1,256 @@
...
+	testutil_check(session->open_cursor(
+	    session, name, NULL, "overwrite", &wcursor));
+	for (i = 0; i < NUM_DOCS-1; i++) {
+		wcursor->set_key(wcursor, i);
+		testutil_check(wcursor->remove(wcursor));
+	}
+	testutil_check(wcursor->close(wcursor));
+	printf("%d documents deleted\n", NUM_DOCS-1);
+
+	/* Random reads, which should see not see the deletes */
```
sulabhM: typo: see not see

src/lsm/lsm_cursor.c
daveh86: In the case of error, we unset the WT_CURSTD_KEY_INT flag. This means that any attempt to read this key will return an error as it is invalid.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,107 @@ err:	__clsm_leave(clsm);
...
+		/*
+		 * Search back through the tree from the top to resolve any
+		 * updates or tombstones. If we find a tombstone, then pick a
+		 * new random doc and try again. We retry forever as eventually
+		 * we should see a current doc or the tombstones will get
+		 * merged out.
+		 */
+		ret = __clsm_lookup(clsm, &cursor->value);
+		if (ret != WT_NOTFOUND)
+			WT_ERR(ret);
```
sulabhM: since we do not update ret after line 982, and we call WT_ERR on it at 982, is this line needed?

sulabhM:
LGTM

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,107 @@ err:	__clsm_leave(clsm);
...
+	const char *cfg[3];
+
+	cfg[0] = WT_CONFIG_BASE(session, WT_SESSION_open_cursor);
+	cfg[1] = "next_random=true";
+	cfg[2] = NULL;
+
+	random_chunk = NULL;
+	lsm_tree = clsm->lsm_tree;
+	rand_doc = __wt_random(&session->rnd);
+	__wt_lsm_tree_readlock(session, lsm_tree);
```
michaelcahill: What I expected is that this operation would use information stashed in `WT_LSM_CURSOR` rather than having to lock the tree for every read.

That said, we don't expect this to be a very common operation, and the readlock will only conflict with structural changes to the tree such as switching to a new in-memory chunk or completing a merge.

Do we know how many random lookups we can do per second on a non-trivial tree (e.g., with 10 chunks)?

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,107 @@ err:	__clsm_leave(clsm);
...
+	rand_doc %= total_docs;
+
+	for (checked_docs = i = 0; i < clsm->nchunks; i++) {
+		checked_docs += lsm_tree->chunk[i]->count;
+		if (rand_doc <= checked_docs) {
+			random_chunk = lsm_tree->chunk[i];
+			break;
+		}
+	}
+	ret = __wt_open_cursor(session, random_chunk->uri, NULL, cfg, cursor);
```
michaelcahill: Opening a new cursor on every random lookup definitely seems unnecessary.  The `WT_LSM_CURSOR` should already have a cursor at `lsm_cursor->cursors[i]`.  It would be a layering violation to call `__curfile_next_random` on that cursor but we do similar things elsewhere.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +902,107 @@ err:	__clsm_leave(clsm);
...
+		 * Search back through the tree from the top to resolve any
+		 * updates or tombstones. If we find a tombstone, then pick a
+		 * new random doc and try again. We retry forever as eventually
+		 * we should see a current doc or the tombstones will get
+		 * merged out.
+		 */
+		ret = __clsm_lookup(clsm, &cursor->value);
+		if (ret != WT_NOTFOUND)
+			WT_ERR(ret);
+	} while (ret == WT_NOTFOUND);
```
michaelcahill: Can you explain what happens if `next_random` is called on an empty LSM tree?  It looks like it will get stuck in an infinite loop...

michaelcahill:
@daveh86 I have a few comments and questions.  It's close: I'm happy to talk through any remaining issues face-to-face on Friday if that works for you.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +944,93 @@ err:	__clsm_leave(clsm);
...
+	 * WT_NOTFOUND.
+	 */
+	if (clsm->nchunks == 0)
+		return (WT_NOTFOUND);
+	for (total_docs = i = 0; i < clsm->nchunks; i++) {
+		total_docs += clsm->iters[i]->count;
+	}
+	if (total_docs == 0)
+		return (WT_NOTFOUND);
+
```
sulabhM: Now that you are not taking locks, maybe you can move random number generation after checking for "total_docs == 0". Can also shorten it to:
rand_doc = __wt_random(&session->and) % total_docs;

src/cursor/cur_file.c
```diff
@@ -117,12 +117,12 @@ err:	API_END_RET(session, ret);
 }
 
 /*
- * __curfile_next_random --
+ * __wt_curfile_next_random --
  *	WT_CURSOR->next method for the btree cursor type when configured with
  * next_random.
```
michaelcahill: I'd add a comment to the effect "exported because it is called directly for LSM cursors"

src/include/lsm.h
```diff
@@ -31,6 +31,17 @@ struct __wt_lsm_worker_args {
 };
 
 /*
+ * WT_LSM_CURSOR_ITER --
+ *	Iterator struct containing all the LSM cursor access points for a chunk.
+ */
+struct __wt_lsm_cursor_iter {
```
michaelcahill: can we please do a rename `s/lsm_cursor_iter/lsm_cursor_chunk/g`?

src/include/lsm.h
```diff
@@ -43,17 +54,12 @@ struct __wt_cursor_lsm {
...
-
-	WT_CURSOR *current;     	/* The current cursor for iteration */
+	WT_CURSOR *current;		/* The current cursor for iteration */
 	WT_LSM_CHUNK *primary_chunk;	/* The current primary chunk */
 
-	uint64_t *switch_txn;		/* Switch txn for each chunk */
-	size_t txnid_alloc;
+	WT_LSM_CURSOR_ITER **iters;	/* Array of LSM cursor units */
+	size_t iters_alloc;		/* Current size iterators array */
+	size_t iters_count;		/* Current number of iterators */
```
michaelcahill: Likewise, can we please rename these `s/iters/chunks/g`?

src/lsm/lsm_cursor.c
```diff
@@ -401,6 +401,44 @@ __clsm_close_cursors(WT_CURSOR_LSM *clsm, u_int start, u_int end)
 }
 
 /*
+ * __clsm_alloc_iter --
+ *	Allocates an array of unit objects for each chunk.
+ */
+static int
+__clsm_alloc_iter(WT_SESSION_IMPL *session, WT_CURSOR_LSM *clsm, u_int nchunks)
```
michaelcahill: Suggest `__clsm_resize_chunks` as the name for this.

src/lsm/lsm_cursor.c
```diff
@@ -401,6 +401,44 @@ __clsm_close_cursors(WT_CURSOR_LSM *clsm, u_int start, u_int end)
...
+	}
+	return (ret);
+}
+
+/*
+ * __clsm_free_iter --
+ *	Allocates an array of unit objects for each chunk.
+ */
+static void
+__clsm_free_iter(WT_SESSION_IMPL *session, WT_CURSOR_LSM *clsm)
```
michaelcahill: `__clsm_free_chunks`

src/lsm/lsm_cursor.c
```diff
@@ -493,9 +532,7 @@ retry:	if (F_ISSET(clsm, WT_CLSM_MERGE)) {
 		 * to detect snapshot isolation conflicts.
 		 */
 		if (F_ISSET(clsm, WT_CLSM_OPEN_SNAPSHOT))
-			WT_ERR(__wt_realloc_def(session,
-			    &clsm->txnid_alloc, nchunks,
-			    &clsm->switch_txn));
+			WT_ERR(__clsm_alloc_iter(session, clsm, nchunks));
```
michaelcahill: Did you consider making this unconditional, so we always allocate the full chunk array here?

We could move the other call at line 594 up into the `WT_CLSM_MERGE` case rather than calling it twice in the common case?

src/lsm/lsm_cursor.c
```diff
@@ -634,25 +667,31 @@ retry:	if (F_ISSET(clsm, WT_CLSM_MERGE)) {
...
 
 		/* Child cursors always use overwrite and raw mode. */
-		F_SET(*cp, WT_CURSTD_OVERWRITE | WT_CURSTD_RAW);
+		F_SET(clsm->iters[i]->cursor,
+		    WT_CURSTD_OVERWRITE | WT_CURSTD_RAW);
+	}
+	/* Setup the count values for each chunk in the iters*/
+	for (i = 0; i != clsm->nchunks; i++) {
+		clsm->iters[i]->count = lsm_tree->chunk[i]->count;
 	}
```
michaelcahill: Our style is to leave out braces for single-line loops.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +944,92 @@ err:	__clsm_leave(clsm);
...
+	CURSOR_API_CALL(cursor, session, next, NULL);
+	WT_CURSOR_NOVALUE(cursor);
+	WT_ERR(__clsm_enter(clsm, false, false));
+
+	do {
+		WT_ERR(__clsm_random_chunk(session, clsm, &c));
+		ret = __wt_curfile_next_random(c);
+		/*
+		 * Sometimes we may run on an empty chunk or for some reason be
+		 * unable to get a random doc. That's okay and we retry.
```
michaelcahill: How does this comment relate to the code?  Does `__wt_curfile_next_random` return `WT_NOTFOUND` in the cases described here?  If so, it would be good to update the comment to make that explicit.

src/lsm/lsm_cursor.c
```diff
@@ -902,6 +944,92 @@ err:	__clsm_leave(clsm);
...
+			WT_ERR(c->get_key(c, &cursor->key));
+			/*
+			 * Search near the current key to resolve any tombstones
+			 * and position to a valid document. If we see a
+			 * WT_NOTFOUND here that is valid, as the tree has no
+			 * documents visible to us.
+			 */
+			WT_ERR(__clsm_search_near(cursor, &exact));
+		}
+	} while (ret == WT_NOTFOUND);
```
michaelcahill: It might be better to restructure this loop as a `for (;;)` with explicit `continue` / `break`.  The fact that `WT_NOTFOUND` sometimes continues and sometimes causes a `goto err` is unnecessarily subtle.

daveh86:
@agorrod. Can you have one final look at this?

michaelcahill:
Thanks @daveh86, LGTM -- I'll merge.

PR #2888, opened 2016-07-19
WT-2778 Python test suite: make scenario initialization consistent

ddanderson:
Added make_scenarios function as a standard way to create scenarios that plays well with the new run.py -s option. Modified all tests that use scenarios to use make_scenarios. A couple other minor changes in scenario naming to ensure that no final scenario ends up with a number as a name.


ddanderson:
@daveh86, can you look this over please?  It regularizes the way we create scenarios, as a followup to WT-2767.

daveh86:
@ddanderson, this LGTM. @agorrod would you like to have a look?

agorrod:
Thanks @ddanderson lgtm - I've pushed a fix for a merge conflict. I'll merge this branch once it's passed testing.

PR #2892, opened 2016-07-21
WT-2773 Fix for reusing page images that are not in cache.

michaelcahill:
We can have a saved record of a disk image in page->modify, but with
multi->disk_image == NULL.  In that case, we don't want to copy the
new data over the top: for one thing, it may be compressed.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -3298,7 +3298,7 @@ supd_check_complete:
 				multi->disk_image = NULL;
 
 				WT_STAT_FAST_DATA_INCR(session, rec_page_match);
-				goto copy_image;
+				goto done;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @michaelcahill we talked this through, and I'm happy with the change.

@keithbostic It's definitely worth you taking a look at this change as well..

keithbostic:
@michaelcahill, @agorrod, this reverts a change I made in 890e89f, and I'm pretty confused about it -- can we please talk through the configuration/failure you saw?

agorrod:
Hi Keith,

The test failure we were using when coming up with this change was from
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress/21206/changes
I found that the failure reproduced within 3 iterations for me.

PR #2893, opened 2016-07-21
WT-2737 Fix a set of places where we ignored the return value of a function.

keithbostic:
@agorrod, @sueloverso, for review/merge.


sueloverso:
Thanks. LGTM

PR #2894, opened 2016-07-21
WT-2670 Inefficient I/O when read full DB (poor readahead)

keithbostic:
@agorrod, @michaelcahill, possible solution to WT-2670 for your consideration/review.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -109,6 +109,7 @@ Encryptor
 Encryptors
 Enqueue
 Eron
+FADV
 FALLOC
 FALLTHROUGH
 FH
--- a/src/include/os.h
+++ b/src/include/os.h
@@ -122,6 +122,9 @@ struct __wt_file_handle_posix {
 	 */
 	int	 fd;				/* POSIX file handle */
 
+	bool	 fadv_random_check;		/* POSIX_FADV_RANDOM */
+	wt_off_t fadv_last_read;		/* Read offset tracking */
+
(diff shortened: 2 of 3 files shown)


agorrod:
@keithbostic I can see that your change would work if an application is doing a single scan at a time. I think it would fail if a user is doing different types of operations along with scans or even multiple scans in parallel. Having said that, it's definitely appealing to figure out the best choice dynamically if we can.

keithbostic:
@agorrod, since we're using a single, shared file handle, I think we're only "sequential" if doing a single scan at a time. I think once there are interspersed reads (for example, multiple scans in parallel), read-ahead is likely to be less useful.

Anyway: I think this change is simple and localized, fixes the specific problem we're seeing (which is, I believe, a single thread scanning a file on startup), and doesn't require an API change. Can we accept this as "better", and design a more general solution if we're pushed in the future?

michaelcahill:
What is the performance test we are using to measure this change?  Are the results for this branch recorded somewhere?

I suspect that modern Linux will do a better job than we can at predictive readahead if we just never call `posix_fadvise` in the first place, but we need to agree on a workload before that's a sensible conversation.

keithbostic:
This branch was only a suggestion to eliminate the current cost of startup in WT-2670, by delaying the setting of `POSIX_FADV_RANDOM` until after any initial sequential reads are completed. I hoped a simple change might resolve the existing problem without much additional work.

Anyway, it sounds like you and @agorrod have a different approach in mind, I'll discard the branch.

PR #2895, opened 2016-07-21
WT-2239 Make sure LSM cursors read up to date dsk_gen, it was racing with compact.

sueloverso:
@michaelcahill and @agorrod Please review this change to fix the assertion failure as we discussed last night.  I have see it detect the assertion-triggering situation a few times and continue on with the fix on bengal.

### DIFF
--- a/src/lsm/lsm_tree.c
+++ b/src/lsm/lsm_tree.c
@@ -771,6 +771,7 @@ __wt_lsm_tree_switch(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree)
 	++lsm_tree->dsk_gen;
 
 	lsm_tree->modified = true;
+	WT_FULL_BARRIER();
 
 	/*
 	 * Set the switch transaction in the previous chunk unless this is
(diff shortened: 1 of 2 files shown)


src/lsm/lsm_tree.c
```diff
@@ -771,6 +771,7 @@ __wt_lsm_tree_switch(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree)
 	++lsm_tree->dsk_gen;
 
 	lsm_tree->modified = true;
+	WT_FULL_BARRIER();
```
agorrod: @sueloverso I'd expected the barrier to be before updating the modified state - is there a reason why it should be after?

sueloverso: The barrier needs to be between the `dsk_gen` and `switch_txn` settings.  It seemed okay to put it after the `modified` line, since we need one somewhere, to make it as late so that all memory stores are visible before setting `switch_txn`.   I don't feel strongly though.

agorrod:
Thanks @sueloverso lgtm - thanks for answering my question. I'll push a little comment and then merge.

PR #2896, opened 2016-07-21
WT-2737 Scrub dirty pages rather than evicting them

keithbostic:


michaelcahill:

PR #2897, opened 2016-07-22
WT-2781 Add checkpoint_wait configuration for bulk cursor open API.

sulabhM:


sulabhM:
@agorrod : Please take a look

sueloverso:
@sulabhM and @agorrod I took a look at this change.  These changes LGTM but opened up different, related questions.

I wondered why `WT_WITH_CHECKPOINT_LOCK_NOWAIT` did not already exist since this functionality is in `session::drop` already.  That opened up a few questions not directly related to this change:
* Should the cursor one also be undocumented?
* The description for `drop: checkpoint_wait` is not accurate.  It does not even attempt to take the checkpoint lock if this is not set.  So it is not returning an error if the lock is not available.  Or really, it appears that setting this to false allows spurious EBUSY from other locks to get returned.  Should the description be changed?

src/include/schema.h
```diff
@@ -121,6 +121,23 @@ struct __wt_table {
 	    &S2C(session)->checkpoint_lock, WT_SESSION_LOCKED_CHECKPOINT, op)
 
 /*
+ * WT_WITH_CHECKPOINT_LOCK_NOWAIT --
+ *	Return error if unable to acquire the checkpoint lock,
+ *	perform an operation if lock acquired, drop the lock.
+ */
+#define	WT_WITH_CHECKPOINT_LOCK_NOWAIT(session, ret, op) do {		\
```
sueloverso: Is there a way to use the generic `WT_WITH_LOCK` that already has support for trylock by setting the `WT_SESSION_LOCK_NO_WAIT` flag instead for this function?  This macro seems far more specific than all the other locking macros here.

sulabhM: @sueloverso : I am looking at an alternative approach to use the underlying function __wt_session_get_btree_ckpt() return an EBUSY. I am investigating if it returns EBUSY while doing a checkpoint as expected. This pull request will change accordingly.

sulabhM: @sueloverso : The description for drop seems to be outdated, and I will work on updating it with this change.
I am inclined to follow drop and keep bulk_cursor's checkpoint_wait parameter undocumented.

sulabhM:
@sueloverso @agorrod : I have updated the change, please have a look.

daveh86:
@sulabhM, this LGTM. Thank you for cleaning up the old checkpoint_wait comments that @sueloverso noticed.

dist/api_data.py
```diff
@@ -820,8 +820,9 @@ def __cmp__(self, other):
 
 'WT_SESSION.drop' : Method([
     Config('checkpoint_wait', 'true', r'''
-        wait for the checkpoint lock, if \c checkpoint_wait=false, fail if
-        this lock is not available immediately''',
+        wait for the checkpoint lock, if \c checkpoint_wait=false, perform
+        the drop operation without taking a lock, failing in case a checkpoint
```
agorrod: Replace "failing in case a checkpoint is under progress" with "returning EBUSY if the drop operation conflicts with a running checkpoint". Similarly below.

sueloverso:
I like where this ended up.  Thanks.  I fixed the text per Alex's comment and merged.

PR #2899, opened 2016-07-22
WT-2737 Page scrubbing: more fixes.

michaelcahill:
* Swap endianness of the page header back after writing (for zSeries).
* Deal with restoring updates when the WT_EVICT_SCRUB flag is not set:
  we want to save disk images for only those pages that have updates to
  restore.  Unfortunately, that isn't easy to figure out down in the
  guts of raw compression.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -385,7 +385,7 @@ __wt_reconcile(WT_SESSION_IMPL *session,
 	/*
 	 * Evicting in-memory uses the update/restore mechanisms.
 	 * The update/restore mechanisms use disk images.
-	 */ 
+	 */
 	if (LF_ISSET(WT_EVICT_IN_MEMORY))
 		LF_SET(WT_EVICT_UPDATE_RESTORE);
 	if (LF_ISSET(WT_EVICT_UPDATE_RESTORE))
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -3250,7 +3250,7 @@ supd_check_complete:
 	}
 
 	/*
-	 * If we saved the disk image in case updated needed to be restored,
+	 * If we saved the disk image in case updates needed to be restored,
 	 * but then found no updates on this page, free the disk image.
 	 */
 	if (F_ISSET(r, WT_EVICT_UPDATE_RESTORE) &&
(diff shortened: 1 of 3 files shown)


michaelcahill:
@keithbostic, can you please review these changes?  They fix some of the problems found in testing but not all: http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-ppc/5076/console

I don't feel strongly about them at all -- if you see better solutions, feel free to toss these, but if you can merge your changes and watch the tests for more failure, that would be great.  Thanks!

keithbostic:
@michaelcahill, I didn't notice in review that you had changed the eviction code to set `WT_EVICT_UPDATE_RESTORE` without setting `WT_EVICT_SCRUB` (9647806). I've never tried to separate creating disk images from doing update/restore. In the interests of stability, I'm going to back out that change and revert to setting `WT_EVICT_SCRUB` in the `WT_EVICT_UPDATE_RESTORE` case.

This probably explains why you and I never saw eye-to-eye on the naming. You think of "evict scrub" as an eviction mode, and I think of "disk images" as an action in the service of some set of eviction modes.  I've tried to move in your direction by setting just the eviction mode (`WT_EVICT_IN_MEMORY`, `WT_EVICT_UPDATE_RESTORE` and `WT_EVICT_SCRUB`) in `__evict_review()`, and then adding additional flags in `__wt_reconcile()`.

keithbostic:
@michaelcahill, reviewed, updated and merged, would you please review when you have a chance?

Thanks!

michaelcahill:
@keithbostic, thanks for merging these fixes: it's good that tests are passing again.

I deliberately don't want `WT_EVICT_SCRUB` to be forced on with `WT_EVICT_UPDATE_RESTORE` -- this makes the latter strictly less useful than it used to be, because we have to save all page images even when there is cache pressure and child pages with no updates to restore.

Can you keep working on this so that we can set `WT_EVICT_UPDATE_RESTORE` without `WT_EVICT_SCRUB`?

src/reconcile/rec_write.c
```diff
@@ -382,6 +382,15 @@ __wt_reconcile(WT_SESSION_IMPL *session,
...
 	mod->last_oldest_id = oldest_id;
 
+	/*
+	 * Evicting in-memory uses the update/restore mechanisms.
+	 * The update/restore mechanisms use disk images.
+	 */
+	if (LF_ISSET(WT_EVICT_IN_MEMORY))
+		LF_SET(WT_EVICT_UPDATE_RESTORE);
+	if (LF_ISSET(WT_EVICT_UPDATE_RESTORE))
+		LF_SET(WT_EVICT_SCRUB);
```
michaelcahill: I don't like changing the flags to change here -- I would like the behavior to be fully specified by the eviction layer.

PR #2903, opened 2016-07-23
WT-2737 Page scrubbing: more fixes.

keithbostic:
Move all flag setting outside of reconciliation, don't save a disk image in the update/restore path unless there are updates to be restored to that segment of the name space.


keithbostic:
@michaelcahill, for your review. It's passing tests, so I'm going to go ahead and merge.

PR #2904, opened 2016-07-25
Fix a JNI thread-safe bug which may cause core dump

jindongh:
The updated ex_thread test may crash without this fix, because JNI wrapper tried to free session->lang_private after it returned session with connection.close().
After java thread A open the session, it will allocate memory and assign it to session->lang_private, when java thread A close the session, it will call connection.close(session) first, and then try to free session->lang_private. The problem is: after thread A call connection.close() and before it free lang_private, another thread may reuse this session and close it, then thread A found lang_private is set to NULL by another thread and it crashes.
So we need to free lang_private before we call connection.close in the JNI wrapper.


wtbuild:

PR #2905, opened 2016-07-25
WT-2785 Scrub dirty pages rather than evicting them: single-page reconciliation

keithbostic:

I don't like the way this turned out (the split between WT_MULTI structures and a single replacement block is beyond baroque), but I can't see anything better -- your good idea goes here.


michaelcahill:
Thanks, @keithbostic -- this lgtm, I'll merge.  I agree that we could revisit in future to try to make the single-page case less special, but this looks like a low risk approach that mostly reuses the existing rewrite code.

PR #2906, opened 2016-07-26
WT-2791 Upload artifacts for each build. Rename current artifacts to "Binaries"

daveh86:


daveh86:
retest this please

daveh86:
@agorrod, this change lets us produce evergreen artifacts on all builds.

agorrod:
Thanks @daveh86 lgtm - should I go ahead and merge?

PR #2908, opened 2016-07-26
WT-2638 ftruncate may not be supported

keithbostic:

There's additional discussion in WT-2638.


agorrod:
@keithbostic I don't think the benefits of the change outweigh the cost of making an API change. I'm inclined to keep the API as it was before, and deal with a lock-free truncate call when a user of the file system API requests the functionality.
If I get a chance I'll back out that part of the change.

keithbostic:
@agorrod, the original question wasn't handling lock-free truncate, it was dealing with filesystem implementations that can only extend the file, they can't shorten it, and there's a known filesystem implementation with that problem. The problem appears in WiredTiger code, for example, the logging code knows both the fallocate and ftruncate APIs can be used to extend the file.

To be clear, I'm not saying we can't support that filesystem without the change: the logging code only falls back to ftruncate if fallocate fails, so things work if fallocate succeeds and WiredTiger doesn't fallback to ftruncate, but it's problematical and the change simplifies both the WiredTiger and OS-layer code.

Second, this is an unreleased API (even via MongoDB), why can't we change it?

agorrod:
Thanks @keithbostic this change lgtm now.

PR #2910, opened 2016-07-27
WT-2798 Fix data consistency bug with table creates during a checkpoint.

michaelcahill:
When logging is disabled, a create operation (and potentially other
metadata updates) could write partially completed checkpoint metadata,
leaving on-disk files inconsistent until the checkpoint completes.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -150,12 +150,9 @@ __wt_txn_get_snapshot(WT_SESSION_IMPL *session)
 	/* For pure read-only workloads, avoid scanning. */
 	if (prev_oldest_id == current_id) {
 		txn_state->snap_min = current_id;
-		__txn_sort_snapshot(session, n, current_id);
-
 		/* Check that the oldest ID has not moved in the meantime. */
 		WT_ASSERT(session, prev_oldest_id == txn_global->oldest_id);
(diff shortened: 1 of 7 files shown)


src/txn/txn.c
```diff
@@ -137,20 +138,26 @@ __wt_txn_get_snapshot(WT_SESSION_IMPL *session)
...
+	 * metadata.  We don't have to keep the checkpoint's changes pinned so
+	 * don't including it in the published snap_min.
+	 */
+	if ((id = txn_global->checkpoint_txnid) != WT_TXN_NONE)
+		txn->snapshot[n++] = id;
+
 	/* For pure read-only workloads, avoid scanning. */
 	if (prev_oldest_id == current_id) {
 		txn_state->snap_min = current_id;
-		__txn_sort_snapshot(session, 0, current_id);
```
agorrod: Is it OK for this to move outside of the read lock because it is sorting the snapshot array inside a single session handle?

agorrod:
@michaelcahill I had one question apart from that lgtm.

src/txn/txn.c
michaelcahill: @agorrod, yes, we've already got everything cached locally by this point -- note that we do the same for the slow path.

PR #2911, opened 2016-07-27
WT-2764 Release checkpoint transaction before syncing

keithbostic:
@michaelcahill, Sue & I were talking about the leaked memory and using the WT-2764 branch, so I built it. Here's some random stuff, hope it saves you 5 minutes. :)


michaelcahill:
Thanks @keithbostic, lgtm.

PR #2912, opened 2016-07-27
WT-2793 Enhance statistics related to overflow values

sueloverso:
@agorrod Here are some changes for the wtperf and overflow tests.  I changed the 130k one to be btree so that we run both with LSM and btree.  That negates the existing plot.  I removed the unused stat and put the overflow related ones from dsrc into the conn as well.

I know we're not supposed to remove existing stats, but this one was always 0 and meaningless.  Again, please let me know and I'll do the merge since it needs Jenkins changes to go along with it.


agorrod:
autobuild: retest this please

agorrod:
@sueloverso This looks good to me.

> I know we're not supposed to remove existing stats, but this one was always 0 and meaningless.

I agree with the stat changes, but they deserve a comment in the upgrade docs - would you mind adding that?

agorrod:
autobuild: retest this please

PR #2914, opened 2016-07-28
WT-2796 Fix a memory leak when using the lookaside table.

michaelcahill:
(This was introduced in WT-2737).


agorrod:

PR #2915, opened 2016-07-28
WT-2800 Fix incorrect error message.

sueloverso:


sueloverso:
self-reviewing and merging

PR #2917, opened 2016-07-29
WT-2802 Copy values during commit before releasing snapshot.

michaelcahill:
When committing a transaction with cursors positioned, copy values
before releasing the transaction snapshot.  Otherwise, there is a window
where the oldest transaction ID could move past the committing
transaction, and the values it needs to copy get freed by another
thread.

This only applies when logging is enabled, and only in applications that
commit transactions with positioned cursors.


sulabhM:
@michaelcahill : LGTM

PR #2919, opened 2016-08-01
WT-2725 Prevent eviction of read-only pages with recent updates.

michaelcahill:
Reverts some of 0136888b3b677d67ff929c6f5b1f8d6aa6793b13.


daveh86:
I've run this against the 3.0 perf regression and the original tests for WT-2725, this passed both.

LGTM

michaelcahill:
Thanks @daveh86, merging.

PR #2920, opened 2016-08-01
WT-2803 Add VERBOSE=1 to all make check jobs

daveh86:


michaelcahill:

PR #2921, opened 2016-08-01
WT-2805 Avoid infinite recursion on error stream failure.

michaelcahill:
Previously, if both a custom and the default error streams failed to
write it caused infinite recursion.


agorrod:

PR #2922, opened 2016-08-01
WT-2801 Prevent eviction of metadata updates by a running checkpoint.

michaelcahill:
Only update btree->checkpoint_gen during a full checkpoint operation
that is holding the checkpoint lock.  Otherwise, updates in the
metadata by a running checkpoint can pass the "visible_all" test and be
evicted.

Followup to WT-2798.


agorrod:
Thanks @michaelcahill lgtm - I'll let you merge this one yourself.

PR #2924, opened 2016-08-02
WT-2804 Don't read values in a tree without a snapshot.

michaelcahill:
Improve two recent assertions, one from WT-2798 relating to writing
metadata updates to disk that are part of a running transaction, and
another from WT-2802 that checks that we don't try to copy values from a
cursor without a transaction pinned.  The latter doesn't apply to
cursors on checkpoints (including chunk cursors in an LSM tree).

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1152,15 +1152,17 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 	if (!skipped &&
 	    (F_ISSET(btree, WT_BTREE_LOOKASIDE) ||
 	    __wt_txn_visible_all(session, max_txn))) {
+#ifdef HAVE_DIAGNOSTIC
 		/*
 		 * The checkpoint transaction is special.  Make sure we never
 		 * write (metadata) updates from a checkpoint in a concurrent
 		 * session.
 		 */
-		WT_ASSERT(session, *updp == NULL ||
-		    (txnid = (*updp)->txnid) == WT_TXN_NONE ||
+		txnid = *updp == NULL ? WT_TXN_NONE : (*updp)->txnid;
+		WT_ASSERT(session, txnid == WT_TXN_NONE ||
 		    txnid != S2C(session)->txn_global.checkpoint_txnid ||
 		    WT_SESSION_IS_CHECKPOINT(session));
+#endif
 		return (0);
 	}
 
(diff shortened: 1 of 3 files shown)


src/session/session_api.c
```diff
@@ -66,11 +66,15 @@ __wt_session_copy_values(WT_SESSION_IMPL *session)
...
-			   WT_TXN_NONE);
+#ifdef HAVE_DIAGNOSTIC
+			/*
+			 * We have to do this with a transaction ID pinned
+			 * unless the cursor is reading from a checkpoint.
+			 */
+			WT_TXN_STATE *txn_state = WT_SESSION_TXN_STATE(session);
+			WT_ASSERT(session, txn_state->snap_min != WT_TXN_NONE ||
+			   (WT_PREFIX_MATCH(cursor->uri, "file:") &&
+			   F_ISSET((WT_CURSOR_BTREE *)cursor, WT_CBT_NO_TXN)));
```
michaelcahill: this is the real fix

src/reconcile/rec_write.c
```diff
@@ -1152,15 +1152,17 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
 		 * write (metadata) updates from a checkpoint in a concurrent
 		 * session.
 		 */
-		WT_ASSERT(session, *updp == NULL ||
-		    (txnid = (*updp)->txnid) == WT_TXN_NONE ||
+		txnid = *updp == NULL ? WT_TXN_NONE : (*updp)->txnid;
+		WT_ASSERT(session, txnid == WT_TXN_NONE ||
 		    txnid != S2C(session)->txn_global.checkpoint_txnid ||
 		    WT_SESSION_IS_CHECKPOINT(session));
+#endif
```
michaelcahill: Coverity complained about this assertion having a side-effect because it set `txnid`.

michaelcahill:
Hold off reviewing this one -- the new assertion has fired in local testing.

michaelcahill:
The new version has a functional change to rollback during autocommit, it has run 25+ iterations of the first `test/format` config from the ticket.  Previously it hasn't made it past 10.

michaelcahill:
@agorrod, this is ready for review.  Can you please take a look?

michaelcahill:
Thanks @agorrod, it's run 100 iterations of the `test/format` workload now.  I'll merge...

PR #2925, opened 2016-08-02
WT-2770 Add statistics tracking schema operations

sulabhM:
Tracking successful and failed attempts separately for each of the following:
create
compact
drop
rebalance
rename
salvage
truncate
verify


sulabhM:
retest this please

sulabhM:
@daveh86 : Please take a look

dist/stat_data.py
```diff
@@ -336,6 +336,22 @@ def __init__(self, name, desc, flags=''):
     ##########################################
     SessionStat('session_cursor_open', 'open cursor count', 'no_clear,no_scale'),
     SessionStat('session_open', 'open session count', 'no_clear,no_scale'),
+    SessionStat('session_table_compact_fail', 'table compact failed attempts', 'no_clear,no_scale'),
```
sueloverso: It would be more consistent to say `calls` instead of `attemtps` for all of these stats.  See AsyncStats and CursorStats for precedent.

daveh86:
@sulabhM, this LGTM

sulabhM:
@sueloverso : Please take a look, I have addressed your comments.

sueloverso:
Thanks @sulabhM then LGTM

src/session/session_api.c
```diff
@@ -530,6 +534,8 @@ __session_create_readonly(
 	WT_UNUSED(uri);
 	WT_UNUSED(config);
 
+	WT_STAT_FAST_CONN_INCR((WT_SESSION_IMPL *)wt_session,
```
agorrod: @sulabhM Casting a ``WT_SESSION`` to a ``WT_SESSION_IMPL`` uses a certain protocol in the source code at the moment. It always happens via a local variable even if it's only used in one place.

While looking at this code, I realised that there is another problem with the read-only stubs, which is that they don't use the ``SESSION_API_CALL`` macros - which mean that errors will be reported incorrectly. The changes in the readonly stubs would be better if they had the form:

```
@@ -531,12 +531,18 @@ static int
 __session_create_readonly(
     WT_SESSION *wt_session, const char *uri, const char *config)
 {
+       WT_SESSION_IMPL *session;
+
        WT_UNUSED(uri);
        WT_UNUSED(config);

-       WT_STAT_FAST_CONN_INCR((WT_SESSION_IMPL *)wt_session,
-           session_table_create_fail);
-       return (__wt_session_notsup(wt_session));
+       session = (WT_SESSION_IMPL *)wt_session;
+       SESSION_API_CALL_NOCONF(session, create);
+
+
+       WT_STAT_FAST_CONN_INCR(session, session_table_create_fail);
+       ret = __wt_session_notsup(wt_session);
+err:   API_END_RET(session, ret);
 }

 /*
```

sulabhM:
@agorrod : Please take a look now.

src/session/session_api.c
```diff
@@ -527,10 +531,18 @@ static int
...
+
 	WT_UNUSED(uri);
 	WT_UNUSED(config);
 
-	return (__wt_session_notsup(wt_session));
+	session = (WT_SESSION_IMPL *)wt_session;
+	SESSION_API_CALL_NOCONF(session, create);
+
+	WT_STAT_FAST_CONN_INCR(session, session_table_create_fail);
+	ret = __wt_session_notsup(wt_session);
```
agorrod: All callers of ``__wt_session_notsup`` now have a ``WT_SESSION_IMPL`` - it would be better to change the parameter type to the function rather than pass in a ``WT_SESSION``

sulabhM:
@agorrod : Done, please take a look

agorrod:
Thanks @sulabhM lgtm, I'll merge.

PR #2926, opened 2016-08-03
WT-2764 Fix a bug calculating the dirty percentage of cache.

michaelcahill:
Also fix the statistic measuring the total time for checkpoint
scrubbing.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -313,7 +313,8 @@ __checkpoint_reduce_dirty_cache(WT_SESSION_IMPL *session)
 	struct timespec start, last, stop;
 	u_int current_dirty;
 	uint64_t bytes_written_start, bytes_written_total;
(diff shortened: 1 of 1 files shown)


agorrod:
Lgtm. I believe it is possible for the cache size to be zero sometimes, but not when the eviction server is doing this calculation.

michaelcahill:
autobuild: retest this please

michaelcahill:
autobuild: please retest this

michaelcahill:
autobuild: retest this please

michaelcahill:
Thanks, @agorrod, I've refined checkpoint scrubbing some more to avoid slowness in cases where the cache is mostly clean but still wait for long enough to start making progress.  I've also rolled back the change that bumped the checkpoint generation of the metadata tree.

I'll go ahead and merge so this gets stress testing overnight.

sueloverso:
@michaelcahill I'll keep an eye on things today and dig into them as tests run.

PR #2927, opened 2016-08-03
WT-2764 Compilation fixes size_t -> uint64_t on OSX.

sueloverso:


sueloverso:
self-review to get evergreen testing.

PR #2929, opened 2016-08-04
WT-2801 Exclude eviction from the metadata until checkpoint is stable

agorrod:
We were releasing the checkpoint transaction too early, and ignoring
it in the metadata file.


agorrod:
Thanks @michaelcahill. I'm happy with these changes now - I'll merge.

PR #2930, opened 2016-08-04
WT-2811 Include the checkpoint transaction if there is no btree.

michaelcahill:
Also make sure there is always a read barrier between reading the
oldest ID and the pinned checkpoint ID.


sueloverso:
@michaelcahill I'm going to merge this.  This branch has passed several iterations of test/format successfully.

PR #2934, opened 2016-08-05
WT-2811 The checkpoint session should not ignore it's own transaction ID

agorrod:
There is a sanity check in reconciliation that transaction IDs always
move forward. The tracked ID is updated by both the checkpoint session
and reconciliation in service of eviction. Recent changes in when
a btree is open for eviction mean that interleaving reconciliation
between eviction and checkpoint made it look like time went backwards.


michaelcahill:

PR #2935, opened 2016-08-05
WT-2813 Configure eviction dirty settings explicitly for LSM.

sueloverso:
@agorrod please review.  Your comment in the ticket was a bit confusing, but I changed the settings if the chosen cache size is less than 20Mb as well as if it is using LSM.  This allows the two CONFIGs in the ticket to repeatedly complete now.


agorrod:
Thanks @sueloverso lgtm - I'll merge

PR #2938, opened 2016-08-08
WT-2820 add gcc warn_unused_result attribute

keithbostic:
@michaelcahill, here are the changes we discussed to turn on gcc error warnings when a function call return is ignored.

I additionally added the gcc `cold` attribute to the standard error, verbose and message functions, see 08cfc8f. It's advertised to improve branch prediction, shrug.

The only big change was to the Btree page debug code (it had to be extensively reworked), but that shouldn't matter to anyone outside the WiredTiger team.

I don't think there are any bug fixes in these changes (other than a few places where we're testing function return values and weren't previously).


keithbostic:
@ddanderson, can you please review 1757fb5? I replaced the explicit calls to `fprintf(stderr, ...)` with calls to `util_err()`, and used class names instead of local variables names in the errors, but I wanted to make sure I didn't mess up something you were doing intentionally.  Thanks!

src/log/log_slot.c
```diff
@@ -262,7 +262,7 @@ __wt_log_slot_new(WT_SESSION_IMPL *session)
 		/*
 		 * If we didn't find any free slots signal the worker thread.
 		 */
-		(void)__wt_cond_auto_signal(session, conn->log_wrlsn_cond);
+		WT_RET(__wt_cond_auto_signal(session, conn->log_wrlsn_cond));
```
agorrod: This looks dangerous to me. If {{__wt_log_slot_new}} returns an error {{__log_slot_switch_internal}} will return with {{WT_MYSLOT_CLOSE}} set which feels broken. @sueloverso do you think this is an OK change?

keithbostic: I modeled the return after the failure line above it, a call to `__wt_log_acquire` (which can fail for a wide variety of reasons, including malloc failures). If this is dangerous, I suspect that return is problematic as well.

That said, my plan is to change `__wt_cond_auto_signal` to not return anything as part of [WT-2822](https://jira.mongodb.org/browse/WT-2822), so this specific return shouldn't be around too much longer.

ddanderson:
@keithbostic, that change LGTM.

src/log/log_slot.c
sueloverso: The assumption is that if `wt_log_slot_new` returns an error then `log->active_slot` is not yet ready.  The `WT_MYSLOT_CLOSE` flag is part of the error handling.  So both should be okay because they're returning an error indicating a new active slot is not ready.

michaelcahill:
Thanks @keithbostic -- lgtm, I'll merge.

PR #2939, opened 2016-08-08
WT-2820 add gcc warn_unused_result attribute

keithbostic:
@michaelcahill, @agorrod, these changes fail our code if it ignores an error return from a WiredTiger method.

I fixed the bugs it found and pushed those fixes into #2938, because I'm pretty sure we'll merge that branch, and, well, these changes are pretty ugly. At the least, we probably shouldn't implement these checks based on --enable-diagnostic.

Anyway, let's talk it over when you have a few minutes.


michaelcahill:
@keithbostic, I'm inclined to discard this branch.  As you say, the changes are pretty ugly.  Add to that we'd need Windows support plus MongoDB SConscript / SConstruct changes to substitute this new macro in the public API and it doesn't seem worth it.

Let me know if you disagree -- we could come up with an approach that doesn't substitute something new into `wiredtiger.h`, or use a wrapper function with the attribute for calls that reenter the API, if you think we should keep the checks in the build.

keithbostic:
> I'm inclined to discard this branch. 

Agreed, I'll close it.

This is the kind of thing we should probably check on every so often out of caution, but the exercise didn't find anything non-trivial this time, I'm not worried about checking regularly.

PR #2940, opened 2016-08-09
WT-2823 support file handles without a truncate method

keithbostic:
Changes to support file systems without file truncate support.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -137,18 +137,9 @@ __wt_block_checkpoint_load(WT_SESSION_IMPL *session, WT_BLOCK *block,
 	 * that was done when the checkpoint was first written (re-writing the
 	 * checkpoint might possibly make it relevant here, but it's unlikely
 	 * enough I don't bother).
-	 *
-	 * If in-memory, we don't read or write the object, and the truncate
-	 * will unnecessarily allocate buffer space.
 	 */
-	if (!checkpoint && !F_ISSET(S2C(session), WT_CONN_IN_MEMORY)) {
-		WT_ERR(__wt_verbose(session, WT_VERB_CHECKPOINT,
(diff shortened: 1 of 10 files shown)


keithbostic:
@michaelcahill, here are the changes to run without ftruncate.

FTR, I'm just doing this to test:
```
diff --git a/src/os_posix/os_fs.c b/src/os_posix/os_fs.c
index 06ec521..d3e01bb 100644
--- a/src/os_posix/os_fs.c
+++ b/src/os_posix/os_fs.c
@@ -723,6 +723,7 @@ directory_open:
        file_handle->fh_sync_nowait = __posix_file_sync_nowait;
 #endif
        file_handle->fh_truncate = __posix_file_truncate;
+       file_handle->fh_truncate = NULL;
        file_handle->fh_write = __posix_file_write;
 
        *file_handlep = file_handle;
```

The interesting changes are:
* Change the block manager to completely ignore whether or not truncate actually succeeds and completely depend on the in-memory understanding of the file's size; I suspect this is the change most likely to break something.
* The logging changes were pretty simple (although there's a new WT_LOG flag to avoid doing log file truncation setup repeatedly if there's no ftruncate call); @sueloverso, would you please do at least that review?
* I removed the in-memory file system's support for ftruncate -- now that ftruncate is optional, that made sense to me. I left that ftruncate code in the example program `examples/c/ex_file_system.c`; @agorrod, I hope that's OK with you.

michaelcahill:

If @sueloverso notices a problem with the logging change, we should be able to fix it later.

sueloverso:
I somehow missed being tagged in this so I am late to this party.  The potential ramifications of this seem to be that if the truncate call simply returns 0 and does nothing, we could possibly end up with a log that cannot be traversed with a log cursor or printlog.  For instance if there is a partial log record at the end of the log, it would not be removed during recovery.  Recovery itself detects and deals with that, so recovery and data are safe.  But the benign operation of walking the log could be affected.

PR #2941, opened 2016-08-10
WT-2817 Add wtperf conf to stress test checkpoints with updates

sulabhM:


sulabhM:
@daveh86 : Can you please review this change

daveh86:
This LGTM

agorrod:
@sulabhM The naming convention for wtperf configurations is to use a ``-`` instead of an ``_``.

bench/wtperf/runners/checkpoint_stress.wtperf
```diff
@@ -0,0 +1,19 @@
...
+# Enough data to fill the cache. 100 million 1k records results in two ~6GB
+# tables
+icount=100000000
+create=true
+compression="snappy"
+populate_threads=1
+checkpoint_interval=60
+checkpoint_threads=1
+report_interval=5
+run_time=600
```
agorrod: Most configurations that are in the code base use a shorter runtime than this - I'm inclined to add a comment explaining that the longer runtime is important to ensure checkpoints are completing.

agorrod:
autobuild: retest this please

PR #2943, opened 2016-08-10
WT-2323 Allocate a transaction id at the beginning of join iteration

ddanderson:


ddanderson:
@agorrod, I need some help on this. I think this is the sort of change you were talking about when the issue was raised. However, I haven't been able to construct a test that shows that this change is an improvement. Can you think of a case where it may make a difference? Thanks!

michaelcahill:
@ddanderson there is a comment from @agorrod in the JIRA ticket, can you please follow up there?

ddanderson:
@agorrod, this is ready for review.  I'll put some discussion in the JIRA ticket.

agorrod:
Thanks @ddanderson lgtm. I agree that the test case is interesting, and it's better to be safe re: having a snapshot allocated at the start of a join.

PR #2944, opened 2016-08-10
WT-2826 clang38 false positive on uninitialized variable.

keithbostic:
Ready for review/merge.


ddanderson:
@keithbostic, this change lgtm.

PR #2946, opened 2016-08-10
WT-2827 Set a reasonable minimum for log_size.

sueloverso:
@agorrod Please review this change.  If a user sets a small `log_size` it will use the log file size.  We may or may not want this semi-silent change to the user's config.

### DIFF
--- a/src/conn/conn_ckpt.c
+++ b/src/conn/conn_ckpt.c
@@ -38,6 +38,13 @@ __ckpt_server_config(WT_SESSION_IMPL *session, const char **cfg, bool *startp)
 	if (conn->ckpt_usecs != 0 ||
 	    (conn->ckpt_logsize != 0 &&
 	    FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))) {
+		/*
+		 * If non-zero, use a minimum of the log file size.  The logging
+		 * subsystem has already been initialized.
+		 */
+		if (conn->ckpt_logsize != 0)
+			conn->ckpt_logsize = WT_MAX(
+			    conn->ckpt_logsize, conn->log_file_max);
 		/* Checkpoints are incompatible with in-memory configuration */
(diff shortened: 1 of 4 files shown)


src/conn/conn_ckpt.c
```diff
@@ -38,6 +38,13 @@ __ckpt_server_config(WT_SESSION_IMPL *session, const char **cfg, bool *startp)
 	if (conn->ckpt_usecs != 0 ||
 	    (conn->ckpt_logsize != 0 &&
 	    FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))) {
+		/*
+		 * If non-zero, use a minimum of the log file size.  The logging
+		 * subsystem has already been initialized.
+		 */
+		if (conn->ckpt_logsize != 0)
+			conn->ckpt_logsize = WT_MAX(
+			    conn->ckpt_logsize, conn->log_file_max);
```
agorrod: I've got two questions about this change:
* Can we guarantee that log_file_max is always configured before getting here.
* Does this work as expected when reconfiguring?

I'm more inclined to put a hard lower bound on the volume of log records, rather than tie it to another configuration option. That makes documentation simpler, and leads to less configuration corner cases.

sueloverso: I bolstered the conditional to check the flag if logging is enabled.  If it is, we are sure that log_file_max is set.  I added some test cases to one of the reconfig unit tests.  I checked that it did the right thing via a printf.

sueloverso:
@agorrod I addressed your two questions with a new change and test case.  If you feel strongly about bounding the volume we should talk.

agorrod:
@sueloverso We talked about this - to close the loop. I'm happy for you to merge this change if you haven't thought of any reasons why a user might want to trigger checkpoints more regularly than per log file.

michaelcahill:
It sounds like we don't have strong opinions about a better fix here.  I'll merge this.

PR #2949, opened 2016-08-12
SERVER-23622 Eviction experiments.

michaelcahill:
*NOT FOR COMMIT* in their current form.


daveh86:
I found the LSM Perf job hung running this branch. Trying to repro now.

agorrod:
@michaelcahill I've done a first pass through this branch. The changes make sense to me, but I'll need to do a more careful review before merging.

I also noticed that there are a couple of code blocks wrapped in ``#if`` that deserve some more work before the branch is merged.

michaelcahill:
@sueloverso, can you please review this branch as well?

In the unlikely event that you are happy with everything, please go ahead and merge as well.

bench/wtperf/wtperf.c
```diff
@@ -1540,8 +1540,7 @@ execute_populate(CONFIG *cfg)
 		print_ops_sec = 0;
 	} else {
 		print_secs = (double)msecs / (double)MSEC_PER_SEC;
-		print_ops_sec =
-		    (uint64_t)((cfg->icount / msecs) / MSEC_PER_SEC);
+		print_ops_sec = (uint64_t)(cfg->icount / print_secs);
```
michaelcahill: Note that this is unrelated but has been bugging me -- at the end of a populate, `wtperf` will now print a meaningful "ops/sec" result.

test/csuite/wt2834_join_bloom_fix/main.c
```diff
@@ -161,7 +161,7 @@ main(int argc, char *argv[])
 	return (0);
 }
 
-void *populate(TEST_OPTS *opts)
+void populate(TEST_OPTS *opts)
```
michaelcahill: These changes are definitely arguable.  The reason I touched this file is that I saw `WT_ROLLBACK` returned from the `WT_CURSOR::insert` call, which the test code doesn't expect.  Using an auto-commit `insert` avoids the issue (WT loops internally on rollback).

I don't know for sure if it relies on the `search` and `insert` operations have to happen in a transaction for correctness of the test, and I don't think the `WT_ROLLBACK` will happen with the current state of the branch.

src/include/cache.h
```diff
@@ -153,15 +158,17 @@ struct __wt_cache {
...
+#define	WT_EVICT_STATE_DIRTY_HARD	0x10 /* Dirty % blocking app threads */
+#define	WT_EVICT_STATE_SCRUB		0x20 /* Scrub dirty pages pages */
+#define	WT_EVICT_STATE_URGENT		0x40 /* Pages are in the urgent queue */
 #define	WT_EVICT_STATE_ALL	(WT_EVICT_STATE_CLEAN | WT_EVICT_STATE_DIRTY)
 	uint32_t state;
 	/*
 	 * Pass interrupt counter.
 	 */
-	uint32_t pass_intr;		/* Interrupt eviction pass. */
+	volatile uint32_t pass_intr;	/* Interrupt eviction pass. */
```
michaelcahill: We have loops that check this without any explicit barriers.  I'm pretty sure we don't want the compiler caching the value in a register.

src/evict/evict_lru.c
```diff
@@ -488,24 +515,6 @@ __wt_evict_destroy(WT_SESSION_IMPL *session)
 }
 
 /*
- * __evict_helper --
- *	Thread to help evict pages from the cache.
- */
-static int
-__evict_helper(WT_SESSION_IMPL *session)
```
michaelcahill: I folded this into `__evict_lru_pages`.

src/evict/evict_lru.c
```diff
@@ -526,16 +535,7 @@ __evict_update_work(WT_SESSION_IMPL *session)
...
 		return (false);
 
-	/*
-	 * Setup the number of refs to consider in each handle, depending
-	 * on how many handles are open. We want to consider less candidates
-	 * from each file as more files are open. Handle the case where there
-	 * are no files open by adding 1.
-	 */
-	cache->evict_max_refs_per_file =
-	    WT_MAX(100, WT_MILLION / (conn->open_file_count + 1));
```
michaelcahill: This variable was no longer used.  The logic has been replaced by the calculation of `target_pages` in `__evict_walk_file`.

src/evict/evict_lru.c
```diff
@@ -682,8 +700,7 @@ __evict_pass(WT_SESSION_IMPL *session)
 			 */
 			WT_STAT_FAST_CONN_INCR(session,
 			    cache_eviction_server_slept);
-			__wt_cond_wait(session,
-			    cache->evict_cond, WT_THOUSAND * WT_MAX(loop, 1));
+			__wt_cond_wait(session, cache->evict_cond, WT_THOUSAND);
```
michaelcahill: This turned out to be important.  If something was blocking eviction for a while (e.g., checkpoints), the loop counter could be large, and it would take a long time to recover.

src/evict/evict_lru.c
```diff
@@ -879,32 +895,25 @@ __evict_lru_pages(WT_SESSION_IMPL *session, bool is_server)
...
-		total_evict = cache->app_evicts +
-		    cache->server_evicts + cache->worker_evicts;
-		app_evict_percent = (100 * cache->app_evicts) /
-			(total_evict + 1);
-		if (app_evict_percent < APP_EVICT_THRESHOLD) {
-			WT_STAT_FAST_CONN_INCR(session,
-			    cache_eviction_server_not_evicting);
-			return (0);
-		}
-	}
```
michaelcahill: This logic has been replaced by `evict_empty_score`, which measures how often the queue has been empty recently (on a scale of 0-100).  If the score is low, the server is queuing pages fast enough and participates in eviction.  Otherwise, it tries harder to keep the queues full.

src/evict/evict_lru.c
```diff
@@ -917,21 +926,45 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
...
+	if ((ret = __evict_walk(cache->walk_session, queue)) != 0)
 		return (ret == EBUSY ? 0 : ret);
 
+	/* Make sure the other queue is current before locking. */
+	if (cache->evict_current_queue != other_queue) {
+		__wt_spin_lock(session, &cache->evict_queue_lock);
+		cache->evict_other_queue = queue;
+		cache->evict_current_queue = other_queue;
+		__wt_spin_unlock(session, &cache->evict_queue_lock);
+	}
```
michaelcahill: I don't love that the eviction server has to wait on `evict_queue_lock` here.  Another thought I had was that it could just set `queue->evict_current = NULL` before acquiring `queue->evict_lock`.

Either way, an unlucky worker or application thread could race and end up waiting on the queue while the server holds the lock...

src/evict/evict_lru.c
```diff
@@ -1256,6 +1295,14 @@ __evict_push_candidate(WT_SESSION_IMPL *session,
 	evict->btree = S2BT(session);
 	evict->ref = ref;
 	evict->score = __evict_read_gen(evict);
+
+	/* Adjust for size when doing dirty eviction. */
+	if (FLD_ISSET(S2C(session)->cache->state, WT_EVICT_STATE_DIRTY) &&
+	    evict->score != WT_READGEN_OLDEST && evict->score != UINT64_MAX &&
+	    !__wt_page_is_modified(ref->page))
+		evict->score += WT_MEGABYTE -
+		    WT_MIN(WT_MEGABYTE, ref->page->memory_footprint);
```
michaelcahill: I don't know how useful this really is: the intent is to make larger dirty pages more attractive eviction candidates when we're trying to reduce the dirty %.  Since this branch also fixes some bugs with the urgent queue, this may no longer be worthwhile.

michaelcahill:
I've done some minor cleanup and added some notes explaining why I made some of the changes and which ones look questionable (to me).  Please take a look and give me any comments or questions you have about these changes.

agorrod:
Thanks @michaelcahill lgtm - I don't fully understand the join test, but I can't see how switching to an auto-commit transaction there would alter the functionality.

src/evict/evict_lru.c
```diff
@@ -555,7 +562,17 @@ __evict_pass(WT_SESSION_IMPL *session)
...
+
+		/*
+		 * If the queue has been empty recently, keep queuing more
+		 * pages to evict.  If the rate of queuing pages is high
+		 * enough, this score will go to zero, in which case the
+		 * eviction server might as well help out with eviction.
+		 */
+		if (cache->evict_empty_score < WT_EVICT_EMPTY_SCORE_CUTOFF ||
+		    (!WT_EVICT_HAS_WORKERS(session) &&
+		    !__evict_queue_empty(cache->evict_urgent_queue)))
```
sueloverso: This conditional is a bit confusing and could use more explanation in the comment.  If there are no eviction workers, and no pages on the urgent queue, the lone eviction server thread won't call `evict_lru_pages`?  Or is that situation solely dependent on the empty score getting low enough before it does so?

src/evict/evict_lru.c
```diff
@@ -810,21 +817,45 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
...
-	WT_EVICT_QUEUE *queue;
+	WT_EVICT_QUEUE *queue, *other_queue;
 	uint64_t read_gen_oldest;
-	uint32_t candidates, entries, queue_index;
+	uint32_t candidates, entries;
 
 	cache = S2C(session)->cache;
 
+	/* Age out the score of how much the queue has been empty recently. */
+	cache->evict_empty_score = (99 * cache->evict_empty_score) / 100;
```
sueloverso: Would this be clearer as:
```
if (cache->evict_empty_score > 0)
    --cache->evict_empty_score;
```
Or am I missing something subtle?

src/evict/evict_lru.c
sueloverso: All the queues here kind of make my head hurt.  In particular (as mentioned below in cache.h), the differences between `evict_fill_queue` and `evict_other_queue` are subtle and not always clear.  They're usually the same, but sometimes, like above, `evict_other_queue` gets swapped with `evict_current_queue` yet the `evict_fill_queue` is not updated.  Either reducing it to two, current and something else, or a large comment somewhere explaining when they're all used and updated would help.

src/evict/evict_lru.c
```diff
@@ -909,20 +940,20 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
...
-	__wt_spin_lock(session, &cache->evict_queue_lock);
-	if (cache->evict_current_queue->evict_current == NULL)
+	if (__evict_queue_empty(queue)) {
+		/*
+		 * This score varies between 0 (if the queue hasn't been empty
+		 * for a long time) and 100 (if the queue has been empty the
+		 * last 10 times we filled up.
+		 */
+		cache->evict_empty_score = WT_MIN(100,
+		    cache->evict_empty_score + WT_EVICT_EMPTY_SCORE_BUMP);
```
sueloverso: A few things here:
1.  The comment describing the meaning of the score is buried here.  Maybe move it to `cache.h`?
2.  The conditional change is confusing or possibly wrong.  The original conditional checked if the previous `cache->evict_current_queue` was empty before replacing it with the newly filled queue.  The new conditional checks if the newly filled `queue` is empty.
3.  I think the entire, fixed, conditional and setting of the stats needs to move to where `cache->evict_current_queue` is set.  We're interested in knowing if the last batch of pages were fully processed or not.
4.  This code now has two locations where `cache->evict_current_queue` is set.  Previously there was only one place, here at the bottom of this function.  Now it is earlier in this fucntion and also in `evict_get_ref`.  That seems like it may end up error-prone.

src/evict/evict_lru.c
```diff
@@ -1149,6 +1186,14 @@ __evict_push_candidate(WT_SESSION_IMPL *session,
 	evict->btree = S2BT(session);
```
sueloverso: Not shown here, but there is a comment without a closing paren at line 1170 in this function.

src/evict/evict_lru.c
sueloverso: A comment or perhaps a stat to determine this would be good.

src/evict/evict_lru.c
```diff
@@ -1421,30 +1467,67 @@ __evict_get_ref(
...
+
+	/*
+	 * Check if the current queue needs to change.
+	 * The current queue could have changed while we waited for the lock.
+	 */
+	queue = cache->evict_current_queue;
+	other_queue = cache->evict_other_queue;
+	if (__evict_queue_empty(queue) && !__evict_queue_empty(other_queue)) {
+		cache->evict_current_queue = other_queue;
+		cache->evict_other_queue = queue;
```
sueloverso: As mentioned earlier this is now a new location where `evict_current_queue` is set and the conditional to set stats should be copied here too (or refactored so there's only one place it is updated).  It seems that this code is now more complicated beyond just getting an eviction ref.

src/include/cache.h
```diff
@@ -110,14 +110,19 @@ struct __wt_cache {
...
 	WT_SESSION_IMPL *walk_session;	/* Eviction pass session */
+	WT_DATA_HANDLE
+		*evict_file_next;	/* LRU next file to search */
 	WT_SPINLOCK evict_queue_lock;	/* Eviction current queue lock */
 	WT_EVICT_QUEUE evict_queues[WT_EVICT_QUEUE_MAX];
-	WT_EVICT_QUEUE *evict_current_queue;/* LRU current queue in use */
-	uint32_t evict_queue_fill;	/* LRU eviction queue index to fill */
+	WT_EVICT_QUEUE *evict_current_queue; /* LRU current queue in use */
+	WT_EVICT_QUEUE *evict_fill_queue;    /* LRU next queue to fill */
+	WT_EVICT_QUEUE *evict_other_queue;   /* LRU queue not in use */
```
sueloverso: The difference between `evict_fill_queue` and `evict_other_queue` is subtle.

src/include/cache.h
```diff
@@ -110,14 +110,19 @@ struct __wt_cache {
...
+	WT_EVICT_QUEUE *evict_urgent_queue;  /* LRU urgent queue */
 	uint32_t evict_slots;		/* LRU list eviction slots */
-	WT_DATA_HANDLE
-		*evict_file_next;	/* LRU next file to search */
-	uint32_t evict_max_refs_per_file;/* LRU pages per file per pass */
+#define	WT_EVICT_EMPTY_SCORE_BUMP	10
+#define	WT_EVICT_EMPTY_SCORE_CUTOFF	10
+	uint32_t evict_empty_score;	/* LRU score of how often queues are
```
sueloverso: AFAICT, this value starts at 0.  It seems like it should start at 100 since queues are empty when we start.  I'd also prefer `#define WT_EVICT_EMTPY_SCORE_MAX 100`.

src/evict/evict_lru.c
michaelcahill: 1. I've moved the comment into `cache.h`.
2. I disagree about the condition -- we have figured out which queue to fill above, in the ommon case we don't choose the current queue the net is the same as previously.  If we are evicting so fast that we start filling the current queue, we want to know whether it was drained and use that for the empty score.
3. I've moved this code earlier.
4. I've changed the code to just mark the queue being filled empty after locking so that subsequent requests for a page will switch to the other queue.

src/evict/evict_lru.c
michaelcahill: As mentioned above, I've made this the one place that switches queues.

The eviction server can't do it efficiently because sometimes it has to fill the current queue in order to keep up with the rate of eviction (e.g., in medium-btree).

src/include/cache.h
michaelcahill: It will be bumped the first time each queue is filled, so the server will go into busy mode immediately.  I've added the named constant.

src/evict/evict_lru.c
michaelcahill: I've updated the comment -- the issue is that if there is just the eviction server, it has to service the urgent queue or applications can get stuck.

src/evict/evict_lru.c
michaelcahill: Agreed -- it started out as something more subtle but now is equivalent to this.

src/evict/evict_lru.c
michaelcahill: I've made the change suggested above so that the eviction server just sets `queue->evict_current = NULL` and doesn't need to wait on the queue lock or switch queues.  I've also added a longer comment to `cache.h`.

The reason for `evict_other_queue` is to avoid repeating logic like this:

```c
other_queue = cache->evict_queues + 1 - (cache->evict_current_queue - cache->evict_queues);
```

I think it's clearer to keep an extra queue pointer around.

PR #2950, opened 2016-08-12
WT-2711 Change statistics log configuration options

keithbostic:
Ignore statistics_log.path configuration as a reconfiguration option.


agorrod:

PR #2952, opened 2016-08-12
WT-2822 panic mutex and other functions that cannot fail

keithbostic:
@agorrod, @michaelcahill, here's a branch with some changes that might be reasonable.

The first set in 9038576 is changing the ticket-lock read/write lock/unlock functions to not return errors, which shouldn't be controversial. (The only errors that could actually happen in those functions were calls to `__wt_verbose`; verbose messages on every lock/unlock seems useless, so I just pulled them.)

The second set in b186c79 is changing the adaptive pthread mutex functions to panic if there is a locking error. This one is more interesting: imagine running with the adaptive pthread mutex functions, and the lock call fails, and we panic, but before returning out of the library, we potentially have multiple threads working on the same chunks of memory because they both hold the lock. I think it's legitimate to question if we can depend on recovery to deal with all possible variants of that? My inclination would be to change the underlying functions to not only panic, but also exit. I'm uncomfortable continuing to run, the failure should never happen, and I'm **really** not interested in adding error handling for every call we make to `__wt_spin_lock/unlock`.

The third set in e309496 is changing the condition mutex functions to not return errors. I don't think these functions have the same problem as the adaptive pthread mutex functions (that is, they only perform signalling functions, not serialization functions), so I don't think these changes should be controversial, and a panic on failure is sufficient.

The fourth set in 61d7dce is changing `__wt_verbose()` to not return errors. I don't think it's worth the error-handling complexity (note we already ignore potential errors from `__wt_err()` and `__wt_errx()`).

One last note: `__wt_try_readlock()` and `__wt_try_writelock()` could be changed to return a boolean instead of an int (busy or not-busy), which would give us a bit more clarity in handling returns from those functions (right now we use `WT_RET` for example). I didn't make that change.


agorrod:
@keithbostic I was expecting to dislike this change, but I actually really like it.

> __wt_try_readlock() and __wt_try_writelock() could be changed to return a boolean instead of an int

I agree that sounds worthwhile. I'll take a look at how that change works this afternoon.

agorrod:
>>__wt_try_readlock() and __wt_try_writelock() could be changed to return a boolean instead of an int

> I agree that sounds worthwhile. I'll take a look at how that change works this afternoon.

After digging into this more, switching trylock to a boolean makes our error handling more complex rather than less, since we rely on propagating the EBUSY return all the way out in a number of places.

So - I'm happy with this change now. Once testing is stabilized I'm inclined to merge. It's significant enough that I think @michaelcahill should agree too.

michaelcahill:
@keithbostic, these changes look good to me.  I'll get them merged.

As a library, I don't think WiredTiger should unilaterally exit in any situation, so if we think there are real situations where lock calls can fail in the wild, we should consider further protection against data corruption.  In particular, maybe we should check for panic before every write?

keithbostic:
@michaelcahill said:

> As a library, I don't think WiredTiger should unilaterally exit in any situation, so if we think there are real situations where lock calls can fail in the wild, we should consider further protection against data corruption. In particular, maybe we should check for panic before every write?

I agree, I'll push that change.

We made the same check of panic-before-IO in Berkeley DB (although it was a lot more vulnerable because of its recovery model). Berkeley DB currently also checks panic before read, write, rename, truncate and unlink.

keithbostic:
I'm glad you guys like the change!

For review: @sueloverso, the most interesting changes are all in the logging code and mutex signaling code. Review shouldn't be too hard (and I think it will be easier to review the individual commits than the full diff, that way there's a single pattern to look for).

One place that deserves a real check is `__wt_cond_wait_signal()`; I changed the code path on `__wt_epoch()` error, and I've messed up that code before, so it makes me nervous.

michaelcahill:
Thanks @keithbostic -- this lgtm, I'm going to go ahead and merge.

PR #2953, opened 2016-08-13
SERVER-25247 MongoDB primary node crashes after SIGINT (ctrl+c)

keithbostic:
@agorrod, I'm giving up on SERVER-25247, I just don't see it.

Here's a workaround that I think will at least avoid the symptom, and given the number of sessions MongoDB allocates, it's a useful change to make.


agorrod:
Thanks @keithbostic lgtm. I've opened WT-2838 to cover the particular change here and I'll link the server ticket, since this is more of a workaround than a fix.

PR #2954, opened 2016-08-13
WT-2835 WT_CONNECTION.leak-memory can skip memory map and cache cleanup

keithbostic:

### DIFF
--- a/src/btree/bt_discard.c
+++ b/src/btree/bt_discard.c
@@ -103,6 +103,15 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	/* Update the cache's information. */
 	__wt_cache_page_evict(session, page);
 
+	dsk = (WT_PAGE_HEADER *)page->dsk;
+	if (F_ISSET_ATOMIC(page, WT_PAGE_DISK_ALLOC))
+		__wt_cache_page_image_decr(session, dsk->mem_size);
+
+	/* Discard any mapped image. */
+	if (F_ISSET_ATOMIC(page, WT_PAGE_DISK_MAPPED))
+		(void)S2BT(session)->bm->map_discard(
+		    S2BT(session)->bm, session, dsk, (size_t)dsk->mem_size);
+
 	/*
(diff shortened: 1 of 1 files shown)


keithbostic:
@agorrod, @michaelcahill:

It would be pretty easy to introduce this problem again, if we updated some cache information in a function page-discard calls, there's nothing to ensure page-discard functions do nothing other than free memory.

What would you think of:
```
diff --git a/src/os_common/os_alloc.c b/src/os_common/os_alloc.c
index cfc7b80..101f873 100644
--- a/src/os_common/os_alloc.c
+++ b/src/os_common/os_alloc.c
@@ -301,8 +301,12 @@ __wt_free_int(WT_SESSION_IMPL *session, const void *p_arg)
         * !!!
         * This function MUST handle a NULL WT_SESSION_IMPL handle.
         */
-       if (session != NULL)
+       if (session != NULL) {
                WT_STAT_FAST_CONN_INCR(session, memory_free);
 
+               if (F_ISSET(S2C(session), WT_CONN_LEAK_MEMORY))
+                       return;
+       }
+
        free(p);
 }
```
It's a cache miss in every free call, but **shrug**?

agorrod:
@keithbostic The change lgtm, I'll go ahead and merge. In regards to your other question in the ticket - won't that mean we never free any memory? I don't think that's a good idea - it would lead to ENOMEM in a bunch of new situations.

keithbostic:
> In regards to your other question in the ticket - won't that mean we never free any memory? I don't think that's a good idea - it would lead to ENOMEM in a bunch of new situations.

@agorrod, I think it means we won't free any memory after we call WT_CONNECTION.close with leak_memory configured (that's what sets WT_CONN_LEAK_MEMORY), otherwise nothing changes. Or am I missing something?

agorrod:
Oh, sorry. I didn't check when the flag was set - I'd presumed it was set when the option is configured (i.e: at startup). Regardless I'd probably rather keep free simpler - surprisingly I have seen free contribute significantly to performance profiles on some workloads - that's generally in the allocator not our code - but I'd rather not add to the complexity.

PR #2955, opened 2016-08-13
WT-2631 nullptr is passed for parameters marked with attribute non-null

keithbostic:
@agorrod, I believe that I have clean Evergreen runs, so this should be ready to merge.


agorrod:
Thanks @keithbostic lgtm.

PR #2957, opened 2016-08-13
WT-2783 wtperf multi-btree.wtperf dumps core on Mac

keithbostic:
@sulabhM, I didn't realize you owned the JIRA ticket until I was already done, I thought it was a backlog ticket. I apologize for that, my mistake.

@agorrod, we could also just agree we're never going to use wtperf to test for memory leaks, and remove every memory free call from the program -- we're playing relatively fast-and-loose with memory in wtperf at the moment and that would make the whole thing a non-issue.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -95,7 +95,7 @@ config_unescape(char *orig)
 
 /*
  * config_init --
- *	CONFIG structure initializaiton, based on a source configuration.
+ *	CONFIG structure initialization, based on a source configuration.
  */
 int
 config_init(CONFIG *dest, const CONFIG *src)
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -94,11 +94,11 @@ config_unescape(char *orig)
 }
 
 /*
- * config_init --
+ * config_copy --
  *	CONFIG structure initialization, based on a source configuration.
  */
 int
-config_init(CONFIG *dest, const CONFIG *src)
+config_copy(CONFIG *dest, const CONFIG *src)
 {
 	CONFIG_QUEUE_ENTRY *conf_line, *tmp_line;
 	size_t i;
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -122,6 +122,10 @@ config_init(CONFIG *dest, const CONFIG *src)
 		for (i = 0; i < src->table_count; i++)
 			dest->uris[i] = dstrdup(src->uris[i]);
 	}
+
+	if (src->async_config != NULL)
+		dest->async_config = dstrdup(src->async_config);
+
 	dest->ckptthreads = NULL;
 	dest->popthreads = NULL;
 	dest->workers = NULL;
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -94,19 +94,28 @@ config_unescape(char *orig)
 }
 
 /*
- * config_assign --
- *	Assign the src config to the dest, any storage allocated in dest is
- * freed as a result.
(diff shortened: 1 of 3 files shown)


sulabhM:
@keithbostic : No prob.

agorrod:
@sulabhM Would you mind doing a first round of review please?

bench/wtperf/config.c
```diff
@@ -94,20 +94,29 @@ config_unescape(char *orig)
...
+		dest->home = dstrdup(src->home);
+	if (src->monitor_dir != NULL)
+		dest->monitor_dir = dstrdup(src->monitor_dir);
+	if (src->partial_config != NULL)
+		dest->partial_config = dstrdup(src->partial_config);
+	if (src->reopen_config != NULL)
+		dest->reopen_config = dstrdup(src->reopen_config);
+	if (src->base_uri != NULL)
+		dest->base_uri = dstrdup(src->base_uri);
+
```
sulabhM: Should "char *async_config" also be duplicated from destination to source config.
If free async_config in the config_free() function, wouldn't we be double freeing this one in case not NULL.

sulabhM:
@keithbostic : Other than my comment on not having duplicated "async_config", this change LGTM.

agorrod:
> we could also just agree we're never going to use wtperf to test for memory leaks

I don't like that idea - allowing memory leaks in wtperf is likely to lead to memory pressure which will skew performance results. Some of the memory management complexity also ties in to how the multi-table and multi-database configurations work, I'm reluctant to reduce error checking there by not thinking about cleanup. I'm also reluctant to give up the ability to run auto-checking tools on wtperf.

Could you point out which part of this change is the bug fix - it isn't obvious to me.

bench/wtperf/config.c
keithbostic: @sulabhM 
> Should "char *async_config" also be duplicated from destination to source config.

Yes. Thank you, Sulabh -- fixed and pushed.

keithbostic:
@agorrod 

>  allowing memory leaks in wtperf is likely to lead to memory pressure which will skew performance results.

Most (all?) of this memory is allocated and never freed until the end of the run, so I doubt that's a problem. That said, I agree we might want to run auto-checking tools on wtperf.

keithbostic:
@agorrod 

> Could you point out which part of this change is the bug fix - it isn't obvious to me.

This change is really the minimal fix, there's no other work here.

The bug is there are multiple CONFIG structures, and when creating a new CONFIG structure we **assign** a previous CONFIG structure to it in order to initialize the shared run configuration information.  That might have worked with a single CONFIG structure, but with multiple CONFIG structures, where we're subsequently allocating and modifying CONFIG structure fields, then freeing them when we're finished with them, it's just asking for problems.

The better change is probably to split the CONFIG structure into two parts, the per-database information and a shared per-run configuration taken from `wtperf_opt.i`, that way we can get rid of the CONFIG assignment and we're unlikely to introduce new bugs going forward.

keithbostic:
@agorrod, I took a run at splitting the CONFIG structure into two parts, the per-database structure and the run-time options shared across all CONFIG structures. It's messy because it touches lots of stuff, but otherwise it's not too bad -- let me know if you want me to create a ticket and clean it up for review and possible merge.

agorrod:
> I took a run at splitting the CONFIG structure into two parts

I like the idea of that change. I'd be happy for you to push it into this branch if you think it's an improvement.

keithbostic:
@agorrod, it's definitely an improvement, I think (and gets us closer to your goal of generalizing the parser).

Let's do it in a separate branch, though.

PR #2958, opened 2016-08-14
WT-2834 Shared bloom filters allow only a partial shortcut.

ddanderson:
This fixes an error in cursor join evaluations that may occur with bloom filters that are shared when false positives are found.  Things to note:

The problem only happens when Bloom filters are specified for parts of the join.  It is less likely when Bloom filters are oversized, and more likely when Bloom filters are undersized. That is the problem occurs with the handling of false positives, which become less likely as the Bloom filter is sized larger. The problem only happens when Bloom filters are shared; that occurs when Bloom filters are specified for multiple parts of a conjunction, and the filters have been given the same size parameter.

The problem will never result in an item not being returned that should be returned.  It will result in item(s) being returned that do not satisfy the conditions of the join.  The fix should ensure that all items returned satisfy the join.

### DIFF
--- a/test/csuite/wt2834_join_bloom_fix/main.c
+++ b/test/csuite/wt2834_join_bloom_fix/main.c
@@ -46,36 +46,6 @@ void (*custom_die)(void) = NULL;
 
 void *populate(TEST_OPTS *opts);
 
-typedef struct {
-	uint32_t seedw;
-	uint32_t seedz;
-} RAND;
-
-static void rand_init(RAND *r)
-{
-	r->seedw = 521288629;
-	r->seedz = 362436069;
-}
-
-/*
- * Returns a random 32 bit integer.
- * We use Marsaglia's Multiply-with-carry method.
- */
-static uint32_t rand_next(RAND *r)
-{
-	uint32_t w;
-	uint32_t z;
-
-	w = r->seedw;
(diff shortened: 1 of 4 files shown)


ddanderson:
@michaelcahill, can you please review?

test/csuite/wt2834_join_bloom_fix/main.c
```diff
@@ -0,0 +1,231 @@
...
+ * Failure mode: We get results back from our join.
+ */
+void (*custom_die)(void) = NULL;
+
+#define	N_RECORDS	100000
+#define	N_INSERT	1000000
+
+void *populate(TEST_OPTS *opts);
+
+typedef struct {
```
agorrod: @ddanderson You can use the WiredTiger __wt_random functions from the test code rather than rolling your own. See https://github.com/wiredtiger/wiredtiger/blob/develop/test/csuite/wt2719_reconfig/main.c for an example.

michaelcahill:
Thanks @ddanderson, I switched to the builtin random number generator as @agorrod suggested. Otherwise this lgtm -- assuming it passes tests with that change, I'll merge.

ddanderson:
@michaelcahill, thanks for the comments and the changes!

PR #2959, opened 2016-08-15
WT-2719 Allow make check without verbose configured.

michaelcahill:


michaelcahill:
@daveh86 can you please review?

michaelcahill:
Thanks for the review @daveh86, merging.

PR #2960, opened 2016-08-15
WT-2719 Fix memory leak in reconfig test

daveh86:


daveh86:
@sulabhM , would you mind taking a look?

I have run this against the failing Jenkins test and it resolves the leak in question.

sulabhM:
LGTM.

I noticed we do not free memory in case when we error out. As a separate ticket we can evaluate if we need to do so in this case and/or all of the csuite.

PR #2962, opened 2016-08-15
WT-2839 lint: Ignoring return value of function

keithbostic:
@ddanderson, for your review/consideration.


ddanderson:
@keithbostic, these LGTM.  I'll merge, thanks!

PR #2963, opened 2016-08-15
WT-2840 clang analysis: garbage values

keithbostic:
@ddanderson, for your review/consideration.


agorrod:
Thanks @keithbostic lgtm.

PR #2964, opened 2016-08-15
WT-2841 Initialize verbose checkpoint timer at start of checkpoint

daveh86:
This initialization was inadvertently removed in WT-2764


daveh86:
@sulabhM , mind taking a look at this one?

sulabhM:
LGTM

PR #2966, opened 2016-08-16
WT-2847 Merge fair locks into read/write locks.

keithbostic:
@agorrod, I think the fair locks are the same as the rwlock write locks -- for your review/consideration.


agorrod:
Thanks @keithbostic lgtm. @michaelcahill this is a change in locking - it looks safe to me, but I'd appreciate another set of eyes before merging.

michaelcahill:
As discussed, I'm all for simplifying our locking -- I'd like to take a look at doing the same exercise for spin locks next.  This lgtm.

FWIW, we use "among" already (in 9 places).

keithbostic:
@michaelcahill 
> I'd like to take a look at doing the same exercise for spin locks next.

What are you thinking about doing with respect to spin locks?

keithbostic:
@agorrod, @michaelcahill -- since you've both signed off, I'm going to merge.

PR #2967, opened 2016-08-16
WT-2695 Integrate s390x accelerated crc32c support

keithbostic:
I think this is ready for review/merge.


keithbostic:
Note, may need [WT-2629 Don't make stacks executable in assembly source](https://github.com/wiredtiger/wiredtiger/commit/0f7ae730d9ab28aaa51a938075502f639947d26f).

keithbostic:
Smoke-testing this change against the WiredTiger big-endian checksum-in-software implementation, this branch is significantly faster, with around a 25x improvement.

SConstruct
```diff
@@ -248,6 +248,8 @@ condition_map = {
     'POSIX_HOST' : env['PLATFORM'] == 'posix',
     'POWERPC_HOST' : False,
     'WINDOWS_HOST' : env['PLATFORM'] == 'win32',
+    'X86_HOST' : True,
+    'ZSERIES_HOST' : False,
```
michaelcahill: FTR, we'll need to make the corresponding changes to MongoDB's `SConscript` file for WiredTiger before this change can be merged.

test/csuite/wt2695_checksum/sw.c
```diff
@@ -0,0 +1,49 @@
...
+
+/*
+ * Build the software version of the CRC code.
+ */
+#define	__wt_cksum	__wt_cksum_notused
+uint32_t __wt_cksum_notused(const void *, size_t);
+#define	__wt_cksum_init	__wt_cksum_init_notused
+void __wt_cksum_init_notused(void);
+
+#include CRC_PATH
```
michaelcahill: This is pretty much inscrutable without context.

I would have written this as `#include "../../src/checksum/x86/checksum.c"` and not bothered with a macro.

michaelcahill:
@keithbostic, thanks, lgtm -- I'll merge.

SConstruct
keithbostic: Sorry, I should have raised this point in a comment (I had a note to myself to re-visit this once the drop happened, but on second thought, that's obviously too late). Thanks for taking this on, @daveh86.

test/csuite/wt2695_checksum/sw.c
keithbostic: @michaelcahill, unless I'm mistaken, you have to use `$(top_srcdir)` in order for all build configurations to work? If that's wrong, happy to make the change.

michaelcahill: We always have "-I$(top_srcdir)/src/include" in CPPFLAGS, that's what I was relying on.  I don't feel that strongly, we probably won't look at this again, it just seemed unnecessarily complicated to have a macro for a #include.

PR #2968, opened 2016-08-17
WT-2829 Fix missing define for BerkeleyDB

daveh86:


test/format/format.h
```diff
@@ -29,6 +29,9 @@
 #include "test_util.h"
 
 #ifdef BDB
+#ifndef DB_DBM_HSEARCH
```
agorrod: This deserves a comment explaining why it's necessary.

agorrod:
@keithbostic Could you review this pull request please? We are open to alternative suggestions. As context - a solution is required before we can switch to using ``--enable-strict`` in Jenkins

keithbostic:
@daveh86, what error do you see, and how can I reproduce it?

I vaguely recall seeing errors around this #ifdef at some point, but building with  `--enable-strict` on `JenkinsBuild` doesn't complain, so I'm not clear on what we're fixing.

keithbostic:
@daveh86, never mind, I found the error, I see what's happening.

I guess the alternative is to add the #ifdef in the `Makefile.am` files:
```
diff --git a/ext/test/kvs_bdb/Makefile.am b/ext/test/kvs_bdb/Makefile.am
index c7d0f58..6e125c0 100644
--- a/ext/test/kvs_bdb/Makefile.am
+++ b/ext/test/kvs_bdb/Makefile.am
@@ -1,4 +1,4 @@
-AM_CPPFLAGS = -I$(top_builddir) \
+AM_CPPFLAGS = -DDB_DBM_HSEARCH=0 -I$(top_builddir) \
     -I$(top_srcdir)/src/include -I$(BERKELEY_DB_PATH)/include
 
 noinst_LTLIBRARIES = libwiredtiger_kvs_bdb.la
diff --git a/test/format/Makefile.am b/test/format/Makefile.am
index 5d946e5..68321cb 100644
--- a/test/format/Makefile.am
+++ b/test/format/Makefile.am
@@ -2,7 +2,7 @@ AM_CPPFLAGS = -I$(top_builddir)
 AM_CPPFLAGS +=-I$(top_srcdir)/src/include
 AM_CPPFLAGS +=-I$(top_srcdir)/test/utility
 if HAVE_BERKELEY_DB
-AM_CPPFLAGS +=-DHAVE_BERKELEY_DB
+AM_CPPFLAGS +=-DHAVE_BERKELEY_DB -DDB_DBM_HSEARCH=0
 AM_CPPFLAGS +=-DBERKELEY_DB_PATH=\"$(BERKELEY_DB_PATH)\"
 AM_CPPFLAGS +=-I$(BERKELEY_DB_PATH)/include
 endif
```

But I think what you did has more clarity, so I like your fix better.

I'm going to hack the comment (hope that's OK!), and then I'll merge.

PR #2969, opened 2016-08-17
WT-2678 The metadata should not imply that an empty value is true

keithbostic:
@michaelcahill, would you be willing to take a fast look at the changes here? I think it's correct, but you wrote all of this code and I'd take comfort from your agreement.

cc: @agorrod


michaelcahill:

PR #2970, opened 2016-08-19
WT-2795 Update read-only documentation

sueloverso:
Explicitly state that parallel read-only handles depend on WT home directory write permission.

@agorrod please review this change to the read-only documentation.


agorrod:
Thanks @sueloverso lgtm.

PR #2971, opened 2016-08-21
WT-2842 split wtperf's configuration into per-database and per-run parts

keithbostic:

The work of making the wtperf configuration code independent of wtperf is here, with one notable exception: in the thread configuration, `cfg->workload` setting is still mixed up with configuration parsing. There are a couple of other odds-and-ends, but that's the only real issue. We could move the `WORKLOAD` structure from `CONFIG` into `CONFIG_OPTS`, but I'm not sure that's the right change -- it's fine if we intend to have standard "workload" parsing in whatever generic configuration code we end up creating (and that looks like where you were going in `testutil_parse_opts()`, so it makes sense). Anyway, if you want that change made, let me know.

One other change we should make before merge is to rename the `CONFIG` structure. It's no longer base configuration at all, it's the runtime database environment information. We can't use `DB`, but we could use `RTE` or `DBH`?


agorrod:
> making the wtperf configuration code independent of wtperf is here

I like that this change brings sharing configuration parsing a step closer. I'm inclined to move there incrementally, rather than expand this change to cover that as well.

I agree that parsing workload groups is a wrinkle - it'd be nice to figure out a more elegant solution, either have our configuration parsing support nested configurations, or unwind the wtperf workload syntax to fit more easily into a generic configuration parser. I've mentioned in other forums that I'd like the ability to specify multiple workload phases to wtperf in the future and there is demand for expanding the thread-group syntax (to eg support different table types), and changes here should take those future directions into consideration.

> One other change we should make before merge is to rename the CONFIG structure. It's no longer base configuration at all, it's the runtime database environment information. We can't use DB, but we could use RTE or DBH?

I see what you are saying. The CONFIG structure holds the shared state for wtperf, I don't think it's particular to database environment information though. If we change the name I'd be OK with "CONTEXT" or "SHARED_STATE". Changing the name of that structure is going to touch nearly all the wtperf code, so I'd prefer if that change was made in a separate pull request that doesn't have any functionality changes.

keithbostic:
I agree with moving incrementally -- my goal was to stop copying the shared global configuration to/from the `CONFIG` structures. For the same reason, I'm fine with leaving workload parsing/configuration in its current state.

I don't like `CONTEXT` or `SHARED_STATE` much. It's the wtperf information for a single WiredTiger database, of which there may be many, and that information isn't shared. (Ummm, that's not right, the information is shared between wtperf threads in a WiredTiger database, is that what you meant?)

Here's an idea: we don't have a `WTPERF` structure in wtperf! It could be the `WTPERF` structure, mapping one-to-one to a WiredTiger database? It's the structure `wtperf` is measuring, makes sense to me.

If the structure renaming will be a separate pull request, this branch is ready for merge. We shouldn't do it on a weekend or when it will affect @michaelcahill's eviction work, since it has the potential to break performance testing in subtle ways if there's a bug.

PR #2972, opened 2016-08-23
WT-2850 clang 4.1 attribute warnings when building

keithbostic:


sueloverso:
LGTM.  Thanks for working through this with me!

PR #2973, opened 2016-08-24
WT-2846 Fix bugs related to reconfiguring eviction settings.

agorrod:
The code to support reconfiguring eviction worker threads had
several bugs. Implemented an abstracted utility thread group API
and switch eviction workers over to using the new abstraction.


agorrod:
@sueloverso Would you mind taking a look at this branch and letting me know if you like the direction it's going in? My plan is to move the async and LSM manager threads over to use the new API as well, so I'd be particularly interested in hearing about any deficiencies you can foresee there.

src/include/thread_group.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+struct __wt_worker_thread {
+	WT_SESSION_IMPL *session;
+	u_int id;
+	wt_thread_t tid;
+#define	WT_WORKER_THREAD_RUN		0x01
+	uint32_t flags;
+
+	/* The runner function used by all workers. */
+	int (*run_func)(WT_SESSION_IMPL *session, WT_WORKER_THREAD *context);
```
agorrod: The function pointer is duplicated in the group and each thread structure - that felt better to me than putting a pointer to the group into each worker structure.

src/include/thread_group.h
```diff
@@ -0,0 +1,56 @@
...
+	uint32_t	 current_workers;	/* Number of active workers */
+
+	WT_RWLOCK	*lock;     /* Protects group changes */
+
+	/*
+	 * Condition signalled when wanting to wake up threads that are
+	 * part of the group - for example when shutting down. This condition
+	 * can also be used by group owners to ensure state changes are noticed.
+	 */
+	WT_CONDVAR      *wait_cond;
```
agorrod: This field is accessed directly by the eviction code, we could wrap condition wakes in the API if that feels cleaner.

sueloverso:
@agorrod I plan to do a more detailed review as well.  My first impressions are:
* I suggest removing all uses of "util" and "utility" in the naming, comments and strings.  I find it confusing to read because it makes me think of threads that are in support of the `wt` command, or `testutil`, etc.  I would prefer just `WT_THREAD` and `WT_THREAD_GROUP` and `wt_thread_group_create`.  Having a prefix implies there may be other kinds too.
* I'm not sure how or if it can fit, but I will toss this out here now - can or should this be expanded to consider threads like logging uses?  Logging has 3 threads, but they are all running different functions and do different things.  They're not a pool like the other 3 subsystems.  It would be nice if all internal threads used the same thing (I guess that includes sweep and checkpoints as well).

dist/api_data.py
```diff
@@ -520,6 +520,7 @@ def __cmp__(self, other):
             'split',
             'temporary',
             'transaction',
+            'util_thread',
```
sueloverso: Per my general comment - I'd call this "thread" or "thread_group"

src/evict/evict_lru.c
```diff
@@ -168,29 +171,27 @@ __wt_evict_server_wake(WT_SESSION_IMPL *session)
...
 /*
- * __evict_thread_run --
- *	General wrapper for any eviction thread.
+ * __wt_evict_worker_run --
+ *	General wrapper for an eviction worker thread.
  */
-static WT_THREAD_RET
-__evict_thread_run(void *arg)
+int
+__wt_evict_worker_run(WT_SESSION_IMPL *session, WT_WORKER_THREAD *worker)
```
sueloverso: To me, `worker` sounds subordinate, implying some other thread is more special or controlling things.  Whereas the old name `thread` made it sound like all threads are equal.  If all eviction threads are using it we should retain the name, adding the prefix, and keep the comment to make it clear any/all eviction threads are in here.

src/evict/evict_lru.c
```diff
@@ -168,29 +171,27 @@ __wt_evict_server_wake(WT_SESSION_IMPL *session)
...
 
-	session = arg;
 	conn = S2C(session);
 	cache = conn->cache;
 
 #ifdef HAVE_DIAGNOSTIC
-	if (session == conn->evict_session)
-		WT_ERR(__wt_epoch(
-		    session, &cache->stuck_ts));	/* -Wuninitialized */
+	WT_ERR(__wt_epoch(session, &cache->stuck_ts));	/* -Wuninitialized */
```
sueloverso: Should this only be initialized with `worker->id == 0`?  With this change, a reconfiguration to add a thread will unconditionally clear/reset the cache timestamp.

src/evict/evict_lru.c
```diff
@@ -317,112 +319,33 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
...
-/*
  * __wt_evict_create --
  *	Start the eviction server thread.
  */
 int
 __wt_evict_create(WT_SESSION_IMPL *session)
 {
 	WT_CONNECTION_IMPL *conn;
-	uint32_t session_flags;
+	WT_DECL_RET;
```
sueloverso: Where are we using `ret` in here?  WT_RET doesn't need it.  So I think we can get rid of that and return (0) at the end.

src/include/thread_group.h
sueloverso: Per my general comment, if we can figure out a clean way to have different functions, then we'd want to get rid of the function in the thread group since different worker threads in the same group could have different functions.  I'd also prefer these all be named WT_THREAD (remove WORKER) - similar to the util comment earlier.

src/include/thread_group.h
sueloverso: Looking at other subsystems, they all need a condvar.  Some of the code feels a bit more clunky but we can add a wrapper later if needed.  It occurs to me that in logging, each different thread has its own condvar so maybe logging would needs 3 different thread groups each with one thread?

src/support/thread_group.c
```diff
@@ -0,0 +1,330 @@
...
+ *	All rights reserved.
+ *
+ * See the file LICENSE for redistribution information.
+ */
+
+#include "wt_internal.h"
+
+/*
+ * __wt_util_thread_run --
+ *	General wrapper for any eviction thread.
```
sueloverso: Remove "eviction" from the comment.

src/support/thread_group.c
```diff
@@ -0,0 +1,330 @@
...
+		if (LF_ISSET(WT_WORKER_CAN_WAIT))
+			session_flags = WT_SESSION_CAN_WAIT;
+		if (F_ISSET(conn, WT_CONN_LAS_OPEN))
+			FLD_SET(session_flags, WT_SESSION_LOOKASIDE_CURSOR);
+		WT_ERR(__wt_open_internal_session(conn, "utility-worker",
+		    false, session_flags, &worker->session));
+		if (LF_ISSET(WT_WORKER_PANIC_FAIL))
+			F_SET(worker, WT_WORKER_PANIC_FAIL);
+		worker->id = i;
+		worker->run_func = group->run_func;
```
sueloverso: It appears the only reason to have a group function is to have it as a holding place for the worker structure to copy it.  It may be needed when adding a new thread unless the caller can pass it in.

src/support/thread_group.c
```diff
@@ -0,0 +1,330 @@
...
+__wt_util_thread_group_start_one(
+    WT_SESSION_IMPL *session, WT_WORKER_THREAD_GROUP *group, bool wait)
+{
+	WT_DECL_RET;
+
+	if (group->current_workers >= group->max)
+		return (0);
+
+	if (!wait) {
+		if (__wt_try_writelock(session, group->lock) != 0)
```
sueloverso: This can be collapsed.

sueloverso:
@agorrod I finished my 2nd pass through this with a bunch of minor comments.  I think it will be a good thing.  Our conversations so far have been from the POV of thread groups that the user can see/think about and reconfigure: async, LSM and eviction.  But I've been thinking about this also from inside.  I think we want all internal WT threads to use this so that all have a consistent structure and control mechanism for starting/stopping.

agorrod:
Thanks for the feedback @sueloverso, I believe I've addressed your comments. Would you mind doing another round of review and merging if you are happy?

src/support/thread_group.c
```diff
@@ -0,0 +1,328 @@
...
+	if (ret != 0 && F_ISSET(thread, WT_THREAD_PANIC_FAIL))
+		WT_PANIC_MSG(session, ret,
+		    "Unrecoverable utility thread error");
+
+	/*
+	 * The three cases when threads are expected to stop are:
+	 * * When recovery is done
+	 * * When the connection is closing
+	 * * When a shutdown has been requested via clearing the run flag.
+	 * Check otherwise fewer threads may be running than expected.
```
sueloverso: I removed this line - we're not checking the number of threads here.

src/include/thread_group.h
```diff
@@ -0,0 +1,56 @@
...
+
+/*
+ * WT_THREAD_GROUP --
+ *	Encapsulation of a group of utility threads.
+ */
+struct __wt_thread_group {
+	uint32_t	 alloc;    /* Size of allocated group */
+	uint32_t	 max;      /* Max threads in group */
+	uint32_t	 min;      /* Min threads in group */
+	uint32_t	 current_threads;/* Number of active threads */
```
sueloverso: I added a name field/arg so that different thread groups can have identifying names.

sueloverso:
I pushed some cosmetic changes.  This LGTM.

PR #2974, opened 2016-08-24
WT-2853 Don't force eviction if multiple cursors are pinning the page.

michaelcahill:
In some patterns of index lookups, we don't want the second cursor in a session to attempt forced eviction: it will just stall itself to no purpose.


ddanderson:
@michaelcahill, this fixes the stall for me - it was very apparent on OSX.  The progress pattern between the insert and get threads is extremely consistent now.  I created another pull request against this one: #2976 for some test modifications that made sense to me. Thanks!

agorrod:
@michaelcahill This change looks good to me - should I go ahead and merge?

test/csuite/wt2853_perf/main.c
```diff
@@ -0,0 +1,323 @@
...
+				fprintf(stderr, "*");
+			else
+				fprintf(stderr, ".");
+			time(&curtime);
+			if (curtime - prevtime > 5) {
+				fprintf(stderr, "\n"
+				    "GAP: %ld secs after %d inserts\n",
+				    curtime - prevtime, i);
+				threadargs->nfail++;
+			}
```
keithbostic: @ddanderson, this should probably be a call to `difftime()`? (Lint noticed.)

PR #2976, opened 2016-08-24
WT-2853 test improvements

ddanderson:
Added testutil_assert for failure condition and reduced test run length. Also cosmetic changes: remove unneeded vars, var declaration reordering, var renaming.


michaelcahill:
Thanks @ddanderson, lgtm.

PR #2978, opened 2016-08-24
WT-2843 Modify test for logging if truncate doesn't exist.

sueloverso:
@michaelcahill or @agorrod Please review this change to the `test/recovery/truncate-log` test.  It adds a step of adding a log record and then reading it back.  This passes in the current (with truncate) tree and fails when using `configure ac_cv_func_ftruncate=no`.

This only is the test portion.  Logging still has assumptions built in that it can truncate.


sueloverso:
I should state that I am working on the logging changes to go along with this.

michaelcahill:

PR #2980, opened 2016-08-25
WT-2859 Fix inconsistencies between fast/all stats usage

sulabhM:


sulabhM:
@agorrod : Please have a look

PR #2981, opened 2016-08-25
WT-2843 If there is no truncate available, manually zero the log file.

sueloverso:
@michaelcahill or @agorrod please review this change to handle log files when truncate is not available.  The change is actually pretty small since `log_zero` already exists.  I'm not thrilled with the goto in there but didn't want to make the code ugly to handle the first-time case.

This passes the new `truncated-log` test both with and without truncate configured.  We need to discuss where/how to get that setting into one of the Jenkins jobs to test this regularly.


agorrod:
Thanks @sueloverso lgtm. I'll go ahead and merge.

PR #2982, opened 2016-08-25
WT-2862 de-linted a test program

ddanderson:


ddanderson:
@keithbostic, can you take a quick look?

michaelcahill:
@ddanderson thanks, this LGTM -- I'll merge.

If there is more lint after the merge, we can always follow up with another change.

PR #2983, opened 2016-08-25
WT-2846 Add flags and fix for destroying thread group.

sueloverso:
@agorrod This fixes the leak and allows the failing `test/format` config finish successfully.


sueloverso:
... but the reconfig test is now failing for me locally.  We can talk tonight as this branch/path is not ready, clearly.

agorrod:
Thanks @sueloverso for cleaning up after me. I made a minor change and will merge.

PR #2984, opened 2016-08-25
WT-2863 WT on Windows does not work with UTF-8 paths

markbenvenuto:
WT on Windows assumes ASCII file names. I am changing it to support UTF-8 by translating all strings that refer to file names to UTF-16, and calling the Unicode Windows API.


michaelcahill:
Thanks @markbenvenuto, I've done a review pass for WT coding style, otherwise this LGTM.

I'll merge.

PR #2985, opened 2016-08-26
WT-2846 Fixes for new thread group code

agorrod:
If it were safe there would be no point locking. The test could fail if a reconfigure races with the eviction server starting a new thread.


agorrod:
autobuild: retest this please

agorrod:
Another fix in this branch: Signal the current eviction server before shutting down the thread group so we don't need to wait for condition variable to timeout

michaelcahill:

PR #2988, opened 2016-08-29
WT-2865 Fix a bug where the eviction server could panic after a WT_NOTFOUND

agorrod:
Recent changes to the eviction server led to a WT_NOTFOUND being propagated unexpectedly.

This isn't ready for merge yet - I'm still working on reproducing the original failure to confirm the fix.


michaelcahill:
@agorrod, when you're ready, this change lgtm.

agorrod:
@michaelcahill As discussed, I have not had any luck reproducing this failure today. It is very likely the root cause of the problem, since the failures we have seen have all been after connection close has started.

sueloverso:
@agorrod I am able to reproduce the problem on my AWS box semi-regularly.  Sometimes under 100 iterations, sometimes several hundred.  I'll start a run with the fix now merged.

PR #2989, opened 2016-08-29
WT-2868 Add sample_interval to checkpoint-stress.wtperf

sulabhM:


sulabhM:
@agorrod : Please review

agorrod:
Thanks @sulabhM lgtm

PR #2991, opened 2016-08-29
WT-2869 Fix a performance regression on secondaries.

michaelcahill:
Don't try to repeatedly evict pages if the eviction state hasn't moved
forwards.


michaelcahill:
I'm going to self-review and merge this one.

PR #2998, opened 2016-08-30
WT-2871 Make verbose formats and argument types match.

michaelcahill:
OS X clang in particular is fussy about size_t vs uint64_t vs uintmax_t.


agorrod:
Thanks @michaelcahill lgtm

PR #2999, opened 2016-08-30
WT-2876 Add an oplog like ability to wtperf utility.

agorrod:
In an attempt to be able to simulate secondary replica nodes.


michaelcahill:
@agorrod, this change doesn't include truncating the "log-like" table -- is that something we should consider adding?

Otherwise, lgtm.

agorrod:
> this change doesn't include truncating the "log-like" table

It would definitely be good to do that. I didn't do it now, since I wasn't prepared to add the complexity and figure out how it fits into the existing truncate functionality in wtperf. I've opened WT-2889 to track the functionality.

michaelcahill:
Fair enough, sounds good to me.  I'll merge.

PR #3000, opened 2016-08-30
WT-2873 Refactor CRC32 code

keithbostic:
Refactor CRC32 code, better support for zSeries boxes that don't support hardware CRC32.


keithbostic:
@michaelcahill, @agorrod, I think this one is ready for review.

If you agree with the approach, I'll start talking to @markbenvenuto about integrating it.

build_posix/aclocal/options.m4
```diff
@@ -47,6 +47,19 @@ AM_CONDITIONAL([HAVE_BUILTIN_EXTENSION_ZLIB],
...
+AC_ARG_ENABLE(crc32-hardware,
+	AC_HELP_STRING([--enable-crc32-hardware],
+	    [Enable CRC32 hardware support.]), r=$enableval, r=yes)
+case "$r" in
+no)	wt_cv_enable_crc32_hardware=no;;
+*)	AC_DEFINE(HAVE_CRC32_HARDWARE)
+	wt_cv_enable_crc32_hardware=yes;;
+esac
+AC_MSG_RESULT($wt_cv_enable_crc32_hardware)
+
```
michaelcahill: Why would anyone disable this?  Is it just for our testing purposes?

michaelcahill:
Thanks @keithbostic, this generally looks good.

A few minor comments:

* Can we have an inlined wrapper function called `__wt_checksum` to call through the function pointer in `__wt_process`, rather than making callers do that?
* Generally, can we stick to "checksum" (the whole word) rather than having function names not match file names?
* If we want to add a Windows implementation of the X86 hardware accelerated checksums one day would this structure have to change again?  I'm concerned that "software" vs "x86" isn't sufficient.

Regardless, please work with @markbenvenuto on the change to the MongoDB build -- we'll need that pushed into MongoDB master before this branch can be merged, or some automated tests that combine the two will break.

build_posix/aclocal/options.m4
keithbostic: Anyone building WiredTiger on the zSeries may have to use this, because the hardware-CRC32 code on the zSeries drops core with some Linux kernels (and @markbenvenuto and I don't yet see any simple way to figure out if we're on such a machine).

keithbostic:
@michaelcahill: 

> Can we have an inlined wrapper function called __wt_checksum to call through the function pointer in __wt_process, rather than making callers do that?

Done & pushed (I used `__wt_cksum`, that way most of the code doesn't change).

> Generally, can we stick to "checksum" (the whole word) rather than having function names not match file names?

I'm happy to make that pass (I thought about doing it too). If we do it everywhere (for example, change `multi->cksum = bnd->cksum` to `multi->checksum = bnd->checksum` in the reconciliation code), it's a big pass, maybe 300 or so occurrences. We could:

1. change the file names to `cksum`
2. only change function names from `cksum` to `checksum`
3. change everything from `cksum` to `checksum`

I'd be inclined to do either 1 or 3, number 2 seems inconsistent (but I won't fight about it). What's your pleasure?

> If we want to add a Windows implementation of the X86 hardware accelerated checksums one day would this structure have to change again? I'm concerned that "software" vs "x86" isn't sufficient.

We could do that by adding `checksum/x86/crc32-win-x86.c` and changing `dist/filelist`:
```
src/checksum/x86/crc32-x86.c        POSIX_HOST
src/checksum/x86/crc32-win-x86.c    WINDOWS_HOST
```
Or if there's a Windows-specific software implementation:
```
src/checksum/software/checksum.c            POSIX_HOST
src/checksum/software/checksum-win-x86.c    WINDOWS_HOST
```

markbenvenuto:
@keithbostic @michaelcahill I just want to clarify that in terms of Windows x64 hardware accelerated checksums, we already have an implementation:
https://github.com/wiredtiger/wiredtiger/blob/2ddc171b47690a42d6477cfc469f4fa62f885517/src/checksum/x86/checksum.c#L1223-L1261
https://github.com/wiredtiger/wiredtiger/blob/2ddc171b47690a42d6477cfc469f4fa62f885517/src/checksum/x86/checksum.c#L1294-L1303

It is subtle because it is not a "Windows" thing, it is conditional instead under `_M_AMD64` instead of say `WIN32` because it is a difference in compiler API. MSVC does not support inline x64 assembly so  we need to use SSE intrinsics instead.

src/checksum/software/checksum.c
```diff
@@ -1171,137 +1163,3 @@ __wt_cksum_sw(const void *chunk, size_t len)
...
-{
-	return (*__wt_cksum_func)(chunk, len);
-}
-
-/*
- * __wt_cksum_init --
- *	WiredTiger: detect CRC hardware and set the checksum function.
- */
-void
-__wt_cksum_init(void)
```
markbenvenuto: For the software CRC32 support, we need a default `__wt_cksum_init` for platforms that do not have hardware CRC32 implementation (i.e., ARM).

michaelcahill:
@keithbostic, can we please go with:

> 3. change everything from cksum to checksum

?

build_posix/aclocal/options.m4
michaelcahill: If I'm understanding correctly, we should turn `HAVE_CRC32_HARDWARE` on in `SConstruct` so we are testing the hardware CRC code on Windows.

michaelcahill:
@keithbostic, thanks -- I'll merge this, it will temporarily break our mongodb-perf-ycsb-develop build in Jenkins but we're closing in on a bump of MongoDB master that will resolve that.

PR #3001, opened 2016-08-30
WT-2875 Add environment variable to disable long tests

daveh86:


daveh86:
@sulabhM, mind reviewing?

test/utility/misc.c
```diff
@@ -139,6 +139,17 @@ testutil_cleanup(TEST_OPTS *opts)
...
 
 /*
+ * testutil_check_long --
+ *	Check if WT_DISABLE_LONG_TESTS is set and exit if so
+ */
+void
+testutil_check_long(void)
+{
+	if (getenv("WT_DISABLE_LONG_TESTS") != NULL)
+		exit(0);
```
sulabhM: Do you want to throw a warning or error message?

sulabhM:
Do you want to throw an error/warning message in case you exit if disable long is set?

Other than that LGTM

daveh86:
I had thought adding an message, but decided against it, as the code is "opt out" not "opt in", meaning you must explicitly do something in order to disable the test.

daveh86:
@agorrod, mind taking a look too?

test/utility/misc.c
```diff
@@ -139,6 +139,17 @@ testutil_cleanup(TEST_OPTS *opts)
 }
 
 /*
+ * testutil_check_long --
+ *	Check if WT_DISABLE_LONG_TESTS is set and exit if so
+ */
+void
+testutil_check_long(void)
```
agorrod: I don't love the name. "check" makes me think the function will return something.

test/utility/misc.c
agorrod: I'm with @sulabhM I think a warning would be valuable here. If it isn't possible without causing problems for "make check" in this configuration, I'd be inclined to tie an output to the ``verbose`` option that is in the ``TEST_OPTS`` structure.

Having this function ``exit`` directly rather than return whether this is a long test, doesn't give the test case a chance to either cleanup after itself or alter it's behavior rather than being skipped altogether.

test/utility/misc.c
```diff
@@ -139,6 +139,17 @@ testutil_cleanup(TEST_OPTS *opts)
...
+{
+	if (getenv("WT_DISABLE_LONG_TESTS") != NULL)
```
agorrod: Is getenv going to work on non-posix platforms? Do any of our tests work on non-posix platforms? I'd be inclined to use ``__wt_getenv`` instead.

daveh86:
@agorrod , I have made this test a boolean to let the application decide how to handle. Can you take a look?

test/utility/misc.c
```diff
@@ -139,6 +139,20 @@ testutil_cleanup(TEST_OPTS *opts)
 }
 
 /*
+ * testutil_check_stop_long --
+ *	Check if WT_DISABLE_LONG_TESTS is set and exit if so
```
agorrod: It would be better if the environment variable and function names match, since they have the same structure. How about making them ``testutil_disable_long_tests`` and ``TESTUTIL_DISABLE_LONG_TESTS``

test/csuite/wt2853_perf/main.c
```diff
@@ -86,6 +86,10 @@ main(int argc, char *argv[])
 
 	opts = &_opts;
 	sharedopts = &_sharedopts;
+
+	/* Check if the WT_DISABLE_LONG_TESTS is set and if so, exit */
```
agorrod: I don't think we need a comment here - the name of the function makes the functionality obvious.

agorrod:
lgtm

PR #3007, opened 2016-08-31
WT-2866 Move stats setting and restore setting current queue.

sueloverso:
@michaelcahill Please review this.  I've chased the failure today since seeing it on the Solaris evergreen run.  Merging your branch this morning makes `stat05` fail most of the time for me on AWS, usually within 5 runs.  Going through your changes, one by one, I found I needed two to make `stat05` pass all the time.  I had to move the setting of whether or not the queue was empty below the return near line 900.  I moved it back to its original location.  I also had to restore the setting of the queues.  Either change without the other still fails.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -851,16 +851,13 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
 	if ((ret = __evict_walk(cache->walk_session, queue)) != 0)
 		return (ret == EBUSY ? 0 : ret);
 
-	/*
-	 * If the queue we are filling is empty, pages are being requested
-	 * faster than they are being queued.
-	 */
-	if (__evict_queue_empty(queue)) {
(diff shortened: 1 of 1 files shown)


sueloverso:
@agorrod Since Michael is out, can you review this so that the tests no longer fail.  I had to revert two parts of WT-2866.  I don't really understand why those two changes broke it, but this does fix it.

agorrod:
@sueloverso I've been digging into this one today (most of the day), and I haven't gotten to the bottom of the failure yet. I don't think this addresses the problem, but rather masks it.

The issue seems to be a timing problem between evicting a page/pages and inserting into a complex table. That isn't a combination of operations that I'd expect to interfere - the only way I can think that they would interfere is that eviction can trigger WT_RESTART in the cursor code. My suspicion is that there is somewhere in the cur_table implementation that isn't handling WT_RESTART errors consistently. I haven't been able to spot it yet though.

sueloverso:
Close this branch in favor of Alex's.

PR #3008, opened 2016-08-31
WT-2879 test_stat05.test_stat_cursor_config.test_stat_cursor_size failures

keithbostic:
Ready for review/merge.


agorrod:
@keithbostic I see that this avoids the failure, but I don't believe it fixes the root cause. The problem appears to be some combination of eviction and complex table insert operations. I'm going to close this for now.

PR #3009, opened 2016-09-01
WT-2866 Don't set eviction stuck unless the cache is full.

agorrod:
Only count the queue being empty towards the calculation of whether the server should evict pages if eviction is needed.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -610,9 +610,9 @@ __evict_pass(WT_SESSION_IMPL *session)
 			 * Mark the cache as stuck if we need space and aren't
 			 * evicting any pages.
 			 */
-			if (FLD_ISSET(
-			    cache->state, WT_EVICT_STATE_CLEAN_HARD) ||
-			    FLD_ISSET(cache->state, WT_EVICT_STATE_DIRTY_HARD))
+			if (FLD_ISSET(cache->state,
+			    WT_EVICT_STATE_CLEAN_HARD |
+			    WT_EVICT_STATE_DIRTY_HARD))
 				F_SET(cache, WT_CACHE_STUCK);
 
 			WT_STAT_FAST_CONN_INCR(session, cache_eviction_slow);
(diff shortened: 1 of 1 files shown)


agorrod:
@sueloverso Please take a look at this branch and merge if it makes sense to you. If you are uncomfortable with the changes I'll be happy to talk them through with @michaelcahill tomorrow instead.

I think we'll want further tuning of both the WT_EVICT_STUCK flag and the condition that decides whether the eviction server should contribute to eviction, but this change is an improvement and should allow tests to pass in the mean time.

sueloverso:
Thanks @agorrod.  I'm going to merge this so that the tests are passing.  I think this area needs more investigation because it was certainly a surprise that the innocuous looking change broke that test.

src/evict/evict_lru.c
```diff
@@ -856,8 +864,11 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
 	 * faster than they are being queued.
 	 */
 	if (__evict_queue_empty(queue)) {
-		cache->evict_empty_score = WT_MIN(WT_EVICT_EMPTY_SCORE_MAX,
-		    cache->evict_empty_score + WT_EVICT_EMPTY_SCORE_BUMP);
+		if (__wt_eviction_needed(session, NULL))
+			cache->evict_empty_score =
+			    WT_MIN(WT_EVICT_EMPTY_SCORE_MAX,
+			    cache->evict_empty_score +
+			    WT_EVICT_EMPTY_SCORE_BUMP);
```
michaelcahill: @agorrod why do ever get here if eviction isn't needed?

agorrod: I believe we get here if the eviction server is woken for another reason. Either a checkpoint or queuing a page for urgent eviction will lead the eviction server to do this. The eviction server uses ``__wt_evict_update_work`` to decide if there is anything to do, wheras ``__wt_eviction_needed`` checks whether the cache is above a trigger value.

The naming could definitely be clearer.

PR #3011, opened 2016-09-01
WT-2555 Make Format run on Windows

markbenvenuto:
This PR adds support for running `test/format` on Windows.

1. Build format without Berkley DB,
2. Fix up various commands to run on Windows. In some cases, I needed to `ifdef` the test code.
3. Add one more pthread function to the test windows shim.
4. Random minor compilation fixes (`wt_internal.h`, `format.h`)

In terms of MCI integration, I put in an example task but I have not enabled it until there is a solution for handling crashes & hangs in Evergreen in general.


agorrod:
Thanks @markbenvenuto lgtm.

PR #3012, opened 2016-09-01
WT-2883 Remove potentially recursive call for verbose handleops

ddanderson:


michaelcahill:
@ddanderson, thanks, fbm -- I can't see any reason to add verbose output for a `printf`.  I'll merge.

PR #3014, opened 2016-09-01
WT-2867 Review and fix barrier usage in __lsm_tree_close

keithbostic:
@agorrod, I took a look at this, and the ticket is correct, the barrier should be a full-barrier, not a read-barrier.

I also cleaned up a couple of routines that can return void, not int.

EDIT: I backed out the tricky change -- on second thought, it was incorrect, and this change is now really vanilla.


agorrod:
Thanks @keithbostic lgtm.

PR #3016, opened 2016-09-02
WT-2859 Remove FAST stats macros

sulabhM:
Having _FAST_ macros gives an impression that when we use them, we are collecting fast statistics only, which is not true.
Except when statistics=none is set, we collect all the stats.
This change removes _FAST_ macros and modifies the basic macros to only collect stats when statistics=none is not set.


sulabhM:
auto build: retest this please

sulabhM:
auto build: retest this please

sulabhM:
Not ready for review - test_inmem01 python test is failing consistently when run through pull-request tester. Same test doesn't fail when run locally.

sulabhM:
@keithbostic , The objective of my change is to be consistent throughout the code in collecting statistics only when statistics is NOT set to "none". For this I removed all the FAST macros and modified the underlying macros to only update stats if stat_flags does NOT have WT_CONN_STAT_NONE set.

In the case of memory allocation early on:
1. We haven't got to process the connection configuration string yet, so stat_flags are set to 0
2. At this point the statistics structure might not have been allocated memory and hence we can't really collect stats
3. I am using WT_CONN_SERVER_RUN flag to determine if we are ready to collect the stats. Before this change the FAST macros used to check for WT_CONN_STAT_FAST to be set to collect stats. Since, I am using WT_CONN_STAT_NONE to be not set, I need another way to determine that we have processed connection configuration.

I can think of other ways to make this check:
1. Can check for "stat_flags != 0 && !FLD_ISSET(..,WT_CONN_STAT_NONE)"
2. Can change enum so that WT_CONN_STAT_NONE is 0

sulabhM:
auto build: retest this please

sulabhM:
@agorrod , while I debug python test test_inmem01, I think this change is ready for review.

bench/wtperf/runners/checkpoint-stress.wtperf
```diff
@@ -1,23 +1,26 @@
 # A stress configuration to create long running checkpoints while doing a lot
```
agorrod: The change to this file is a tidy up related to WT-2868, it's not particularly relevant for this change.

agorrod:
@keithbostic Would you mind reviewing this branch please?

src/conn/conn_api.c
```diff
@@ -1677,14 +1677,21 @@ __conn_statistics_config(WT_SESSION_IMPL *session, const char *cfg[])
...
 		LF_SET(WT_CONN_STAT_CLEAR);
+	}
 	WT_RET_NOTFOUND_OK(ret);
 
 	if (set > 1)
 		WT_RET_MSG(session, EINVAL,
 		    "only one statistics configuration value may be specified");
 
+	WT_ASSERT(session, LF_ISSET(WT_CONN_STAT_NONE | WT_CONN_STAT_FAST));
+
```
keithbostic: @sulabhM, I'm not sure what this assertion is guarding against; maybe a comment would be worthwhile?

keithbostic:
@sulabhM, I don't like testing `WT_CONN_SERVER_RUN` in OS-specific code (and I think it's a bad flag to overload with a "should we track statistics" semantic).

I don't care much about statistics consistency (that is, I don't care if we accumulate statistics until statistics are configured and then quit based on the configuration); that can happen at any point during the run, I can't see why it matters.

I agree with you we don't want to have to deal with when/if the statistics structures are allocated.

Of the ideas you listed, I rather like your idea of making `WT_CONN_STAT_NONE` == 0. We could get rid of `WT_CONN_STAT_NONE` and add a new macro, using either non-zero or the all/fast flags as the test:
```
#define WT_STATISTICS_CONFIGURED(session)    (S2C(session)->stat_flags != 0)

#define WT_STATISTICS_CONFIGURED(session)
        FLD_ISSET(S2C(session)->stat_flags, WT_CONN_STAT_ALL | WT_CONN_STAT_FAST)
```
then use that macro where we're currently testing for `WT_CONN_STAT_NONE`:
```
#define WT_STAT_INCRV(session, stats, fld, value) do {
        if (!WT_STATISTICS_CONFIGURED(session))
                (stats)[WT_STATS_SLOT_ID(session)]->fld += (int64_t)(value);
```

keithbostic:
@sulabhM, this change as written lgtm, with a couple of minor comments.

There is one general question I'd like to ask. (I wasn't in on the original discussions of this change, nor have I thought nearly as hard about this change as you have, so please forgive me if I'm raising topics you have already resolved or just missing the obvious.)

IIRC, the reason "fast" statistics existed was because there was an idea we might some day add more statistics categories, and then applications might optionally **not** track "fast" statistics. For example, imagine some new "evict" statistics which are sufficiently expensive to gather they aren't included in "fast". Applications could then configure `statistics=(fast,evict)` (if they wanted the eviction statistics in addition to the "fast" statistics), or `statistics=(evict)` if they wanted the eviction statistics alone. And there'd be a new macro family `WT_STAT_EVICT_CONN_XXX` that only maintained a statistic if the `WT_CONN_STAT_EVICT` flag was set.

I think this change moves to a model where "fast" statistics are always collected, they're no longer optional for an application.

I've been staring at the Reference Guide wording and I think the existing arguments to `wiredtiger_open` and the open of a statistics cursor ("all", "fast" and "none), continue to make sense. (Although "fast" is really the "default" statistics now, not a separate group of statistics, so "on" might make more sense -- regardless, I doubt it's worth changing.) If we ever add new groups of statistics, we should document "fast" is configured whenever either "all" or the new group is specified, of course.

That said, I'm unsure of the original problem description:
> Having FAST macros gives an impression that when we use them, we are collecting fast statistics only, which is not true.
Except when statistics=none is set, we collect all the stats.
This change removes FAST macros and modifies the basic macros to only collect stats when statistics=none is not set.

IIRC, the intent was "all" be a superset of "fast" (and, "all" be a superset of any future groups of statistics).

It looks to me like the `WT_FAST_XXX` macros as written would isolate the "fast" statistics from other groups of statistics, and doing a quick scan, I don't see any places where the wrong macro was being used. Were there problems in how those macros were being used?

Was there consensus that applications would never wish to turn off "fast" statistics?

sulabhM:
@keithbostic , while working on writing a script to create documentation categorising statistics between "fast" and "all", I found the usage of FAST macros misleading. For both "all" and "fast" we set the WT_CONN_STAT_FAST flag, and use that to check when we are updating statistics. Effectively we update almost all the statistics irrespective of whether fast or all is set, and we update a remaining separate set of statistics only if WT_CONN_STAT_ALL is set while we walk btree in a stats cursor context to obtain these expensive stats. Also, as of today we do not have any other category of statistics, all being a superset for all the statistics. This did not come obvious to me when I started looking at the code.

> It looks to me like the WT_FAST_XXX macros as written would isolate the "fast" statistics from other groups of statistics, and doing a quick scan, I don't see any places where the wrong macro was being used. Were there problems in how those macros were being used?

There isn't a functional problem with the way these macros have been used. I haven't really made any change to the logic of how we collect/update/report stats as before. The only change is to get away from naming macros as FAST, because that is kind of misleading. When we are collecting stats we have always have FAST flag set, so it makes more sense to look for the NONE flag not being set and renaming these statistics accordingly.

>Was there consensus that applications would never wish to turn off "fast" statistics?

The idea is to make it obvious that as of today we can't really turn off "fast" statistics. When enabled these are the only statistics we have (apart from a small set of expensive stats that we walk the btree to collect as and when needed)

> I think this change moves to a model where "fast" statistics are always collected, they're no longer optional for an application.

That is true and reflects the fact that today if you are collecting statistics you can't avoid collecting the "fast" statistics.
We can always move to a model where we collect different set of statistics with appropriately named macros, but as of today since we can't avoid collecting "fast" statistics, I think this change makes it more obvious and avoids that confusion.

src/lsm/lsm_stat.c
```diff
@@ -39,7 +39,7 @@ __curstat_lsm_init(
 	WT_ERR(__wt_scr_alloc(session, 0, &uribuf));
 
 	/* Propagate all, fast and/or clear to the cursors we open. */
-	if (!F_ISSET(cst, WT_CONN_STAT_NONE)) {
+	if (cst != 0) {
```
keithbostic: `cst` is a pointer, so this should be a comparison against `NULL`

keithbostic: It looks to me like `cst` can't be NULL here, so adding this test is almost certainly going to trigger a Coverity complaint, Coverity doesn't like it if you check a pointer against NULL, but don't check it against NULL everywhere.

keithbostic:
A couple of minor comments, otherwise, lgtm.

sueloverso:
>>  Was there consensus that applications would never wish to turn off "fast" statistics?

> The idea is to make it obvious that as of today we can't really turn off "fast" statistics. When enabled these are the only statistics we have (apart from a small set of expensive stats that we walk the btree to collect as and when needed)

>>  I think this change moves to a model where "fast" statistics are always collected, they're no longer optional for an application.

> That is true and reflects the fact that today if you are collecting statistics you can't avoid collecting the "fast" statistics.  We can always move to a model where we collect different set of statistics with appropriately named macros, but as of today since we can't avoid collecting "fast" statistics, I think this change makes it more obvious and avoids that confusion.

I think @keithbostic brings up some very valid points that leads me to think this change is gratuitous and has downside.  As pointed out the downside is that fast stats will now always be on for any stats turned on and this change removes flexibility that currently exists even though today that flexibility is under-utilized.  What is the upside?

His example of adding an `evict` category is timely and prescient.  I foresee in the very near future that we will add stats to distinguish the couple dozen conditions where we reject a page, at a minimum.  There has been frequent talk to do so recently, particularly debugging eviction-related issues.  But stats are not free and so it is possible, if not very likely, that if those stats are added, there'd be a new `evict` category that would now be off by default, but turned on by Bruce or us for deeper debugging.

Yes, mongoDB always has `fast` stats turned on.  That is unlikely to ever change.  But it is easy to believe that we might also do similar eviction debugging for other WT-only customers that do not currently run stats.  So if `stats=(evict)` exists soon, then immediately we'd be avoiding collecting `fast` stats in that particular case.  I believe the (original) code is correct in that manner and adding such a category is likely to mean all/many of these proposed changes have to be undone again.  Today's code is also structured such that adding a new category is clean and easy right now.

So, even though today `fast` is the only sub-category of `all`, that doesn't sufficiently motivate assuming it will always be the case.  Removing the ability of adding other distinct sub-categories is a loss, IMO.  I think a user turning on `stats=(evict)` would be surprised to get a huge number of completely unrelated stats in addition to getting eviction-specific stats.

sueloverso:
TBH, I had forgotten about potential new statistics categories until reading Keith's message.  But I immediately recognized it as a clear path forward in the area I've been discussing with @michaelcahill the last week or two.  There was a dilemma between really wanting stats on each eviction conditional to know why we skip pages for eviction, but stats on paths like that can be expensive so we have been hesitant to add them.

So while I do not think this branch should be merged, this has been very beneficial to pointing out an ideal path forward in a related area.

src/lsm/lsm_stat.c
sulabhM: I wanted to check if the flags are 0, and copy them only if the value for flags is not 0.
I have fixed that now.

sulabhM:
I have started the general discussion about this change in the jira ticket: https://jira.mongodb.org/browse/WT-2859

sulabhM:
@agorrod : Please review.

src/os_common/os_alloc.c
```diff
@@ -45,8 +45,9 @@ __wt_calloc(WT_SESSION_IMPL *session, size_t number, size_t size, void *retp)
 	 */
 	WT_ASSERT(session, number != 0 && size != 0);
 
-	if (session != NULL)
-		WT_STAT_FAST_CONN_INCR(session, memory_allocation);
+	/* Update statistics only if a connection has been established. */
```
agorrod: I think it's worth expanding on this comment to make it clear why it is important. Maybe "Only update statistics once we can be sure the connection and hence statistics tracking structures have been fully initialized."

agorrod:
@sulabhM I have one nit about comment wording (that is duplicated a couple of times). Apart from that I like this branch.

agorrod:
Thanks @sulabhM lgtm.

PR #3018, opened 2016-09-02
WT-2886 Ignore dirty eviction settings when in_memory is true.

michaelcahill:


michaelcahill:
@sueloverso can you please review this one as well?

I made the documentation in `api_data.py` omit periods at the end of strings -- feel free to omit that part of the change if you prefer and I'll open a separate ticket.

agorrod:
Thanks @michaelcahill lgtm - should I go ahead and merge?

PR #3019, opened 2016-09-02
WT-2888 void functions incorrectly typed as returning int

keithbostic:
It seems like we've been finding/fixing functions that return an int, but don't really have a return value, one-by-one, for the last couple of weeks, so I went looking this morning.

Ready for review/merge.


agorrod:
Thanks @keithbostic lgtm. I'll merge.

PR #3021, opened 2016-09-05
WT-2876 Fix a memory leak of new wtperf URI

agorrod:


agorrod:
I'm self-reviewing this one. It's a simple memory leak fix.

PR #3022, opened 2016-09-05
WT-2233 Make aggressive / stuck a sliding scale rather than booleans.

michaelcahill:


michaelcahill:
autobuild: retest this please

src/evict/evict_lru.c
```diff
@@ -1144,25 +1162,31 @@ retry:	while (slot < max_entries && ret == 0) {
...
 	 * candidates and we aren't finding more.
 	 */
-	if (cache->pass_intr == 0 && ret == 0 &&
-	    slot < max_entries && (retries < 2 ||
+	if (slot < max_entries && (retries < 2 ||
 	    (retries < 10 &&
 	    (slot == queue->evict_entries || slot > start_slot)))) {
 		start_slot = slot;
 		++retries;
 		goto retry;
```
agorrod: What is the benefit of holding the dhandle lock across retries?

src/evict/evict_lru.c
```diff
@@ -574,52 +558,70 @@ __evict_pass(WT_SESSION_IMPL *session)
 		 * workers, it must service the urgent queue in case all
 		 * application threads are busy.
 		 */
-		if (cache->evict_empty_score < WT_EVICT_EMPTY_SCORE_CUTOFF ||
+		if (cache->evict_empty_score < WT_EVICT_SCORE_CUTOFF ||
 		    (!WT_EVICT_HAS_WORKERS(session) &&
 		    !__evict_queue_empty(cache->evict_urgent_queue)))
 			WT_RET(__evict_lru_pages(session, true));
 
+		if (cache->pass_intr != 0)
```
agorrod: Your change doesn't really change how pass_intr is being used, I am suspicious that it could be disruptive to the eviction server when handles are closed after a checkpoint has finished, since I suspect that checkpoint handles will be discarded by sweep via ``__wt_conn_dhandle_discard_single``, which will interrupt what the eviction server is doing.

src/include/cache.i
```diff
@@ -68,6 +79,21 @@ __wt_cache_read_gen_new(WT_SESSION_IMPL *session, WT_PAGE *page)
...
+ * __wt_cache_stuck --
+ *      Indicate if the cache is stuck (i.e., not making progress).
+ */
+static inline bool
+__wt_cache_stuck(WT_SESSION_IMPL *session)
+{
+	WT_CACHE *cache;
+
+	cache = S2C(session)->cache;
+	return (cache->evict_aggressive_score == WT_EVICT_SCORE_MAX &&
```
agorrod: I generally like the new aggressive score scheme. I'm a little bit nervous that it is prone to off-by-ones and potentially wrapping. The usages look correct at the moment, but I'm inclined to abstract out all updates to ``cache->evict_aggressive_score`` eventually to make it less likely that we will introduce bugs in the future.

agorrod:
@michaelcahill I had a couple of questions, but other than that the changes in here make sense to me.

src/evict/evict_lru.c
michaelcahill: I don't think there is any performance impact: this should be the rare case where the eviction server gets to the end of the handle list without finding a handle it wants to visit, then decides to restart.  I made the change so I could use `WT_ERR` rather than `break` for getting out of the loop: tracing through the `ret == 0` checks made the code brittle.

src/evict/evict_lru.c
michaelcahill: I agree that sweep could cause interruptions in that case...

We use `pass_intr` for a couple of different things, but I think it's mostly to encourage the eviction server to release the handle list lock.

Should we discuss alternatives to this change?

src/include/cache.i
michaelcahill: That sounds fine to me: this code has evolved while this branch has been open.

One question for you: the intent of the timing was that eviction would get aggressive after 100ms of not making progress and stuck after 1s.  Leaving the implementation details aside, do you think those times are reasonable?

src/evict/evict_lru.c
agorrod: I'm not inclined to chase this now - I think it's more likely to be relevant in different workloads. I mentioned it because I knew I'd completely forget otherwise.

src/include/cache.i
agorrod: I think it's short for both of them - especially for stuck, since it is the trigger for adding LAS entries in eviction. You have more experience of tuning than I do, so I didn't argue.

michaelcahill: How about 200ms / 2s?

There is a balancing act -- we don't want to start forcing transactions to roll back unnecessarily but if the lookaside table is really required to make progress with some workload, it would be nice to figure that out without waiting too long.

agorrod: > How about 200ms / 2s?

Works for me.

michaelcahill: Thanks, done.

agorrod:
Thanks @michaelcahill lgtm, I'll merge.

PR #3023, opened 2016-09-05
WT-2892 hot backup can race with block truncate

keithbostic:


keithbostic:
@michaelcahill, there are additional unprotected accesses to the `hot_backup` flag:

* https://github.com/wiredtiger/wiredtiger/blob/develop/src/log/log.c#L837-L843
* https://github.com/wiredtiger/wiredtiger/blob/develop/src/lsm/lsm_work_unit.c#L589-L598
* https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn_log.c#L369-L380

I think they may be worth review, I doubt `log/log.c` and `lsm/lsm_work_unit.c`, I'm less sure of `txn/txn_log.c`.

keithbostic:
Here's the workload I've been using to test:
```cache=200
backups=1
checkpoints=1
leak_memory=1
logging_prealloc=1
logging=1
checksum=uncompressed
compression=lz4
#in_memory=1
data_source=table
file_type=row
timer=5
ops=0
rows=10000
salvage=0
```

keithbostic:
I've lost the failure I saw when the block-manager raced with hot-backup; the failure I saw with the logging files was generally something like this:
```
t: /media/ephemeral1//RUNDIR.18004/WiredTigerLog.0000000002: handle-read: pread: failed
to read 90496 bytes at offset 104636032: WT_ERROR: non-specific WiredTiger error
```

keithbostic:
@sueloverso, for your review/consideration, I think it's ready to merge if you're happy with it.

src/log/log.c
```diff
@@ -834,30 +832,39 @@ __log_newfile(WT_SESSION_IMPL *session, bool conn_open, bool *created)
...
-		/*
-		 * If we get any error other than WT_NOTFOUND, return it.
-		 */
-		WT_RET_NOTFOUND_OK(ret);
+		__wt_readlock(session, conn->hot_backup_lock);
+		if (conn->hot_backup)
+			__wt_readunlock(session, conn->hot_backup_lock);
+		else {
+			ret = __log_alloc_prealloc(session, log->fileid);
+			__wt_readunlock(session, conn->hot_backup_lock);
```
sueloverso: Why not just:
```
read lock
if (!conn->hot_backup)
    log_alloc_prealloc...
readunlock
```

src/log/log.c
```diff
@@ -968,11 +975,17 @@ __log_truncate_file(WT_SESSION_IMPL *session, WT_FH *log_fh, wt_off_t offset)
...
+		__wt_readlock(session, conn->hot_backup_lock);
+		if (conn->hot_backup)
+			__wt_readunlock(session, conn->hot_backup_lock);
+		else {
+			ret = __wt_ftruncate(session, log_fh, offset);
+			__wt_readunlock(session, conn->hot_backup_lock);
+			if (ret != ENOTSUP)
+				return (ret);
+			F_SET(log, WT_LOG_TRUNCATE_NOTSUP);
+		}
```
sueloverso: I see why you did that here though.  I suspect it was just copied and lines deleted earlier!

sueloverso:
@keithbostic LGTM with one minor rearrangement.  Your changes do not modify the LSM or txn related uses.  Have you decided those are not needed?  I need to think more deeply about them.

sueloverso:
I have one more general comment.  It would be nice to add a `backup.sh` that gets run, similar to `recover.sh` so that we continue to test backup-related races specifically.  Also I'd like to see a Jenkins job like `wiredtiger-test-format-backup` that would be similar to `wiredtiger-test-format-recovery` using the script.

src/log/log.c
keithbostic: I tried it that way, and handling the error conditions got messy (for example, avoiding signaling the server when `hot_backup` was set and we'd never called `__log_alloc_prealloc`, I ended up needing more variables.

If you prefer that form, I have no objection!

keithbostic:
Thank you, @sueloverso.

AFAIK, @michaelcahill believes LSM & txn are OK.

I'd like to merge this one as soon as you're ready.

src/log/log.c
sueloverso: My mistake, I completely "unindented" the rest of the 'else' statement when I read through this.  All is well.

sueloverso:
@keithbostic I think the changes to the default settings for backup and logging will be sufficient to get enough coverage for this.  No need for a backup.sh and new job.  I'll merge when PR tests complete.

PR #3029, opened 2016-09-09
WT-2901 Make checkpoint scrubbing configurable.

michaelcahill:
Also avoid eviction of dirty pages entirely in the eviction server if
worker threads are configured.


agorrod:
Thanks @michaelcahill lgtm. I'll merge and kick off a set of Jenkins perf tests.

PR #3030, opened 2016-09-09
WT-2902 Keep populating the eviction queue if it is being emptied.

agorrod:
Not yet ready for merge - this is just a starting point.


agorrod:
This has been replaced by a different solution.

PR #3032, opened 2016-09-12
WT-2903 Change the eviction_scrub_target default to 15%

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'm happy for this to be merged once it passes testing.

For the record, this change improves overall throughput in some random update workloads with data sets larger than dirty-cache. The trade off is more instability in throughput during checkpoints.

michaelcahill:
Thanks @agorrod, merging.

PR #3033, opened 2016-09-13
WT-2908 Add a dry-run option to python suite

daveh86:


daveh86:
@ddanderson , would you be able to have a look at this? It provides a way to iterate through all the tests/scenarios in a manner that we can use to parallelise on Windows.

ddanderson:
@daveh86, lgtm.  I'll merge.

PR #3035, opened 2016-09-13
WT-2908 Change output of shortDesc to use simpleName

daveh86:


daveh86:
@agorrod, can you look over this? 

It's a change of variable names to correctly reference the tests originating file.

agorrod:
lgtm

PR #3036, opened 2016-09-13
WT-2902 Track per-tree the efficiency of eviction walks.

michaelcahill:
Some workloads create "deserts" within a tree where no good eviction
candidates can be found (e.g., zipfian distributions, append-only
workloads, etc.).  Rather than walking all the way through such
regions, give up and try in the opposite direction next time.


agorrod:
Thanks @michaelcahill lgtm.

PR #3039, opened 2016-09-14
WT-2898 evict dynamic workers

fedorova:
This branch dynamically tunes the number of eviction workers in the range specified by threads_max and threads_min eviction variables. It is recommended to set threads_min to 1 and threads_max to a larger number (e.g., 8). The eviction server should pick the best number of eviction workers automatically.


src/evict/evict_lru.c
```diff
@@ -808,6 +800,147 @@ __wt_evict_file_exclusive_off(WT_SESSION_IMPL *session)
...
+	pgs_evicted_cur = cache->pages_evict;
+
+	/*
+	 * If we have recorded the number of pages evicted at the end of
+	 * the previous measurement interval, we can compute the eviction
+	 * rate in evicted pages per second achieved during the current
+	 * measurement interval.
+	 * Otherwise, we just record the number of evicted pages and return.
+	 */
+	if (pgs_evicted_prev == 0)
```
sulabhM: Looks like this will always be true, since pgs_evicted_prev is initialised to 0

fedorova:
@agorrod, thanks for the fixes!

sueloverso:
@fedorova I took a quick look.  It would be good to add some (perhaps temporary) stats so that we can see how things change when this runs in the wild.

src/support/thread_group.c
```diff
@@ -334,3 +335,30 @@ __wt_thread_group_start_one(
...
+ * __wt_thread_group_stop_one --
+ *	Stop one thread if possible.
+ */
+int
+__wt_thread_group_stop_one(
+    WT_SESSION_IMPL *session, WT_THREAD_GROUP *group, bool wait)
+{
+	WT_DECL_RET;
+
+	if (group->current_threads >= group->max)
```
fedorova: Yes, indeed. Good catch! I will push a fix.

fedorova:
This one is ready for review and testing. Thank you, to everyone who contributed. Don't forget that you have to contribute eviction threads_max to a reasonably large number and eviction threads_min to one.

michaelcahill:
I've merged the tip of `develop` into here and switched performance testing to this branch to see whether our automated tests pick up any improvements.

agorrod:
I just reviewed the results from the Jenkins performance jobs. The results looked stable apart from the jobs that measure latencies, which showed max latency increasing from ~750ms to 3500ms. (http://build.wiredtiger.com:8080/job/wiredtiger-perf-stress/plot/). I've switched the jobs back to run on develop for now.

sueloverso:
autobuild: retest this please

fedorova:
Please take a look at my comment on the ticket: 

https://jira.mongodb.org/browse/WT-2898?focusedCommentId=1401530&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1401530

In a nutshell: I figured out why we are seeing higher latencies with the dynamic-workers branch. This branch is trying to avoid interference in the eviction so it is creating the eviction threads less aggressively. As a result, application threads are doing most of the eviction and are seeing high latencies. At the same time, dynamic-workers branch is giving us MUCH HIGHER THROUGHPUT, because there is less interference. So we are gaining throughput but are sacrificing tail latencies. 

My hypothesis is that we could have both good throughput and short latencies if we made application threads less eager about helping with eviction, letting the eviction server do the work instead. I'll try a few things and let you know how it goes.

fedorova:
I tried to experiment with a branch where application threads are not helping to evict and all eviction is done by 20 eviction workers. That did not make a difference: in fact I observed slightly higher maximum latencies and lower throughput. Perhaps that workload needed more eviction workers. I will continue with further experiments.

src/support/thread_group.c
```diff
@@ -334,3 +335,30 @@ __wt_thread_group_start_one(
...
+int
+__wt_thread_group_stop_one(
+    WT_SESSION_IMPL *session, WT_THREAD_GROUP *group, bool wait)
+{
+	WT_DECL_RET;
+
+	if (group->current_threads <= group->min)
+		return (0);
+
+	if (wait)
```
sulabhM: I spent some time looking at places we create and destroy threads.
Apart from the usual create/destroy the thread group and reconfigure, it looks like __evict_pass() is the only caller that could have modified the number of threads.
Since with this change __evict_pass() can only create/delete threads in the context of thread[0], do we have to worry about thread group being modified with parallel threads. If not we can do away with the group->lock.

src/evict/evict_lru.c
```diff
@@ -843,6 +835,164 @@ __wt_evict_file_exclusive_off(WT_SESSION_IMPL *session)
...
+	 * If we previously removed threads and so saw no degradation, we will
+	 * keep removing.
+	 * If we previously removed threads and saw degradation, we will add.
+	 */
+	if (conn->evict_tune_last_action == EVICT_ADDED)
+		if (pct_diff > EVICT_TUNE_THRESHOLD)
+			try_create_threads = true;
+		else
+			try_remove_threads = true;
+	else if (conn->evict_tune_last_action == EVICT_REMOVED)
```
sulabhM: The logic here leads to either always creating or removing a thread. Does that mean we will either reach the min or max configured threads for the system, or keep removing and adding threads around an optimal thread count for the workload.

src/evict/evict_lru.c
```diff
@@ -843,6 +835,164 @@ __wt_evict_file_exclusive_off(WT_SESSION_IMPL *session)
...
+		cur_threads = conn->evict_threads.current_threads;
+		target_threads = (cur_threads - conn->evict_threads.min) / 2;
+
+		for (i = 0; i < target_threads; i++) {
+			WT_ERR(__wt_thread_group_stop_one(
+				       session, &conn->evict_threads, false));
+
+			if (conn->evict_threads.current_threads == cur_threads)
+				break;
+
```
sulabhM: missed out doing "cur_threads--;" ?

src/evict/evict_lru.c
```diff
@@ -843,6 +835,164 @@ __wt_evict_file_exclusive_off(WT_SESSION_IMPL *session)
...
+			conn->evict_tune_last_action = EVICT_REMOVED;
+			WT_STAT_CONN_INCR(session,
+			    cache_eviction_worker_removed);
+			WT_STAT_CONN_SET(session, cache_eviction_active_workers,
```
sulabhM: Since a thread is always created or removed, maybe we can take this stat update and the one in the if loop outside the loops and just update once.

dist/stat_data.py
```diff
@@ -213,6 +213,9 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_eviction_walks_active', 'files with active eviction walks', 'no_clear,no_scale'),
     CacheStat('cache_eviction_walks_started', 'files with new eviction walks started'),
     CacheStat('cache_eviction_worker_evicting', 'eviction worker thread evicting pages'),
+    CacheStat('cache_eviction_worker_created', 'eviction worker thread created'),
+    CacheStat('cache_eviction_worker_removed', 'eviction worker thread removed'),
+    CacheStat('cache_eviction_active_workers', 'number of active eviction workers'),
```
fedorova: @sulabhM, sure. I can do that.

src/evict/evict_lru.c
fedorova: @sulabhM: yes, we will keep adding/removing threads to avoid getting stuck in a local minima and to adapt to the dynamically changing workload.

src/evict/evict_lru.c
fedorova: @sulabhM. Yep, good catch!

fedorova:
@sueloverso, I found and fixed a few bugs in the eviction worker tuning logic. Could you please retest on your instance again, as I did not manage to get myself access on a similar instance yet? Would be great to see what the throughput and latencies look like. I don't expect to see huge changes, but you never know. I will investigate the causes for long tail latencies in the meantime.

dist/api_data.py
```diff
@@ -396,17 +396,17 @@ def __cmp__(self, other):
     Config('eviction', '', r'''
         eviction configuration options''',
         type='category', subconfig=[
-            Config('threads_max', '1', r'''
+            Config('threads_max', '30', r'''
```
agorrod: This part of the change makes me nervous. I really like the benefit, but setting the default limit to 30 makes me uncomfortable. Would it be reasonable to set a default of 6 or 8?

src/include/connection.h
```diff
@@ -107,7 +107,7 @@ struct __wt_named_extractor {
  * Allocate some additional slots for internal sessions so the user cannot
  * configure too few sessions for us to run.
  */
-#define	WT_EXTRA_INTERNAL_SESSIONS	10
+#define	WT_EXTRA_INTERNAL_SESSIONS	40
```
agorrod: Some of our users are sensitive to the number of session handles WiredTiger uses. In part because their application can run many instances of WiredTiger on a single machine, so increasing this count significantly concerns them because the system may end up overwhelmed by threads. This number also plays into the amount of memory allocated per connection and per transaction.

Having said that, it's a safety net re: internal threads, though I think it's another vote for having a lower default maximum eviction thread count.

fedorova: @agorrod, setting the number to 8 would be reasonable.

fedorova:
@agorrod, I have changed the new defaults to eight and pushed the changes. I will re-do performance tests with that change and follow up.

fedorova:
@agorrod , I re-did performance tests. All seems in order.

src/evict/evict_lru.c
```diff
@@ -1421,7 +1625,6 @@ __evict_walk_file(WT_SESSION_IMPL *session, WT_EVICT_QUEUE *queue,
 
 		page = ref->page;
 		modified = __wt_page_is_modified(page);
-		page->evict_pass_gen = cache->evict_pass_gen;
```
sueloverso: @fedorova This seems unrelated to your change.  This line was added fairly recently (Oct 21) and I think it likely that it was lost in a merge issue.  I'm going to restore it.  Speak up if you removed it intentionally at some point.

src/evict/evict_lru.c
```diff
@@ -1570,7 +1773,7 @@ __evict_get_ref(
 	server_only = is_server && !WT_EVICT_HAS_WORKERS(session);
 	urgent_ok = (!is_app && !is_server) ||
 	    !WT_EVICT_HAS_WORKERS(session) ||
-	    (is_app && __wt_cache_aggressive(session));
+	    __wt_cache_aggressive(session);
```
sueloverso: Same with this removal/change.  This line was changed Oct23 and seems unrelated to your work.

PR #3042, opened 2016-09-15
WT-2898 Sasha's evict-dynamic-worker changes in WT coding style

sulabhM:


src/include/connection.h
```diff
@@ -309,8 +309,17 @@ struct __wt_connection_impl {
 	bool		 evict_server_running;/* Eviction server operating */
 
 	WT_THREAD_GROUP  evict_threads;
-	uint32_t	 evict_threads_max;/* Max eviction threads */
-	uint32_t	 evict_threads_min;/* Min eviction threads */
+	uint32_t         evict_threads_max;/* Max eviction threads */
```
agorrod: Did you consider adding these fields to the ``WT_CACHE`` structure instead of ``WT_CONNECTION_IMPL``?

src/include/connection.h
```diff
@@ -309,8 +309,17 @@ struct __wt_connection_impl {
...
+	uint32_t         evict_threads_min;/* Min eviction threads */
+
+	struct timespec  evict_tune_last_time;/* Last evict thread check */
+	struct timespec	 evict_time_last_create;/* Last evict thread create */
+	uint64_t	 evict_tune_pgs_last;/* Number of pages evicted */
```
agorrod: Why is this indented differently to the other fields?

sulabhM:
@fedorova , David and I made some changes to help you with the WT coding style. Please see if this helps.

Instead of using static variables, we have put the variables that need their value persist in the connection structure along with other eviction variables (and renamed them eviction_tune_*). To do that, we had to restructure code a bit, so please review if this change doesn't break the logic you intend to checkin.

src/include/connection.h
sulabhM: We initially did, then thought these fields tune evict_threads group, and would go well with thread related fields.

src/include/connection.h
sulabhM: Looks like an tab/space issue, I have pushed a fix to it.

fedorova:
Thanks, Sulabh and David! The code is much cleaner now. I fixed the pct_change to be signed. It has to be, because it can be negative if we evicted fewer pages in the current interval than in the previous one. Looks good otherwise. I'll add stats and verbose messages before considering this done.

PR #3043, opened 2016-09-15
WT-2913 Re-sort eviction queues if we find no new candidates.

michaelcahill:
We could be in a situation with a tiny number of pages in cache, all of
which are already queued.  The queue needs to be re-sorted so that some
of those entries can become candidates.  This caused hangs in some
test/format configs.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -876,27 +876,25 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
 		return (0);
 
 	/* Get some more pages to consider for eviction. */
-	switch (ret = __evict_walk(cache->walk_session, queue)) {
-	case 0:
-		break;
-	case EBUSY:
-		return (0);
-	case WT_NOTFOUND:
-		/*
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @michaelcahill lgtm assuming it passes tests.

src/evict/evict_lru.c
```diff
@@ -1355,7 +1353,7 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		 * no good eviction candidates can be found.  Abandon the walk
 		 * if we get into that situation.
 		 */
-		give_up = pages_seen > 100 &&
+		give_up = !__wt_cache_aggressive(session) && pages_seen > 100 &&
```
michaelcahill: @agorrod I didn't call this out but we discussed it.  I can move this part into a separate branch if you like (e.g., for WT-2891)?

agorrod: I did see that - it seems reasonable to me. I don't think we need to split it out.

PR #3044, opened 2016-09-15
WT-2910 When running in-memory, only evict dirty pages.

michaelcahill:
Also, don't relax the clean / dirty requirements when we get aggressive:
if application threads are blocked by the dirty limit, evicting a clean
page doesn't unblock them so there's no point trying.

### DIFF
--- a/src/conn/conn_cache.c
+++ b/src/conn/conn_cache.c
@@ -56,6 +56,13 @@ __cache_config_local(WT_SESSION_IMPL *session, bool shared, const char *cfg[])
 		cache->eviction_dirty_target = (u_int)cval.val;
 
 		/*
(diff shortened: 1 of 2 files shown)


michaelcahill:
@daveh86, this branch is a version of the changes we talked about today.  I think they're necessary (but may not be sufficient) even if the hang isn't reproducing on the tip of develop.  Let's discuss in person tomorrow.

michaelcahill:
autobuild: please retest

michaelcahill:
autobuild: retest this please

daveh86:
This LGTM

michaelcahill:
Thanks @daveh86, merging.

PR #3046, opened 2016-09-15
WT-2121 Use instruction-parallel crc32c implementation

keithbostic:
@agorrod, @michaelcahill, finally figured out how to get libtool to build from `.asm` files, so I'm pushing a branch that integrates Intel's CRC32 implementation, using pclmulqdq instructions.

I'll add some discussion to the WT-2121 ticket, but in summary, I'm seeing between .9% and 1.3% improvement over our existing code, so I'm not feeling much excitement.

It's not going to build on Windows (at least), and we're not ready to merge, so as far as I'm concerned, we should just close this one, keeping the branch around for if and when.

### DIFF
--- a/build_posix/Make.base
+++ b/build_posix/Make.base
@@ -78,6 +78,19 @@ if HAVE_BUILTIN_EXTENSION_ZLIB
 libwiredtiger_la_LIBADD += ext/compressors/zlib/libwiredtiger_zlib.la
 endif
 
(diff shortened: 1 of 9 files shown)


michaelcahill:
@keithbostic I agree with your original assessment, let's close this for now and come back to it if we see evidence that it's worthwhile.

PR #3047, opened 2016-09-16
WT-2916 Fix and simplify s_whitespace.

michaelcahill:
Use the same rules for all source files: strip trailing whitespace,
collapse multiple blank lines and strip trailing empty lines.


agorrod:
lgtm

PR #3048, opened 2016-09-16
WT-2918 the dist scripts create C files s_whitespace complains about

keithbostic:
I'm seeing complaints when I run the dist scripts about `.c` files in dist. The `s_export` script creates the file `__a.c` and shouldn't, the `s_tags` script can avoid creating a `.c` file.

Random minor cleanups along the way.

Ready for review/merge.

### DIFF
--- a/dist/s_tags
+++ b/dist/s_tags
@@ -1,10 +1,16 @@
 #! /bin/sh
 
 # Build tags files.
-trap 'rm -f tags; exit 0' 0 1 2 3 13 15
+trap 'rm -f tags' 0 1 2 3 13 15
 
 # Skip this when building release packages
 test -n "$WT_RELEASE_BUILD" && exit 0
+
+# We have to be in the dist directory to run.
+test -f s_tags || {
+	echo "s_tags requires dist be the current working directory"
+	exit 1
+}
 
 # We require ctags which may not be installed.
 type ctags > /dev/null 2>&1 || {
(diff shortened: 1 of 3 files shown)


dist/s_tags
```diff
@@ -1,41 +1,37 @@
 #! /bin/sh
 
-# Build tags file.
-#
-t=__a.c
-trap 'rm -f $t; exit 0' 0 1 2 3 13 15
+# Build tags files.
+trap 'rm -f tags; exit 0' 0 1 2 3 13 15
```
ddanderson: @keithbostic, If this script is run while current directory is build_posix or src/include, the resulting tags file will be removed.  Probably unlikely anyone would do that.  Maybe use `ctags -f tmptags` when test the ctags options?

ddanderson:
@keithbostic, looks good modulo a small nit, not sure it's worth fixing.

dist/s_tags
keithbostic: I thought of that too, but currently this script can't be run from anywhere other than `dist` (it assumes there are `../src/include` and a `../build_posix/` directories). I agree it might be a little late at that point, though, I'll put in a quick test to ensure we're where we think we are.

ddanderson:
@keithbostic, new changes LGTM, I'll merge.

ddanderson:
beat me to it.

PR #3049, opened 2016-09-16
WT-2919 dist scripts force exit 0 on error

keithbostic:
@agorrod, when fixing `s_tags` today, I noticed we were forcing script exits of 0 almost everywhere (we probably just cut-and-pasted from one place it made sense). This doesn't clean up error handling completely, but I think it makes it incrementally better.

I'm pointing you at this one because I thought there might be something in the `s_all` changes you made awhile back that require scripts to always exit 0?

Anyway, feel free to simply discard this one without further discussion if you don't like it, entirely up to you.

### DIFF
--- a/dist/s_all
+++ b/dist/s_all
@@ -124,3 +124,4 @@ echo 'dist/s_all run finished'
 if test $errmode -ne 0; then
 	exit $errfound;
 fi
+exit 0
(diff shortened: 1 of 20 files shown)


agorrod:
@keithbostic good catch - I like this change. I'm trying to keep testing stable today. I'll merge it in a bit later this week.

PR #3051, opened 2016-09-19
WT-2864 Reconfiguring the checkpoint server can lead to hangs

keithbostic:
@agorrod, for your review/consideration, I thought this might be a reasonable way to attack the problem in WT-2864.


agorrod:
Thanks @keithbostic this change is a good way to move forward.

PR #3053, opened 2016-09-20
WT-2923 heap-use-after-free on address in compaction

keithbostic:

### DIFF
--- a/src/btree/bt_compact.c
+++ b/src/btree/bt_compact.c
@@ -171,30 +171,64 @@ int
 __wt_compact_page_skip(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
 {
 	WT_BM *bm;
+	WT_DECL_RET;
 	size_t addr_size;
 	u_int type;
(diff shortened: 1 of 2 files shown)


keithbostic:
@michaelcahill, @agorrod, while I was staring at this today, I remembered compaction can cause checkpoint to wait (it acquires the `WT_BTREE.flush_lock`), while it walks the tree. I was thinking a better solution would be for compact to periodically check if sync is also walking the tree, in which case it could either wait or return EBUSY.

Thoughts?

michaelcahill:
@keithbostic, waiting for `btree->flush_lock` seems orthogonal to this fix.  If we have a workload that demonstrates an issue, we could look at some other way to serialize checkpoints and compact operations, but it isn't obviously bad to me to wait on the `flush_lock`.

Regardless, the fix lgtm -- FTR, my understanding is that compact was racing with eviction and the fix is to have compact swap pages to locked before examining them.  I'll merge.

PR #3054, opened 2016-09-21
WT-2924 hang when thread holding table lock tapped for eviction

keithbostic:
@agorrod, @michaelcahill, there's some discussion in WT-2924, but the change here is to wrap all high-level locks in a set/clear of the `WT_SESSION_NO_EVICTION` flag so we never tap a thread holding a high-level lock for eviction.

This may cast too wide a net -- I suppose the alternative/obvious change is to enumerate more locks in `__wt_cache_eviction_check()`, but I prefer this change because I think it makes it unlikely we'll introduce this bug again.

EDIT: we can't declare local variables in the WT_WITH_XXX lock macros (because we nest them), so I fell back to enumerating the locks when we check.


keithbostic:
autotest: rebuild this please

keithbostic:
autobuild: retest this please

keithbostic:
This branch has been replaced by #3056.

PR #3057, opened 2016-09-22
WT-2926 WT_CONNECTION.reconfigure can attempt unlock of not-locked lock

keithbostic:


keithbostic:
@agorrod, I believe the reason I found this one is because FreeBSD apparently detects attempts to unlock pthread mutexes that haven't been locked.

AWS offers FreeBSD 10 AMIs, maybe we should occasionally spin one up and run the static and stress tests? Not sure if that's easy to do in our current test structure, but it might be useful.

agorrod:
Thanks @keithbostic lgtm.

PR #3058, opened 2016-09-23
WT-2921 Reset the count when the last hazard pointer is cleared.

michaelcahill:


agorrod:

PR #3059, opened 2016-09-23
WT-2336 syscall tester to test file system operations

ddanderson:
This is a rough version of a system call tester.  The idea proposed by @agorrod is to have a simple program that does a few small WT operations, run strace on it, capture the output and compare it to a 'known good' output.  In this case, the source of the program to be tested is in `test/syscall/wt2336_base/main.c` and the script that runs and compares is `test/syscall/syscall.py`.  The script with no arguments will look for `.run` files under `test/syscall/*`; each `.run` file lives in the directory with the C source for the program to be run.  (There can be multiple `.run` files, for cases when the C program is to be run with different arguments).  The `.run` file contains some header information (what syscalls to collect, what system this test runs under) terminated by a `RUN` statement, following that is a list of strace output to be compared.  The .run file is intended to be read like a C program but is really a rather constrained syntax.  It does allow variables to be assigned to results of system calls (for file descriptors) to be used later in the script.  It also allows some degree of fuzziness in matching.

At the moment, the script runs on Linux (using strace) and OSX (using a modified dtruss that allows one to specify a subset of system calls to be traced).  There are some small hacks to make it work on OSX -- I can go into this if there is interest, and probably make some of it better.  We would need an approximate equivalent to strace on Windows to get it ported there.

This is _rough_ because there are some TODOs listed in the python code. The script could use a little more cleanup, refactoring, documenting.  I'd like to get some feedback about the overall approach, the syntax and capabilities.

The `.run` files I've given do work (they match correctly), I'm not sure if they are matching too much detail (for example, are the movements of the turtle file important for "proving" what we want to know about the file system operations?)

@agorrod, can you take a look?


ddanderson:
retest this please

ddanderson:
retest this please

agorrod:
Thanks @ddanderson I really like the shape of this change. I think we should iterate on the content of the expected calls file, I'm happy to work with you on that.

>> OSX (using a modified dtruss that allows one to specify a subset of system calls to be traced)

I'm not inclined to chase OS X support to the ground - since it appears to be complex. I'm inclined to get this working on Linux first, then think about support on other platforms.

One other thing that occurs to me is that a different set of configure settings that change supported system methods may alter correctness. For example ``HAVE_FDATASYNC`` in ``wiredtiger_config.h``. I think we should either choose a set of supported calls per platform or define a set of different configurations to test.

ddanderson:
Thanks @agorrod, I definitely need your help on determining what's required in the expected calls.  I'm going to move the high level discussion to the [JIRA ticket.](https://jira.mongodb.org/browse/WT-2336)

ddanderson:
This pull request is *not* yet ready for final review.  I'm trying to get it to run with a jenkins job, and it's not quite working yet.

ddanderson:
@agorrod, whenever you are available, this is ready for review or merge. With a bit of struggle, I figured out what was going wrong with my Jenkins job (wiredtiger-syscall-linux), and it runs successfully now.

ddanderson:
I've left some TODO items in the JIRA ticket about modifications to jenkins jobs needed to really finish off this work.

agorrod:
@daveh86 Could you work with @ddanderson to get this branch merged and make the necessary Jenkins updates?

PR #3060, opened 2016-09-24
WT-2928 eviction failing to switch queues leads to starvation

keithbostic:


agorrod:
Thanks @michaelcahill lgtm - I'm happy to merge once testing has finished.

PR #3061, opened 2016-09-24
WT-2924 Ensure we are doing eviction when threads are waiting for it

keithbostic:

There are a couple of different things in here that I noticed reading the code, you may want to cherry-pick them into whatever you end up merging.


michaelcahill:
Thanks @keithbostic, lgtm.  I'll merge so this gets some testing over the weekend.

PR #3062, opened 2016-09-26
WT-1592 Add per-dhandle current cache usage statistics

agorrod:
Exposed via a new 'cache_walk' statistics configuration option.

Can be gathered by using:
``statistics=(fast,cache_walk),statistics_log=(wait=10,sources=[file:])``
Can be viewed using wtstats.py visualizer.


agorrod:
@sulabhM would you mind doing a review for me?

src/conn/conn_api.c
```diff
@@ -1671,12 +1671,23 @@ __conn_statistics_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+		    WT_CONN_STAT_CACHE_WALK | WT_CONN_STAT_FAST);
 		++set;
 	}
 	WT_RET_NOTFOUND_OK(ret);
 
 	if ((ret = __wt_config_subgets(
+	    session, &cval, "cache_walk", &sval)) == 0 && sval.val != 0) {
+		if (flags == 0)
+			WT_RET_MSG(session, EINVAL,
+			    "the value \"cache_walk\" is not compatible with "
```
sulabhM: If I provide "cache_walk" only, and don't necessarily provide "none", it will still get set to none and error out saying "not compatible with none statistics configuration". Do you instead want to say on the lines of "cache_walk can only be provided with either fast or all".

src/cursor/cur_stat.c
```diff
@@ -606,24 +606,37 @@ __wt_curstat_open(WT_SESSION_IMPL *session,
...
 			F_SET(cst, WT_CONN_STAT_FAST);
 		}
 		WT_ERR_NOTFOUND_OK(ret);
+
+		if ((ret = __wt_config_subgets(session,
+		    &cval, "cache_walk", &sval)) == 0 && sval.val != 0) {
+			if (!F_ISSET(cst, WT_CONN_STAT_ALL | WT_CONN_STAT_FAST))
+				WT_ERR_MSG(session, EINVAL,
+				    "the value \"cache_walk\" is not "
+				    "compatible with \"none\" statistics "
```
sulabhM: If I provide "cache_walk" only, and don't necessarily provide "none", it will still get set to none and error out saying "not compatible with none statistics configuration". Do you instead want to say on the lines of "cache_walk can only be provided with either fast or all".

src/conn/conn_api.c
agorrod: I was torn about this one. I agree that none is implicit, so the message isn't as useful as ideal to the user. I wanted to avoid having a growing list of options in the error message that match. I'll switch it to be consistent with the message for "clear"

agorrod:
autobuild: retest this please

agorrod:
@michaelcahill would you please take a look at these changes?

src/btree/bt_stat.c
```diff
@@ -44,6 +44,9 @@ __wt_btree_stat_init(WT_SESSION_IMPL *session, WT_CURSOR_STAT *cst)
 	WT_STAT_SET(session, stats, cache_bytes_inuse,
 	    __wt_btree_bytes_inuse(session));
 
+	if (F_ISSET(cst, WT_CONN_STAT_CACHE_WALK))
+		WT_RET(__wt_curstat_cache_walk(session));
+
 	/* Everything else is really, really expensive. */
```
michaelcahill: It seems like this comment isn't quite right any more -- or are we saying that a cache walk is "really expensive" but not "really, really expensive"?  ;-)

src/include/connection.h
```diff
@@ -285,12 +285,13 @@ struct __wt_connection_impl {
...
-#define	WT_CONN_STAT_JSON	0x08	/* output JSON format */
-#define	WT_CONN_STAT_ON_CLOSE	0x10	/* output statistics on close */
-#define	WT_CONN_STAT_SIZE	0x20	/* "size" statistics configured */
+#define	WT_CONN_STAT_ALL	0x01	/* All statistics configured */
+#define	WT_CONN_STAT_CACHE_WALK	0x02	/* Cache walk statistics configured */
+#define	WT_CONN_STAT_CLEAR	0x04	/* Clear statistics after gathering */
+#define	WT_CONN_STAT_FAST	0x08	/* Fast statistics configured */
+#define	WT_CONN_STAT_JSON	0x10	/* Output in JSON format */
+#define	WT_CONN_STAT_ON_CLOSE	0x20	/* Output statistics on close */
+#define	WT_CONN_STAT_SIZE	0x40	/* Size statistics configured */
```
michaelcahill: It surprised me on revisiting this code that these flags are also used in `WT_CURSOR_STAT`.  They are also a mix of statistics categories and behavior of the statistics logging thread.

How would you feel about renaming the category flags `WT_STAT_TYPE_*` and moving them into `flags.py`?

dist/stat_data.py
```diff
@@ -498,6 +508,28 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_write_restore', 'pages written requiring in-memory restoration'),
 
     ##########################################
+    # Cache content statistics
+    ##########################################
+    CacheWalkStat('cache_state_avg_written_size', 'Average on-disk page image  size seen', 'no_clear,no_scale'),
```
michaelcahill: double space

dist/stat_data.py
```diff
@@ -498,6 +508,28 @@ def __init__(self, name, desc, flags=''):
...
+    CacheWalkStat('cache_state_count_memory', 'Number of pages that have never been written from cache', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_count_not_queueable', 'Number of pages that could not be queued for eviction', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_count_queued', 'Number of pages currently queued for eviction', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_count_smaller_alloc_size', 'Number of on-disk page image sizes smaller than a single allocation unit', 'no_clear,no_scale'),
```
michaelcahill: In all of the above, "count" seems redundant in the name.  We mostly assume stats are a count unless they have "size" in them, I think...

dist/stat_data.py
```diff
@@ -498,6 +508,28 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_write_restore', 'pages written requiring in-memory restoration'),
 
     ##########################################
+    # Cache content statistics
+    ##########################################
+    CacheWalkStat('cache_state_avg_written_size', 'Average on-disk page image  size seen', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_count_memory', 'Number of pages that have never been written from cache', 'no_clear,no_scale'),
```
michaelcahill: From the description, I can't tell if this means "newly created in memory and never written" or "read from disk and clean" or "read from disk and dirtied but not written".  Can you clarify?

dist/stat_data.py
```diff
@@ -498,6 +508,28 @@ def __init__(self, name, desc, flags=''):
...
+    CacheWalkStat('cache_state_gen_avg_gap', 'Average difference between current eviction generation when the page was last considered', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_gen_current', 'Current eviction generation', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_gen_max_gap', 'Maximum difference between current eviction generation when the page was last considered', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_max_pagesize', 'Maximum page size seen', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_min_written_size', 'Minimum on-disk page image size seen', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_pages', 'Total number of pages currently in cache', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_pages_clean', 'Number of clean pages currently in cache', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_pages_dirty', 'Number of dirty pages currently in cache', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_pages_internal', 'Number of internal pages currently in cache', 'no_clear,no_scale'),
+    CacheWalkStat('cache_state_pages_leaf', 'Number of leaf pages currently in cache', 'no_clear,no_scale'),
```
michaelcahill: Do we track these breakdowns of page states elsewhere?  Another way to ask the question: could we track these in "fast" stats without a cache walk?

agorrod: We don't currently, but it would be simple enough to add something into the cache tracking functions in ``btree.i`` - would you like me to do that instead?

agorrod:
autobuild: retest this please

agorrod:
@sulabhM I've made some fairly significant changes in this branch, including making btree {{tree_walk}} statistics gathering an explicit option, not just implied by {{all}}. Would you mind doing another review?

michaelcahill:
Thanks @agorrod, I've taken a look and this LGTM now.

sueloverso:
While I know this has been merged, I've become very sensitive to the size of the dhandle this week.  These new stats add 3312 bytes to the already large size of the dhandle structure.

PR #3063, opened 2016-09-26
WT-2928 Don't give up when the urgent queue is half empty.

michaelcahill:


src/evict/evict_lru.c
```diff
@@ -1579,13 +1579,12 @@ __evict_get_ref(
 	if (is_server &&
 	    (cache->evict_empty_score > WT_EVICT_SCORE_CUTOFF ||
 	    __evict_queue_empty(cache->evict_fill_queue, false))) {
-		do {
+		while ((ret = __wt_spin_trylock(
+		    session, &cache->evict_queue_lock)) == EBUSY)
 			if ((!urgent_ok ||
 			    __evict_queue_empty(urgent_queue, false)) &&
 			    !__evict_queue_full(cache->evict_fill_queue))
 				return (WT_NOTFOUND);
```
michaelcahill: This is just a cleanup change, as far as I'm concerned -- `while >> do`.

src/evict/evict_lru.c
```diff
@@ -1621,7 +1620,8 @@ __evict_get_ref(
 	 */
 	for (;;) {
 		/* Verify there are still pages available. */
-		if (__evict_queue_empty(queue, is_server)) {
+		if (__evict_queue_empty(
+		    queue, is_server && queue != urgent_queue)) {
```
michaelcahill: This is the real fix -- don't give up on the urgent queue.

src/evict/evict_lru.c
agorrod: Fine by me. Have you seen the potential extra try lock call when the queues are empty add to contention on the eviction queue lock?

michaelcahill: I haven't measured the perf impact at all: the case where the queues are completely empty should be caught right at the top of `__evict_get_ref`...

PR #3064, opened 2016-09-27
WT-2931 Enforce limits on dirty data for in-memory.

michaelcahill:
Reverts WT-2886 commit ac1f7401dcb8be345973f7787d9121c5c321bf7b.


agorrod:
Thanks @michaelcahill lgtm

PR #3065, opened 2016-09-27
WT-2858 rename wtperf's CONFIG structure

keithbostic:
@agorrod, the next chunk of work, there should be no real changes in this one.


agorrod:
Thanks @keithbostic lgtm.

PR #3066, opened 2016-09-28
WT-2907 Fixed java concurrent close test to have both insert threads and scan threads.

ddanderson:
Additionally some minor code cleanup.
Confirmed that the test still passes.


agorrod:
Thanks @ddanderson lgtm - I've merged.

PR #3067, opened 2016-09-28
WT-2831 Don't checkpoint if there have been no changes

sulabhM:
Add a modified flag to the connection structure. Do a checkpoint if the flag is set, not otherwise.
Set this flag whenever we set btree->modified to true. Clear the connection->modified flag after taking a transaction id for checkpoint.


sulabhM:
@daveh86 : Mind taking a look

sulabhM:
autobuild: retest this please

src/include/connection.h
```diff
@@ -172,6 +172,12 @@ struct __wt_connection_impl {
 	WT_SPINLOCK turtle_lock;	/* Turtle file spinlock */
 
 	/*
+	 * Is there a data/schema change that needs to be the part of a
+	 * checkpoint.
+	 */
+	bool modified;
```
sulabhM: Moved.

src/conn/conn_ckpt.c
```diff
@@ -87,22 +87,26 @@ __ckpt_server(void *arg)
...
+				 * In case we crossed the log limit during the
+				 * checkpoint and the condition variable was
+				 * already signalled, do a tiny wait to clear
+				 * it so we don't do another checkpoint
+				 * immediately.
+				 */
+				__wt_cond_wait(session, conn->ckpt_cond, 1);
+			}
+		} else
+			WT_STAT_CONN_INCR(session, txn_checkpoint_skipped);
```
sulabhM: For a small window btree->modified might actually be true while conn->modified is false and yet to be changed to true. This is expected and in this case the next checkpoint will not be skipped.

Since btree->modified and conn->modified are not read/written together under a lock, throughout the program execution there will be short instances when they will not be in sync with each other.

daveh86: This seems like worthwhile detail to add in a comment at this point then.

sulabhM:
LSMs get taken care of: when any btree that is a part of LSM is marked dirty, the connection is marked dirty as well.

src/conn/conn_ckpt.c
sulabhM: Done, added a comment.

sulabhM:
autobuild: retest this please

daveh86:
This LGTM

src/include/btree.i
```diff
@@ -530,6 +530,9 @@ __wt_page_modify_set(WT_SESSION_IMPL *session, WT_PAGE *page)
 		WT_FULL_BARRIER();
 	}
 
+	if (S2C(session)->modified == 0)
+		S2C(session)->modified = 1;
+
```
michaelcahill: Consider moving this up inside the `if` before setting `S2BT(session)->modified = 1`.

Or is there some case where the btree is dirty on entry but the connection could be clean?

src/txn/txn_ckpt.c
```diff
@@ -1440,8 +1451,10 @@ __checkpoint_tree(
 	 * If the checkpoint didn't complete successfully, make sure the
 	 * tree is marked dirty.
 	 */
-	if (ret != 0 && !btree->modified && was_modified)
+	if (ret != 0 && !btree->modified && was_modified) {
 		btree->modified = 1;
+		S2C(session)->modified = 1;
```
michaelcahill: Can we check that `conn->modified == 0` before setting it?  Otherwise, there could be lots of updates to that shared location here (this path happens a lot in some workloads).

src/include/btree.i
sulabhM: > Or is there some case where the btree is dirty on entry but the connection could be clean?

Yes, the intention is to make sure we update connection to dirty in the case btree is marked dirty and connection is not while entering the function. I don't have a concrete example on how this might happen. Consider the duration between marking connection clean after taking transaction id and marking btree clean after checkpointing this specific btree. During this time period btree is marked dirty and connection clean for any thread calling __wt_page_modify_set() in parallel to the checkpoint thread.

I think it is cleaner to not assume a coupling this tight between btree and connection modified states. We should make sure that __wt_page_modify_set() marks connection dirty if btree is supposed to be dirty, irrespective of what state these flags were when starting to execute the function.

src/txn/txn_ckpt.c
sulabhM: Done.

sulabhM:
graph


@michaelcahill , the series in black colour is after the change vs red in before the change.
The graph plots maximum read latencies with time. The extreme peaks for the red series corresponds to the checkpoints, which is missing in the black series. The workload is readonly over 5000 collections, with connection configuration is asking for checkpoints to run every 20 secs.
Timeseries data for the black series run show that the checkpoints get skipped.

I am considering a run of perf tests on this branch for further testing.

agorrod:
@sulabhM the Jenkins performance tests showed a regression, but I suspect it is because the base is an older version of develop. I've merged in the latest develop, we should check the performance tests again tomorrow for regressions.

sulabhM:
@michaelcahill : Can you take another look at this review.

src/conn/conn_ckpt.c
```diff
@@ -87,22 +87,36 @@ __ckpt_server(void *arg)
...
+		 * A connection is marked dirty whenever a btree gets marked
+		 * dirty, which reflects upon a change in the database that
+		 * needs to be checkpointed. Said that, there can be short
+		 * instances when a btree gets marked dirty and the connection
+		 * is yet to be. We might skip a checkpoint in that short
+		 * instance, which is okay because by the next time we get to
+		 * checkpoint, the connection would have been marked dirty and
+		 * hence the checkpoint will not be skipped this time.
+		 */
+		if (conn->modified == 1) {
```
michaelcahill: style nit: since `conn->modifed` is a `bool`, just write `if (conn->modified) ...`.

src/include/btree.i
```diff
@@ -530,6 +530,9 @@ __wt_page_modify_set(WT_SESSION_IMPL *session, WT_PAGE *page)
 		WT_FULL_BARRIER();
 	}
 
+	if (S2C(session)->modified == 0)
+		S2C(session)->modified = 1;
```
michaelcahill: Similarly, this should be:

```c
if (!S2C(session)->modified)
    S2C(session)->modified = true;
```

src/txn/txn_ckpt.c
```diff
@@ -1440,8 +1451,11 @@ __checkpoint_tree(
 	 * If the checkpoint didn't complete successfully, make sure the
 	 * tree is marked dirty.
 	 */
-	if (ret != 0 && !btree->modified && was_modified)
+	if (ret != 0 && !btree->modified && was_modified) {
 		btree->modified = 1;
+		if (S2C(session)->modified == 0)
+			S2C(session)->modified = 1;
+	}
```
michaelcahill: As well as changing `S2C(session)->modified` here, please update the line above to be `btree->modified = true`.

src/txn/txn_ckpt.c
```diff
@@ -1354,7 +1365,7 @@ __checkpoint_tree(
 	 * checkpoint is created, and then things get bad.
 	 */
 	WT_ERR(__wt_page_modify_init(session, btree->root.page));
-	__wt_page_modify_set(session, btree->root.page);
+	__wt_page_only_modify_set(session, btree->root.page);
```
michaelcahill: Please add a comment explaining why we don't want to make the tree modified here.

src/reconcile/rec_write.c
```diff
@@ -564,10 +564,12 @@ __rec_write_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
 		 * barrier after the change for clarity (the requirement is the
 		 * flag be set before a subsequent checkpoint reads it, and
 		 * as the current checkpoint is waiting on this reconciliation
-		 * to complete, there's no risk of that happening)
+		 * to complete, there's no risk of that happening).
 		 */
 		btree->modified = 1;
 		WT_FULL_BARRIER();
+		if (S2C(session)->modified == 0)
+			S2C(session)->modified = 1;
```
michaelcahill: Similarly use `bool` style.

src/include/btree.i
michaelcahill: OK, fair enough.  Maybe add a comment to that effect.

sulabhM:
Addressed Michael's comments.

agorrod:
Thanks @sulabhM lgtm - I'll merge

PR #3068, opened 2016-09-28
WT-2933 Fix a race between named snapshots and checkpoints.

michaelcahill:
The changes in WT-2798 introduced a bug into the allocation of named
snapshots when a checkpoint is running.  This bug should not have led to
invalid reads or data corruption but violated some sanity checking in
diagnostic builds.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -303,7 +303,6 @@ __wt_txn_update_oldest(WT_SESSION_IMPL *session, uint32_t flags)
 	    __wt_try_readlock(session, txn_global->scan_rwlock)) != 0)
 		return (ret == EBUSY ? 0 : ret);
 	__txn_oldest_scan(session, &oldest_id, &last_running, &oldest_session);
-	WT_ASSERT(session, !WT_TXNID_LT(oldest_id, txn_global->oldest_id));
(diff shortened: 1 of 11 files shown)


michaelcahill:
Note that the assertions here have paranoia dialled up to 11.  I'll make a pass through tomorrow and strip out any that aren't needed.

src/txn/txn.c
```diff
@@ -250,8 +250,14 @@ __txn_oldest_scan(WT_SESSION_IMPL *session,
 
 	/* The oldest ID can't move past any named snapshots. */
 	if ((id = txn_global->nsnap_oldest_id) != WT_TXN_NONE &&
-	    WT_TXNID_LT(id, oldest_id))
-		oldest_id = id;
+	    WT_TXNID_LT(id, oldest_id)) {
+#ifdef HAVE_DIAGNOSTIC
+		WT_DIAGNOSTIC_YIELD;
+		if ((id = txn_global->nsnap_oldest_id) != WT_TXN_NONE &&
+		    WT_TXNID_LT(id, oldest_id))
```
agorrod: Making this conditional is an odd difference between diagnostic and non-diagnostic modes. Could it be an assertion instead?

agorrod:
Thanks @michaelcahill lgtm.

PR #3069, opened 2016-09-29
WT-2932 Add a configuration option allowing tables to ignore cache limits

agorrod:
Only available for in-memory configurations, designed to be used for
applications that maintain a small amount of content that is required
for correctness.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -183,7 +183,7 @@ file_config = format_meta + [
         configure Huffman encoding for values.  Permitted values are
         \c "none", \c "english", \c "utf8<file>" or \c "utf16<file>".
         See @ref huffman for more information'''),
-    Config('in_memory_ignore_cache_size', 'false', r'''
+    Config('ignore_in_memory_cache_size', 'false', r'''
         allow update and insert operations to proceed even if the cache is
         already at capacity. Only valid in conjunction with in-memory
         databases. Should be used with caution - this configuration allows
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -183,6 +183,12 @@ file_config = format_meta + [
         configure Huffman encoding for values.  Permitted values are
         \c "none", \c "english", \c "utf8<file>" or \c "utf16<file>".
         See @ref huffman for more information'''),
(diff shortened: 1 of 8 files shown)


agorrod:
autobuild: retest this please

src/include/api.h
```diff
@@ -139,7 +139,9 @@
 	(s) = (WT_SESSION_IMPL *)(cur)->session;			\
 	TXN_API_CALL_NOCONF(s, WT_CURSOR, n, cur,			\
 	    ((bt) == NULL) ? NULL : ((WT_BTREE *)(bt))->dhandle);	\
-	if (F_ISSET(S2C(s), WT_CONN_IN_MEMORY) && __wt_cache_full(s))	\
+	if (F_ISSET(S2C(s), WT_CONN_IN_MEMORY) &&			\
+	    !F_ISSET((WT_BTREE *)(bt), WT_BTREE_IGNORE_CACHE) &&	\
+	    __wt_cache_full(s))						\
```
michaelcahill: whitespace before backslash

agorrod: I believe that is a git diff-ism, there is a tab character after the ``&``

dist/api_data.py
```diff
@@ -183,6 +183,12 @@ def __cmp__(self, other):
         configure Huffman encoding for values.  Permitted values are
         \c "none", \c "english", \c "utf8<file>" or \c "utf16<file>".
         See @ref huffman for more information'''),
+    Config('in_memory_ignore_cache_size', 'false', r'''
+        allow update and insert operations to proceed even if the cache is
+        already at capacity. Only valid in conjunction with in-memory
+        databases. Should be used with caution - this configuration allows
+        WiredTiger to consume memory over the configured cache limit''',
+        type='boolean'),
```
michaelcahill: How about renaming this `ignore_in_memory_cache_size`?  I don't feel super strongly, but having the verb at the beginning reads better to me.

src/include/api.h
michaelcahill: OK, that's an odd one.

PR #3070, opened 2016-09-29
WT-2936 Enable strict by default. Setup for --disable-strict

daveh86:


michaelcahill:
As discussed elsewhere, let's drop this change to avoid causing build failures in future compiler releases.

agorrod:
We decided not to take this change - see the JIRA ticket for reasoning.

PR #3071, opened 2016-09-29
WT-2937 Only do an eviction walk after a page has been taken from the current queue.

michaelcahill:


src/evict/evict_lru.c
```diff
@@ -506,12 +506,6 @@ __evict_update_work(WT_SESSION_IMPL *session)
 		F_CLR(cache, WT_CACHE_EVICT_CLEAN | WT_CACHE_EVICT_CLEAN_HARD);
 	}
 
-	/* If threads are blocked by eviction we should be looking for pages. */
-	WT_ASSERT(session, !F_ISSET(cache, WT_CACHE_EVICT_CLEAN_HARD) ||
-	    F_ISSET(cache, WT_CACHE_EVICT_CLEAN));
-	WT_ASSERT(session, !F_ISSET(cache, WT_CACHE_EVICT_DIRTY_HARD) ||
-	    F_ISSET(cache, WT_CACHE_EVICT_DIRTY));
-
```
michaelcahill: This assertion is now trivially satisfied so there is no point keeping it.

src/evict/evict_lru.c
```diff
@@ -1569,44 +1562,39 @@ __evict_get_ref(
...
 	    (cache->evict_empty_score > WT_EVICT_SCORE_CUTOFF ||
-	    __evict_queue_empty(cache->evict_fill_queue, false))) {
-		while ((ret = __wt_spin_trylock(
-		    session, &cache->evict_queue_lock)) == EBUSY)
-			if ((!urgent_ok ||
-			    __evict_queue_empty(urgent_queue, false)) &&
-			    !__evict_queue_full(cache->evict_fill_queue))
-				return (WT_NOTFOUND);
+	    __evict_queue_empty(cache->evict_fill_queue, false)))
+		return (WT_NOTFOUND);
```
michaelcahill: I don't see much point spinning here.  If we're going to fill queues, just do it.

This condition probably deserves a (much) longer comment, though.

src/evict/evict_lru.c
```diff
@@ -1569,44 +1562,39 @@ __evict_get_ref(
...
-		 * configured), or application threads if necessary.
+		 * for more, but should only switch queues if there are no
+		 * other eviction workers.
 		 */
 		queue = cache->evict_current_queue;
 		other_queue = cache->evict_other_queue;
-		if (__evict_queue_empty(queue, is_server) &&
-		    !__evict_queue_empty(other_queue, is_server)) {
+		if (__evict_queue_empty(queue, server_only) &&
+		    !__evict_queue_empty(other_queue, server_only)) {
```
michaelcahill: This is subtle -- if there is a single eviction server, it will switch queues as soon as it gets half way through one.  Otherwise, it could give up unnecessarily.

We *don't* do that when there are multiple eviction workers because it leads to unnecessary contention on the queue while it is being filled.

agorrod:
Thanks @michaelcahill - these changes make sense to me. I'm going to switch the Jenkins perf tests to this branch. If they look OK I think we should merge.

michaelcahill:
There is a regression with inserts in `update-btree`.  This will need some more investigation.

sueloverso:
The latest changes impact `evict-lsm` plots with 4 eviction threads.  That is now consistently running slower than with 1 thread.

michaelcahill:
autobuild: retest this please

agorrod:

PR #3072, opened 2016-09-30
WT-2938 Change assembly file extensions to .sx

daveh86:


daveh86:
@agorrod , all of our platforms seem OK with this change. Can you review?

agorrod:
@daveh86 Will this change require any changes to the MongoDB build? I'd prefer to know before merging.

michaelcahill:
Out of interest, can we do an Evergreen patch build against the WiredTiger project to test this?

daveh86:
To @agorrod's question. No, it shouldnt' as the Scons script mongo uses reads from dist/filelist
to @michaelcahill's question. Running now.

agorrod:
Thanks @daveh86 lgtm

PR #3073, opened 2016-09-30
WT-2941 format should use faster key-generation functions

keithbostic:
@agorrod, since it made a difference for wtperf, might as well speed up format, too.

### DIFF
--- a/test/format/util.c
+++ b/test/format/util.c
@@ -78,7 +78,7 @@ key_gen_setup(WT_ITEM *key)
 }
 
 static void
-key_gen_common(WT_ITEM *key, uint64_t keyno, int suffix)
+key_gen_common(WT_ITEM *key, uint64_t keyno, const char * const suffix)
 {
 	int len;
 	char *p;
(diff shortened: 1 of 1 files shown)


keithbostic:
@ddanderson, there's some code in `cur_json.c:__json_unpack_put()`:
```
        case 'b':
        case 'h':
        case 'i':
        case 'l':
        case 'q':
                return (s +
                    (size_t)snprintf((char *)buf, bufsz, "%" PRId64, pv->u.i));
        case 'B':
        case 't':
        case 'H':
        case 'I':
        case 'L':
        case 'Q':
        case 'r':
        case 'R':
                return (s +
                    (size_t)snprintf((char *)buf, bufsz, "%" PRId64, pv->u.u));
```
that's calling `sprintf` to convert `int64_t` values to strings. We have some functions now that are hugely faster than `sprintf` at doing that work, is there any value in doing faster conversion here?

sueloverso:
I'd say the more consistency the better.  None of these are on performance critical paths anyway but faster is always better.  We would need to move those functions into the src/support directory rather than the test directory.

ddanderson:
@keithbostic, I agree with Sue.  It doesn't seem performance critical, but if we have the routines already, then why not.

test/format/util.c
```diff
@@ -78,19 +78,26 @@ key_gen_setup(WT_ITEM *key)
...
-	 * The key always starts with a 10-digit string (the specified cnt)
-	 * followed by two digits, a random number between 1 and 15 if it's
-	 * an insert, otherwise 00.
+	 * The key always starts with a 10-digit string (the specified row
+	 * number), followed by a suffix if it's an insert.
 	 */
-	len = sprintf(p, "%010" PRIu64 ".%02d", keyno, suffix);
+	u64_to_string_zf(keyno, key->mem, 11);
+	if (suffix == NULL)
+		len = 10;
```
agorrod: Does it matter that this is functionally different to the previous implementation? The old version always created the same length strings, this one will create strings with a different length depending on whether the suffix is NULL.

keithbostic: I totally missed that, I thought I'd matched the previous behavior! Thank you!

I pushed a change to get back to the previous behavior. I have no idea if having those two extra bytes is material to how format runs, but I'm not interested in figuring it out, either.

agorrod:
Thanks @keithbostic lgtm. I agree that the json usage isn't on a performance path, so I'm inclined to leave it alone

PR #3075, opened 2016-10-02
WT-2880 Add Zstandard compression support

keithbostic:
Add Zstd compression to WiredTiger.


keithbostic:
@agorrod, @michaelcahill, I think Zstd is in and running correctly.

I don't believe we can make it run in raw-compression mode, that is, we'll always want to configure Zstd the way we configure snappy, not the way we configure LZ4. I thought about simply ripping out the raw-compression code, but left it in with a big comment, let me know if you think it should simply be removed (and if so, should we leave the possibility open, or just rename Zstd support to "zstd", and forget about "zstd" and "zstd-noraw").

I suspect we'll also want to change the default compression-level, it's currently 19, but engineering at Facebook said "our most wanted compression levels are typically 3, 8 and 1, in that order, depending primarily on the expected speed / compression trade-off".

ext/compressors/lz4/lz4_compress.c
```diff
@@ -277,8 +277,9 @@ lz4_compress_raw(WT_COMPRESSOR *compressor, WT_SESSION *session,
 	(void)final;
 
 	sourceSize = (int)offsets[slots];		/* Type conversion */
-	targetDestSize =
-	    (int)((dst_len < page_max ? dst_len : page_max) - extra);
+	targetDestSize = (int)
+	    (((dst_len - sizeof(LZ4_PREFIX)) < page_max ?
+	    (dst_len - sizeof(LZ4_PREFIX)) : page_max) - extra);
```
michaelcahill: I'm struggling to parse this and when I do I'm not sure it's right.

`zlib_compress_raw` ignores `dst_len` completely.  Why is that okay, but lz4 checks it?

If we have to do something here, how about `extra += sizeof(LZ4_PREFIX)` before this?

ext/compressors/zstd/zstd_compress.c
```diff
@@ -0,0 +1,552 @@
...
+ * destination buffer as well.
+ *
+ * Use fixed-size, 4B values (WiredTiger never writes buffers larger than 4GB).
+ *
+ * The unused field is available for a mode flag if one is needed in the future,
+ * we guarantee it's 0.
+ */
+typedef struct {
+	uint32_t compressed_len;	/* True compressed length */
+	uint32_t uncompressed_len;	/* True uncompressed source length */
```
michaelcahill: Since zstd has `ZSTD_getDecompressedSize`, can we avoid storing it and only have an 8 byte prefix?

I see in the docs that the uncompressed size is optional but always included by `ZSTD_compress`, so this change would imply that we are very sure we won't support raw mode in the future.

ext/compressors/zstd/zstd_compress.c
```diff
@@ -0,0 +1,552 @@
...
+	 *	LZ4	2.101	620	3100
+	 *	snappy	2.091	480	1600
+	 *
+	 * Set the zstd compression level to 3: according to the zstd web site,
+	 * that reduces zstd's compression speed to around 200 MB/s, increasing
+	 * the compression ratio to 3.100 (close to zlib's best compression
+	 * ratio). In other words, position zstd as a zlib replacement, having
+	 * similar compression at much higher compression/decompression speeds.
+	 */
+	zstd_compressor->compression_level = 3;
```
michaelcahill: Maybe (for initial testing at least) we should have "zstd-1", "zstd-3", "zstd-8", "zstd-19", each with a different compression level?

ext/compressors/zstd/zstd_compress.c
```diff
@@ -0,0 +1,552 @@
...
+zstd_extension_init(WT_CONNECTION *connection, WT_CONFIG_ARG *config)
+{
+	int ret;
+
+	(void)config;    				/* Unused parameters */
+
+	if ((ret = zstd_add_compressor(connection, 1, "zstd")) != 0)
+		return (ret);
+	if ((ret = zstd_add_compressor(connection, 0, "zstd-noraw")) != 0)
+		return (ret);
```
michaelcahill: As discussed, let's at least flip this so we have the default "zstd" not include raw compression (matching snappy).  If testing suggests that the overhead of "zstd-raw" is justified given the on-disk space saving, we can leave it in, otherwise let's remove that code.

ext/compressors/zstd/zstd_compress.c
keithbostic: What we need in the decompression function is the length of the compressed bytes, not the length of the uncompressed bytes, so I don't think `ZSTD_getDecompressedSize()` helps us.

If I'm wrong in that statement, I don't think it's a problem to go in that direction, though. We can have completely different block prefixes for the raw and no-raw versions of the engine by having separate decompression functions (instead of the merged one like in the LZ4 driver), so I think we're safe.

ext/compressors/zstd/zstd_compress.c
keithbostic: I've gone ahead and removed the `raw` code, it greatly simplifies the driver, and I don't see any path to supporting raw compression in Zstd.

ext/compressors/lz4/lz4_compress.c
keithbostic: I think it's a bug in the`zlib_compress_raw()` function.

In short, `page_max` is the target page size taken from `WT_BTREE.{maxleafpage,maxintlpage}`, and `dst_len` is either the return value of the `WT_COMPRESSOR.pre_size` function or the source size. Currently, `page_max` can be larger or smaller than `dst_len`.

Since zlib doesn't have a pre_size function, `dst_len` is the source size, and the source size may be smaller than the target page size, which `zlib_compress_raw()` doesn't take into account. We don't fail because zlib is very, very unlikely to exceed the source size during compression, but I think we'd drop core if it ever did grow.

I've pushed a change in 78f145c to fix the zlib code, explain what's going on and to make the lz4 and zlib code look the same.

We could also change the reconciliation code to ensure the destination buffer is never smaller than the target page size (currently we only guarantee it's never smaller than the return of the pre_size function or the length of the source if there's no pre_size function). That might be safer for future raw compression functions.

keithbostic:
@michaelcahill, I've noticed this problem before, but never had the guts to dig in and fix it.

The zlib raw compression routine didn't handle errors gracefully, it just returned on Zlib function failure without calling deflateEnd, which would leak memory and scratch buffers. The rework in 1ff5d2c is pretty big, but I don't see a simpler solution.

keithbostic:
@michaelcahill, d12df74 is the commit of the changes to fallback to a larger number of reserved bytes if the default number fails. (It also fixes a screwup I made in the error handling in 1ff5d2c, so if you decide against this commit, I'll have to fix up the branch.)

I think this branch is now ready for merge, once it passes testing and you're happy with it.

keithbostic:
@agorrod, this branch adds configuration support to both zstd and zlib, so you can do something like:

```
ret = wiredtiger_open(home, NULL,
            "create,"
            "extensions=[/usr/local/lib/"
            "libwiredtiger_zlib.so=[config=[compression_level=3]]]", &conn);
```
and set the compression level from the application. Ideally, that will make it easier to benchmark compression engines from applications.

Unfortunately, it doesn't work for MongoDB, when the compression engines are built-in, because no `CONFIG_ARG` values are passed when `wiredtiger_open` calls the underlying init functions.

@michaelcahill and I were talking about this, and I wanted to ask your opinion.

Could we add a new `wiredtiger_open` configuration option to specify arguments passed to the underlying built-in initialization functions?

Maybe `builtin_config=[config=[compression_level=3]]`, and `wiredtiger_open` could extract `builtin_config` and build a `CONFIG_ARG` structure for the built-in init functions?

Unfortunately, that doesn't work if the underlying init functions have common values (imagine both zlib and zstd use `compression_level`, but the initialization values have different meanings). I guess we'd need to target a specific built-in?

Maybe `zlib_builtin=[config=[compression_level=3]]` would work?

Anybody have a better idea?

michaelcahill:
@keithbostic, I did most of this in a local branch on Friday -- I think it should be:
```c
builtin_extension_config={zlib=(compression_level=3)}
```
That allows each extension to be configured separately.  I only realized when doing this work that the `config` key is being passed through to extensions, rather than just the value of the config key as intended.  The fix for *that* is more involved because it requires some extra entries in the extension API.

I'll open a separate ticket for this: it is orthogonal to the work on zstd.

keithbostic:
@michaelcahill, @agorrod, now that WT-2985 is confirmed to be in the develop branch, I think this branch is ready to merge.

FTR, someone needs to review `zlib_compress_raw()`. There were a set of commits that changed it in significant ways (78f145c, 1ff5d2c, d12df74, da7ac14, 0020bc4, d54033f). I think there's value there (serious error-handling fixes and avoiding a `deflateCopy` call in the case of small blocks), but the changes are big.

agorrod:
FWIW: I just switch stress testing onto this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress

build_win/wiredtiger_config.h
```diff
@@ -70,6 +73,9 @@
 /* Define to 1 if you have the `z' library (-lz). */
 /* #undef HAVE_LIBZ */
 
+/* Define to 1 if you have the `zstd' library (-lzstd). */
+#define HAVE_LIBZSTD 1
```
agorrod: I'm surprised we default zstd to enabled on Windows - was that deliberate?

ext/compressors/zlib/zlib_compress.c
```diff
@@ -32,18 +32,10 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <wiredtiger_config.h>
 #include <wiredtiger.h>
 #include <wiredtiger_ext.h>
 
-/*
- * We need to include the configuration file to detect whether this extension
```
agorrod: This was called out explicitly, since we don't expect user implemented compressors to do it, and it's not "normal" for applications to include that header file. I don't feel super strongly if you do. I also wonder why the Windows specific define was there if it's not necessary any more.

build_win/wiredtiger_config.h
keithbostic: No, my mistake -- thanks for catching it!

Updated & pushed.

ext/compressors/snappy/snappy_compress.c
```diff
@@ -108,91 +108,78 @@ wt_snappy_compress(WT_COMPRESSOR *compressor, WT_SESSION *session,
...
-			/*
-			 * On decompression, snappy requires an exact compressed
-			 * byte count (the current value of snaplen). WiredTiger
-			 * does not preserve that value, so save snaplen at the
-			 * beginning of the destination buffer.
-			 *
-			 * Store the value in little-endian format.
-			 */
+	if (snret == SNAPPY_OK && snaplen + sizeof(size_t) < src_len) {
+		*result_lenp = snaplen + sizeof(size_t);
```
michaelcahill: A couple of comments here that aren't directly related to your change...

* I think it would be good to have a macro constant `SNAPPY_PREFIX` rather then repeating `sizeof(size_t)` through this code.
* It should also really be a `uint64_t`, since we're assuming that it is 64-bits.

ext/compressors/snappy/snappy_compress.c
```diff
@@ -108,91 +108,78 @@ wt_snappy_compress(WT_COMPRESSOR *compressor, WT_SESSION *session,
...
-	size_t snaplen;
-
-	wt_api = ((SNAPPY_COMPRESSOR *)compressor)->wt_api;
 
 	/*
 	 * Retrieve the saved length, handling little- to big-endian conversion
 	 * as necessary.
 	 */
-	snaplen = *(size_t *)src;
+	src_len = *(size_t *)src;
```
michaelcahill: I'm not a fan of using the parameter for a different purpose.  It was telling us something real about how much data is safe for the function to use.  Now that information gets replaced with whatever happened to be in the on-disk buffer.

ext/compressors/snappy/snappy_compress.c
```diff
@@ -108,91 +108,78 @@ wt_snappy_compress(WT_COMPRESSOR *compressor, WT_SESSION *session,
...
 #ifdef WORDS_BIGENDIAN
-	snaplen = snappy_bswap64(snaplen);
+	src_len = snappy_bswap64(src_len);
 #endif
-	if (snaplen + sizeof(size_t) > src_len) {
-		(void)wt_api->err_printf(wt_api,
-		    session,
-		    "wt_snappy_decompress: stored size exceeds buffer size");
-		return (WT_ERROR);
-	}
```
michaelcahill: This test for buffer overflow seems pretty important.  Why is it okay to ignore the real in-memory size of the buffer and trust whatever bits we happen to read from disk?  Can't this change turn reading corrupted data from an error case into a segfault?

ext/compressors/zlib/zlib_compress.c
keithbostic: Now I understand. Some of the compression engines had the comment, some didn't, so I just dropped it. I've added it back in for all of the engines.

I didn't chase the reason for the Windows-specific define. Zlib was the only engine that had it, and both snappy and lz4 used `inline`, so I was reasonably sure it wasn't necessary.

ext/compressors/zlib/zlib_compress.c
```diff
@@ -32,17 +32,15 @@
...
 /*
  * We need to include the configuration file to detect whether this extension
- * is being built into the WiredTiger library.
+ * is being built into the WiredTiger library; application-loaded compression
+ * functions won't need it.
  */
-#include "wiredtiger_config.h"
-#ifdef _MSC_VER
-#define	inline __inline
-#endif
```
michaelcahill: This change is required unless we remove `inline` from `zlib_find_slot`.

I'm pretty sure lz4 has never been built on Windows, and snappy only uses it for big endian swapping.

ext/compressors/zlib/zlib_compress.c
```diff
@@ -234,121 +232,163 @@ zlib_compress_raw(WT_COMPRESSOR *compressor, WT_SESSION *session,
...
 		/* Find the next slot we will try to compress up to. */
 		if ((curr_slot = zlib_find_slot(
-		    zs.total_in + zs.avail_out, offsets, slots)) > last_slot) {
-			zs.avail_in = offsets[curr_slot] - offsets[last_slot];
-			while (zs.avail_in > 0 && zs.avail_out > 0)
-				if ((ret = deflate(&zs, Z_SYNC_FLUSH)) != Z_OK)
-					return (zlib_error(compressor,
-					    session, "deflate", ret));
+		    zs->total_in + zs->avail_out, offsets, slots)) >
+		    last_slot) {
```
michaelcahill: This is unweildy, is it any better as:
```
		curr_slot = zlib_find_slot(
		    zs->total_in + zs->avail_out, offsets, slots);
		if (curr_slot > last_slot) {
```
?

ext/compressors/zstd/zstd_compress.c
```diff
@@ -0,0 +1,334 @@
...
+	size_t zstd_ret;
+
+	/*
+	 * Retrieve the saved length, handling little- to big-endian conversion
+	 * as necessary.
+	 */
+	src_len = *(size_t *)src;
+#ifdef WORDS_BIGENDIAN
+	src_len = zstd_bswap64(src_len);
+#endif
```
michaelcahill: Likewise here, I think we need to sanity check the bits from disk in case they are corrupted.

ext/compressors/snappy/snappy_compress.c
keithbostic: > I think it would be good to have a macro constant `SNAPPY_PREFIX` rather then repeating `sizeof(size_t)` through this code.
> It should also really be a `uint64_t`, since we're assuming that it is 64-bits.

Done & pushed.

One thing, we now assume a `size_t` can be stored in a `uint64_t`, which isn't necessarily true. It's hard to imagine a compression buffer exceeding 8B, so I can't imagine anything breaking, but technically we're living dangerously.

ext/compressors/snappy/snappy_compress.c
keithbostic: > I'm not a fan of using the parameter for a different purpose. It was telling us something real about how much data is safe for the function to use. Now that information gets replaced with whatever happened to be in the on-disk buffer.

Fixed & pushed.

ext/compressors/snappy/snappy_compress.c
keithbostic: > This test for buffer overflow seems pretty important. Why is it okay to ignore the real in-memory size of the buffer and trust whatever bits we happen to read from disk? Can't this change turn reading corrupted data from an error case into a segfault?

I agree, I've reverted that change, and added corresponding changes to LZ4 and Zstd, commit da25f35.

ext/compressors/zlib/zlib_compress.c
keithbostic: > I'm pretty sure lz4 has never been built on Windows, and snappy only uses it for big endian swapping.

Oh, that explains it -- sorry about that. Fixed and pushed, commit 87ef223.

ext/compressors/zstd/zstd_compress.c
keithbostic: > Likewise here, I think we need to sanity check the bits from disk in case they are corrupted.

Included in commit da25f35.

keithbostic:
@michaelcahill, I believe I've addressed all of your review comments for #3075.

michaelcahill:

mdcallag:
will this show up in a future 3.2 branch?

keithbostic:
> will [Zstd] show up in a future 3.2 branch?

I don't know the plan, hoping @agorrod can reply.

michaelcahill:
@mdcallag, even if we bump MongoDB 3.2 to the latest version of WiredTiger, I don't expect the MongoDB side of these changes to be backported to 3.2.  It may be possible to work around that for testing purposes, however.

Also, please note that there is another change coming in WT-2962 that is required to configure the Zstandard compression level in MongoDB.

PR #3076, opened 2016-10-03
WT-2920 New eviction statistics

keithbostic:

It occurs to me usecs is an awfully small unit for an aggregated statistic, but it is what we use elsewhere, so I guess it makes sense for consistency.

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -381,7 +381,8 @@ connection_stats = [
     ##########################################
     # Yield statistics
     ##########################################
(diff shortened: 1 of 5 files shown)


agorrod:
autobuild: retest this please

The last results weren't reported due to a GitHub outage.

keithbostic:
@agorrod, lgtm, merged.

Thanks for finding/fixing the `WT_STAT_ENABLED` bug I put in there!

PR #3079, opened 2016-10-04
WT-2946 dist/s_docs incompatible with OS X Xcode installation

keithbostic:
@ddanderson, a fix to use `-` instead of a temporary file.


ddanderson:
@keithbostic, works for me.

PR #3080, opened 2016-10-04
WT-2948 simplify error handling by making epoch time return never fail

keithbostic:
@agorrod, I've been meaning to do this for awhile, and the changes we're considering for WT-2920 are going to be easier with this change.

If you want the change, I'll flag a few places that might deserve real review, in the comments.

### DIFF
--- a/bench/wtperf/idle_table_cycle.c
+++ b/bench/wtperf/idle_table_cycle.c
@@ -34,21 +34,15 @@ check_timing(WTPERF *wtperf,
 {
 	CONFIG_OPTS *opts;
 	uint64_t last_interval;
-	int ret;
 
 	opts = wtperf->opts;
 
-	if ((ret = __wt_epoch(NULL, stop)) != 0) {
-		lprintf(wtperf, ret, 0,
-		    "Get time failed in cycle_idle_tables.");
-		wtperf->error = true;
-		return (ret);
-	}
+	__wt_epoch(NULL, stop);
 
 	last_interval = (uint64_t)(WT_TIMEDIFF_SEC(*stop, start));
 
 	if (last_interval > opts->idle_table_cycle) {
-		lprintf(wtperf, ret, 0,
+		lprintf(wtperf, ETIMEDOUT, 0,
 		    "Cycling idle table failed because %s took %" PRIu64
 		    " seconds which is longer than configured acceptable"
 		    " maximum of %" PRIu32 ".",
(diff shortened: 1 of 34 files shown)


bench/wtperf/idle_table_cycle.c
```diff
@@ -34,21 +34,15 @@ check_timing(WTPERF *wtperf,
...
-		wtperf->error = true;
-		return (ret);
-	}
+	__wt_epoch(NULL, stop);
 
 	last_interval = (uint64_t)(WT_TIMEDIFF_SEC(*stop, start));
 
 	if (last_interval > opts->idle_table_cycle) {
-		lprintf(wtperf, ret, 0,
+		lprintf(wtperf, ETIMEDOUT, 0,
```
keithbostic: Bug fix, `ret` wasn't being set.

sueloverso:
I like this change a lot.  I'll let @agorrod chime in, but I'm a fan and it LGTM.

src/os_posix/os_time.c
```diff
@@ -12,26 +12,26 @@
...
 {
 	WT_DECL_RET;
 
 #if defined(HAVE_CLOCK_GETTIME)
 	WT_SYSCALL(clock_gettime(CLOCK_REALTIME, tsp), ret);
 	if (ret == 0)
-		return (0);
-	WT_RET_MSG(session, ret, "clock_gettime");
+		return;
+	WT_PANIC_MSG(session, ret, "clock_gettime");
```
agorrod: @keithbostic How would you feel about changing this panic to a diagnostic assertion that retrieving the time doesn't fail, and then eating any error codes here.

That would mean that any caller of __wt_epoch needs to handle time not moving forward, but that's probably true regardless. I'd feel more comfortable with that change.

keithbostic: @agorrod, it's not that time wouldn't move forward, it's that time would be zero'd out, so the callers would have to deal with:

1. a valid time followed by a zero'd out time,
2. a zero'd out time followed by a valid time, or
3. one or two zero'd out times (there are consumers of `__wt_epoch()` that use the time for purposes other than start/stop computations, for example, statistics logging file names).

We could handle number 1 by saving a copy of the first valid time somewhere (the `WT_SESSION_IMPL` structure?), but I don't see any way to handle the other two without significantly complicating the `__wt_epoch()` API.

Obviously, I don't think the change is risky or I wouldn't have suggested it (we've never seen this API fail, and the Windows implementation can't fail). But, if consensus is there's too much risk, then I think we have to kill the branch, I don't see a path to minimizing the risk that's less complicated than the existing error handling.

sueloverso: I think the panic is okay given that we really have no expectation that it will fail.  I think the risk of a panic is less than the impact of, say, dealing with additional zeroed out time conditionals in the WT_TIMEDIFF macros and other places every time they're used.

michaelcahill: Could we at least switch to `WT_SYSCALL_RETRY` then?  My concern is some rare case where `EINTR` is returned if a signal arrives at the wrong time...

agorrod:

PR #3081, opened 2016-10-05
WT-2950 Free all reconciliation memory between calls to eviction.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm, but I'd like to see a round of performance testing before merging - which probably means this lines up behind WT-2937

agorrod:
Thanks @michaelcahill I'm going to merge this. I'm a little bit concerned about performance consequences, but the benefit outweighs that at the moment. If we see workloads where there is a degradation we can use them to fine-tune how aggressive we are at releasing memory.

PR #3082, opened 2016-10-05
WT-2949 Add option to skip closing the connection after test runs.

sueloverso:
@agorrod Please review this change to add a `close_conn` option to `wtperf`.  I default it to false so that most tests just exit.  I added the option to the config files where the workload is a populate phase and a later test is expecting to set `create=false`.

It saves 9 minutes off the `wiredtiger-perf-stress` test, from 59 down to 50 minutes.  I expect the savings to be less for the smaller and shorter test groups.

### DIFF
--- a/bench/wtperf/idle_table_cycle.c
+++ b/bench/wtperf/idle_table_cycle.c
@@ -34,21 +34,15 @@ check_timing(WTPERF *wtperf,
 {
 	CONFIG_OPTS *opts;
 	uint64_t last_interval;
-	int ret;
 
 	opts = wtperf->opts;
 
-	if ((ret = __wt_epoch(NULL, stop)) != 0) {
-		lprintf(wtperf, ret, 0,
-		    "Get time failed in cycle_idle_tables.");
-		wtperf->error = true;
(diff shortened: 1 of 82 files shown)


michaelcahill:
@sueloverso I'm uncomfortable having this default `false` -- can we go the other way and set it to false for the long-running tests where we're sure we'll never care about using the data again?

agorrod:
I'm voting with @michaelcahill on this one - enabling the option default doesn't feel right.

> can we go the other way and set it to false for the long-running tests where we're sure we'll never care about using the data again?

We could set it to false in the ``wtperf_run.sh`` script used by Jenkins, using ``-o close_conn=false`` as part of the wtperf command line. That shouldn't introduce any problems when reproducing Jenkins regressions locally, since the shutdown time isn't included in any of the throughput calculations.

sueloverso:
I made the changes and this is ready to merge.

agorrod:
Thanks @sueloverso I'm happy with this change now, though I'd prefer if the option description mentioned that data can be lost if it's enabled.

PR #3083, opened 2016-10-06
WT-2947 replace test suite populate functions with *DataSet classes

ddanderson:
The initial commit is part 1: introduce *DataSet and create temporary compatability layer.  I wanted to get some confirmation on this approach before part 2: making the tedious changes, as there's about 50 some files that currently use the the populate functions.  When that is completed, the temporary functions in helper.py as well as all the *_populate* functions can be removed.

The motivation for this change is as part of WT-2833, I'm going to add some tests to do projections of data sets for dump/load, and it seemed like the function based approach had started to become unmanageable.  As an aside, it's possible that having an easy way to mix in projections may be helpful in other tests.


ddanderson:
@agorrod, can you take a look at this before I change all the callers in test/suite ?

agorrod:
Thanks @ddanderson looks like a nice improvement to me.

ddanderson:
Hi @agorrod, can you review? This is a big set of diffs, some are mechanical, some not. There's not a huge rush on this, I have some other tasks on my plate.  Thanks!

michaelcahill:
Thanks @ddanderson, lgtm.  I've looked through these changes and like the simplification and opportunity for future extensions.  I'll merge.

PR #3085, opened 2016-10-07
WT-2956 utility tests -h option is always overridden by the default setup

keithbostic:


keithbostic:
@agorrod, I also pointed lint at the csuite programs and cleaned up the results.

agorrod:
Thanks @keithbostic lgtm. The custom handler simplification is nice - it took me a while to figure out why we didn't do it in the first place, it was because not all users of test_utils linked against the library (some just included the header)

keithbostic:
Oh, sorry, @agorrod, I should have said something. Yeah, I went through the exact same exercise trying to figure that out.

PR #3086, opened 2016-10-09
WT-2955 Add statistics tracking the amount of time threads spend waiting for high level locks

keithbostic:


agorrod:
Thanks @keithbostic I like the shape of this. Did you do any performance testing? Did you play with splitting the statistics into internal vs application threads as well?

keithbostic:
I didn't do any real performance testing, just a couple of simple wtperf runs that finished in roughly the same amount of time as before (just checking in case I did something stupid).

In short, I believe we don't acquire a significant number of high-level locks in performance paths, plus @fedorova and @sueloverso measured clock calls in WT-2663 and didn't see problems.

I did not split the statistics into internal vs. application threads. I agree with your note in WT-2955 that it would lead to quite a few more statistics and I didn't think the additional information granularity was likely to be worth that cost. It would be an easy change to make, of course, happy to do it if you want.

sueloverso:
It might be worth adding tracking for the eviction related locks too.  They've been a contention point in the past.

agorrod:
Thanks @keithbostic. I generated a timeseries that includes the new stats, which made me think:

* it would be good if the stats are alphabetically grouped via lock type in the description so they sort together.
* I do think splitting internal/application session is worthwhile. Even in my basic example I found myself wondering whether threads waiting on the table lock were internal or application.

Following is a snapshot of the timeseries I ended up with - let's chat if you feel differently.

screen shot 2016-10-11 at 10 14 36 am

keithbostic:
@agorrod:

> Rename the lock statistics so they group together.
> Split lock wait times between internal and application threads.

done and pushed into the branch.

keithbostic:
@agorrod, I tucked a couple of additional changes into this branch:

* 41afee9 is just cleanup, putting the initialization of the connection's dummy session in one place. It would be good to merge that with generic session initialization, and I can do that if you want, but it's real work, and should probably have its own ticket.

* e43529b is a performance hack. The session's statististics bucket is fixed when its ID is selected, and we can cache a pointer to that bucket instead of doing repeated integer division every time we update a connection statistic. We could do the same hack for data-statistics, but because the session's data-handle comes and goes, it's trickier, I didn't make that change.

dist/stat_data.py
```diff
@@ -245,6 +264,25 @@ def __init__(self, name, desc, flags=''):
     DhandleStat('dh_sweeps', 'connection sweeps'),
 
     ##########################################
+    # Locking statistics
+    ##########################################
```
sueloverso: I'd still like to see eviction locks included.

src/include/mutex.h
```diff
@@ -74,6 +74,10 @@ struct __wt_rwlock {
 
 struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_spinlock {
 	volatile int lock;
+
+	int16_t stat_count;		/* acquisitions */
+	int16_t stat_app_usecs;		/* waiting application threads */
+	int16_t stat_int_usecs;		/* waiting server threads */
```
sueloverso: It took me a while to figure out these were the offset for the stat field rather than collecting the actual usecs value.  Can the comments and/or the fields add "offset of" or something similar?

src/include/mutex.i
```diff
@@ -255,3 +259,46 @@ __wt_spin_unlock(WT_SESSION_IMPL *session, WT_SPINLOCK *t)
 #error Unknown spinlock type
 
 #endif
+
+/*
+ * WT_SPIN_INIT_TRACKED --
+ *	Spinlock initialization, with tracking.
+ *
+ * Implemented as a macro so we can pass in a statistics field and convert
+ * it into a slot number.
```
sueloverso: I find the use of `slot` confusing here (and in the FIELD_TO_SLOT macro name).  I think of the `slot` as the hashed value you're now storing in `stat_bucket`.  The macro converts field name to stat structure offset.

src/include/session.h
```diff
@@ -147,6 +147,9 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_session_impl {
 	void	*reconcile;		/* Reconciliation support */
 	int	(*reconcile_cleanup)(WT_SESSION_IMPL *);
 
+	/* Sessions have an associated statistics bucket based on its ID. */
+	u_int	stat_bucket;		/* Statistics bucket offset */
```
sueloverso: This is a good change.

src/include/mutex.i
```diff
@@ -111,6 +112,9 @@ __wt_spin_init(WT_SESSION_IMPL *session, WT_SPINLOCK *t, const char *name)
 #endif
 
 	t->name = name;
+
+	t->stat_count = t->stat_app_usecs = t->stat_int_usecs = -1;
+
```
sueloverso: Personally, I prefer without all the whitespace here.

dist/stat_data.py
keithbostic: I don't have an opinion; I'm happy to do the work if @sueloverso and @agorrod identify which locks should be tracked.

sueloverso: I would suggest starting with cache->evict_queue_lock, cache->evict_walk_lock and queue->evict_lock.  We can run our eviction tests, look at stats and timeseries and reassess if we want to stop tracking any of those.

src/include/mutex.i
keithbostic: > Personally, I prefer without all the whitespace here.

Done and pushed.

keithbostic:
@sueloverso: after c51c8d8, changing the appropriate `__wt_spin_init` calls to `WT_SPIN_INIT_TRACKED` (in `__wt_cache_create())`, and adding the `dist/stat_data.py` fields, is probably all you need to do to add the new tracking you & I discussed.

src/include/mutex.h
```diff
@@ -74,6 +74,16 @@ struct __wt_rwlock {
...
 struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_spinlock {
 	volatile int lock;
+
+	/*
+	 * We track acquisitions and time spent waiting for some locks. For
+	 * performance reasons and to make it possible to write generic code
+	 * that tracks statistics for different locks, we store the offset
+	 * of the statistics fields to be updated during lock acquisition.
+	 */
+	int16_t stat_count;		/* acquisitions */
```
agorrod: It took me quite a while to figure out how this works, even with the comment. I spent a while thinking "why can these never overflow".. Would you mind renaming the fields to make it clear that they are offsets to the counter location?

src/include/mutex.h
```diff
@@ -83,7 +93,17 @@ struct WT_COMPILER_TYPE_ALIGN(WT_CACHE_LINE_ALIGNMENT) __wt_spinlock {
...
-	const char *name;		/* Statistics: mutex name */
+	const char *name;		/* Mutex name */
+
+	/*
+	 * We track acquisitions and time spent waiting for some locks. For
+	 * performance reasons and to make it possible to write generic code
+	 * that tracks statistics for different locks, we store the offset
+	 * of the statistics fields to be updated during lock acquisition.
+	 */
+	int16_t stat_count;		/* acquisitions */
```
agorrod: Since the stats content is duplicated across ``__wt_spinlock`` implementations anyway, would it maybe make sense to create a single:

```
struct __wt_spinlock_stat {
    WT_SPINLOCK l;
    /* Stat offset fields */
    int16_t stat_count;
    int16_t stat_app_usecs;
    int16_t stat_int_usecs;
};
```

Then __wt_spin_lock_track would be something like:

```
int
__wt_spin_lock_track(WT_SESSION *sessoin, WT_SPIN_LOCK *lock) {
    WT_SPIN_LOCK_STAT *lock_stat;

    lock_stat = (WT_SPIN_LOCK_STAT *)lock;
   ...
}
```

It's a little bit horrible - so I don't mind if you don't like it. The other advantage is that non-tracked spin locks don't have the additional fields in them either.

src/include/mutex.h
keithbostic: Sure, done & pushed.

src/include/mutex.h
keithbostic: @agorrod, I looked at something like this too, but decided against it. The code I came up with wasn't any better or smaller and I'd rather not have a new kind of spinlock (with new init/destroy functions and so on), appear in WiredTiger.

And, since spinlock structures are cache-line aligned, I don't think there's any space saving in the change, I think the compilers are creating (at a minimum), 64B structures.

agorrod:
Thanks. Please go ahead and merge

PR #3087, opened 2016-10-09
WT-2959 don't use WT_SESSION_IMPL before it's initialized

keithbostic:
@agorrod, I managed to get a core dump with an experimental change I was playing with, and the reason is we're using the WT_SESSION_IMPL we're initializing before we've finished initializing it.

Feel free to throw this away if you don't like it, it wouldn't have fired if I hadn't been modifying session initialization.


agorrod:
Thanks @keithbostic lgtm.

PR #3088, opened 2016-10-10
WT-2954 Try to evict pages that exceed memory_page_max even if the tr…

agorrod:
…ansaction ID isn't increasing

Workloads that insert large values checking only for the transaction ID
moving forward can lead to very large in-memory pages. Allow the page to exceed
the split size, but not the configured maximum.

Skipping such pages is a performance optimization - allow the optimization to
work as long as pages are between split size and maximum size.

Also don't require so many entries on a leaf page to trigger a split when
there are large values.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -328,8 +328,10 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_REF *ref)
 		return (false);
 
 	/*
-	 * If we have already tried and the transaction state has not moved on,
-	 * eviction is highly likely to fail.
+	 * Allow some leeway if the transaction ID isn't moving forward since
+	 * it is unlikely eviction will be able to evict the page. Don't keep
+	 * skipping the page indefinitely or large records can lead to
+	 * extremely large memory footprints.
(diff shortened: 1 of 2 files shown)


agorrod:
autobuild: retest this please

agorrod:
@michaelcahill Would you mind doing a review of this change please?

src/btree/bt_read.c
```diff
@@ -331,7 +331,8 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * If we have already tried and the transaction state has not moved on,
 	 * eviction is highly likely to fail.
 	 */
-	if (page->modify->last_eviction_id == __wt_txn_oldest_id(session))
+	if (page->memory_footprint < btree->maxmempage &&
+	    page->modify->last_eviction_id == __wt_txn_oldest_id(session))
```
michaelcahill: Can you please add a sentence to the comment above?  Something like "if the page grows above the maximum size, keep trying to split it or inserts of large records can lead to extremely large memory footprints".

PR #3089, opened 2016-10-10
WT-2961 Add support for custom version strings when generating docs.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - should I go ahead and merge?

michaelcahill:
Yes, please.  I'll need to do some manual work to update the old versions of docs, but that won't be committed.

My main concern about this change is that there is yet another place to maintain versions, and `src/docs/style/header-web.html` isn't somewhere we're all going to remember to look.  Maybe that file should be generated from a list in `dist`?

agorrod:
@daveh86 This change adds a version dropdown to our documentation, that means we have a section that records a set of versions. We need to remember to update that when cutting a release, the other place we do that is in ``src/docs/top/main.dox``, there is a script in ``dist/s_release_docs`` that attempts to automatically change the version strings in that ``main.dox`` file when cutting a new release.

Could you add a new section to ``s_release_docs`` that similarly updates the versions mentioned in the drop-down? The versions are (and will remain) different, since the dropdown mentions MongoDB versions, wheras ``main.dox`` is for WiredTiger versions.

michaelcahill:
This branch has been changed to use symbolic names for "current" versions of documentation.  I think that's good enough for now.  Merging...

PR #3091, opened 2016-10-12
WT-2965 Remove sleep and retry loop from __evict_exclusive

daveh86:


daveh86:
@michaelcahill, can you have a look?

michaelcahill:

PR #3093, opened 2016-10-14
WT-2971 Add details on raw-compression into WT documentation

daveh86:


src/docs/tune-compression.dox
```diff
@@ -19,6 +19,22 @@ An example of turning on row-store or column-store dictionary compression:
 
 @snippet ex_all.c Configure dictionary compression on
 
+@section raw_compression Raw Compression
```
agorrod: Thanks @daveh86. This is a good starting point, but I'd like to push it further and split file-format compression out into a more obvious section, and describe both the non-raw and raw compression approaches.

I'm a bit torn about whether all of the file format compression description should be in the file_formats_compression section or here.

agorrod:
@ddanderson Would you mind looking over the ``src/docs/tools`` part of this change that adds the ability to link from a language-versioned doc to a non-language versioned one.

daveh86:
I've tested locally and the @x_ref linkage worked for going from the "_lang_java" files into the non-language files worked well!

agorrod:
@sueloverso Would you mind reviewing this documentation change? The goal is to improve block-compression coverage in our docs.

sueloverso:
@daveh86 and @agorrod I pushed some reorg changes and some wordsmithing and typos.  I put `noraw` first since that is the more common.  I changed the use of `documents` and `rows` to just say `data` since both page blocks and log records are compressed.  I wanted it more general.  I tried to use `noraw` consistently (i.e. without a - in the middle of it)

One thing I'd like to see added is a sentence about when you want/don't want to use raw compression.  It's always been a special case for me.  Something that says "Most typical users want to use noraw compression.  Only in <special case> would raw compression make sense."

michaelcahill:
@daveh86 is my last edit okay with you?

daveh86:
Yes, LGTM. And thanks.

michaelcahill:

PR #3094, opened 2016-10-14
WT-2954 Enable fast appends with small in-memory pages.

michaelcahill:


agorrod:
This change in it's current form introduces a regression in update-btree workload on Jenkins:
http://build.wiredtiger.com:8080/job/wiredtiger-perf-btree/plot/

agorrod:
The regression in update-btree appears to be because aggressive is being set more often for the workload on the branch than develop. Aggressive is being increased because of dirty content. Will figure that out tomorrow, since it's counter to my intuition with these changes. Interestingly the workload is using the in-memory restoration path.

michaelcahill:
@agorrod, I think this change is ready to go now.  I've switched our perf testing over and done a new Evergreen sys-perf patch build.  Can you please review?

agorrod:
Thanks @michaelcahill these changes look good to me now. I'll watch the perf tests and merge if they look good, or chase any regressions. I'll also re-run the update-btree workload standalone against the new version to see if performance is improved.

src/evict/evict_lru.c
```diff
@@ -907,24 +907,7 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
...
-	 */
-	if (ret == WT_NOTFOUND) {
-		if (F_ISSET(cache,
-		    WT_CACHE_EVICT_CLEAN_HARD | WT_CACHE_EVICT_DIRTY_HARD))
-			cache->evict_aggressive_score = WT_MIN(
-			    cache->evict_aggressive_score + WT_EVICT_SCORE_BUMP,
-			    WT_EVICT_SCORE_MAX);
-		WT_STAT_CONN_SET(session, cache_eviction_aggressive_set,
-		    cache->evict_aggressive_score);
-	}
```
agorrod: @michaelcahill I didn't end up merging this branch today (and therefore didn't update MongoDB master). The reason I didn't is because I couldn't get the performance of update-btree back without removing this change - i.e: the change to no longer return WT_NOTFOUND from ``__evict_walk``. I tested the various cases I have around related to WT-2954, and none of them relied on this part of the pull request - I'm not sure what case it improved.

I also couldn't get to the bottom of why adding it back in helps - generally switching into aggressive mode hurts the update-btree workload, and this change makes that happen less often.

michaelcahill: I'm happy for that part of the change to be reverted.  Early on, some of the wtperf workloads were much slower when we got aggressive, which motivated this change.  It may not be necessary at all given the other changes in this branch, and I'd prefer to get the fix in asap.

I can't do the revert now: Monday is the earliest I'll get to it.

agorrod: I just reverted that part of the change - I'll watch the Jenkins performance builds

agorrod:
Perf looks good on this branch now, so I'm going to merge.

PR #3095, opened 2016-10-14
WT-2974 lint

keithbostic:
Ready for review/merge.


keithbostic:
@agorrod, there was a value calculated in `__evict_stat_walk()`, `bytes_inmem`, that was never used. I removed it, but just in case the error was not using it, I wanted to flag it for your attention.

agorrod:
Thanks @keithbostic

PR #3096, opened 2016-10-16
WT-2976 Add a statistic tracking how long app threads spend reading

agorrod:


agorrod:
@sueloverso Would you please review this change?

agorrod:
FWIW: I don't like how the statistics naming groups the new stats, but I struggled to word the descriptions in a way that made sense while keeping order. Feel free to make changes to the wording.

sueloverso:
@agorrod I did rework the wording so that they will group together.  See if you like it.

agorrod:
Thanks @sueloverso I did a little more word shuffling, but preserved the order. I'll go ahead and merge.

PR #3098, opened 2016-10-17
WT-2945 Occasional hang running reconfigure fuzz test

keithbostic:

### DIFF
--- a/src/lsm/lsm_manager.c
+++ b/src/lsm/lsm_manager.c
@@ -55,6 +55,7 @@ __lsm_general_worker_start(WT_SESSION_IMPL *session)
 	 * as many worker threads as are required to keep up with demand.
 	 */
 	WT_ASSERT(session, manager->lsm_workers > 0);
+	WT_ASSERT(session, manager->lsm_workers < manager->lsm_workers_max);
 	for (; manager->lsm_workers < manager->lsm_workers_max;
 	    manager->lsm_workers++) {
 		worker_args =
@@ -116,17 +117,18 @@ __lsm_stop_workers(WT_SESSION_IMPL *session)
 {
 	WT_LSM_MANAGER *manager;
 	WT_LSM_WORKER_ARGS *worker_args;
-	uint32_t i;
 
 	manager = &S2C(session)->lsm_manager;
 	/*
-	 * Start at the end of the list of threads and stop them until we
-	 * have the desired number.  We want to keep all active threads
-	 * packed at the front of the worker array.
+	 * Start at the end of the list of threads and stop them until we have
+	 * the desired number. We want to keep all active threads packed at the
+	 * front of the worker array.
 	 */
-	WT_ASSERT(session, manager->lsm_workers != 0);
-	for (i = manager->lsm_workers - 1; i >= manager->lsm_workers_max; i--) {
-		worker_args = &manager->lsm_worker_cookies[i];
+	WT_ASSERT(session, manager->lsm_workers > manager->lsm_workers_max);
+	for (; manager->lsm_workers > manager->lsm_workers_max;
+	    manager->lsm_workers--) {
+		worker_args =
+		    &manager->lsm_worker_cookies[manager->lsm_workers - 1];
 		/*
 		 * Clear this worker's flag so it stops.
 		 */
@@ -136,7 +138,6 @@ __lsm_stop_workers(WT_SESSION_IMPL *session)
 		worker_args->tid = 0;
 		worker_args->type = 0;
 		worker_args->flags = 0;
-		manager->lsm_workers--;
 		/*
 		 * We do not clear the session because they are allocated
 		 * statically when the connection was opened.
--- a/src/os_posix/os_mtx_cond.c
+++ b/src/os_posix/os_mtx_cond.c
@@ -107,8 +107,14 @@ __wt_cond_signal(WT_SESSION_IMPL *session, WT_CONDVAR *cond)
 	__wt_verbose(session, WT_VERB_MUTEX, "signal %s", cond->name);
 
 	/*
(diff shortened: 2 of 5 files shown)


keithbostic:
@agorrod, the fact the start function is named `__lsm_general_worker_start` and the stop function is named `__lsm_stop_workers` confused me, why are they different?

Also, the start function asserts `WT_ASSERT(session, manager->lsm_workers > 0);`, while the stop function asserts `WT_ASSERT(session, manager->lsm_workers > manager->lsm_workers_max);`, that also looked odd.

src/os_posix/os_mtx_cond.c
```diff
@@ -103,27 +103,30 @@ void
...
 	__wt_verbose(session, WT_VERB_MUTEX, "signal %s", cond->name);
 
-	/* Fast path if already signalled. */
-	if (cond->waiters == -1)
-		return;
+	/*
+	 * Our callers are often setting flags to cause a thread to exit. Add
+	 * a barrier to ensure the flags are seen by the threads.
+	 */
+	WT_WRITE_BARRIER();
```
agorrod: I'd like to understand the performance consequences of this change. It would be a bit horrible to have barrier and non-barrier versions of this function, but it may be worthwhile. Could you run some performance tests to see if adding a barrier reduces throughput?

src/lsm/lsm_manager.c
```diff
@@ -116,17 +117,18 @@ __lsm_stop_workers(WT_SESSION_IMPL *session)
...
-	 * have the desired number.  We want to keep all active threads
-	 * packed at the front of the worker array.
+	 * Start at the end of the list of threads and stop them until we have
+	 * the desired number. We want to keep all active threads packed at the
+	 * front of the worker array.
 	 */
-	WT_ASSERT(session, manager->lsm_workers != 0);
-	for (i = manager->lsm_workers - 1; i >= manager->lsm_workers_max; i--) {
-		worker_args = &manager->lsm_worker_cookies[i];
+	WT_ASSERT(session, manager->lsm_workers > manager->lsm_workers_max);
```
agorrod: This assertion doesn't feel right to me, I think this function could be called with less than max workers and would end up as a no-op.

src/lsm/lsm_manager.c
```diff
@@ -116,17 +117,18 @@ __lsm_stop_workers(WT_SESSION_IMPL *session)
...
+	for (; manager->lsm_workers > manager->lsm_workers_max;
+	    manager->lsm_workers--) {
```
agorrod: The decrement was outside of the loop control originally because I was being paranoid about decrementing more than expected on error. Your change looks right, but I thought I'd point it out.

src/lsm/lsm_manager.c
keithbostic: `__lsm_stop_workers()` is only called from `__wt_lsm_manager_reconfig()`, where we've already asserted `manager->lsm_workers_max < the previous manager->lsm_workers_max`.

I only changed the assert because checking `manager->lsm_workers != 0` didn't seem to add anything at all, and the assert I added matched the assert in `__lsm_general_worker_start()`.

I'm happy to remove it if you like.

src/lsm/lsm_manager.c
keithbostic: This is another change I made for clarity (I like it when the add and the decrement loops match, I think it makes it more likely they're correct).

You've done a lot more work in LSM than I have, I'm happy to be guided by your intuition here, feel free to simply discard the changes I made in LSM, it's commit f84f8be.

src/os_posix/os_mtx_cond.c
keithbostic: I ran `medium-multi-btree-log.wtperf` and `evict-btree-stress-multi.wtperf` (since eviction and logging are heavy consumers of `__wt_cond_signal())`, and they're about the same as before. Let me know if you have other workloads you think might exercise this change.

keithbostic:
@agorrod, @michaelcahill, tl;dr

I think this branch is ready to merge, Alex may want additional performance tests, and may want to revert one LSM-related commit (but that commit is standalone and shouldn't require any additional testing).

src/os_posix/os_mtx_cond.c
```diff
@@ -103,27 +103,30 @@ void
...
-		WT_ERR(pthread_mutex_lock(&cond->mtx));
-		locked = true;
-		WT_ERR(pthread_cond_broadcast(&cond->cond));
-	}
+	/*
+	 * Fast path if we are in (or can enter), a state where the next waiter
+	 * will return immediately as already signaled.
+	 */
+	if (cond->waiters == -1 ||
+	    (cond->waiters == 0 && __wt_atomic_casi32(&cond->waiters, 0, -1)))
```
keithbostic: The hang happens rarely enough we won't know until the branch has been running awhile. I think this branch fixes a bug (flags to cause the server thread to exit not getting flushed), and makes things better/cleaner, but I'm not sure it fixes the hang.

I don't think that reading `cond->waiters` multiple times is the problem: if the first read matches -1 then we must have read `waiters` before any threads waiting on the mutex, and the next thread attempting a wait will return immediately; if the first read isn't -1 and the second read matches 0, then we atomically swap `waiters` to -1: any thread trying to wait atomically increments `waiters` and so either that thread wins and our swap must fail and we'll proceed to wake the thread up, or we win and our swap to -1 followed by that thread's increment will force that thread to return immediately.

I *think*. :)

I'd be fine with turning that code into a single read, but I don't see a bug there.

agorrod:
Thanks @keithbostic I'm good to merge this, but I want to flush out the performance questions in WT-2954 first, so I'm holding off for now.

michaelcahill:
I'll merge and we'll see whether the original issue happens again...

PR #3099, opened 2016-10-18
WT-2415 Add option for joins to return false positives from the Bloom filter.

ddanderson:


ddanderson:
@michaelcahill, can you please review this?

agorrod:

PR #3100, opened 2016-10-19
WT-2985 checkpoint core dump

keithbostic:


michaelcahill:
@keithbostic, sorry for the radio silence -- I've had back to back meetings for several days.

The latest version of this is what I was expecting: checkpoint should visit these pages, I think.  That test is intended to skip simple pages that were clean at the beginning of the checkpoint (or on disk) and become dirty while the checkpoint is running.  It should not be a common enough case for pages to be this state after a split has failed that there's any measurable performance penalty for visiting them.

src/btree/bt_sync.c
```diff
@@ -9,6 +9,66 @@
...
+		    i = 0; i < mod->mod_multi_entries; ++multi, ++i)
+			if (multi->addr.addr == NULL)
+				return (false);
+
+	/*
+	 * Mark the tree dirty if skipping a page, the checkpoint marked it
+	 * clean and we can't skip future checkpoints until this page is
+	 * written.
+	 */
+	__wt_tree_modify_set(session);
```
michaelcahill: Would you mind pull this up into the caller so the inlined function has no side effects?

keithbostic:
@michaelcahill, this last set of changes is passing all of my tests, so if you're happy with it, it's ready to merge.

The reason I resisted the second approach was I don't like the checkpoint sync code looking inside the page modify structure, that's relatively ugly (not that the additional code in reconciliation was all that pretty, but at least it was in reconciliation).

Anyway, the only other idea I had was to add a new "multiblock with failed addresses" flag to the page modify structure, and that's not much better.

src/btree/bt_sync.c
keithbostic: Yes, you're right, that's better.

michaelcahill:

PR #3101, opened 2016-10-20
WT-2968 Don't open file handles on backup.

sueloverso:
@agorrod Take a look at this branch.  It evolved as I went along and after chatting with @keithbostic we realized we should be able to protect schema operations adequately and we don't need the open handles in backup.


agorrod:
Thanks @sueloverso the change generally makes sense to me. I'm a bit apprehensive about checking a list of names read from a file, especially if that list could be 100,000 long.

The other concern I have is that schema operations will block waiting for the hot backup lock (which I presume is held until the hot backup file has finished being created).

i.e: I have a few questions about performance - I'd like to talk them through next time we are online at the same time.

src/include/cursor.h
```diff
@@ -61,7 +61,6 @@
 
 struct __wt_cursor_backup_entry {
 	char *name;			/* File name */
-	WT_DATA_HANDLE *handle;		/* Handle */
 };
```
keithbostic: Should we get rid of `__wt_cursor_backup_entry` and replace it with `char *[]`?

src/include/schema.h
```diff
@@ -79,6 +79,27 @@ struct __wt_table {
...
+		    uri, &in_backup)) == 0 && in_backup) {		\
+			__wt_readunlock(session,			\
+			    S2C(session)->hot_backup_lock);		\
+			return (EBUSY);					\
+		}							\
+	} else								\
+		__wt_readunlock(session,				\
+		    S2C(session)->hot_backup_lock);			\
+} while (0)
+/*
```
keithbostic: * Why not merge this #define into (a possibly renamed) `__wt_metadata_walk_backup()` function? I can't think of a reason this needs to be a #define.

* The value of `__ret` is being ignored, I think this needs a `WT_RET`.

* Is there any reason to acquire the read-lock before checking the `hot_backup` value? It seems to me that if `hot_backup` is 0, we have no reason to do the lock/unlock, and that's likely to be true most of the time.

src/meta/meta_turtle.c
```diff
@@ -54,6 +54,47 @@ __metadata_init(WT_SESSION_IMPL *session)
...
+	WT_DECL_ITEM(value);
+	WT_DECL_RET;
+	WT_FSTREAM *fs;
+	bool exist;
+
+	/* Look for a hot backup file. We expect to have one. */
+	*in_backup = false;
+	WT_RET(__wt_fs_exist(session, WT_METADATA_BACKUP, &exist));
+	WT_ASSERT(session, exist);
+	WT_RET(__wt_fopen(session, WT_METADATA_BACKUP, 0, WT_STREAM_READ, &fs));
```
keithbostic: Since `__wt_fopen` should fail with a non-existent file, is there any reason to have the `__wt_fs_exist` call?

src/include/schema.h
sueloverso: Solved all of these by taking the first suggestion.

src/schema/schema_util.c
```diff
@@ -9,6 +9,52 @@
...
+		__wt_readunlock(session, conn->hot_backup_lock);
+	}
+err:	if (fs != NULL) {
+		WT_TRET(__wt_fclose(session, &fs));
+		__wt_scr_free(session, &key);
+		__wt_scr_free(session, &value);
+	}
+	return (ret);
+}
+/*
```
sueloverso: Yes.  I like that.  I'll fix it up (and the whitespace) and push that.

sueloverso:
@agorrod and @keithbostic Take another look.  I moved to walking the in-memory list from the cursor instead of opening and reading the on-disk file.  I had to move the location of the schema check call down into the `file` schema functions since that is what is in the cursor list.

src/cursor/cur_backup.c
```diff
@@ -229,6 +229,7 @@ __backup_start(
 	 */
 	__wt_writelock(session, conn->hot_backup_lock);
 	conn->hot_backup = true;
+	conn->hot_backup_list = NULL;
 	__wt_writeunlock(session, conn->hot_backup_lock);
```
michaelcahill: @sueloverso, can you please update the comment above to explain that we're relying on the schema lock to prevent `drop` operations while building the list?

keithbostic:
lgtm.

sueloverso:
autobuild: retest this please

PR #3102, opened 2016-10-21
WT-2904 reported checkpoint size many times data size

keithbostic:
We weren't adjusting the rolling checkpoint size for the size of the root page.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -615,8 +615,6 @@ live_update:
 	WT_CKPT_FOREACH(ckptbase, ckpt)
 		if (F_ISSET(ckpt, WT_CKPT_ADD)) {
 			/*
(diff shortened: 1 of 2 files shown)


michaelcahill:
@keithbostic, thanks, I think I understand this change.  Is there a sanity checking assertion we can make about `ci->size` being no larger than the file size?

keithbostic:
> Is there a sanity checking assertion we can make about `ci->size` being no larger than the file size?

Yes, we could assert that (and at run-time force it to be true).

keithbostic:
> Is there a sanity checking assertion we can make about `ci->size` being no larger than the file size?

Done and pushed in commit 7cdf96c.

agorrod:
Thanks @keithbostic lgtm.

PR #3103, opened 2016-10-21
WT-2987 opening a cursor on an incomplete table drops core

keithbostic:
Ready for review/merge.


michaelcahill:
Thanks @keithbostic, this change looks fine.  Can you please translate the original test case into Python?

keithbostic:
>  Can you please translate the original test case into Python?

Sure, pushed in commit c1b9008.

agorrod:
Thanks @keithbostic lgtm.

PR #3107, opened 2016-10-24
WT-2975 Reduce size of statistics in dhandle structure

keithbostic:


michaelcahill:
@keithbostic, there are definitely debugging advantages to having the stats in a separate allocation.

If we could defer the allocation until `WT_DHANDLE_OPEN` is set, then we could save this memory in many common cases (e.g., the handles used during checkpoints to make sure it is safe to drop old checkpoints).  Can you take a look and see if that is feasible?  I don't know of any places where it would make sense to keep statistics on a handle that isn't open, but I won't be shocked if you tell me they exist.

keithbostic:
@michaelcahill 

> If we could defer the allocation until `WT_DHANDLE_OPEN` is set

I agree that makes sense, it's a good idea.

My concern is probably the same as yours, we might drop core because we don't run some code path very often and there's a path where we modify a statistic before `WT_DHANDLE_OPEN` is set.

That said, I can't find one, so I think it's safe.

The safer alternative would be to change the data-source macros to check explicitly:
```
diff --git a/src/include/stat.h b/src/include/stat.h
index d0b0b60..5fceae1 100644
--- a/src/include/stat.h
+++ b/src/include/stat.h
@@ -232,7 +232,8 @@ __wt_stats_clear(void *stats_arg, int slot)
 #define        WT_STAT_DATA_DECR(session, fld)                                 \
        WT_STAT_DATA_DECRV(session, fld, 1)
 #define        WT_STAT_DATA_INCRV(session, fld, value) do {                    \
-       if ((session)->dhandle != NULL)                                 \
+       if ((session)->dhandle != NULL &&                               \
+           (session)->dhandle->stat_array != NULL)                     \
                WT_STAT_INCRV(                                          \
                    session, (session)->dhandle->stats, fld, value);    \
 } while (0)
```

I'd rather not add another memory access to the data-source statistics macros (but that may be a foolish concern). What do you think?

sueloverso:
I think the one additional memory access and cmp is not something to be concerned about.  I don't see that being the heavy piece of incrementing a stat.  What about an assertion since we don't expect any to be there?  It would find those with diagnostic but disappear when in production code.

michaelcahill:
I agree with @sueloverso: let's play it safe and always check `stat_array != NULL`.  We're about to use it to calculate an offset, I'll be surprised if we can measure the difference.

keithbostic:
@michaelcahill, I think this one is ready to merge and WT-2975 ready to close, we've gone as far in this direction as we're going, for now.

src/conn/conn_dhandle.c
```diff
@@ -332,6 +330,16 @@ __wt_conn_btree_open(
...
 	F_SET(btree, LF_MASK(WT_BTREE_SPECIAL_FLAGS));
 
+	/*
+	 * Allocate data-source statistics memory. We don't allocate that memory
+	 * when allocating the data-handle because not all data handles need
+	 * statistics (for example, handles used for checkpoint locking). If we
+	 * are reopening the handle, then it may already have statistics memory,
+	 * check to avoid the leak.
+	 */
+	if (!F_ISSET(dhandle, WT_DHANDLE_OPEN))
```
michaelcahill: It probably isn't necessary to check the `WT_DHANDLE_OPEN` flag here, or at least something is wrong if we get here and the handle is already open.  That said, I don't mind it for clarity.

src/conn/conn_dhandle.c
```diff
@@ -554,8 +562,7 @@ __wt_conn_dhandle_discard_single(
 
 	dhandle = session->dhandle;
 
-	if (F_ISSET(dhandle, WT_DHANDLE_OPEN) ||
-	    (final && F_ISSET(dhandle, WT_DHANDLE_DEAD))) {
```
michaelcahill: Why was this changed?  It looked like it used to deal with discarding dead handles: how is that connected with a change to how statistics are allocated?

michaelcahill:
LGTM, I'll merge.

PR #3110, opened 2016-10-29
WT-2998 add error messages to error returns that might be confusing.

keithbostic:
Add error messages to error returns that might be confusing.

### DIFF
--- a/src/async/async_api.c
+++ b/src/async/async_api.c
@@ -232,7 +232,7 @@ __async_start(WT_SESSION_IMPL *session)
 	uint32_t i, session_flags;
 
 	conn = S2C(session);
-	conn->async_cfg = 1;
+	conn->async_cfg = true;
 	/*
 	 * Async is on, allocate the WT_ASYNC structure and initialize the ops.
 	 */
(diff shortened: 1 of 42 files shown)


keithbostic:
@michaelcahill, I went looking for HELP-2962, where we apparently returned `EINVAL` somewhere in the recovery path. I couldn't find it, but I did a pretty systematic review of everywhere we return a naked value without some kind of associated error message, and this branch is the result.

If you think this is worth merging, then it needs a proper review, there are a few real changes.

I have a few open questions that could suggest further changes for this branch, basically, all places I didn't change because I didn't understand them well enough (or my attempt to change them broke something).

@michaelcahill: this code in `__find_column_format()` looks like it could benefit from a verbose message:
```
if (k.len == colname->len &&
    strncmp(colname->str, k.str, k.len) == 0) {
        if (value_only && inkey)
                return (EINVAL);
        return (0);
}
```

@sueloverso: I had to back out 0b20e1a because it broke `test/test_async01.py`. Thinking it over, it seemed reasonable an application wouldn't want verbose complaints every time it exceeds the async handle limit, but I mention it to you in case there's something messier going on, I was surprised that test hit that error message.

@ddanderson: the `__wt_json_token()` error handling is a little strange. All but one error breaks out (or falls out), of the switch statement, setting the returned token type to -1 and returning `EINVAL`. The exception simply returned `-1`, not a valid error. I changed that place to return `EINVAL` instead of `-1`, but I was concerned a caller might expect `*toktype` to be set on error return and that was incorrect. I would suggest switching all of the errors to immediately return instead of falling out of the switch statement and depending on result not having been set to something other than `-1`: that seems cleaner to me and I think it makes it less likely we'll get into trouble in the future.

@michaelcahill: any reason `WT_CURSOR_INDEX.set_value` (that's the `__curindex_set_value()` function) can't just be set to `__wt_cursor_set_value_notsup()`? It calls `JOINABLE_CURSOR_API_CALL` and clears `WT_CURSTD_VALUE_SET`, which made me nervous.

keithbostic:
@sueloverso, I've pushed the changes to add verbose errors anywhere recovery fails to unpack a log record, in df8af00. Would you please take a look?

sueloverso:
Regarding the async test, it is intentional that the test sets the number of available operation handles to half the number of entries we want to insert/search.  We want to test the condition of running out of handles.

src/async/async_api.c
```diff
@@ -144,7 +144,7 @@ __async_new_op_alloc(WT_SESSION_IMPL *session, const char *uri,
 	 */
 	if (op == NULL || op->state != WT_ASYNCOP_FREE) {
 		WT_STAT_CONN_INCR(session, async_full);
-		WT_RET(EBUSY);
+		return (EBUSY);
```
sueloverso: I think a comment saying why we have a naked error return here would be helpful for the future so we know it was intentional.

michaelcahill:

PR #3111, opened 2016-11-01
WT-2953 During reconciliation only check for checkpoint+LAS collision…

daveh86:
… if we have used the LAS table.


michaelcahill:
Thanks, @daveh86, lgtm.  I'll merge.

PR #3112, opened 2016-11-01
WT-2964 Alter evict walk to not fill all its slots with internal page…

daveh86:
…s when running in aggressive mode

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1453,14 +1453,14 @@ __evict_walk_file(WT_SESSION_IMPL *session, WT_EVICT_QUEUE *queue,
 		if (modified && !F_ISSET(cache, WT_CACHE_EVICT_DIRTY))
 			continue;
 
-		/* If eviction gets aggressive, anything else is fair game. */
-		if (__wt_cache_aggressive(session))
-			goto fast;
-
 		/* Limit internal pages to 50% of the total. */
 		if (WT_PAGE_IS_INTERNAL(page) &&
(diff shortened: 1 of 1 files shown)


src/evict/evict_lru.c
```diff
@@ -1453,15 +1453,15 @@ __evict_walk_file(WT_SESSION_IMPL *session, WT_EVICT_QUEUE *queue,
 		if (modified && !F_ISSET(cache, WT_CACHE_EVICT_DIRTY))
 			continue;
 
-		/* If eviction gets aggressive, anything else is fair game. */
-		if (__wt_cache_aggressive(session))
-			goto fast;
-
 		/* Limit internal pages to 50% of the total. */
 		if (WT_PAGE_IS_INTERNAL(page) &&
 		    internal_pages >= (int)(evict - start) / 2)
```
michaelcahill: @daveh86, unrelated to your change, this test has bugged me for a while.  Can we try changing `>=` to `>`?

In particular, I'm concerned that if a tree goes idle, we will evict all the leaf pages from cache, then never consider any of the internal pages (because we have to find one leaf page before we will queue the first internal page).

We've tried changing this before: I don't think the results were conclusive.

daveh86: @michaelcahill, that change makes sense. I will make it in a moment and re-execute my reproducer to confirm it doesn't have a negative impact there.

michaelcahill:
Thanks @daveh86, I'll go ahead and merge this, then look at the changes in #3123.

PR #3113, opened 2016-11-01
SERVER-26753 Don't spin on a read-lock in a tight loop.

michaelcahill:
We could be starving a thread that we are waiting on of CPU.


michaelcahill:
Note that this reverts one part of WT-2560, which was done to avoid performance regressions in automated tests.  We'll need to run a full round of performance tests on this branch to measure the impact of this change.

michaelcahill:
@daveh86 can you please review?

daveh86:
LGTM

PR #3114, opened 2016-11-01
WT-2975 Fix a leak of statistics data when reopening handles.

michaelcahill:


michaelcahill:
autobuild: retest this please

michaelcahill:
@keithbostic, we're seeing leaks since WT-2975 was merged.  There also seems to be some general instability, with segfaults and hangs, but that could be from one of the other commits today.

I believe this branch fixes the leaks (at least the ones I've reproduced).  Can you please review, and see if there are other places we need to worry about?

keithbostic:
lgtm.

PR #3115, opened 2016-11-01
WT-3001 WT_EXTENSION_API references are named inconsistently.

keithbostic:
@michaelcahill, I noticed we were naming `WT_EXTENSION_API` objects inconsistently when I was working on the zstd changes (probably because you and I each wrote a chunk). The core code consistently uses `wt_api`, so I switched the extension code to match.

There's no other change in this branch.


michaelcahill:

PR #3116, opened 2016-11-01
WT-3002 Allow applications to exempt threads from eviction.

keithbostic:
@michaelcahill, Bruce Lucas & I were discussing how we lost ftdc data if the ftdc data thread was picked to help with eviction.

Would it be reasonable to offer applications a configuration to designate high-latency threads exempt from WiredTiger tasks?


keithbostic:
@michaelcahill, on second thought this branch introduces a bug.

The code in `__session_reconfigure()` doesn't start with the current session configuration, and so setting one of the reconfiguration values unintentionally resets the other reconfiguration values, that is:
```
session.reconfigure(session, "exempt=true,isolation=snapshot");
session.reconfigure(session, "exempt=false");        /* XXX: resets isolation too */
```
also resets the isolation value because the reconfiguration code starts from the base API configuration, not the current configured state.

See `__conn_reconfigure` for the fix, you have to save a copy of the "current" configured values after each configuration, and use that as the starting point instead of the base values.

This works at the moment because there's only a single possible session reconfiguration value, as soon as there's more than one configuration value, it breaks.

If you decide you want this branch, I'll do that work.

Otherwise, we should add a comment to the `__session_reconfigure()` code so when we add a second configuration value, we don't get surprised.

michaelcahill:
The session config key should be called something like `ignore_cache_size` to make it clear what this option does.

I'd suggest making this a `WT_CONNECTION::open_session` option only for now and avoiding `WT_SESSION::reconfigure` (though a comment is obviously fine).

That said, I'm not sure how useful this is, tbh -- in many cases, the FTDC thread ends up blocked on something else (e.g., getting a read ticket), so this wouldn't be sufficient to ensure that the FTDC thread always makes progress.

keithbostic:
@michaelcahill, I deliberately chose a less specific name, so that it would be reasonable for the option to ignore other housekeeping tasks (should any be added in the future).

I'm good with just discarding this branch: if it's not likely to improve MongoDB statistics gathering (or if I'm solving a non-problem), let's just kill it.

bdlucas1:
+Mark Benvenuto

Agreed that it doesn't guarantee FTDC won't block somewhere else. However,
at this point we don't have any ideas for a silver bullet to prevent that
and are stuck playing whack-a-mole.

That said, this will only be one piece of the puzzle for this particular
mole; doing the plumbing in the integration layer to make use of this may
be non-trivial, given session caching.

keithbostic:
> That said, this will only be one piece of the puzzle for this particular mole; doing the plumbing in the integration layer to make use of this may be non-trivial, given session caching.

Understood. That's why I suggested `WT_SESSION.reconfigure`, to make it easy to turn it on/off if given a random session to work with.

michaelcahill:
@keithbostic, I've pushed a change that makes `WT_SESSION::reconfigure` work and changes the config key as I suggested above. Please take a look.

If there are other settings we want to make configurable in the future, let's revisit then. I don't want to export a "go faster" switch and then deal with people being upset that there are consequences.

keithbostic:
lgtm, merged.

PR #3117, opened 2016-11-01
WT-2999 free memory related to cursor used with custom extractors during joins.

ddanderson:


ddanderson:
retest this please

ddanderson:
@michaelcahill, could you please review this?  It's a small change.

michaelcahill:
Thanks @ddanderson, lgtm.  I'll merge so that this is fixed, but should we add a test case involving extractors to avoid a regression?

PR #3119, opened 2016-11-02
WT-3004 lint: declare functions that don't return a value as void

keithbostic:
Ready for review/merge.


michaelcahill:
@keithbostic, I guess this is fine, what tool do you use to detect cases like this?

It's entirely possible that some future change will reintroduce an error condition into `__wt_txn_get_snapshot` but it happens that `__wt_try_readlock` was the only source of errors.
If we're going to enforce a rule that as soon as a function is known to return zero it must be changed to a `void` return, we need to have a tool that enforces that as part of the regular commit checking process.

keithbostic:
@michaelcahill:

> what tool do you use to detect cases like this?

The `dist/s_void` script, written for WT-2888.

I added it to the tree but didn't add it to `s_all`, I thought it might report too many false positives. So far it's been pretty quiet, we could try running it all the time. I don't care one way or the other.

>  entirely possible that some future change will reintroduce an error condition into `__wt_txn_get_snapshot`

Sure, but that's true in all cases?

I think the change is worthwhile: it removes branching from heavily used code as well as simplifying the code.

I understand your point: it's more error prone to remove error handling and then re-introduce it a month later, but I can't think of any way to reliably delay cleanup until after error handling is unlikely to be re-introduced.

michaelcahill:
@keithbostic, I don't buy the performance argument -- with link-time optimization, checking for a non-zero return from a function that returns a constant zero should be optimized away.

That said, I think it's reasonable to make a blanket rule that we won't have any functions that just return zero.  If we're going to do that, it should be enforced on each commit by running `s_void` as part of `s_all`.  Can you please make this change and look out for false positives?

keithbostic:
@michaelcahill, does anybody do LTO with shared libraries?

I've been running `s_void` for a month or so now, and there haven't been any false positives. I'll push that change, and if it turns out to be noisy we can pull it out.

PR #3120, opened 2016-11-02
WT-3005 Add top-level .gitignore file.

keithbostic:
I don't (yet) use a `.gitignore` file in my workflow and someone asked on wiredtiger-users, so I played with it a bit out of curiosity.

Feel free to discard without a second thought if nobody wants it.


sueloverso:
I have not used `.gitignore` in my workflow either.  The man page says it is used to determine what files to ignore for a commit.  Git already seems to do that fine with the untracked files (as shown when committing or via `git status`).  How is this different than the untracked and uncommitted files shown there?

michaelcahill:
@sueloverso, the big advantage I see is for `git status`.

@keithbostic, thanks, lgtm, I'll merge.

PR #3121, opened 2016-11-02
WT-3003 Don't generate log record and op types.  They must be static.

sueloverso:
@keithbostic and @michaelcahill  Please review.  I added comments in log_data.py to remind people to add them to the include file.  Is it clear and visible enough?  Is there an easy way to verify they did add the `#define` in one of the`s_` scripts?

### DIFF
--- a/dist/log.py
+++ b/dist/log.py
@@ -98,34 +98,6 @@ def printf_line(f, optype, i, ishex):
     return ifbegin + body + ifend
 
 #####################################################################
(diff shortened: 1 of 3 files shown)


keithbostic:
WiredTiger won't compile if they don't add the `#define`, that seems like enough of a check.

lgtm.

michaelcahill:
Thanks: I don't feel strongly about this but the change looks fine.  I'll merge.

PR #3123, opened 2016-11-03
WT-2964 Walk for longer when only looking for dirty pages

michaelcahill:


michaelcahill:
@daveh86, I think this branch is ready to go: it's looking no worse than develop in automated testing.  Can you please review?

daveh86:
This LGTM

PR #3125, opened 2016-11-03
SERVER-26889 Got signal: 11 (Segmentation fault) in 3.0.13

keithbostic:
Fix 552a33b (cherry-picked from 521270d). When the commit was merged, a line was dropped.

@michaelcahill, for your review: you can see the difference if you compare 521270d and 552a33b, the line `ref->addr = addr;` should have been removed and wasn't. I think the result is `ref->addr` can potentially point to freed memory.


michaelcahill:
Thanks @keithbostic, lgtm -- I'll merge (and kick off MongoDB patch builds).

PR #3126, opened 2016-11-04
SERVER-26753 Don't spin indefinitely trying to get a read lock.

michaelcahill:
Even if it appears to be available, we're burning CPU.


daveh86:
The final change adding exclusive has had resolved the perf problems seen in YCSB testing.

daveh86:
This LGTM

michaelcahill:
Thanks @daveh86, I'll merge and we'll see what happens with our perf tests.  I'll also kick off new MongoDB tests.

PR #3127, opened 2016-11-06
WT-2968 Fix a segfault when a drop races with closing a backup cursor.

michaelcahill:
We were clearing the shared pointer to the list of filenames in the
backup, then releasing the hotbackup lock before clearing the hotbackup
flag.  That means drops need to check both the flag and that the pointer
is non-NULL.


michaelcahill:
@sulabhM, can you please review this?  It's holding up a bump of MongoDB master and v3.2...

PR #3131, opened 2016-11-06
WT-3008 Move wtperf stress jobs to new stress runner folder

daveh86:


michaelcahill:
@daveh86, this seems to conflate the two correctness tests in `wiredtiger-test-wtperf` with the performance tests in `wiredtiger-perf-stress`.  I only care about the former for now.

michaelcahill:
Thanks, @daveh86, I'll merge.  We should keep in mind the potential for confusion between "stress" performance tests and "stress" correctness tests.

PR #3132, opened 2016-11-07
WT-3009 Remove the eviction_dirty_target from test/format runs with l…

daveh86:
…ess than 20MB cache


michaelcahill:
Thanks @daveh86, lgtm.  Let's see if this is sufficient to resolve the recent test hangs.

PR #3135, opened 2016-11-07
WT-3011 __wt_curjoin_open() saves the wrong URI in the cursor.

keithbostic:
@ddanderson, this one is for you, please.

Two changes.

First, Coverity had a false positive about some code in `__wt_curjoin_open()`, and while looking at it, I noticed we were storing `table:xxx` in the cursor's `internal_uri` field rather than `join:table:xxx`, which looks wrong to me. By happy coincidence, this change should also keep Coverity quiet.

Second, there was one thing I noticed awhile ago was that the __wt_json_token() error handling was a little strange. All but one error wrote out a message and then  fell out of the switch statement to fail (setting the returned token type to -1), and the one other error simply returned (not setting the returned token type). I switched all of the errors to immediately return instead of falling out of the switch statement.

If I'm misunderstanding what's going on here, just kill either or both changes!

### DIFF
--- a/src/cursor/cur_join.c
+++ b/src/cursor/cur_join.c
@@ -1326,7 +1326,6 @@ __wt_curjoin_open(WT_SESSION_IMPL *session,
 	cursor = &cjoin->iface;
 	*cursor = iface;
 	cursor->session = &session->iface;
-	cursor->internal_uri = table->name;
 	cursor->key_format = table->key_format;
 	cursor->value_format = table->value_format;
 	cjoin->table = table;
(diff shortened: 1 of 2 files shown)


ddanderson:
Hi @keithbostic, I agree with the changes, except that it didn't change the ``internal_uri``.  Actually , leaving ``internal_uri`` untouched will do the right thing in the call to ``__wt_cursor_init``, it ``strdups`` it to ``uri``.  I've pushed a mod, see if you agree.

The other change to JSON code lgtm.

keithbostic:
@ddanderson: thank-you, I missed the fact that `__wt_curjoin_open()` was handling `cursor->internal_uri` locally.

Your change looks right to me, I'll merge.

PR #3137, opened 2016-11-08
WT-2962 Allow configuration of builtin extensions.

michaelcahill:
While in the area, fix sending "config={values}" to extensions: just
the values should be passed in.

### DIFF
--- a/examples/c/ex_all.c
+++ b/examples/c/ex_all.c
@@ -1085,6 +1085,15 @@ main(void)
 	 * The connection has been closed.
 	 */
 
+	/*! [Configure zlib extension with compression level] */
+	ret = wiredtiger_open(home, NULL,
+	    "create,"
+	    "extensions=[/usr/local/lib/"
+	    "libwiredtiger_zlib.so=[config=[compression_level=3]]]", &conn);
+	/*! [Configure zlib extension with compression level] */
+	if (ret == 0)
+		(void)conn->close(conn, NULL);
+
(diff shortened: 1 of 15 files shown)


examples/c/ex_all.c
```diff
@@ -1085,6 +1085,15 @@ main(void)
...
 
+	/*! [Configure zlib extension with compression level] */
+	ret = wiredtiger_open(home, NULL,
+	    "create,"
+	    "extensions=[/usr/local/lib/"
+	    "libwiredtiger_zlib.so=[config=[compression_level=3]]]", &conn);
+	/*! [Configure zlib extension with compression level] */
+	if (ret == 0)
+		(void)conn->close(conn, NULL);
+
```
michaelcahill: This is debugging to test the change: I'll revert it.

test/format/config.c
```diff
@@ -301,7 +301,7 @@ config_compression(const char *conf_name)
 		break;
 #endif
 #ifdef HAVE_BUILTIN_EXTENSION_ZSTD
-	case 15: case 16 case 17:		/* 15% zstd */
```
michaelcahill: This is a bug that only shows up when building with `builtins=zstd`, which we don't yet have automated.

michaelcahill:
@sulabhM, can you please review this change?

michaelcahill:
Thanks @sulabhM, I'll merge.

PR #3138, opened 2016-11-08
WT-3014 Add GCC/clang support for ELF symbol visibility.

keithbostic:
@michaelcahill, reading up on your LTO comment in #3119, I realized the compilers we use support attributes to declare symbols hidden/visible, which should result in smaller symbol tables, less need for GOT/relocation instructions and less chance of unexpected API use.

This doesn't help MongoDB because it loads against our object files, not our libraries, although it would potentially help if the build is configured with `--link-model=dynamic`.


michaelcahill:
@keithbostic, in principle this seems like a good idea, but maintaining all of those entry points for tests looks like a pain.  We already try to link most tests statically: could we link them directly against the objects to avoid exporting so many symbols?

keithbostic:
@michaelcahill 
> but maintaining all of those entry points for tests looks like a pain.

First, I agree, I was surprised and unhappy with how many APIs I had to list (38).

That said, the list shouldn't change much, since we don't add many new APIs to the list we "export" outside of the library, so it shouldn't be difficult to maintain.

I'll take a run at linking against the objects.

Another idea: a list of function pointers in the library's `global` or `WT_CONNECTION` structure instead, then test and/or other programs could get to the hidden APIs via a function pointer.

I could live with either that or linking with the objects, whichever looks less fragile going forward.

keithbostic:
@michaelcahill: linking against objects doesn't work well.

The wt utility pulls in a lot of code: it needs `cursor/cur_json.o` and that object in turn requires configuration, scratch buffer, error handling and lots of other code. I didn't get to the python `wiredtiger_wrap.c` code, but I think it's likely to be difficult as well.

Linking against objects will require constant tweaking as library source code changes what underlying functions are called, it's worse than listing the 38 APIs explicitly.

We could load the test code and `wt` utility against all of the library objects, I suppose, that would work and would keep us from having to constantly tweak the change, but I don't like that much.

keithbostic:
@ddanderson, in 10b45b8 you added an explicit call to `__wt_err()` in `lang/python/wiredtiger.i`, and I was wondering why that was the right choice? Is there something we could use other than an internal WiredTiger function?

keithbostic:
@michaelcahill, I took a run at this today based on our conversations, and I can't think of a way to load the test Makefiles against the underlying set of library objects. It'd be great if you can think of one, but I don't see the path.

Otherwise, I suggest we agree we're not likely to export new symbols to future test programs and so the maintenance of the list of "default" functions won't be a problem, and take the branch as is, else, discard the branch.

I prefer the former, but not strongly -- I don't think the functionality here is sufficiently interesting to push for, so I'm happy to simply discard the branch if you prefer.

keithbostic:
@michaelcahill, I briefly talked with @ddanderson yesterday, and his opinion is we can replace the Java and Python `wiredtiger.i` calls to internal WiredTiger functions with other calls.

There are 4: `__wt_calloc_def/__wt_free`, `__wt_sleep` and `__wt_err()`. All of them except `__wt_err` are easy, Don suggests replacing `__wt_err` with `WT_EXTENSION_API->err_printf()`.

If we decide to discard this branch, I'll create a separate ticket for that change, and I'll cherry-pick the changes to `examples/c/ex_file_system.c` so it doesn't require WiredTiger internal functions.

michaelcahill:
@keithbostic, I took another look through these changes and I'm inclined to take this branch.  I can't see a maintainable way to change the build, but even if we figure that out, it can be a commit on top of these changes.

If we can further reduce the set of exported functions by making changes to the Java and Python APIs, that's great but can happen in a followup ticket.

Let me know if you're happy for this branch to be merged.

keithbostic:
> Let me know if you're happy for this branch to be merged.

fbm.

PR #3139, opened 2016-11-08
WT-3000 Fix missing records at end of previous log file

sueloverso:
Not yet ready to merge.  Just pushed the test changes to reproduce the problem.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -863,6 +863,7 @@ lu
 lwsync
 lz
 lzo
+mT
 madvise
 majorp
 malloc
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -62,18 +62,8 @@ __log_fs_write(WT_SESSION_IMPL *session,
 	 * writes to the previous log file to complete otherwise there could
 	 * be a hole at the end of the previous log file that we cannot detect.
 	 */
-	if (slot->slot_release_lsn.l.file < slot->slot_start_lsn.l.file) {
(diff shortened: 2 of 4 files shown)


sueloverso:
@michaelcahill this is ready for review and merge.  I took a different approach than the one we discussed.  I think this is less impactful and simpler code changes and doesn't introduce any version changes.

michaelcahill:
@sueloverso, I thought we talked about this: my concern with this approach is that I don't think it is crash safe.  In other words, I think we need to wait for `sync_lsn`, not `write_lsn`.  Otherwise, after a crash, we still don't know whether the last write(s) in a log file will be on disk.

We could make that change so this version is safe, but my concern is that waiting for an fsync when we switch log files will introduce large latencies in some workloads.  That's why I suggested adding information to the log file header so that we can directly detect whether all of the previous log file was seen during recovery.

sueloverso:
That introduces a bunch of complexities.  First, almost all log files are pre-allocated, so they already have a header in them.  So to rewrite the header in N+1 with this information means an open/write/sync at that time and so is effectively equivalent.  Then there is added complexity in recovery where we have to read N+1 to compare to processing log file N as well as the complexity of versions.

sueloverso:
I've made the change to do the fsync and started a Jenkins logging wtperf run on that branch.  I'll run against mongodb workloads tomorrow.

sueloverso:
@michaelcahill I put some performance info in the ticket.  I have not seen any performance slowdown doing the sync.  The background `log_flush` thread keeps that down a couple megabytes, at most.  This makes the code change quite small.

michaelcahill:

PR #3141, opened 2016-11-11
WT-3017 Don't set NULL hazard pointers.

michaelcahill:


michaelcahill:

While it's obviously scary to have this bug here, I think what we've seen (failing during `__wt_hazard_clear`) would always be one of the results, and the workload is causing an extreme number of page replacements (which explains why we're seeing this, and I'll chase separately).

I have a much bigger set of paranoia changes that led me here, mostly asserting that `__wt_hazard_count(session, page) > 0` in more and more places until I got back here.

src/support/hazard.c
```diff
@@ -101,7 +112,8 @@ __wt_hazard_set(WT_SESSION_IMPL *session, WT_REF *ref, bool *busyp
 		 * found this page using the tree's key space, whatever page we
 		 * find here is the page for us to use.)
 		 */
-		if (ref->page == hp->page && ref->state == WT_REF_MEM) {
+		if (ref->page == page && ref->state == WT_REF_MEM) {
+			WT_ASSERT(session, hp->page != NULL);
```
michaelcahill: On further thought, we may want:

```
        WT_READ_BARRIER();
        if (ref->page == page) {
                ...
        }
}
```

i.e., what we need to be sure about is that `ref->page` is correct *after* we check `ref->state`.  Let me know what you think: I don't have enough runs yet to be confident that what's in the branch is sufficient.

src/support/hazard.c
```diff
@@ -101,9 +112,12 @@ __wt_hazard_set(WT_SESSION_IMPL *session, WT_REF *ref, bool *busyp
...
 		 */
-		if (ref->page == hp->page && ref->state == WT_REF_MEM) {
-			++session->nhazard;
-			return (0);
+		if (ref->state == WT_REF_MEM) {
+			WT_READ_BARRIER();
+			if (ref->page == page) {
+				++session->nhazard;
+				return (0);
+			}
```
keithbostic: First, I don't understand the original commit 3baa16b; I don't think using a local variable for `page` makes a difference because `hp->page` must be re-read after the barrier? What am I missing there?

Second, I agree with you on the final version of this change.

My reading of ISO/IEC 9899:201x is the read-barrier isn't necessary because the "ordinary" read of `WT_REF.page` can't be reordered across the volatile read of `WT_REF.state`. Section 5.1.2.3 says "The presence of a sequence point between the evaluation of expressions A and B implies that every value computation and side effect associated with A is sequenced before every value computation and side effect associated with B." Since a logical AND is a "sequence point" and accessing a volatile is a "side effect", the ordinary read of the page cannot be done before the volatile read of the state.

However, this hasn't always been the case, apparently, [Dr. Dobbs from 2009](http://www.drdobbs.com/parallel/volatile-vs-volatile/212701484?pgno=2) says:
```
Second, what about nearby ordinary reads and writes -- can those still 
be reordered around unoptimizable reads and writes? Today, there is no
practical portable answer because C/C++ compiler implementations vary
widely and aren't likely to converge anytime soon. For example, one
interpretation of the C++ Standard holds that ordinary reads can move
freely in either direction across a C/C++ volatile read or write, but
that an ordinary write cannot move at all across a C/C++ volatile read
or write -- which would make C/C++ volatile both less restrictive and
more restrictive, respectively, than an ordered atomic. Some compiler
vendors support that interpretation; others don't optimize across
volatile reads or writes at all; and still others have their own
preferred semantics.
```

The bottom line is I don't expect the read-barrier to cost us any performance (it's just forcing the ordering of a read we have to do anyway), we might reasonably remove `volatile` from `WT_REF.state` and that would introduce a nasty bug, so I'd leave the read-barrier in.

michaelcahill:
Thanks for the review, @keithbostic: I'll merge.

FTR, the reason the first change avoided the failure we've seen in testing is not (just) the local variable, it's the NULL check.

Also, regardless of what the spec says and what order the compiler emits reads, I believe the read barrier is required for correctness.  Without it, the CPU could schedule reads of `ref->page` before the read of `ref->state`.

keithbostic:
@michaelcahill 
> regardless of what the spec says and what order the compiler emits reads, I believe the read barrier is required for correctness.  Without it, the CPU could schedule reads of `ref->page` before the read of `ref->state`.

Sorry to push back on this, but I'm missing something -- my reading of the spec is the compiler cannot reschedule the read of `ref->page` before `ref->state`?

michaelcahill:
@keithbostic, the compiler can't but the CPU can, without a barrier.

PR #3143, opened 2016-11-12
WT-3012 Check a btree is LSM primary before switching.

michaelcahill:


michaelcahill:
Self review: lgtm and has been running a stress job for most of the weekend without failing.

PR #3144, opened 2016-11-14
WT-2984 Keep sufficient history in the metadata for queries.

michaelcahill:
Since we treat the checkpoint transaction specially, we also have to
track the amount of history required for the metadata specially.
Previously, there was a window where a query started while a checkpoint
was running could fail to see the results of the checkpoint when it
queried the metadata.


michaelcahill:
@sueloverso, can you please review this and check that it fixes the failure for you (it does for me)?

sueloverso:
LGTM and I have run the failing case several times successfully.

PR #3146, opened 2016-11-14
WT-3015 Change when we will evict internal pages

daveh86:


daveh86:
Have pushed this into the perf_branch for performance testing.

michaelcahill:
Thanks @daveh86, lgtm and the perf numbers are unchanged.  I'll merge...

PR #3147, opened 2016-11-15
WT-3020 Always make LSM chunks evictable when they are switched out.

michaelcahill:
Reverts a bug introduced in WT-3012.


sueloverso:
LGTM - merge once the pull request tester finishes.

PR #3148, opened 2016-11-15
WT-3022 Change lsm_tree flags to fields to prevent race conditions

sulabhM:


sulabhM:
@daveh86 : Can you review this change please.

daveh86:
Only thing I see is that you are implicit about evaluating bools as true, but explicit when evaluating as false.

I think the standard is to always be implicit in the case of a bool, so you can drop the "== false" from all your checks.

Beyond that LGTM.

michaelcahill:
Thanks @sulabhM, lgtm.  I'll merge.

PR #3153, opened 2016-11-16
WT-3000 Backport fixes and test to 3.0.

sueloverso:
@michaelcahill Please review this backport of the fixes and test for WT-3000 for 3.0.  It was a bit more complicated than originally thought.
There are a few things to note:
* I backported the whole `test/recovery` directory.  That includes the `truncated-log` test as well.  That was easier than teasing it apart, but I can do that if you don't want that second test.
* I needed to backport other fixes in order to get the test to pass.  In particular, the fixes for WT-2184 and WT-2559 were required to get the test to pass.
* The fix for WT-2559 says it is fixed in 3.0.13 but that is not the case.  There is no 3.0 backport in that ticket and I'm not sure why/how that got set there.
* The fix for WT-2184 was needed for recovery.
* I needed to fix the if/else in `log_write_internal` otherwise some calls weren't correctly waiting in write-no-sync mode.  That mode will have a perf impact, but MongoDB doesn't use that mode.


michaelcahill:

PR #3154, opened 2016-11-16
WT-3025 Fix error path in log_force_sync.

sueloverso:
@michaelcahill Please review this minor fix.  While doing WT-3000 I noticed this bug in the error path.


michaelcahill:
Thanks @sueloverso, lgtm.

For the record, my understanding is that without this fix, we could leak a file descriptor, but only in the case where {{__wt_fsync}} failed, which always logs an error message.

PR #3155, opened 2016-11-16
WT-2670 Add access_pattern_hint configuration option

sueloverso:
Not ready for review.


sueloverso:
This is ready for review.  I ran the `500m-btree-populate.wtperf` with each of the different settings.  I then ran the `500m-btree-rdonly.wtperf` workload to observe performance differences.  When run with `default` of `POSIX_FADV_NORMAL` I see a 2x performance increase over either `random` or `sequential`.  This is similar to what the user observed in their app.

Some things to note:
* I changed the name from `normal` in the ticket to `default`.
* I made `default` the default value, which is different from today's `POSIX_FADV_RANDOM`.
* I have kicked off a Jenkins perf build on this branch to see if it shows any differences.

dist/api_data.py
```diff
@@ -120,6 +120,14 @@ def __cmp__(self, other):
...
 # Per-file configuration
 file_config = format_meta + [
+    Config('access_pattern', 'default', r'''
+        It is recommended that workloads that consist primarily of
+        updates and/or point queries specify \c random.  Workloads that
+        do many cursor scans through large ranges of data specify
+        \c sequential and other workloads specify \c default.  The
+        option leads to an advisory call to an appropriate operating
+        system API where available''',
+        choices=['default', 'random', 'sequential']),
```
agorrod: I don't love ``default`` here - it's not clear what that means. Maybe ``mixed`` would fit better? or ``normal`` to match the flag that is being passed to ``fadvise`` (though I'm also inclined to skip the call completely in mixed mode).

src/os_posix/os_fs.c
```diff
@@ -681,12 +681,19 @@ __posix_open_file(WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session,
 	 * interesting.
 	 */
 	if (!pfh->direct_io && file_type == WT_FS_OPEN_FILE_TYPE_DATA) {
+		advise_flag = POSIX_FADV_NORMAL;
```
agorrod: I believe there is a matching change required in  ``src/os_win/os_fs.c``, we currently pass in ``FILE_FLAG_RANDOM_ACCESS``.

src/os_posix/os_fs.c
```diff
@@ -681,12 +681,19 @@ __posix_open_file(WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session,
 	 * interesting.
 	 */
 	if (!pfh->direct_io && file_type == WT_FS_OPEN_FILE_TYPE_DATA) {
```
agorrod: I would quite like to skip the call to posix_fadvise completely in the default, or normal case. Then we wouldn't override external configurations (I don't know if there is such a facility?).

src/os_posix/os_fs.c
sueloverso: Yes, and also `FILE_FLAG_SEQUENTIAL_SCAN`.  There is no equivalent to `POSIX_FADV_NORMAL`.  I'll fix that up.

dist/api_data.py
sueloverso: AFAICT the man page states that `POSIX_FADV_NORMAL` is essentially a no-op.  It is what is used if no advice is given.  So I don't care much either way if we skip it or pass in normal.  The current code always calls it to be explicit and I kept that.

sueloverso: What about `none` which we use in other places?

sueloverso:
FTR so I remember tomrrow, we'll change it to `access_pattern_hint` and use `none` and skip the call.

sueloverso:
autobuild: retest this please

sueloverso:
@agorrod This is now changed per our discussion last night.  @michaelcahill please review as I changed `api_config.py` after battling the config demon a long time today.  My original change failed several python tests, particularly dump/load tests like `util02`.  The reason is that having `none` as the default and a list of `choices` made the default string `access_pattern_hint=` and a load/session create failed since `choices` implied a non-empty setting.

Other solutions I rejected were using some other string like `off` or `skip` as a replacement for `none`, getting rid of `choices` and manually checking, trying various quoting in `api_data.py` and those were all ugly or felt like hack workarounds.

src/os_win/os_fs.c
```diff
@@ -521,10 +521,16 @@ __win_open_file(WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session,
...
-	if (file_type == WT_FS_OPEN_FILE_TYPE_DATA)
+	/* If the user indicated a random workload, disable read-ahead. */
+	if (file_type == WT_FS_OPEN_FILE_TYPE_DATA &&
+	    LF_ISSET(WT_FS_OPEN_ACCESS_RAND))
 		f |= FILE_FLAG_RANDOM_ACCESS;
 
+	/* If the user indicated a sequential workload, set that. */
+	if (file_type == WT_FS_OPEN_FILE_TYPE_DATA &&
+	    LF_ISSET(WT_FS_OPEN_ACCESS_RAND))
+		f |= FILE_FLAG_SEQUENTIAL_SCAN;
```
michaelcahill: This looks like a copy/paste bug: should be `LF_ISSET(WT_FS_OPEN_ACCESS_SEQ)`

src/os_posix/os_fs.c
```diff
@@ -676,17 +676,25 @@ __posix_open_file(WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session,
...
 	/*
-	 * Disable read-ahead on trees: it slows down random read workloads.
+	 * If the user set an access pattern hint, call fadvise now.
 	 * Ignore fadvise when doing direct I/O, the kernel cache isn't
 	 * interesting.
 	 */
-	if (!pfh->direct_io && file_type == WT_FS_OPEN_FILE_TYPE_DATA) {
+	if (!pfh->direct_io && file_type == WT_FS_OPEN_FILE_TYPE_DATA &&
+	    LF_ISSET(WT_FS_OPEN_ACCESS_RAND |
+	    LF_ISSET(WT_FS_OPEN_ACCESS_SEQ))) {
```
michaelcahill: should be `LF_ISSET(WT_FS_OPEN_ACCESS_RAND | WT_FS_OPEN_ACCESS_SEQ)` (no nested `LF_ISSET`: that does something very strange).

src/os_posix/os_fs.c
```diff
@@ -676,17 +676,25 @@ __posix_open_file(WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session,
...
+		if (LF_ISSET(WT_FS_OPEN_ACCESS_RAND))
+			advise_flag = POSIX_FADV_RANDOM;
+		if (LF_ISSET(WT_FS_OPEN_ACCESS_SEQ))
+			advise_flag = POSIX_FADV_SEQUENTIAL;
```
michaelcahill: I'm surprised this doesn't generate an unused variable warning: how can the compiler figure out that `advise_flag` will always be set here?

dist/api_data.py
```diff
@@ -120,6 +120,14 @@ def __cmp__(self, other):
...
 # Per-file configuration
 file_config = format_meta + [
+    Config('access_pattern_hint', 'none', r'''
+        It is recommended that workloads that consist primarily of
+        updates and/or point queries specify \c random.  Workloads that
+        do many cursor scans through large ranges of data specify
+        \c sequential and other workloads specify \c none.  The
+        option leads to an advisory call to an appropriate operating
+        system API where available''',
+        choices=['none', 'random', 'sequential']),
```
michaelcahill: We're making a change to the default behavior here (from "random" to "none").  I suspect that's right but would like some data to back it up.  I'll kick off MongoDB perf tests against this branch to see if there are any regressions there...

dist/api_config.py
```diff
@@ -193,12 +193,14 @@ def checkstr(c):
...
 def get_default(c):
+    cdef = c.flags
     t = gettype(c)
+    choices = cdef.get('choices', [])
     if c.default == 'false':
         return 'false'
     elif c.default == 'true':
         return 'true'
-    elif t == 'string' and c.default == 'none':
+    elif t == 'string' and c.default == 'none' and choices == []:
```
michaelcahill: This seems fine to me.

It could just be `not c.flags.get(choices, [])`, I think?

sueloverso:
I've been seeing an improvement (and occasionally the same perf) when running the 500m readonly test with the new default compared to random or sequential.  I have restarted the perf tests and kicked off a long perf test against this final version of the branch to see any differences on the plots.

agorrod:
Thanks @sueloverso lgtm. I pushed something into the upgrading documentation. @michaelcahill Would you like any more testing before we merge this?

sueloverso:
I have run the long perf tests a few times and the regular Jenkins perf tests many times with this branch and there are no differences.  I ran with both defaults (none and random) on evergreen sys-perf.  Results there are very inconsistent so the default setting does not seem to matter.

michaelcahill:
Thanks @sueloverso and @agorrod, lgtm.  I'll merge.

PR #3156, opened 2016-11-18
WT-3026 log trickle

sueloverso:
Not yet ready for review.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -583,11 +583,11 @@ wiredtiger_open_log_configuration = [
         Config('recover', 'on', r'''
             run recovery or error if recovery needs to run after an
             unclean shutdown''',
-            choices=['error','on'])
+            choices=['error','on']),
         Config('trickle', '100', r'''
             start a background flush of the log file every time this
             many milliseconds passes.  A value of zero disables the flush''',
-            min='0', max='100000'),
+            min='0', max='100000')
     ]),
 ]
 
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -575,15 +575,19 @@ wiredtiger_open_log_configuration = [
         Config('file_max', '100MB', r'''
             the maximum size of log files''',
             min='100KB', max='2GB'),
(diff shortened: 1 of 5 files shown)


sueloverso:
@agorrod and @michaelcahill this is ready for review.  With this I see fewer latency warnings with the 500m populate configuration and the maximum latencies are lower as well.  We may be able to remove the log flush background thread in MongoDB now after this is merged.

dist/api_data.py
```diff
@@ -575,15 +575,19 @@ def __cmp__(self, other):
...
+            the name of a directory into which log files are written. The
+            directory must already exist. If the value is not an absolute
+            path, the path is relative to the database home (see @ref
+            absolute_path for more information)'''),
         Config('recover', 'on', r'''
             run recovery or error if recovery needs to run after an
             unclean shutdown''',
-            choices=['error','on'])
+            choices=['error','on']),
+        Config('trickle', '100', r'''
```
agorrod: I prefer ``background_flush_period`` to ``trickle``. How did you choose 100ms as the default? I wonder if the default should be 0, since we don't really want to add in additional fsync calls for applications that configure synchronous logging.

src/conn/conn_log.c
```diff
@@ -163,6 +163,9 @@ __logmgr_config(
 			FLD_SET(conn->log_flags, WT_CONN_LOG_RECOVER_ERR);
 	}
 
+	WT_RET(__wt_config_gets(session, cfg, "log.trickle", &cval));
+	conn->log_trickle = (uint32_t)cval.val;
```
agorrod: It's not just the user-visible name I don't really like ``trickle`` for, I'd quite like for a name change to flow down.

sueloverso:
FTR, this is on hold as we may not want this, due to WT-3039.

agorrod:
@sueloverso Can we close this branch? I believe we have committed to the approach in WT-3039 now.

sueloverso:
Yes.  I will close the PR and ticket.

PR #3157, opened 2016-11-24
WT-3030 Test failure indicating invalid key order during traversal

keithbostic:


keithbostic:
The fix is in #3166, but merge this pull request (#3157), it has useful changes.

PR #3158, opened 2016-11-25
WT-3028 Count eviction of dirty pages as progress.

michaelcahill:
Fixes part of WT-3023, which was only supposed to avoid counting page
rewrites as progress...


michaelcahill:
@sulabhM, can you please review?

This means the change of WT-3023 now only applies to page rewrites.  All other cases of eviction and splits will increment the count of evicted pages, so eviction will treat them as making progress.  Rewrites will not increment that count, so if that is all we are doing, eviction will stay "stuck".

PR #3159, opened 2016-11-28
WT-3028 Directly track if reconciliation makes progress.

michaelcahill:
Give up on page rewrites that don't use any updates.


agorrod:
Thanks @michaelcahill I like this approach, it makes sense to give up on the rewrite if it won't save any space. LGTM
@keithbostic would you mind doing a review as well? In particular I'd like more confidence that the change to ``__rec_write_status`` doesn't capture any cases we aren't expecting.

michaelcahill:
A little more context: I tried to leave the reconciliation and split code alone and change cache accounting so that these operations wouldn't count as progress in WT-3023.  That change was too broad and led to some workloads that weren't stuck failing the diagnostic check in the eviction server.

This version is targeted as the specific situation where we are rewriting a page and didn't use any of the updates.  I can't think of any reason to go discard the original page in that case.

src/reconcile/rec_write.c
```diff
@@ -555,6 +560,14 @@ __rec_write_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
...
 
 	/*
+	 * If there is only one block and no update was used, no progress has
+	 * been made and there is no point swapping the new page into place.
+	 */
+	if (r->bnd_next == 1 && F_ISSET(r, WT_EVICT_UPDATE_RESTORE) &&
+	    r->bnd[0].supd != NULL && !r->update_used)
+		return (EBUSY);
+
+	/*
```
keithbostic: To be honest, I'm just stunned this case occurs enough to make it worth catching.

The change looks reasonable to me, but we're starting to accumulate tests in `__rec_write_status`, and that's not a great place for them.

And... in staring at this, isn't it a problem if a failed reconciliation in the service of checkpoint doesn't reset the tree's `modified` field?

We call `__rec_write_status()` before we find out if the `__rec_write_wrapup()` call (which frees blocks, actually performs the root-page checkpoint write, and so on) has succeeded.

Can we make things better by splitting the `__rec_write_status()` code into two parts, "final tests", and "final status", and calling the functions in this order:
* final-tests
* reconciliation wrap-up
* final-status

Just for clarity, here's what I'm thinking about:
```
index 2692f35..1a07897 100644
@@ -315,6 +315,8 @@ static int  __rec_col_var(WT_SESSION_IMPL *,
 static int  __rec_col_var_helper(WT_SESSION_IMPL *, WT_RECONCILE *,
                WT_SALVAGE_COOKIE *, WT_ITEM *, bool, uint8_t, uint64_t);
 static int  __rec_destroy_session(WT_SESSION_IMPL *);
+static int  __rec_final_checks(WT_SESSION_IMPL *, WT_RECONCILE *);
+static void __rec_final_status(WT_SESSION_IMPL *, WT_RECONCILE *, int);
 static int  __rec_root_write(WT_SESSION_IMPL *, WT_PAGE *, uint32_t);
 static int  __rec_row_int(WT_SESSION_IMPL *, WT_RECONCILE *, WT_PAGE *);
 static int  __rec_row_leaf(WT_SESSION_IMPL *,
@@ -334,7 +336,6 @@ static int  __rec_update_las(
                WT_SESSION_IMPL *, WT_RECONCILE *, uint32_t, WT_BOUNDARY *);
 static int  __rec_write_init(WT_SESSION_IMPL *,
                WT_REF *, uint32_t, WT_SALVAGE_COOKIE *, void *);
-static int  __rec_write_status(WT_SESSION_IMPL *, WT_RECONCILE *, WT_PAGE *);
 static int  __rec_write_wrapup(WT_SESSION_IMPL *, WT_RECONCILE *, WT_PAGE *);
 static int  __rec_write_wrapup_err(
                WT_SESSION_IMPL *, WT_RECONCILE *, WT_PAGE *);
@@ -426,9 +427,9 @@ __wt_reconcile(WT_SESSION_IMPL *session,
        WT_ILLEGAL_VALUE_SET(session);
        }
 
-       /* Get the final status for the reconciliation. */
+       /* Final checks for a successful reconciliation. */
        if (ret == 0)
-               ret = __rec_write_status(session, r, page);
+               ret = __rec_final_checks(session, r);
 
        /* Wrap up the page reconciliation. */
        if (ret == 0)
@@ -436,6 +437,9 @@ __wt_reconcile(WT_SESSION_IMPL *session,
        else
                WT_TRET(__rec_write_wrapup_err(session, r, page));
 
+       /* Set the final status for the reconciliation. */
+       __rec_final_status(session, r, ret);
+
        /* Release the reconciliation lock. */
        __wt_writeunlock(session, &page->page_lock);
 
@@ -540,18 +544,12 @@ __rec_las_checkpoint_test(WT_SESSION_IMPL *session, WT_RECONCILE *r)
 }
 
 /*
- * __rec_write_status --
- *     Return the final status for reconciliation.
+ * __rec_final_checks --
+ *     Some final checks before we wrap things up.
  */
 static int
-__rec_write_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
+__rec_final_checks(WT_SESSION_IMPL *session, WT_RECONCILE *r)
 {
-       WT_BTREE *btree;
-       WT_PAGE_MODIFY *mod;
-
-       btree = S2BT(session);
-       mod = page->modify;
-
        /*
         * If we have used the lookaside table, check for a lookaside table and
         * checkpoint collision.
@@ -567,10 +565,28 @@ __rec_write_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
            r->bnd[0].supd != NULL && !r->update_used)
                return (EBUSY);
 
+       return (0);
+}
+
+/*
+ * __rec_final_status --
+ *     Return the final status for reconciliation.
+ */
+static void
+__rec_final_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, int error)
+{
+       WT_BTREE *btree;
+       WT_PAGE *page;
+       WT_PAGE_MODIFY *mod;
+
+       btree = S2BT(session);
+       page = r->page;
+       mod = page->modify;
+
        /*
         * Set the page's status based on whether or not we cleaned the page.
         */
-       if (r->leave_dirty) {
+       if (error || r->leave_dirty) {
                /*
                 * The page remains dirty.
                 *
@@ -625,8 +641,6 @@ __rec_write_status(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
                else
                        WT_ASSERT(session, !F_ISSET(r, WT_EVICTING));
        }
-
-       return (0);
 }
 
 /*
@@ -3616,7 +3630,7 @@ __wt_bulk_wrapup(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk)
 
        WT_RET(__rec_split_finish(session, r));
        WT_RET(__rec_write_wrapup(session, r, r->page));
-       WT_RET(__rec_write_status(session, r, r->page));
+       __rec_final_status(session, r, 0);
 
        /* Mark the page's parent and the tree dirty. */
        parent = r->ref->home;
```

michaelcahill: I'm happy to split `__rec_write_status` but don't love those names.  I'll take a run at this.

Note that this is currently causing tests to fail so I'm keen to get the fix into shape and merged today if I can.

src/reconcile/rec_write.c
```diff
@@ -50,6 +50,11 @@ typedef struct {
 	bool leave_dirty;
 
 	/*
+	 * Track if an update was copied onto a page during reconciliation.
+	 */
+	bool update_used;
+
+	/*
```
michaelcahill: I originally called it `progress`.  It's tracking whether a `WT_UPDATE` is used when writing a new page image.  I honestly can't see why any eviction should go ahead if we don't set this, but constrain the check to rewrites for now because constant rewriting of pages is the worst case we see in real workloads (generally when pages grow larger than `memory_page_max` for various reasons).

I'll rename it to something clearer.

src/reconcile/rec_write.c
michaelcahill: BTW, I don't think there is any issue for checkpoints: this test is restricted to `WT_EVICT_UPDATE_RESTORE`.

keithbostic: fbm, @agorrod and I are both good with your original change, so if this isn't as simple as it looks, let's merge your patch and we can open another ticket for the checkpoint fix (assuming you agree with me there's a problem there).

keithbostic: The issue for checkpoints is more wide-ranging: imagine we call `__rec_write_status`, then `__rec_write_wrapup` fails: the checkpoint will fail but we've lost our chance to set the tree's `modified` flag back to true. Or maybe I'm just missing something...

michaelcahill: @keithbostic, checkpoints marking trees dirty is already handled: https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn_ckpt.c#L1453-L1457

agorrod:
Thanks @michaelcahill lgtm assuming it passes tests.

src/reconcile/rec_write.c
keithbostic: @michaelcahill, if you make a change here, I think the code in `txn_ckpt.c` to re-mark the tree dirty could use `__wt_tree_modify_set()` instead of inlining it.

PR #3160, opened 2016-11-28
WT-2969 Possible snapshot corruption during compaction

keithbostic:
@michaelcahill, @agorrod, I believe this is ready for review.


keithbostic:
autobuild: retest this please

dist/api_data.py
```diff
@@ -847,9 +847,10 @@ def __cmp__(self, other):
 
 'WT_SESSION.drop' : Method([
     Config('checkpoint_wait', 'true', r'''
-        wait for the checkpoint lock, if \c checkpoint_wait=false, perform
-        the drop operation without taking a lock, returning EBUSY if the
-        operation conflicts with a running checkpoint''',
+        wait for concurrent checkpoints to complete before attempting the drop
+        operation. If \c checkpoint_wait=false, attempt the drop operation
+        without waiting, returning EBUSY if the operation conflicts with a
+        running checkpoint''',
```
keithbostic: trying to clarify that `checkpoint_wait` isn't about locking, that is, `lock_wait` configures whether or not we "wait for the checkpoint lock", `checkpoint_wait` only configures if we try to acquire the checkpoint lock.

src/session/session_api.c
```diff
@@ -804,27 +804,33 @@ __wt_session_drop(WT_SESSION_IMPL *session, const char *uri, const char *cfg[])
 	WT_RET(__wt_config_gets_def(session, cfg, "checkpoint_wait", 1, &cval));
 	checkpoint_wait = cval.val != 0;
 	WT_RET(__wt_config_gets_def(session, cfg, "lock_wait", 1, &cval));
-	lock_wait = cval.val != 0 || F_ISSET(session, WT_SESSION_LOCK_NO_WAIT);
-
-	if (!lock_wait)
-		F_SET(session, WT_SESSION_LOCK_NO_WAIT);
```
keithbostic: @michaelcahill, notice I removed the `F_ISSET(session, WT_SESSION_LOCK_NO_WAIT);` part of the `lock_wait` test. I couldn't figure out what that did, it just didn't make any sense to me.

src/include/schema.h
```diff
@@ -113,12 +112,16 @@ struct __wt_table {
 } while (0)
 
 /*
- * WT_WITH_CHECKPOINT_LOCK --
+ * WT_WITH_CHECKPOINT_LOCK, WT_WITH_CHECKPOINT_LOCK_NOWAIT --
  *	Acquire the checkpoint lock, perform an operation, drop the lock.
  */
-#define	WT_WITH_CHECKPOINT_LOCK(session, ret, op)			\
-	WT_WITH_LOCK(session, ret,					\
+#define	__WT_WITH_CHECKPOINT_LOCK(session, ret, nowait, op)		\
```
michaelcahill: Can we find some way to do this without adding a new class of macro that starts with underscores?

src/include/schema.h
```diff
@@ -98,10 +98,9 @@ struct __wt_table {
  * WT_WITH_LOCK --
  *	Acquire a lock, perform an operation, drop the lock.
  */
-#define	WT_WITH_LOCK(session, ret, lock, flag, op) do {			\
+#define	WT_WITH_LOCK(session, ret, nowait, lock, flag, op) do {		\
```
michaelcahill: How about we just make this `WT_WITH_LOCK_NOWAIT` and avoid the extra argument and the `else`?

src/txn/txn_ckpt.c
```diff
@@ -894,13 +917,22 @@ __wt_txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
 	 *
 	 * Application threads are not generally available for potentially slow
 	 * operations, but checkpoint does enough I/O it may be called upon to
 	 * perform slow operations for the block manager.
+	 *
+	 * Application checkpoints wait until the checkpoint lock is available,
+	 * compaction checkpoints don't.
 	 */
+#define	WT_TXN_SESSION_MASK						\
+	(WT_SESSION_CAN_WAIT | WT_SESSION_NO_EVICTION)
```
michaelcahill: Can we get rid of `WT_SESSION_CAN_WAIT`?  Is there any case where we know it helps with a workload?

src/session/session_api.c
keithbostic: > The variable was badly named, but that was determining if `WT_SESSION_LOCK_NO_WAIT` was already set on a session before trying to set it again (and later clear it).

OK, good -- I don't think there's a path where that can happen, but obviously we don't have to worry about it unless we revert back to using a `WT_SESSION_IMPL` flag.

src/txn/txn_ckpt.c
keithbostic: > The original intent of lock_wait=false was that the operation would not wait for any locks, which is why it used a session flag. We had some exceptions already, but this branch makes a bigger change of only applying lock_wait to the high-level checkpoint / schema locks.
> Can you explain why that is necessary? What problem did the compact change run up against with drops and lock_wait=false?

I'm sorry! (I hate it when people make me figure out why they made a change, my apologies.)

The compact problem wasn't with drop, that was a side-effect of removing the `WT_SESSION_LOCK_NO_WAIT` flag.

Compact wants to do a checkpoint, but not wait on the checkpoint lock. I set `WT_SESSION_LOCK_NO_WAIT` in `__wt_txn_checkpoint()` and that led to a pretty subtle bug.  Checkpoint eventually calls `__sync_file()`, which calls `__wt_evict_file_exclusive_on()`, which calls `WT_WITH_PASS_LOCK()`, which returns `EBUSY` because `WT_SESSION_LOCK_NO_WAIT` is set.

I thought about clearing `WT_SESSION_LOCK_NO_WAIT` after its use in `__wt_txn_checkpoint()`, but you have to do the clear in the lock `op` argument (which gets ugly fast), or in the called function, which separates the flag set from the clear, and kind of defeats the whole idea of a per-session flag.

I also thought about changing `WT_WITH_PASS_LOCK` to call `WT_WITH_LOCK_WAIT` instead of `WT_WITH_LOCK`, but the bug was not only hard to find, it's easy to imagine ways we could re-introduce it in other code paths in the future, that is, it's impossible to guarantee there aren't any problematic lock acquisitions downstream from setting `WT_SESSION_LOCK_NO_WAIT` (and those bugs are potentially workload dependent, so we could fail in the field). To be safe, I think we'd have to change every lock macro other than the top-level ones to ignore the per-session flag, in which case the per-session flag isn't doing anything useful.

The reason I didn't make no-wait versions of all of the macros was because the checkpoint, schema and table locks were the only ones I could find where `WT_SESSION_LOCK_NO_WAIT` looked applicable, and no-wait wasn't needed for the other macros, as far as I could tell. It's certainly possible there are code paths I missed where the per-session flag should reasonably be applied to locks downstream: I looked pretty carefully and didn't see any, but I could easily be wrong.

I could buy the argument there's a downstream lock to which `WT_SESSION_LOCK_NO_WAIT` should apply, and it's easier to pass the flag in the `WT_SESSION_IMPL` structure than drill a hole for a no-wait flag through the API, but as I said, I didn't see any.

In summary, I still prefer this change to a per-session flag, but if you really want the per-session flag, we could put the per-session flag back in and change the non-top-level locking calls to use `WT_WITH_LOCK_WAIT` instead of `WT_WITH_LOCK`.

src/include/schema.h
keithbostic: > How about we just make this WT_WITH_LOCK_NOWAIT and avoid the extra argument and the else?

I didn't do that because it implies the waiting version of the lock macros no longer need a `ret` variable and so they all change, and I figured the compiler would drop the unused code since it's testing a constant.

I'll go ahead and push that change, if you don't like it just revert it.

src/include/schema.h
keithbostic: > Can we find some way to do this without adding a new class of macro that starts with underscores?

Sure, it just means duplicating the asserts in each pair of macros. It's not that bad, I'll go ahead and push that change, if you don't like it, feel free to revert.

src/txn/txn_ckpt.c
keithbostic: > Can we get rid of `WT_SESSION_CAN_WAIT`? Is there any case where we know it helps with a workload?

It's a latency question, not a throughput question: application threads don't set `WT_SESSION_CAN_WAIT`, and so won't get highjacked by the block manager to call `fsync` if `os_cache_dirty_max` or `os_cache_max` is configured to `WT_SESSION::create`.

We could replace it with a test of `WT_SESSION_INTERNAL`, though, if there's no internal thread we're not willing to tap for an fsync call. (That change gets rid of `WT_THREAD_CAN_WAIT` too).

Let me know if you want me to make that change.

michaelcahill:
autobuild: test this please

michaelcahill:

PR #3164, opened 2016-12-01
WT-3048 WiredTiger maximum size warning uses the wrong format.

keithbostic:
WT-3048 WiredTiger maximum size warning uses the wrong format.


agorrod:
Thanks @keithbostic lgtm

PR #3166, opened 2016-12-02
WT-3052 Improve search if index hint is wrong

michaelcahill:
After a split, child references in an internal page can end up in a different location in the index array.  WiredTiger maintains a hint of the expected offset, and searches if the expected reference is not found at the location of the hint.

This change fixes a race where multiple threads could set the hint to different values (if they are searching different versions of an index).  This could lead to a search returning the wrong slot, because we were writing the correct index into the hint, *then* reading from the hint into the returned value.


src/btree/bt_walk.c
```diff
@@ -40,21 +40,15 @@ retry:	WT_INTL_INDEX_GET(session, ref->home, pindex);
...
-			*slotp = ref->pindex_hint = i;
-			return;
-		}
+		if (pindex->index[i] == ref)
+			goto found;
 	for (i = 0; i < pindex->entries; ++i)
 		if (pindex->index[i] == ref) {
-			*pindexp = pindex;
-			*slotp = ref->pindex_hint = i;
+found:			*pindexp = pindex;
```
agorrod: Did you deliberately leave out setting the hint altogether now? I'd sort of expected there to be a new line with:

```
ref->pindex_hint = *slotp;
```

michaelcahill: Yes, as @keithbostic, we're already doing that at https://github.com/wiredtiger/wiredtiger/blob/develop/src/btree/bt_walk.c#L437-L439

tbh, since we already set the hint during splits, I don't know if we should ever update it during reads.

michaelcahill:

This needs to be backported as far as `mongodb-3.2`, but I don't think the `mongodb-3.0` branch needs this fix.  Can you please confirm?

I'm inclined to go further and only update hints during splits, but that assumes all paths through splits do that.  As far as I can see they do, but that isn't trivial to confirm.  What do you think?

```diff
diff --git a/src/btree/bt_walk.c b/src/btree/bt_walk.c
index b687d45..b89be7b 100644
--- a/src/btree/bt_walk.c
+++ b/src/btree/bt_walk.c
@@ -426,15 +426,8 @@ restart:	/*
 			++*walkcntp;
 
 		for (;;) {
-			/*
-			 * Move to the next slot, and set the reference hint if
-			 * it's wrong (used when we continue the walk). We don't
-			 * update those hints when splitting, so it's common for
-			 * them to be incorrect in some workloads.
-			 */
+			/* Move to the next slot. */
 			ref = pindex->index[slot];
-			if (ref->pindex_hint != slot)
-				ref->pindex_hint = slot;
 
 			/*
 			 * If we see any child states other than deleted, the
```

src/btree/bt_walk.c
```diff
@@ -17,54 +17,52 @@ __ref_index_slot(WT_SESSION_IMPL *session,
...
-	__wt_yield();
-	goto retry;
+		/*
+		 * If we don't find our reference, the page split and our home
+		 * pointer references the wrong page. When internal pages
+		 * split, their WT_REF structure home values are updated; yield
+		 * and wait for that to happen.
+		 */
+		__wt_yield();
+	}
```
keithbostic: @michaelcahill, I don't like this loop -- I had to write a test program to convince myself it works.

I put a different version in pull request #3167 that is (very slightly) more efficient, and which I think reads a lot more simply.

If you don't want that change, I think you can simplify this loop, the code moving to smaller slots doesn't need to test against `entries`:
```
*** src/btree/bt_walk.c.orig    Fri Dec  2 08:33:44 2016
--- src/btree/bt_walk.c Fri Dec  2 08:35:03 2016
***************
*** 46,53 ****
                for (j = 1; j <= maxslots; j++) {
                        if ((i = hint + j) < entries && pindex->index[i] == ref)
                                goto found;
!                       if (j <= hint &&
!                           (i = hint - j) < entries && pindex->index[i] == ref)
                                goto found;
                }

--- 46,52 ----
                for (j = 1; j <= maxslots; j++) {
                        if ((i = hint + j) < entries && pindex->index[i] == ref)
                                goto found;
!                       if (j <= hint && && pindex->index[i = hint - j] == ref)
                                goto found;
                }
```

keithbostic:
> I'm inclined to go further and only update hints during splits, but that assumes all paths through splits do that. As far as I can see they do, but that isn't trivial to confirm. What do you think?

My preference is to update hints in the walk code.

When doing the split implementation, I found that hints were often wrong and updating them in the walk code made them a lot more likely to be correct. (For example, there are places in the split code where we just guess at the hints and there's no validation the split code actually gets the hints correct. The walk code actually knows a hint is correct, absent another split.)

Whether this matters in terms of performance I can't say, I don't think we ever did measurements to validate the hints. Intuitively, in big internal pages, I expect them to make a difference.

src/btree/bt_walk.c
keithbostic: OK... I just figured out why the code moving to smaller slots has to test against `entries`, but I think that also means that if there's a big number in the hint field, we're potentially going to be in this loop for a long time.

keithbostic:
> This needs to be backported as far as mongodb-3.2, but I don't think the mongodb-3.0 branch needs this fix. Can you please confirm?

It looks to me like all of the branches (3.0, 3.2 and 3.4) need this fix, the code is essentially identical in all three.

michaelcahill:
Note: moved from WT-3030 to WT-3052 to keep the former focussed on the fix.

michaelcahill:
autobuild: test this please

michaelcahill:
autobuild: test this please

michaelcahill:
autobuild: test this please

agorrod:
@michaelcahill and @keithbostic I like this branch now. I think the search scheme is more likely to efficiently find the right slot than the previous one and the loop is OK to reason through.

michaelcahill:
@keithbostic, I'm happy with this change too, and it's your version of the loop.  Can you please do the (hopefully final) review?

keithbostic:
@michaelcahill, @agorrod: I reviewed the change, and in 8094620, I removed a `WT_ORDERED_READ` call, I don't think it's necessary. (I removed it because I don't want to try and figure out why this read needs to be ordered 6 months from now, not because it's likely to cause problems.)

Otherwise, this lgtm, I'm ready for it to be merged.

keithbostic:
@agorrod, @michaelcahill, looking at this problem again.

If you recall, the root cause was this statement:

`*slotp = ref->pindex_hint = i;`

being logically rewritten as:

```
ref->pindex_hint = i;
*slotp = ref->pindex_hint;
```

I'm suspicious the reason this might happen is we're falling out of the register set (this is an inlined function, called from a function with its own register variables).

If that's what's happening, I'm worried we haven't fixed the problem, we've only fixed a symptom.

Absent the use of volatile in some form, I believe the compiler can rewrite the current code:

```
slot = ref->pindex_hint;
if (slot >= entries)
        slot = entries - 1;
if (pindex->index[    >>> slot <<<    ] == ref)
        goto found;
```
as:
```
slot = ref->pindex_hint;
if (slot >= entries)
        slot = entries - 1;
if (pindex->index[    >>> ref->pindex_hint <<<    ] == ref)
        goto found;
```
which can obviously race.

The Linux kernel handles this with read/write-once macros ([here's a link](http://lxr.free-electrons.com/source/tools/include/linux/compiler.h#L73)).

We could make `WT_REF.pindex_hint` volatile (as we do the page index), but I'm not a big fan of volatile. (See the Linux kernel rants, with which I generally agree; in short, I think adding volatile makes it harder to reason about threaded behavior because the "fix" is separate from the code that deals with threaded behavior.)

My preference would be something like the `__wt_atomic_xxx` macros:
```
#define __wt_access_once32(var) (*((volatile uint32_t *)(&(var))))
...
slot = __wt_access_once32(ref->pindex_hint);
```
instead of the current naked read.

Finally, in reviewing our use of `WT_REF.pindex_hint`, I think the `__check_leaf_key_range` functions in the page-search code are also vulnerable to this same problem, for example, here:

```
WT_INTL_INDEX_GET(session, leaf->home, pindex);
indx = leaf->pindex_hint;
if (indx + 1 < pindex->entries && pindex->index[indx] == leaf)
        if (recno >= pindex->index[indx + 1]->ref_recno) {
                cbt->compare = -1;      /* page keys < search key */
                return (0);
        }
```

michaelcahill:
@keithbostic, I agree that absent `volatile` or a barrier, compilers could generate code that reads from a shared location multiple times.  I don't believe we see that in practice regardless of optimization level (because compilers are able to find a register), but we shouldn't rely on it.

This is exactly the (one and only) problem that `volatile` solves (the compiler will issue a read instruction for each access), and what the `ACCESS_ONCE` macro in Linux does.  I don't believe we need the more complex `READ_ONCE` macro because we don't rely on reads larger than 64-bits.  We use a `volatile` type qualifier already in many places to solve this problem.

This is why you asked to put a read barrier there, which is why we ended up with a `WT_ORDERED_READ`.  I agree that `WT_BARRIER` would be sufficient instead of a read barrier, but I would prefer to mark `pindex_hint` as volatile and not have to think about this each time it is accessed.

The alternative is to add a `WT_ACCESS_ONCE` macro, but then we either need to remove all uses of `volatile` and/or do some type or name munging to make sure all accesses are protected.  Marking `pindex_hint` as `volatile` is a much simpler path.

michaelcahill:
After further offline discussion, we've agreed to declare `pindex_hint` volatile.  If that change passes tests, I'll merge.

PR #3168, opened 2016-12-02
WT-3030  Fix a race between scans and splits reading the index hint.

keithbostic:
WT-3030  Fix a race between scans and splits reading the index hint.


michaelcahill:
Closing: we don't generally use pull requests for backports: `git cherry-pick -x` provides better tracking in the git history.

PR #3171, opened 2016-12-04
WT-2960 Inserting multi-megabyte values can cause pathological lookaside usage

keithbostic:
@michaelcahill, @agorrod, here are the reconciliation changes I described to you.

With this change, it's pretty hard to get the lookaside table to fire at all (fixing the workload described in WT-2960, that workload no longer uses the lookaside table).

I don't see a significant slow-down in any of the workloads I've tested so far. (Alex, that includes random, out-of-cache eviction-stressing workloads.)

I'm going to do some more performance testing on various workloads, but if there are any specific workloads you'd like tested, let me know.


agorrod:
Thanks @keithbostic I'm happy with this change. The code changes make sense to me and my testing hasn't shown up any performance or correctness issues.

src/reconcile/rec_write.c
```diff
@@ -553,14 +567,18 @@ __rec_write_check_complete(WT_SESSION_IMPL *session, WT_RECONCILE *r)
 		return (EBUSY);
 
 	/*
-	 * If we are doing eviction and restoring updates, there is only one
-	 * block and all update were skipped, no progress has been made and
-	 * there is no point swapping the new page into place.
+	 * If we are doing update/restore based eviction, confirm part of the
+	 * page is being discarded, or at least 10% of the updates won't have
```
agorrod: 10% appears a bit arbitrary, I'm a little concerned that a workload will fail to make progress because of this change. OTOH I don't like wasting CPU on update-restore that isn't useful, so I see a definite benefit.

I'm inclined to merge, but watch test/format stress testing, and take a closer look at this check if we start seeing timeouts.

PR #3173, opened 2016-12-05
WT-3030 Fix a race between scans and splits reading the index hint.

michaelcahill:


michaelcahill:
This was reviewed elsewhere, merging...

PR #3174, opened 2016-12-05
WT-3053 Python and Java support use internal WiredTiger functions

keithbostic:
Remove calls to internal WiredTiger functions from the Java and Python APIs, meaning we no longer need to expose a few more internal WiredTiger functions.

### DIFF
--- a/bench/wtperf/idle_table_cycle.c
+++ b/bench/wtperf/idle_table_cycle.c
@@ -83,7 +83,7 @@ cycle_idle_tables(void *arg)
 		snprintf(uri, sizeof(uri),
 		    "%s_cycle%07d", wtperf->uris[0], cycle_count);
 		/* Don't busy cycle in this loop. */
-		__wt_sleep(1, 0);
+		(void)sleep(1);
 
 		/* Setup a start timer. */
 		__wt_epoch(NULL, &start);
@@ -126,7 +126,7 @@ cycle_idle_tables(void *arg)
 		 */
 		while ((ret = session->drop(
 		    session, uri, "force,checkpoint_wait=false")) == EBUSY)
-			__wt_sleep(1, 0);
+			(void)sleep(1);
 
 		if (ret != 0 && ret != EBUSY) {
 			lprintf(wtperf, ret, 0,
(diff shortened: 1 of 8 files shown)


keithbostic:
@ddanderson, would you please take this review? Thanks!

ddanderson:
@keithbostic, this lgtm.  Just wondering what is the problem these changes are solving?  Or is it more about information hiding and best practice?

keithbostic:
> Just wondering what is the problem these changes are solving? Or is it more about information hiding and best practice?

The latter. Ideally, we wouldn't export any symbols in the WiredTiger library other than the external API, that gives us a smaller GOT, faster/better-optimized code, less chance of symbol collision. Right now we export 38 function names we'd really rather not export, and this change gets rid of 4 of those.

bench/wtperf/idle_table_cycle.c
```diff
@@ -83,7 +83,7 @@ cycle_idle_tables(void *arg)
 		snprintf(uri, sizeof(uri),
 		    "%s_cycle%07d", wtperf->uris[0], cycle_count);
 		/* Don't busy cycle in this loop. */
-		__wt_sleep(1, 0);
+		(void)sleep(1);
```
agorrod: We switched to using ``__wt_sleep`` to avoid needing to write two versions of platform specific sleep functions.

lang/java/wiredtiger.i
```diff
@@ -362,7 +362,7 @@ javaClose(JNIEnv *env, WT_SESSION *session, JAVA_CALLBACK *jcb, jfieldID *pfid)
 
 	(*env)->SetLongField(env, jcb->jobj, fid, 0L);
 	(*env)->DeleteGlobalRef(env, jcb->jobj);
-	__wt_free(jcb->session, jcb);
+	free(jcb);
```
agorrod: This change may be OK since the memory is completely owned by Java. On the other hand, if there is a custom memory allocator configured in WiredTiger, I'd rather the Java language binding uses the same allocator rather than the system allocator.

PR #3176, opened 2016-12-05
WT-3053 Python and Java support use internal WiredTiger functions

keithbostic:
Make __wt_err() visible again.


agorrod:
The change that triggered this change was reverted. I also made a couple of review comments in the original pull request: https://github.com/wiredtiger/wiredtiger/pull/3174#pullrequestreview-11490573

PR #3178, opened 2016-12-05
WT-3053 Revert: Python and Java support use internal WiredTiger functions

agorrod:
Reverts wiredtiger/wiredtiger#3174


agorrod:
autobuild: retest this please

PR #3179, opened 2016-12-06
WT-3053 Make Python use internal memory allocation again.

michaelcahill:
Since we aren't going to make allocation functions hidden, revert
changes to the Python API in WT-3014 so it uses internal memory
allocation functions, matching the Java API.


michaelcahill:

agorrod:
Thanks @michaelcahill lgtm.

PR #3181, opened 2016-12-06
WT-3057 WiredTiger hazard pointers should use the WT_REF, not the WT_PAGE.

keithbostic:
@michaelcahill, this is the change we talked about, I think it turned out to be as simple as we thought.

I didn't drive the change down as deeply into the split code as I probably should have, but trying to was relatively messy: code that needs to lock both WT_REF structures and WT_PAGE structures is inherently messy in our code.

### DIFF
--- a/src/support/hazard.c
+++ b/src/support/hazard.c
@@ -106,6 +106,12 @@ __wt_hazard_set(WT_SESSION_IMPL *session, WT_REF *ref, bool *busyp
 		 */
 		if (ref->state == WT_REF_MEM) {
 			++session->nhazard;
+
+			/*
+			 * Callers require a barrier here so operations holding
+			 * the hazard pointer see consistent data.
+			 */
+			WT_READ_BARRIER();
 			return (0);
 		}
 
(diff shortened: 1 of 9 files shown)


src/support/hazard.c
```diff
@@ -103,21 +102,11 @@ __wt_hazard_set(WT_SESSION_IMPL *session, WT_REF *ref, bool *busyp
...
-		 * found this page using the tree's key space, whatever page we
-		 * find here is the page for us to use.)
+		 * state of WT_REF_MEM.
 		 */
 		if (ref->state == WT_REF_MEM) {
-			WT_READ_BARRIER();
-			if (ref->page == page) {
-				++session->nhazard;
-				return (0);
-			}
```
michaelcahill: I don't think we can omit the read barrier here.

Otherwise, we could publish the ref, check `ref->state == WT_REF_MEM`, then see `ref->page == NULL` because that read is scheduled before the `ref->state` check.

michaelcahill:

PR #3182, opened 2016-12-06
WT-3055 Fix timing related failures in Java AsyncOp tests

ddanderson:
When an AsyncOp (java object) is created, cache whether the cursor is "Raw".
Later, when an application is fetching keys/values from the AsyncOp,
the JAVA_CALLBACK object may be detached from the (java) AsyncOp, which makes
finding out information about the cursor impossible.  See the JIRA for more detailed info.


agorrod:
Thanks @ddanderson lgtm.

PR #3184, opened 2016-12-08
WT-3061 Ensure checkpoint syncs are run as part of syscall testing

daveh86:
Not 100% if this is the best way to proceed here, re pwrite vs pwrite64.

If it is, then we would have to exclude some of the Jenkins machines from testing. The alternative is to find a way to check if the system has pwrite or pwrite64 registered in strace.

### DIFF
--- a/test/syscall/syscall.py
+++ b/test/syscall/syscall.py
@@ -636,12 +636,7 @@ class Runner:
         errfile = open(self.errfilename, 'w')
         if self.args.verbose:
             print('RUNNING: ' + str(callargs))
-        # We have to overload the WIREDTIGER_CONFIG to ensure we perform
-        # fdatasync operations on checkpoints
-        my_env = { k : v for k, v in os.environ.iteritems()
-            if k != 'WIREDTIGER_CONFIG' }
-        subret = subprocess.call(
-            callargs, stdout=outfile, stderr=errfile, env=my_env)
+        subret = subprocess.call(callargs, stdout=outfile, stderr=errfile)
         outfile.close()
         errfile.close()
         if subret != 0:
--- a/test/syscall/syscall.py
+++ b/test/syscall/syscall.py
@@ -475,7 +475,7 @@ class Runner:
         if not em:
             self.fail(errline, 'Unknown strace/dtruss output: ' + errline)
             return False
-        gotcall = em.groups()[0]
+        gotcall = re.sub(pwrite_in, pwrite_out, em.groups()[0])
         # filtering syscalls here if needed.  If it's not a match,
         # mark the errline so it is retried.
         if self.strip_syscalls != None and gotcall not in self.strip_syscalls:
--- a/test/syscall/syscall.py
+++ b/test/syscall/syscall.py
@@ -638,8 +638,8 @@ class Runner:
             print('RUNNING: ' + str(callargs))
         # We have to overload the WIREDTIGER_CONFIG to ensure we perform
         # fdatasync operations on checkpoints
-        my_env = os.environ.copy()
-        my_env["WIREDTIGER_CONFIG"] = "checkpoint_sync=true"
+        my_env = { k : v for k, v in os.environ.iteritems()
+            if k != 'WIREDTIGER_CONFIG' }
         subret = subprocess.call(
             callargs, stdout=outfile, stderr=errfile, env=my_env)
         outfile.close()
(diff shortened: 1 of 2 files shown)


daveh86:
@ddanderson, can you have a look at this? I'd love to hear your thoughts.

test/syscall/syscall.py
```diff
@@ -628,7 +628,12 @@ def run(self):
         errfile = open(self.errfilename, 'w')
         if self.args.verbose:
             print('RUNNING: ' + str(callargs))
-        subret = subprocess.call(callargs, stdout=outfile, stderr=errfile)
+        # We have to overload the WIREDTIGER_CONFIG to ensure we perform
+        # fdatasync operations on checkpoints
+        my_env = os.environ.copy()
+        my_env["WIREDTIGER_CONFIG"] = "checkpoint_sync=true"
```
michaelcahill: I'd recommend just `del my_env["WIREDTIGER_CONFIG"]` here -- we should test the defaults, it's only because we run this test with Jenkins and in most other cases want to avoid hammering disks unnecessarily.

ddanderson: @daveh86, I think it would be better to set `wiredtiger_open` options in the program that is running, that is `test/syscall/wt2336_base/main.c`.  Then if we wanted to try other durability or checkpoint options, we could do that by making other programs (e.g. additional subdirectories under test/syscall) .  As much as possible, syscall.py should not know about WiredTiger, it should just be a pattern matching program.

test/syscall/wt2336_base/base.run
```diff
@@ -46,15 +46,15 @@ SYSTEM("Darwin");
...
 
-TRACE("close,fdatasync,fsync,ftruncate,open,pwrite,rename");
+TRACE("close,fdatasync,fsync,ftruncate,open,pwrite64,rename");
 RUN("");
 ...
 OUTPUT("--------------wiredtiger_open");
 // lock == 3
 lock = open("./WiredTiger.lock", O_RDWR|O_CREAT|O_CLOEXEC, 0666);
-pwrite(lock, "WiredTiger lock file\n", 0x15, 0x0);
+pwrite64(lock, "WiredTiger lock file\n", 0x15, 0x0);
```
ddanderson: @daveh86, I think the `pwrite64` changes are fine for Linux but may break running on OSX.  I don't know how much we care about that, but it works on OSX now.  (FTR, a way to resolve this may be to create a `PWRITE()` macro in the .run file, like how `FTRUNCATE` is defined now.  I'd vote for keeping things simple and not do that now.  That is, keep the pwrite64 changes you've made as is.)

test/syscall/syscall.py
michaelcahill: Since `WIREDTIGER_CONFIG` overrides `wiredtiger_open` by design (so sysadmins can override a setting), it isn't sufficient to change this in the C code.

We could just change the Jenkins recipe, since it's Jenkins setting `WIREDTIGER_CONFIG` in the first place...

test/syscall/wt2336_base/base.run
daveh86: I've made a halfway change for this; the code now simply converts all the input pwrite64's to pwrite's using a regex. This made testing work on both OSX and on Linux (the flavours that use pwrite64 at-least).

test/syscall/syscall.py
daveh86: I've made the change Michael suggested. The ```WIREDTIGER_CONFIG``` config mentioned is set Jenkins wide. We could remove it, but that would make it removed from all Jenkins jobs.  At the moment, I lean towards having this over-ride here, as having ```checkpoint_sync``` set either on or off is an implicit assumption for the task.

ddanderson:
@daveh86, can we just remove the WIREDTIGER_CONFIG variable for just this Jenkins job?  I think something like `unset WIREDTIGER_CONFIG` before the `python` command in the set of build commands should do it. Another option would be to do `setenv` (and `_putenv` for Windows) in the `main.c`.  Either of these feels like a better place for it to me.

daveh86:
I've removed the over-ride of WIREDITGER_CONFIG and added the unset to the test.

Can you take another look @ddanderson

ddanderson:
@daveh86, that looks good to me.  I'll let you do the merge since I know you're coordinating w/ a Jenkins job.

PR #3186, opened 2016-12-08
WT-3056 Allow projected table and join cursors to use primary keys

ddanderson:


ddanderson:
@michaelcahill could you review please?

src/cursor/cur_join.c
```diff
@@ -1333,7 +1333,7 @@ __wt_curjoin_open(WT_SESSION_IMPL *session,
 	WT_ERR(__wt_scr_alloc(session, 0, &tmp));
 	if (columns != NULL) {
 		WT_ERR(__wt_struct_reformat(session, table,
-		    columns, strlen(columns), NULL, 1, tmp));
+		    columns, strlen(columns), NULL, false, tmp));
```
agorrod: @ddanderson Does this change mean that projections will always include the primary key? I thought the goal was to make it optional?

ddanderson: @agorrod, projections only include columns specified, so keys are not included unless mentioned.  The parameter flag indicates if keys are allowed.  `__wt_struct_reformat` also processes column lists for column groups, which is why that flag is needed.

agorrod:
Thanks for the explanation @ddanderson. lgtm, I'll merge.

PR #3187, opened 2016-12-08
SERVER-26545 Remove fixed-size limitation on WiredTiger hazard pointers

keithbostic:
SERVER-26545 Remove fixed-size limitation on WiredTiger hazard pointers


keithbostic:
Some comments for review:

1. When growing the hazard array, I put the old arrays in the connection's list of free-on-close pointers. It wouldn't be too difficult to periodically process that list and get free those entries, let me know if you want me to do that.

2. I rewrote the hazard-set code to tighten down the loops, make them faster/smaller. It was a separate commit, if you hate it, kill it.

3. We could probably save some space in the session structure by using `uint16_t` instead of `uint32_t` for the hazard pointer array information. The `WT_SESSION_IMPL` structure is 832 bytes at the  moment, so I assumed we don't care about the space, let me know if that's wrong.

4. I removed a `WT_PUBLISH` that appeared in the original __wt_hazard_set() code, from here:
```
if (hp >= session->hazard + session->hazard_size) {
        if (session->nhazard < session->hazard_size &&
            restarts++ == 0)
                hp = session->hazard;
        else if (session->hazard_size >= conn->hazard_max)
                break;
        else
                WT_PUBLISH(session->hazard_size, WT_MIN(    <<< This WT_PUBLISH
                    session->hazard_size + WT_HAZARD_INCR,
                    conn->hazard_max));
}
``` 
I think it was a cut-and-paste error in the original code, but I wanted to call it out for a close look.


5. There's some code in `__wt_hazard_set`: 
```
/*
 * If there are no available hazard pointer slots, make another one
 * visible.
 */
if (session->nhazard >= session->hazard_inuse) {
        WT_ASSERT(session,
            session->nhazard == session->hazard_inuse &&
            session->hazard_inuse < session->hazard_size);
        hp = &session->hazard[session->hazard_inuse++];    <<< This increment
} 
```
Not ordering that increment in any way bothers me, because it means we're relying on the subsequent WT_FULL_BARRIER, after the hazard reference has been initialized, to make both that increment and the hazard reference visible to the eviction thread. I think it's safe, but it took me some time to convince myself of that and I wanted to call it out for a close look.

michaelcahill:
@keithbostic, I pushed a few minor changes:
* upgrading note for `hazard_max`;
* restored the old behavior of limiting how many trees the eviction server would walk until we have better tests for workloads with tens of thousands of active trees.

I've also pointed some tests at this branch [stress-sanitizer](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/) and performance.  Assuming they don't turn up any surprises, this LGTM.

I'll open a separate project to add general-purpose epoch management: I think that can replace some existing code and could also be used here to figure out when it is safe to free old hazard arrays.

keithbostic:
@michaelcahill, your changes look fine to me.

I've been pounding this change (plus some minor code changes so we always allocate new hazard arrays), on AWS for a few days now, without any failures.

And, I noticed some testing around setting `hazard_max`, and removed it in 2447221.

The only thing that's interesting there is `test_bug011`, which attempted to "Test having eviction working on more files than the number of allocated hazard pointers." by setting `hazard_max` to 25 and then pushing eviction. We could rewrite that test for the `WT_EVICT_MAX_TREES` value by opening more than 1000 files in that test, but that's likely to make it a relatively long-running test. Let me know if you want me to do that.

Ready to merge as far as I'm concerned.

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill: this PR failed to build today, but the failure was unrelated to the hazard changes, I’m chasing it separately. I’ve kicked off another PR build, if it finishes and you’re OK with the test changes I pushed today, I’m good to merge.

michaelcahill:
Thanks @keithbostic, I reintroduced `test_bug011`, bumped the number of tables to 2000 and made it only run with `--long` so it doesn't cause problems day-to-day.  This is now looking good to me, I'll merge...

PR #3188, opened 2016-12-11
WT-3063 Reserve records for read-modify-write.

keithbostic:
@michaelcahill, @agorrod, a possible approach to WT-3063.


michaelcahill:
@keithbostic, this looks pretty much the way I expected.  There are some details I'll work through tomorrow, but it's definitely looking like the right direction to me.

keithbostic:
@michaelcahill, great!

A couple of things I'd like to touch base on:

* In LSM, I'm ignoring everything but the primary chunk: it looked correct to me, but I'm not familiar enough with LSM to be confident.

* In the table cursor, I'm hitting the primary and ignoring the indices. Again, not an area I'm terribly familiar with, and also there's a question of modifying the table update code to update the primary first to get the fastest possible collision with a reserve operation.

* We walk the list of updates on every commit to handle relatively rare conditions: clean up the reserve operations and to free truncate logging memory. It's a linked list so it's not a cheap walk, I was thinking of adding a flag so we only walk the list of updates if we need to do so. That muddies the layering a bit, though. Thoughts?

michaelcahill:
I've pointed perf tests and a stress test at this branch, just to be sure before merging it.  It's generally looking good to me but I need to stare at it some more tomorrow.

michaelcahill:
autobuild: retest this please

michaelcahill:
A `rebalance` test just timed out.  AFAIK, we haven't seen that elsewhere:

  http://build.wiredtiger.com:8080/job/wiredtiger-pull-request-python/1244/console

Any ideas?

keithbostic:
> A rebalance test just timed out.

We haven't seen this elsewhere, and I haven't the slightest clue. I can't find the build, is it possible to know if it was the rebalance test that took too long, or if some other test ran out the clock but managed to finish?

Regardless, I can't think of any reason the reserve changes would affect rebalance.

agorrod:
@keithbostic I'm not sure how this API would fit into an application. My expectation was that applications would use the functionality to avoid needing to rollback a lot of transactions because a single record in a table is hot. Do you expect the use to be:

```
WT_CURSOR *c;
c->set_key(foo);
c->reserve();
/* Do we need another set_key in here? */
c->search();
c->get_value(&blah);
/* Do other stuff, including creating new new_blah from blah */
c->set_key(foo);
c->set_value(new_blah);
c->update();
```

I'd love to see a little sample application we could use to figure out a good flow for the new API, and see how it interacts with the existing ``WT_CURSOR`` methods.

keithbostic:
>My expectation was that applications would use the functionality to avoid needing to rollback a lot of transactions because a single record in a table is hot.

My expectation is similar, but I would generalize to say "avoid rolling back other updates or wasting work because of a subsequent failed update". In other words, an operation might read or update many records before updating a final record and you don't want to waste work if the final update fails due to a collision.

I don't see any reason we'd need a `set_key` between the `WT_CURSOR.reserve` and `WT_CURSOR.search`. We don't currently enforce any consistent "key/value is set" semantics between cursor calls, if the key isn't "unset" by some action, then it's "set" for the next operation (For example, we don't document if `WT_CURSOR.equals` clears a set key or not. Intuitively, I would expect `WT_CURSOR.equals` to have no effect on a currently set key, but our documentation is silent.)

My intuition is the `WT_CURSOR.reserve` call isn't valuable absent a reasonable amount of read/write work between the reserve and the update, so I'd expect to see something like:
```
WT_CURSOR *c;
c->set_key(result);
c->reserve();

for key in a lot of documents:
        c->set_key(key);
        c->search();
        c->get_value();
        /* Modify accumulated value based on this document. */

c->set_key(result);
c->set_value(value);
c->update();
```

That said, reserve is a fast enough call that for contentious records it might even be useful for WiredTiger to internally "reserve" the primary before updating the indices of complex WiredTiger tables so colliding updates fail early (although there might be easier ways to get the same effect).

src/cursor/cur_table.c
```diff
@@ -623,6 +623,29 @@ err:	CURSOR_UPDATE_API_END(session, ret);
...
+static int
+__curtable_reserve(WT_CURSOR *cursor)
+{
+	WT_CURSOR_TABLE *ctable;
+	WT_DECL_RET;
+	WT_SESSION_IMPL *session;
+
+	ctable = (WT_CURSOR_TABLE *)cursor;
+	JOINABLE_CURSOR_UPDATE_API_CALL(cursor, session, update, NULL);
+	WT_ERR(__curtable_open_indices(ctable));
```
michaelcahill: I don't understand why this is necessary, since `reserve` isn't looking at indices.

src/cursor/cur_table.c
```diff
@@ -623,6 +623,29 @@ err:	CURSOR_UPDATE_API_END(session, ret);
...
+
+	/* Reserve in the primary, ignore the indices. */
+	APPLY_CG(ctable, reserve);
```
michaelcahill: In tables, `primary` means the first column group, i.e., `ctable->cg_cursors[0]`.

We could consider doing what the comment says, and only reserving in one column group.  That will only make a difference in complex tables, and the downside is if we one day get smarter about updates to complex tables (only applying the update to column groups that have changed), then primary-only reserve would not be sufficient to guarantee subsequent updates to the record will succeed.

TL;DR this looks right to me.

src/lsm/lsm_cursor.c
```diff
@@ -1458,6 +1458,12 @@ __clsm_put(WT_SESSION_IMPL *session, WT_CURSOR_LSM *clsm,
 	if (position)
 		clsm->current = primary;
 
+	/* Reserve: only update the primary chunk, ignore the record count. */
+	if (reserve) {
+		primary->set_key(primary, key);
+		return (primary->reserve(primary));
+	}
```
michaelcahill: I think we need to reserve on the same set of chunks that we would update in (i.e., move this into the loop below).  Otherwise, we could miss a conflict with a concurrent transaction and a subsequent `update` can fail unexpectedly.

michaelcahill: More specifically, I think reserve should exactly replace the `update` in the loop below and we should continue to call `insert` as we do for updates, because this `insert` is special in that it is only doing a conflict check in overlapping chunks (usually `nupdates == 1`).

src/cursor/cur_table.c
keithbostic: > I don't understand why this is necessary, since reserve isn't looking at indices.

I went back-and-forth on this one, but I think it's reasonable to leave it in: it makes the reserve method's initial cursor code look the same as the other methods, it's not a performance cost because the eventual update will call `__curtable_open_indices` anyway, and there's general error checking in `__curtable_open_indices` (for example, checking for bulk cursors). The reserve method is an update under the covers, and I'd rather the indices be open before we do updates.

src/cursor/cur_table.c
keithbostic: > TL;DR the code looks right to me but the comment should say "Reserve in column groups, ignore indices".

Fixed in commit a55acd0.

src/lsm/lsm_cursor.c
keithbostic: > I think we need to reserve on the same set of chunks that we would update in (i.e., move this into the loop below).

I think I did what you wanted in commit e2d19f7, but please review the change to be sure.

keithbostic:
> Should we make it an error to call reserve without a transaction? I can squint and imagine cases where it could possibly be useful, but they're not obvious. More likely is that someone will try to reserve, do work, then update without a transaction and be surprised...

I agree, and commit 5993f63 makes `WT_CURSOR.reserve` fail outside of a transactional context.

keithbostic:
> We need to decide whether a cursor is positioned by calling reserve. In particular, can the application rely on calling WT_CURSOR::get_value afterwards or not? I would argue that it match update -- that is, the cursor is positioned after a reserve and the value is available.

I can see why that makes sense. (Note we don't have this information when reserve completes, we'll have to turn `WT_CURSOR.reserve` into an update followed by a search, but that's easy enough to do.)

If so, what about renaming `reserve` to be `search_update` or even `search_for_update`? I think that will be clearer: search doesn't expect a value to be set, search always returns not-found if the record isn't there, search makes it clear `WT_CURSOR.reset` should be called, and it's reasonably close to SQL's "select for update".

Let me know what you decide and I'll make this change.

michaelcahill:
@keithbostic, let's talk this through some more.

I think a more common case will be to locate a record somehow (via search or scan), then reserve it.  If we could go back in time, I'd make `update` and `remove` only work that way.  The problem is that we have this ambiguity already where we always do a search regardless of whether the cursor is positioned.

But if the cursor is positioned at the end of a `reserve` operation (and I think it should be), then it's reasonable for the application to call `get_value`, in which case it has to return the same value that would be returned by `search`.

That said, I don't see think need to do another search: the cursor is already positioned, and we must have had the real value at some point.  Ideally reserve would just leave that in place.  At worst, we should be able to call `__cursor_valid` and `__wt_kv_return` to put the original value into `cursor->value`.

keithbostic:
> I think a more common case will be to locate a record somehow (via search or scan), then reserve it.

The usage I've had in my head is there's a "well-known" row in which some work results will be stored, which wouldn't require a preceding search/scan; what's the workflow you're thinking of?

Regardless, I see your point, but making "reserve" behave differently will be confusing.

> But if the cursor is positioned at the end of a reserve operation (and I think it should be), then it's reasonable for the application to call get_value, in which case it has to return the same value that would be returned by search.

Yes.

> That said, I don't see think need to do another search: the cursor is already positioned, and we must have had the real value at some point. Ideally reserve would just leave that in place.

The cursor is positioned, but after the search, it's positioned for an update with the magic reserve cookie, not a normal return value, and after the update, it's positioned for return of the magic reserve cookie. I'd have to stare at some search code to convince myself either of those states are OK for reading a key's value.

Imagine an update list consisting of only aborted values: after positioning the cursor for an update into that list, will a "read" of the cursor's position find the on-page value? I have no idea.

And if we get the return value after the search but before the update, and some other update slips in, then get-value after the reserve call could be different from get-value from subsequent reads. I realize that's valid, but it feels strange for the caller of the reserve operation to see a different value than what other, subsequent readers will see.

I guessed reserve wouldn't be a heavily used operation, I figured re-doing the search wasn't a big deal -- we have the leaf page, it won't even be a tree walk.

keithbostic:
I'm shutting down this branch -- I'm been trying to layer the WT-98 update changes on top of it, the whole mess has gotten pretty incoherent. I'm going to back off and try again.

PR #3189, opened 2016-12-11
WT-3066 lint

keithbostic:
@agorrod, lint.


agorrod:
Thanks @keithbostic lgtm.

PR #3190, opened 2016-12-12
WT-3068 Copy artefacts of test runs in wtperf_run script.

agorrod:


agorrod:
I've switched the Jenkins performance jobs to run on this branch to make sure it works as expected there.

bench/wtperf/runners/wtperf_run.sh
```diff
@@ -99,6 +99,11 @@ while test "$run" -le "$runmax"; do
 	if test "$?" -ne "0"; then
 		exit 1
 	fi
+
+	# Copy the artifacts from the run
+	backup_dir=${home}_$(basename $wttest)_${run}_$(date +"%s")
+	rsync -r -m --exclude=*Log* --exclude=*.wt --exclude=WiredTiger --exclude=*Preplog* --exclude=WiredTiger.lock --exclude=WiredTiger.turtle $home/ $backup_dir
+
```
michaelcahill: FTR, {{$home}} here is the local {{WT_TEST}} directory, completely unrelated to {{$HOME}}, so this is just making a local copy of parts of {{wtperf}}'s database directory.  In Jenkins, it will be cleaned up each time the workspace is refreshed.

michaelcahill: I think if we wanted any archiving across runs, that logic should be in Jenkins (like what we have for test failure tarballs).

michaelcahill:
Thanks @agorrod, LGTM.  I'll merge.

PR #3192, opened 2016-12-13
WT-2833 Add projections to wt dump utility

ddanderson:
When dumping with a projection, the dumped metadata associated with the table is modified so that the list of columns matches the list from the projection, and the value format corresponds to projected format.  We use an open cursor using the projection to obtain the value format to show, that requires a slight reordering of when cursors are opened in the dump utility.

Also fix a problem in the JSON dump cursor which did not handle projections.

### DIFF
--- a/src/cursor/cur_json.c
+++ b/src/cursor/cur_json.c
@@ -369,11 +369,11 @@ __wt_json_unpack_char(u_char ch, u_char *buf, size_t bufsz, bool force_unicode)
  *	of column names.
  */
 void
-__wt_json_column_init(WT_CURSOR *cursor, const char *keyformat,
+__wt_json_column_init(WT_CURSOR *cursor, const char *uri, const char *keyformat,
     const WT_CONFIG_ITEM *idxconf, const WT_CONFIG_ITEM *colconf)
 {
 	WT_CURSOR_JSON *json;
-	const char *p, *end, *beginkey;
+	const char *beginkey, *end, *lparen, *p;
 	uint32_t keycnt, nkeys;
 
 	json = (WT_CURSOR_JSON *)cursor->json_private;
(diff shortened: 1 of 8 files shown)


ddanderson:

src/cursor/cur_json.c
```diff
@@ -400,8 +400,16 @@ __wt_json_column_init(WT_CURSOR *cursor, const char *keyformat,
 			keycnt++;
 		p++;
 	}
-	json->value_names.str = p;
-	json->value_names.len = WT_PTRDIFF(end, p);
+	if ((lparen = strchr(uri, '(')) != NULL) {
+		/* This cursor is a projection. */
+		json->value_names.str = lparen;
+		json->value_names.len = strlen(lparen) - 1;
+		WT_ASSERT((WT_SESSION_IMPL *)cursor->session,
```
agorrod: Should this case be caught and return an explicit error rather than an assertion, since its a string passed in through the API? Or is already caught earlier?

src/utilities/util_dump.c
```diff
@@ -225,16 +251,107 @@ dump_json_table_end(WT_SESSION *session)
...
+	char *newconfig;
+	const char *keyformat, *p;
+
+	len = strlen(config) + strlen(cursor->value_format) +
+	    strlen(cursor->uri) + 20;
+	if ((newconfig = malloc(len)) == NULL)
+		return util_err(session, errno, NULL);
+	*newconfigp = newconfig;
+
+#define	DUMP_ADD_CONFIG(fmt, ...) do {					\
```
agorrod: Would you mind implementing this using a function instead of a macro?

src/cursor/cur_json.c
ddanderson: This is caught when doing a general syntax check of the URI.  I added a test case to make sure.

src/utilities/util_dump.c
ddanderson: I turned the macro into a function, just pushed the commit.

ddanderson:
@agorrod, can you take another look please?

PR #3193, opened 2016-12-13
WT-3069 LevelDB APIs no longer build.

keithbostic:
WT-3069 LevelDB APIs no longer build.


agorrod:
Thanks @keithbostic lgtm, and builds for me now.

PR #3195, opened 2016-12-14
WT-3070 Fix search_near on indexes.

michaelcahill:
Since WT-2773, the sign of the `exact` result was flipped and
WT_NOTFOUND was returned for search_near calls with a key past the end
of the index.


michaelcahill:
@ddanderson, can you please review?

ddanderson:
@michaelcahill, lgtm. I added a couple more tests, I'll merge.

PR #3198, opened 2016-12-14
WT-3078 test reconfiguration hang in the statlog server.

keithbostic:
Flush the caller's flags before reading the condition information, we may not write anything before joining the exiting thread, and a write-barrier() would not necessarily have any effect.


keithbostic:
@michaelcahill, the only thing I can see here is we're doing no writes at all and the write-barrier has no effect, and/or the read of the condition variable is getting moved to before the write-barrier(). Thoughts?

michaelcahill:
@keithbostic, I can squint and convince myself that a full barrier is needed before we read the condition variable state.  That said, I'm still concerned that something suprising happens as a result of reading `cond->waiters` multiple times, so we'll have to keep an eye out for any more hangs in that test.

PR #3199, opened 2016-12-15
WT-3079 Always remember the last file eviction visited.

michaelcahill:


michaelcahill:
I don't know that this is a problem, but it looks to me as if the eviction server can systematically clear where it is up to in the list of btree handles (if we give up on the last tree we walk, we'll call `__evict_clear_walk`, then next time start with the first tree again).

Pushing a branch for further discussion.

agorrod:
@michaelcahill This change almost exactly reverses a change in WT-2313 - looks like that was intended to be a bug fix, but I can't spot the bug.

michaelcahill:
FTR, in my testing this change makes eviction behave more like what I expect: with more trees than it is permitted to walk simultaneously, the eviction server keeps walks in the maximum number most of the time.  Unfortunately, that also introduces regressions into some workloads, which needs more investigation independent of deciding if this change is safe.

michaelcahill:
Closing: as noted, this reverts a change from WT-2313 that fixed a race between eviction and sweep.  We have bigger fish to fry.

PR #3200, opened 2016-12-15
WT-283 Fix data source examples to include alter.

sueloverso:
Fix some places using data sources to include alter.  Found after merge.


sueloverso:
@keithbostic Can you review these two fixes for session->alter on data sources found after the merge?

keithbostic:
lgtm.

PR #3201, opened 2016-12-15
WT-3080 Python test suite: add elapsed time for tests

ddanderson:
This does increase the amount of console output, but having an effective way to reduce the test suite time seems worth it.


ddanderson:
@daveh86, can you take a look at this please?  @keithbostic was getting a timeout in the test suite during a pull request, and suggested some way to measure where we're spending our time.  I'll follow up on another ticket with some suggestions for putting certain tests into the --long run.

ddanderson:
BTW, this is a simplistic solution, perhaps you have better ideas?

ddanderson:
retest this please

daveh86:
Simple works for me. LGTM.

PR #3202, opened 2016-12-15
WT-3068 Have Jenkins include specific files for copy rather than exclude

agorrod:


agorrod:
I'm going to go ahead and self review here. Don't save LSM chunks - they can take up a lot of space.

PR #3206, opened 2016-12-16
WT-3082 Python test suite: shorten default run to avoid timeouts.

ddanderson:
Tests that have lots of scenarios will have the number of scenarios pruned in the default case.  Running with --long restores the previous larger number of scenarios.

This runs about 25% faster in my standalone testing.

### DIFF
--- a/test/suite/test_backup03.py
+++ b/test/suite/test_backup03.py
@@ -74,7 +74,7 @@ class test_backup_target(wttest.WiredTigerTestCase, suite_subprocess):
         ('backup_9', dict(big=3,list=[])),         # Backup everything
     ]
 
-    scenarios = make_scenarios(list)
+    scenarios = make_scenarios(list, prune=3, prunelong=1000)
     # Create a large cache, otherwise this test runs quite slowly.
     conn_config = 'cache_size=1G'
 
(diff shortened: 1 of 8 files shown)


ddanderson:
retest this please

ddanderson:
The `wiredtiger-pull-request-python` build on tinderbox previously took 7:04, now it takes 5:07.  Now, this isn't going to improve pull request test times, but it has benefits.  First, the python test suite is not in the critical path and probably won't be for a while, we can add more tests and extend existing tests without lengthening pull request test times. Also, I sometimes run the entire test suite on my local box, this will certainly be faster.  This can also be overridden in a selective way, for example:  `python run.py --long -v 2 backup03` will run the long suite of  `test_backup03` tests.

ddanderson:
@daveh86, can you review please?

daveh86:
The changes make sense to me, so LGTM.

I've assumed the 1000 value for prunelong is just to run all the scenarios in long mode. Is that correct?

Do we know how many scenarios the touched tests run? Just to get an idea of how much trimming is occurring?

ddanderson:
@daveh86, I've answered your questions and filled in some more details in the JIRA: https://jira.mongodb.org/browse/WT-3082

PR #3207, opened 2016-12-19
WT-2771 Add a statistic to track per-btree dirty cache usage.

michaelcahill:


michaelcahill:
Note that there is also some reordering to a couple of lines here so that we're consistently updating the btree stat, then the cache-wide stat.

agorrod:
Thanks @michaelcahill. This change lgtm - should I merge, or are you thinking about adding more statistics as well?

Looking at this also made me wonder if we should be skipping clean trees in eviction when only looking for dirty candidates.

michaelcahill:
Thanks @agorrod, this is the whole change for this ticket so please go ahead and merge.

w.r.t. skipping trees, my hope is that the `target_pages` calculation (which takes into account clean vs dirty) will mean that we skip clean trees 99% of the time when looking for dirty pages.

PR #3209, opened 2016-12-19
WT-2402 Fix setting the Autoconf options too late for strict checking

daveh86:


daveh86:
@sueloverso would you mind testing this on your OSX machine to confirm this resolves the issue you were seeing?

sueloverso:
Thanks @daveh86.  It builds quietly now until it fails on the `util_dump.c` issue that is another ticket.  That is a success for this issue!

PR #3211, opened 2016-12-20
WT-3092 Quiet autogen.sh, change local Python build permissions

keithbostic:
WT-3092 Quiet autogen.sh, change local Python build permissions

### DIFF
--- a/build_posix/aclocal/options.m4
+++ b/build_posix/aclocal/options.m4
@@ -57,7 +57,7 @@ AH_TEMPLATE(
     HAVE_CRC32_HARDWARE, [Define to 1 to configure CRC32 hardware support.])
 AC_MSG_CHECKING(if --enable-crc32-hardware option specified)
 AC_ARG_ENABLE(crc32-hardware,
-	AC_HELP_STRING([--enable-crc32-hardware],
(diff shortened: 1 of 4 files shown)


michaelcahill:
@keithbostic, it seems dangerous to install Python code into system locations with different permissions that the sysadmin expects.  I never install the Python modules as part of my ordinary development process: what problem is that solving?

keithbostic:
@michaelcahill 
>  it seems dangerous to install Python code into system locations with different permissions that the sysadmin expects. I never install the Python modules as part of my ordinary development process: what problem is that solving?

Agreed, I didn't read the `setup.py` documentation closelly enough, I thought there was a separate install step from the build step.

Anyway, the problem is you can't clean the tree after a make install, so the sequence
```
% make
% sudo make install
% make clean
```
fails because there are files owned by root the user can't remove.

That's annoying enough to my workflow I noticed it.

I pushed a different change in 1f84fb1 to remove the build directory after the install, instead.

If you don't like that change either, let me know and I'll revert both versions of that change and merge the rest.

michaelcahill:
Thanks @keithbostic, fbm.  I'll merge.

(I still think we should figure out why you need to "make install", but let's do that separately).

PR #3213, opened 2016-12-20
WT-3039 Add previous LSN record to log files

sueloverso:
Not ready for review.  Add a new log record to record the end of the previous log to detect a hole in records at the end of a log file.


sueloverso:
autobuild: retest this please

sueloverso:
@agorrod this is ready for review.  The long wtperf runs show that it solves both the large number and duration of latency issues introduced from WT-3000.  Another performance "issue" is discussed in the ticket for this.

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

sueloverso:
@agorrod and @michaelcahill don't merge this yet.  I started a test loop of the `random-abort` recovery test and it uncovered an error path bug after several hundred iterations.  If it dies after writing a log file header but before writing the new system record, it spits out an error message and won't continue recovery.  That needs to be fixed.

sueloverso:
@agorrod This is ready for an in-depth review.  Given that I found two bugs since the on-site visits, it definitely needs review!

src/conn/conn_log.c
```diff
@@ -40,6 +40,116 @@ __logmgr_sync_cfg(WT_SESSION_IMPL *session, const char **cfg)
...
+		 * progress with their writes.
+		 */
+		WT_RET(tmp_session->iface.checkpoint(
+		    &tmp_session->iface, "force=1"));
+		/*
+		 * We need to wait for the archival to reach up to the
+		 * log number needed to be assured no previous version
+		 * log files exist.  There could be outstanding writes or a
+		 * cursor open preventing archiving.
+		 */
```
agorrod: Is there a situation that could cause the file to not move forward to lognum? Is forcing the checkpoint guaranteed to roll over the log file?

I don't understand why an open cursor would prevent archiving? And when you say "outstanding write" do you mean a write that has been issued to the filesystem but not processed?

src/conn/conn_log.c
```diff
@@ -40,6 +40,116 @@ __logmgr_sync_cfg(WT_SESSION_IMPL *session, const char **cfg)
...
+		if (!first)
+			__wt_yield();
+		WT_RET(WT_SESSION_CHECK_PANIC(tmp_session));
+		WT_RET(__wt_log_truncate_files(tmp_session, NULL, true));
+		first = false;
```
agorrod: This loop deserves a statistic - it seems unlikely that we'd get stuck in here, but I can't think how we'd figure it out for a MongoDB user without a statistic.

src/conn/conn_log.c
sueloverso: Actually the earlier code {{wt_log_set_version}} will force a roll over into the changed-version log file.  We then force a checkpoint in the new log file so that all old ones can be archived.

An open `log:` cursor that is positioned on an LSN as it walks the log prevents all archiving.  We have to wait for any open log cursor to close.

And by 'outstanding write' I mean a group of earlier threads that are using an earlier consolidated slot that they're still copying into, and will eventually write to the OS.

sueloverso: I'll try to make some clarifying changes to the comments.

src/conn/conn_log.c
sueloverso: Good point.  I'll add a new statistic.  I'll also be busy merging up with the release change too.

sueloverso:
Don't merge this until WT-3355 is resolved.

agorrod:
@sueloverso I've switched the Jenkins http://build.wiredtiger.com:8080/job/wiredtiger-configure-combinations/ and http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/ jobs over to this branch in preparation for merging.

agorrod:
The configure combinations job failed, but that's possibly due to a recent configuration change to the job. We should try again once this is up to date on develop.

sueloverso:
I merged current develop into this branch and retried the `wiredtiger-configure-combinations` and it passes now.  I ran it twice and then reset the job back to develop.  The `wiredtiger-test-format-stress-sanitizer` has successfully run about 20 times on the branch.  I reset that back to develop also.

agorrod:
@sueloverso Please go ahead and merge this branch when you are ready.

PR #3218, opened 2016-12-23
WT-3099 lint: static function declarations, non-text characters in documentation

keithbostic:
lint: static function declarations, non-text characters in documentation


michaelcahill:
Thanks @keithbostic, I don't love the loss of information about the error in the utility code, but we already have that issue.  I'll merge.

PR #3219, opened 2016-12-27
WT-3100 test bug: format is weighted to delete, insert, then write operations.

keithbostic:
test bug: format is weighted to delete, insert, then write operations.


keithbostic:
Note to reviewer: I deleted some configuration code that skewed LSM workloads to insert, I couldn't think of a reason we should constrain LSM workloads that way.
```
/*
 * If this is an LSM run, set the cache size and crank up the insert
 * percentage.
 */
if (DATASOURCE("lsm")) {
        if (!config_is_perm("cache"))
                g.c_cache = 30 * g.c_chunk_size;

        if (!config_is_perm("insert_pct"))
                g.c_insert_pct = mmrand(NULL, 50, 85);
}
```

michaelcahill:

PR #3221, opened 2016-12-29
WT-3104 Fix single threaded eviction configurations.

sueloverso:


sueloverso:
self-reviewing to get the correct configs running in jenkins.

PR #3222, opened 2016-12-30
SERVER-16796 Add recovery messages for progress.

sueloverso:
@michaelcahill and @agorrod I have two branches for this work.  One uses the `wt_progress` call and the other `wt_msg`.  I'm on the fence which I prefer, but I'm leaning toward this one.  Having the ability to give an actual message is both simpler and can be more informative.  I went with something pretty basic.  If you're aware of more information wanted let me know.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -594,7 +594,10 @@ wiredtiger_open_log_configuration = [
         Config('recover', 'on', r'''
             run recovery or error if recovery needs to run after an
             unclean shutdown''',
-            choices=['error','on'])
+            choices=['error','on']),
+        Config('recover_progress', 'true', r'''
+            print messages indicating progress of recovery''',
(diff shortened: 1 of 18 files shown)


dist/api_data.py
```diff
@@ -594,7 +594,10 @@ def __cmp__(self, other):
         Config('recover', 'on', r'''
             run recovery or error if recovery needs to run after an
             unclean shutdown''',
-            choices=['error','on'])
+            choices=['error','on']),
+        Config('recover_progress', 'true', r'''
+            print messages indicating progress of recovery''',
+            type='boolean'),
```
sueloverso: Using `wt_msg` means I had to make this configurable in order to make it quiet in order for the tests to pass.

sueloverso:
Close this in favor of using verbose messages.

PR #3223, opened 2017-01-02
WT-3137 Fix a hang in logging due to a race condition

sueloverso:
Not for merge, just for debugging for now.


sueloverso:
@agorrod This is ready for review.  It has run for a few days on my AWS box and passed on Bruce's as well.  Some things to keep in mind:
- The real fix is in `log_slot_activate` where we initialize the `slot_state` last.
- Other changes reduce but do not eliminate the chance of using a stale pointer.  I refactored a bunch of `log_slot_join` to clean up things that don't change during the loop.  I added a pool index to rotate among the slots round-robin rather than always starting at 0.
- I left in all the debugging I added under HAVE_DIAGNOSTIC.  That includes yields in key places to encourage races and the 10-second hang detection and slot dumping.  I may have been over-zealous in leaving that in.

agorrod:
@sueloverso I've take a first look at this today, but I'd like the weekend to digest. Sorry for the slow turnaround.

src/log/log_slot.c
```diff
@@ -30,12 +72,25 @@ __wt_log_slot_activate(WT_SESSION_IMPL *session, WT_LOGSLOT *slot)
...
 	slot->slot_last_offset = log->alloc_lsn.l.offset;
 	slot->slot_fh = log->log_fh;
 	slot->slot_error = 0;
-	slot->slot_unbuffered = 0;
+#ifdef	HAVE_DIAGNOSTIC
+	/*
+	 * Yield here to encourage races during slot switches.
+	 */
+	__wt_yield();
+#endif
```
michaelcahill: Suggest `WT_DIAGNOSTIC_YIELD`.

src/log/log_slot.c
```diff
@@ -440,6 +560,7 @@ __wt_log_slot_join(WT_SESSION_IMPL *session, uint64_t mysize,
 		WT_ASSERT(session, slot->slot_unbuffered == 0);
 		WT_STAT_CONN_INCR(session, log_slot_unbuffered);
 		slot->slot_unbuffered = (int64_t)mysize;
+		WT_WRITE_BARRIER();
```
michaelcahill: A comment would help: what writes need to be ordered here?

src/log/log_slot.c
```diff
@@ -30,12 +72,25 @@ __wt_log_slot_activate(WT_SESSION_IMPL *session, WT_LOGSLOT *slot)
...
+	 */
+	__wt_yield();
+#endif
+	WT_BARRIER();
+	/*
+	 * Set the slot state last.  Other threads may have a stale pointer
+	 * to this slot everything else needs to be there before the state
+	 * is cleared.  Use a barrier to make sure it is last.
+	 */
+	slot->slot_state = 0;
```
michaelcahill: It isn't clear to me that the `WT_BARRIER` above guarantees anything in terms of correctness.  (I suspect the compiler is generating code in the order specified here in the absence of any reason to reorder, but the `WT_BARRIER` doesn't guarantee that cores will see writes in the order they are issued).

Do you mean `WT_PUBLISH(slot->slot_state, 0);`?

src/log/log_slot.c
```diff
@@ -101,9 +160,32 @@ __log_slot_close(
...
+				__wt_epoch(session, &now);
+				if (WT_TIMEDIFF_SEC(now, begin) > 10) {
+					__wt_errx(session, "SLOT_CLOSE: Slot %"
+					PRIu32 " Timeout unbuffered, state 0x%"
+					PRIx64 " unbuffered %" PRIu64,
+					(uint32_t)(slot - &log->slot_pool[0]),
+					slot->slot_state,
+					slot->slot_unbuffered);
+					__log_slot_dump(session);
+					abort();
```
michaelcahill: Should be `__wt_abort(session);`

src/log/log_slot.c
```diff
@@ -370,53 +479,65 @@ __wt_log_slot_join(WT_SESSION_IMPL *session, uint64_t mysize,
...
+			join_offset = WT_LOG_SLOT_JOINED(old_state);
+			if (unbuffered)
+				new_join = join_offset + WT_LOG_SLOT_UNBUFFERED;
+			else
+				new_join = join_offset + (int32_t)mysize;
+			new_state = (int64_t)WT_LOG_SLOT_JOIN_REL(
+			    (int64_t)new_join, (int64_t)released,
+			    (int64_t)flag_state);
+
+			slot = log->active_slot;
```
michaelcahill: Is there a race possible here when reading `log->active_lot`?  In other words, does it matter if `slot` ends up with a different location than we read `old_state` from above?

src/log/log_slot.c
sueloverso: Fixed

src/log/log_slot.c
sueloverso: Fixed.

src/log/log_slot.c
sueloverso: Fixed.

src/log/log_slot.c
sueloverso: No it doesn't matter and there is no way to prevent it.  If `old_state` matches the state everything is fine, if it doesn't then we try again.

src/log/log_slot.c
sueloverso: done.

sueloverso:
autobuild: retest this please

src/log/log_slot.c
michaelcahill: OK, but is there any disadvantage to setting `slot` above line 501 (the way it used to be)?  As written, I feel like it almost needs a comment explaining why we're not doing that...

michaelcahill:

PR #3224, opened 2017-01-03
WT-3083 For wtperf config dump, concat config options in multiple lines

sulabhM:


sulabhM:
@autobuild :  retest this please

sulabhM:
@daveh86 , please have a look

bench/wtperf/wtperf.c
```diff
@@ -2502,52 +2502,105 @@ main(int argc, char *argv[])
...
+
+		append_comma = false;
+		pos = 0;
+		req_len = 20;
+		req_len += (debug_cconfig ? strlen(debug_cconfig) : 0);
+		req_len += (user_cconfig ? strlen(user_cconfig) : 0);
+		req_len +=
+		    (wtperf->async_config ? strlen(wtperf->async_config) : 0);
+		req_len +=
+		    (wtperf->compress_ext ? strlen(wtperf->compress_ext) : 0);
```
keithbostic: Stylistic question, test should compare against `NULL`, it's not a boolean.

keithbostic:
@sulabhM, can you add a paragraph in this PR that talks about the problem that's being fixed? WT-3083 just says there's a problem, but there's nothing talking about what's fundamentally going wrong.

sulabhM:
wtperf.c main() has logic that generates non default configuration strings. These generated strings are added as an entry in the queue "opts->config_head". This ticket fixes two issues:

1. The above code even when no non-default configuration to generate, creates empty configuration with commas, something like ",,,". This in itself is a bug
2. The above code generates an entry and adds to the queue (an empty comma filled entry if no non-default configuration was generated). config_consolidate() is used to consolidate queue entries with same keys. Consolidation is done before dumping the configuration in WT_TEST/CONFIG.wtperf. Consolidation function looks for entries with same configuration key, and discards all except the last one in the queue. (this would be empty ",,," in most cases). Consolidate should concatenate the strings and not discard the previous ones, because WiredTiger would have actually applied them during the test.

daveh86:
Another thing, can we change the title to reflect that this changes wtperf only?

daveh86:
This LGTM.

agorrod:
@sulabhM These changes look reasonable to me. I think it'd be good to add tests verifying that ``CONFIG.wtperf`` file corresponds to the expected content. You could either add something to Jenkins (my preference) or to ``bench/wtperf/smoke.sh`` so that ``make check`` tests it.

I'd like those tests to be added before the pull request is merged.

sulabhM:
@daveh86 : This changed significantly, also python test got added. Can you have a look at this change again.

daveh86:
The new change LGTM. But I found one issue when trying to run this with --enable-snappy, rather than --with-builtins=snappy.

Can you make a change so that the script changes its working directory to that of the wtperf binary?

agorrod:
Thanks @sulabhM lgtm

PR #3225, opened 2017-01-03
SERVER-16796 Recovery progress via verbose messages.

sueloverso:
@michaelcahill Here's the change using verbose.  This is much better than the other two (I'm going to close those PRs).  This should be reviewed and merged before WT-3039 so that it can get into the mongo 3.4 branch before that version change.  There will need to be a mongo change to turn on the verbose setting too, separate from this.


michaelcahill:

PR #3226, opened 2017-01-04
WT-3108 Add disk memory size to verify debug output

sulabhM:


sulabhM:
verify output looks like following after the change:

0x6f77d0: row-store internal
        disk 0x78db70, mem_size 60447, entries 1895, clean, unlocked, disk-alloc
        K {#00}
        0x7a06d0 disk [1129979904-1130012672, 32768, 4245046585]
        K {0000000000000000682}
        0x7a0710 disk [1339924480-1339957248, 32768, 3546965311]
        K {0000000000000001363}
        0x7a0750 disk [1125883904-1125916672, 32768, 850466594]

sulabhM:
@michaelcahill, would you mind reviewing this one.

michaelcahill:
fbm -- @keithbostic, you've been the primary maintainer of debug output code: is this change okay with you?

sulabhM:
The motivation to add the page size to verify output is being able to use that information to compare each page's size with max leaf page size set. I have been using that information for evaluating my changes for WT-2439.

src/btree/bt_debug.c
```diff
@@ -679,8 +679,9 @@ __debug_page_metadata(WT_DBG *ds, WT_REF *ref)
 	}
 
 	WT_RET(ds->f(ds, ": %s\n", __wt_page_type_string(page->type)));
-	WT_RET(ds->f(ds,
-	    "\t" "disk %p, entries %" PRIu32, (void *)page->dsk, entries));
+	WT_RET(ds->f(ds, "\t" "disk %p, mem_size %" PRIu32,
+	    (void *)page->dsk, page->dsk->mem_size));
+	WT_RET(ds->f(ds, ", entries %" PRIu32, entries));
```
keithbostic: @sulabh, I think this will drop core if the page was created in memory and `page->dsk` is `NULL`.

sulabhM: In the latest commit, I have put a check for page->dsk to be not null.

sulabhM:
@keithbostic , addressed your comments. Please have another look.

keithbostic:
lgtm.

PR #3230, opened 2017-01-04
Correct snapshot documentation on transactions page.

milkie:


wtbuild:

sueloverso:
@milkie The change itself is fine.  However we only merge into the develop branch and it has to have a ticket associated with it.  The master branch is only updated from a release via develop.  I can create the ticket and PR if you'd like.

milkie:
Oops.  I can take care of that!  Let me try again.

PR #3232, opened 2017-01-05
WT-3110 New WT Utility tests

daveh86:
Not 100% ready yet.

I need to think about what else can be done with stats.

### DIFF
--- a/test/suite/test_util17.py
+++ b/test/suite/test_util17.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python
+#
+# Public Domain 2014-2017 MongoDB, Inc.
+# Public Domain 2008-2014 WiredTiger, Inc.
+#
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
+# distribute this software, either in source code form or as a compiled
+# binary, for any purpose, commercial or non-commercial, and by any
+# means.
+#
+# In jurisdictions that recognize copyright laws, the author or authors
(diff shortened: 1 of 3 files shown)


daveh86:
@sulabhM can you take a look at this?

test/suite/test_util17.py
```diff
@@ -0,0 +1,55 @@
...
+class test_util17(wttest.WiredTigerTestCase, suite_subprocess):
+    tablename = 'test_util17.a'
+
+    def test_stat_process(self):
+        """
+        Test alter in a 'wt' process
+        """
+        params = 'key_format=S,value_format=S'
+        outfile = "wt-stat.out"
+        expected_string = "cursor: cursor create calls=12"
```
sulabhM: Please put a comment why "12".

daveh86:

test/suite/test_util17.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_stat_process(self):
+        """
+        Test stat in a 'wt' process
+        This test is just here to confirm that stat produces a correct looking
+        output, it isn't here to do statistics validation.
+        """
+        params = 'key_format=S,value_format=S'
+        outfile = "wt-stat.out"
+        expected_string = "cursor: cursor create calls=12"
```
agorrod: This is the only thing that worries me in the test. Could you exclude the ``12``? I'm not sure why it is (or will always be) 12.

agorrod:
Thanks @daveh86 lgtm

PR #3233, opened 2017-01-05
WT-3112 Add lock statistics to try_lock path.

sueloverso:
@keithbostic Please review this change to add the locking stats to the try-lock path.  I ran the evict-btree.wtperf workload on develop and this branch.  Develop acquires the dhandle list lock 47 times.  This branch is 116721 acquisitions.


keithbostic:
lgtm.

PR #3236, opened 2017-01-09
WT-3115 Convert the dhandle list lock into a read/write lock.

agorrod:
It was a spinlock, but most acquirers only need shared access and it can be a contetion point in many-table workloads.

This change isn't ready for review yet - I've pushed it up to get some more testing before I iron out some stats and lock-upgrade questions.


agorrod:
@michaelcahill We have talked about this change - I'm still crafting a workload that demonstrates contention on the lock, but we know they exist.

Would you mind taking a look and let me know if you think it's heading in the direction you expected?

michaelcahill:
@agorrod, happy to talk this through -- I'll try to catch you on Tuesday morning.  This generally looks good, thanks!

The main thing that surprised me here was the low-level reacquisition of the lock in write mode.  What I'd imagined was another pass in `__wt_session_get_btree` that tries a read-only lookup of an existing handle before getting a write lock in order to open the tree for real.  I'm probably missing something, but that seems cleaner to me than having the macro do the work.

michaelcahill:
@agorrod, I'm happier with the shape of the change now, but it will need further review and discussion.

michaelcahill:
In particular, switching to a rwlock conflicts with @sueloverso's recent change to track time spent waiting on the handle list lock.  That needs to be resolved properly before this change is really ready for review.

src/txn/txn_ckpt.c
```diff
@@ -640,9 +640,8 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_ASSERT(session, session->ckpt_handle_next == 0);
 	WT_WITH_SCHEMA_LOCK(session,
 	    WT_WITH_TABLE_LOCK(session,
-		WT_WITH_HANDLE_LIST_LOCK(session,
-		    ret = __checkpoint_apply_all(
-		    session, cfg, __wt_checkpoint_get_handles, NULL))));
+		ret = __checkpoint_apply_all(
```
agorrod: Will switching to a read-lock that is released during the traverse mean that checkpoint can potentially retrieve an inconsistent set of handles during this traverse? The traverse is done before the checkpoint transaction has started, so maybe this change doesn't make anything less consistent than it was before? Or does the schema lock protect against parallel operations that would interfere?

src/cursor/cur_backup.c
```diff
@@ -346,13 +346,9 @@ __backup_stop(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb)
...
 __backup_all(WT_SESSION_IMPL *session)
 {
-	WT_DECL_RET;
-
 	/* Build a list of the file objects that need to be copied. */
-	WT_WITH_HANDLE_LIST_LOCK(session, ret =
-	    __wt_meta_apply_all(session, NULL, __backup_list_uri_append, NULL));
-
-	return (ret);
+	return (__wt_meta_apply_all(
```
agorrod: Is a read lock that is re-acquired during the traverse enough to guarantee the consistent set of handles backup is looking for?

src/txn/txn_ckpt.c
michaelcahill: @agorrod, I believe the schema lock protects against creates / drops during this walk.

Also note that there is already some ambiguity about what set of files are include in a checkpoint (if a file is opened for the first time and dirtied in between here and the transaction starting, it won't be included in the checkpoint).  I'd like to leave fixing that out of scope for this change.

src/cursor/cur_backup.c
michaelcahill: I believe so: backup holds the schema lock here and is walking the metadata to find its list of files.

michaelcahill:
@agorrod, is there any new testing or other steps I should consider for this change?

src/conn/conn_dhandle.c
```diff
@@ -97,7 +95,9 @@ __wt_conn_dhandle_find(
 	conn = S2C(session);
 
 	/* We must be holding the handle list lock at a higher level. */
-	WT_ASSERT(session, F_ISSET(session, WT_SESSION_LOCKED_HANDLE_LIST));
+	WT_ASSERT(session,
+	    F_ISSET(session, WT_SESSION_LOCKED_READ_HANDLE_LIST |
```
agorrod: Is it worth creating a synonym of ``WT_SESSION_LOCKED_READ_HANDLE_LIST | WT_SESSION_LOCKED_WRITE_HANDLE_LIST`` as ``WT_SESSION_LOCKED_HANDLE_LIST``?

src/conn/conn_handle.c
```diff
@@ -64,6 +63,8 @@ __wt_connection_init(WT_CONNECTION_IMPL *conn)
 	WT_RET(__wt_spin_init(session, &conn->turtle_lock, "turtle file"));
 
 	/* Read-write locks */
+	/* TODO: Add stats tracking back in */
```
agorrod: Remove TODO

src/conn/conn_sweep.c
```diff
@@ -233,7 +233,11 @@ __sweep_remove_handles(WT_SESSION_IMPL *session)
 		if (!WT_DHANDLE_CAN_DISCARD(dhandle))
 			continue;
 
-		WT_WITH_HANDLE_LIST_LOCK(session,
+		/*
+		 * The read lock we take here will be upgraded to a write lock
```
agorrod: This comment is no longer accurate

src/evict/evict_lru.c
```diff
@@ -24,40 +24,48 @@ static int  __evict_walk_file(
...
 
 	conn = S2C(session);
 	cache = conn->cache;
 	dh_lock = &conn->dhandle_lock;
-	stats = (int64_t **)conn->stats;
+	/*
+	 * We want to know if a lot of time is spent waiting for this readlock
+	 * but don't (yet) have stats for read/write locks.
+	 */
+#if 0 /* XXX */
```
agorrod: As a concrete suggestion, I think we should remove the three handle-list lock stats:
```

    LockStat('lock_handle_list_count', 'handle-list lock acquisitions'),
    LockStat('lock_handle_list_wait_application', 'handle-list lock application thread wait time (usecs)'),
    LockStat('lock_handle_list_wait_internal', 'handle-list lock internal thread wait time (usecs)'),
```

And add a new stat, ``Eviction server time spent waiting for handle list lock (usecs)``, and re-enable this code.

src/include/dhandle.h
```diff
@@ -37,6 +37,23 @@
...
+    (void)__wt_atomic_add32(&dhandle->session_ref, 1)
+
+#define	WT_DHANDLE_RELEASE(dhandle)					\
+    (void)__wt_atomic_sub32(&dhandle->session_ref, 1)
+
+#define	WT_DHANDLE_NEXT(session, dhandle, head, field) do {		\
+	if (dhandle == NULL)						\
+		dhandle = TAILQ_FIRST(head);				\
+	else {								\
+		    WT_DHANDLE_RELEASE(dhandle);			\
```
agorrod: The ordering is safe because a handle list read-lock needs to be held - maybe we could assert that in the macro?

src/lsm/lsm_tree.c
```diff
@@ -520,11 +521,11 @@ __wt_lsm_tree_get(WT_SESSION_IMPL *session,
...
 	WT_DECL_RET;
 
-	WT_ASSERT(session, F_ISSET(session, WT_SESSION_LOCKED_HANDLE_LIST));
-
-	ret = __lsm_tree_find(session, uri, exclusive, treep);
+	WT_WITH_HANDLE_LIST_READ_LOCK(session,
+	    ret = __lsm_tree_find(session, uri, exclusive, treep));
 	if (ret == WT_NOTFOUND)
-		ret = __lsm_tree_open(session, uri, exclusive, treep);
+		WT_WITH_HANDLE_LIST_WRITE_LOCK(session,
```
agorrod: With this change, is it possible that multiple threads will try to open the tree sequentially? i.e: multiple threads could get WT_NOTFOUND returns from the ``__lsm_tree_find`` call, and all attempt to open the tree. Maybe they need to re-check find once the write lock is acquired.

Hmm - answering my own question - the ``__lsm_tree_open`` implementation already does that. Maybe adding a comment to that effect would be worthwhile.

agorrod:
@michaelcahill I believe this one is ready to go - I'll leave you to merge, since I made the last round of changes.

dist/flags.py
```diff
@@ -117,13 +117,14 @@
         'SESSION_CAN_WAIT',
         'SESSION_INTERNAL',
         'SESSION_LOCKED_CHECKPOINT',
-        'SESSION_LOCKED_HANDLE_LIST',
         'SESSION_LOCKED_METADATA',
         'SESSION_LOCKED_PASS',
+        'SESSION_LOCKED_READ_HANDLE_LIST',
```
sueloverso: What would you think of `SESSION_LOCKED_HANDLE_LIST_READ` and `SESSION_LOCKED_HANDLE_LIST_WRITE` so they sort together?

agorrod:
@michaelcahill There is a real problem in this branch, the failing stack is:

```

#1  0x0000003e6ca3665a in abort () from /lib64/libc.so.6
#2  0x00007fc6b5dfa8b4 in __wt_open (session=0x3e6caf329d <write+45>, name=0xb9b4f8 "0X\270", file_type=WT_FS_OPEN_FILE_TYPE_CHECKPOINT, flags=454301078,
    fhp=0x588e941a) at ../src/os_common/os_fhandle.c:250
#3  0x00007fc6b5e3ef6f in CityHash64 (s=0xb9b4f8 "0X\270", len=140491432067936) at ../src/support/hash_city.c:338
#4  0x00007fc6b5dee5d3 in __wt_lsm_tree_truncate (session=0x7fc6b5e6f757, name=0xb9b4f8 "0X\270", cfg=0x0) at ../src/lsm/lsm_tree.c:1049
#5  0x00007fc6b5de3ba6 in __wt_lsm_manager_destroy (session=0x0) at ../src/lsm/lsm_manager.c:332
#6  0x00007fc6b5e31617 in __session_create (wt_session=0x0, uri=0x0, config=0x7fc6b4ceee90 "") at ../src/session/session_api.c:578
#7  0x00007fc6b5e31d3f in __session_log_flush_readonly (wt_session=0x7fc6b5e67100, config=0x0) at ../src/session/session_api.c:657
```

agorrod:
Thanks @michaelcahill - I'm happy with this change now - I'll go ahead and merge.

PR #3237, opened 2017-01-09
WT-3105 Create all eviction sessions initially to avoid deadlock.

sueloverso:
@agorrod Here's the change to allocate and create all the eviction thread sessions at the start to avoid the deadlock,


agorrod:

PR #3238, opened 2017-01-09
WT-3114 Avoid archive immediately after recovery.

sueloverso:
@agorrod Please review this change.  It was smaller than I originally thought it would be even.  Essentially one line and a few archiving related tests.  This will avoid archiving log files after recovery until after the first "ongoing" checkpoint is written.


agorrod:
@sueloverso I added a comment into the JIRA ticket.

agorrod:
Thanks @sueloverso lgtm - I'll merge. Is the user-visible behavior change significant enough to warrant a not in the upgrading docs?

PR #3240, opened 2017-01-10
WT-3039 Make new log record more general for potential future expansion.

sueloverso:
@michaelcahill and @agorrod Here's a slightly different take on the new prevlsn log record.  I've renamed it to `system` and included a currently unused WT_ITEM in the record.  The thought is that we may want to include other information in this record someday and then all the plumbing is there (if the "other information" maps only to a WT_ITEM).  This is a branch off the other branch, wt-3039-prevlsn, so can be merged if you like it.

### DIFF
--- a/dist/log_data.py
+++ b/dist/log_data.py
@@ -42,7 +42,8 @@ rectypes = [
     LogRecordType('message', 'message', [('string', 'message')]),
 
     # System record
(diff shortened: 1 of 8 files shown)


sueloverso:
autobuild: retest this please

agorrod:
@sueloverso I'm a big fan of this change.

sueloverso:
I am merging into the other branch.

PR #3241, opened 2017-01-12
WT-3121 In the test suite, create a standard way to load extensions

ddanderson:
Many examples of overriding setUpConnectionOpen() can now be handled by a combination of conn_config (as variable or method) and conn_extensions (as variable or method).


ddanderson:
@daveh86, can you review this please?

agorrod:
Thanks @ddanderson this looks like a good simplification to me. @daveh86 Do you think this is ready to merge?

daveh86:
@agorrod , I've read through this twice now, and run the suite with changes. LGTM

PR #3242, opened 2017-01-12
WT-3105 Avoid thread group deadlock on close.

sueloverso:
@agorrod Please review this.  Unfortunately I've been running both the format CONFIG and the evict btree wtperf workload on develop without reproducing the deadlock.  So while I think this fixes the deadlock, I have not been able to confirm either way.  The branch runs fine too for my testing.


agorrod:
@sueloverso I can see how this fixes the deadlock, I'm inclined to sit on this until we've talked through WT-3123.

agorrod:
Thanks @sueloverso as discussed, I think we should consider broader changes to the thread group semantics, but this should fix the bug so I'll get it merged.

PR #3243, opened 2017-01-12
WT-3126 bug: dist/s_all script has misplaced quote causing bad error reporting

keithbostic:
bug: dist/s_all script has misplaced quote causing bad error reporting


agorrod:
Thanks @keithbostic lgtm.

PR #3245, opened 2017-01-12
WT-3128 wt printlog returns operation-not-supported if it doesn't find any log files

keithbostic:
Add a verbose message if we don't find any log files, it's likely the user pointed us at the wrong directory.

@sueloverso, would you please review this change?


daveh86:
@keithbostic is there a way that we could test this?

keithbostic:
@daveh86
> is there a way that we could test this within the python suite?

Create a WiredTiger database w/o any logging configured, then run the wt utility printlog command. It won't find log files and you should see the error.

PR #3246, opened 2017-01-13
WT-3116 Change s_all to avoid missing error returns on older platforms

daveh86:


daveh86:
@agorrod , can you take a look at this?

agorrod:
@keithbostic Would you mind taking this one please? I noticed that ``s_all -F -A`` works differently for me than @daveh86 and this is an attempt to fix it.

@daveh86 Could you explain why the change works?

daveh86:
I found that one of our test boxes had an older version of sh. Some diving into the run showed that when we hit an error we could "break" from the loop that checks all the results files early.

I removed this problematic break call and restructured the code to avoid the use of continue.

keithbostic:
nice cleanup, lgtm, merging.

PR #3248, opened 2017-01-17
WT-3118 Protect test against unexpectedly slow child start.

sueloverso:
@agorrod Please review this change to the random-abort test program to wait for the table to exist before starting the timeout waiting.  It runs fine on my systems.  I manually added in a sleep at the beginning of the child process to simulate the delay and reproduce the original problem and test this change.


agorrod:

PR #3249, opened 2017-01-18
WT-3131 Enable ROTN on Windows

daveh86:


daveh86:
@keithbostic, can you take a look at this and confirm if its suitable?

daveh86:
retest this please

daveh86:
@ddanderson , can you take a look at this and confirm if its suitable/needed?

ext/encryptors/rotn/rotn_encrypt.c
```diff
@@ -64,6 +64,9 @@
  * In this case, we transform all bytes in the input.
  */
 
+/* Add u_char type for windows compatibility */
+typedef unsigned char   u_char;
+
```
ddanderson: @daveh86, probably it's better to replace all uses of u_char by uint8_t, which should already be defined (it's in stdint.h).

PR #3252, opened 2017-01-18
WT-3105 Fix the thread group usage on eviction reconfigure and add test.

sueloverso:
@agorrod Please review this omission from the original change.  I discovered that reconfigure was not fixed and that we had no test to eviction thread changes.  I did run the new test the first time with `verbose=(thread_group)` to confirm the right stuff is happening.


sueloverso:
autobuild: retest this please

agorrod:
Thanks @sueloverso lgtm.

PR #3253, opened 2017-01-18
WT-3097 Race on reconfigure or shutdown can lead to waiting for statistics log server

keithbostic:


src/async/async_api.c
```diff
@@ -540,8 +539,13 @@ __wt_async_flush(WT_SESSION_IMPL *session)
 	WT_ASSERT(session, async->flush_op.state == WT_ASYNCOP_FREE);
 	async->flush_op.state = WT_ASYNCOP_READY;
 	WT_RET(__wt_async_op_enqueue(session, &async->flush_op));
-	while (async->flush_state != WT_ASYNC_FLUSH_COMPLETE)
-		__wt_cond_wait(session, async->flush_cond, 100000);
+	while (async->flush_state != WT_ASYNC_FLUSH_COMPLETE) {
+		/*
+		 * No quit function needed, we're only pausing for 1/10th of a
+		 * second.
+		 */
```
michaelcahill: Let's be consistent and call this a "run function" everywhere.

keithbostic: > Let's be consistent and call it a "run function" everywhere.

Done.

> Also, consider adding a function __wt_cond_wait_nocheck or similar that asserts the timeout is less than ~1s to avoid repeating this comment everywhere.

I added an assert to ensure we don't sleep for more than 5 seconds w/o a run check function.

I'm hesitant to add a check function. There are 4 flavors of this call (`__wt_cond_wait_signal()` and its wrappers `__wt_cond_wait()`, `__wt_cond_auto_wait_signal()` and `__wt_cond_auto_wait()`), we'd need 4 more names to do that. Let me know if you want that change.

keithbostic:
@sueloverso, can you take on this review?

Notes:
* I’ve broken out the commits by change, so going by commit is likely to be clearer than looking at the diff as a whole, at least to start.
* I got around the problems with lots of calls not having "stop running test flags" by ignoring any calls that were waiting for short periods of time (turns out, the calls without test flags aren't waiting very long, so they're not a race problem, they're going to wake up shortly, regardless). There's probably a chunk of remaining work to avoid hanging if `WT_CONNECTION.close` is called unexpectedly, but that's outside the scope of this work. (I have no objection to including that work in this branch if you want, though.)
* I made some minor changes in server threads to stop immediately when quitting or being reconfigured, instead of doing one last chunk of work and then stopping.
* I changed `WT_CONNECTION.close` to set the `WT_CONN_CLOSING` flag before calling the async, lsm-manager and sweep destroy functions. I couldn't see any reason they'd be special, and it allows those functions to check that flag.
* I reworked `WT_SESSION.transaction_sync` as you and I discussed, to wait for the full period, rather than waking periodically.

keithbostic:
autobuild: retest this please

src/conn/conn_ckpt.c
```diff
@@ -78,14 +94,20 @@ __ckpt_server(void *arg)
...
+		if (!__ckpt_server_run_chk(session))
+			break;
 		/*
 		 * Wait...
 		 * NOTE: If the user only configured logsize, then usecs
 		 * will be 0 and this wait won't return until signalled.
 		 */
-		__wt_cond_wait(session, conn->ckpt_cond, conn->ckpt_usecs);
+		__wt_cond_wait(session,
+		    conn->ckpt_cond, conn->ckpt_usecs, __ckpt_server_run_chk);
```
sueloverso: If anything, this use of passing in a run function does deserve a comment since it is the rare case.

src/conn/conn_stat.c
```diff
@@ -525,10 +541,17 @@ __statlog_server(void *arg)
 	WT_ERR(__wt_buf_init(session, &path, strlen(conn->stat_path) + 128));
 	WT_ERR(__wt_buf_init(session, &tmp, strlen(conn->stat_path) + 128));
 
-	while (F_ISSET(conn, WT_CONN_SERVER_RUN) &&
-	    F_ISSET(conn, WT_CONN_SERVER_STATISTICS)) {
+	for (;;) {
+		if (!__statlog_server_run_chk(session))
+			break;
```
sueloverso: Same here

src/conn/conn_sweep.c
```diff
@@ -266,13 +282,20 @@ __sweep_server(void *arg)
...
 	 * Sweep for dead and excess handles.
 	 */
-	while (F_ISSET(conn, WT_CONN_SERVER_RUN) &&
-	    F_ISSET(conn, WT_CONN_SERVER_SWEEP)) {
-		/* Wait until the next event. */
-		__wt_cond_wait(session,
-		    conn->sweep_cond, conn->sweep_interval * WT_MILLION);
+	for (;;) {
+		if (!__sweep_server_run_chk(session))
+			break;
```
sueloverso: And here.

src/include/extern.h
```diff
@@ -613,11 +613,9 @@ extern void __wt_session_close_cache(WT_SESSION_IMPL *session) WT_GCC_FUNC_DECL_
...
-extern void __wt_cond_auto_wait_signal( WT_SESSION_IMPL *session, WT_CONDVAR *cond, bool progress, bool *signalled) WT_GCC_FUNC_DECL_ATTRIBUTE((visibility("hidden")));
-extern void __wt_cond_auto_wait( WT_SESSION_IMPL *session, WT_CONDVAR *cond, bool progress) WT_GCC_FUNC_DECL_ATTRIBUTE((visibility("hidden")));
-extern int __wt_cond_auto_destroy(WT_SESSION_IMPL *session, WT_CONDVAR **condp) WT_GCC_FUNC_DECL_ATTRIBUTE((warn_unused_result)) WT_GCC_FUNC_DECL_ATTRIBUTE((visibility("hidden")));
+extern int __wt_cond_auto_alloc(WT_SESSION_IMPL *session, const char *name, uint64_t min, uint64_t max, WT_CONDVAR **condp) WT_GCC_FUNC_DECL_ATTRIBUTE((warn_unused_result)) WT_GCC_FUNC_DECL_ATTRIBUTE((visibility("hidden")));
+extern void __wt_cond_auto_wait_signal(WT_SESSION_IMPL *session, WT_CONDVAR *cond, bool progress, bool (*run_func)(WT_SESSION_IMPL *), bool *signalled) WT_GCC_FUNC_DECL_ATTRIBUTE((visibility("hidden")));
+extern void __wt_cond_auto_wait(WT_SESSION_IMPL *session, WT_CONDVAR *cond, bool progress, bool (*run_func)(WT_SESSION_IMPL *)) WT_GCC_FUNC_DECL_ATTRIBUTE((visibility("hidden")));
```
sueloverso: I realize this is a matter of personal taste.  I prefer having balanced/matching functions and names even if they're just wrappers.  @agorrod and I discussed this in [WT-2318](https://jira.mongodb.org/browse/WT-2318) and [PR 2550](https://github.com/wiredtiger/wiredtiger/pull/2550).

src/conn/conn_ckpt.c
```diff
@@ -78,14 +94,20 @@ __ckpt_server(void *arg)
 	conn = S2C(session);
 	wt_session = (WT_SESSION *)session;
 
-	while (F_ISSET(conn, WT_CONN_SERVER_RUN) &&
-	    F_ISSET(conn, WT_CONN_SERVER_CHECKPOINT)) {
+	for (;;) {
+		if (!__ckpt_server_run_chk(session))
+			break;
```
keithbostic: It could be, but the loop begins like this:
```
for (;;) {
        if (!__ckpt_server_run_chk(session))
                break;
        /*
         * Wait...
         * NOTE: If the user only configured logsize, then usecs
         * will be 0 and this wait won't return until signalled.
         */
        __wt_cond_wait(session,
            conn->ckpt_cond, conn->ckpt_usecs, __ckpt_server_run_chk);

        /* Check if we're quitting or being reconfigured. */
        if (!__ckpt_server_run_chk(session))
                break;
```
That is, `__ckpt_server_run_chk()` appears 3 times in the first few lines of the loop, and I hoped repeating the construct would make things clearer, than calling the function different ways.

src/async/async_api.c
```diff
@@ -540,8 +539,13 @@ __wt_async_flush(WT_SESSION_IMPL *session)
 	WT_ASSERT(session, async->flush_op.state == WT_ASYNCOP_FREE);
 	async->flush_op.state = WT_ASYNCOP_READY;
 	WT_RET(__wt_async_op_enqueue(session, &async->flush_op));
-	while (async->flush_state != WT_ASYNC_FLUSH_COMPLETE)
-		__wt_cond_wait(session, async->flush_cond, 100000);
+	while (async->flush_state != WT_ASYNC_FLUSH_COMPLETE) {
+		/*
+		 * No run function needed, we're only pausing for 1/10th of a
+		 * second.
+		 */
```
keithbostic: OK, I've removed all of the comments. My fear was it was too easy to change for how long a wait was configured without realizing the wait would then need a run-time check function where it didn't before. However, there's now an assert in the condition variable code that should catch that happening, so we're probably safe.

src/conn/conn_sweep.c
```diff
@@ -266,13 +282,20 @@ __sweep_server(void *arg)
...
+
 		__wt_seconds(session, &now);
```
sueloverso: I just realized you moved the call to `wt_seconds` to before the `cond_wait`.  It originally was after the wait.  Was that change intentional?

src/conn/conn_ckpt.c
sueloverso: I see that everywhere now that I pull the branch and look at the code instead of the diff.  All of the 4-5 functions are identically structured. What about encapsulating in a function:
`bool __wt_server_wait_and_run(same 4 args as cond_wait)` and that then makes it caller look like this (statlog for example):
```
while (__wt_server_wait_and_run(session, conn->stat_cond, conn->stat_usecs, __statlog_server_run_chk)
    if (WT_STAT_ENABLED(session))
        WT_ERR(__statlog_log_one(session, ...);
```
```

File: src/include/extern.h (hunk 5 above)
Comment (keithbostic) [inline src/include/extern.h]: I disagree, I don't see any point in an additional function call, it's the same structure just initialized slightly differently. That said, I won't die on this hill, if you & @agorrod want the wrapper functions back, let me know, I'm happy to make that change.

Comment (sueloverso) [inline src/include/extern.h]: In IM you suggested an assert to ensure the correct `wait` functions are called.  If we go with this naming then please add that.

File: src/conn/conn_ckpt.c (hunk 6 above)
Comment (keithbostic) [inline src/conn/conn_ckpt.c]: Ummm, I'd rather add the  `while` back in that push it down another level.

File: src/conn/conn_ckpt.c (hunk 2 above)
Comment (keithbostic) [inline src/conn/conn_ckpt.c]: Fixed.

File: src/conn/conn_sweep.c (hunk 4 above)
Comment (keithbostic) [inline src/conn/conn_sweep.c]: Resolved.

File: src/conn/conn_sweep.c (hunk 8 above)
Comment (keithbostic) [inline src/conn/conn_sweep.c]: No, it was not -- I deliberately tried (and failed!) to leave it alone.  Thank you, nice catch.

File: src/include/extern.h (hunk 5 above)
Comment (keithbostic) [inline src/include/extern.h]: > In IM you suggested an assert to ensure the correct wait functions are called. If we go with this naming then please add that.

Done.

Comment (keithbostic) [inline src/include/extern.h]: So 435478e failed pretty quickly: in the eviction code, the condition variable `cache->evict_cond` is used both as an automatically adjusted condition variable, and as a normal condition variable (to wait 1/1000th of a second in `evict_pass()`).

@agorrod, @michaelcahill, is this a bug in eviction, or is that a deliberate use, and it's OK to use automatically adjusting condition variables in other ways?

Comment (keithbostic) [inline src/include/extern.h]: @michaelcahill confirms: this is an expected use case, I'm going to revert 435478e.

**Comment by keithbostic** [comment]
autobuild: retest this please

File: src/include/extern.h (hunk 5 above)
Comment (michaelcahill) [inline src/include/extern.h]: It's "expected" in the sense that we have different flavors of eviction threads sharing the same condition variable -- we don't want the eviction server sleeping a long time while holding locks.

We could probably restructure eviction (i.e., add a special-purpose condition variable or a short `__wt_sleep`) for this case.  It has gone through multiple iterations already.

File: src/conn/conn_ckpt.c (hunk 6 above)
Comment (michaelcahill) [inline src/conn/conn_ckpt.c]: Why do we need to check before calling `__wt_cond_wait`?  It's going to check before going to sleep...

I don't mind the `while(__wt_server_wait_loop(...))` construct (though I don't like `_and_` function names), since we're repeating the same logic in lots of places.  What's your objection @keithbostic?

File: src/os_win/os_mtx_cond.c [hunk 9]
```diff
@@ -59,6 +58,23 @@ __wt_cond_wait_signal(
...
+	 * only a single wakeup, our waker is likely waiting for us to exit.
+	 * After acquiring the mutex (so we're guaranteed to be awakened by any
+	 * future wakeup call), optionally check if we're OK to keep running.
+	 * This won't ensure our caller won't just loop and call us again, but
+	 * at least it's not our fault.
+	 *
+	 * Assert we're not waiting longer than a few seconds if not checking
+	 * the run status.
+	 */
+	WT_ASSERT(session, run_func != NULL || usecs < 5 * WT_MILLION);
```
michaelcahill: If we make this 1 second, will anything break?  That's easier to remember / conform to than "a few".

src/session/session_api.c
```diff
@@ -1555,22 +1570,20 @@ __session_transaction_sync(WT_SESSION *wt_session, const char *config)
...
-			 * Note, we will wait an increasing amount of time
-			 * each iteration, likely doubling.  Also note that
-			 * the function timeout value is in usecs (we are
-			 * computing the wait time in msecs and passing that
-			 * in, unchanged, as the usecs to wait).
-			 */
-			__wt_cond_wait(session, log->log_sync_cond, waited_ms);
-		else
+		if (forever || waited_ms < timeout_ms) {
+			remaining_usec = (timeout_ms - waited_ms) * 1000;
```
sueloverso: Should be WT_THOUSAND.

keithbostic:
@sueloverso, courtesy of Alex & Michael:

The problem is that Jenkins runs the test suite with a special environment variable, and that leads to the failure. You can reproduce the failure with:
```
env \
    'WIREDTIGER_CONFIG=checkpoint_sync=0' \
    LD_LIBRARY_PATH=.libs \
    python test/suite/run.py test_txn14.test_txn14.test_log_flush
```
It seems to me it's likely an error to call `WT_SESSION.transaction_sync` if `checkpoint_sync` isn't configured, does this change make sense to you?
```
diff --git a/src/session/session_api.c b/src/session/session_api.c
index 020216037..23867b333 100644
@@ -1535,6 +1535,12 @@ __session_transaction_sync(WT_SESSION *wt_session, const char *config)
        if (!FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))
                WT_ERR_MSG(session, EINVAL, "logging not enabled");
 
+       /*
+        * If checkpoint-sync not enabled, there is nothing to do.
+        */
+       if (!F_ISSET(conn, WT_CONN_CKPT_SYNC))
+               WT_ERR_MSG(session, EINVAL, "checkpoint-sync not enabled");
+
        log = conn->log;
        timeout_ms = waited_ms = 0;
        forever = true;
diff --git a/test/suite/test_txn14.py b/test/suite/test_txn14.py
index f9ccabaab..e9a41156a 100644
--- a/test/suite/test_txn14.py
+++ b/test/suite/test_txn14.py
@@ -92,11 +92,11 @@ class test_txn14(wttest.WiredTigerTestCase, suite_subprocess):
             c[i+self.entries] = i + self.entries + 1
         c.close()
         self.session.log_flush(cfgarg)
+        # If doing a background flush, wait a few seconds. I have seen an
+        # individual log file's fsync take more than a second on some systems,
+        # give it time to flush perhaps a few files.
         if self.sync == 'background':
-            # If doing a background flush, wait a few seconds.  I have
-            # seen an individual log file's fsync take more than a second
-            # on some systems.  So give it time to flush perhaps a few files.
-            self.session.transaction_sync('timeout_ms=4000')
+            time.sleep(5)
         self.simulate_crash_restart(".", "RESTART")
         c = self.session.open_cursor(self.t1, None, None)
         i = 0
```

michaelcahill:
@keithbostic, the environment variable is only intended to avoid the low-level call to `fsync` so the disks on our test servers last longer and tests complete faster.

There must be some other issue here that is causing this test to fail on the branch?

sueloverso:
I agree with @michaelcahill.  The test and background `transaction_sync` deals with moving LSNs.  Yes, it does so by syncing the log files in the background which is independent of the purpose of checkpoint_sync.  None of that underlying code changed here so something else must have.

keithbostic:
@michaelcahill, @sueloverso: thanks for pushing back. I'm an idiot, the code was testing the wrong flag.

src/conn/conn_ckpt.c
keithbostic: > Why do we need to check before calling __wt_cond_wait? It's going to check before going to sleep...

I removed the call of the `XXX_run_chk()` functions before the call to `__wt_cond_wait()`.

src/os_win/os_mtx_cond.c
keithbostic: I don't believe anything will break because all of our current calls to `__wt_cond_auto_alloc` set a max of 1 second.

That said, I don't see a need to tighten the test down. It's not wrong to create a condition variable with a max-wait of more than 1 second, not specifying a race-check function is only a problem if the wait is long enough to affect shutdown.

If you still think it's better at 1 second, feel free to tighten it down.

src/session/session_api.c
keithbostic: Fixed.

keithbostic:
@michaelcahill, as far as I'm concerned, this one is ready to merge.

src/evict/evict_lru.c
```diff
@@ -698,7 +699,7 @@ __evict_pass(WT_SESSION_IMPL *session)
 				WT_STAT_CONN_INCR(session,
 				    cache_eviction_server_slept);
 				__wt_cond_wait(
-				    session, cache->evict_cond, WT_THOUSAND);
+				    session, cache->evict_cond, 1000, NULL);
```
sueloverso: This 1000 should go back to WT_THOUSAND as well.

keithbostic: Sure, done.

michaelcahill:
autobuild: retest this please

michaelcahill:

PR #3254, opened 2017-01-19
WT-3135 WT-3159 Fix search_near() with custom collators for index keys of variable length.

ddanderson:
As part of the fix for search_near, this includes a fix for how format 'U' is interpreted.  'U' should always include the size, before this change, it did not if the 'U' appeared at the end of the format. As a side effect of this change, WT-3159 is also fixed.

### DIFF
--- a/src/cursor/cur_index.c
+++ b/src/cursor/cur_index.c
@@ -246,7 +246,7 @@ __curindex_search(WT_CURSOR *cursor)
 	 * all the visible fields so it unpacks correctly.
 	 */
 	if (cindex->index->collator != NULL)
-		WT_ERR(__wt_struct_repack(session, cindex->child->key_format,
+		WT_ERR(__wt_struct_repack(session, child->key_format,
 		    cindex->iface.key_format, &child->key, &found_key));
 	else
 		found_key.size = cursor->key.size;
--- a/src/cursor/cur_index.c
+++ b/src/cursor/cur_index.c
@@ -307,8 +307,19 @@ __curindex_search_near(WT_CURSOR *cursor, int *exact)
 	 * so we flip the sign of the result to match what callers expect.
 	 */
 	found_key = child->key;
-	if (found_key.size > cursor->key.size)
-		found_key.size = cursor->key.size;
+	if (found_key.size > cursor->key.size) {
+		/*
+		 * Custom collators will use unpack functions to
+		 * get the visible key fields. We need to provide
+		 * those fields in their entirety.
+		 */
+		if (cindex->index->collator != NULL)
+			WT_ERR(__wt_struct_repack(session,
+			    cindex->child->key_format,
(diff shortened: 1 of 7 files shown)


daveh86:
retest this please

ddanderson:
This is not yet ready for review, I'm adding more tests.

daveh86:
retest this please

daveh86:
@ddanderson , I added ASAN checking into the build testers today. And this PR was first on the list in requests, so I tagged it to be tested.

ddanderson:
retest this please

ddanderson:
@michaelcahill, can you review please?  There will be a small hit when using search/search_near with custom collators - one of the items passed to the collator needs to be rescanned using the format.  At least there is no extra copying.

ddanderson:
Also please take note, there's a fix in `__pack_next` for the `'U' `case.

src/cursor/cur_index.c
```diff
@@ -240,7 +240,17 @@ __curindex_search(WT_CURSOR *cursor)
...
-	found_key.size = cursor->key.size;
+
+	/*
+	 * Custom collators will use unpack functions to
+	 * get the visible key fields. We need to provide
+	 * those fields in their entirety.
+	 */
+	if (cindex->index->collator != NULL)
+		WT_ERR(__wt_struct_repack(session, cindex->child->key_format,
+		    cindex->iface.key_format, &child->key, &found_key));
```
michaelcahill: Minor point: we've got `cindex->child` in local variable `child` (I think) -- can we use that consistently?

Can you please expand the comment a little and reformat?  Something like "Custom collators expect to see complete keys, pass an item containing all of the visible fields so it unpacks correctly".

src/include/packing.i
```diff
@@ -168,10 +168,13 @@ next:	if (pack->cur == pack->end)
...
 		return (0);
 	case 'u':
-	case 'U':
 		/* Special case for items with a size prefix. */
 		pv->type = (!pv->havesize && *pack->cur != '\0') ? 'U' : 'u';
 		return (0);
+	case 'U':
+		/* Special case for items with a size prefix. */
+		pv->type = !pv->havesize ? 'U' : 'u';
+		return (0);
```
michaelcahill: Can you add more detail to the comments here about why these cases are different?

ddanderson:
Hello @michaelcahill, can you take another look, please?  I simplified the `'U'` case, and added more to the comments.

ddanderson:
@michaelcahill, can you please review this?  There's no compatibility problem that I'm aware of with this latest fix.  I've tested an "upgrade" (creating database in develop, then reading it with this branch) using various schemas, as well as crashing the "develop" database and recovering with this branch.

michaelcahill:
Thanks @ddanderson, this lgtm.  I'll merge.

PR #3256, opened 2017-01-20
WT-3130 Proposal to change initialization of custom filesystem

keithbostic:

I changed it so `WT_CONNECTION.get_home()` will work before we load the extensions. I also configure error handling a little earlier, I thought that made sense to do before loading extensions.


agorrod:
Thanks @keithbostic lgtm.

PR #3259, opened 2017-01-22
WT-3144 bug fix: random cursor returns not-found when descending to an empty page

keithbostic:
@michaelcahill, these changes make things better (and pass the existing test), but they don't entirely solve the problem as we won't necessarily recover from descending into an internal page that contains nothing other than deleted leaf pages. That condition should have triggered a reverse split to clean up the parent, but it's obviously not guaranteed we'll be able to reverse split the parent.

I'm not sure how far we want to push this: for example, if the entire tree is filled with empty pages and there's a single valid key/value pair, are we committed to returning that item from a sample call?

I looked at the page-in `WT_READ_NO_EMPTY` flag, but didn't see any obvious way to have it help solve the problem.

### DIFF
--- a/src/btree/row_srch.c
+++ b/src/btree/row_srch.c
@@ -831,7 +831,7 @@ restart:	/*
 		}
 		if (i == entries)
 			for (i = 0; i < entries; ++i) {
-				descent = pindex->index[entries];
+				descent = pindex->index[i];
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, I agree that this seems better so I'll merge.  I'll follow up in the ticket about next steps.

PR #3261, opened 2017-01-23
WT-3120 Fix ordering problem in connection_close for filesystem loaded in an extension

keithbostic:
Lint/Coverity. Some of the code was copied from example code, which is why there are changes there as well.


ddanderson:
LGTM, merging.

PR #3262, opened 2017-01-23
WT-3120 Fix ordering problem in connection_close for filesystem loaded in an extension

keithbostic:
WT_CURSOR.{get_key,get_value} can fail.


agorrod:
Thanks @keithbostic lgtm.

PR #3263, opened 2017-01-23
WT-2790 Look at removed statistics instead of files closed.

sueloverso:
Ready for review once testing on evergreen completes.  After looking at some debugging statistics during a failure, I realized there is a lag between when sweep closes files and when it can remove data handles from the connection list.  On slow machines, we can hit the test failure if we catch it between those two.  So I modified the test where it looks at statistics to look for any change in the removed statistic instead of closed files.

### DIFF
--- a/test/suite/test_sweep01.py
+++ b/test/suite/test_sweep01.py
@@ -106,7 +106,6 @@ class test_sweep01(wttest.WiredTigerTestCase, suite_subprocess):
         k = 0
         sleep = 0
         max = 60
-        final_nfile = 4
         while sleep < max:
             self.session.checkpoint()
             k = k+1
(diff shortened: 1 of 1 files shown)


sueloverso:
autobuild: retest this please

sueloverso:
@agorrod This is ready for review. It has passed on evergreen as well (although admittedly the machine there did not fail every time anyway).

agorrod:
> a lag between when sweep closes files and when it can remove data handles from the connection list

I'm not seeing why this works. I would expect a successful call to ``__sweep_remove_one`` to result in both the ``dh_sweep_remove`` and ``stat.conn.file_open`` statistics being bumped by one. Could you help me understand what I'm missing?

sueloverso:
The test originally looked for the number of open files to be the expected number (4) from a large number, 30-ish.  As sweep closes files, in `sweep_discard_trees` it is decrementing the `conn->open_file_count` from the call to `conn_btree_sync_and_close`.  Only after all files have been closed do we then walk the entire list of handles again and remove them if possible.  There is a window where the open file count has been decremented and the sweep thread has not yet called `sweep_remove_handles`.  The test thread could see the statistics between those two steps and result in this failure.  The `sweep_remove_handles` is the last step in the sweep.  So I adjusted the test to look for any movement in that statistic as well as the original expected number of files.

agorrod:
Oh, I see. Thanks.

PR #3264, opened 2017-01-24
WT-3148 Improve efficiency of eviction with many small trees.

michaelcahill:


src/evict/evict_lru.c
```diff
@@ -1587,6 +1589,13 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 	    !F_ISSET(cache, WT_CACHE_EVICT_CLEAN))
 		min_pages *= 10;
 
+	walk_flags =
+	    WT_READ_CACHE | WT_READ_NO_EVICT | WT_READ_NO_GEN | WT_READ_NO_WAIT;
+
+	/* Randomize the walk direction. */
+	if (btree->evict_walk_reverse)
+		FLD_SET(walk_flags, WT_READ_PREV);
+
```
michaelcahill: Note: this reordering should not have any effect, it just seemed cleaner to me to deal with page counts, then walk flags rather than interleaving the two.

agorrod:
I've switched the Jenkins perf jobs over to this branch.

agorrod:
@michaelcahill I plan to run the Jenkins performance tests on this branch next week before merging.

sueloverso:
I changed the Jenkins PERF_BRANCH to this one so that it can run all weekend and you can look at it on your Monday morning.

sueloverso:
FYI I had to move the perf branch back to develop.  All of the perf tests were failing on Jenkins when it was set to this branch.  I have not looked at why.

sueloverso:
Ok. I took a quick look, I had a typo in the branch name...  I reset it to this branch and will let it go now.

src/evict/evict_lru.c
```diff
@@ -1539,44 +1541,44 @@ __evict_walk_file(WT_SESSION_IMPL *session,
...
-		if (__wt_random(&session->rnd) > UINT32_MAX / 100)
-			return (0);
+	/*
+	 * Randomly walk trees with a small fraction of the cache in case there
+	 * are so many trees that none of them use enough of the cache to be
+	 * allocated slots.  Make the chance of walking a tree equal to the
+	 * chance that a random byte in cache belongs to the tree.
+	 */
+	if (target_pages == 0 && (cache_inuse == 0 ||
+	    __wt_random64(&session->rnd) % cache_inuse > btree_inuse))
```
agorrod: If both dirty and clean eviction are enabled ``cache_inuse`` and ``btree_inuse`` reflect the dirty proportion of data.

I'm struggling a bit to see how this calculation actually works. I feel like the number of trees should play into the left side of the calculation as well. If there are 1000 trees in a 100GB cache. 800 type A trees use 1MB (round to 0), 100 type B trees use 100MB (10GB), 50 type C trees use 500MB (25GB) and 50 type D trees use 1.3GB each.

| Tree type | Frequency of selection (one selection per X passes) | Number of selections per traverse of handles |
|---|---|---|
| Type A | 100,000 | 0.008 |
| Type B | 1000 | 0.01 |
| Type C | 200 | 0.25 |
| Type D | 76 | 0.64 |

Frequency of selection = Cache in use / (# trees of type * tree size)
Number of selections = Frequency of selection / count of similar handles

michaelcahill: @agorrod, I'll push a fix for the dirty / clean calculation shortly.

Here is a slightly different way to slice that scenario:

| Tree type | Number of trees | Chance of selection | Pages per pass
|---|---|---|---|
| Type A | 800 | 0.001% | 0.08 |
| Type B | 100 | 0.1% | 1 |
| Type C | 50 | 0.5% | 2.5 |
| Type D | 50 | 1.3% | 6.5 |

Pages per pass = Number of trees * Chance of selection * 10 (pages per tree is constant here)

I think you are saying that in this case where all trees are "small", we should expect to find more than 10 pages after walking through all the trees in the cache.  Is that right?

The actual proportions there look right to me -- i.e., we are putting more effort into the bigger trees.  If we were to bias the probability based on the total number of *slots* (rather than trees), would that address this concern?

agorrod: @michaelcahill Following up from our phone call, I think we are on the same page. If we were filling the queue about 10x from a single pass of the handle list I'd be happy.

agorrod:
Thanks @michaelcahill the latest set of changes look good to me. I've switched the Jenkins perf jobs over to this branch, and will merge if all looks good there.

PR #3265, opened 2017-01-24
WT-3150 Reduce impact of checkpoints on eviction.

michaelcahill:
In particular, don't have the eviction server give up all walks each
time it is interrupted, and only wait for requesting threads to make
progress: don't go to sleep.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -280,7 +280,7 @@ __wt_evict_thread_run(WT_SESSION_IMPL *session, WT_THREAD *thread)
 	WT_CACHE *cache;
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
-	bool did_work;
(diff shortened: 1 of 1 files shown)


agorrod:
@michaelcahill I plan to run the Jenkins performance tests on this branch next week before merging.

agorrod:
@michaelcahill this lgtm pending the results of the automated performance tests that are running now.

agorrod:
The perf testing looks good, except that the test_checkpoint_stress job went from completing two checkpoints per run to only completing a single checkpoint. I think this change deserves some local testing to ensure it doesn't interfere with checkpoint performance too much.

agorrod:
The regression didn't reproduce for me locally, I've kicked off another run on Jenkins.

agorrod:
Thanks @michaelcahill the perf change is ephemeral, and this is definitely an improvement. I'm going to merge.

PR #3266, opened 2017-01-24
WT-3137 Fix a hang in logging due to a race condition

keithbostic:
Lint:
Don't print int32_t's with %d.
WT_LOGSLOT.slot_error is an int, not an int32_t.
Don't print off_t's as 32-bits, use the maximum size unsigned object.


keithbostic:
@sueloverso, it occurs to me, the diagnostic messages should be sent to `__wt_msg`, not `__wt_errx`?

sueloverso:
@keithbostic LGTM, thanks.  The diagnostic messages are be written immediately before aborting because there is likely a problem.

PR #3267, opened 2017-01-24
WT-3123 Modify thread groups to create and pause threads

sueloverso:
This is not yet ready for review.  Still testing.


daveh86:
retest this please

daveh86:
retest this please

daveh86:
retest this please

sueloverso:
@agorrod This is ready for review.  I confirmed that the statistics are equivalent for eviction workers starting and stopping compared to develop.  Between `evict-btree-stress.wtperf, test_reconfig01` and `wt2719_reconfig` I think we have adequate testing of this.

sueloverso:
I changed the Jenkins perf branch to run with this several times.  Interestingly enough, the medium-btree shows a slight but consistent degradation.  The eviction tests showed no change.  A couple tests showed some improvement (update-checkpoint-lsm).

src/evict/evict_lru.c
```diff
@@ -293,31 +303,44 @@ __wt_evict_thread_run(WT_SESSION_IMPL *session, WT_THREAD *thread)
...
-			__wt_verbose(session, WT_VERB_EVICTSERVER, "sleeping");
-
-			/* Don't rely on signals: check periodically. */
-			__wt_cond_auto_wait(
-			    session, cache->evict_cond, did_work, NULL);
-			__wt_verbose(session, WT_VERB_EVICTSERVER, "waking");
+	for (;;) {
+		if (!F_ISSET(thread, WT_THREAD_RUN))
+			break;
+		if (F_ISSET(thread, WT_THREAD_ACTIVE)) {
```
agorrod: Would it be possible to hide the ``WT_THREAD_ACTIVE`` state in the thread group? I guess it would require an additional layer of indirection in thread start functions, where there is a general

```
void
__thread_run_owner(void *thread)
{
    WT_THREAD *thread;
    for (;;) {
        if (!F_ISSET(thread, WT_THREAD_RUN))
            return;
        if (!F_ISSET(thread, WT_THREAD_ACTIVE)) {
            __wt_cond_wait(paused_thread_wait_period);
            continue;
       }
       WT_ERR(thread->worker_func(thread));     
    }
err:    /* General error reporting */
}
````

sueloverso:
autobuild: retest this please

src/evict/evict_lru.c
sueloverso: Yes, it hides one set of code looking at the thread state for another.  The eviction code has thread shutdown code that it needs to execute.  Since in this style, the eviction code function returns after one pass we now need a shutdown function to call in addition to the run function.  I'll code that up and we can see whether or not it is too gross.

michaelcahill:
@sueloverso, I reviewed this branch.

Generally I'm a fan -- the more generic we can make this code, the more useful it will be long term.

The `medium-btree` regression is curious: any ideas what is going on there?

I'm honestly not sure about keeping threads alive but asleep vs stopping them.  There is a debugging cost to starting / stopping threads, and of course there is some CPU cost to having inactive threads wake up periodically.  If this is an experiment, how are we going to decide which way to go?

sueloverso:
I investigated the regression.  I am not able to reproduce it on my AWS box with all the same build and configuration settings.  All my runs come out comparable for both develop and this branch (with the earlier plot runs a couple weeks ago).  I'll investigate again with the newer code.

agorrod:
Closing this in favor of pull request 3286

PR #3268, opened 2017-01-25
WT-3139 Enhance wtperf to support periodic table scans

agorrod:
This is not ready for review, the functionality works but isn't connected into any other reporting.


agorrod:
@sueloverso Would you mind glancing at the configuration part of this change. It adds a new thread type, that has the following syntax:

```
threads=((count=1,scan=1000000,pause=300,table_name=test00003))
```

Which results in a single thread, scanning a million items from table "test00003" once every 300 seconds.

That syntax is pretty different to other thread groups, the ``table_name`` is different - and I haven't generalized it to other operation types in the workload (I don't know if I should).

The ``pause`` is also unique - and again it's currently specific to the ``scan`` operation, but I think that's reasonable since most other operations use ``throttle`` to achieve something similar.

sueloverso:
@agorrod I've been thinking about this.  You're right this is very different.  I think some changes could be considered.

The biggest one for me is `table_name`.  I would prefer just `table=3`.  There is already a `table_name` general config setting and the user should not need to know the default value nor how tables are numbered.  With `table=3` all we need to check is that it is `<= table_count` and the code can generate the proper name.  I think `table` could be a general addition to any/all workloads.  If it is specified, then that workload set only applies to that table.  I think that is a useful generalization in the case of multiple tables.

Scans are really reads.  I'm going to think about coming up with some form of `read` workload that does the same functionality but fits better with the current set of operations.

sueloverso:
Ok, here's what I propose:
* we use `table=N` and `pause=SEC` as general workload modifiers.
* we already have `read_range` as a global configuration setting.  It is not used in any configuration in `runners`.  I believe it is more correct to have `read_range` as a workload setting anyway.  So remove it from the global setting and use it in workloads instead.

Your example would become:
`threads=((count=1,read=1,read_range=1000000,pause=300,table=3))`

sueloverso:
I have merged the sub-branch and pointed the Jenkins perf tests at this branch.  If it looks good we can merge this.

sueloverso:
@agorrod This has run through two iterations of the perf tests and there are a couple changes.  I suspect that all of the changes (some a slight degradation, some an improvement) are due to your change to `traverse_all_content`.  I also started a `perf-long` run because that change to `traverse_all_content` may not play well with the huge 500m workload.  I don't know if it will affect the performance measurement itself.  But it may add a noticeable amount of time to run the job (currently 6:55).  We'll see how long it takes to run the 4 jobs with that.

Also just to be clear, there is no change to the read reporting.  It does not count the range walk with the cursor as a read (nor did it before when `read_range` was a global setting).

agorrod:
Thanks @sueloverso I'm inclined to have the ``traverse_all_content`` as optional, and default to off. I'll add a ``pre_load_dataset`` configuration option.

PR #3269, opened 2017-01-25
WT-3138 Enhance eviction server statistics range.

agorrod:
This change is excessive - the set of stats added here should be reviewed
and pruned down, but putting it up so I can see better how it looks in
code-review.


daveh86:
retest this please

daveh86:
retest this please

agorrod:
@daveh86 I'm trying to resuscitate this branch. Would you please do a first round of review?

agorrod:
@daveh86 ping

daveh86:
Another thought, it may be worth adding an "oldest unvisited page" statistic into this change. Currently we have the oldest visited, average visited and average unvisited. I can imagine a case where you have one very old page, that eviction can never seem to visit (as it is busy or something) which would skew the average unvisited value. Having an oldest unvisited would help with diagnosing that case.

src/evict/evict_lru.c
```diff
@@ -1678,6 +1679,31 @@ __evict_walk_file(WT_SESSION_IMPL *session,
...
 		target_pages = remaining_slots;
+
+	/*
+	 * These statistics generate a histogram of the number of pages targeted
+	 * for eviction each round. The range of values here are targeted at the
+	 * likely ranges for eviction based on some initial testing. The aim
+	 * is to cover these likely ranges in as few statistics as possible to
+	 * reduce the overall overhead.
+	 */
+	if (target_pages < 16) {
```
agorrod: I think it's interesting to track how often this value is 0. It's also valuable to see finer granularity for low numbers - it gives good insight into how sampling is working when there are a lot of trees in cache.

dist/stat_data.py
```diff
@@ -511,8 +522,21 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_eviction_fail', 'data source pages selected for eviction unable to be evicted'),
     CacheStat('cache_eviction_hazard', 'hazard pointer blocked page eviction'),
     CacheStat('cache_eviction_internal', 'internal pages evicted'),
+    CacheStat('cache_eviction_pages_seen', 'pages seen by eviction walk'),
     CacheStat('cache_eviction_split_internal', 'internal pages split during eviction'),
     CacheStat('cache_eviction_split_leaf', 'leaf pages split during eviction'),
+    CacheStat('cache_eviction_target_page_ge512', 'Histogram - Per tree page target greater than 512'),
+    CacheStat('cache_eviction_target_page_lt16', 'Histogram - Per tree page target 1-16'),
```
agorrod: I believe this range includes 0.

src/evict/evict_lru.c
```diff
@@ -1760,12 +1794,26 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 			btree->evict_start_type =
 			    (btree->evict_start_type + 1) %
 			    WT_EVICT_WALK_START_NUM;
+			WT_STAT_CONN_INCR(
+			    session, cache_eviction_walks_gave_up);
+			WT_STAT_DATA_INCR(
+			    session, cache_eviction_walks_gave_up);
```
agorrod: I believe more granularity here is interesting. i.e: Whether it's because we didn't find any pages to queue, or if it's the ratio that failed - what that ratio was (somehow).

dist/stat_data.py
```diff
@@ -230,10 +230,21 @@ def __init__(self, name, desc, flags=''):
...
     CacheStat('cache_eviction_walk', 'pages walked for eviction'),
+    CacheStat('cache_eviction_walk_from_root', 'eviction walks started from root of tree'),
+    CacheStat('cache_eviction_walk_passes', 'eviction passes of a file'),
+    CacheStat('cache_eviction_walk_saved_pos', 'eviction walks started from saved location in tree'),
     CacheStat('cache_eviction_walks_abandoned', 'eviction walks abandoned'),
     CacheStat('cache_eviction_walks_active', 'files with active eviction walks', 'no_clear,no_scale'),
+    CacheStat('cache_eviction_walks_ended', 'eviction walks reached end of tree'),
+    CacheStat('cache_eviction_walks_gave_up', 'eviction walks gave up because they saw too many pages'),
     CacheStat('cache_eviction_walks_started', 'files with new eviction walks started'),
+    CacheStat('cache_eviction_walks_stopped', 'eviction walks gave up because they walked the tree twice'),
```
agorrod: The description here isn't very accurate. No walks really walk the whole tree twice - it's an idiosyncrasy of the loop structure that the loop terminates at two restarts.

src/evict/evict_stat.c
```diff
@@ -16,28 +16,41 @@ static void
...
+	visited_age_gap_sum = unvisited_age_gap_sum = 0;
+	max_pagesize = 0;
+	num_memory = num_queued = num_not_queueable = num_smaller_allocsz = 0;
+	pages_clean = pages_dirty = pages_internal = pages_leaf = 0;
+	seen_count = unvisited_count = walk_count = 0;
 	min_written_size = UINT64_MAX;
 
 	while (__wt_tree_walk_count(session, &next_walk, &walk_count,
-	    WT_READ_CACHE | WT_READ_NO_EVICT | WT_READ_NO_WAIT) == 0 &&
+	    WT_READ_CACHE | WT_READ_NO_EVICT |
```
agorrod: Yes

src/evict/evict_stat.c
```diff
@@ -100,13 +127,15 @@ __evict_stat_walk(WT_SESSION_IMPL *session)
...
 	    cache_state_smaller_alloc_size, num_smaller_allocsz);
-	WT_STAT_DATA_SET(session, cache_state_pages, walk_count);
+	WT_STAT_DATA_SET(session, cache_state_pages, seen_count);
 	WT_STAT_DATA_SET(session, cache_state_pages_clean, pages_clean);
 	WT_STAT_DATA_SET(session, cache_state_pages_dirty, pages_dirty);
 	WT_STAT_DATA_SET(session, cache_state_pages_internal, pages_internal);
 	WT_STAT_DATA_SET(session, cache_state_pages_leaf, pages_leaf);
 	WT_STAT_DATA_SET(session,
 	    cache_state_refs_skipped, walk_count - seen_count);
+	WT_STAT_DATA_SET(session,
```
agorrod: I believe `pages seen by eviction walk` already tracks that

src/evict/evict_lru.c
daveh86: As discussed, 0 doesn't make sense here but tracking times when there are less than MIN_PAGES_PER_TREE does.

daveh86:
Thanks for the clarifications @agorrod. I've made changes to the things you picked up and merged in develop. I think this is ready for another round.

PR #3270, opened 2017-01-25
WT-3139 Implement scans as read_range.

sueloverso:
@agorrod I took a pass at implementing my suggestions and I think it is a bit cleaner and requires less new code than your original implementation.  I didn't take it any further than testing the one `stress` workload I found using `read_range`.  Review and merge if you like this better (it is based off your branch).

### DIFF
--- a/bench/wtperf/stress/btree-split-stress.wtperf
+++ b/bench/wtperf/stress/btree-split-stress.wtperf
@@ -6,5 +6,4 @@ run_time=300
 reopen_connection=false
 populate_threads=2
 value_sz=256
-read_range=100
-threads=((count=4,inserts=1,throttle=100000),(count=8,reads=1))
+threads=((count=4,inserts=1,throttle=100000),(count=8,reads=1,read_range=100))
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -619,7 +619,7 @@ worker(void *arg)
 
 	while (!wtperf->stop) {
 		if (workload->pause != 0)
-			(void)sleep((uint)workload->pause);
+			(void)sleep((unsigned int)workload->pause);
 		/*
 		 * Generate the next key and setup operation specific
 		 * statistics tracking objects.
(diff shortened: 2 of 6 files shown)


agorrod:
@sueloverso I added some comments to the JIRA ticket.

sueloverso:
@agorrod This branch is ready for review and merging into your initial branch.  I fixed and confirmed the random cursor and table index bug you found.  I added a sanity check to disallow inserts and a table identifier.  We could allow them if there are no overlapping inserts in other workloads, but it seemed too complex for the moment for key space management.  We can fix it when/if we need that functionality.

bench/wtperf/stress/btree-split-stress.wtperf
```diff
@@ -6,5 +6,4 @@ run_time=300
 reopen_connection=false
 populate_threads=2
 value_sz=256
-read_range=100
-threads=((count=4,inserts=1,throttle=100000),(count=8,reads=1))
+threads=((count=4,inserts=1,throttle=100000),(count=8,reads=1,read_range=100))
```
agorrod: Did you mean to commit this change? I'm fine if it was deliberate - just checking.

bench/wtperf/wtperf.c
```diff
@@ -2957,14 +2889,35 @@ static uint64_t
...
+			lprintf(wtperf, ret, 0, "worker: rand next failed");
+			/* 0 is outside the expected range. */
+			return (0);
+		}
+		if ((ret = rnd_cursor->get_key(rnd_cursor, &key_buf)) != 0) {
+			lprintf(wtperf, ret, 0,
+			    "worker: rand next key retrieval");
+			return (0);
+		}
+		extract_key(key_buf, &rval);
```
agorrod: I think it's worth calling ``rand_cursor->reset`` here - I'd feel more comfortable both not keeping a positioned cursor and starting the random search from the beginning each time.

bench/wtperf/stress/btree-split-stress.wtperf
sueloverso: Yes.  The `read_range` configuration option is now part of specific workloads rather than at the top-level.  This change is now necessary for correct running of this file.

bench/wtperf/wtperf.c
sueloverso: Done.

PR #3274, opened 2017-02-02
WT-3157 Fix checkpoint error path

sueloverso:
@keithbostic Please review this new pull request for the checkpoint error path.  I believe it needs to go into develop instead of @ddanderson 's branch.  Otherwise it is the same as the other branch.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -181,7 +181,7 @@ __checkpoint_apply(WT_SESSION_IMPL *session, const char *cfg[],
 	int (*op)(WT_SESSION_IMPL *, const char *[]))
 {
 	WT_DECL_RET;
-	u_int i;
+	u_int i, j;
 
 	/* If we have already locked the handles, apply the operation. */
 	for (i = 0; i < session->ckpt_handle_next; ++i) {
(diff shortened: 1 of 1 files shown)


sueloverso:
Merging - Keith gave a lgtm in IM.

src/txn/txn_ckpt.c
```diff
@@ -181,18 +181,30 @@ __checkpoint_apply(WT_SESSION_IMPL *session, const char *cfg[],
...
 	}
 
-	return (0);
+err:
+	/*
+	 * If we have an error somewhere in processing the handles, then
+	 * we need to mark earlier trees dirty.
+	 */
+	if (ret != 0)
+		for (j = 0; j < i; ++j) {
```
agorrod: I'm not convinced this is always adequate. I believe it's necessary to mark any tree that any phase of checkpoint has visited as dirty, not just handles that have been visited in this phase of checkpoint.

By phase, I mean any place that calls ``__checkpoint_apply``.

keithbostic: In this kind of error handling, I'd bend over backwards to be more aggressive.

For the same reason, I'd support setting `S2C(session)->modified` as well.

PR #3275, opened 2017-02-02
WT-3157 More aggressive error handling.

sueloverso:
@agorrod Please review this more aggressive branch (since the other was merged).


agorrod:
@michaelcahill Would you mind taking a look at this change? I'm happy with it, but it's in a sensitive code path, so another set of eyes are warranted.

michaelcahill:
@sueloverso, I opened #3281 with my suggestion here (i.e., move this into the error handler for `__wt_txn_checkpoint`).  Can you please review that branch?

PR #3276, opened 2017-02-03
WT-3150 Fix: don't spin forever during eviction interrupts.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm

PR #3277, opened 2017-02-03
WT-2439 Improve page layout: keep pages more than half full

sulabhM:


sulabhM:
@keithbostic : Please take a look at the approach.
The failures with the pull request tester are in the column store related functionality, which I am yet to work on. I tested locally and all row store related tests in python test suite pass successfully.

This is a review of a draft change, to go through the approach that I am following. I have commented where work still needs to be done.

keithbostic:
@sulabhM, I'll take a look soon. Does test/format run yet?

src/reconcile/rec_write.c
```diff
@@ -344,6 +327,15 @@ static int  __rec_dictionary_init(WT_SESSION_IMPL *, WT_RECONCILE *, u_int);
...
 static void __rec_dictionary_reset(WT_RECONCILE *);
+static inline int __rec_split_crossing_bnd(
+    WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len);
+
+#define	CROSSING_ALT_BND(r, next_len)			\
+	(r->bnd[r->bnd_next].alt_offset == 0 && next_len > r->alt_space_avail)
+#define	CROSSING_SPLIT_BND(r, next_len)			\
+	(next_len > r->space_avail)
+#define	NEED_KEY_PROMOTE(r, next_len)			\
+	(CROSSING_ALT_BND(r, next_len) || CROSSING_SPLIT_BND(r, next_len))
```
keithbostic: @sulabhM, you'll want to change these macro names, they step on name spaces outside of ours.

keithbostic: You'll also need to parenthesize `r` and `next_len` to avoid operator precedence problems.

keithbostic:
@sulabhM, I like where this is going, I think what you've got is simpler & cleaner than the original code.

sulabhM:
Successfully ran test/format for 17 hours (appx 1500 runs), following config:

```
runs=10000
file_type=row
data_source=table
```

src/reconcile/rec_write.c
```diff
@@ -2350,39 +2398,132 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
 	WT_BTREE *btree;
-	size_t corrected_page_size, len;
+	size_t corrected_page_size, inuse, len;
 
 	btree = S2BT(session);
 	bm = btree->bm;
 
 	len = WT_PTRDIFF(r->first_free, r->disk_image.mem);
-	corrected_page_size = len + add_len;
+	inuse = len - r->bnd[r->bnd_next].offset +
```
agorrod: Line wrapping isn't quite right here, and the block is a bit hard to read (but that's my personal preference)

src/reconcile/rec_write.c
```diff
@@ -2350,39 +2398,132 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+ *	Write the previous split chunk to the disk as a page.
+ *	Shift the contents of the current page to the start of the buffer,
+ *	making space for a new chunk to be written.
+ *	If the caller asks for a chunk resizing, the boundary between the two
+ *	chunks is readjusted to the alternate boundary details stored in the
+ *	previous chunk, letting the current chunk grow at the cost of the
+ *	previous chunk.
+ */
+static int
+__rec_split_write_prev_shift_cur(
```
agorrod: I'd like an ``and`` in the middle

src/reconcile/rec_write.c
```diff
@@ -2350,39 +2398,132 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+    WT_SESSION_IMPL *session, WT_RECONCILE *r, bool resize_chunks)
+{
+	WT_BTREE *btree;
+	WT_DECL_ITEM(tmp_buf);
+	WT_DECL_RET;
+	WT_BOUNDARY *bnd_cur, *bnd_prev;
```
agorrod: ``WT_BOUNDARY`` belongs above ``WT_BTREE``

src/reconcile/rec_write.c
```diff
@@ -2350,39 +2398,132 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+	len = WT_PTRDIFF(r->first_free, dsk) - bnd_cur->offset;
+
+	/*
+	 * Resize chunks if the current is smaller than the minimum,
+	 * and there is an alternate boundary available in the previous
+	 * boundary details.
+	 */
+	if (resize_chunks &&
+	    len < r->alt_split_size &&
+	    bnd_prev->alt_offset != 0) {
```
agorrod: Our code style says to use the minimum number of lines

src/reconcile/rec_write.c
```diff
@@ -2350,39 +2398,132 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+		bnd_cur->offset = bnd_prev->alt_offset;
+		bnd_cur->entries += bnd_prev->entries - bnd_prev->alt_entries;
+		bnd_prev->entries = bnd_prev->alt_entries;
+		bnd_cur->recno = bnd_prev->alt_recno;
+
+		// Are we overwriting here, and not losing the memory that was
```
agorrod: Comment style

src/reconcile/rec_write.c
```diff
@@ -2391,147 +2532,58 @@ __rec_split(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len)
...
-		 * We can get here if the first key/value pair won't fit.
-		 * Additionally, grow the buffer to contain the current item if
-		 * we haven't already consumed a reasonable portion of a split
-		 * chunk.
-		 */
-		if (inuse < r->split_size / 2)
-			break;
+	last = &r->bnd[r->bnd_next];
+	inuse = WT_PTRDIFF(r->first_free, dsk)
+	    - last->offset + WT_PAGE_HEADER_BYTE_SIZE(btree);
```
agorrod: The ``-`` should be on the previous line. Better to add braces to avoid ambiguity

src/reconcile/rec_write.c
```diff
@@ -2391,147 +2532,58 @@ __rec_split(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len)
...
-		 * Set the key for the next chunk (before writing the block, a
-		 * key range is needed in that code).
-		 */
-		next->recno = r->recno;
-		if (dsk->type == WT_PAGE_ROW_INT ||
-		    dsk->type == WT_PAGE_ROW_LEAF)
-			WT_RET(__rec_split_row_promote(
-			    session, r, &next->key, dsk->type));
+	/*
+	 * Hitting a page boundary resets the dictionary,
```
agorrod: This comment fits on a single line now.

src/reconcile/rec_write.c
```diff
@@ -3063,20 +3096,66 @@ __rec_split_finish_std(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+			 */
+			bnd_prev = bnd_cur - 1;
+			bnd_prev->entries += bnd_cur->entries;
+			r->bnd_next--;
+			grow_bnd = false;
+		}
+	}
+
+	/*
+	 * We need an extra boundary at the end, we use it to compare the keys
```
agorrod: This comment should be updated

src/reconcile/rec_write.c
```diff
@@ -3688,25 +3673,27 @@ __wt_bulk_insert_row(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk)
...
-				if (!ovfl_key)
-					WT_RET(__rec_cell_build_leaf_key(
-					    session, r, NULL, 0, &ovfl_key));
-			}
+	if (r->raw_compression) {
+		if (key->len + val->len > r->space_avail)
+			WT_RET(__rec_split_raw(
+			    session, r, key->len + val->len));
+	}
+	else if (WT_CHECK_CROSSING_BND(r, key->len + val->len)) {
```
agorrod: Merge two lines

src/reconcile/rec_write.c
```diff
@@ -5054,6 +5056,68 @@ __rec_row_merge(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
...
+		if (alt_offset == WT_PAGE_HEADER_BYTE_SIZE(btree)) {
+			/*
+			 * This is possible if the first record doesn't fit in
+			 * the alternate boundary size, we write this record
+			 * without setting up an alternate boundary. We will get
+			 * the opportunity to setup the alternate boundary
+			 * before writing out the next record.
+			 */
+			return (0);
+		}
```
agorrod: Remove brackets around single line ``if``

api/leveldb/leveldb_wt.h
```diff
@@ -52,7 +52,7 @@
         "session_max=8192,mmap=false,"                                  \
         "transaction_sync=(enabled=true,method=none),"
 // Note: LSM doesn't split, build full pages from the start
-#define WT_TABLE_CONFIG "type=lsm,split_pct=100,leaf_item_max=1KB,"	\
```
sulabhM: confirm is it is okay to split at the default 90% here

daveh86:
retest this please

src/reconcile/rec_write.c
```diff
@@ -1974,40 +1992,79 @@ __rec_split_bnd_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+ *	chunk, hence increasing the size of the last page without decreasing the
+ *	penultimate page size beyond the alternate split size.
+ */
+static uint32_t
+__wt_alt_split_page_size(WT_SESSION_IMPL *session, uint32_t maxpagesize)
+{
+	WT_BTREE *btree;
+
+	btree = S2BT(session);
+#define	WT_ALT_SPLIT_PCT 45
```
sulabhM: As discussed offline, this can be anything that is less than the split pct chosen. Since we are restricting split percentage between 50 and 100, alt_split_pct can be anything < 50. 45 is just a random value that I picked for now.

sulabhM:
@keithbostic , can you review this change please. This is ready for a final pass of reviews.

src/reconcile/rec_write.c
```diff
@@ -26,6 +26,11 @@ typedef struct {
 	uint32_t flags;			/* Caller's configuration */
 
 	WT_ITEM	 disk_image;		/* Temporary disk-image buffer */
+	/*
+	 * Temporary buffer used to write out a disk image when managing two
+	 * chunks worth of data in memory
+	 */
+	WT_ITEM *interim_buf;
```
sulabhM: yes, I see performance improvement when I create the buffer once and keep it around, instead of creating and destroying each time in __rec_split_write_prev_and_shift_cur()

src/reconcile/rec_write.c
```diff
@@ -187,35 +193,35 @@ typedef struct {
...
+		 * written without decreasing the penultimate page size beyond
+		 * the minimum split size.
+		 */
+		uint32_t min_bnd_entries; /* No of entries at min split pct */
+		/* Next chunk's starting offset at min split pct */
+		size_t min_bnd_offset;
+		/* Next chunk's starting record number at min split pct */
+		uint64_t min_bnd_recno;
+		/* Next chunk's starting key at min split pct */
+		WT_ITEM min_bnd_key;
```
sulabhM: I like the idea of renaming boundary information as max/min. Though min suggests a global minimum of 50% (the one user can set and the one we will readjust to if the last chunk is not large enough), max will be 50 to 100 %, equal to the split percentage user has configured. Max doesn't signify the maximum split percentage that can be set (100%). If "max" doesn't create a confusion like that, I am happy to rename.

You are right, we do not need min_XXX for all the boundaries. Initially I thought of and discussed that approach, and then decided to have this information for each boundary. With respect to performance, I don't think there will be any difference and I suppose extra information per boundary doesn't consume too much memory either. It looked simpler and cleaner to track per boundary. Having only one set of min_XXX fields requires me to have an additional bool flag to determine the state of this information for the current boundary. At the moment I initialise min_bnd_offset to 0, and can check if I have crossed the minimum by checking if it is still 0, not possible without extra field or more complex calculations on the offset and current size if maintaining one set of min_XXX fields.

If you don't have strong reasons against tracking information for all the boundaries, I am inclined to keep it.

src/reconcile/rec_write.c
```diff
@@ -1974,40 +1998,64 @@ __rec_split_bnd_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+ *	Minimum split size boundary calculation: To track a boundary at the
+ *	minimum split size that we could have split at instead of splitting at
+ *	the split page size.
+ */
+static uint32_t
+__rec_min_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
+{
+	return (__rec_split_page_size_from_pct(
+	    WT_BTREE_MIN_SPLIT_PCT, maxpagesize, btree->allocsize));
+}
```
sulabhM: You get the idea, I am wrapping around a common functionality to generate two functions. Earlier revisions of this change had more functionality in these two functions and it made more sense, now they both are single lined. I thought of doing #define for them, but __wt_split_page_size() gets accessed from src/btree/bt_handle.c as well, and I didn't want to put one of the defines in a header file and the other one here. Since these functions are not in any performance path, I just left them as single lined for now. I am happy to inline them or #define them. Unless you suggest a change, I will keep them like this.

src/reconcile/rec_write.c
```diff
@@ -2350,19 +2399,105 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+	dsk = r->disk_image.mem;
+	cur_len = WT_PTRDIFF(r->first_free, dsk) - bnd_cur->offset;
+
+	/*
+	 * Resize chunks if the current is smaller than the minimum, and there
+	 * are details on the minimum split size boundary available in the
+	 * previous boundary details.
+	 */
+	if (resize_chunks &&
+	    cur_len < r->min_split_size && bnd_prev->min_bnd_offset != 0) {
```
sulabhM: This function will NOT be called if we have only one boundary, I have put assert to validate that: WT_ASSERT(session, r->bnd_next != 0)

Ideally we shouldn't have bnd_prev->min_bnd_offset == 0 here. It is possible, when say while writing to the chunk we get a K/V pair that is large enough that we cross the minimum and the split_size at the same time. The minimum and split boundary will be the same, and it wont make sense to set the min_XXX fields then. This should not happen frequently, because the buffer will adjust to the oversized items, but is still possible if this is the first K/V pair that is being written. Check for "bnd_prev->min_bnd_offset == 0" acts as a failsafe in such cases.

src/reconcile/rec_write.c
```diff
@@ -3680,25 +3655,26 @@ __wt_bulk_insert_row(WT_SESSION_IMPL *session, WT_CURSOR_BULK *cbulk)
...
-				r->key_pfx_compress = false;
-				if (!ovfl_key)
-					WT_RET(__rec_cell_build_leaf_key(
-					    session, r, NULL, 0, &ovfl_key));
-			}
+	if (r->raw_compression) {
+		if (key->len + val->len > r->space_avail)
+			WT_RET(__rec_split_raw(
+			    session, r, key->len + val->len));
+	} else if (WT_CHECK_CROSSING_BND(r, key->len + val->len)) {
```
sulabhM: That is what I did in the earlier revisions. I was getting perf degradation in the highly optimised bulk insert case. I fixed some amount of that degradation, but couldn't get rid of all of it. Most of it came from memmove of the second chunk to the start of interim buffer.

Also, since bulk load creates pages of page_size (split_pct as 100), it doesn't make sense to do all this change for this path (like it doesn't make sense for fixed size column store).
So I am bypassing my change for the bulk load routine.

src/reconcile/rec_write.c
```diff
@@ -293,6 +299,15 @@ typedef struct {
...
 } WT_RECONCILE;
 
+#define	WT_CROSSING_MIN_BND(r, next_len)			\
+	(!(r)->is_bulk_load && (r)->bnd[(r)->bnd_next].min_bnd_offset == 0 && \
+	    ((r)->space_avail - (next_len)) <			\
+	    ((r)->split_size - (r)->min_split_size))
+#define	WT_CROSSING_SPLIT_BND(r, next_len) ((next_len) > (r)->space_avail)
+#define	WT_CHECK_CROSSING_BND(r, next_len)			\
+	(WT_CROSSING_MIN_BND(r, next_len) || WT_CROSSING_SPLIT_BND(r, next_len))
+
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -293,6 +299,15 @@ typedef struct {
...
+static uint32_t __rec_min_split_page_size(WT_BTREE *, uint32_t);
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -1943,6 +1964,9 @@ __rec_split_bnd_init(WT_SESSION_IMPL *session, WT_BOUNDARY *bnd)
 
 	bnd->already_compressed = false;
 
+	bnd->min_bnd_offset = 0;
+	bnd->min_bnd_entries = 0;
+
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -2122,7 +2150,42 @@ __rec_split_init(WT_SESSION_IMPL *session,
 		r->split_size = __wt_split_page_size(btree, r->page_size);
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -2350,19 +2399,105 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+	    cur_len < r->min_split_size && bnd_prev->min_bnd_offset != 0) {
+		bnd_cur->offset = bnd_prev->min_bnd_offset;
+		bnd_cur->entries +=
+		    bnd_prev->entries - bnd_prev->min_bnd_entries;
+		bnd_prev->entries = bnd_prev->min_bnd_entries;
+		bnd_cur->recno = bnd_prev->min_bnd_recno;
+
+		__wt_buf_set(
+		    session, &bnd_cur->key,
+		    bnd_prev->min_bnd_key.data, bnd_prev->min_bnd_key.size);
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -2350,19 +2399,105 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+		    session, prev_len_aligned, &r->interim_buf));
+	else
+		WT_RET(__wt_buf_init(
+		    session, r->interim_buf, prev_len_aligned));
+
+	dsk_tmp = r->interim_buf->mem;
+	memcpy(dsk_tmp, dsk, bnd_cur->offset);
+	dsk_tmp->recno = bnd_prev->recno;
+	dsk_tmp->u.entries = bnd_prev->entries;
+	dsk_tmp->mem_size = (uint32_t)bnd_cur->offset;
```
sulabhM: That is what I intended to do, have a single disk_image large enough to accomodate two chunks, write the first chunk out and memmove the second to the start of the buffer, without using an interim buffer. I made that change and figured that if you provide a buffer and a size to __rec_split_write(), it doesn't guarantee that it wont modify that buffer past that size. In-fact it appends zeroes at the end of the given size, I suppose to better layout the page on disk, hence overwriting into the second chunk. It took some amount of memory disassembly to figure that bug in my initial revision. I discussed this with Michael, and went ahead with an interim buffer.
I was concerned about performance degradation because of having to copy each time to an interim buffer, but my tests show that memmove is a much bigger offender.

src/reconcile/rec_write.c
```diff
@@ -2391,147 +2529,62 @@ __rec_split(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len)
...
+	    dsk->type == WT_PAGE_ROW_LEAF)
+		WT_RET(__rec_split_row_promote(
+		    session, r, &next->key, dsk->type));
 
+	r->entries = 0;
 	/*
+	 * Set the space available to another split-size chunk.
+	 */
+	r->space_avail =
+	    r->split_size - WT_PAGE_HEADER_BYTE_SIZE(btree);
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -3063,20 +3090,63 @@ __rec_split_finish_std(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+		if (WT_PTRDIFF(r->first_free, dsk) > r->page_size &&
+		    r->bnd_next != 0) {
+			/*
+			 * We hold two boundaries worth of data in the buffer,
+			 * and this data doesn't fit in a single page.  If the
+			 * last chunk is too small, readjust the boundary to a
+			 * pre-computed minimum.
+			 * Write out the penultimate chunk to the disk as a page
+			 */
+			__rec_split_write_prev_and_shift_cur(session, r, true);
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -3844,10 +3824,11 @@ __wt_bulk_insert_var(
...
 	/* Boundary: split or write the page. */
-	if (val->len > r->space_avail)
-		WT_RET(r->raw_compression ?
-		    __rec_split_raw(session, r, val->len) :
-		    __rec_split(session, r, val->len));
+	if (r->raw_compression) {
+		if (val->len > r->space_avail)
+			WT_RET(__rec_split_raw(session, r, val->len));
+	} else if (WT_CHECK_CROSSING_BND(r, val->len))
+		WT_RET(__rec_split_crossing_bnd(session, r, val->len));
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -5046,6 +5037,67 @@ __rec_row_merge(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
 }
 
 /*
+ * __rec_split_crossing_bnd --
+ * 	Save the details for the minimum split size boundary or call for a
+ * 	split.
+ */
+static inline int
+__rec_split_crossing_bnd(
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -5046,6 +5037,67 @@ __rec_row_merge(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
...
+	 * large enough, just split at this point.
+	 */
+	if (WT_CROSSING_MIN_BND(r, next_len) &&
+	    !WT_CROSSING_SPLIT_BND(r, next_len)) {
+		btree = S2BT(session);
+		bnd = &r->bnd[r->bnd_next];
+		dsk = r->disk_image.mem;
+		min_bnd_offset =
+		    WT_PTRDIFF(r->first_free, dsk) - bnd->offset
+		    + WT_PAGE_HEADER_BYTE_SIZE(btree);
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
```diff
@@ -5460,29 +5507,28 @@ __rec_row_leaf_insert(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins)
...
+			 * working with an overflow key), rebuild the
+			 * key without compression.
+			 */
+			if (r->key_pfx_compress_conf) {
+				r->key_pfx_compress = false;
+				if (!ovfl_key)
+					WT_RET(
+					    __rec_cell_build_leaf_key(
+					    session,
+					    r, NULL, 0, &ovfl_key));
```
sulabhM: fixed in latest revision

src/reconcile/rec_write.c
keithbostic: > Max doesn't signify the maximum split percentage that can be set (100%). If "max" doesn't create a confusion like that, I am happy to rename.

I think of "max" as the "maximum for this boundary", so I don't think it creates confusion.

> At the moment I initialise min_bnd_offset to 0, and can check if I have crossed the minimum by checking if it is still 0,

I'm not sure why that same check wouldn't work?

I still think it's worthwhile moving the minimum boundary information out of the boundary structure. It's an additional 64B and we can have thousands of these structures.

src/reconcile/rec_write.c
keithbostic: > Check for "bnd_prev->min_bnd_offset == 0" acts as a failsafe in such cases.

I think that's worth a comment.

src/reconcile/rec_write.c
keithbostic: > In-fact it appends zeroes at the end of the given size, I suppose to better layout the page on disk, hence overwriting into the second chunk.

Ouch. OK.

> but my tests show that memmove is a much bigger offender.

Which memmove is that?

src/reconcile/rec_write.c
keithbostic: I think we're miscommunicating -- what I'm suggesting is something like this:
```
*** rec_write.c.orig    Tue Mar 28 06:47:56 2017
--- rec_write.c.new   Tue Mar 28 07:50:52 2017
***************
*** 300,307 ****
  } WT_RECONCILE;

  #define       WT_CROSSING_MIN_BND(r, next_len)
\
!       (!(r)->is_bulk_load &&                                          \
!           (r)->bnd[(r)->bnd_next].min_bnd_offset == 0 &&              \
            ((r)->space_avail - (next_len)) <                           \
            ((r)->split_size - (r)->min_split_size))
  #define       WT_CROSSING_SPLIT_BND(r, next_len) ((next_len) > (r)->space_avai
l)
--- 300,306 ----
  } WT_RECONCILE;

  #define       WT_CROSSING_MIN_BND(r, next_len)
\
!       ((r)->bnd[(r)->bnd_next].min_bnd_offset == 0 &&         \
            ((r)->space_avail - (next_len)) <                           \
            ((r)->split_size - (r)->min_split_size))
  #define       WT_CROSSING_SPLIT_BND(r, next_len) ((next_len) > (r)->space_avai
l)
***************
*** 3711,3717 ****
                if (key->len + val->len > r->space_avail)
                        WT_RET(__rec_split_raw(
                            session, r, key->len + val->len));
!       } else if (WT_CHECK_CROSSING_BND(r, key->len + val->len)) {
                /*
                 * Turn off prefix compression until a full key written to the
                 * new page, and (unless already working with an overflow key),
--- 3710,3716 ----
                if (key->len + val->len > r->space_avail)
                        WT_RET(__rec_split_raw(
                            session, r, key->len + val->len));
!       } else if (WT_CROSSING_SPLIT_BND(r, key->len + val->len)) {
                /*
                 * Turn off prefix compression until a full key written to the
                 * new page, and (unless already working with an overflow key),
***************
*** 3880,3886 ****
                if (val->len > r->space_avail)
                        WT_RET(__rec_split_raw(session, r, val->len));
        } else
!               if (WT_CHECK_CROSSING_BND(r, val->len))
                        WT_RET(__rec_split_crossing_bnd(session, r, val->len));

        /* Copy the value onto the page. */
--- 3879,3885 ----
                if (val->len > r->space_avail)
                        WT_RET(__rec_split_raw(session, r, val->len));
        } else
!               if (WT_CROSSING_SPLIT_BND(r, val->len))
                        WT_RET(__rec_split_crossing_bnd(session, r, val->len));

        /* Copy the value onto the page. */
```

src/reconcile/rec_write.c
```diff
@@ -2350,19 +2395,104 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+	 * Create an interim buffer if not already done to prepare the previous
+	 * chunk's disk image.
+	 */
+	prev_len_aligned = WT_ALIGN(bnd_cur->offset, btree->allocsize);
+	if (r->interim_buf == NULL)
+		WT_RET(__wt_scr_alloc(
+		    session, prev_len_aligned, &r->interim_buf));
+	else
+		WT_RET(__wt_buf_init(
+		    session, r->interim_buf, prev_len_aligned));
```
keithbostic: I just realized this is wrong. Reconciliation has to ask the block manager how much space it needs to write a chunk of data, so it  should be something like this:
```
diff --git a/src/reconcile/rec_write.c b/src/reconcile/rec_write.c
index 0ac5b34db..7c0633d2c 100644
@@ -2431,15 +2431,17 @@ static int
 __rec_split_write_prev_and_shift_cur(
     WT_SESSION_IMPL *session, WT_RECONCILE *r, bool resize_chunks)
 {
+       WT_BM *bm;
        WT_BOUNDARY *bnd_cur, *bnd_prev;
        WT_BTREE *btree;
        WT_PAGE_HEADER *dsk, *dsk_tmp;
-       size_t cur_len, prev_len_aligned;
+       size_t cur_len, len;
        uint8_t *dsk_start;
 
        WT_ASSERT(session, r->bnd_next != 0);
 
        btree = S2BT(session);
+       bm = btree->bm;
        bnd_cur = &r->bnd[r->bnd_next];
        bnd_prev = bnd_cur - 1;
        dsk = r->disk_image.mem;
@@ -2469,13 +2471,12 @@ __rec_split_write_prev_and_shift_cur(
         * Create an interim buffer if not already done to prepare the previous
         * chunk's disk image.
         */
-       prev_len_aligned = WT_ALIGN(bnd_cur->offset, btree->allocsize);
+       len = bnd->cur_offset;
+       WT_RET(bm->write_size(bm, session, &len));
        if (r->interim_buf == NULL)
-               WT_RET(__wt_scr_alloc(
-                   session, prev_len_aligned, &r->interim_buf));
+               WT_RET(__wt_scr_alloc(session, len, &r->interim_buf));
        else
-               WT_RET(__wt_buf_init(
-                   session, r->interim_buf, prev_len_aligned));
+               WT_RET(__wt_buf_init(session, r->interim_buf, len));
 
        dsk_tmp = r->interim_buf->mem;
        memcpy(dsk_tmp, dsk, bnd_cur->offset);
```

src/reconcile/rec_write.c
keithbostic: I'm wondering if the fact we're using `memmove` is the problem? We should switch to using `memcpy` because the strings do not overlap (they're in the same buffer, but they're not overlapping). I didn't think that made any difference on a modern compiler/library, but maybe it does?

src/reconcile/rec_write.c
```diff
@@ -4961,25 +5008,25 @@ __rec_row_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
...
+			/*
+			 * In one path above, we copied address blocks
+			 * from the page rather than building the actual
+			 * key.  In that case, we have to build the key
+			 * now because we are about to promote it.
+			 */
+			if (key_onpage_ovfl) {
+				WT_ERR(__wt_buf_set(session, r->cur,
+				    WT_IKEY_DATA(ikey), ikey->size));
+				key_onpage_ovfl = false;
```
keithbostic: @sulabhM, there are 4 more places where the if/else don't match up.

keithbostic:
@sulabhM, I think I'm good to go on this one when you're ready. In summary, my remaining comments:

* I think the min values should move outside the boundary structure.
* I still like renaming boundary structures to min/max so they match.
* I'd like to understand why the buffer copy isn't showing up in your performance measurements; we always could avoid the copy in the case of `btree->kencryptor != NULL`, but that violates some layering and other solutions are messier. (Your good idea goes here!)
* I think we can call `WT_CROSSING_SPLIT_BND` in the bulk-load routines, and then remove the test of `r->bulk_load` in those macros, that is, we know by what function we're in if it's worth testing `r->bulk_load`.
* There's a bug in sizing the interim buffer, we have to ask the block manager how large it needs to be.
* There are still 4 (I think) places that we have if/else constructs that read badly, that is, it's something like:
```
if (raw->compression) {
        if (boundary conditions)
} else if (boundary conditions) {
}
```
and I much prefer:
```
if (raw->compression) {
        if (boundary conditions)
} else
        if (boundary conditions) {
        }
```
Ignore the comment I made about `memmove`, I was wrong, we can't switch those calls to `memcpy`.

I don't feel terribly strongly about the style comments at this point, if you & @michaelcahill disagree, happy to go with you two.

In short, lgtm.

src/reconcile/rec_write.c
sulabhM: Added a comment here

src/reconcile/rec_write.c
sulabhM: renamed to max/min. Had a discussion with Keith, if we want one set of min_XXX, will need to add an additional field that will say if for current chunk minimum boundary has been set or not. We decided to not make that change for now. Will follow up in another ticket on improvements that we can make on top of this change.

src/reconcile/rec_write.c
sulabhM: Will follow up in a new ticket on making improvements, like using two buffers that can be interchanged every split, hence avoiding an interim buffer and memory operations.

src/reconcile/rec_write.c
sulabhM: Fixed this in latest revision.

src/reconcile/rec_write.c
sulabhM: Fixed this bug in latest revision.

src/reconcile/rec_write.c
sulabhM: Fixed in the latest revision

michaelcahill:
Thanks @sulabhM and @keithbostic for this epic change.  LGTM, merging.

PR #3278, opened 2017-02-03
WT-3161 Panic on a write error in logging.

sueloverso:
@agorrod and @ddanderson Here's a fix for the logging hang.  Please review.  I changed it so that if writing to the log file gets an error we panic.


ddanderson:
@sueloverso, I ran it against the wt2909 test a few times and it passes. The code change LGTM, but I don't feel qualified to to make a deep judgement about how it affects logging.

sueloverso:
@agorrod Give this a look for merging please.

src/log/log.c
```diff
@@ -21,6 +21,27 @@ static int __log_write_internal(
 #define	WT_LOG_OPEN_VERIFY	0x02
 
 /*
+ * __log_panic --
+ *	Report an error, then panic the system.
+ */
+static int
```
agorrod: I believe the ``WT_PANIC_MSG`` macro mirrors the functionality you added here.

I'd be happy to change our ``__wt_panic`` to require an error message instead - I think that's a reasonable requirement for an operation that intends to panic the database.

I looked and the check for ``WT_CONN_PANIC`` on API entry doesn't result in another call to ``__wt_panic``, so changing that function shouldn't result in getting an error message from every thread.

sueloverso: I think `WT_PANIC_MSG` is the right thing here.  Fixed.

agorrod:
Thanks @sueloverso lgtm

PR #3280, opened 2017-02-05
WT-3088 bug: WiredTiger can evict the tree's current eviction walk point

keithbostic:


agorrod:
@keithbostic I believe this change closes a window, but I also think the worst consequence is an assertion firing (which we've never seen). Would you like me to merge, or do you anticipate having other changes for WT-3088 as well?

keithbostic:
> I believe this change closes a window, but I also think the worst consequence is an assertion firing (which we've never seen). 

I believe that's correct.

This isn't worth merging yet (this change might even go away once we understand WT-3088).

src/evict/evict_lru.c
```diff
@@ -1838,14 +1837,16 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		if (__wt_ref_is_root(ref) || evict == start || give_up ||
 		    ref->page->read_gen == WT_READGEN_OLDEST ||
 		    ref->page->memory_footprint >= btree->splitmempage) {
-			btree->evict_ref = ref;
-			WT_RET(__evict_clear_walk(session, restarts == 0));
-		} else if (ref->page->read_gen == WT_READGEN_OLDEST) {
+			if (restarts == 0)
+				WT_STAT_CONN_INCR(
+				    session, cache_eviction_walks_abandoned);
+			WT_RET(__wt_page_release(cache->walk_session,
```
agorrod: Switching the order of clearing ``evict_ref`` and releasing the page is contrary to the comment in ``__evict_clear_walk`` that reads:

```
  * Clear evict_ref first, in case releasing it forces eviction (we
  * assert we never try to evict the current eviction walk point).
```

Oh, I see now. In the larger change ``evict_ref`` is never set while the tree is actually being walked. I don't think the new comment is quite adequate, but I think the change is OK

agorrod:
@keithbostic I took a look at these changes and they all look correct to me.

I feel like this bug fix is likely to be back ported, and something I'm probably going to dig up again in the future. As such I felt more comfortable with the bug fix split from the cleanup changes. I've gone ahead and made that change (https://github.com/wiredtiger/wiredtiger/pull/3284).

If you are happy with both branches I'm happy for them to be merged.

keithbostic:
@agorrod,
> As such I felt more comfortable with the bug fix split from the cleanup changes.

Makes sense, I should have thought of that!

keithbostic:
LGTM, merging.

PR #3281, opened 2017-02-06
WT-3157 Alternative checkpoint cleanup.

michaelcahill:


michaelcahill:
@sueloverso, I know there was some discussion about this earlier, but `__checkpoint_apply` doesn't seem like the best place to do this cleanup.  There is already cleanup (including marking the connection dirty) in the error handler for `__wt_txn_checkpoint` (the only caller of `__checkpoint_apply`).  Additionally, there is an identical loop over all of the handles there in order to release them.

Given that, how about we mark btree handles dirty in the same place?

sueloverso:
This ended up quite clean so let's go with it.  I ran several iterations of the failing test in Don's branch too.  I'll merge all the way up.

PR #3282, opened 2017-02-06
WT-3123 Have thread active/pause wrapper in thread group code.

sueloverso:
@agorrod Here are the changes to make the thread group code the wrapper for the activate and pause of threads.  It required a bunch of changes.  I think conceptually it is cleaner, but in some ways the eviction code is a bit less intuitive since it is no longer a server loop.  I had to add both a check function and a stop function for threads to call.  Let me know what you think.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -282,7 +282,8 @@ __wt_evict_thread_chk(WT_SESSION_IMPL *session)
 
 /*
  * __wt_evict_thread_run --
- *	Starting point for an eviction thread.
+ *	Entry function for an eviction thread.  This is called repeatedly
+ *	from the thread group code so it does not need to loop itself.
  */
 int
 __wt_evict_thread_run(WT_SESSION_IMPL *session, WT_THREAD *thread)
@@ -295,14 +296,10 @@ __wt_evict_thread_run(WT_SESSION_IMPL *session, WT_THREAD *thread)
 	conn = S2C(session);
 	cache = conn->cache;
 
-#if defined(HAVE_DIAGNOSTIC) || defined(HAVE_VERBOSE)
 	/*
-	 * Ensure the cache stuck timer is initialized when starting eviction.
+	 * The thread group code calls us repeatedly.  So each call is one pass
+	 * through eviction.
 	 */
-	if (thread->id == 0)
-		__wt_epoch(session, &cache->stuck_ts);
-#endif
-
 	if (conn->evict_server_running &&
 	    __wt_spin_trylock(session, &cache->evict_pass_lock) == 0) {
 		/*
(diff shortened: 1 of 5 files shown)


src/support/thread_group.c
```diff
@@ -22,11 +22,21 @@ __wt_thread_run(void *arg)
 	thread = (WT_THREAD*)arg;
 	session = thread->session;
 
+	for (;;) {
+		if (!F_ISSET(thread, WT_THREAD_RUN))
+			return (WT_THREAD_RET_VALUE);
```
agorrod: I think it would be better if all paths resulted in a call to the stop function

src/support/thread_group.c
```diff
@@ -22,11 +22,21 @@ __wt_thread_run(void *arg)
...
+		if (!F_ISSET(thread, WT_THREAD_ACTIVE))
+			__wt_cond_wait(session, thread->pause_cond,
+			    WT_THREAD_PAUSE * WT_MILLION, thread->chk_func);
```
agorrod: Pausing for 10 seconds here makes me nervous. I know the condition variable is signaled if something more urgent is required, but we've seen cases before where having such long sleeps can lead to very slow shutdown (for example). I'd be happier if this was 10x/second - 0.1 of a second isn't really noticeable.

src/evict/evict_lru.c
```diff
@@ -303,79 +303,81 @@ __wt_evict_thread_run(WT_SESSION_IMPL *session, WT_THREAD *thread)
 		__wt_epoch(session, &cache->stuck_ts);
```
agorrod: Will this timer be re-initialized more regularly with this change?

src/evict/evict_lru.c
```diff
@@ -303,79 +303,81 @@ __wt_evict_thread_run(WT_SESSION_IMPL *session, WT_THREAD *thread)
...
-				    WT_SESSION_LOCKED_PASS);
-				ret = __evict_server(session, &did_work);
-				F_CLR(cache->walk_session,
-				    WT_SESSION_LOCKED_PASS);
-				F_CLR(session, WT_SESSION_LOCKED_PASS);
-				was_intr = cache->pass_intr != 0;
-				__wt_spin_unlock(session,
-				    &cache->evict_pass_lock);
-				WT_ERR(ret);
+	if (conn->evict_server_running &&
```
agorrod: I actually like this structure more, but I think it deserves a comment explaining that the thread group repeatedly calls this function.

src/support/thread_group.c
sueloverso: Good catch.  I agree.

src/support/thread_group.c
sueloverso: I strongly disagree with this.  Paused threads shouldn't even exist. I.e. this mechanism is replacing thread create/destroy.  We have customers who run 40 WT databases.  If most of those use the default eviction threads that is several hundred threads, most of which are paused.  There have been tickets where internal worker threads doing actual work waking 0.1 second was too much for that scenario and wasted a lot of CPU.  The algorithm is supposed to settle on a stable number so this is an experiment to see if keeping these threads existing is better than create/destroy.

src/evict/evict_lru.c
sueloverso: Done

src/evict/evict_lru.c
sueloverso: Yes, good catch and a bad side effect.  My choices are to move it, or add an initialization function for threads in a group.  I'll move it.

agorrod:
Thanks @sueloverso I'm going to go ahead and merge this into the other branch, so we can do a full review together.

PR #3283, opened 2017-02-08
WT-3164 Reset ckpt field on error.  Add error reset function.

sueloverso:
@agorrod and @ddanderson Here is a fix for WT-3164 that is based off develop, but tested with Don's branch.  With this fix (and my hand placed error per the ticket) the command now proceeds further and fails with the same error as WT-3165.  Without any hand-injected error at that specific location the test command succeeds.  But the fix needs to be in develop and can then be merged into Don's branch.  Please review.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -525,11 +525,11 @@ __checkpoint_verbose_track(WT_SESSION_IMPL *session,
 }
 
 /*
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @sueloverso lgtm - I changed the name of the function and am going to merge.

sueloverso:
Thanks.  I like your name better.

PR #3284, opened 2017-02-09
WT-3170 Change how eviction walk point is saved during walk.

agorrod:


agorrod:
This was an offshoot of chasing WT-3088. @keithbostic Would you please review?

src/evict/evict_lru.c
```diff
@@ -773,14 +773,14 @@ __evict_clear_walk(WT_SESSION_IMPL *session, bool count_stat)
...
+	WT_STAT_CONN_INCR(session, cache_eviction_walks_abandoned);
 
 	/*
-	 * Clear evict_ref first, in case releasing it forces eviction (we
-	 * assert we never try to evict the current eviction walk point).
+	 * Clear evict_ref before releasing it in case that forces eviction (we
+	 * assert that we never try to evict the current eviction walk point).
 	 */
-	btree->evict_ref = NULL;
+	WT_PUBLISH(btree->evict_ref, NULL);
```
keithbostic: @agorrod, I think this change is wrong.

`WT_PUBLISH` is a write-barrier before the assignment, and what we need here is to ensure the update of `btree->evict_ref` be done before any subsequent reads of the value.  I think this should be:
```
btree->evict_ref = NULL;
WT_FULL_BARRIER();
```
I've made this same mistake. @michaelcahill, how about:
```
/*
 * Write a shared location and guarantee that subsequent reads do not see any
 * earlier state.
 */
#define WT_ORDERED_WRITE(v, val) do {
        (v) = (val);
        WT_FULL_BARRIER();
} while (0)
```

src/evict/evict_lru.c
```diff
@@ -1662,8 +1662,15 @@ __evict_walk_file(WT_SESSION_IMPL *session,
...
 	/*
-	 * Get some more eviction candidate pages.
-	 *
+	 * Get some more eviction candidate pages, starting at the last saved
+	 * point. Clear the saved point immediately, we assert when discarding
+	 * pages we're not discarding an eviction point, so this clear must be
+	 * complete before the page is released.
+	 */
+	ref = btree->evict_ref;
+	WT_PUBLISH(btree->evict_ref, NULL);
```
keithbostic: Ditto, this change -- we need a barrier after the assignment, not before.

keithbostic:
@agorrod, @michaelcahill: I pushed some cleanup in f376415. With 0b9e453, I don't think we do any non-DIAGNOSTIC work after completing the split, so we can skip that entire step. That should make our splits marginally faster, we won't separately walk through all the child internal pages created by the split (that are still in the cache).

I didn't do this cleanup in PR #3280 because it wasn't required for the bug fix.

src/evict/evict_lru.c
michaelcahill: I disagree -- first up, why do you think any barrier is required here?

Second, I want to avoid sprinkling explicit barriers around, so some more general mechanism would be better (and at least the pairing of `WT_PUBLISH` and `WT_ORDERED_READ` has reasonable mapping into the C11 memory model).

Thirdly, a read barrier here implies that you're worried that *this* thread is going to read something from before it set `evict_ref`.  What reads are you concerned about ordering in that way?

src/evict/evict_lru.c
keithbostic: I retract my comment about line 783, the fact this thread is the one to discard the page means it's OK.

I still think the barrier is required at line 1671, I don't see any guarantee that setting `btree.evict_ref` won't be delayed until after another thread tests in `__wt_ref_out` that `btree.evict_ref != the ref being discarded`.

keithbostic: @michaelcahill covered this one the call; I'll back out these changes entirely.

src/btree/bt_split.c
```diff
@@ -1180,16 +1165,13 @@ __split_internal(WT_SESSION_IMPL *session, WT_PAGE *parent, WT_PAGE *page)
 	WT_ASSERT(session, WT_INTL_INDEX_GET_SAFE(page) == pindex);
 	WT_INTL_INDEX_SET(page, replace_index);
 
+	/* The split is live. */
+	WT_WRITE_BARRIER();
```
michaelcahill: There's already a write barrier in `WT_INTL_INDEX_SET` before the pointer is published, which I think it is in the right order.  In other words, it guarantees that everything in the allocated index is set *before* the pointer becomes visible to other threads.

What does a barrier after the pointer is visible achieve?

michaelcahill:
@keithbostic, I think this change is right but I'd like to stare at `bt_split.c` some more on Monday.

src/btree/bt_split.c
keithbostic: I didn't review the barrier code when I made that change, I just preserved it in place.

But, yes, I agree with you, and looking more closely, there are a few barrier calls that aren't needed and one that I think is incorrect.

I've created a PR #3288 with my suggested changes to the barriers in the split code, for your consideration.

michaelcahill:
Thanks @agorrod and @keithbostic, this lgtm.  Since the last change has already been tested, I'll merge.

PR #3285, opened 2017-02-09
WT-3149 Have eviction choose a random point when walking a tree.

agorrod:
Only choose a random point when there is no saved walk point.


src/btree/row_srch.c
```diff
@@ -785,7 +785,7 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
  *	Find a random leaf page in a row-store tree.
  */
 int
-__wt_row_random_descent(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
```
michaelcahill: I'm not seeing much here that is row-store specific.  If we change the `page->type` check to use `WT_PAGE_IS_INTERNAL(page)`, can we use this on any tree?

agorrod:
@keithbostic Could you take a look at this change and let me know what you think?

There are two things I don't like stylistically about this change:
1) The eviction server is aware of row storeness.
2) The extra flags to the search feel a bit clunky.

A more invasive change could fix that by adding a general random search function that returns ENOTSUP for non row-stores. I'm happy to go down that path if you think it'd be better?

I'm also a bit apprehensive about how long the search could take in a sparse tree, but can't think of a good way to check that (and it should be rare that we don't have an existing walk point.

src/evict/evict_lru.c
```diff
@@ -1662,6 +1662,21 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		FLD_SET(walk_flags, WT_READ_PREV);
```
agorrod: @keithbostic I think we probably also want to revisit ``evict_walk_reverse`` - I suspect we only want to set a walk direction when setting from a NULL evict_ref.

src/btree/row_srch.c
keithbostic: I agree with this, and I've pushed that change in 7666562.

I left the function itself in `row_srch.c` because I can't think of a better place for it.

keithbostic:
@agorrod 
> The eviction server is aware of row storeness.

Updated per @michaelcahill's suggestion.

> The extra flags to the search feel a bit clunky.

I assume you mean the `mem_only` flag to `__wt_row_random_descent()`. Yes, I agree it's a little clunky, but I can't think of anything better.

There is a retry in here (retry 100 times if we don't find a valid leaf page), that may not be right for eviction, depending on how often this function is called?

If there are other changes to this function to support eviction, we could create an eviction-specific version: it's only about 80 lines of code.

> A more invasive change could fix that by adding a general random search function that returns ENOTSUP for non row-stores.

I think that question is resolved with @michaelcahill's suggested change. The `WT_CURSOR.next_random` call stack doesn't call into this function for anything other than row-stores (it checks both when setting the cursor's `next_random` method, and again when the underlying `next_random` function is called).

> I'm also a bit apprehensive about how long the search could take in a sparse tree, but can't think of a good way to check that (and it should be rare that we don't have an existing walk point).

If this function is only called rarely, I don't think sparse trees will be an issue (including any questions about the retry loop in the code).

keithbostic:
lgtm.

src/evict/evict_lru.c
keithbostic: @agorrod 
> I think we probably also want to revisit ``evict_walk_reverse`` - I suspect we only want to set a walk direction when setting from a NULL evict_ref.

If by "revisit", you mean "remove", I think that's a good idea.

Frankly, the less reverse traversal of the tree we do, the happier I am.

keithbostic:
@agorrod 
A couple of fixes chasing the test suite failures: b9dd358 and ea9f063.

src/btree/row_srch.c
```diff
@@ -781,23 +781,28 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
...
 
 /*
- * __wt_row_random_descent --
- *	Find a random leaf page in a row-store tree.
+ * __wt_random_descent --
+ *	Find a random leaf page in a tree.
  */
 int
-__wt_row_random_descent(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
+__wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, bool eviction)
```
agorrod: My vote would be for this to be in a different location. Either a new file; maybe ``src/btree/btree_random.c`` or maybe into ``bt_cursor.c``

agorrod: In response to the other question - I'm inclined to keep the single function for now. The function isn't that long, but I can easily imaging bug fixes coming here, and I don't see a way to share the common code paths.

michaelcahill:
I'd like to talk more about the walk direction when walks start at a random point.  Can we have that discussion early next week?

keithbostic:
@agorrod, lgtm.

I agree with you & @michaelcahill, it would be good to move this somewhere else.

I don't want this code in `bt_cursor.c`, that already has too much stuff.

We could create `bt_random.c`, and put `__wt_btcur_next_random()`, `__wt_random_descent()` and `__wt_row_random_leaf()` there?

(If we do that, let's merge PR #3289 first, it changes `__wt_btcur_next_random()`.)

michaelcahill:
@keithbostic, agreed and done.

michaelcahill:
@agorrod, I resurrected the code to reverse the direction of eviction walks and make it reverse each time we start a new random walk.  Let's talk about whether we want to go this way: in my local testing of `evict-btree.wtperf`, I don't measure any performance impact.

michaelcahill:
autobuild: retest this please

agorrod:
Thanks @keithbostic and @michaelcahill these changes look good to me. I'm going to merge and watch the perf jobs over night.

PR #3286, opened 2017-02-09
WT-3123 Add check and stop functions.  Move loop into thread_group code.

sueloverso:
Not yet ready to merge.  I split out the restructuring changes from the active/pause changes for thread groups.  This branch is only the restructuring, adding chk and stop functions, etc.  I want to run the Jenkins perf tests on this branch to see how it performs.


sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

sueloverso:
@agorrod Can you take a pass over these changes with fresh eyes?  This is a split out of just the function call portion of the thread group changes overall.  That is, there should be no functional difference other than the eviction code for `wt_evict_thread_run` is a single pass and the loop is now in the thread group code in `thread_run`.  Perhaps something got unintentionally changed that clearly affects how eviction is run (see my discussion in the ticket about `update-checkpoint-btree.wtperf` and the Jenkins perf plots).

I suspect the cause of this is the overall cause of the differences seen when I did the full functional and active/paused thread group changes.

agorrod:
@sueloverso I've lost track of where we are up to with this change. Did the [1fd9e82](https://github.com/wiredtiger/wiredtiger/pull/3286/commits/15f9e82758eda74f112e0d92473aac14da37cb8e) change bring the performance back up?

sueloverso:
@agorrod this branch is now performing equal to develop.  However this branch only adds the function call structural changes by moving the loop calls into the thread group code and adding chk and stop functions.  It still stops/starts threads rather than pausing/activating them.  So this is half of the set of needed changes.

agorrod:
This is the final version and ready for review.

agorrod:
Thanks @sueloverso lgtm - I'll go ahead and merge.

PR #3288, opened 2017-02-10
WT-3170 Change how eviction walk point is saved during walk.

keithbostic:
Review split barriers.

@michaelcahill, here is my suggested change for the barriers in the split code.


michaelcahill:

PR #3289, opened 2017-02-10
WT-3144 bug fix: random cursor returns not-found when descending to an empty page

keithbostic:
If random descent through the tree fails, fallback to skipping through the tree's pages; if skipping through the tree's pages fails, fallback to a random entry from the first page in the tree that contains anything at all.

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -862,40 +862,26 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	WT_STAT_CONN_INCR(session, cursor_next);
 	WT_STAT_DATA_INCR(session, cursor_next);
 
+#ifdef HAVE_DIAGNOSTIC
 	/*
-	 * If retrieving random values without sampling, or we don't have a
-	 * page reference, pick a roughly random leaf page in the tree.
+	 * Under some conditions we end up using the underlying cursor.next to
+	 * walk through the object. Since there are multiple calls, we can hit
+	 * the cursor-order checks, turn them off.
+	 */
+	__wt_cursor_key_order_reset(cbt);
+#endif
+
+	/*
+	 * If we don't have a current position in the tree, or if retrieving
+	 * random values without sampling, pick a roughly random leaf page in
+	 * the tree and return an entry from it.
 	 */
 	if (cbt->ref == NULL || cbt->next_random_sample_size == 0) {
-		/*
-		 * Skip past the sample size of the leaf pages in the tree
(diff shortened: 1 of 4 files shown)


src/btree/bt_cursor.c
```diff
@@ -896,24 +896,53 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
...
 		    session, ret = __wt_row_random_descent(session, cbt));
+
+		/*
+		 * Random descent may return not-found: the tree might be empty
+		 * or have so many deleted items we didn't find any valid pages.
+		 * We can't return WT_NOTFOUND to the application unless a tree
+		 * is really empty, fallback to skipping through tree pages.
+		 */
+		if (ret == WT_NOTFOUND)
+			goto fallback_walk;
```
michaelcahill: As far as I can see, we won't have done the calculation of `cbt->next_random_leaf_skip`, so it looks like we'll always find the first record in the tree.  Am I missing something?

test/suite/test_cursor_random.py
```diff
@@ -71,6 +71,15 @@ def test_cursor_random(self):
...
 
+    # Check that next_random fails with an empty tree, repeatedly.
+    def test_cursor_random_empty(self):
+        uri = self.type
+        self.session.create(uri, 'key_format=S,value_format=S')
+        cursor = self.session.open_cursor(uri, None, self.config)
+        for i in range(1,5):
+            self.assertTrue(cursor.next(), wiredtiger.WT_NOTFOUND)
+        cursor.close
+
```
michaelcahill: It looks like we should have a test that creates a tree with enough data for multiple pages, reopens the connection so we have a real tree, then truncates most / all of the tree and makes sure random lookups find data / fail (respectively).  That way we're testing {{WT_REF_DELETED}}, not just empty pages.

src/btree/bt_cursor.c
keithbostic: > we won't have done the calculation of cbt->next_random_leaf_skip, so it looks like we'll always find the first record in the tree.

Agreed, I've done another run at this one, ready for your review.

test/suite/test_cursor_random.py
keithbostic: > It looks like we should have a test that creates a tree with enough data for multiple pages, reopens the connection so we have a real tree, then truncates most / all of the tree and makes sure random lookups find data / fail (respectively). That way we're testing WT_REF_DELETED, not just empty pages.

Added in fc30d0e.

michaelcahill:

PR #3291, opened 2017-02-13
WT-3152 Switch the table lock to a rwlock.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'd like to wait until the current test hangs are resolved prior to merging.

PR #3292, opened 2017-02-13
WT-3135 lint.

michaelcahill:
test/csuite/wt3135_search_near_collator/main.c:75:22: error: implicit conversion loses integer precision: 'int64_t' (aka 'long long') to 'int' [-Werror,-Wshorten-64-to-32]
                *cmp = compare_int(pkey1, pkey2);


keithbostic:
@michaelcahill, I included the WT-3135 (#3292) lint change in WT-3174 (#3293), so I'll close #3292 without merging.

PR #3294, opened 2017-02-14
WT-3175 Don't verify children during splits up the tree.

michaelcahill:
Reverts part of 7f5d0f9981214c723f2ed90cf4533887ed406176.

Fixes a deadlock in diagnostic mode.  Also revert a change that could
cause diagnostic code to read pages into cache: we don't want diagnostic
adding cache pressure and we already verify pages as they are evicted.


agorrod:

PR #3298, opened 2017-02-14
WT-3180 bug fix: disable long tests in the top-level main program, disable core files for subtests

ddanderson:
Also fixed some uint64 values/parameters that were declared as uint32.


sueloverso:
Thanks @ddanderson this LGTM.  Merge when ready.

PR #3299, opened 2017-02-14
WT-3176 Add compatibility API.

sueloverso:
Not yet ready for merge.


src/conn/conn_api.c
```diff
@@ -190,6 +190,42 @@ __wt_conn_remove_collator(WT_SESSION_IMPL *session)
...
+	} else
+		conn->compat_major = WIREDTIGER_VERSION_MAJOR;
+	WT_RET(__wt_config_gets(session, cfg,
+	    "compatibility.minor", &cval));
+	/*
+	 * We don't currently have knowledge of major/minor pairs.  If the
+	 * user gave us zero set it to the current version.  Check for some
+	 * out of bounds settings.
+	 */
+	conn->compat_minor = (uint16_t)cval.val;
```
agorrod: Would it make more sense to specify the version as a string than two separate integers? Then there is no ambiguity about which version is being specified.

sueloverso: We can discuss this.  I chose this path (and wrote it up in the scope) for a couple of reasons:
* There already exists `version=(major=,minor=)` in the `api_data.py` file.  This API followed that precedent.
* This API allows a user to not have to know the specific release necessarily.  If this change goes into a 3.0, then just `compat=(major=2)` is sufficient.
* Given those reasons above and that it is easier to deal with integer args instead of bursting apart a string led to this API.

However, these are not strong reasons.  The `version` string only appears in the meta-data file, not something visible to users.  How strongly do you feel about this?

agorrod: Per our phone conversation, I think it's worth exploring to see how the string version ends up, but I'm not wedded to the suggestion.

sueloverso:
@agorrod I pushed the user-facing change. I like how it turned out, it didn't clutter up the code at all.  One minor question:  do you anticipate other compatibility factors other than release?  Right now I have it as a subcategory that only has one option:  `compatibility=(release="2.9"))`.  It could be simplified to `compatibility="2.9"`.

sueloverso:
autobuild: retest this please

sueloverso:
FTR, our discussion concluded that we want to keep the subcategory even if it only has one entry for now.  We can contemplate other conditions such as `on_close` or some other way to defer this.

sueloverso:
@agorrod I believe this is ready to merge.  I have added it to test/format and have it running.  I want to let that run at least overnight without a failure (I debugged one hang today so it has found a bug or two).  Here are some things to note:

* This is based off the wt-3039-prevlsn branch, so merging this doesn't impose it on develop yet.
* I chose to bump the release to 3.0 due to the file format change.  If you disagree with that then I have to add code back in for minor version checking.
* The test/format changes use `v10` and `v11` for the log version numbers.  I'm not thrilled with it so open to better suggestions.  Internally I use "2.6" and "3.0" as a known old/new release number.
* There was some code movement in `log.c` that makes the diff very large.
* I manually tested a downgrade followed by running printlog from an "old" tree to make sure the old tree could read the log files.  I do not have a regularly run test to continue that automatically.  Is there any testing infrastructure like that?  I didn't see any in the test suite.

michaelcahill:
@sueloverso, I didn't get to this today -- either @agorrod  or I will work through it tomorrow.

sueloverso:
FYI, the test/format run is still going strong from overnight up to almost 500 iterations.

src/include/log.h
```diff
@@ -261,8 +265,9 @@ struct __wt_log {
 	uint64_t	 write_calls;		/* Calls to log_write */
 #endif
 
-#define	WT_LOG_OPENED		0x01	/* Log subsystem successfully open */
-#define	WT_LOG_TRUNCATE_NOTSUP	0x02	/* File system truncate not supported */
+#define	WT_LOG_FORCE_NEWFILE	0x01	/* Log subsystem successfully open */
```
agorrod: Copy-pasted description for the new flag

src/log/log.c
```diff
@@ -1005,6 +1121,91 @@ __log_newfile(WT_SESSION_IMPL *session, bool conn_open, bool *created)
...
+	WT_WITH_SLOT_LOCK(session, log,
+	    ret = __log_set_version(session,
+	    major, minor, first_rec, downgrade));
+	if (!live_chg)
+		return (ret);
+	WT_ERR(ret);
+	/*
+	 * XXX - Any gap problem between setting version above and this flag
+	 * now? Also may want to clear it in log_newfile when it sees it so that
+	 * additional log records don't also cause new log files repeatedly.
```
agorrod: I can't see a problem with a race here - the `force_write` call follows setting the flag, and switching doesn't clear the flag.

src/include/log.h
sueloverso: Fixed

src/log/log.c
sueloverso: I decided to move this in to the function that sets up other fields under lock and clear it once we create the new file so that we don't get multiple file switches potentially.

agorrod:
Thanks @sueloverso - these changes look good to me now. I'm going to merge into the WT-3176 branch, and we can do a full review of all changes before merging into develop.

PR #3301, opened 2017-02-14
WT-3179 test bug: clang sanitizer failure in fail_fs

ddanderson:
@keithbostic noticed these issues:
* there’s a fprintf to stderr in fail_fs.c.
* The fail_file_truncate() function calls `ftruncate`, and I don't think `ftruncate` is always thread-safe, does that matter here?
* There's an access(2) call that uses and argument of 0, instead of F_OK.

This commit has fixes for first and last.  I don't know the answer for ftruncate, but I see that the truncate function in the default posix file system (see `os_posix/os_fs.c`) also calls ftruncate here. If it matters here, it will matter there as well.

### DIFF
--- a/ext/test/fail_fs/fail_fs.c
+++ b/ext/test/fail_fs/fail_fs.c
@@ -536,7 +536,7 @@ fail_fs_exist(WT_FILE_SYSTEM *file_system,
 	(void)file_system;					/* Unused */
 	(void)session;						/* Unused */
 
-	*existp = (access(name, 0) == 0);
+	*existp = (access(name, F_OK) == 0);
 	return (0);
 }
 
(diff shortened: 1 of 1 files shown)


ddanderson:
@keithbostic, here's the clean up for the issues you brought up.  Can you review/merge please?

keithbostic:
> I don't know the answer for ftruncate, but I see that the truncate function in the default posix file system (see os_posix/os_fs.c) also calls ftruncate here. If it matters here, it will matter there as well.

I agree this is OK -- we document "WT_FILE_HANDLE::fh_truncate is not called by multiple threads concurrently (on the same file handle)."

The change looks good to me, I'll merge.

PR #3302, opened 2017-02-15
WT-3149 Make random lookups for eviction more lightweight.

michaelcahill:
Eviction walks don't need to start on leaf pages: just try to descend
through the tree and as soon as we can't swap to a child page, start
the walk from the parent.

### DIFF
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -165,11 +165,11 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
 }
 
 /*
- * __random_leaf --
- *	Find a random leaf page in a tree.
+ * __wt_random_descent --
+ *	Find a random page in a tree for either sampling or eviction.
  */
-static int
-__random_leaf(WT_SESSION_IMPL *session, WT_REF **refp)
+int
+__wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, bool eviction)
 {
 	WT_BTREE *btree;
 	WT_DECL_RET;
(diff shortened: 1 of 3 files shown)


agorrod:
This branch gets back most of the performance degradation for me, I still see about a 5% drop, but it's within the standard run-to-run variance now.

The reason I decided not to look for an internal page in the first place, was because I thought avoiding "deserts" in the file, where we walk a lot of out of cache refs may outweigh the cost of a more expensive random walk. I guess this workload indicates that isn't the case, though I'd be curious to see whether larger trees see the same cost/benefit. It'd be nice if we could get this "find a good eviction walk start point" out of the eviction server fast path somehow (potential future work).

It's also not clear to me why this workload is abandoning it's walk point at all, my intuition is that it shouldn't. There are about 1000 dirty pages, eviction is doing enough scrubbing that eviction is also looking for clean candidates, there is enough cache pressure that the targets aren't being reached. I'm going to take a bit of a look at those things, but it's more out of curiosity than needing more fixes than are in this branch

michaelcahill:
@agorrod, it sounds like some of the new stats in WT-3138 might help to answer some of those questions?  In particular, knowing whether walks "ended" or "gave up" vs the current "abandoned" stat might clarify what's going?

agorrod:
Thanks @michaelcahill FWIW: This workload is giving up the walk because it's finding "deserts" in the middle of the file, where there aren't enough candidates being queued vs seen. Most of the pages seen are clean, and eviction is only looking for dirty candidates.

I'm going to merge this one, but there is definitely space to improve how the eviction server is finding pages for this workload.

PR #3303, opened 2017-02-15
WT-3184 bug fix: duplication needs special case searching an index that has a custom collator.

ddanderson:
In this case, we must use the entire (raw) key to duplicate the position,
instead of truncating to the visible part.

### DIFF
--- a/src/cursor/cur_index.c
+++ b/src/cursor/cur_index.c
@@ -245,7 +245,8 @@ __curindex_search(WT_CURSOR *cursor)
 	 * Custom collators expect to see complete keys, pass an item containing
 	 * all the visible fields so it unpacks correctly.
 	 */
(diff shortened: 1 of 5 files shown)


ddanderson:
@michaelcahill, this is ready for review.  For a little more context, see the JIRA ticket: https://jira.mongodb.org/browse/WT-3184

michaelcahill:
@ddanderson, thanks, I see what's going on here.  I didn't think about this case when WT-3135 went in but you're correct that it has to be handled specially.

I think this change is fine and I will merge.  If we find in future that there are more special cases when duplicating cursors we should consider a method `new_cursor->duplicate(new_cursor, old_cursor)` that makes `new_cursor` point to the same position as `old_cursor` to replace the shared `__wt_cursor_dup_position`.

PR #3304, opened 2017-02-15
WT-3186 Fix error path and panic detection in logging loops.

sueloverso:
@agorrod Here is a fix for one path I found looking at the stack from the failure.  I have not been able to reproduce the hang even with some simple error injection directly but based on the stack and errors, this is definitely a problem path.  This fixes two loops that didn't break and weren't checking for panic.


agorrod:
Thanks @sueloverso lgtm

PR #3305, opened 2017-02-16
WT-3149 Use a range of eviction walk start points.

agorrod:
Choosing a random point isn't very efficient in append only workloads.


michaelcahill:
Thanks @agorrod, I'm happy for this change to be merged when we have confidence it fixes the performance regression.

PR #3306, opened 2017-02-16
WT-3187 Ramp up aggressiveness in reducing cache pool usage

agorrod:
We could get into situations where no participants looked like good candidates.
Also put a failsafe into the balance loop, to ensure future failures to reduce usage won't lead to hang on shutdown.


agorrod:
@sueloverso Would you mind reviewing this change please?

sueloverso:
LGTM.  I made a couple minor wording changes to comments.  Overall I'm not a fan of large comments in between if-else clauses, but the comments are needed in this file and there is precedent elsewhere in this file with that.  I'll merge once testing completes.

PR #3307, opened 2017-02-17
WT-3188 More log loops needing to check panic.

sueloverso:
@agorrod Here are a few more panic checking changes.  I had another similar hang today.  I went through all of the logging source files looking at loops where we're waiting for something and added a check rather than fault them in.


agorrod:
Thanks @sueloverso lgtm

PR #3308, opened 2017-02-19
WT-3189 Fix a segfault in eviction random page search.

agorrod:
A NULL page could be encountered when traversing a tree that is
being used by exclusive access.


agorrod:

michaelcahill:
Thanks @agorrod, I've edited the comment and this LGTM.  I'll merge.

PR #3309, opened 2017-02-20
WT-3149 Fix a compiler warning on OS X.

agorrod:
src/evict/evict_lru.c:1683:2: error: default label in switch which covers all enumeration values [-Werror,-Wcovered-switch-default]


michaelcahill:
autobuild: retest this please

michaelcahill:
@agorrod, I've confirmed that this fixes the warning locally.  I'll merge when tests pass.

PR #3310, opened 2017-02-20
WT-3191 lint

keithbostic:
Alex, this one needs your review because of dfee9c8. I think your original code is legal, but requires multiple casts to/from "ints" because of the default types for arithmetic operations. It was making my head hurt trying to figure it out, so I turned it into a switch statement.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1720,10 +1720,22 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 			 * Try a different walk start point next time if a
 			 * walk gave up.
 			 */
-			btree->evict_walk_state =
-			    (btree->evict_walk_state + 1) %
-			    WT_EVICT_WALK_MAX_LEGAL_VALUE;
-			break;
+			switch (btree->evict_walk_state) {
+			case WT_EVICT_WALK_NEXT:
+				btree->evict_walk_state = WT_EVICT_WALK_PREV;
+				break;
+			case WT_EVICT_WALK_PREV:
+				btree->evict_walk_state =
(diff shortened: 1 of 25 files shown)


src/evict/evict_lru.c
```diff
@@ -1720,10 +1720,22 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 			 * Try a different walk start point next time if a
 			 * walk gave up.
 			 */
-			btree->evict_walk_state =
```
agorrod: @keithbostic I preferred the original version of this code, since it meant we didn't need to update another section of code when adding a new walk state - it'd be easy to forget to update the new case statement, and it's not an obvious bug. How horrible are the required casts?

keithbostic: > since it meant we didn't need to update another section of code when adding a new walk state

Yes, but it also means the enum's *order of declaration* is part of the switching algorithm, which is just weird. :)

It's your call. If you hate the change, revert it and I'll either add in the casts or ignore the complaints.

agorrod: @michaelcahill Do you mind being the arbiter? I don't feel super strongly - it's possible that we'll end up choosing walk start point strategies in a more sophisticated fashion in the future which would make this moot.

michaelcahill: I pushed something that keeps the arithmetic with minimal casting.

@keithbostic, is that approach okay with you?

michaelcahill:
@keithbostic, I'm happy for this to be merged whenever you're ready.

keithbostic:
lgtm, merging.

PR #3311, opened 2017-02-20
WT-3193 Close a race between verify and the eviction server

keithbostic:
@agorrod, can we return not-found on this failure instead of success?

### DIFF
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -37,7 +37,7 @@ __btree_initialize(WT_BTREE *btree, bool closing)
 		 * Closing: clear everything except cache/eviction information
 		 * and one LSM flag.
 		 */
-		memset(btree, 0, WT_BTREE_CLEAR_SIZE(btree));
+		memset(btree, 0, WT_BTREE_CLEAR_SIZE);
 		F_CLR(btree, ~(WT_BTREE_LSM_PRIMARY | WT_BTREE_NO_EVICTION));
 	} else {
 		/* Opening: clear everything except the special flags. */
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -15,6 +15,39 @@ static int __btree_preload(WT_SESSION_IMPL *);
 static int __btree_tree_open_empty(WT_SESSION_IMPL *, bool);
 
 /*
+ * __btree_initialize --
+ *	Initialize the WT_BTREE structure.
+ */
+static void
+__btree_initialize(WT_BTREE *btree, bool closing)
+{
+	uint32_t mask;
+
+	/*
+	 * This function exists as a place to talk about how the WT_BTREE
+	 * structure is initialized (or re-initialized, when the object is
+	 * re-opened). The upper-level handle code currently sets/clears
+	 * information in the WT_BTREE structure, specifically flags, cache
+	 * and eviction information. All of that information must persist
+	 * after the WT_BTREE object is closed, and some of it has already
+	 * been initialized when the object is opened/re-opened.
+	 */
+	if (closing) {
+		/*
+		 * Closing: clear everything except cache/eviction information
+		 * and one LSM flag.
+		 */
+		memset(btree, 0, WT_BTREE_CLEAR_SIZE(btree));
+		F_CLR(btree, ~(WT_BTREE_LSM_PRIMARY | WT_BTREE_NO_EVICTION));
+	} else {
+		/* Opening: clear everything except the special flags. */
+		mask = F_MASK(btree, WT_BTREE_SPECIAL_FLAGS);
+		memset(btree, 0, sizeof(*btree));
+		btree->flags = mask;
+	}
+}
+
+/*
  * __wt_btree_open --
  *	Open a Btree.
  */
@@ -33,7 +66,10 @@ __wt_btree_open(WT_SESSION_IMPL *session, const char *op_cfg[])
 	bool creation, forced_salvage, readonly;
 
 	dhandle = session->dhandle;
+
 	btree = S2BT(session);
+	__btree_initialize(btree, false);
+	btree->dhandle = session->dhandle;
 
 	/* Checkpoint files are readonly. */
 	readonly = dhandle->checkpoint != NULL ||
@@ -126,6 +162,20 @@ __wt_btree_open(WT_SESSION_IMPL *session, const char *op_cfg[])
 		}
 	}
 
+	/*
+	 * Eviction ignores trees until the handle's open flag is set, configure
+	 * eviction before that happens.
+	 *
+	 * Files that can still be bulk-loaded cannot be evicted.
+	 * Permanently cache-resident files can never be evicted.
+	 * Special operations don't enable eviction. (The underlying commands
+	 * may turn on eviction, but it's their decision.)
+	 */
+	if (btree->bulk_load_ok ||
+	    F_ISSET(btree, WT_BTREE_IN_MEMORY | WT_BTREE_REBALANCE |
+	    WT_BTREE_SALVAGE | WT_BTREE_UPGRADE | WT_BTREE_VERIFY))
+		WT_ERR(__wt_evict_file_exclusive_on(session));
+
 	if (0) {
 err:		WT_TRET(__wt_btree_close(session));
 	}
@@ -155,12 +205,14 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 
 		/* Close the underlying block manager reference. */
 		WT_TRET(bm->close(bm, session));
-
-		btree->bm = NULL;
 	}
 
 	/* Close the Huffman tree. */
 	__wt_btree_huffman_close(session);
+
+	if (btree->collator_owned && btree->collator->terminate != NULL)
+		WT_TRET(btree->collator->terminate(
+		    btree->collator, &session->iface));
 
 	/* Destroy locks. */
 	__wt_rwlock_destroy(session, &btree->ovfl_lock);
@@ -170,18 +222,7 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 	__wt_free(session, btree->key_format);
 	__wt_free(session, btree->value_format);
 
-	if (btree->collator_owned) {
-		if (btree->collator->terminate != NULL)
-			WT_TRET(btree->collator->terminate(
-			    btree->collator, &session->iface));
-		btree->collator_owned = 0;
-	}
-	btree->collator = NULL;
-	btree->kencryptor = NULL;
-
-	btree->bulk_load_ok = false;
-
-	F_CLR(btree, WT_BTREE_SPECIAL_FLAGS);
+	__btree_initialize(btree, true);
 
 	return (ret);
 }
@@ -267,9 +308,9 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
 
 	WT_RET(__wt_config_gets(session, cfg, "cache_resident", &cval));
 	if (cval.val)
-		F_SET(btree, WT_BTREE_IN_MEMORY | WT_BTREE_NO_EVICTION);
+		F_SET(btree, WT_BTREE_IN_MEMORY);
 	else
-		F_CLR(btree, WT_BTREE_IN_MEMORY | WT_BTREE_NO_EVICTION);
+		F_CLR(btree, WT_BTREE_IN_MEMORY);
 
 	WT_RET(__wt_config_gets(session,
 	    cfg, "ignore_in_memory_cache_size", &cval));
@@ -482,13 +523,10 @@ __btree_tree_open_empty(WT_SESSION_IMPL *session, bool creation)
 	/*
 	 * Newly created objects can be used for cursor inserts or for bulk
 	 * loads; set a flag that's cleared when a row is inserted into the
-	 * tree. Objects being bulk-loaded cannot be evicted, we set it
-	 * globally, there's no point in searching empty trees for eviction.
+	 * tree.
 	 */
-	if (creation) {
+	if (creation)
 		btree->bulk_load_ok = true;
-		__wt_btree_evictable(session, false);
-	}
 
 	/*
 	 * A note about empty trees: the initial tree is a single root page.
@@ -578,27 +616,6 @@ __wt_btree_new_leaf_page(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 		break;
 	}
 	return (0);
-}
-
-/*
- * __wt_btree_evictable --
- *      Setup or release a cache-resident tree.
- */
-void
-__wt_btree_evictable(WT_SESSION_IMPL *session, bool on)
-{
-	WT_BTREE *btree;
-
-	btree = S2BT(session);
-
-	/* Permanently cache-resident files can never be evicted. */
-	if (F_ISSET(btree, WT_BTREE_IN_MEMORY))
-		return;
-
-	if (on)
-		F_CLR(btree, WT_BTREE_NO_EVICTION);
-	else
-		F_SET(btree, WT_BTREE_NO_EVICTION);
 }
 
 /*
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -24,13 +24,14 @@ __btree_initialize(WT_BTREE *btree, bool closing)
 	uint32_t mask;
 
 	/*
-	 * This function exists as a place to talk about how the WT_BTREE
-	 * structure is initialized (or re-initialized, when the object is
-	 * re-opened). The upper-level handle code currently sets/clears
-	 * information in the WT_BTREE structure, specifically flags, cache
-	 * and eviction information. All of that information must persist
-	 * after the WT_BTREE object is closed, and some of it has already
-	 * been initialized when the object is opened/re-opened.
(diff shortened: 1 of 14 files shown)


michaelcahill:
@keithbostic, I don't know if that's safe or not.  It might be okay if we are sure `ref->page == NULL` only ever happens with the root page.  Otherwise we can leak hazard pointers.

Regardless, `WT_NOTFOUND` is ignored by the calling code so this won't have any effect.

keithbostic:
>  I don't know if that's safe or not. It might be okay if we are sure ref->page == NULL only ever happens with the root page. Otherwise we can leak hazard pointers.

I think that this can only happen with the root page; that said, I think there's a serious problem here we're not dealing with, and this change isn't sufficient. Maybe I'm just missing something, let's talk it through.

michaelcahill:
@keithbostic, needless to say, neither @agorrod or I made it to the change we discussed today (having `verify` use `__wt_evict_file_exclusive_on`/`off` to set `btree->root.page`).  If you get a chance, please take a look to see if that is a reasonable way to avoid checking for a `NULL` page.

keithbostic:
@agorrod, @michaelcahill: changes for your review.

I left the check for the root page in the walk code, there are paths where we're "discarding handles from the cache" and we've never read in a root page. Rather than check that in the callers, I left the check in the walk code and added a comment explaining it.

src/btree/bt_vrfy.c
```diff
@@ -215,38 +215,55 @@ __wt_verify(WT_SESSION_IMPL *session, const char *cfg[])
...
-			WT_TRET(__wt_cache_op(session, WT_SYNC_DISCARD));
+			/* Discard the checkpoint from the cache. */
+			WT_ERR(__wt_cache_op(session, WT_SYNC_DISCARD));
 		}
 
 		/* Unload the checkpoint. */
-		WT_TRET(bm->checkpoint_unload(bm, session));
-		WT_ERR(ret);
+		bm_unload = false;
+		WT_ERR(bm->checkpoint_unload(bm, session));
```
agorrod: Do we need to do the ``__wt_evict_file_exclusive_on`` call before unloading the checkpoint? Eviction walks the ``dhqh`` list to find handles, so this handle is visitable at any time. Or is this only a problem if the eviction server sees a handle that is partially populated via ``__wt_btree_tree_open`` - if it comes in between now, and where you added the ``exclusive_on``, it'll see a NULL root page and do the right thing?

keithbostic:
> Do we need to do the __wt_evict_file_exclusive_on call before unloading the checkpoint? Eviction walks the dhqh list to find handles, so this handle is visitable at any time. Or is this only a problem if the eviction server sees a handle that is partially populated via __wt_btree_tree_open - if it comes in between now, and where you added the exclusive_on, it'll see a NULL root page and do the right thing?

Good catch -- I think you're correct.

When verify finishes with a checkpoint, it calls `__wt_cache_op(session, WT_SYNC_DISCARD)`, which clears `WT_BTREE.root.page`. Eviction won't ignore the handle: if eviction tries to do a walk the NULL root page check in the walk code will handle the problem, but there's no matching NULL root page check in the `__wt_random_descent()` code (and I think we've agreed we don't want to add one there).

Thinking about this more, there's a fundamental problem in turning on eviction (by setting `WT_DHANDLE_OPEN`), before the underlying verify (or salvage, or rebalance, or upgrade), has a  chance to turn off eviction. I think the solution is probably to turn off eviction before setting `WT_DHANDLE_OPEN`, but unfortunately, that's tricky.

keithbostic:
@agorrod, @michaelcahill, here's a set of changes to block eviction for the entire life of a "special" command btree structure. This change is more invasive than the alternative, but I think it's cleaner.

There's a change in here that's not necessary, a change to be aggressive clearing the WT_BTREE structure when it's opened/closed: I made that change while trying to figure out what parts of the WT_BTREE structure needed to persist after closing that structure, and left it in -- I think it may break testing in the short-term, but I think it's worth it to get clarity on what in the WT_BTREE structure is persisted after it's closed.

The simplest alternative to this change is one Alex & I came up with:
```
*** evict/evict_lru.c.orig        Wed Feb 22 15:25:35 2017
--- evict/evict_lru.c.   Thu Feb 23 09:31:01 2017
***************
*** 1430,1436 ****
                 */
                if (!F_ISSET(btree, WT_BTREE_NO_EVICTION) &&
                    !__wt_spin_trylock(session, &cache->evict_walk_lock)) {
!                       if (!F_ISSET(btree, WT_BTREE_NO_EVICTION)) {
                                cache->evict_file_next = dhandle;
                                WT_WITH_DHANDLE(session, dhandle, ret =
                                    __evict_walk_file(session, queue,
--- 1430,1437 ----
                 */
                if (!F_ISSET(btree, WT_BTREE_NO_EVICTION) &&
                    !__wt_spin_trylock(session, &cache->evict_walk_lock)) {
!                       if (btree->root.page != NULL &&
!                           !F_ISSET(btree, WT_BTREE_NO_EVICTION)) {
                                cache->evict_file_next = dhandle;
                                WT_WITH_DHANDLE(session, dhandle, ret =
                                    __evict_walk_file(session, queue,
```
Alex and I think this change works as long as no code changes `WT_BTREE.root.page` without first blocking eviction. I think that's currently true except for the `WT_SESSION.rebalance` command, which is safe. My vote is for the larger change, because I think this change is more fragile in the long run.

Michael, I took a look at changing the upper-level handle code to not set `WT_DHANDLE_OPEN` in the special-command open path, and that looks a lot messier to me (it involves changing the sweep code, for example, and more special-case flag checking in the handle code). Let's talk that over, I may be missing where you were going with that.

If we go with this change, I think we should also remove the calls `__wt_evict_file` makes to `__wt_evict_file_exclusive_on/__wt_evict_file_exclusive_off`, because that function evicts the root page, setting the `WT_BTREE.root.page` pointer to NULL, which means callers of `__wt_evict_file` should be preventing eviction access to the handle after `__wt_evict_file` returns in some way, either by calling `__wt_evict_file_exclusive_on` separately, or by clearing `WT_DHANDLE_OPEN`. (We should assert in `__wt_evict_file` that eviction is prevented from accessing the handle, of course.)

keithbostic:
> We should assert in __wt_evict_file that eviction is prevented from accessing the handle, of course.

I made this change in 82dd36f, plus a minor fix in `__wt_session_lock_checkpoint`, and it seems to run.

src/btree/bt_handle.c
```diff
@@ -15,6 +15,39 @@ static int __btree_preload(WT_SESSION_IMPL *);
...
+	 * after the WT_BTREE object is closed, and some of it has already
+	 * been initialized when the object is opened/re-opened.
+	 */
+	if (closing) {
+		/*
+		 * Closing: clear everything except cache/eviction information
+		 * and one LSM flag.
+		 */
+		memset(btree, 0, WT_BTREE_CLEAR_SIZE(btree));
+		F_CLR(btree, ~(WT_BTREE_LSM_PRIMARY | WT_BTREE_NO_EVICTION));
```
michaelcahill: I think the only way the `WT_BTREE_LSM_PRIMARY` flag can be set here is during `WT_CONNECTION::close` -- I don't think there's any reason to preserve that flag after a btree handle is closed.

Its purpose is for accounting, it could be called something like `WT_BTREE_IGNORE_DIRTY_SIZE` to clarify the intent.

src/btree/bt_handle.c
```diff
@@ -581,27 +619,6 @@ __wt_btree_new_leaf_page(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 }
 
 /*
- * __wt_btree_evictable --
- *      Setup or release a cache-resident tree.
- */
-void
-__wt_btree_evictable(WT_SESSION_IMPL *session, bool on)
```
michaelcahill: Happy to see this go.  The overlap with `evict_exclusive_on/off` was confusing.

src/btree/bt_walk.c
```diff
@@ -348,16 +348,24 @@ __tree_walk_internal(WT_SESSION_IMPL *session,
...
+		/*
+		 * If the root page is NULL, treat it as an empty tree. This is
+		 * not meant to avoid races when accessing a WT_REF, any thread
+		 * setting the root page must have exclusive access to the root
+		 * WT_REF (that is, must have locked out eviction). This is only
+		 * to avoid checking in every caller if the tree has been read
+		 * into memory.
+		 */
+		if (btree->root.page == NULL)
 			goto done;
```
michaelcahill: This leaves me confused about the point of this change.  I think the random descent code should match walk, because we do a random descent just before calling walk in eviction.

Why can't a random descent see a tree before it has been read into memory, but walk can?  Is this for some other case, like `__wt_evict_file`?

src/include/btree.h
```diff
@@ -130,6 +126,30 @@ struct __wt_btree {
...
+	bool modified;			/* If the tree ever modified */
+	bool bulk_load_ok;		/* Bulk-load is a possibility */
+
+	/*
+	 * The tree's cache and eviction information persist past after the
+	 * handle is closed (clean cache pages may remain after the tree is
+	 * closed). Be careful clearing the WT_BTREE structure.
+	 */
+#define	WT_BTREE_CLEAR_SIZE(b)						\
+	(WT_PTRDIFF(&(b)->root, b))
```
michaelcahill: I think it's clearer to use:

`#define	WT_BTREE_CLEAR_SIZE offsetof(WT_BTREE, root)`

I note that `WT_SESSION_CLEAR_SIZE` uses the same kind of calculation.

The bigger issue is why do we need to keep the fields below on a closed tree (particularly since they will be cleared on next open)?  Ideally we could free `dhandle->handle` in that case -- can't we just assert that eviction isn't active in a closed tree?

src/btree/bt_vrfy.c
```diff
@@ -216,37 +216,59 @@ __wt_verify(WT_SESSION_IMPL *session, const char *cfg[])
...
+		 */
 		WT_ERR(ret);
 
 		/* Display the tree shape. */
 		if (vs->dump_layout)
 			WT_ERR(__verify_layout(session, vs));
 	}
 
-done:
-err:	/* Inform the underlying block manager we're done. */
```
michaelcahill: Just curious, why reorder?  I'd normally go for alpha order in the absence of some compelling reason...

keithbostic: An unintentional change, I've reverted it.

src/btree/bt_handle.c
keithbostic: Failing to preserve the flag makes `test_cursor06` fail:
```
test_cursor06.test_cursor06.test_reconfigure_overwrite(lsm-S): ,
file:reconfigure-000002.lsm, WT_CONNECTION.close:
WT_BTREE.bytes_dirty_leaf underflow: decrementing 11608
```
The problem is drop closes the btree handle, but the bytes aren't discarded until later. Here's the stack from drop closing the handle:
```
#0  __btree_initialize (btree=0x807440380, closing=true)
    at src/btree/bt_handle.c:43
#1  0x0000000803aa83ed in __wt_btree_close (session=0x8071618a0)
    at src/btree/bt_handle.c:227
#2  0x0000000803afe698 in __wt_conn_btree_sync_and_close (session=0x8071618a0, 
    final=false, force=true) at src/conn/conn_dhandle.c:194
#3  0x0000000803aff7e4 in __wt_conn_dhandle_close_all (session=0x8071618a0, 
    uri=0x8074230a0 "file:reconfigure_cgroup1-000002.lsm", force=true)
    at src/conn/conn_dhandle.c:508
#4  0x0000000803ba86c9 in __drop_file (session=0x8071618a0, 
    uri=0x8074230a0 "file:reconfigure_cgroup1-000002.lsm", force=true, 
    cfg=0x7fffffffcb80) at src/schema/schema_drop.c:33
#5  0x0000000803ba8148 in __wt_schema_drop (session=0x8071618a0, 
    uri=0x8074230a0 "file:reconfigure_cgroup1-000002.lsm", cfg=0x7fffffffcb80)
    at src/schema/schema_drop.c:181
#6  0x0000000803b6bf9a in __wt_lsm_tree_drop (session=0x8071618a0, 
    name=0x8070e5e20 "lsm:reconfigure_cgroup1", cfg=0x7fffffffcb80)
    at src/lsm/lsm_tree.c:919
#7  0x0000000803ba81ee in __wt_schema_drop (session=0x8071618a0, 
    uri=0x8070e5e20 "lsm:reconfigure_cgroup1", cfg=0x7fffffffcb80)
    at src/schema/schema_drop.c:185
#8  0x0000000803ba8a84 in __drop_table (session=0x8071618a0, 
    uri=0x806e2686c "table:reconfigure", cfg=0x7fffffffcb80)
    at src/schema/schema_drop.c:129
#9  0x0000000803ba823d in __wt_schema_drop (session=0x8071618a0, 
    uri=0x806e2686c "table:reconfigure", cfg=0x7fffffffcb80)
    at src/schema/schema_drop.c:187
#10 0x0000000803bbc339 in __session_drop (wt_session=0x8071618a0, 
    uri=0x806e2686c "table:reconfigure", config=0x806dc6984 "force")
    at src/session/session_api.c:856
```
and then we eventually close the connection:
```
#6  0x0000000803aa6027 in __wt_cache_page_byte_dirty_decr (
    session=0x807027000, page=0x80780e1a0, size=11608) at btree.i:295
#7  0x0000000803aa5de2 in __wt_cache_dirty_decr (session=0x807027000, 
    page=0x80780e1a0) at btree.i:388
#8  0x0000000803aa5032 in __wt_page_modify_clear (session=0x807027000, 
    page=0x80780e1a0) at btree.i:594
#9  0x0000000803aa4ca0 in __wt_page_out (session=0x807027000, 
    pagep=0x808011100) at src/btree/bt_discard.c:80
#10 0x0000000803aa4c43 in __wt_ref_out (session=0x807027000, ref=0x808011100)
    at src/btree/bt_discard.c:59
#11 0x0000000803b3fa88 in __wt_evict_file (session=0x807027000, 
    syncop=WT_SYNC_DISCARD) at src/evict/evict_file.c:91
#12 0x0000000803ad1868 in __wt_cache_op (session=0x807027000, 
    op=WT_SYNC_DISCARD) at src/btree/bt_sync.c:319
#13 0x0000000803bdcf87 in __wt_checkpoint_close (session=0x807027000, 
    final=true) at src/txn/txn_ckpt.c:1641
#14 0x0000000803afee8a in __wt_conn_btree_sync_and_close (session=0x807027000, 
    final=true, force=false) at src/conn/conn_dhandle.c:191
#15 0x0000000803b00142 in __wt_conn_dhandle_discard_single (
    session=0x807027000, final=true, force=false)
    at src/conn/conn_dhandle.c:575
#16 0x0000000803b00721 in __wt_conn_dhandle_discard (session=0x807027000)
    at src/conn/conn_dhandle.c:642
#17 0x0000000803b053cd in __wt_connection_close (conn=0x8023dc000)
    at src/conn/conn_open.c:131
#18 0x0000000803af3d6d in __conn_close (wt_conn=0x8023dc000, config=0x0)
    at src/conn/conn_api.c:1089
```
and `__wt_cache_page_byte_dirty_decr` is looking for that flag in the btree handle.

src/btree/bt_walk.c
keithbostic: > Why can't a random descent see a tree before it has been read into memory, but walk can? Is this for some other case, like `__wt_evict_file`?

Exactly, it's for `__wt_evict_file`, we "evict" handles that never had backing pages all the time.

I almost made the change in 9690941 before, but chickened out...

That commit removes the code to handle empty trees in the walk code, and asserts in both the walk code and the eviction code that we never see an empty tree in either place.

src/include/btree.h
keithbostic: > I think it's clearer to use:
> `#define	WT_BTREE_CLEAR_SIZE offsetof(WT_BTREE, root)`

Agreed, pushed in 8673a4f.

keithbostic: > The bigger issue is why do we need to keep the fields below on a closed tree (particularly since they will be cleared on next open)?  Ideally we could free `dhandle->handle` in that case -- can't we just assert that eviction isn't active in a closed tree?

There's some code in `__wt_conn_btree_sync_and_close` that I don't entirely understand, but it's possible to get through there without discarding the contents of the tree, and with `WT_DHANDLE_DEAD` set on the handle, which I think implies we've closed the tree but eviction will still be active in the tree.

We could maybe fix this by not discarding the `WT_BTREE` structure until we're discarding the contents of the cache, but the obvious changes to do that immediately break with:
```
lsm-worker: wt-000003.lsm: file-remove: file has open handles.
```

We could also move the eviction information from the `WT_BTREE` structure into the `WT_DATA_HANDLE` structure (or allocate an eviction structure on demand if that makes the data-handle too large).

michaelcahill:
@keithbostic, lgtm, I'm happy for this to be merged whenever you're ready.

We can revisit the issue of "dead" trees at some point -- that happens in between a forced drop and sweep successfully discarding the tree.  In other words, the tree is still "open" and has pages in cache, but it's no longer part of the namespace.  That may fall out as soon as we make tables or LSM trees into first class data handles.

keithbostic:
@michaelcahill 
> We can revisit the issue of "dead" trees at some point

Agreed -- this isn't yet correct, but it's significantly better than it was, and we can fix it incrementally.

I'll merge.

PR #3312, opened 2017-02-21
WT-3192 Truncate obsolete updates when all updates are obsolete

keithbostic:


michaelcahill:
I don't think this is safe: I'll follow up on the ticket.

keithbostic:
> I don't think this is safe: I'll follow up on the ticket.

Thanks -- I had a vague recollection of proposing this a long time ago and agreeing it *wasn't* safe, but I couldn't find the email or remember the reason.

PR #3313, opened 2017-02-27
WT-3182 Switch make-check to run the short test suite by default

keithbostic:
@agorrod, here's a first cut at this one.

I think this is all that's necessary in the source code, but maybe I'm missing the big picture!

Let me know if you agree -- we can merge, which will just turn off long tests everywhere.

We can talk over what test harnesses should enable long tests and I'll make/test those changes.


agorrod:
Thanks @keithbostic - this lgtm, I'll merge and sort out the automated testing side of things.

PR #3314, opened 2017-02-28
WT-3182 Update Evergreen configuration to include long make check

agorrod:


agorrod:
I'm going to self review, since this is a tested change to Evergreen.

PR #3316, opened 2017-03-01
WT-3193 Close a race between verify opening a handle and eviction visiting it

keithbostic:
@michaelcahill, this change seems to work for me, with light testing.

Two notes:

* we could wait to free memory until the handle is "discarded", that is, `__wt_free(session, btree->key_format);` can live in `__wt_btree_close` or `__wt_b tree_discard`. If it's in "close", then there's some risk another thread of control will try and look at it; if it's in "discard", we have to somehow figure out if it's a re-open or a new open, because in the case of a re-open, we need to discard that memory, and in the case of a new open, we can't touch it. I left that work in close, but I don't feel strongly.

* note I'm clearing the `WT_BTREE_SPECIAL_FLAGS` in `__wt_conn_btree_sync_and_close`, because that layer currently "owns" those flags.

### DIFF
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -15,46 +15,6 @@ static int __btree_preload(WT_SESSION_IMPL *);
 static int __btree_tree_open_empty(WT_SESSION_IMPL *, bool);
 
 /*
- * __btree_initialize --
- *	Initialize the WT_BTREE structure.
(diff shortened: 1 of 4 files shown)


michaelcahill:
Thanks @keithbostic, I prefer this approach.  I'll merge...

PR #3317, opened 2017-03-01
WT-3199 bug: eviction assertion failure

keithbostic:


keithbostic:
@michaelcahill, this is only lightly tested, but seems to work and I can no longer reproduce the failure from WT-3199.

The only necessary fix is 9a70331, of course, 80497b3 is cleanup.

I pushed 8b218e6 (removing `WT_BTREE_NO_EVICTION`), but it broke right after I pushed it, so I reverted it. If you like this change, feel free to merge, I'll look at removing `WT_BTREE_NO_EVICTION` in a separate ticket if it makes sense.

As I said on the call, I never did figure out the exact path 9a70331 fixes. I'm guessing it's some race clearing `WT_BTREE_NO_EVICTION` without correctly adjusting `WT_BTREE.evict_disabled`, but I couldn't find it.

For testing, the CONFIG in the WT-3199 ticket generally fires for me in under 500 runs.

michaelcahill:
Thanks @keithbostic, this change looks right to me.  I think the issue was that eviction could start a walk in a btree just before it was switched in to become primary.  I'm not sure why this didn't happen before WT-3193 but this fix looks reasonable.

Hopefully the round trip to interrupt the eviction server every time we start a new LSM chunk won't be significant, but we've already talked about using an epoch in eviction rather than the locking we're currently doing in there.

I'll merge.

PR #3319, opened 2017-03-06
WT-3207 Don't hold clean handles during checkpoints.

michaelcahill:
Previously, we gathered handles, then started a transaction, then
figured out which handles were clean and released them.  However,
* checkpoints were keeping every handle in both its handle list and in
the meta_tracking list because the *_apply_all functions were saving all
handles when meta_tracking was active; and
* we had acquired exclusive locks on checkpoints to be dropped before
determining that we could skip a checkpoint in a clean tree.  These
locks blocked drops (among other things) until the checkpoint completed.

The solution here is to first start the transaction, then check for
clean handles as checkpoint visits them.  However, this has to cope with
races where a handle changes state in between the transaction starting
and getting the handle (e.g., table creates, bulk loads completing).

### DIFF
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -400,10 +400,7 @@ __conn_btree_apply_internal(WT_SESSION_IMPL *session, WT_DATA_HANDLE *dhandle,
 		return (ret == EBUSY ? 0 : ret);
 
 	WT_SAVE_DHANDLE(session, ret = file_func(session, cfg));
-	if (WT_META_TRACKING(session))
-		WT_TRET(__wt_meta_track_handle_lock(session, false));
-	else
-		WT_TRET(__wt_session_release_btree(session));
+	WT_TRET(__wt_session_release_btree(session));
 	return (ret);
 }
 
@@ -497,7 +494,12 @@ __wt_conn_dhandle_close_all(
 
 		session->dhandle = dhandle;
 
-		/* Lock the handle exclusively. */
+		/*
+		 * Lock the handle exclusively.  If this is part of
+		 * schema-changing operation (indicated by metadata tracking
+		 * being enabled), hold the lock for the duration of the
+		 * operation.
+		 */
 		WT_ERR(__wt_session_get_btree(session,
 		    dhandle->name, dhandle->checkpoint,
 		    NULL, WT_DHANDLE_EXCLUSIVE | WT_DHANDLE_LOCK_ONLY));
--- a/src/meta/meta_apply.c
+++ b/src/meta/meta_apply.c
@@ -45,11 +45,7 @@ __meta_btree_apply(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
 		    session, uri, NULL, NULL, 0)) != 0)
 			return (ret == EBUSY ? 0 : ret);
 		WT_SAVE_DHANDLE(session, ret = file_func(session, cfg));
-		if (WT_META_TRACKING(session))
-			WT_TRET(__wt_meta_track_handle_lock(
-			    session, false));
-		else
-			WT_TRET(__wt_session_release_btree(session));
+		WT_TRET(__wt_session_release_btree(session));
 		WT_RET(ret);
 	}
 	WT_RET_NOTFOUND_OK(ret);
--- a/src/session/session_dhandle.c
+++ b/src/session/session_dhandle.c
@@ -560,7 +560,7 @@ __wt_session_get_btree(WT_SESSION_IMPL *session,
 int
 __wt_session_lock_checkpoint(WT_SESSION_IMPL *session, const char *checkpoint)
 {
-	WT_DATA_HANDLE *dhandle, *saved_dhandle;
+	WT_DATA_HANDLE *saved_dhandle;
 	WT_DECL_RET;
 
 	WT_ASSERT(session, WT_META_TRACKING(session));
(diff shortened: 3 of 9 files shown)


src/conn/conn_dhandle.c
```diff
@@ -400,10 +400,7 @@ __conn_btree_apply_internal(WT_SESSION_IMPL *session, WT_DATA_HANDLE *dhandle,
 		return (ret == EBUSY ? 0 : ret);
 
 	WT_SAVE_DHANDLE(session, ret = file_func(session, cfg));
-	if (WT_META_TRACKING(session))
-		WT_TRET(__wt_meta_track_handle_lock(session, false));
-	else
-		WT_TRET(__wt_session_release_btree(session));
```
michaelcahill: Don't save handles for the duration just because they are visited -- checkpoint filters the handles it needs to visits and stashes them.

src/meta/meta_apply.c
```diff
@@ -45,11 +45,7 @@ __meta_btree_apply(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
 		    session, uri, NULL, NULL, 0)) != 0)
 			return (ret == EBUSY ? 0 : ret);
 		WT_SAVE_DHANDLE(session, ret = file_func(session, cfg));
-		if (WT_META_TRACKING(session))
-			WT_TRET(__wt_meta_track_handle_lock(
-			    session, false));
-		else
-			WT_TRET(__wt_session_release_btree(session));
```
michaelcahill: Matching `__conn_btree_apply`

src/session/session_dhandle.c
```diff
@@ -560,18 +560,23 @@ __wt_session_get_btree(WT_SESSION_IMPL *session,
...
-	 * while we are creating the new checkpoint.
+	 * while we are creating the new checkpoint.  Hold the lock until the
+	 * checkpoint completes.
 	 */
 	WT_ERR(__wt_session_get_btree(session, saved_dhandle->name,
 	    checkpoint, NULL, WT_DHANDLE_EXCLUSIVE | WT_DHANDLE_LOCK_ONLY));
+	if ((ret = __wt_meta_track_handle_lock(session, false)) != 0) {
+		WT_TRET(__wt_session_release_btree(session));
+		goto err;
+	}
```
michaelcahill: While reviewing, found a potential handle leak if operations fail before the lock handle is saved.

src/txn/txn_ckpt.c
```diff
@@ -239,22 +239,67 @@ int
...
+		WT_RET(ret);
+		if (metadata_race) {
+			__wt_verbose(session, WT_VERB_CHECKPOINT,
+			    "skipped checkpoint of %s with metadata conflict",
+			    session->dhandle->name);
+			F_CLR(&session->txn, WT_TXN_ERROR);
+			F_SET(btree, WT_BTREE_SKIP_CKPT);
+			return (0);
+		}
+	}
```
michaelcahill: This is the most horrible part of this change by a long way.  Your better idea goes here.

src/txn/txn_ckpt.c
```diff
@@ -631,29 +642,23 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
 	 */
-	WT_ASSERT(session, session->ckpt_handle_next == 0);
-	WT_WITH_SCHEMA_LOCK(session,
-	    WT_WITH_TABLE_READ_LOCK(session,
-		ret = __checkpoint_apply_all(
-		    session, cfg, __wt_checkpoint_get_handles, NULL)));
-	WT_ERR(ret);
+	__wt_spin_lock(session, &conn->schema_lock);
+	F_SET(session, WT_SESSION_LOCKED_SCHEMA);
+	locked_schema = true;
```
michaelcahill: It's possible that the horribleness above in `__wt_checkpoint_get_handles` makes it unnecessary to hold the schema lock across the transaction begin and handle gathering.  This was my first attempt to close the window and it wasn't sufficient.

src/txn/txn_ckpt.c
keithbostic: I'm not clear on what "a recent durable point" means, exactly; since the durable point must precede the transaction's durable point, I don't see any problems. with one question: presumably the durable point was a default checkpoint, what if this checkpoint is named?

src/txn/txn_ckpt.c
keithbostic: It seems intuitive to me that would be the case, at least I can't think of a reason the test in `__wt_checkpoint_get_handles` wouldn't be sufficient.

agorrod: If you end up holding the schema lock across this block of code, I'd prefer if it was pulled into a separate function so we can use a ``WT_WITH_LOCK`` macro. FWIW I think is reasonable to hold the schema lock here since it blocks other schema level operations (create, drop, rename) from updating the metadata while checkpoint is deciding which handles should be included in the checkpoint. It may not be strictly necessary, but there are enough different co-ordination mechanisms in play here, that holding the schema lock makes me feel more comfortable.

agorrod:
@michaelcahill I didn't have much to add here - I'll take another look once you've answered the questions from @keithbostic

michaelcahill:
@keithbostic, you had a question in here about named checkpoints that disappeared due to code shuffling.  We are already skipping clean handles during named checkpoints: i.e., named checkpoints don't imply `force=true`.  I'll put in a fix for that and make them fail if they skip any trees due to metadata update conflicts.

agorrod:
Thanks @michaelcahill lgtm, I'll go ahead and merge

PR #3320, opened 2017-03-06
WT-2978 Python: make a pip-compatible installer.

ddanderson:
This pull request is currently for discussion, as it has a caveat in how it works.  That is, this version will not create a wheel (the preferred method for the pip installer), but it will fall back to an older way of building/installing the distribution. [JIRA description of the issue.](https://jira.mongodb.org/browse/WT-2978?focusedCommentId=1517052&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1517052) I'm going to continue to work on a better way to solve this.

Then, to create a distribution do: `cd lang/python; python setup_pip.py sdist`, that creates a `.tar.gz` file in `lang/python/dist`.

To test it out on a target machine, you need pip, and it's helpful to have virtualenv (so you don't need to install anything as root).  [If you don't have virtualenv installed, here's one way to get it, and pip, without privileges.](http://stackoverflow.com/questions/4324558/whats-the-proper-way-to-install-pip-virtualenv-and-distribute-for-python/5177027#5177027)  You also need a compiler, SWIG, some autoconf tools and snappy, zlib.  That is, the usual stuff needed for building WT.

On the target machine:
```
  $ virtualenv v1
  $ . v1/bin/activate
  $ pip install --no-binary :all: wiredtiger-2.9.2.tar.gz
  $ python examples/python/ex_access.py    # try it out
```
If you try just `pip install wiredtiger-2.9.2.tar.gz` it will work, but with an error message (that's the caveat above).

### DIFF
--- a/lang/python/setup_pip.py
+++ b/lang/python/setup_pip.py
@@ -27,10 +27,11 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 
-# This script builds a Python source distribution that can built be
-# installed via pip install. This must be run in a git repository to
-# determine the files to package. In this directory, run
-# "python setup_pip.py sdist" to create a tar.gz file under ./dist
+# This script builds a Python source distribution that can built be installed
+# via pip install. This must be run in a git repository to determine the files
+# to package. Also as a prerequisite, SWIG must be run as the generated files
+# are part of the package. To create the distribution, in this directory, run
+# "python setup_pip.py sdist", this creates a tar.gz file under ./dist .
 from __future__ import print_function
 import os, os.path, re, shutil, site, sys
 from setuptools import setup, Distribution
@@ -126,7 +127,6 @@ def get_build_path():
     build_paths = []
     find_executable('autoreconf', build_paths)
     find_executable('aclocal', build_paths)
-    find_executable('swig', build_paths)
     build_path = os.environ['PATH'] + ':' + ':'.join(build_paths)
     return build_path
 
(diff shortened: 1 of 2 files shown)


ddanderson:
Pretty close, but not quite ready for final review, I need to test snappy/zlib. The latest commits remove the big caveat mentioned above. That is, `pip install *.tar.gz` installs without error.

ddanderson:
@michaelcahill, this is ready for review. That is, `pip install *.tar.gz` installs without error, and the python examples and simple examples using snappy/zlib can be run, at least on my OSX laptop and kodkod (CentOS).

Known limitations: Windows not supported (probably wouldn't be difficult to do), python3 not supported. snappy header files must be present on the target system (as well as snappy and zlib libraries).

Once we get this where we like it, we should probably set up a jenkins job to build a distribution, install it, and run some simple examples.

michaelcahill:
@ddanderson, this looks good, and is self-contained so should be completely safe to commit.

One thing I haven't completely understood is the connection between this package and https://pypi.python.org/pypi -- the upload instructions seems to expect `setuptools` or `twine` will upload the package (I don't know anything about `twine`).

Do you understand what is involved there, so we can eventually just run `pip install wiredtiger` and have it grab the package from the repository?  Will that force us to support Windows sooner rather than later?

ddanderson:
@michaelcahill, I'm going to investigate this using the testpypi site. I don't know the answer about whether it requires us to do Windows (I suspect not), but I'll find out.

ddanderson:
@michaelcahill, I tested this with testpypi and wrote up the procedures (for testpypi and pypi) in the JIRA https://jira.mongodb.org/browse/WT-2978 . With the tweaks just committed, it seems to install and work, so I don't think we need to address the Windows part of this until we want to.  It's ready to merge, I think.  The next step could be to push something to pypi at the next release point, or I could package up a distribution based on WT-2.9.1 now.  We'll need to update the release procedure.

michaelcahill:
Thanks @ddanderson, lgtm, I'll merge.

Thanks also for the upload instructions: we'll need to follow up when we cut the next release of WiredTiger.

PR #3322, opened 2017-03-08
WT-3202 wtperf should not reopen in-memory connections

gugemichael:
fix the potential problem on wtperf test wiredtiger in_memory mode. check wt_conn in_memory flag directly and without the "reopen_connection=false" option which is likely to lead to ambiguity

### DIFF
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -1653,6 +1653,10 @@ close_reopen(WTPERF *wtperf)
 	CONFIG_OPTS *opts;
 	int ret;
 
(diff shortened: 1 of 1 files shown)


wtbuild:

gugemichael:
@agorrod This is about https://jira.mongodb.org/browse/WT-3202

michaelcahill:
autobuild: ok to test

michaelcahill:
@gugemichael, thank you for submitting this change.  We have merged a different version of the fix in #3341, so I'm going to close this one for now.

PR #3323, opened 2017-03-08
WT-3213 Only error if fixed-length and long_running_txn is set.

sueloverso:
@keithbostic Please review this small fix to allow the generated CONFIG to be used on the next run.


keithbostic:
lgtm.

PR #3324, opened 2017-03-08
WT-3190 perform a complete re-tune of eviction workers every 30 seconds.

fedorova:
This branch addresses performance issues in MongoDB workloads with phase changes, where the eviction server was not proactive enough in creating workers.


fedorova:
autobuild: retest this please

fedorova:
autobuild: retest this please

agorrod:
@fedorova Sorry I've let this one sit - as discussed on the call I've been letting perfect be the enemy of good.

@sulabhM Could you do a review of this change?
@daveh86 Could you run this change through the Evergreen system performance tests, ensuring that eviction threads are configured to enable auto-tuning and compare with/without this change?

fedorova:
@agorrod, I do plan to investigate how we can detect phase changes more accurately, rather than using a fixed 30 second heuristic. So this is just a temporary improvement, which should not hurt any workloads and should help some.

src/evict/evict_lru.c
```diff
@@ -926,28 +930,63 @@ __evict_tune_workers(WT_SESSION_IMPL *session)
...
+		}
+	}
+	/*
+	 * If we have not reached stable state, don't do anything
+	 * unless enough time has passed since the last
+	 * time we have taken any action in this function.
 	 */
-	if (WT_TIMEDIFF_SEC(
-	    current_time, conn->evict_tune_last_time) < EVICT_TUNE_PERIOD)
+	else if (time_diff < EVICT_TUNE_PERIOD)
```
sulabhM: the else needs to be in the same line where if block ends. So this block might end up looking like:

```
        if (conn->evict_tune_stable) {
                if (time_diff < EVICT_FORCE_RETUNE)                              
                        return (0);                                                       
                /*                                                               
                 * Stable state was reached a long time ago. Let's re-tune. Reset all the state.
                 */                                                              
                conn->evict_tune_stable = 0;                             
                ....
                ....
        } else if (time_diff < EVICT_TUNE_PERIOD)                                
                /*                                                               
                 * If we have not reached stable state, don't do anything unless 
                 * enough time has passed since the last time we have taken any  
                 * action in this function.                                      
                 */                                                              
                return (0);
```

src/evict/evict_lru.c
```diff
@@ -926,28 +930,63 @@ __evict_tune_workers(WT_SESSION_IMPL *session)
...
 		return (0);
 
+	/*
+	 * Measure the number of evicted pages so far. Eviction rate
```
sulabhM: 80 character comment formatting needed

agorrod:
Thanks @fedorova and @sulabhM this seems like an improvement to me - I'll go ahead and merge.

PR #3326, opened 2017-03-10
WT-3207 Report a message for conflicting forced checkpoints, rather than an error

agorrod:
Continue to return EBUSY. Have test/fops handle EBUSY returns from forced checkpoints.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -319,10 +319,12 @@ __wt_checkpoint_get_handles(WT_SESSION_IMPL *session, const char *cfg[])
 			 * safely be skipped here.
 			 */
 			F_CLR(&session->txn, WT_TXN_ERROR);
-			if (force)
-				WT_RET_MSG(session, EBUSY,
+			if (force) {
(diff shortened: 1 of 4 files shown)


src/txn/txn_ckpt.c
```diff
@@ -319,10 +319,12 @@ __wt_checkpoint_get_handles(WT_SESSION_IMPL *session, const char *cfg[])
 			 * safely be skipped here.
 			 */
 			F_CLR(&session->txn, WT_TXN_ERROR);
-			if (force)
-				WT_RET_MSG(session, EBUSY,
+			if (force) {
+				WT_RET(__wt_msg(session,
```
agorrod: I chose to propagate the error from `__wt_msg` rather than EBUSY, since errors from `__wt_msg` seem more concerning. I could be convinced to go the other way

agorrod:
autobuild: retest this please

agorrod:
@michaelcahill I'm going to go ahead and merge this change - please take a look to let me know if you think any of the changes I made to fops are unreasonable.

PR #3327, opened 2017-03-10
WT-3212 Table cursors should not free memory owned by the table.

ddanderson:


ddanderson:
@michaelcahill, can you please take a quick review?  Unfortunately I don't yet have a test case that exposes the problem.

src/cursor/cur_table.c
```diff
@@ -989,10 +989,14 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 
 	if (0) {
 err:		if (*cursorp != NULL) {
-			if (*cursorp != cursor)
+			if (*cursorp != cursor) {
+				/* The URI is owned by the table. */
+				(*cursorp)->internal_uri = NULL;
 				WT_TRET(__wt_cursor_close(*cursorp));
+			}
```
michaelcahill: I think this generally requires a great big comment.  I had to go back to understand how we could end up with `*cursorp != NULL` and `*cursorp != cursor` (i.e., a dump cursor wrapping the normal handle).

Would it be sufficient to call `(*cursorp)->close(*cursorp)` without clearing `internal_uri`?

src/cursor/cur_table.c
```diff
@@ -989,10 +989,14 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
...
 			*cursorp = NULL;
 		}
+		cursor->internal_uri = NULL;
```
michaelcahill: This one isn't required because `__curtable_close` takes care of it.

ddanderson:
@michaelcahill, please review the latest changes, I hope it's clearer.

ddanderson:
I tested this by adding temp code to inject an error, and ran it with and without the change.  I haven't been able to create a test case that covers this.

PR #3328, opened 2017-03-12
WT-3216 add support for clang-tidy

keithbostic:


michaelcahill:
@keithbostic, I've looked through the changes and can do a more thorough review, but was expecting to see changes to run `clang-tidy` either via `make` or a script.  What's the plan there so we don't reintroduce these issues?

Also, is `clang-tidy` sufficiently functional that we can retire `flexelint`?  It would be great to have something automated that everyone can run locally...

keithbostic:
> ... but was expecting to see changes to run `clang-tidy` either via `make` or a script.  What's the plan there so we don't reintroduce these issues?

Absent a better idea, I'll periodically run `clang-tidy` and review the output.

> Also, is `clang-tidy` sufficiently functional that we can retire `flexelint`?  It would be great to have something automated that everyone can run locally...

Agreed, but no, `flexelint` is a lot more aggressive/interesting than `clang-tidy`. As far as I can tell, `clang-tidy` doesn't even replace `scan-build` (another `clang` static analyzer API).

Here's the current situation...

Over time I've added various "lint" tools in my local environment, a set of tools I run periodically and look for warnings and/or compiler complaints. The list includes: `gcc 4.7`, `gcc 4.9`, `gcc 5.4`, `gcc 6.3`, `clang 3.4`, `clang 3.8`, `scan-build`, `clang-tidy`, a handful of small scripts that look for problems (but which aren't guaranteed to be errors, which is why they're not in scripts in `dist`), and, of course, `flexelint`.

Of those tools, `scan-build`, `clang-tidy`, `flexelint` and some of the scripts produce lots of output we don't care about. As an example, the `google-readability-casting` flag to `clang-tidy` complains if you screw up a cast. Unfortunately, it also complains if you potentially screwed up a cast on a particular machine. For example, `clang-tidy` complains about both:
```
size_t a, b;
a = (size_t)b;
```
and
```
uintmax_t x;
size_t y;
x = (uintmax_t)y;
```
because `uintmax_t` might be the same size as `size_t` for a specific machine/compiler combination. We want to fix the former and ignore the latter, but unfortunately there's a lot of noise in that signal.

The clang tool output we ignore isn't that bad, a few hundred lines, but we're ignoring more than 3,000 lines of `flexelint` output at the moment.

The way I deal with all of the noisy tools is to run the tool and diff the output against a previous run. It's pretty easy to skim the output and ignore the paired lines that only means a complaint I'm ignoring changed line numbers in the file.

I often do a run about the time there's a pull request I'm pretty sure is going to be merged, both to make sure I've not created any problems and to avoid the hassle of creating new tickets for trivial changes (9f72430).

I try not to only do that in tickets we'll never backport, but I realize that's not an exact science.

I don't have any interest in "owning" this process, and the more automation, the better, of course. I do think we should continue making lint-style checks, I think it helps us keep our tree cleaner and more readable.

Anyway, all that said, other than wanting to keep doing general lint and other checks, I'm happy to modify the process in any way that would benefit the group and minimize the effort.

cc: @agorrod

keithbostic:
Reviewing the changes I've made for `clang-tidy` so far, the only one I care about a lot is the tool finds places where we're not correctly enclosing macro arguments to avoid operator precedence bugs. I'm kind of shocked that none of our other tools figure that out...

Since failing to correctly enclose a macro's arguments is pretty much a time-bomb waiting to happen, I think it's worth continuing to run `clang-tidy`.

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill, I'm done with this one for now, I think I've gotten all the low-hanging fruit.

There still remain recommendations from `clang-tidy` that make sense to take, but the changes all require thought.

michaelcahill:
@keithbostic, I'd like `clang-tidy` to be automated if we're going down this path.  We already do that for various versions on gcc, clang and for scan-build, so having you run it manually means (a) nobody else can do this work, (b) nobody writes code that is clean by default and (c) lots of "lint" commits that really should have been part of some other change.

We already automate scan-build, e.g., http://build.wiredtiger.com:8080/job/wiredtiger-clang-analyzer/3122/clangScanBuildBugs/

There is a Jenkins plugin to run `clang-tidy` the same way: https://github.com/jenkinsci/clang-tidy-plugin -- I think we should set up a job that runs `clang-tidy` like we have for `scan-build`.

Both jobs should be configured to fail when the number of results increases past some threshold level: the reason the current scan-build job doesn't get much attention is that it never fails.

src/log/log.c
```diff
@@ -312,11 +312,11 @@ __wt_log_written_reset(WT_SESSION_IMPL *session)
...
 	conn = S2C(session);
-	if (!FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED))
-		return;
-	log = conn->log;
-	log->log_written = 0;
-	return;
+
+	if (FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED)) {
+		log = conn->log;
+		log->log_written = 0;
```
sueloverso: I guess this could really just become a single line `conn->log->log_written = 0;`.

src/btree/bt_io.c
```diff
@@ -184,6 +184,7 @@ __wt_bt_write(WT_SESSION_IMPL *session, WT_ITEM *buf,
 	int compression_failed;		/* Extension API, so not a bool. */
 	uint8_t *dst, *src;
 	bool data_checksum, encrypted;
+	bool timer;
```
sueloverso: Minor nit - join to the line above it.

src/log/log.c
keithbostic: Come to think of it, why are we testing `WT_CONN_LOG_ENABLED`, does it matter?

sueloverso: If `WT_CONN_LOG_ENABLED` is not set then `conn->log` is NULL so we want to avoid deref'ing the NULL pointer.  So we need some conditional in some form.

michaelcahill:

PR #3329, opened 2017-03-13
WT-3207 Fix a leak if a checkpoint fails.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm.

PR #3330, opened 2017-03-14
WT-98 Update the current cursor value without a search

keithbostic:


keithbostic:
@agorrod, @michaelcahill, here are the rest of the changes for WT-98.

keithbostic:
@agorrod, @michaelcahill, I changed some of the LSM cursor routines to not copy cursor keys into local memory if the cursor is positioned, in commit af2c787.

Would someone please check that carefully? It passes the test suite, but I'm not that comfortable making changes in LSM.

keithbostic:
@michaelcahill, we don't currently document the state of the `WT_SESSION.truncate` cursors on return.

In the current implementation, the `start` cursor has moved to the `stop` cursor's position or the end of the table, but I'm hesitant to guarantee that, and I don't see much value there.

In 9378cca, I changed `WT_SESSION.truncate` to reset any specified cursors, and documented it. Let me know if you want something different.

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

michaelcahill:
@keithbostic sorry I didn't get this done today.  Tomorrow for sure.

src/btree/bt_cursor.c
```diff
@@ -409,6 +414,15 @@ __wt_btcur_search(WT_CURSOR_BTREE *cbt)
...
 	WT_STAT_DATA_INCR(session, cursor_search);
 
+	WT_CURFILE_OP_PUSH;
+
+	/*
+	 * The pinned page goes away if we do a search, make sure there's a
+	 * local copy of any key, then re-save the cursor state.
+	 */
+	WT_ERR(__cursor_copy_int_key(cursor));
+	WT_CURFILE_OP_PUSH;
```
michaelcahill: Even with the comment, it seems confusing to "push" twice here.  It's really more "save / restore state" than "push / pop".

Since we're in the area and I had to go back and page this in, I think it would be a lot clearer if `WT_CURFILE_OP_DECL` declared a named variable (an opaque struct) that is passed to inlined `save` and `restore` functions, to make it clear they have some state and operate on the `cursor` local variable.

src/btree/bt_cursor.c
```diff
@@ -558,15 +588,18 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
 			exact = -1;
 	}
 
+err:	if (exactp != NULL && (ret == 0 || ret == WT_NOTFOUND))
+		*exactp = exact;
```
michaelcahill: Can you please add a comment explaining why we need to set `exactp` when `ret == WT_NOTFOUND`?

src/btree/bt_cursor.c
```diff
@@ -945,11 +1092,13 @@ err:	if (ret == WT_RESTART) {
...
 	 */
-	if (ret == 0)
+done:	if (ret == 0)
 		WT_TRET(__wt_kv_return(session, cbt, cbt->modify_update));
-
-	if (ret != 0)
+	else {
 		WT_TRET(__cursor_reset(cbt));
+		WT_CURFILE_OP_POP;
+	}
```
michaelcahill: This code was previously structured to handle an error from `__wt_kv_return` the same as other errors (and reset the cursor).  Switching to an `else` has lost that, can it go back to an `if (ret != 0)`?

src/btree/bt_ret.c
```diff
@@ -142,8 +142,20 @@ __value_return(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd)
...
+	/*
+	 * We may already have an internal key, in which case the cursor may
+	 * not be set up to get another copy (for example, when we rely on a
+	 * search-function result).
+	 */
+	F_CLR(cursor, WT_CURSTD_KEY_EXT);
+	if (!F_ISSET(cursor, WT_CURSTD_KEY_INT)) {
+		WT_RET(__key_return(session, cbt));
+		F_SET(cursor, WT_CURSTD_KEY_INT);
+	}
```
michaelcahill: This seems obscure to me -- we'll silently fail to return the current key if one is already set in the cursor.  Can we assert `!F_ISSET(cursor, WT_CURSTD_KEY_INT)` here instead?

src/btree/bt_cursor.c
keithbostic: Yeah, I think that makes things a lot better, I totally agree.

src/btree/bt_cursor.c
keithbostic: Agreed.

src/btree/bt_ret.c
keithbostic: If you call `WT_CURSOR.update` with the current key set to a pinned page, successfully update that row without doing a search, and then call `__wt_key_return`, `WT_CURSTD_KEY_INT` will be set, and the cursor will reference the correct on-page key (which hasn't changed).

What's worse, we can't "update" the key, even if we wanted to do the extra work, because absent a search we don't have the information we need to set the key.

That's why the key return code looks different from value return code, the value return code doesn't have the "it was set and hasn't changed" case.

src/btree/bt_cursor.c
keithbostic: I think it's a mistake -- a change went in about 4 years ago in 0572f6c that looks incorrect to me, there's no reason we should be testing `ret == WT_NOTFOUND`.

The test suite passes when that test is removed.

keithbostic:
@michaelcahill, I've pushed the changes from your review.

michaelcahill:

PR #3332, opened 2017-03-16
WT-3206 Fix a race allocating split generations.

michaelcahill:
We use split generations to detect when readers may be looking at
structures that are replaced by a split.  For correctness, we should
only increment the global split generation *after* a split becomes
public.  Only then can we safely check that no thread is still reading
with the old generation.

Previously, a split could increment the global split generation, then a
thread could start reading with the new split generation but see the old
index structure.

This issue was introduced by WT-3088, where we wanted a way to ensure
that newly-allocated pages don't split until it is safe.  That is solved
here by having the split code pin a split generation in the ordinary way
(without allocating a new one) for the duration that splits of new
pages need to be prevented.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -187,7 +187,7 @@ __split_safe_free(WT_SESSION_IMPL *session,
 		exclusive = true;
 
 	if (exclusive) {
-		__wt_free(session, p);
+		__wt_overwrite_and_free_len(session, p, s);
 		return (0);
 	}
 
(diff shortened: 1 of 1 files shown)


src/btree/bt_split.c
```diff
@@ -187,7 +187,7 @@ __split_safe_free(WT_SESSION_IMPL *session,
 		exclusive = true;
 
 	if (exclusive) {
-		__wt_free(session, p);
+		__wt_overwrite_and_free_len(session, p, s);
```
michaelcahill: Note to reviewers: this is intended to catch future bugs more reliably.  It matches what we do in `__wt_split_stash_discard`.

agorrod:
Thanks @michaelcahill lgtm

PR #3336, opened 2017-03-17
WT-3196 Prevent eviction in LSM primaries after they are flushed.

michaelcahill:
Once an LSM primary is known to be on disk, we expect readers to use the
checkpoint.  The original page image for the primary will then be
discarded by an LSM worker thread.

We previously allowed the LSM primary to be evicted in between so that
eviction workers can deal with cache pressure ahead of the LSM worker
threads discarding the chunk.  However, that leads to cases where
application threads end up evicting a 100MB page, and also means that
discarding the chunk needs to worry about split generations (the cause
of the assertion failure here).

The solution suggested here is simple: never enable eviction in LSM
primaries, which also means we never need to fix up cache accounting.

### DIFF
--- a/src/lsm/lsm_work_unit.c
+++ b/src/lsm/lsm_work_unit.c
@@ -256,51 +256,6 @@ err:
 }
 
 /*
- * __lsm_switch_primary_off --
- *      Switch when a btree handle is no longer the current primary chunk of
- * an LSM tree.
- */
-static void
-__lsm_switch_primary_off(WT_SESSION_IMPL *session)
-{
-	WT_BTREE *btree;
-	WT_CACHE *cache;
-	WT_PAGE *child, *root;
-	WT_PAGE_INDEX *pindex;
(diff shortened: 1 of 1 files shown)


michaelcahill:
@sueloverso can you please run the original workload on this branch over the weekend to see whether it is fixed?  Also please take a look at the change.

@agorrod, there is an explanation above, we should talk some more about whether this is reasonable but as you can see it does simplify some things.

agorrod:
autobuild: retest this please

agorrod:
Thanks @michaelcahill this change works for me.

sueloverso:
I kicked this off over the weekend and after a day it hit the assertion failure in WT-2990 (which I have reopened).  Otherwise it appears to fix hitting the specific problem in that ticket.  WT-2990 is almost 6 months old so I don't think this change caused that problem, just exposed or widened a window to it.

PR #3337, opened 2017-03-17
SERVER-28194 Missing WiredTiger.turtle file loses data

keithbostic:
@markbenvenuto, can you please take a look at this change? -- thanks!


src/os_win/os_fs.c
```diff
@@ -86,23 +86,11 @@ __win_fs_rename(WT_FILE_SYSTEM *file_system,
...
-		if (DeleteFileW(to_wide->data) == FALSE) {
-			windows_error = __wt_getlasterror();
-			__wt_errx(session,
-			    "%s: file-rename: DeleteFileW: %s",
-			    to, __wt_formatmessage(session, windows_error));
-			WT_ERR(__wt_map_windows_error(windows_error));
-		}
-
-	if (MoveFileW(from_wide->data, to_wide->data) == FALSE) {
+	if (MoveFileExW(from_wide->data,
```
agorrod: There is a slightly offputting article here about the semantic of MoveFileExW with MOVEFILE_REPLACE_EXISTING:
http://www.tek-tips.com/viewthread.cfm?qid=570894

I can't find the knowledge base article it refers to, and it's pretty old, so maybe it's not relevant?

Possibly more relevant:
http://stackoverflow.com/a/167475

Which says:
```
Setting this value guarantees that a move performed as a copy and delete operation
is flushed to disk before the function returns. The flush occurs at the end of the
copy operation.
```

That's not reassuring, because what happens if there is a hard power failure between the copy and delete?

keithbostic: My reading of the `MoveFileEx Deletes the File Being Renamed` article is the problem only appears when the `from` and `to` names are the same (or differ only in case). WiredTiger never renames a file to the same name or to a name that differs only in case, so that shouldn't be a problem for us.

My reading of the StackOverflow article (and searching a bit more), is MSDN doesn't offer a guaranteed atomic rename (except maybe [MoveFileTransacted](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365241(v=vs.85).aspx), and Microsoft says not to use that function).

[A Microsoft paper referenced in the StackOverflow article](http://research.microsoft.com/pubs/64525/tr-2006-45.pdf) includes the paragraph:

> To perform a safe write, a new version of the file is created with a temporary name. Next, the new data is written to the temporary file and those writes are flushed to disk. Finally, the new file is renamed to the permanent file name, thereby deleting the file with the older data. Under UNIX, rename() is guaranteed to atomically overwrite the old version of the file. Under Windows, the ReplaceFile() call is used to atomically replace one file with another.

However, the [ReplaceFile documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365512(v=vs.85).aspx) doesn't declare it's atomic, although it does say:
> The backup file, replaced file, and replacement file must all reside on the same volume.

which makes me think it's likely to be atomic, I think that wording points to a metadata update, which should be atomic on any modern filesystem.

There's also a comment in an MSDN discussion forum [MoveFileEx(MOVEFILE_REPLACE_EXISTING) + NTFS + same volume = atomic?](https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/449bb49d-8acc-48dc-a46f-0760ceddbfc3/movefileexmovefilereplaceexisting-ntfs-same-volume-atomic) from a Microsoft employee:

> If the MoveFile stays on the same volume and no incompatible flags are specified then MoveFile will TRY to fulfill the request by calling ZwSetInformationFile(Rename), which may or may not be atomic depending on the filesystem involved. But if ZwSetInformationFile fails or if your MoveFile request is incompatible with ZwSetInformationFile for any reason, MoveFile will silently fall back to the normal CopyFile + DeleteFile, which is not atomic. That means you can never rely on MoveFile to be atomic. If it succeeds, it probably was atomic, but it might fail halfway.
>
> It is very possible for MoveFile to be atomic, but it is also possible for something to go wrong and for it to fall back to a non-atomic behavior even if the underlying filesystem supports atomic rename.
>
> If you were to use the unsupported and semi-documented NtSetInformationFile function directly, you could avoid any possibility of falling back to CopyFile + DeleteFile. You would still be at the mercy of the filesystem's implementation.
>
> I would guess that all major filesystems (including NTFS and FAT) implement rename-within-the-same-directory as a single metadata operation. Metadata operations are always atomic (they either happen, do not happen, or corrupt the filesystem). Note that NTFS differs from FAT only in that it promises to not corrupt the filesystem in the case of an interrupted metadata operation.

WiredTiger renames are done within a single directory, so I expect `MoveFileEx` to be atomic on any modern filesystem. We could switch to `ReplaceFile`, but I don't think it will make any difference.

I've added the `MOVEFILE_WRITE_THROUGH` flag to the `MoveFileEx` call. If we somehow end up in a copy-then-delete path, that flag adds a disk flush after the copy phase, so the window of vulnerability is as short as possible.

cc: @markbenvenuto

michaelcahill:

PR #3338, opened 2017-03-17
WT-3227 Python test suite inserts unnecessary whitespace in error output

keithbostic:


ddanderson:
@keithbostic, looks good to me, I'll merge.

PR #3339, opened 2017-03-17
WT-3228 table cursor update/remove overwrite configuration errors

keithbostic:


src/lsm/lsm_cursor.c
```diff
@@ -1250,10 +1250,10 @@ __clsm_lookup(WT_CURSOR_LSM *clsm, WT_ITEM *value)
 	WT_ERR(WT_NOTFOUND);
 
 done:
-err:	F_CLR(cursor, WT_CURSTD_KEY_SET | WT_CURSTD_VALUE_SET);
-	if (ret == 0) {
-		clsm->current = c;
+err:	if (ret == 0) {
+		F_CLR(cursor, WT_CURSTD_KEY_SET | WT_CURSTD_VALUE_SET);
```
agorrod: Is the change to clearing the flags safe? The content of `cursor->key` is overwritten at line 1234 above, and the content of `cursor->value` isn't.

keithbostic: I believe it's safe, but I doubt it's correct in all error paths.

By that I mean we're not going to drop core or anything, but the cursor's external key/value references should be left untouched on error, and I don't think this code gets that semantic correct in the error cases. (To be fair, the previous version of the code didn't either, I didn't make anything worse.)

To make all this correct, I think we'd need to do something similar to the way the btree cursors work in WT-98, that is, save the "external" state of the cursor at a high level, then when an error is about to be returned, reset the state of the cursor's key and value to their original state. It's too hard to reason through all of the cursor flags vs. error paths (and in this case, the various functions that call `__clsm_lookup`), and even if we got it correct once, the next LSM code change would break it. To fix this we should wrap the underlying functions, including `__clsm_lookup`.

I can do that work if you want, it's not too difficult and it shouldn't cost any performance.

michaelcahill:
Thanks @keithbostic, lgtm, I'll merge.  Let's revisit LSM cursor flag handling post WT-98.

PR #3340, opened 2017-03-18
WT-3136 bug fix: WiredTiger doesn't check sprintf calls for error return

keithbostic:
This change checks the sprintf, snprintf, vsprintf and vsnprintf calls in the tree for error returns.

The big picture is to create 3 pairs of functions that cover WiredTiger's usage of the sprintf family of functions:
1. `__wt_snprintf/__wt_vsnprintf`: fail if the buffer is too small for the formatted contents. Used in most cases where we have a buffer and we want to format into it.
2. `__wt_snprintf_len_set/__wt_vsnprintf_len_set`: set the returned length of the formatted contents, don't fail if the buffer is too small for the formatted contents. Used when we want to format into a buffer, with a return of the length of the results.
3. `__wt_snprintf_len_incr/__wt_snprintf_len_incr`: increment the returned length of the formatted contents, don't fail if the buffer is too small for the formatted contents. Used when we want to format into a buffer repeatedly, incrementing the total length as we go.

The reason for number 2 is because my first effort ended up with a lot of calls that looked like this:
```
size_t len;
...
len = 0;
__wt_snprintf_len_incr(..., &len, ...);
```
and I thought it was too easy to forget to set `len = 0;`, and if `len` is on the stack, that might bite us in the field.

@agorrod, @michaelcahill, this might be misguided, but it's the simplest solution I saw. Can you please take a look and consider this approach?

If Alex/Michael sign off on this change:

@sueloverso, would you please review `log/log.c`? And could we get a comment on the `len = 1;` line, why are we adding 1 to the length?

@ddanderson, would you please review `cursor/cur_json.c`?

@sulabhM, would you please own the full review? I'll add some review notes to flag changes that might be worth a closer look.


bench/wtperf/config.c
```diff
@@ -438,14 +438,13 @@ config_opt(WTPERF *wtperf, WT_CONFIG_ITEM *k, WT_CONFIG_ITEM *v)
...
 			begin = newstr = dstrdup(v->str);
 		else {
-			newlen += strlen(*strp) + 1;
-			newstr = dcalloc(newlen, sizeof(char));
-			snprintf(newstr, newlen,
-			    "%s,%*s", *strp, (int)v->len, v->str);
+			newlen = strlen(*strp) + v->len + strlen(",") + 1;
+			newstr = dmalloc(newlen);
+			testutil_check(__wt_snprintf(newstr, newlen,
+			    "%s,%.*s", *strp, (int)v->len, v->str));
```
keithbostic: Note the change from `%*s` to `%.*s`, this is a bug fix, additionally, the length of the allocated memory is too short and fails in testing.

test/recovery/truncated-log.c
```diff
@@ -193,9 +189,9 @@ fill_db(void)
 	max_key = min_key * 2;
 	first = true;
 	for (i = 0; i < max_key; ++i) {
-		snprintf(k, sizeof(k), "key%03d", (int)i);
-		snprintf(v, sizeof(v), "value%0*d",
-		    (int)(V_SIZE - strlen("value")), (int)i);
+		testutil_check(__wt_snprintf(k, sizeof(k), "key%03d", (int)i));
+		testutil_check(__wt_snprintf(v, sizeof(v), "value%0*d",
+		    (int)(V_SIZE - (strlen("value") + 1)), (int)i));
```
keithbostic: There's an off-by-one here, we weren't taking the buffer's nul-termination into account.

src/log/log.c
```diff
@@ -2246,8 +2246,10 @@ __wt_log_vprintf(WT_SESSION_IMPL *session, const char *fmt, va_list ap)
 		return (0);
 
 	va_copy(ap_copy, ap);
-	len = (size_t)vsnprintf(NULL, 0, fmt, ap_copy) + 1;
```
sueloverso: We account for the terminating NUL when we get the initial length of the string to write into the log file and allocate a log record to hold it.  We use that length in computing the log size record and log messages assume a NUL-terminated string.

src/os_win/os_snprintf.c
```diff
@@ -8,17 +8,47 @@
...
+		*retsizep += (size_t)len;
+		return (0);
+	}
 
-	va_start(args, _Format);
-	ret = _wt_vsnprintf(_DstBuf, _MaxCount, _Format, args);
-	va_end(args);
+	/* Return the buffer size required. */
+	if (len == -1)
+		*retsizep += (size_t)_vscprintf(fmt, ap) + 1;
```
keithbostic: Thanks, @sulabhM, I think you're correct. (The reason the `+1` was there was because it was in the original code and I just copied that code.)

I've pushed a fix in 72ca3ef.

PR #3343, opened 2017-03-21
WT-3230 add automated support for clang static analysis tools

keithbostic:


michaelcahill:
@keithbostic, I expected we would put `clang-tidy-diff.py` on the test machines: do we need to copy it into our tree?

With `dist/s_clang-tidy`, it isn't going to be sufficient to test `HEAD^`: we need to test all of the changes in the pull request, which is some incantation based on `$ghprbSourceBranch` and `$ghprbTargetBranch` (see https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin#GitHubpullrequestbuilderplugin-EnvironmentVariables).  We could add an argument with the git revision spec (that could default to `HEAD^`, if that's useful).

I guess my real question is whether you plan to run this locally, or if the intent is just to get it into Jenkins.  If it's only to be run by Jenkins, I think we can create a build based on what's here.

keithbostic:
> I expected we would put clang-tidy-diff.py on the test machines: do we need to copy it into our tree?

Yes -- we needed local changes in order to see complaints about our header files.

> With dist/s_clang-tidy, it isn't going to be sufficient to test HEAD^: we need to test all of the changes in the pull request, which is some incantation based on $ghprbSourceBranch and $ghprbTargetBranch

Agreed. I haven't done the work to switch this over to "Jenkins mode", I figured that could wait until you decided how we were going to address false positives (see the last posting in WT-3230). I went far enough to convince myself there was a reasonable path to automation here and then stopped, sorry if that wasn't clear.

> I guess my real question is whether you plan to run this locally, or if the intent is just to get it into Jenkins. If it's only to be run by Jenkins, I think we can create a build based on what's here.

I was planning to get this into Jenkins, I understood that to be what you wanted.

But, to do that, we have to resolve the question of false positives.

keithbostic:
I'm going to backlog this ticket. The problem I see is we don't have a good way to ignore false positives, and turning off all the tests for which there are false positives makes the output significantly less useful, at least for current WiredTiger branches and using the clang 4.0 release.

PR #3344, opened 2017-03-22
WT-3155 Remove WT_CONN_SERVER_RUN flag

keithbostic:
@michaelcahill, I added barriers to the standard thread create/join APIs (similar to how we have a barrier in the yield API), and I thought that might interest you.

@agorrod, @sueloverso, most of the changes are in and around LSM manager startup and shutdown, so I hoped you could do the review.

@sueloverso, in 85336ab I renamed `WT_CONN_LOG_SERVER_RUN` to `WT_CONN_SERVER_LOG` to match the other server flags, please let me know if that's not right.


src/conn/conn_dhandle.c
```diff
@@ -314,8 +314,6 @@ __wt_conn_btree_open(
 	    F_ISSET(dhandle, WT_DHANDLE_EXCLUSIVE) &&
 	    !LF_ISSET(WT_DHANDLE_LOCK_ONLY));
 
-	WT_ASSERT(session, !F_ISSET(S2C(session), WT_CONN_CLOSING));
```
michaelcahill: What is the motivation for removing this assertion?  It has caught real bugs relating to thread management in the past...

src/lsm/lsm_worker.c
```diff
@@ -93,11 +92,13 @@ __lsm_worker(void *arg)
 
 	cookie = (WT_LSM_WORKER_ARGS *)arg;
 	session = cookie->session;
-	conn = S2C(session);
 
 	entry = NULL;
-	while (F_ISSET(conn, WT_CONN_SERVER_RUN) &&
-	    F_ISSET(cookie, WT_LSM_WORKER_RUN)) {
+	for (;;) {
+		WT_BARRIER();			/* Don't cache the run flag. */
```
agorrod: Do we use this scheme anywhere else? I don't understand why this check is unique.

src/include/lsm.h
```diff
@@ -162,6 +162,9 @@ struct __wt_lsm_manager {
 #define	WT_LSM_MAX_WORKERS	20
 #define	WT_LSM_MIN_WORKERS	3
 	WT_LSM_WORKER_ARGS lsm_worker_cookies[WT_LSM_MAX_WORKERS];
+
+#define	WT_LSM_MANAGER_SHUTDOWN	0x01	/* Manager has shut down */
```
agorrod: I don't love introducing a special shutdown path for LSM worker threads. The path I'd like to take this down is switching LSM over to using the `thread_group` code that the eviction server currently uses. I thought there was a JIRA ticket describing that change, but I can't find it. There is outstanding work in WT-3123 that changes the (internal) thread group API, so I'm inclined to hold off until that's resolved.

src/conn/conn_dhandle.c
keithbostic: The reason I removed it was because it meant we could set `WT_CONN_CLOSING` before connection close started shutting down subsystems. In other words, I wanted `WT_CONN_CLOSING` to mean "we're closing", rather than, "we've reached a point in closing after which we'll no longer open files".

What kind of bugs are you thinking of? Why would it be an error to open a file after connection close has started?

We could, of course, add another flag `WT_CONN_CLOSING_NO_OPEN`, that would mean no files should be opened after this time.

src/include/lsm.h
keithbostic: I certainly agree it would be better to switch over to using the `thread_group` code in LSM.

src/lsm/lsm_worker.c
keithbostic: What reason do you have to know these flags will be seen by the worker threads?

While I agree the existing code is likely to work, I don't believe that it's guaranteed to work, and a chunk of the changes in this branch fix potential races.

src/conn/conn_open.c
```diff
@@ -100,6 +94,10 @@ __wt_connection_close(WT_CONNECTION_IMPL *conn)
 		__wt_yield();
 	}
 
+	/* Shut down the subsystems, ensuring workers see the state change. */
+	F_SET(conn, WT_CONN_CLOSING);
```
agorrod: I believe this was after the subsystem destroy calls so that those subsystems had a chance to do an orderly shutdown of their worker threads. I'm not sure if that's currently required - but I think it's safer to keep the ordering. Why the order change in here? Why is the barrier required? Do some operations need to see the change to `WT_CONN_CLOSING`? The only place I can see that uses this flag to assert anything other than correctness is avoiding having application threads contribute to eviction.

src/lsm/lsm_worker.c
```diff
@@ -93,11 +92,13 @@ __lsm_worker(void *arg)
 
 	cookie = (WT_LSM_WORKER_ARGS *)arg;
 	session = cookie->session;
-	conn = S2C(session);
 
 	entry = NULL;
-	while (F_ISSET(conn, WT_CONN_SERVER_RUN) &&
```
agorrod: I'm personally a fan of having a pair of flags controlling thread shutdown. My concern is what happens if one of the manager threads stops due to an error, and skips clearing the sub-system flag. In that case shutdown will hang because the worker threads will never get the signal to stop.

Having said that..  other subsystems don't rely on a shared flag, so I don't have much of a leg to stand on here.

src/lsm/lsm_worker.c
agorrod: Can I ask a different question: What problem are you trying to fix with the barrier? Are you worried that the result of `F_ISSET(cookie, WT_LSM_WORKER_RUN)` will be cached across passes in this loop, and thus we won't exit because we don't see the state change? I understand your dislike of the `volatile` keyword, but that seems like exactly the case `volatile` is designed for.

A quick `grep -rnI "while.*F_ISSET" src/` reveals a dozen places where we use this same structure - I'd much rather not have this `WT_BARRIER` here - it's going to make me think every time I encounter the code.

agorrod:
@sueloverso I've done a round of review on these changes. Would you mind picking up the next round while I'm in transit?

I'm keen to keep in mind that we'd like to shift the LSM work threads to using the thread group functionality once WT-3129 is resolved.

src/conn/conn_dhandle.c
keithbostic: > `WT_ASSERT(session, !F_ISSET(S2C(session), WT_CONN_CLOSING));`
> What is the motivation for removing this assertion?

I've restored this assertion in b7e94da, using a new flag, `WT_CONN_CLOSING_NO_MORE_OPENS`.

src/lsm/lsm_worker.c
keithbostic: @agorrod 
> I'd much rather not have this WT_BARRIER here - it's going to make me think every time I encounter the code.

I've changed this to using a new `WT_LSM_WORKER_ARGS.running` field that's declared volatile, in commit e65259f.

src/conn/conn_open.c
keithbostic: @agorrod 
> I believe this was after the subsystem destroy calls so that those subsystems had a chance to do an orderly shutdown of their worker threads. I'm not sure if that's currently required - but I think it's safer to keep the ordering.

It's currently required because there's an assert when opening a Btree that we're not in the "closing" path, and that's why we didn't set `WT_CONN_CLOSING` until after async and LSM had exited.

> Why the order change in here?

The order change is because I wanted `WT_CONN_CLOSING` to mean "closing", rather than "at the point in closing where we no longer should open files". I've added a new flag `WT_CONN_CLOSING_NO_MORE_OPENS` to support the previous diagnostic check, hopefully that works for everybody!

> Why is the barrier required? Do some operations need to see the change to WT_CONN_CLOSING?

I doubt the barrier is required (I did the same review as you, and I agree no thread needs to see the flag at this time). That said: it's a flag other threads are expected to use to modify their actions, and it should have a barrier for that reason. Asking a future programmer to realize `WT_CONN_CLOSING` shouldn't be used for anything critical because it just might not be visible, is an unreasonable requirement.

src/lsm/lsm_worker.c
keithbostic: @agorrod 
> I'm personally a fan of having a pair of flags controlling thread shutdown.

I don't strongly disagree, although it potentially makes shutdown more complicated because a worker thread might exit earlier than the manager wants, for example, imagine the manager wants worker threads to finish up some work before exiting.

keithbostic:
@sueloverso, I believe I've responded to all of the comments.

The original goal of this ticket was to remove the `WT_CONN_SERVER_RUN` flag, and this set of changes does that. However, doing that required a number of changes in the LSM tree, and while looking through that code I found a few additional problems (for example, potential races and assumptions about opaque types).

All that said, I believe these changes fix existing problems in the LSM manager/worker code, and do not make it more (or less) difficult to switch the LSM worker threads to use the thread group code.

sueloverso:
I think your changes are agnostic for LSM using thread groups.  LGTM.

src/lsm/lsm_worker.c
agorrod: That ordering problem is why I probed about the change in ordering re: when `WT_CONN_CLOSING` is set in `__wt_connection_close`. I'll open a new ticket requesting a safety mechanism around shutdown of worker threads.

agorrod:
Thanks @keithbostic this change looks OK to me.

PR #3348, opened 2017-03-24
WT-3136 bug fix: WiredTiger doesn't check sprintf calls for error return

keithbostic:
Fix a typo.


keithbostic:
autobuild: retest this please

PR #3350, opened 2017-03-24
WT-3207 Use config to determine checkpoint force value.

sueloverso:
@michaelcahill Here's the change we discussed.  Please review.


michaelcahill:

PR #3352, opened 2017-03-25
WT-3063 Reserve records for read-modify-write

keithbostic:
`WT_CURSOR.reserve` implementation.


keithbostic:
@michaelcahill, all the interesting work is in 101bfa6.

keithbostic:
@michaelcahill, I think this one is ready for your review.

One change I'd like to consider is this branch does a search after each `WT_CURSOR.reserve` in order to support a subsequent `WT_CURSOR.get_value`. Given the low probability of a subsequent `get_value` call, I think it's likely wasted work. It's not too bad for a btree (a leaf page search in a pinned page), but in LSM it's a "read", so there is potential access to non-primary chunks in LSM, which kind of attacks the whole point of LSM.

We could set the cursor's value in the lower-level Btree code if the reserve call did a search (so if the reserve call wasn't positioned when the call was done), but that change is a little fiddly. And it only helps if the cursor isn't positioned when the reserve call is made, of course.

Or, I was wondering about postponing the search until the application actually calls `WT_CURSOR.get_value`, so, doing something like this in the get-value calls:
```
diff --git a/src/cursor/cur_std.c b/src/cursor/cur_std.c
index e48d77668..fe15b30d5 100644
--- a/src/cursor/cur_std.c
+++ b/src/cursor/cur_std.c
@@ -425,8 +425,19 @@ __wt_cursor_get_valuev(WT_CURSOR *cursor, va_list ap)
 
 	CURSOR_API_CALL(cursor, session, get_value, NULL);
 
-	if (!F_ISSET(cursor, WT_CURSTD_VALUE_EXT | WT_CURSTD_VALUE_INT))
-		WT_ERR(__wt_cursor_kv_not_set(cursor, false));
+	/*
+	 * A WT_CURSOR.reserve call might not have done a search and so wouldn't
+	 * have a value (and, in cases like LSM, the reserve call might not have
+	 * read the chunk where the key's value is held). We don't retrieve the
+	 * value during the original call because it's an expensive thing to do
+	 * (turning an LSM write into an LSM read), and because the likelihood
+	 * the application will ask for it is small. For that reason we can end
+	 * up here with an internal key and no value.
+	 */
+	if (!F_ISSET(cursor, WT_CURSTD_VALUE_SET) &&
+	    F_ISSET(cursor, WT_CURSTD_KEY_INT))
+		WT_ERR_NOTFOUND_OK(cursor->search(cursor));
+	WT_ERR(__cursor_checkvalue(cursor));
```
This change passes the test suite with one exception. If you do:
```
cursor.set_key()
cursor.search()        /* position the cursor */
cursor.remove()       /* cursor remains positioned */
cursor.get_value()    /* under-the-covers search invalidates cursor position */
cursor.next()             /* cursor starts at beginning of the table */
```

I could tighten the test down so that it only triggers if the previous call was a reserve call, but wanted to talk this over with you, maybe there's a better idea.

michaelcahill:
Thanks @keithbostic, I've been through this and it looks right to me. I'll merge.

PR #3353, opened 2017-03-26
SERVER-28168 Cannot start or repair mongodb after unexpected shutdown.

keithbostic:
Panic if there's an error in reading/writing from/to the turtle file, there's no point in continuing. This change avoids user confusion when the turtle file is corrupted or zero'd out by the filesystem.


sueloverso:
LGTM

PR #3355, opened 2017-03-27
WT-3238 Java: Fix Cursor.compare and Cursor.equals to return int values.

ddanderson:
Non-zero int values for these functions should not raise exceptions.
This commit includes various tests for compare/equals.


ddanderson:
@mwberry, you are welcome to review too.

mwberry:
LGTM!

michaelcahill:
Thanks @ddanderson, lgtm -- I'll merge.

PR #3356, opened 2017-03-29
WT-3208 Don't count page rewrites as eviction making progress.

michaelcahill:


michaelcahill:
Note: reinstates some logic from WT-3023 that was removed in WT-3028.

src/btree/bt_discard.c
```diff
@@ -16,13 +16,14 @@ static void __free_skip_array(
...
+static void __page_out_int(WT_SESSION_IMPL *, WT_PAGE **, bool);
 
 /*
- * __wt_ref_out --
+ * __wt_ref_out_int --
  *	Discard an in-memory page, freeing all memory associated with it.
  */
 void
-__wt_ref_out(WT_SESSION_IMPL *session, WT_REF *ref)
+__wt_ref_out_int(WT_SESSION_IMPL *session, WT_REF *ref, bool rewrite)
```
michaelcahill: In case it isn't clear, these "_int" functions were me not wanting to update every caller of `__wt_ref_out` and `__wt_page_out` to add an extra `false` argument.

michaelcahill:
Further note: it's likely that WT-3028 was sufficient at the time, but the solution was changed in WT-2960 to be something more heuristic and that may now be causing workloads to appear to make progress when they aren't.

agorrod:
Thanks @michaelcahill lgtm.

PR #3357, opened 2017-03-29
WT-3160 Queue empty internal pages for eviction.

michaelcahill:
Now that queuing is smarter, remove the skew against internal pages.


agorrod:
@michaelcahill I agree with the idea of this change, but I'm nervous that we'll evict more internal pages in out of cache workloads now. I came across a wtperf workload in WT-2260 that was used to tune internal page eviction efficiency in the past - I'd like to run it against this branch before merging.

michaelcahill:
@agorrod thanks for taking a look.  Using the workload from WT-2260 to evaluate this change makes sense.  I'll kick that off tomorrow if you don't beat me to it.

agorrod:
@michaelcahill I ran the workload in WT-2260 and this branch is doing eviction of internal pages - wheras develop doesn't do any:

```
$ grep "internal pages evic" branch/WiredTigerStat.* | tail -n 1
Mar 30 17:32:10 1700059 WT_TEST cache: internal pages evicted
$ grep "internal pages evic" develop/WiredTigerStat.* | tail -n 1
Mar 30 16:06:53 0 WT_TEST cache: internal pages evicted
```

Those numbers are accumulated over ~1200 seconds, so the rate of internal page eviction is about 1400/second on this branch for this workload. The internal page eviction doesn't hurt overall throughput for me, but it does make me feel uncomfortable.

michaelcahill:
Hi Alex,

Fair enough, I'll try with the skew against internal pages back in the
calculation.  My concern is with workloads that have lots of idle trees --
I think in that case we should evict internal pages over leaf pages in busy
trees.

Michael.

michaelcahill:
@agorrod I've made changes so we don't always consider internal pages for eviction, but only when getting aggresive or in idle trees.  Can you please take another look?

agorrod:
Thanks @michaelcahill I reran the test case, and it works better with this patch now.

I also increased the skew for the metadata file - because I really don't think we should be evicting pages from there. Though I suspect they are generally hot enough that skew doesn't matter.

I'm going to go ahead and merge.

PR #3358, opened 2017-03-29
WT-2439 Enhance reconciliation page layout

keithbostic:


keithbostic:
@sulabhM, the change to look at hard is in b2aca4c, all the rest of the churn was trying to change it so we verify all images we create during reconciliation, not just the ones we write to disk.

michaelcahill:
The fixes for dictionaries make sense to me and the tip of this branch has been running stress tests for a few hours.  LGTM, I'll merge.

PR #3359, opened 2017-03-29
WT-3244 metadata operations failing in in-memory configurations

keithbostic:
@michaelcahill, turning off in-memory cache-full checks on the metadata file is obviously a relatively big hammer.

If that's too dangerous, we could turn off the cache checks on just the unroll operations? Let me know what you prefer.


agorrod:
Thanks @keithbostic lgtm

PR #3360, opened 2017-03-29
WT-3243 slot release

sueloverso:
@agorrod or @michaelcahill please review this change.  With this I see a bit better performance with the repro case and I've tweaked the slot join statistics to give a better idea of what is going on during a switch.


sueloverso:
autobuild: retest this please

sueloverso:
This is not yet ready to merge.  I switched the perf tests to run this branch and one of the logging tests hung.  I am debugging that.

src/log/log_slot.c
```diff
@@ -371,6 +281,102 @@ __wt_log_slot_new(WT_SESSION_IMPL *session)
...
+		return (0);
+	WT_RET(WT_SESSION_CHECK_PANIC(session));
+
+	/*
+	 * We may come through here multiple times if we were not able to
+	 * set up a new one.  If we closed it already,
+	 * don't try to do it again but still set up the new slot.
+	 */
+	if (!F_ISSET(myslot, WT_MYSLOT_CLOSE)) {
+		ret = __log_slot_close(session, slot, &release, forced);
```
michaelcahill: Hi @sueloverso, I don't know if it's the problem you're facing, but it looks like the `release` value here can be lost in the new version of the code.  In particular, if `__log_slot_new` fails then when we retry the call, `release` will be false.   Maybe the `release` variable needs to be scoped in the calling function instead?

michaelcahill:
@sueloverso I realized my comment was probably cryptic so I pushed the change I was trying to describe in 1d57ee4.  Feel free to revert that if I'm on the wrong track.

src/log/log_slot.c
sueloverso: Actually the comment makes perfect sense, since this is fully paged into my head.  Thanks for taking a look at this.  I think your fix may address the hang I see.  I will try the repro case, which has hit in under 5 iterations repeatedly so far.

sueloverso:
@michaelcahill This is now working and is ready for review.  I changed the mechanism to retain release information so that it uses `myslot->flags` in the same way we retain the did-we-close information across calls.  We can send in a flags pointer or use the myslot flags, but both close and release should use the same mechanism.

sueloverso:
I should note that the statistics for `log: consolidated slot join races` has been split and has a new meaning.  This statistic used to mean anytime we couldn't join the slot for any reason and we yielded.  With this branch it only means we lost a race to update the atomic state.  I added a new statistic to account for being unable to join because the active slot is not open.

sueloverso:
autobuild: retest this please

michaelcahill:
Thanks @sueloverso, LGTM.

The statistics change should be fine: apart from this ticket, I don't think anyone outside our team has been looking closely at log statistics.  And using a flag rather than a `bool *` parameter makes perfect sense.  I figured out the build failure (it was an environment issue), so this change is ready to go.  Merging...

PR #3361, opened 2017-03-29
WT-3245 Avoid hangs on shutdown when a utility thread encounters an error

keithbostic:
@sueloverso, the log threads didn't panic on unexpected exit, the async worker threads checked for panic (unlike the other server threads).

Can you please review?


michaelcahill:
Thanks for the changes @keithbostic and the review @sueloverso.  I'll merge.

PR #3362, opened 2017-04-02
WT-3076 Add a general-purpose epoch manager

keithbostic:

* I created the structure and supporting functions, including the drain functions.
* I switched over table schema, split-generations, hazard pointers and checkpoint generation numbers, all of which were pretty easy. I did not switch over LSM trees or eviction generations (happy to take a run at them, but there are some subtleties I want to discuss with you or @agorrod first).
* I did not inline these functions, which might make a difference in some cases. (The problem is that `btree.h` calls the generation functions, and there's some header file trickiness in inlining functions called by include files.)
* I  used the name "generation" instead of "epoch", mostly because we already have `__wt_epoch` (which returns the seconds since the POSIX epoch), and that looked confusing. Happy to switch to "epoch" if you & Alex prefer, but I think that means renaming `__wt_epoch`. I don't care much, either way is fbm.
* The hazard pointer change: I set it up so `__wt_txn_release` and `__session_reset` free any session memory that can be freed, which means they'll identify hazard pointer arrays that can be free'd, rather than having any special support.


src/support/generation.c
```diff
@@ -0,0 +1,360 @@
...
+	return (session->generations[which]);
+}
+#endif
+
+/*
+ * __wt_session_gen_publish --
+ *	Publish a thread's resource generation.
+ */
+uint64_t
+__wt_session_gen_publish(WT_SESSION_IMPL *session, int which)
```
michaelcahill: I'd call this `__wt_gen_enter`, and I think it should loop until it sees that the published generation is still current.  I note that the split code is the only thing using it at the moment but I suspect there will be more cases (e.g., eviction).

src/support/generation.c
```diff
@@ -0,0 +1,360 @@
...
+	for (oldest = conn->generations[which] + 1,
+	    s = conn->sessions, i = 0; i < session_cnt; ++s, ++i) {
+		if (!s->active)
+			continue;
+
+		/* Ensure we only read the value once. */
+		WT_ORDERED_READ(v, s->generations[which]);
+		if (v != 0 && v < oldest)
+			oldest = v;
+	}
```
michaelcahill: One thing I'd like to talk through with this change is whether we can avoid this loop (and in fact any loop over sessions).

Ignoring performance issues for a second, imagine each generation number is hashed to a WT_RWLOCK.  Entering generation 100 would acquire a read lock on that generation's lock.

When some thread wants to be sure that everyone has left generation 100, it first increments the generation to 101, then waits for a write lock on the lock for generation 100.  This will be a huge win when there are 100K sessions open.

I think we can make the performance reasonable by having N locks per generation (where 8 <= N <= 32).  Each session entering a generation uses `session->id` to hash onto one of these locks, making conflicts between sessions very unlikely.  There is more work for the thread checking everyone has left a generation (it has to get N writelocks serially), but still much less than scanning all the sessions.

keithbostic:
@michaelcahill 
> You mentioned eviction: I don't think it makes sense to change the read generation to use this code (but also wouldn't object) -- that's really just a globally incrementing counter.

Agreed, by eviction I meant the `__wt_evict_file_exclusive_on/off` code.

src/support/generation.c
keithbostic: @michaelcahill 

> I'd call this `__wt_gen_enter`, and I think it should loop until it sees that the published generation is still current.  I note that the split code is the only thing using it at the moment but I suspect there will be more cases (e.g., eviction).

Fixed as part of commit 93f54f8.

src/include/schema.h
```diff
@@ -68,8 +68,8 @@ struct __wt_table {
 	bool cg_complete, idx_complete, is_simple;
 	u_int ncolgroups, nindices, nkey_columns;
 
-	uint32_t refcnt;	/* Number of open cursors */
-	uint32_t schema_gen;	/* Cached schema generation number */
+	uint32_t refcnt;		/* Number of open cursors */
+	uint64_t schema_generation;	/* Cached schema generation number */
```
sueloverso: Everything else still uses `_gen` so I'd prefer this stay with that name as well.

src/support/hazard.c
```diff
@@ -22,6 +22,7 @@ hazard_grow(WT_SESSION_IMPL *session)
 	WT_HAZARD *nhazard;
 	size_t size;
 	void *ohazard;
+	uint64_t generation;
```
sueloverso: Similarly, I'd prefer `hazard_gen`.

src/include/schema.h
keithbostic: Fixed in commit dfd6eea.

src/support/hazard.c
keithbostic: Fixed in commit dfd6eea.

keithbostic:
@michaelcahill 
> What I would like to do is add a generation to avoid __wt_evict_file_exclusive_on needing to interact with the eviction server.

I've stared at this some, and there are some subtleties (for example, the eviction server gives up on getting the handle list lock if it's "interrupted"). I'm happy to take a run at this, but I need you to sketch out how you think it should work, otherwise I'm not going to get the change correct.

keithbostic:
**Don't merge this branch, it's wrong.**

michaelcahill:
Thanks @keithbostic, this change is a good consolidation and looks right to me.  I'll merge, we can follow up in separate tickets on ideas for improving the implementation or using generations in other parts of the code.

PR #3364, opened 2017-04-03
WT-3250 Have one function initializing the WT portion of the spinlock.

sueloverso:
@agorrod and @michaelcahill Here are my suggested changes for the spin lock fix.

### DIFF
--- a/src/include/mutex.h
+++ b/src/include/mutex.h
@@ -62,31 +62,17 @@ union __wt_rwlock {			/* Read/write lock */
 #define	SPINLOCK_PTHREAD_MUTEX		2
 #define	SPINLOCK_PTHREAD_MUTEX_ADAPTIVE	3
 
-#if SPINLOCK_TYPE == SPINLOCK_GCC
-
 struct __wt_spinlock {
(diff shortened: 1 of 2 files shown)


sueloverso:
@agorrod The merge of the earlier branch caused `wiredtiger-test-spinlock` test to fail because the spinlock for `SPINLOCK_GCC` doesn't contain the `name` or `initialized` fields.  Do you recall why we don't want or have those there?  I have pushed a separate change to fix that by unifying the structures, but it seems quite intentional in code I haven't touched much.
@keithbostic Can you look at these changes?

sueloverso:
autobuild: retest this please

keithbostic:
> Do you recall why we don't want or have those there?

I don't think there's any reason those fields weren't there, other than they weren't used (for example, since gcc spinlocks can't ever fail, there wasn't any opportunity to use the name as part of an error message).

This approach looks good to me.

sueloverso:
autobuild: retest this please

agorrod:
Thanks @sueloverso - this change is a good improvement - sorry I merged the original change from underneath you.

PR #3365, opened 2017-04-04
WT-3249 Look at slot_state during force while holding lock.

sueloverso:
We could race an in-progress switch that set a new, empty active slot
but has not yet released the previously active slot and get an
incorrect LSN.


agorrod:
autobuild: retest this please

agorrod:
autobuild: retest this please

agorrod:
autobuild: retest this please

sueloverso:
@agorrod I pushed the fix for your question about the non-forced case and added a comment.  That was a good catch.

agorrod:
Thanks @sueloverso - this change looks good to me now.

@michaelcahill I'm going to go ahead and merge this change to get it into stress testing. Would you please do a review as well, since it's in subtle code.

PR #3367, opened 2017-04-04
WT-3247 Test should exit instead of abort to avoid a core dump.

sueloverso:
@michaelcahill Please review these changes.  I don't recall what Jenkins job prompted this.  However when I run `develop` I get a core file generated and with this change I do not, as expected and the test passes.  This test is not a randomly aborting test.


sueloverso:

michaelcahill:

PR #3368, opened 2017-04-05
WT-3253 Update Python transaction tests to have correct flush configuration

sueloverso:
@ddanderson Please review this changes to a few txn tests.  If we are only writing log records in memory we need to flush the log before making a backup to make sure all records are there before copying.  This bug has been in there a long time but recent changes widened a window.


agorrod:
autobuild: retest this please

ddanderson:
LGTM.

PR #3369, opened 2017-04-05
WT-3222 Add additional log statistics

sueloverso:
Not yet ready for review.  First pass at adding a few interesting stats.


agorrod:
@sueloverso the code changes in here look fine to me - do you think it's worth running the performance tests on the branch?

The other thing I noticed is that {{log_slot_joins}} is going away - I think we need to add an entry to {{src/dox/upgrading.dox}} to reflect that.

sueloverso:
There are two pairs of stats that were identical (just noticed I haven't committed the other) so two stats are going away.  I think we should chat about this because I definitely feel like I went overboard to try to consider anything that might be helpful and may have overshot the target.  Stats aren't free so we should consider if they're useful.  In any case I'll make the doc changes too.

sueloverso:
@agorrod This is ready for review after our discussion yesterday.  @bdlucas1 you may want to glance at this too.  Here are some points for consideration:
* I'm torn on including the change to add a sleep if we've yielded a bunch.  Both 1. if the change should happen at all and 2. if it should happen within this change or be completely separate.
* The number of join calls, which should be about the same as `log_writes`, is now the sum of `log_slot_immediate` and `log_slot_yield`.  Then there are various subsets of `log_slot_yield` where we record if we yielded due to atomic op race/collision or we found the slot closed.  The good news is that 90% of all join calls immediately join and never yield.

agorrod:
Thanks @sueloverso lgtm - I'm going to go ahead and merge.

PR #3371, opened 2017-04-05
WT-3133 Track operation duration

fedorova:
To allow us to investigate the cause of slow operations


sulabhM:
I merged develop into this branch and got it to compile.
The tests mostly fail because the code isn't setting the directory to create tracking ops files. As I get time from the other tickets I will fix that issue and do a review for the change.

fedorova:
@sulabhM, sorry for not testing for that prior to committing. I'll be away for a week starting this Friday, but I'll see if I can fix it while travelling. If not, back to this September 6.

sulabhM:
@fedorova, don't worry about it, I understand this branch is still being worked on. Enjoy your travels, I will take care of putting in a fix as part of the review. We can have a discussion about the change when you are back.

fedorova:
@sulabhM, the build currently fails on Windows. I think I know how to fix it, but since I don't have a Windows machine I did not want to put in the fix without the ability to test it. If you or anyone have access to a windows machine, I suggest trying the following. 

So the problem is that __asm__ directive used in src/include/misc.i is not liked by non-gcc compilers on Windows. Based on what I have read, the __asm will work fine. So we'd need to use

  `#ifdef _M_AMD64`

to detect if we are using the visual studio compiler and if so, use the __asm directive in place of __asm__.

I do not know if we support any other non-gcc compilers on Windows.

fedorova:
@sulabhM, I've made a bunch of changes to make the branch build properly on all systems, including  Windows, and pass the tests. At the moment it fails ex_all with the wiredtiger-pull-request-asan. That test passes on my system and with the linux build (I have not seen failures in other configurations yet either). I'm wrapping up for today. If you know how to set up that configuration and can figure out what's going on, that would be great! I don't yet know how to reproduce that test failure, so won't make any progress until I figure that out. Thanks!

fedorova:
@sulabhM, my bad: ex_all failed with the Linux build. I hope I'll be able to reproduce it on the Linux AWS machine. Hopefully once that's fixed, the wiredtiger-pull-request-asan configuration will pass as well.

fedorova:
@sulabhM, ex_all crashes on some system, because WT_CONNECTION_IMPL became much larger with optracking and we overflow the stack in some cases where we allocate WT_CONNECTION_IMPL on the stack. I am working on a fix.

sulabhM:
@fedorova , I started looking at it too, I will wait for your fix and take it from there. To reproduce failures from Jenkins jobs, I usually look at the job configuration and run the same steps as in the configuration. Let me know if you need help.

fedorova:
@sulabhM, I fixed the bug that made ex_all fail. There are more errors in other tests. I am making progress on those. Will keep you posted.

dist/api_data.py
```diff
@@ -774,6 +774,10 @@ def __cmp__(self, other):
         RPC server for primary processes and use RPC for secondary
         processes). <b>Not yet supported in WiredTiger</b>''',
         type='boolean'),
+    Config('optrack', 'true', r'''
```
agorrod: I'm inclined to set the default value to 'false' for this option since it generates so much information.

I also think we should move this configuration option into the `connection_runtime_config` section - so that the tracing can be enabled/disabled via a call to `WT_CONNECTION:reconfigure`. Maybe we should split that into a different ticket though?

src/conn/conn_api.c
```diff
@@ -1022,6 +1022,112 @@ err:	API_END_RET_NOTFOUND_MAP(session, ret);
...
+ * __conn_optrack_dir --
+ *	Set the directory for operation logging
+ */
+static int
+__conn_optrack_dir(WT_SESSION_IMPL *session, const char *cfg[])
+{
+	WT_CONFIG_ITEM cval;
+
+	/* Only use the environment variable if configured. */
+	WT_RET(__wt_config_gets(session, cfg, "use_environment", &cval));
```
agorrod: I'm inclined to change this, so the configuration option becomes `optrack=(path=,enabled=false)`. We went through an exercise a while ago of avoiding allowing WiredTiger users to create files in arbitrary locations on the system for security reasons, so using a `path` configuration option that has to be a subdirectory of the database is better here.
If a user wants to map that onto a different device then they have the option of pre-creating the directory as a symbolic link.

src/conn/conn_api.c
```diff
@@ -1022,6 +1022,112 @@ err:	API_END_RET_NOTFOUND_MAP(session, ret);
...
+	return (0);
+}
+
+/*
+ * A global variable that tells us whether or not we have created a
+ * map file for the current process. Map files, translating binary
+ * numbers into function names are valid per-process, so we have to
+ * set them once per process, not per connection or per session.
+ */
+static bool __wt_optrack_map_setup = 0;
```
agorrod: I'm confused by this. The WT_CONNECTION handle is single-process, so it should be OK to track the field there. I see that it's a recent response to a test failure so I'll dig deeper.

fedorova:
@agorrod, I saw that some tests failed after you merged the develop branch. I'll take a look at this next week, as I am signing off for the day.

src/conn/conn_api.c
fedorova: @agorrod, this must be an oversight on my part. I probably intended to move this to the connection structure, but forgot. No reason to have it as a static. I will move and retest.

agorrod:
Thanks - I've taken a quick look and there are two test cases that failed:
test_compact02 has been flaky in regular testing at the moment, and appears
to be sensitive to timing - I wonder if enabling the operation tracking
slows things down enough to cause the test to fail more regularly.
test_encrypt06 fails fairly reliably for me - it searches for string
matches in the database directory of an encrypted database, and finds one
in the optrack files. I suspect this is a genuine information leak when
operation tracking is enabled.

I'd expect both of those failures to go away if we switch the default
operation tracking to false, and I don't think they are worth further
investigation right now.

fedorova:
@agorrod, I got to the bottom of the "double free" bug in fops. Turned out to be memory corruption. I had assumed that a session equals a thread. That's not always true. There could be many threads using the same session. As a result there was a race that led to writing beyond the length of the buffer and corrupting the memory. Since the session!=thread scenario does not happen often, I am inclined to not put heavy synchronization into the very light operation tracking macro, but instead put safeguards to prevent memory corruption and let us lose a few log records if there is a race. I'll put the comment in the code, so you get a better idea of the worst-case scenario. 

As a result of chasing this bug the whole evening, I didn't get to fixing other errors. I'll get to them next week unless you get to them before me. I'll clean up the code and push the fix, hopefully before falling asleep today.

sueloverso:
@fedorova can you expand on the "many threads using the same session"?  Sessions are not thread-safe.  Different threads may use the same session in succession, serially, but there should not be multiple threads using the session at the same time.  If you are seeing the latter that is a bug in our test/app.  Where did you see that and should we investigate it?

src/conn/conn_api.c
```diff
@@ -1022,6 +1022,112 @@ err:	API_END_RET_NOTFOUND_MAP(session, ret);
...
+	/* Set up the directory for operation logs. */
+	WT_RET(__conn_optrack_dir(session, cfg));
+
+	/* Operation tracking files will include the ID
+	 * of the creating process in their name, so we
+	 * can distinguish between log files created by
+	 * different WiredTiger processes in the same
+	 * directory. We cache the process id for future use.
+	 */
+	conn->optrack_pid = __wt_process_id();
```
sulabhM: With operation tracking path being always relative to a wiredtiger directory would this be needed?
We wont be having two WT instances dumping to same directory, but a restart will change the id.

agorrod: I believe it is useful - for example if we used this feature with wtperf, this means we get a different set of files for the load and workload phase(s).

fedorova:
@sulabhM, @agorrod, I have implemented lazy creation of output files. I might have a few more hours on Friday, otherwise I'll pick up next Wednesday. If you have a suggestion on what to address next, please let me know before you leave for the week-end. Otherwise I'll pick up where Sulabh leaves off.

src/include/optrack.h
```diff
@@ -0,0 +1,73 @@
...
+	if (s->optrackbuf_ptr == WT_OPTRACK_MAXRECS) {			\
+		if (s->optrack_fh != NULL) {				\
+			size_t bytes = __wt_optrack_flush_buffer(s);	\
+			s->optrack_offset += bytes;			\
+		}                                                       \
+		s->optrackbuf_ptr = 0;					\
+	}
+
+/*
+ * We do not synchronize access to optrack buffer pointer under the assumption
```
sulabhM: A session is a thread context in WT. Sessions aren't thread safe, we can't have multiple threads using the same session. This assumption always holds good. Do we need to just update the comment, or is there more to it ?

fedorova: @sulabhM, there were multiple threads using the same session. That's why I had the race. It was a corner case, but it did happen. This assumption may not always hold true.

sueloverso: There definitely should not be multiple threads using the same session.  They can use the same session sequentially but not simultaneously.

fedorova: I hope I am wrong. I saw that in the fops test. It was only for session #1 and only at the very end of the execution. So could be the corner case. I will see if I can produce a patch with printfs that reproduce the situation.

fedorova: Ran into issues. Operation tracking was turned off by default on all tests. When I turned it back on in fops test to reproduce the issue, that test blows up due to "WT_SESSION.create: unknown configuration key: 'operation_tracking'". I'll investigate that first. Then I will be set up to try and reproduce the issue with two threads accessing session 1 simultaneously.

fedorova: @sueloverso, first of all, my apologies that I missed your question about threads using the same session that you posted two weeks ago. I just saw it now. Here is some more information on the issue:

1. I have not been able to reproduce it in the latest version of the branch; I will keep trying, because if the issue is there, we need to know about it.

2. I was able to reproduce it in the version of the branch as of commit 95d056a2bb8a3bd8f5fba27aef262e7cfc6c2241.

3. It is visible if we insert some printfs. Here is what I see:

16384, s 0, tid 139660021970688, begin track
16386, s 0, tid 139660136994560, begin track
16386, s 0, tid 139660136994560, end track
16385, s 0, tid 139660162172672, begin track
16387, s 0, tid 139660162172672, end track
16387, s 0, tid 139660136994560, begin track
16388, s 0, tid 139660136994560, end track
16388, s 0, tid 139660162172672, begin track
16388, s 0, tid 139660162172672, end track
16386, s 0, tid 139660021970688, end track

The first number is just an operation tracking buffer pointer; ignore it for the time being, it is being printed in order to help us find the place where the interleaving occurs. The second number is the session id ("s 0" means session 0). The third number is the thread ID obtained via pthread_self(). Threads print the "begin track" message when they enter an operation tracking macro and the "end track" message when they are done with that macro. The macro is just a few lines of non-reentrant code without any synchronization. If there is only one thread using a session at a time, we expect that any thread that prints a "begin track" message in a given session will print the corresponding "end track" message before any other thread using the same session prints the "begin track" message. That is NOT what we see in the snipped above.

What you notice instead is that thread with tid 139660021970688 prints the "begin track", then a bunch of other threads print "begin track" and "end track" messages, and then tid 139660021970688 prints its "end track" message. All of these threads are using session 0, so we see interleaving among multiple threads using the same session.

I observed that behaviour on Alex's AWS instance. To reproduce it:  

1. Check out wt-3133-track-long-ops branch.
2. Check out commit 95d056a2bb8a3bd8f5fba27aef262e7cfc6c2241.
3. Apply the attached patch.
4. Build.
5. Run the fops test: cd build_posix/test/fops; ./t
6. Wait for the test to blow up.
7. Scroll up to search for print statements, where the operation tracking buffer points (the first number printed) reaches 16384. Then look for interleaving among threads.

Note that the attached file is actually a diff. I had to give it the TXT extension, because otherwise GitHub won't upload it. 

I will work on reproducing the bug on the current branch. I won't spend too much time on it, because it is likely that the many merges from develop that happened since the old commit, where I am able to reproduce the bug, eliminated the bug, making this a moot point.

[many-threads-same-session-patch.txt](https://github.com/wiredtiger/wiredtiger/files/1415680/many-threads-same-session-patch.txt)

fedorova: @sueloverso, @sulabhM, bad news. The interleaving does occur in the latest version of the branch. Here is the output from the fops test with additional print statements (see my previous email for instructions on how to interpret the output):

16383, s 0, tid 139730455553792, begin track
16383, s 0, tid 139730184304384, begin track
1, s 0, tid 139730184304384, end track
1, s 0, tid 139730184304384, begin track
2, s 0, tid 139730184304384, end track
0, s 0, tid 139730455553792, end track

The catch here is that we see the 'begin track' message from thread 139730455553792, then a bunch of begin/end messages from thread 139730184304384, then the 'end track' message from thread 139730455553792.

If you apply the attached patch to the current version of the branch (commit baaf96b5cd9a068a35574100d3329fe9e66c7ec5), you should be able to catch the interleaving after a few tries. 

Instructions:

1. Check out wt-3133-track-long-ops branch.
2. Check out commit baaf96b5cd9a068a35574100d3329fe9e66c7ec5 (currently the head).
3. Apply the attached patch.
4. In test/fops/t.c change the operation tracking directory "/mnt/fast/sasha" to any directory you'd like. 
4. Build.
5. Run the fops test: cd build_posix/test/fops; ./t
6. The test will pass.  Look for interleaved begin/end statements among different threads using session 0.


[many-threads-same-session-interleaved.txt](https://github.com/wiredtiger/wiredtiger/files/1416439/many-threads-same-session-interleaved.txt)

src/include/optrack.h
```diff
@@ -0,0 +1,67 @@
...
+#define	WT_TRACK_OP(s, optype)						\
+	tr->timestamp = __wt_rdtsc(s);					\
+	tr->op_type = optype;						\
+	tr->op_id = (uint64_t)&__func__;				\
+									\
+	if (optype == 0 && !id_recorded)				\
+		__wt_optrack_record_funcid(s, tr->op_id,		\
+		    (void*)&__func__, sizeof(__func__), &id_recorded);	\
+	if (s->optrackbuf_ptr == WT_OPTRACK_MAXRECS) {			\
+		if (s->optrack_fh != NULL)				\
```
sulabhM: I doubt this condition will ever be true. __wt_optrack_flush_buffer() allocates the optrack_fh, and __wt_optrack_flush_buffer() gets called either at session close or here. And since optrack_fh will be NULL here, it will not call __wt_optrack_flush_buffer().

fedorova: Good catch! We used to allocate optrack_fh on session open, so it would typically not be NULL here. Then I changed to lazy allocation of operation tracking files. Now it will always be NULL. I will fix the logic.

sulabhM:
autobuild: retest this please

sulabhM:
autobuild: retest this please

sulabhM:
autobuild: retest this please

sulabhM:
autobuild: retest this please

michaelcahill:
Note that we're seeing an assertion fire in [MongoDB tests](https://logkeeper.mongodb.org/build/b6c7b5864a60f2c7fd2c18942012b335/test/5a2601c7c2ab68214111b775): 

```
[js_test:find] 2017-12-05T02:17:43.652+0000 d21011\| 2017-12-05T02:17:43.648+0000 E
STORAGE  [initandlisten] WiredTiger error (0) [1512440263:648093][24095:0x7f090e7e09a0],
wiredtiger_open: __wt_open, 250: lock_file || !LF_ISSET(WT_FS_OPEN_CREATE)
| [js_test:find] 2017-12-05T02:17:43.652+0000 d21011| 2017-12-05T02:17:43.648+0000 E
STORAGE  [initandlisten] WiredTiger error (0) [1512440263:648120][24095:0x7f090e7e09a0],
wiredtiger_open: aborting WiredTiger library
```

agorrod:
Thanks @fedorova I'm going to merge this - sorry for the extended delay..

PR #3372, opened 2017-04-06
WT-3265 Allow eviction of recently split pages when tree is locked.

michaelcahill:
When pages split in WiredTiger, internal pages cannot be evicted
immediately because there is a chance that a reader is still looking at
an index pointing to the page.  We check for this when considering pages
for eviction, and assert that we never evict an internal page in an
active generation.

However, if a page splits and then we try to get exclusive access to
the tree (e.g., to verify it), we could fail to evict the tree from
cache even though we have guaranteed exclusive access to it.

Relax the check on internal pages to allow eviction from trees that are
locked exclusive.

### DIFF
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -1354,8 +1354,13 @@ __wt_page_can_evict(
 	 * the original parent page's index, because evicting an internal page
 	 * discards its WT_REF array, and a thread traversing the original
 	 * parent page index might see a freed WT_REF.
+	 *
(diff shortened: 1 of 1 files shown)


michaelcahill:
@agorrod, I have not been able to reproduce the MongoDB failure (still trying), but I'm reasonably sure this change (a) is safe and (b) will avoid the assertion.

agorrod:
Thanks @michaelcahill the change makes sense to me, I'll go ahead and merge.

PR #3375, opened 2017-04-06
WT-3266 Drop lock while joining threads.

sueloverso:
@keithbostic or @agorrod Do you want to review this?

### DIFF
--- a/src/support/thread_group.c
+++ b/src/support/thread_group.c
@@ -23,7 +23,6 @@ __thread_run(void *arg)
 	session = thread->session;
 
 	for (;;) {
-		WT_ERR(WT_SESSION_CHECK_PANIC(session));
 		if (!F_ISSET(thread, WT_THREAD_RUN))
 			break;
 		if (!F_ISSET(thread, WT_THREAD_ACTIVE))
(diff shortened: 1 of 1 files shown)


sueloverso:
This fix has run more than 3x the iterations successfully (over 2000 so far) than the original repro (602).

src/support/thread_group.c
```diff
@@ -23,6 +23,7 @@ __thread_run(void *arg)
 	session = thread->session;
 
 	for (;;) {
+		WT_ERR(WT_SESSION_CHECK_PANIC(session));
```
keithbostic: I don't understand why we're checking panic here -- at best I don't think it makes anything better. There's some related discussion in WT-3245.

src/support/thread_group.c
```diff
@@ -92,8 +93,27 @@ __thread_group_shrink(
 		 */
```
keithbostic: This comment looks strange to me, preceding a bunch of code that has nothing to do with waking threads? (I realize it's in the original code, but figured I'd mention it.)

src/support/thread_group.c
sueloverso: If the system panics, I don't see anything that will break this loop.  However, I'm happy to let that be considered under WT-3245 with other threads too.

sueloverso:
I completed 10000 iterations of this branch with the CONFIG.

michaelcahill:
Thanks @sueloverso, this change looks right to me.  I'll merge.

PR #3376, opened 2017-04-07
WT-3262 Don't check if the cache is full when accessing metadata.

michaelcahill:
Also don't check for a full cache while holding the table lock (we're
likely reading the metadata in that case, just being extra careful).


agorrod:
Thanks @michaelcahill lgtm.

PR #3378, opened 2017-04-07
WT-3268 Failure to close cursor can get wiredtiger stuck in a cursor-close loop

keithbostic:
@sulabhM, this was a great catch!

Here's a possible solution for your review -- I found two related infinite loops and fixed them as well.

What makes this change a little messy is there wasn't a single change that matched the problems: I added a generic pair of macros that wrap the TAILQ walk, so if we hit the same element twice in the walk, we remove it from the queue, and used that in two cases. I think that approach is appropriate for walks where there are a variety of underlying functions that are expected to remove the element from the queue, where we won't know if the element was removed after they fail.

In a third case, I added a force flag to an underlying function so it always removes the queue element.

And, there was one restructuring that just cleaned up some messy code in a close function.

These are all separate commits, hopefully that makes the review easier.

### DIFF
--- a/src/include/misc.h
+++ b/src/include/misc.h
@@ -278,22 +278,20 @@ union __wt_rand_state {
 };
 
 /*
- * WT_TAILQ_SAFE_REMOVE_BEGIN/WT_TAILQ_SAFE_REMOVE_END --
- *	Macros to safely walk a TAILQ where we're expecting some underlying
+ * WT_TAILQ_SAFE_REMOVE --
+ *	Macro to safely walk a TAILQ where we're expecting some underlying
  * function to remove elements from the list, but we don't want to stop on
  * error, nor do we want an error to turn into an infinite loop. Used during
- * during shutdown, when we're shutting down various lists.
+ * shutdown, when we're shutting down various lists. Unlike TAILQ_FOREACH_SAFE,
+ * this macro works even when the next element gets removed along with the
+ * current one.
  */
-#define	WT_TAILQ_SAFE_REMOVE_BEGIN(head, elm, field) do {		\
-	void *__prev;							\
-	for (__prev = NULL; ((elm) = TAILQ_FIRST(head)) != NULL;) {	\
-		if (__prev == (elm)) {					\
+#define	WT_TAILQ_SAFE_REMOVE(var, head, field, tvar)			\
+	for ((tvar) = NULL; ((var) = TAILQ_FIRST(head)) != NULL;	\
+	    (tvar) = (var))						\
+		if ((tvar) == (var)) {					\
 			/* Leak the structure. */			\
-			TAILQ_REMOVE(head, (elm), field);		\
+			TAILQ_REMOVE(head, (var), field);		\
 			continue;					\
-		}							\
-		__prev = (elm);
+		} else
 
-#define	WT_TAILQ_SAFE_REMOVE_END					\
-	}								\
-	} while (0)
--- a/src/include/misc.h
+++ b/src/include/misc.h
@@ -276,3 +276,24 @@ union __wt_rand_state {
 		uint32_t w, z;
 	} x;
 };
+
+/*
+ * WT_TAILQ_SAFE_REMOVE_BEGIN/WT_TAILQ_SAFE_REMOVE_END --
+ *	Macros to safely walk a TAILQ where we're expecting some underlying
+ * function to remove elements from the list, but we don't want to stop on
+ * error, nor do we want an error to turn into an infinite loop. Used during
+ * during shutdown, when we're shutting down various lists.
+ */
+#define	WT_TAILQ_SAFE_REMOVE_BEGIN(head, elm, field) do {		\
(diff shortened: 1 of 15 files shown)


sulabhM:
@keithbostic ,
For your consideration, I pushed in a change to the macro so that it's usage looks similar to the TAILQ_MACROS. When I thought about the solution to this issue, I was planning to use the standard TAILQ_FOREACH_SAFE, but realised that cursor close can end up removing multiple cursors (and hence invalidating the temp next one) from the queue. So I have put that comment in there for anyone who wonders why we can't use the standard TAILQ_FOREACH_SAFE.

sulabhM:
I am still reviewing the other parts of the change

src/include/misc.h
```diff
@@ -278,22 +278,20 @@ union __wt_rand_state {
...
+	for ((tvar) = NULL; ((var) = TAILQ_FIRST(head)) != NULL;	\
+	    (tvar) = (var))						\
+		if ((tvar) == (var)) {					\
 			/* Leak the structure. */			\
-			TAILQ_REMOVE(head, (elm), field);		\
+			TAILQ_REMOVE(head, (var), field);		\
 			continue;					\
-		}							\
-		__prev = (elm);
+		} else
```
keithbostic: @sulabhM, I agree with your approach.

The "cursor removes more than one element" problem didn't occur to me -- the reason I didn't use TAILQ_FOREACH_SAFE was I wanted to leave the linked list empty. If we've already "closed" a structure, leaving it on the linked list isn't going to make anything better, and some other thread of control might somehow decide to use it.

The hidden `else` is a little disconcerting, but I can't think of any reason it's bad.

I don't think we need to put `var` between parentheses in the TAILQ_REMOVE call (I got it wrong originally).

src/session/session_api.c
```diff
@@ -215,7 +215,7 @@ __session_close(WT_SESSION *wt_session, const char *config)
 			WT_TRET(session->event_handler->handle_close(
 			    session->event_handler, wt_session, cursor));
 		WT_TRET(cursor->close(cursor));
-	} WT_TAILQ_SAFE_REMOVE_END;
+	}
```
keithbostic: Yeah, the `else` bugs me, it feels scary (but I can't think of a bug it would introduce, either). Can you run it by Alex or Michael for a another opinion? Happy to go along with what you and they decide.

src/include/misc.h
```diff
@@ -276,3 +276,21 @@ union __wt_rand_state {
...
+ * current one.
+ */
+#define	WT_TAILQ_SAFE_REMOVE(var, head, field, tvar)			\
+	for ((tvar) = NULL; ((var) = TAILQ_FIRST(head)) != NULL;	\
+	    (tvar) = (var))						\
+		if ((tvar) == (var)) {					\
+			/* Leak the structure. */			\
+			TAILQ_REMOVE(head, (var), field);		\
+			continue;					\
+		} else
```
sueloverso: I glanced at this diff today.  I will admit that this naked else really makes me nervous about introducing subtle bugs in the future.

michaelcahill: @sueloverso and @keithbostic I hear you about the dangling `else`.  I don't think it's actually more dangerous that a macro that just includes a `for` without a body, but it's certainly not obvious.

@sulabhM and I kicked this around before he pushed that change -- I could live with BEGIN / END macros, but having `BEGIN { ... } END` seems messy (the braces aren't necessary but code won't autoformat correctly without them).  That said, we have exactly one other case of this, `WT_INTL_FOREACH_BEGIN { ... } END`, and it hasn't caused trouble.

TL;DR we should go back to the way this was and I'll ignore the unnecessary braces.

michaelcahill:
@sulabhM, other places where we should use the new macro:

`async.c:462` -- the loop to free `WT_ASYNC_FORMAT` objects
`async_worker.c:344` -- async cursors
`lsm_tree.c:141` -- closing LSM trees
`os_fhandle.c`:364 -- closing file handles (even though `__handle_close` is currently structured so the remove will always happen, don't rely on that in the loop)
`os_fs_inmemory.c:523` -- closing in-memory handles
`schema_list.c:249` -- closing tables

If we're not going to use this for all of the `TAILQ_FIRST` loops in `conn_api.c`, we should at least move the explicit `TAILQ_REMOVE` to the top of each loop so they are consistent.

The loops at `lsm_manager.c:525` and `lsm_manager.c:537` would be better as `TAILQ_FOREACH_SAFE`.  Similarly for `session_dhandle.c:403`.

sueloverso:
The direction of these changes looks better to me.  Once the other locations Michael noted are pushed, I'll look one more time.

sulabhM:
@michaelcahill : I reviewed all the places I found TAILQ_FIRST in a loop and have cleaned them up. Can you please review this change.

michaelcahill:
Thanks @keithbostic and @sulabhM -- this LGTM.  I'll merge.

PR #3379, opened 2017-04-10
WT-3271 Prevent integer overflow in eviction tuning.

michaelcahill:


michaelcahill:
I thought about going further here and changing the types of all thread counts to `int32_t` -- I have that change if anyone is interested, but I think fixing the bug (including potentially backporting) is more important.

agorrod:
Thanks @michaelcahill lgtm - I'm going to go ahead and merge. @fedorova please take a look as well, and let me know if you see any problems.

PR #3380, opened 2017-04-10
WT-3197 aarch64 CRC32C support fails to compile on non-linux ARM platforms

keithbostic:
getauxval is a Linux system call, don't attempt to call it unless we're running on a Linux system.


agorrod:
Thanks @keithbostic lgtm - I'm happy for you to merge this once it's passed automated testing.

PR #3381, opened 2017-04-10
WT-3197 aarch64 CRC32C support fails to compile on non-linux ARM platforms

keithbostic:
First change was insufficient, need to test __linux__ when including header files.


keithbostic:
autobuild: retest this please

PR #3384, opened 2017-04-12
WT-3263 Allow recovery log file archiving on clean shutdown.

sueloverso:
Revert test changes that needed a checkpoint for archiving.

@agorrod This is ready for review.

### DIFF
--- a/src/txn/txn_recover.c
+++ b/src/txn/txn_recover.c
@@ -535,6 +535,8 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
 	 * this is not a read-only connection.
 	 * We can consider skipping it in the future.
 	 */
+	if (needs_rec)
+		FLD_SET(conn->log_flags, WT_CONN_LOG_RECOVER_DIRTY);
 	if (WT_IS_INIT_LSN(&r.ckpt_lsn))
 		WT_ERR(__wt_log_scan(session, NULL,
 		    WT_LOGSCAN_FIRST | WT_LOGSCAN_RECOVER,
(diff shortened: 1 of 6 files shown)


sueloverso:
autobuild: retest this please

agorrod:
autobuild: retest this please

src/txn/txn_recover.c
```diff
@@ -556,7 +558,8 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
 	 */
 	WT_ERR(session->iface.checkpoint(&session->iface, "force=1"));
 
-done:	FLD_SET(conn->log_flags, WT_CONN_LOG_RECOVER_DONE);
+done:	FLD_SET(conn->log_flags,
+	    WT_CONN_LOG_RECOVER_DIRTY | WT_CONN_LOG_RECOVER_DONE);
```
agorrod: It looks to me as though this will mean the `WT_CONN_LOG_RECOVER_DIRTY` flag will always be set because of this line - am I missing something?

sueloverso: We only set it above if we detect `needs_rec` (i.e. we detect an unclean shutdown), but recovery is single-threaded and I simply unconditionally clear it.  It seemed messier in the code to conditionalize that clear.

agorrod: I'm still missing it. Do we need the new call to `FLD_SET` in `__wt_txn_recover` in case `__wt_log_scan` or `checkpoint` return an error, and this latter FLD_SET for other cases? It might be faster for us to talk on the phone next week.

sueloverso: Thanks for the push back!  I didn't read closely enough and thought it was a FLD_CLR.  I have fixed this and added a clear statement.

agorrod:
Thanks @sueloverso this lgtm

PR #3387, opened 2017-04-14
WT-3281 stress test sanitizer failure

keithbostic:


src/btree/bt_cursor.c
```diff
@@ -682,22 +662,24 @@ __wt_btcur_insert(WT_CURSOR_BTREE *cbt)
...
-		 * use the pinned page, but that's an unlikely path.) Re-save
-		 * the cursor state: we may retry but eventually fail.
+		 * The pinned page goes away if we fail for any reason, ensure
+		 * there's a local copy of any key or value. (Restart could
+		 * still use the pinned page, but that's an unlikely path.)
+		 * Re-save the cursor state: we may retry but eventually fail.
 		 */
-		WT_TRET(__cursor_copy_int_key(cursor));
+		WT_TRET(__cursor_localkey(cursor));
+		WT_TRET(__cursor_localvalue(cursor));
```
michaelcahill: This case seems fine to me: it will be very unusual to have an internal value here, so normal cases won't involve any copying.

src/btree/bt_cursor.c
```diff
@@ -432,9 +409,10 @@ __wt_btcur_search(WT_CURSOR_BTREE *cbt)
 
 	/*
 	 * The pinned page goes away if we do a search, make sure there's a
-	 * local copy of any key, then re-save the cursor state.
+	 * local copy of any key or value, then re-save the cursor state.
 	 */
-	WT_ERR(__cursor_copy_int_key(cursor));
+	WT_ERR(__cursor_localkey(cursor));
+	WT_ERR(__cursor_localvalue(cursor));
```
keithbostic: I had not thought about the middle ground of only leaving external references untouched on error, that makes sense to me.

I've changed the cursor functions that called `__cursor_localvalue()` to call `__cursor_novalue()` instead (that's cursor `search`, `search-near`, `insert`, `insert-check` and `update`).

@michaelcahill, you didn't flag all of those functions, and I agree it's not performance-critical in functions like `cursor.insert`, but it's easier to make the functions look the same than it is to put in a comment why they're different. Let me know if I'm missing something...

We don't document what cursor key/value references are left untouched on error.

Also, we're inconsistent in the Btree code, and likely more inconsistent in higher-level stacks like LSM.

(The example I'm thinking about in Btree is `cursor.remove`, we currently clear any external cursor value reference on failure. I can't see why that would be a problem, but there you have it.)

Should we document this semantic? I could add a sentence like:
```
On failure, any previous cursor position is lost. Any previously set cursor key or
value that references application-memory (in other words, keys or values set by
the application separately from previous cursor operations), are left unchanged.
```
If we document the semantic, I'll have to fix `cursor.remove` and should add tests to `test_scope` to verify we get it right.

I don't feel strongly either way, it's hard to imagine application-writers upset that `cursor.remove` didn't leave their cursor value set.

Let me know if we should document this semantic, and, in that case, if anybody has better wording.

src/btree/bt_cursor.c
```diff
@@ -516,10 +494,11 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
...
 
 	/*
-	 * The pinned page goes away if we do a search, make sure there's a
-	 * local copy of any key, then re-save the cursor state.
+	 * Any pinned page goes away if we do a search, make sure there's a
+	 * local copy of any key or value, then re-save the cursor state.
 	 */
-	WT_ERR(__cursor_copy_int_key(cursor));
+	WT_ERR(__cursor_localkey(cursor));
+	WT_ERR(__cursor_localvalue(cursor));
```
keithbostic: Fixed.

src/btree/bt_cursor.c
keithbostic: Fixed anyway.

michaelcahill:

PR #3388, opened 2017-04-14
WT-3275 Fix a bug in LSM cursor semantics.

agorrod:
If an update, remove or reserve operation required a search it was possible for a race with eviction to cause a segfault between the lookup and requested operation.

### DIFF
--- a/src/lsm/lsm_cursor.c
+++ b/src/lsm/lsm_cursor.c
@@ -1530,6 +1530,11 @@ __clsm_insert(WT_CURSOR *cursor)
 	WT_ERR(__cursor_needvalue(cursor));
 	WT_ERR(__clsm_enter(clsm, false, true));
 
+	/*
+	 * It isn't necessary to copy the key out after the lookup in this
+	 * case because any non-failed lookup results in an error, and a
+	 * failed lookup leaves the original key intact.
+	 */
 	if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE) &&
 	    (ret = __clsm_lookup(clsm, &value)) != WT_NOTFOUND) {
 		if (ret == 0)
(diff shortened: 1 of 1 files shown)


src/lsm/lsm_cursor.c
```diff
@@ -1574,8 +1579,14 @@ __clsm_update(WT_CURSOR *cursor)
 	WT_ERR(__cursor_needvalue(cursor));
 	WT_ERR(__clsm_enter(clsm, false, true));
 
-	if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE))
+	if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE)) {
 		WT_ERR(__clsm_lookup(clsm, &value));
+		/*
```
agorrod: I thought about pushing the "save a key" semantic into `__clsm_lookup`, but that seemed more confusing than this change. I'm happy to switch to that approach if it's better than the duplication in this change.

agorrod:
Another approach would have been to move the call to `__clsm_reset_cursors` in `__clsm_put` until after the insert, but that means cursors would be pinned across the insert operation which isn't ideal.

agorrod:
@keithbostic This is a superset of the change you suggested in WT-3275. I'd appreciate if you can take a look.

michaelcahill:
Thanks @agorrod, I'm inclined to take the fix and worry about performance issues later if they show up in testing.  At some point I'd like to discuss `reserve` in LSM trees -- I have a feeling that there's something wrong there (i.e., if the key we're reserving doesn't exist in the primary) but I can't put my finger on it.

agorrod:
> At some point I'd like to discuss reserve in LSM trees

I agree and I've opened WT-3290 to explore.

PR #3389, opened 2017-04-15
WT-3284 tree-walk restart bug

keithbostic:


michaelcahill:
Thanks @keithbostic, splitting up page skipping to keep the restart handling sane makes sense to me.  This change looks good, I'll merge.

PR #3390, opened 2017-04-15
WT-3286 Improve search if an index hint is wrong

keithbostic:


agorrod:
Thanks @keithbostic - this change looks good and complete to me. I'm going to hold off merging until I can touch base with Ramon to make sure I don't cause JIRA problems when merging.

PR #3391, opened 2017-04-15
WT-3287 review WiredTiger internal panic checks

keithbostic:
Here's a set of changes based on error-handling discussions @sueloverso, @ddanderson, @agorrod  and I had in NYC last week.

In summary, we added several panic checks inside logging loops to avoid hangs in `wt2909_checkpoint_integrity` child processes, after fault injection. The change is to remove those panic checks, and to instead immediately exit the child process when a panic happens.

There are 3 commits:

* da99e9d changes the standard error WT_EVENT_HANDLER to exit with a non-zero value on a WT_PANIC error: if the best thing an application can do on a panic is exit and restart, then it's reasonable for the standard error handler to exit. (To be fair, I think this change pushes the envelope some. On the balance, I like the change, but I won't argue if the consensus is the change is a bad idea.) If we don't take the change, we need to make a change in `test/csuite/wt2909_checkpoint_integrity/main.c`, giving the test program an error handler that exits on panic.

* aa67cf1 is a minor cleanup in `include/api.h`, it's largely unrelated to the point of this pull request.

* 71016d0 removes the `WT_SESSION_CHECK_PANIC()` calls we added to keep the `wt2909_checkpoint_integrity` test from hanging (as well as a couple of additional ones).

@ddanderson, my understanding from you is that `wt2909_checkpoint_integrity` is fine with child programs that simply exit, and that's the path I've taken.

### DIFF
--- a/dist/s_void
+++ b/dist/s_void
@@ -88,8 +88,8 @@ func_ok()
 	    -e '/int handle_progress$/d' \
 	    -e '/int helium_cursor_reset$/d' \
 	    -e '/int helium_session_verify$/d' \
-	    -e '/int index_compare_primary$/d' \
 	    -e '/int index_compare_S$/d' \
+	    -e '/int index_compare_primary$/d' \
 	    -e '/int index_compare_u$/d' \
 	    -e '/int index_extractor_u$/d' \
 	    -e '/int log_print_err$/d' \
@@ -103,7 +103,6 @@ func_ok()
 	    -e '/int nop_pre_size$/d' \
 	    -e '/int nop_sizing$/d' \
 	    -e '/int nop_terminate$/d' \
-	    -e '/int nop_terminate$/d' \
 	    -e '/int os_errno$/d' \
 	    -e '/int revint_terminate$/d' \
 	    -e '/int rotn_error$/d' \
@@ -111,6 +110,7 @@ func_ok()
 	    -e '/int rotn_terminate$/d' \
 	    -e '/int snappy_pre_size$/d' \
 	    -e '/int snappy_terminate$/d' \
+	    -e '/int subtest_error_handler$/d' \
 	    -e '/int uri2name$/d' \
 	    -e '/int usage$/d' \
 	    -e '/int util_err$/d' \
--- a/src/async/async_worker.c
+++ b/src/async/async_worker.c
@@ -57,7 +57,6 @@ retry:
 			return (0);
 		if (!F_ISSET(conn, WT_CONN_SERVER_ASYNC))
 			return (0);
-		WT_RET(WT_SESSION_CHECK_PANIC(session));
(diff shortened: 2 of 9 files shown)


keithbostic:
If we don't take da99e9d, here's an (untested) change that I think works:
```
diff --git a/test/csuite/wt2909_checkpoint_integrity/main.c b/test/csuite/wt2909_checkpoint_integrity/main.c
index 478373564..2a6614a6a 100644
--- a/test/csuite/wt2909_checkpoint_integrity/main.c
+++ b/test/csuite/wt2909_checkpoint_integrity/main.c
@@ -445,6 +445,31 @@ run_process(TEST_OPTS *opts, const char *prog, char *argv[], int *status)
 }
 
 /*
+ * subtest_error_handler --
+ *     Error event handler.
+ */
+static int
+subtest_error_handler(WT_EVENT_HANDLER *handler,
+    WT_SESSION *session, int error, const char *message)
+{
+       (void)(handler);
+       (void)(session);
+       (void)(message);
+
+       /* Exit on panic, there's no checking to be done. */
+       if (error == WT_PANIC)
+               exit (0);
+       return (0);
+}
+
+static WT_EVENT_HANDLER event_handler = {
+       subtest_error_handler,
+       NULL,   /* Message handler */
+       NULL,   /* Progress handler */
+       NULL    /* Close handler */
+};
+
+/*
  * subtest_main --
  *     The main program for the subtest
  */
@@ -478,7 +503,8 @@ subtest_main(int argc, char *argv[], bool close_test)
            WT_FAIL_FS_LIB
            "=(early_load,config={environment=true,verbose=true})]"));
 
-       testutil_check(wiredtiger_open(opts->home, NULL, config, &opts->conn));
+       testutil_check(wiredtiger_open(
+           opts->home, &event_handler, config, &opts->conn));
        testutil_check(
            opts->conn->open_session(opts->conn, NULL, NULL, &session));
```

ddanderson:
@keithbostic, as for the untested change, I would change `exit(0)` to `exit(1)`.  I've tested this change as well and it works (without da99e9d).

For future reference (because I forgot too), the way to test this is:
```
$ cd build_posix/test/csuite
$ TESTUTIL_ENABLE_LONG_TESTS=1 ./test_wt2909_checkpoint_integrity -v
```
It takes a few minutes.

keithbostic:
@ddanderson, thank you for running that test -- I've gone ahead and pushed that change into place.

agorrod:
@keithbostic This lgtm now. Should I go ahead and merge, or were you thinking of using this as the basis for more changes?

keithbostic:
@agorrod 
> Should I go ahead and merge, or were you thinking of using this as the basis for more changes?

This is done as far as I'm concerned.

PR #3392, opened 2017-04-17
WT-3288 fix error codes for event_handler to be consistent in file operations

halex2005:
I used WiredTiger library as dependency.
I have updated wiredtiger v2.7.0 => v2.9.1 and found an issue when I call `__wt_open_session` with directory which doesn't exists: 

- `__wt_open_session` return code is correct = 2 (ENOENT)
- `event_handler` receive errorCode = 0, and `wiredtiger_strerror` return `Successful` string for this code

This PR is intended to return consistency to error codes, now `event_handler`'s error code and return code will be consistent for file operations.

### DIFF
--- a/src/os_win/os_fs.c
+++ b/src/os_win/os_fs.c
@@ -54,11 +54,12 @@ __win_fs_remove(WT_FILE_SYSTEM *file_system,
 	WT_RET(__wt_to_utf16_string(session, name, &name_wide));
 
 	if (DeleteFileW(name_wide->data) == FALSE) {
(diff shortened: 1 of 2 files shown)


wtbuild:

keithbostic:
Thank you, @halex2005! I'll be looking at your change as soon as I can.

michaelcahill:
autobuild: ok to test

keithbostic:
Replaced by pull-request #3393, closing.

PR #3394, opened 2017-04-18
SERVER-28835 Coverity analysis defect 101330: Resource leak

keithbostic:
Discard any allocated thread structure on error.


agorrod:
Thanks @keithbostic lgtm

PR #3395, opened 2017-04-18
WT-3292 review/cleanup full-barrier calls in WiredTiger

keithbostic:
@michaelcahill, I reviewed the places where we use full-barriers after our discussions last week, intending to remove the places I'd been (unnecessarily) using them for visibility.

There's only one "real" fix in this branch, commit 13b9f16. It ensures calls to `__wt_sleep` are barriers (similarly to the barrier we added in `__wt_yield`).

Feel free to discard this branch without further discussion if you don't want it. I incrementally made these changes while convincing myself things were correct, and most of the changes involve fixing comments or other similar cleanups.

@sueloverso, I've included the logging change we discussed.

### DIFF
--- a/src/session/session_compact.c
+++ b/src/session/session_compact.c
@@ -226,7 +226,7 @@ __compact_checkpoint(WT_SESSION_IMPL *session)
 	 */
 	txn_global = &S2C(session)->txn_global;
 	for (txn_gen = __wt_gen(session, WT_GEN_CHECKPOINT);;) {
-		WT_READ_BARRIER();
+		/* Loop checks volatile-declared objects, no barriers needed. */
 		if (!txn_global->checkpoint_running ||
 		    txn_gen != __wt_gen(session, WT_GEN_CHECKPOINT))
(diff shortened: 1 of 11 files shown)


src/session/session_compact.c
```diff
@@ -226,7 +226,7 @@ __compact_checkpoint(WT_SESSION_IMPL *session)
 	 */
 	txn_global = &S2C(session)->txn_global;
 	for (txn_gen = __wt_gen(session, WT_GEN_CHECKPOINT);;) {
-		WT_READ_BARRIER();
+		/* Loop checks volatile-declared objects, no barriers needed. */
```
sueloverso: I had to read this comment several times to parse it.   Can we expand the text a bit?  Something like:
* This loop only checks objects that are declared volatile.  Therefore no barriers are needed.

keithbostic: @sueloverso, fixed in 7aba20c.

michaelcahill:

PR #3396, opened 2017-04-19
WT-3282 Split cache flags and cache_pool flags.

sueloverso:
@agorrod Here's the branch for your review.  All the tests are still running.  This is ready to merge after they have run for another day or two.  But you can see the changes now.

I decided to split the flags into a separate field rather than use the {{flags_atomic}} macros for this fix.  The cache pool flags are rarely used/modified and I didn't want to impose atomic operations on every eviction use of the flags.


agorrod:
Thanks @sueloverso I understand this fix, but it makes me sad. If we need this change, I think we also need to review any place we use flags on shared data structures and make sure they aren't relying on state transitions for correctness.

sueloverso:
@agorrod all the tests are still running fine (3 instances each on ocelot and margay).  Assuming all is well at the end of the day today (which will be over 48 hours of running), I'll merge.

PR #3397, opened 2017-04-19
WT-3296 LAS table fixes/improvements

keithbostic:
@michaelcahill, @agorrod, some changes to fix problems in the LAS table code.

Michael, there's a change in `__rec_txn_read()` that I'd appreciate a close review on.


src/reconcile/rec_write.c
```diff
@@ -573,16 +572,29 @@ __rec_write_check_complete(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+		if (r->update_mem_skipped == 0)
+			return (EBUSY);
+
+		/*
+		 * Don't rewrite pages where we're not going to get back enough
+		 * memory to care. There's no empirical evidence the 10KB limit
+		 * is a good configuration, but it should keep us from wasting
+		 * time on tiny pages and pages with only a few updates.
+		 */
+		if (r->update_mem - r->update_mem_skipped < 10 * WT_KILOBYTE)
```
michaelcahill: I'm uncomfortable about the fixed 10KB in here (as I was uncomfortable when the previous "10% of updates must be discarded" check went in).  It's not hard to imagine workloads where every page would have less than 10KB saving so we wouldn't make progress (e.g., in-memory workloads with small updates to lots of pages).

Would this achieve the same goal as the fixed 10KB check:

```c
if (r->update_mem_skipped > 9 * r->update_mem / 10)
    return (EBUSY);
```

?

More generally, eviction tracks whether the something has changed since the last time we tried eviction on a page.  What is the evidence for having reconciliation override that by giving up here?

michaelcahill:
The check on rewrites made me review what we actually do during an update/restore, and as far as I can tell it's binary: either the most recent update for some key is globally visible, in which case we can put it on the page and discard the history, or we keep the *previous* on-page value (however old) and all of the history.

We could do better: if there is a globally visible update in the list, we could put that on page and discard everything after, including the previous on-page value.

Here is a step in that direction: at least throw away anything obsolete as part of eviction:

```
diff --git a/src/reconcile/rec_write.c b/src/reconcile/rec_write.c
index a667a28..7cb833e 100644
@@ -1127,6 +1127,14 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 	} else
 		upd_list = ins->upd;
 
+	/* Discard obsolete updates if evicting. */
+	if (F_ISSET(r, WT_EVICTING)) {
+		WT_UPDATE *obsolete = __wt_update_obsolete_check(
+		    session, page, upd_list->next);
+		if (obsolete != NULL)
+			__wt_update_obsolete_free(session, page, obsolete);
+	}
+
 	++r->update_cnt;
 	for (skipped = false,
 	    max_txn = WT_TXN_NONE, min_txn = UINT64_MAX,
```

Going further would mean returning the first stable update in `updp` from `__rec_txn_read` rather than clearing `*updp` at line 1289.  It would also invalidate the `update_mem_skipped` calculation, which is why I'm discussing this here rather than opening a new ticket...

michaelcahill:
I don't want this change to wait for my return -- I think it's basically correct but as noted, I'm concerned about a fixed magic number of 10KB in savings below which reconciliation gives up.

@agorrod, if you and @keithbostic agree on this change, please go ahead and merge without waiting for me.  The discussion about improving update/restore to discard obsolete updates probably deserves a new ticket (unless I've misunderstood what's going on in which case ignore me).

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

src/reconcile/rec_write.c
keithbostic: I don't much like the 10KB limit, either, but in heavy insert workloads under cache stress, I saw lots of small pages being evicted, pages only a few K in size, with only a few update chains.

```c
if (r->update_mem_skipped > 9 * r->update_mem / 10)
    return (EBUSY);
```

This won't catch pages with only a couple of update chains we had to skip. Imagine one update chain of 100 bytes we had to skip, `100 > (9 * 100 / 10) == true`.

We don't want rewrite pages in memory if we're not going to get anything back. We could drop the limit to 1-2KB? Is it worth doing any work at all to recover 2KB?

keithbostic:
@michaelcahill 
> either the most recent update for some key is globally visible, in which case we can put it on the page and discard the history, or we keep the previous on-page value (however old) and all of the history.

I believe that's correct.

> We could do better: if there is a globally visible update in the list, we could put that on page and discard everything after, including the previous on-page value.

I definitely want the change to discard obsolete updates, and I've pushed it in d8f2aa2.

Replacing the on-page element with a globally visible update from the list, is (as you noted), trickier. We'd also have to remove it from the update list too, otherwise we'd attempt to re-apply it when rewriting the page.

keithbostic:
autobuild: retest this please

src/reconcile/rec_write.c
agorrod: @keithbostic I believe this change would cause test/format to hang with cache full more regularly.

keithbostic: @agorrod, happy to reduce the limit, but at some point it makes no sense to rewrite a page in memory (which is a ton of work).

keithbostic:
> This change generally looks good to me, but I'm suspicious about stress testing fallout. I'm going to switch stress testing onto this branch for a while before merging.

Thanks, @agorrod. You're right, it never occurred to me, for example, that we've never tested `reserve` updates with the LAS table, because LAS has been busted for awhile. I've seen two distinct failures so far, I'm chasing.

michaelcahill:
This change looks okay to me: I still don't love the arbitrary 2KB cutoff for trying rewrites, but with the inMemory storage engine out of the picture I can't construct a sensible case where it would matter.  The tip of the branch has been through many rounds of stress testing without hitting any failures.  I'll merge.

PR #3398, opened 2017-04-19
WT-3293 Don't explicitly mark internal symbols hidden.

agorrod:
It messes with external stack decoders.

The details of this change are open for discussion. I've left the explicit `visibility("default")` tags in the code, since that is most of the hard work - though if we don't test with `visibility("hidden")` as the default via `-fvisibility=hidden` it's likely our set of marked functions will get out of date over time.

I'm going to switch the performance runs over to this branch, to see if there is a noticeable difference.


agorrod:
The perf branches didn't show any variation when running with this change.

michaelcahill:

PR #3401, opened 2017-04-24
WT-3300 Coverity 1374542: Dereference after null check

keithbostic:
False positive, but explicitly checking the stop variable should make the complaint go away.


agorrod:
Thanks @keithbostic lgtm

PR #3402, opened 2017-04-24
SERVER-28820 Add a few error path messages.

sueloverso:
@agorrod Please review this change.  It adds a few error messages for some paths in `log_scan` to know why we got `WT_NOTFOUND` including the path in the ticket.


agorrod:
@sueloverso I generally really like this change... I believe this changes the semantic of log cursors. In the past they would return WT_NOTFOUND, which is dealt with as a non-error case. Is that an expected/desired change?

sueloverso:
@agorrod That is fair enough.  I modified it so that we only give the error message if coming in via recovery.  A user will just get the non-error WT_NOTFOUND but recovery should only be passing in valid LSNs that it finds in the meta-data.

agorrod:
Thanks @sueloverso lgtm.

PR #3403, opened 2017-04-26
WT-3289 update thread group alloc size only at successful thread creation

sulabhM:


agorrod:
@sulabhM the change looks reasonable to me, but I'm not going to do a review because of the test failure.

sulabhM:
WT-3302 takes care of fixing wt-3289, so closing this one

PR #3405, opened 2017-04-27
WT-3302 Introduce a thread struct with a member telling if thread created successfully

sulabhM:


sulabhM:
@michaelcahill , for your consideration.

agorrod:
Thanks @sulabhM lgtm - please go ahead and merge this when you are ready (I'm not going to do it, in case you aren't ready).

PR #3408, opened 2017-04-28
WT-3142 Add a workload generator application

agorrod:
The workload generator runs core workloads in C++, which directly calls the C API and thus minimizes the overhead of the test program. The workload setup is done in Python to allow for an expressive workload definition language.

Initial commit has a number of example workloads that match equivalent wtperf workloads.


ddanderson:
@agorrod, the last 3 commits should resolve 3 of the current build complaints:
  

- evergreen build should be fixed by https://github.com/wiredtiger/wiredtiger/pull/3408/commits/5480a3aef1dfb95303639dbc6116116dd10976e4 (compilation warnings)
- wiredtiger-test-spinlock should be fixed by https://github.com/wiredtiger/wiredtiger/pull/3408/commits/4d749c121c197e87062eeeb3cc37272b26e673ca (stray test target in Makefile.am)
- wiredtiger-configure-combinations should be fixed by https://github.com/wiredtiger/wiredtiger/pull/3408/commits/f156e257d223a2d436de6724a300e64d2de7c71e (config tests, building in .../build instead of build_posix)

As for wiredtiger-test-unit-ppc, I think the problem is that there is no C++ compiler installed on the machine.  If you look at http://build.wiredtiger.com:8080/job/wiredtiger-test-unit-ppc/623/consoleFull, in the configure output, you see
```
checking for c++... no
checking for g++... no
```
I guess we just never installed a C++ compiler on ppc?  I think this is our first real use of C++, so we probably need to put it in as a prerequisite for a complete build (is there some doc that needs to be updated?)  There probably is a way to fail early (in configure) if there's no c++.

agorrod:
@ddanderson Sorry I left you hanging on this. I'll test the changes here tomorrow.

Re: C++ compiler support - I think it would be good to skip building the workload generator if there is no C++ compiler.

@keithbostic Have you done any work on managing the compilers available on the PPC test machine?

keithbostic:
@agorrod 
> Have you done any work on managing the compilers available on the PPC test machine?

No, I have not -- I have an action item to upgrade it to clang 4.0, but wasn't planning a "real" installation.

agorrod:
@ddanderson I've tested the failed Jenkins tests except for the C++ failure. I'm working on getting C++ installed on the PPC machine, I think it'd be good to add conditional compilation into configure in the mean time - do you feel comfortable doing that work?

ddanderson:
@agorrod, sure I'll add the conditional compilation stuff for C++ into configure.

agorrod:
Thanks @ddanderson - this lgtm, I'll merge when it's a convenient time to monitor testing afterward.

PR #3411, opened 2017-05-05
WT-3158 backport test

daveh86:


daveh86:
@agorrod mind taking a look and seeing if we want to merge these? It may be good to have @keithbostic have a look also, as he wrote the originals.

michaelcahill:
@daveh86 I triggered [a run of the Windows perf job](http://build.wiredtiger.com:8080/job/mongodb-perf-ycsb-windows/1051/) against this branch and it does appear to fix the regression.

Do you understand the root cause (i.e., why perf went down and why these changes bring it back up again)?

daveh86:
@michaelcahill, no I do not. I found that this resolved the performance regression by brute force. I tested all the likely commits between the mongodb-3.6 branch and the tip of develop. 

It is entirely possible that this is an unrelated performance improvement, so I think my next action would be to test the commits here vs the known good branch and see what the result is

agorrod:
@daveh86 That sounds like a good plan of action for Monday

daveh86:
It's taken me a while, but I now have an answer as to why this fix works, it increases the size of the session structure. I've tested by simply adding a void* pad to the session structure and this counters the performance regression.

daveh86:
I'm closing this in favour of https://github.com/wiredtiger/wiredtiger/pull/3414

PR #3413, opened 2017-05-08
WT-3312 encryption btree configuration test

keithbostic:


agorrod:
@sueloverso This one is more in your area of expertise - it makes sense to me - do you think there is value in explicitly calling out the metadata table?

sueloverso:
This LGTM.  Keith and I talked about this change beforehand.  The special case is unnecessary because the metadata should take the other half of the conditional anyways.  (It came about because I asked if the LAS file needed something similar.)

PR #3414, opened 2017-05-11
WT-3158 Add a void* padding variable to WT_SESSION_IMPL

daveh86:


michaelcahill:
Closing, the fix for the underlying issue is in #3416, we will backport that after it is merged into develop.

PR #3415, opened 2017-05-11
WT-3158 Fix structure padding on Windows.

michaelcahill:


keithbostic:
@michaelcahill, I tossed a test into dist to check for an even number of `WT_PACKED_STRUCT_BEGIN/END` lines, not sure if we can do something better?

Nice catch... it's been busted since January 2015!

michaelcahill:
This change fixed the performance regression.  It has been superseded by #3416, which applies to develop (where the perf regression was inadvertently fixed by a change that grew `WT_SESSION_IMPL`).

PR #3416, opened 2017-05-11
WT-3158 Fix structure layout on Windows.

michaelcahill:
We use a pragma on Windows to force a struct to be packed, but were
missing the "end" pragma that restores normal layout.  The result was
that most structs were being packed, leading to poor performance for
workloads (particularly when accessing session structures).


michaelcahill:
@keithbostic, I moved the check into `s_style`: turns out we had an obsolete check there anyway and we can do a better job of explaining where the bug is...

PR #3417, opened 2017-05-12
WT-3158 Fix structure layout on Windows.

keithbostic:
@michaelcahill, the last merge breaks on FreeBSD, the shell/wc combination leaves white space in the `cnt` variable and `expr` complains.


michaelcahill:

PR #3419, opened 2017-05-12
WT-3264 Fix log file cleanup when toggling logging.

sueloverso:
Not yet ready for review.

### DIFF
--- a/src/txn/txn_recover.c
+++ b/src/txn/txn_recover.c
@@ -20,6 +20,7 @@ typedef struct {
 	} *files;
 	size_t file_alloc;		/* Allocated size of files array. */
 	u_int max_fileid;		/* Maximum file ID seen. */
+	WT_LSN max_lsn;			/* Maximum checkpoint LSN seen. */
 	u_int nfiles;			/* Number of files in the metadata. */
 
 	WT_LSN ckpt_lsn;		/* Start LSN for main recovery loop. */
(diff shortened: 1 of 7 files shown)


sueloverso:
@michaelcahill and @agorrod this change is ready for review.  With it WT will detect logging getting toggled on back and forth and if the log files have been removed it will start logging at a log file larger than the largest checkpoint LSN we find in the meta-data.  I have a change that I'm preparing for mongoDB to remove all the log files if they're turning logging off so that we don't continually do this toggling dance forever.

src/txn/txn_recover.c
```diff
@@ -443,9 +449,24 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
...
+		 */
 		WT_ERR(__recovery_file_scan(&r));
 		conn->next_file_id = r.max_fileid;
-		goto done;
+
+		if (FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED) &&
+		    WT_IS_MAX_LSN(&metafile->ckpt_lsn) &&
+		    !WT_IS_MAX_LSN(&r.max_lsn)) {
+			WT_ERR(__wt_log_reset(session, r.max_lsn.l.file));
+			goto ckp;
```
agorrod: nit: could you make this ``ckpt``?

sueloverso: Just to make sure I'm interpreting that correctly, do you mean you'd like `max_lsn` to be `max_ckpt_lsn`?

sueloverso: Oh, I think you meant `goto ckp` to be `goto ckpt`.

agorrod: It was the ``goto``

PR #3423, opened 2017-05-17
WT-3313 Replace the deprecated LZ4_compress call with one to LZ4_comp…

daveh86:
…ress_default


daveh86:
@keithbostic would you mind taking a look at this? I've changed our LZ4 implementation slightly as the LZ4_compress call is now marked deprecated

michaelcahill:

PR #3424, opened 2017-05-17
WT-3327 If timediff detects backwards time, return 0.

sueloverso:
@agorrod and @michaelcahill please review this change.  It is a small change that if timediff sees backward time, it returns 0.  I considered adding a statistic to note that but that got messy pretty quickly (because those macros do not have a session, and that added multiple statements to something used as an expression).  If you look at the ticket I could semi-reliably get checkpoint to hang, and with this fix, it never did (admittedly with a fairly small number of runs in both cases).

This can be merged except I have not do an in-depth code review for where we use timestamps in general.

### DIFF
--- a/src/include/os.h
+++ b/src/include/os.h
@@ -55,21 +55,22 @@
 	}								\
 } while (0)
 
+#define	WT_TIMECMP(t1, t2)						\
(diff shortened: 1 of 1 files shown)


michaelcahill:
Closing in favor of #3427.

PR #3425, opened 2017-05-18
WT-3169 Add verbose messages regarding lookaside file usage

daveh86:
This adds the new "lookaside" verbose option and WT_VERB_LOOKASIDE along with 2 new verbose messages.


daveh86:
@keithbostic this was the commit I mentioned to you yesterday evening.

Can you take a look at this and confirm that the output matches what you and Alex discussed?

src/reconcile/rec_write.c
```diff
@@ -6591,3 +6594,29 @@ __rec_dictionary_lookup(
...
+	__wt_eviction_dirty_needed(session, &pct_dirty);
+	wt_session->transaction_pinned_range(wt_session, &pinned_range);
+
+	__wt_verbose(session, WT_VERB_LOOKASIDE,
+	    "Lookaside written to. Number of TXN ID's Pinned: %" PRIu64
+	    ". Cache is %" PRIu32 "%% dirty. "
+	    "Cache is %" PRIu32 "%% full. "
+	    "Page entries requiring LAS support: %" PRId64,
+	    pinned_range, pct_dirty, pct_full, inserted_cnt);
+}
```
keithbostic: Style comments: the first line of  the explanatory text for the function is indented (don't ask me why, just the way it is), and, we sort fixed size objects in the arg list (uint32_t, uint64_t), by size, so uint64_t comes first (again, just the way it is), `WT_SESSION* wt_session;` should be `WT_SESSION *wt_session;` (whitespace before the `*`).

keithbostic: `__wt_verbose_lookaside_write()` can be static to this file.

I'd suggest doing something like we have in the `__wt_verbose()` function, there's an `#ifdef` for `HAVE_VERBOSE` in the called function, and we would test `WT_VERBOSE_ISSET` in `__wt_verbose_lookaside_write()`, instead of in the caller.

keithbostic: Can we make the message read naturally, instead of being so choppy? Off the top of my head, something like:
```
__wt_verbose(session, WT_VERB_LOOKASIDE,
    "Page reconciliation required %" PRId64 " lookaside entries: %" 
    PRIu64 " pinned TXN ID's, cache is %" PRIu32 "/%" PRIu32 "%% dirty/clean",
    inserted_cnt, pinned_range, pct_dirty, pct_full);
```
Gets rid of "LAS", too. :)

keithbostic: I don't think there's any reason for `insert_cnt` to be signed, looks like my mistake in the original code. Can you please fix that?

keithbostic: We'll have to `(void)` the returns from `__wt_eviction_clean/dirty_needed`, they return a boolean.

michaelcahill:
I'm okay with this code change but please see my query [in jira](https://jira.mongodb.org/browse/WT-3169?focusedCommentId=1575131&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1575131) about the goals for this change.

src/reconcile/rec_write.c
```diff
@@ -6591,3 +6594,37 @@ __rec_dictionary_lookup(
...
 	return (0);
 }
+
+/*
+ * __verbose_lookaside_write --
+ *	Create a verbose message to display with details about the transaction
+ * state when performing a lookaside table write.
+ */
+static void
+__verbose_lookaside_write(WT_SESSION_IMPL *session, uint64_t inserted_cnt)
```
agorrod: The usual protocol is for static functions to have names that match the subsystem. All other static functions in this file start with a ``__rec_`` prefix - why is this one different?

Why did you call the parameter ``inserted_cnt`` when the variable in the only calling function is ``insert_cnt``?

src/reconcile/rec_write.c
```diff
@@ -6591,3 +6594,37 @@ __rec_dictionary_lookup(
...
+	uint64_t pinned_range;
+	uint32_t pct_dirty, pct_full;
+
+	if (WT_VERBOSE_ISSET(session, WT_VERB_LOOKASIDE)) {
+
+		wt_session = &session->iface;
+
+		(void)__wt_eviction_clean_needed(session, &pct_full);
+		(void)__wt_eviction_dirty_needed(session, &pct_dirty);
+		wt_session->transaction_pinned_range(wt_session, &pinned_range);
```
agorrod: It's not normal for internal calls to redirect through a session handle. When that happens a number of fields get updated in the WT_SESSION structure. The general solution here would be to change ``__session_transaction_pinned_range`` to ``__wt__session_transaction_pinned_range`` and make it non-static.

src/reconcile/rec_write.c
```diff
@@ -6591,3 +6594,37 @@ __rec_dictionary_lookup(
...
+static void
+__verbose_lookaside_write(WT_SESSION_IMPL *session, uint64_t inserted_cnt)
+{
+#ifdef HAVE_VERBOSE
+	WT_SESSION *wt_session;
+	uint64_t pinned_range;
+	uint32_t pct_dirty, pct_full;
+
+	if (WT_VERBOSE_ISSET(session, WT_VERB_LOOKASIDE)) {
+
```
agorrod: We wouldn't normally have a blank line here.

src/reconcile/rec_write.c
```diff
@@ -6591,3 +6594,37 @@ __rec_dictionary_lookup(
...
+		(void)__wt_eviction_clean_needed(session, &pct_full);
+		(void)__wt_eviction_dirty_needed(session, &pct_dirty);
+		wt_session->transaction_pinned_range(wt_session, &pinned_range);
+
+		__wt_verbose(session, WT_VERB_LOOKASIDE,
+		    "Page reconciliation required %" PRIu64
+		    " lookaside entries: %" PRIu64
+		    " pinned TXN ID's, cache is %" PRIu32
+		    "/%" PRIu32 "%% dirty/clean",
+		    inserted_cnt, pinned_range, pct_dirty, pct_full);
```
agorrod: The formatting looks inconsistent to me here, I think it'll come out as:

```
Page reconciliation required 99 lookaside entries: 99 pinned TXN ID's, cache is 99/99% dirty/clean
```

It'd be better as:

```
Page reconciliation triggered lookaside write. Entries written: 99, pinned transaction IDs: 99, cache dirty: 99%, cache clean: 99%
```

That uses common separators between terms. There are also other grammar corrections.

agorrod:
@daveh86 I like the suggestions from @michaelcahill to:
* Log at most one of each type of message per checkpoint
* Rename the option to reflect that this is a sampling of events.

Please make those changes

src/reconcile/rec_write.c
```diff
@@ -6591,3 +6594,37 @@ __rec_dictionary_lookup(
 	*dpp = next;
 	return (0);
 }
+
+/*
+ * __verbose_lookaside_write --
+ *	Create a verbose message to display with details about the transaction
+ * state when performing a lookaside table write.
```
daveh86: Can you elaborate on what is wrong with that indentation? I'm not certain what I'm missing.

daveh86:
Putting a hold on this review for now while I address the larger changes suggested.

src/reconcile/rec_write.c
agorrod: The examples I found of multi-line function level comments re-indent to the same level, for example:
https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn_ckpt.c#L1104

Having looked a bit further, it seems we aren't consistent, some places do that and others don't.

daveh86:
@michaelcahill can you take a look at this one?

michaelcahill:
Thanks @daveh86, with a minor comment change to match the signed type change for `record_cnt`, lgtm.  I'll merge.

PR #3426, opened 2017-05-18
WT-3303 Deadlock during first access to lookaside table

keithbostic:
Don't acquire the eviction walk-lock when releasing exclusive eviction access to a file, it can deadlock. Instead, atomically decrement the counter.


michaelcahill:

PR #3427, opened 2017-05-18
WT-3327 Check for backward time in wt_epoch

sueloverso:
@agorrod and @michaelcahill here's a version of fixing the time problems in `wt_epoch` instead of in `WT_TIMEDIFF`.  @keithbostic you might want to look at it too.  I prefer this version over the one I created yesterday in `WT_TIMEDIFF` for a few reasons:
* It allows us to add in a statistic easily to detect and record this situation.
* It is equally as compact and localized for backporting.
* It deals with it at the source rather than in the use of timestamps.

By keeping it in the session, users of `wt_epoch` with a NULL session do not get the benefit of this change but that is only `wtperf` and other local test programs so I'm okay with that.


michaelcahill:

PR #3428, opened 2017-05-19
WT-3258 Add timers tracking time spent on failed evictions of large p…

daveh86:
…ages


daveh86:
@agorrod would you mind having a look at this?

dist/stat_data.py
```diff
@@ -207,6 +207,7 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_eviction_force', 'pages evicted because they exceeded the in-memory maximum'),
     CacheStat('cache_eviction_force_delete', 'pages evicted because they had chains of deleted items'),
     CacheStat('cache_eviction_force_fail', 'failed eviction of pages that exceeded the in-memory maximum'),
+    CacheStat('cache_eviction_force_fail_time', 'time spent on failed eviction of pages that exceeded the in-memory maximum (usecs)'),
```
agorrod: It's better when the descriptions for pairs of stats have matching prefixes, so they sort together when viewed in tools.

src/evict/evict_page.c
```diff
@@ -86,8 +88,12 @@ __wt_page_release_evict(WT_SESSION_IMPL *session, WT_REF *ref)
 			 */
 			WT_STAT_CONN_INCR(
 			    session, cache_eviction_force_delete);
```
agorrod: Looking at how this falls out, I think it makes sense to add time tracking for ``cache_eviction_force`` and ``cache_eviction_force_delete`` as well. Maybe it could be structured as:

```
/* Describe why using non-standard error handling */
ret = __wt_evict(session, ref, false);
__wt_epoch(session, &stop);
if (ret == 0) {
    if (too_big)
        ....
} else {
    ....
}
```

That's a bit of an odd ordering with the handling of error returns, so I think it would deserve a comment.

daveh86:
Thanks @agorrod. Those changes make a lot of sense. Would you be able to take another look?

src/evict/evict_page.c
```diff
@@ -75,19 +77,37 @@ __wt_page_release_evict(WT_SESSION_IMPL *session, WT_REF *ref)
 	(void)__wt_atomic_addv32(&btree->evict_busy, 1);
 
 	too_big = page->memory_footprint >= btree->splitmempage;
-	if ((ret = __wt_evict(session, ref, false)) == 0) {
-		if (too_big)
+
+	/*
+	 * We want to track the timing of how long the call to evict took. The
+	 * reason for forced evict depends on the size of the page's size. If
```
agorrod: `the size of the page's size`, and generally I think this comment could be made shorter while conveying the necessary information - could you take a stab at that.

daveh86:
@agorrod , I've made what I hope is the final change. Can you have a last look?

agorrod:
Thanks @daveh86 this change lgtm now. There is a conflict against develop in an auto-generated file. Could you please update the branch and then merge.

PR #3429, opened 2017-05-20
WT-2972 partial updates: fix Python to allow a no-op modify.

ddanderson:
To transmit the length of the WT_MODIFY array from the typemap to the call site in SWIG, we put the length into an (otherwise unused) first element.  Changed from using a blank entry as a sentinel, as a blank can legally appear in an list of Modify elements.


ddanderson:
@keithbostic, I think this fixes the problem you're seeing with no-op Modify entries in Python.

keithbostic:
Thank you, @ddanderson, I've merged those changes in.

PR #3430, opened 2017-05-22
WT-3331 Get time into a local variable so we can read and use a consistent time

sueloverso:
@agorrod Here's a fix for the LSM failure seen over the weekend.  The new clock check was reading the location passed in by the caller and that was getting set/reset out from under it.  I now read the time into a local variable to do all of the new time checks.

It seems that it could be dangerous to have timestamps being freely overwritten and its possible we could get parts of two timestamps set into a field.  My new checks, reading the same location over and over, was seeing different values each time and was storing a value into the session that was not one it read.

### DIFF
--- a/src/os_posix/os_time.c
+++ b/src/os_posix/os_time.c
@@ -50,7 +50,7 @@ __wt_epoch(WT_SESSION_IMPL *session, struct timespec *tsp)
 		tmp.tv_sec = v.tv_sec;
 		tmp.tv_nsec = v.tv_usec * WT_THOUSAND;
 		__wt_time_check_monotonic(session, &tmp);
-		*tsp->tv_sec = tmp;
+		*tsp = tmp;
 		return;
 	}
 	WT_PANIC_MSG(session, ret, "gettimeofday");
--- a/src/os_posix/os_time.c
+++ b/src/os_posix/os_time.c
@@ -31,14 +31,14 @@ __wt_epoch(WT_SESSION_IMPL *session, struct timespec *tsp)
 	/*
 	 * Read into a local variable so that we're comparing the correct
 	 * value when we check for monotonic increasing time.  There are
-	 * many places we read into an unlocked global varible.
+	 * many places we read into an unlocked global variable.
 	 */
 #if defined(HAVE_CLOCK_GETTIME)
 	WT_SYSCALL_RETRY(clock_gettime(CLOCK_REALTIME, &tmp), ret);
 	if (ret == 0) {
+		__wt_time_check_monotonic(session, &tmp);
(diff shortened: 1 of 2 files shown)


src/os_posix/os_time.c
```diff
@@ -27,10 +28,17 @@ __wt_epoch(WT_SESSION_IMPL *session, struct timespec *tsp)
...
+	 * many places we read into an unlocked global varible.
+	 */
 #if defined(HAVE_CLOCK_GETTIME)
-	WT_SYSCALL_RETRY(clock_gettime(CLOCK_REALTIME, tsp), ret);
+	WT_SYSCALL_RETRY(clock_gettime(CLOCK_REALTIME, &tmp), ret);
 	if (ret == 0) {
-		__wt_time_check_monotonic(session, tsp);
+		tsp->tv_sec = tmp.tv_sec;
+		tsp->tv_nsec = tmp.tv_nsec;
+		__wt_time_check_monotonic(session, &tmp);
```
agorrod: Doesn't this change mean the check in monotonic isn't working any more, since it'll set the ``tmp`` variable to the older timestamp instead of ``tsp`` that is being returned to the user?

Torn writes into a timestamp that is shared (unlocked) across multiple threads sounds worrying. Should we use a memcpy from a local variable instead?

sueloverso: Yes, good catch.  I have pushed a fix.

I agree that the shared fields are worrying, but I added a WT_DIAGNOSTIC_YIELD call in between the setting of the two fields and the CONFIG ran a long time, until I killed it.

src/os_posix/os_time.c
```diff
@@ -39,9 +47,10 @@ __wt_epoch(WT_SESSION_IMPL *session, struct timespec *tsp)
 
 	WT_SYSCALL_RETRY(gettimeofday(&v, NULL), ret);
 	if (ret == 0) {
-		tsp->tv_sec = v.tv_sec;
-		tsp->tv_nsec = v.tv_usec * WT_THOUSAND;
-		__wt_time_check_monotonic(session, tsp);
+		tmp.tv_sec = v.tv_sec;
+		tmp.tv_nsec = v.tv_usec * WT_THOUSAND;
+		__wt_time_check_monotonic(session, &tmp);
+		*tsp->tv_sec = tmp;
```
michaelcahill: This doesn't look right -- I'd expect `*tsp = tmp;`

Do we build and test anywhere with `gettimeofday` but not `clock_gettime`?

sueloverso: Yikes, you are correct.  That was a cut-paste error. Fixed

agorrod:

PR #3433, opened 2017-05-24
WT-3326 Added wtperf.py script to run wtperf files.

ddanderson:
@agorrod, this could be merged or not, your choice.  There is plenty more potential work to be done on this ticket, but it may be quite useful as is.  wtperf.py currently generates valid Python code for 17 out of 59 of wtperf/runners/*.wtperf.  The others fail gracefully, complaining about wtperf directives that we don't yet support.  I've run a few of the good ones, they run successfully.  I think the next step should be to pick one or two Jenkins jobs that currently use wtperf that we could run in parallel and get that set up.

### DIFF
--- a/bench/workgen/runner/runner/__init__.py
+++ b/bench/workgen/runner/runner/__init__.py
@@ -88,5 +88,5 @@ except:
 shutil.rmtree('WT_TEST', True)
 os.mkdir('WT_TEST')
 
-from .core import txn, extensions_config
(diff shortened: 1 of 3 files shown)


michaelcahill:
@ddanderson I really like that the changes here are all Python.  @agorrod I haven't done a close review, let me know if you'd like me to take this one.

agorrod:
Thanks @ddanderson looks good to me

PR #3436, opened 2017-05-24
WT-3339 The CURSOR_UPDATE_API_CALL macro will dump core on a NULL btree handle

keithbostic:
Split CURSOR_UPDATE_API_CALL into two versions, one of which expects a btree handle, one of which doesn't.


agorrod:
Thanks @keithbostic lgtm.

PR #3437, opened 2017-05-25
WT-3338 Implement optimized cursor modify.

keithbostic:


keithbostic:
@agorrod, @michaelcahill: I believe the partial-update pull request branch is now feature complete, that is, all of the code we'll need to fast-path btree partial updates is included. It's not debugged yet, but this set of changes should give you a good idea of the shape of the change if you want to do a review to ensure we're all on the same page.

keithbostic:
@agorrod, @sueloverso, @michaelcahill, I think PR #3437 is ready for review.

Sue, I probably don't require a full review from you (although I'd certainly welcome one!) Regardless, would you please take a look at my changes to logging and recovery? A few specific notes:
* your approval of the (unformatted) log record I'm creating. Should there be any kind of a mode flag, or can we just rev the log version if this record changes?
* the assumption that not finding a record during recovery can be safely ignored because it implies the record was subsequently deleted.
* is there any additional testing you want?

Alex, I think either you or Michael will own a general review (there are lots of interesting changes to `WT_UPDATE` handling). And maybe @sulabhM should look at reconciliation? Regardless, I'll leave that all up to you.

agorrod:
Thanks @keithbostic I've switched a few of the Jenkins stress jobs over to this branch, and will get started on reviewing.

agorrod:
There was a stress test failure in Jenkins on this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/15494/console

keithbostic:
@agorrod, I've pushed a fix for last night's stress test and address sanitizer failures (I think both were the same underlying problem).

This one is a little nasty: first, it's another problem with the already merged changes in WT-3402 (PR #3493) -- that merge broke variable-length column-store overflow values with RLEs greater than 1.

The details, from the fix I pushed in fa123ff:

> Imagine an overflow value with an RLE greater than 1. We append a copy to the end of an update chain, but it's possible it's the overflow value for more than one record, and appending it to the end of one record's update chain means a subsequent enter of a globally visible value to one of the records would allow the truncation of the overflow chain that leaves other records without a value.
> 
> The fix it that, if appending such an overflow record, set the transaction ID to the first possible transaction ID. That ID is old enough to be globally visible, but we can use it as a flag if an update record cannot be discarded when truncating an update chain.

Obviously, this means that for variable-length column-store overflow records with RLEs greater than 1, update chains can get longer than we'd like, because they might be impossible to truncate. The previous implementation eventually discarded such overflow records, so this is a step backward.

On the other hand, this is only a problem for variable-length column-store overflow records with RLEs greater than 1 (and I still think this implementation is an improvement for everything else, that is, row-store), plus, I never did figure out how to make the previous implementation work for `WT_CURSOR.modify`, both of which make me want to keep moving forward.

If these update lists are ever a problem, I can see some approaches that would allow us to truncate them (generation numbers, or reference counting), but I don't plan to go there now.

Let me know if you have any objections to my (misuse) of `WT_TXN_FIRST` in this way, especially in the brave new world of timestamp transaction IDs. As far as I can tell, it's not an error to have both a transaction ID and a timestamp, but I wanted to call it out.

cc: @michaelcahill

src/reconcile/rec_write.c
```diff
@@ -294,6 +294,15 @@ typedef struct {
 	bool cache_write_restore;	/* Used update/restoration */
 
 	uint32_t tested_ref_state;	/* Debugging information */
+
+	/*
+	 * XXX
+	 * In the case of a modified update, we may need a copy of the current
+	 * value as a set of bytes. We call back into the btree code using a
+	 * fake cursor to do that work. This a layering violation and fragile,
+	 * we need a better solution.
```
agorrod: I agree we need a better solution, but none jumps to mind.

keithbostic:
> I'm happy for you to merge it if you like, or to wait for another round of review from Sue or Michael next week.

I'm inclined to wait on review from Sue & Michael, unless there's a push to get it merged.

agorrod:
Sounds good to me. No rush that I'm aware of.

src/cursor/cur_file.c
```diff
@@ -513,6 +547,15 @@ __curfile_create(WT_SESSION_IMPL *session,
...
 	__wt_btcur_open(cbt);
 
+	/*
+	 * WT_CURSOR.modify supported on 'u' value formats, but the fast-path
+	 * through the btree code requires log file format changes, it's not
+	 * available in all versions.
+	 */
+	if (WT_STREQ(cursor->value_format, "u") &&
+	    S2C(session)->compat_major >= WT_LOG_V2)
+		cursor->modify = __curfile_modify;
```
sueloverso: Is this the only line needed to support compatibility versions?  That's amazing.

test/csuite/Makefile.am
```diff
@@ -57,6 +57,9 @@ noinst_PROGRAMS += test_wt3135_search_near_collator
 test_wt3184_dup_index_collator_SOURCES = wt3184_dup_index_collator/main.c
 noinst_PROGRAMS += test_wt3184_dup_index_collator
 
+test_wt3338_partial_update_SOURCES = wt3338_partial_update/main.c
+noinst_PROGRAMS += test_wt3338_partial_update
+
```
sueloverso: I looked at all the tests below and nothing specifically tests modify with logging enabled that I can tell.  The csuite program and the python program never use logging AFAICT.  Am I missing it somewhere?  Also, it would be good to have a new test section in `test_compat01.py` or even a new `test_compat02.py` that tests modify and compatibility, confirming that the new record never ends up in the old version log file.

src/txn/txn_log.c
```diff
@@ -78,31 +77,47 @@ __txn_op_log(WT_SESSION_IMPL *session,
...
+	 * caller doesn't log reserve operations, we shouldn't see them here.
 	 */
-	WT_ASSERT(session, upd->type != WT_UPDATE_RESERVED);
 	if (cbt->btree->type == BTREE_ROW) {
 #ifdef HAVE_DIAGNOSTIC
 		__txn_op_log_row_key_check(session, cbt);
 #endif
-		if (upd->type == WT_UPDATE_DELETED)
+		switch (upd->type) {
+		case WT_UPDATE_STANDARD:
```
keithbostic: > Or did you put STANDARD first because it is expected to be the most common?

It's hold-over from when we didn't pass a `WT_UPDATE` type into the row-/column-store modification routines, so 0 was the "default" type.

That's no longer the case, and we could sort all of the places we use `upd->type`, but I think it makes the code a little less readable, having the common case first helps a little.

I doubt there's any performance reason for having the `WT_UPDATE_STANDARD` first, since the type value is 0, 1, 2, or 3, it's going to be compiled as a jump table, I think.

tl;dr: it's a holdover from an old implementation, let me know if you want it changed.

src/txn/txn_recover.c
```diff
@@ -121,6 +121,20 @@ __txn_op_apply(
...
 	switch (optype) {
+	case WT_LOGOP_COL_MODIFY:
+		WT_ERR(__wt_logop_col_modify_unpack(session, pp, end,
+		    &fileid, &recno, &value));
+		GET_RECOVERY_CURSOR(session, r, lsnp, fileid, &cursor);
+		cursor->set_key(cursor, recno);
+		if ((ret = cursor->search(cursor)) != 0)
+			WT_ERR_NOTFOUND_OK(ret);
+		else {
+			WT_ERR(__wt_modify_apply(
```
keithbostic: IIRC, I was thinking there might be upgrade/downgrade implications involving reading/writing log records for the right version of the log file, but even though I convinced myself that wasn't possible, this way I didn't have even think about it.

I'll add a comment.

src/cursor/cur_file.c
keithbostic: You tell me.... :)

test/csuite/Makefile.am
keithbostic: > and nothing specifically tests modify with logging enabled that I can tell.

`test/format` tests modify with logging enabled.

> Also, it would be good to have a new test section in test_compat01.py or even a new test_compat02.py that tests modify and compatibility, confirming that the new record never ends up in the old version log file.

That would probably have to wait on the resolution of WT-3417.

Should we instead add upgrade/downgrade support to `test/format`?

src/txn/txn_log.c
sueloverso: Yeah, I'd prefer them alphabetical, thanks.

test/csuite/Makefile.am
sueloverso: There already is a 'compat' option to `test/format`.  Hmm, so it is probably actually generating invalid log files as there likely are transactions in flight when the version changes.  It would be interesting to run printlog over the logs if logging, compatibility and modify are all going on in test/format.

I also don't think WT-3417 is needed for a unit test.  That is only needed in the case where there can be outstanding transactions in progress when the compatibility version changes.  It would be easy to construct a test that doesn't do that.  (And I definitely agree we need a test that does have it and specifically tests the contents of log files.)

src/txn/txn_log.c
keithbostic: > Yeah, I'd prefer them alphabetical

Pushed in b38c837 for your review.

keithbostic:
@agorrod, I added the assert you and I discussed in e0af4ab, the lookaside table should never include a reserved update.

PR #3438, opened 2017-05-25
WT-2972: In Python API for WT_CURSOR::modify, WT_NOTFOUND should not raise an exception.

ddanderson:


keithbostic:
I'm pulling this one into my local branch, there's no need to merge it into develop.

PR #3441, opened 2017-05-26
WT-3248 Mark reconciliation a success if no skipped updates

sulabhM:


michaelcahill:
@keithbostic and @agorrod, please take a look at this change and merge if you reach agreement.

This fixes the perf instability in the `overflow-130k.wtperf` workload, and fixes the extreme slowdown in the join visibility test in WT-3340.

The fundamental problem in both cases was giving up on update/restore evictions unnecessarily.  In the case of the join visibility test, a page has grown large (hence forced eviction), and contains a lot of tombstones.

Previously, we were only keeping track of the size of *saved* updates on a page, and the fraction of saved updates that were "skipped" (i.e., uncommitted).  With this change, we track *all* updates seen on the page, and only give up on an update / restore if all the updates are saved.

In other words, if we manage to apply any updates during reconciliation (or we split), we count that as progress and the update/restore is allowed to complete.  Only for single pages where we don't split and don't apply any updates do we give up (and potentially go to lookaside eviction based on cache pressure).  I checked the long-mode-only `test_las.py` and the lookaside table is still being used when needed.

This is a change to an area of the code that we argued about last time it was changed, so please take a look and see what you think.

agorrod:
Thanks @sulabhM and @michaelcahill this change lgtm. I'd really like if how aggressive we are with update/restore was somehow linked to the difficulty we are having finding space in the cache, but I think this is better in the general case than switching to LAS more often.

agorrod:
The perf tests showed improvement for the expected config with this branch, and no degradation on other tests.

src/reconcile/rec_write.c
```diff
@@ -48,8 +48,9 @@ typedef struct {
 	/* Track the page's maximum transaction ID. */
 	uint64_t max_txn;
 
-	uint64_t update_mem;		/* Total update memory */
-	uint64_t update_mem_skipped;	/* Skipped update memory */
+	uint64_t update_mem_all;	/* Total update memory size */
+	uint64_t update_mem_saved;	/* Saved update memory size */
+	uint64_t update_mem_uncommitted;/* Uncommitted update memory size */
```
agorrod: I'm inclined to leave these as-is even if they hold more information than we use at the moment. It feels like we are still iterating on the right solution, I've created:
https://jira.mongodb.org/browse/WT-3344
to make sure we don't forget to approach this more scientifically in the future.

PR #3442, opened 2017-05-26
WT-3329 Visit trees using a tiny fraction of cache.

michaelcahill:
For workloads where no tree takes up a large enough fraction of cache,
we were using a randomized approach to deciding when eviction should
visit trees.  That led to slow performance for workloads with uniform
updates over thousands of trees.


michaelcahill:
FTR, mostly so this doesn't get lost.  The million collection workload uncovered a perf degradation when there are thousands of active trees.  This change fixed it, but has not been evaluated for other workloads yet.

agorrod:
I've switched the Jenkins perf jobs to run on this branch.

agorrod:
@michaelcahill I'm concerned this unwinds some of the benefit required for SERVER-27700 - I'd like to be confident that it won't introduce a regression for those cases before merging.

michaelcahill:
@agorrod, I share your concern.  To be clear, this change is important in scaling to a large number of tables that are all the same size (i.e., once the number of tables grows to 10K, each occupies less than 0.01% of cache).

We have `multi-btree-zipfian*.wtperf` (with varying numbers of tables) that can be used to test whether these changes introduce regressions.  Are there other workloads you can think of that we should check?

michaelcahill:
autobuild: please retest this

michaelcahill:
autobuild: retest this please

daveh86:
I've run the multi-btree-zipfian*.wtperf workload with 1, 100, 1000, 5000 and 10000 tables. This saw no performance degradation and often massive performance wins (5000 tables being the biggest leap by far with 7x better performance). I'm going to make this the perf branch and see if that turns anything up, otherwise I believe this change should be merged.

daveh86:
This has run as the perf-branch for a run and showed no degradation. I will let it run for 24H, but as it stands I am confident this is a good change.

agorrod:
@daveh86 I've added a comment to https://jira.mongodb.org/browse/WT-3329 regards the testing I think is necessary.

daveh86:
@agorrod the testing requested was a success. The runs against WT's perf branch showed consistent performance. The run with the specific config you requested showed a small performance gain (~1%). I think this can be merged.

agorrod:
I've updated this to the latest develop, and have switched the Jenkins perf branches over again (call me paranoid). If all looks good I'll go ahead and merge

PR #3443, opened 2017-05-26
WT-3343 WiredTiger database close can attempt unlock of a lock that's not held.

keithbostic:
Closing the file handles that remain open at database close will attempt to unlock a lock that's not currently held.


agorrod:
Thanks for the fix @keithbostic lgtm.

PR #3445, opened 2017-05-29
WT-3307 Close btree/dhandle properly when handling __wt_checkpoint_close failure

sulabhM:


sulabhM:
@michaelcahill , please have a look.

sueloverso:
@sulabhM I've been studying this change today and would like to discuss it with you.  My concern started when I considered this part of the comment in `wt_conn_btree_sync_and_close`:
```
         * The close can fail if an update cannot be written, return the EBUSY
         * error to our caller for eventual retry.
```
If I look at `wt_checkpoint_close` it explicitly returns EBUSY when transaction visibility checks fail.  I'm worried that your change will alter the state of the dhandle in an unexpected way in that case where we're expecting a retry.  I'm wondering if the change should look something like this instead:
```
ret = __wt_checkpoint_close(session, final);
if (ret == EBUSY)
    WT_ERR(ret);
else
    WT_TRET(ret);
```
In the case of EBUSY we want to bail out and in all other errors (or 0) we want to continue.

sulabhM:
@sueloverso , this is ready for another round of review.

sueloverso:
Thanks @sulabhM LGTM

PR #3446, opened 2017-05-29
WT-3345 Tune WiredTiger's read/write locks.

michaelcahill:
Add a workload that stresses rwlock performance under various conditions
(including threads >> cores), tune read and write lock operations to
only spin when it is likely to help, and to back off to a condition
variable when there is heavy contention.


michaelcahill:
Not yet ready for review, pushing in order to run automated performance tests against these changes.

src/btree/bt_page.c
```diff
@@ -65,6 +65,9 @@ __wt_page_alloc(WT_SESSION_IMPL *session,
 	page->type = type;
 	page->read_gen = WT_READGEN_NOTSET;
 
+	/* Initialize the page lock. */
+	__wt_rwlock_init(session, &page->page_lock);
+
```
michaelcahill: This is a real bug regardless of other changes: we were relying on `__wt_calloc` to initialize a `WT_RWLOCK`.

michaelcahill:
@keithbostic, please take a look.  I haven't had a chance to double check the results in the ticket: I may have run the read-only test on a diagnostic build or something, will try to reproduce tomorrow.

I'm ambivalent about keeping `test_rwlock` -- it has been super useful for this exercise but probably won't be much use for anything else.

Also I note that I've left some `XXX`s in the code: in particular, we will need to deal with errors from allocating and destroying condition variables.

Lastly, we discussed changing `__wt_cond_alloc` to `__wt_cond_init` (i.e., to not always allocate memory).  I'm still keen but didn't get to that today.

src/include/serial.i
```diff
@@ -154,7 +154,7 @@ __col_append_serial_func(WT_SESSION_IMPL *session, WT_INSERT_HEAD *ins_head,
 static inline int
 __wt_col_append_serial(WT_SESSION_IMPL *session, WT_PAGE *page,
     WT_INSERT_HEAD *ins_head, WT_INSERT ***ins_stack, WT_INSERT **new_insp,
-    size_t new_ins_size, uint64_t *recnop, u_int skipdepth)
+    size_t new_ins_size, uint64_t *recnop, u_int skipdepth, bool exclusive)
```
michaelcahill: Note that these `exclusive` changes are required to avoid self-deadlock where the update/restore path is holding a page lock while re-applying changes.  Now they share the same locks, we need to skip page locking in that path.

I don't love it, very open to better ideas.

src/cursor/cur_file.c
```diff
@@ -520,7 +520,14 @@ __curfile_create(WT_SESSION_IMPL *session,
...
 	if (0) {
-err:		WT_TRET(__curfile_close(cursor));
+err:		/*
+		 * Our caller expects to release the data handle if we fail.
+		 * Disconnect it from the cursor before closing.
+		 */
+		if (session->dhandle != NULL)
+			__wt_cursor_dhandle_decr_use(session);
+		cbt->btree = NULL;
+		WT_TRET(__curfile_close(cursor));
```
michaelcahill: This was a bug uncovered by having sanity checking in the new rwlocks.

yjhjstz:
any benchmark show the improvement ?

keithbostic:
> I'm ambivalent about keeping `test_rwlock` -- it has been super useful for this exercise but probably won't be much use for anything else.

@michaelcahill 

I'd keep it, and not configured to run in the short-test case seems reasonable.

I pushed a bunch of changes for `test_rwlock` -- nothing interesting, but my compiler setup complained about some stuff, figured it couldn't hurt.

src/support/mtx_rw.c
```diff
@@ -198,37 +163,75 @@ __wt_readlock_spin(WT_SESSION_IMPL *session, WT_RWLOCK *l)
 void
 __wt_readlock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
```
keithbostic: Just noticed: the `__wt_readlock` and `__wt_writelock` functions have read barriers, but the "try" versions don't. I think that's probably a bug.

src/support/mtx_rw.c
```diff
@@ -198,37 +163,75 @@ __wt_readlock_spin(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
+		 * Note: don't re-read from the lock or we could race with a
+		 * writer unlocking.
+		 */
+		if (old.u.s.readers_queued == UINT16_MAX ||
+		    (old.u.s.next != old.u.s.current &&
+		    old.u.s.readers_queued > old.u.s.next - old.u.s.current)) {
+			__wt_cond_wait(
+			    session, l->cond_readers, WT_THOUSAND, NULL);
+			goto restart;
+		}
```
keithbostic: I would flip the `old.u.s.next != old.u.s.current` test to be `current != next` to match the test above.

I do not understand why we're testing `readers_queue > next - current`, that needs a comment.

src/support/mtx_rw.c
```diff
@@ -149,46 +132,28 @@ __wt_try_readlock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
 
-	/*
-	 * The replacement lock value is a result of allocating a new ticket and
-	 * incrementing the reader value to match it.
-	 */
-	new.s.readers = new.s.next = old.s.next + 1;
-	return (__wt_atomic_cas64(&l->u, old.u, new.u) ? 0 : EBUSY);
+	/* The replacement lock value is a result of adding an active reader. */
+	new.u.s.readers_active++;
+	return (__wt_atomic_casv64(&l->u.v, old.u.v, new.u.v) ? 0 : EBUSY);
```
keithbostic: I think `readers_active` can overflow.

src/support/mtx_rw.c
```diff
@@ -198,37 +163,75 @@ __wt_readlock_spin(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
-		 * paused enough, yield so we don't burn CPU to no purpose. This
-		 * situation happens if there are more threads than cores in the
-		 * system and we're thrashing on shared resources.
+		 * Check for wrapping: if we have more than 64K lockers
+		 * waiting, the ticket value will wrap and two lockers will
+		 * simultaneously be granted the lock.
 		 */
-		if (++pause_cnt < WT_THOUSAND)
+		WT_ASSERT(session, new.u.s.readers_queued != 0);
+	} while (!__wt_atomic_casv64(&l->u.v, old.u.v, new.u.v));
```
keithbostic: The `Check for wrapping: if we have more than 64K lockers` comment no longer applies to this code.

I think we can remove the assert on `readers_queued != 0`, we know it's safe, we incremented it immediately before testing it.

src/support/mtx_rw.c
```diff
@@ -328,31 +365,36 @@ __wt_writelock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
+			new.u.s.readers_queued = 0;
+		}
+	} while (!__wt_atomic_casv64(&l->u.v, old.u.v, new.u.v));
 
 	WT_DIAGNOSTIC_YIELD;
+
+	if (new.u.s.readers_active != 0)
+		__wt_cond_signal(session, l->cond_readers);
+	else if (new.u.s.current != new.u.s.next)
+		__wt_cond_signal(session, l->cond_writers);
```
keithbostic: I'd move the diagnostic yield after signalling readers/writers, I think it's a better test, and otherwise we could seriously impact throughput, there won't be anybody holding the lock.

src/support/mtx_rw.c
```diff
@@ -264,22 +276,35 @@ __wt_try_writelock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
+	 * on the lock, that is, if this thread's ticket would be the next
+	 * ticket granted.  Check if this can possibly succeed (and confirm the
+	 * lock is in the correct state to grant this write lock).
 	 */
-	if (old.s.writers != old.s.next)
+	old.u.v = l->u.v;
+	if (old.u.s.current != old.u.s.next || old.u.s.readers_active != 0)
 		return (EBUSY);
 
+	WT_ASSERT(session, old.u.s.readers_queued == 0);
```
keithbostic: The `WT_ASSERT(session, old.u.s.readers_queued == 0);` makes sense to me, but I think it probably deserves a comment about why it's an interesting test, something like "if there are no active readers or writers, there should be no queued readers".

src/support/mtx_rw.c
```diff
@@ -198,37 +163,75 @@ __wt_readlock_spin(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
 	WT_DIAGNOSTIC_YIELD;
 
-	/*
-	 * Possibly wrap: if we have more than 64K lockers waiting, the ticket
-	 * value will wrap and two lockers will simultaneously be granted the
-	 * lock.
-	 */
-	ticket = __wt_atomic_fetch_add16(&l->s.next, 1);
-	for (pause_cnt = 0; ticket != l->s.readers;) {
+	pause_cnt = 0;
```
keithbostic: `pause_cnt` doesn't need to be set here.

src/support/mtx_rw.c
```diff
@@ -289,29 +314,41 @@ __wt_try_writelock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
 			WT_PAUSE();
-		else
-			__wt_sleep(0, 10);
+		else if (pause_cnt < 1200)
+			__wt_yield();
+		else {
+			session->current_rwlock = l;
+			session->current_rwticket = ticket;
+			__wt_cond_wait(
+			    session, l->cond_writers, 0, __write_blocked);
```
keithbostic: `pause_cnt` should be reset to 0 after waiting on the condition variable (at least, that's the approach in `__wt_readlock()`).

src/include/serial.i
michaelcahill: Done.

src/support/mtx_rw.c
michaelcahill: I added comments explaining why: the atomic operations in the trylock functions are sufficient, but in the slow path we need a barrier after the thread enters the lock.

src/support/mtx_rw.c
michaelcahill: I've removed both: that was an experiment when I wasn't confident about the notifications arriving.

michaelcahill:
@keithbostic, I think I've addressed everything in your review.  Could you please take another look?

src/support/mtx_rw.c
```diff
@@ -149,46 +121,30 @@ __wt_try_readlock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
-	 * The replacement lock value is a result of allocating a new ticket and
-	 * incrementing the reader value to match it.
+	 * The replacement lock value is a result of adding an active reader.
+	 *
+	 * We rely on this atomic operation to provide a barrier.
 	 */
-	new.s.readers = new.s.next = old.s.next + 1;
-	return (__wt_atomic_cas64(&l->u, old.u, new.u) ? 0 : EBUSY);
+	new.u.s.readers_active++;
+	return (__wt_atomic_casv64(&l->u.v, old.u.v, new.u.v) ? 0 : EBUSY);
```
keithbostic: I think `readers_active` can technically overflow here, can't it? Imagine 65K `__wt_try_readlock` calls.

I would agree it's wildly unlikely.

keithbostic:
I was thinking maybe we should add a mode to the `thread_rwlock` function in the test program to have a memory location it's locking and checking between the locks? It could explicitly yield to get strong testing.

src/btree/bt_handle.c
```diff
@@ -442,7 +442,7 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
 	}
 
 	/* Initialize locks. */
-	__wt_rwlock_init(session, &btree->ovfl_lock);
+	WT_RET(__wt_rwlock_init(session, &btree->ovfl_lock));
```
michaelcahill: I've made them consistent with what we do elsewhere, let's follow up on that discussion outside this change.

src/include/btmem.h
```diff
@@ -420,11 +420,13 @@ struct __wt_page_modify {
...
-#define	WT_PAGE_UNLOCK(s, p)						\
-	__wt_spin_unlock((s), &S2C(s)->page_lock[(p)->modify->page_lock])
-	uint8_t page_lock;		/* Page's spinlock */
+#define	WT_PAGE_LOCK(s, p) 						\
+	__wt_spin_lock((s), &(p)->modify->page_lock)
+#define	WT_PAGE_TRYLOCK(s, p)						\
+	__wt_spin_trylock((s), &(p)->modify->page_lock)
+#define	WT_PAGE_UNLOCK(s, p) 						\
+	__wt_spin_unlock((s), &(p)->modify->page_lock)
+	WT_SPINLOCK page_lock;		/* Page's spinlock */
```
michaelcahill: We could save some bytes by using an unpadded version of `WT_SPINLOCK`.  Also note that this is only in modified pages.

I don't see any easy alternative: the earlier version where we had shared locks isn't going to work when reconciliation is holding pages locked and doing inserts into the lookaside table (and there are probably other cases I'm not aware of).

src/support/mtx_rw.c
michaelcahill: I've pushed the fix.

michaelcahill:
Happy to enhance `test_rwlock`: any thoughts on how to detect whether there are readers active concurrent with a write?

michaelcahill:
Thanks for the review @keithbostic.  FTR, we discussed status and agreed these changes are ready to merge, so I'm going to go ahead and do that.

PR #3450, opened 2017-06-01
WT-3346 create JSON output for latency sampling.

ddanderson:
Also added a script to merge workgen and WT outputs.


agorrod:
Thanks @ddanderson lgtm

PR #3451, opened 2017-06-02
WT-3354 Fix bugs found by Coverity.

michaelcahill:
* two cases where error checking for rwlocks should goto the error label for cleanup.
* LSM code not restoring isolation if a checkpoint fails part way through

### DIFF
--- a/src/support/mtx_rw.c
+++ b/src/support/mtx_rw.c
@@ -235,8 +235,6 @@ stall:			__wt_cond_wait(
 		}
 	}
 
-	WT_ASSERT(session, l->u.s.readers_active > 0);
-
 	/*
 	 * Applications depend on a barrier here so that operations holding the
 	 * lock see consistent data.  The atomic operation above isn't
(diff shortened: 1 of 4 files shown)


michaelcahill:
Also, there was just a failure in [a PPC stress test](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-ppc/10140/console): my reading is that this assertion needs to be ordered after the read barrier to be safe.

src/support/mtx_rw.c
```diff
@@ -245,6 +243,8 @@ stall:			__wt_cond_wait(
 	 * meantime.
 	 */
 	WT_READ_BARRIER();
+
+	WT_ASSERT(session, l->u.s.readers_active > 0);
```
keithbostic: @michaelcahill, I'm not seeing this one. The lock structure is volatile, so this read cannot be cached/re-ordered, and the thread of control is holding a read lock so `readers_active` must be non-zero?

michaelcahill: As discussed, `volatile` prevents the _compiler_ from reordering instructions, but a CPU with a weak memory model can schedule a read earlier (i.e., use a cached value).

PR #3453, opened 2017-06-05
WT-3251 remove interim buffer when splitting during reconciliation 

sulabhM:


sulabhM:
@daveh86 , mind taking a look at this change?

src/reconcile/rec_write.c
```diff
@@ -3172,15 +3128,112 @@ __rec_split_raw(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len)
...
+	uint32_t combined_size, len_to_move;
+	uint8_t *cur_dsk_start;
+
+	WT_ASSERT(session, r->prev_img_ptr != NULL);
+
+	btree = S2BT(session);
+	bnd_cur = &r->bnd[r->bnd_next];
+	bnd_prev = bnd_cur - 1;
+	*chunks_merged = false;
+	combined_size = bnd_prev->size +
```
daveh86: Maybe a comment here, explaining why we want to remove one HEADER's worth of size.

src/reconcile/rec_write.c
```diff
@@ -3172,15 +3128,112 @@ __rec_split_raw(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len)
 }
 
 /*
+ * __rec_split_finish_process_prev --
+ * 	If the two split chunks together fit in a single page, merge them into
+ * 	one. If they do not fit in a single page but the last is smaller than
+ * 	the minimum desired, move some data from the penultimate chunk to the
+ * 	last chunk and write out the previous/penultimate. Finally, update the
+ * 	pointer to the current image buffer.
```
daveh86: I think its worth mentioning that if we have two merged chunks, we will write them from the "prev" slot, not the current slot.

src/reconcile/rec_write.c
```diff
@@ -2473,89 +2457,59 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
+    WT_SESSION_IMPL *session, WT_RECONCILE *r)
 {
-	WT_BM *bm;
 	WT_BOUNDARY *bnd_cur, *bnd_prev;
-	WT_BTREE *btree;
-	WT_PAGE_HEADER *dsk, *dsk_tmp;
-	size_t cur_len, len;
-	uint8_t *dsk_start;
-
-	WT_ASSERT(session, r->bnd_next != 0);
```
sulabhM: r->bnd_next starts initialised as 0 and r->prev_img_ptr starts initialised as NULL. bnd_next is index into the boundary array, we increment it when we create a boundary. To create the first boundary, first we initialise the second buffer, let prev_img_ptr point to the first image and current image pointer point to the second image, then we increment bnd_next to 1.

So assert is right in checking "r->prev_img_ptr != NULL && r->bnd_next != 0".
Said that I am updating assert to check after first boundary creation.

Assert will check the following now:

        WT_ASSERT(session, (r->prev_img_ptr == NULL && r->bnd_next == 0) ||      
            (r->prev_img_ptr != NULL && r->bnd_next != 0));

src/reconcile/rec_write.c
sulabhM: After this function exits, we will only have one buffer, the current image pointer. At the end of the function:
r->cur_img_ptr = r->prev_img_ptr;

I will update comment to address that expectation is to write out of current image pointer after this function exits

sulabhM:
@keithbostic , please have a look.
I have addressed Dave's comments. Perf tests on this change doesn't show any regression. Depending upon the update size, there seems to be 1-5% improvement in the number of updates in the workload attached to the original reconciliation page layout ticket.

daveh86:
retest this please

src/reconcile/rec_write.c
```diff
@@ -976,8 +978,8 @@ __rec_destroy(WT_SESSION_IMPL *session, void *reconcilep)
 		return;
 	*(WT_RECONCILE **)reconcilep = NULL;
 
-	__wt_buf_free(session, &r->disk_image);
-	__wt_scr_free(session, &r->interim_buf);
+	__wt_buf_free(session, &r->disk_image[0]);
+	__wt_buf_free(session, &r->disk_image[1]);
```
sulabhM: @michaelcahill , what are your thoughts about this.

src/reconcile/rec_write.c
```diff
@@ -3172,15 +3143,119 @@ __rec_split_raw(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t next_len)
...
+	} else {
+		if (bnd_cur->size < r->min_split_size &&
+		    bnd_prev->min_bnd_offset != 0 ) {
+			/*
+			 * The last chunk, pointed to by the current image
+			 * pointer, has less than the minimum data. Let's move
+			 * any data more than the minimum from the previous
+			 * image into the current.
+			 */
+			len_to_move = bnd_prev->size - bnd_prev->min_bnd_offset;
```
sulabhM: Fixed these. David helped fix the pull request tester to catch such issues.

src/reconcile/rec_write.c
```diff
@@ -2473,89 +2471,60 @@ __rec_split_grow(WT_SESSION_IMPL *session, WT_RECONCILE *r, size_t add_len)
...
-	(void)memmove(dsk_start, (uint8_t *)dsk + bnd_cur->offset, cur_len);
-
-	/* Fix boundary offset */
-	bnd_cur->offset = WT_PAGE_HEADER_BYTE_SIZE(btree);
-	/* Fix where free points */
-	r->first_free = dsk_start + cur_len;
+	/* swap previous and current buffers */
+	tmp_img_ptr = r->prev_img_ptr;
+	r->prev_img_ptr = r->cur_img_ptr;
+	r->cur_img_ptr = tmp_img_ptr;
```
sulabhM: I think other than "type", all header fields change for each split chunk that we write to the disk. So we set the type field when we initialise the buffer and do not touch it again. I didn't realize that we have a page field in the reconciliation structure, so based the code on `__rec_split_init`, I thought i can copy from the other buffer. Using `r->page->type` is cleaner, I have made that change.

I scanned rec_write.c , there are probably several places we can use `r->page->type` instead of `dsk->type`. But I don't totally understand the code around all these places. Also, for instance, in the case of type being `WT_PAGE_OVFL`, `r->page->type` can be different from `dsk->type`.

keithbostic: > Also, for instance, in the case of type being WT_PAGE_OVFL, r->page->type can be different from dsk->type.

Overflow pages should be a separate case, we don't reconcile overflow pages.

If you'd like me to take a run at this and you could do the review, let me know.

sulabhM: @keithbostic , I will be happy to review, please go ahead with the change.

src/reconcile/rec_write.c
michaelcahill: I'd steer clear of using scratch buffers (for pretty much anything, actually, but that's a bigger discussion).  Modern memory allocators like tcmalloc and jemalloc already stash recently freed memory for fast reuse.

I'd particularly avoid scratch buffers in reconciliation because having large chunks of memory hang around after eviction can be problematic when there are lots of active sessions.

PR #3454, opened 2017-06-05
WT-3356 Use atomic reads of rwlocks.

michaelcahill:
Previously we had some conditions that checked several fields within a
rwlock by indirecting to the live structure.  Switch to always doing a
read of the full 64-bit value, then using local reads from the copy.

Otherwise, we're relying on the compiler and the memory model to order
the structure accesses in "code execution order".  That could explain
assertion failures and/or incorrect behavior with the new rwlock
implementation.


michaelcahill:
Discussed with @keithbostic, this code has been in stress testing all day with no new failures.  This seems like a definite improvement, merging to get more stress and perf testing overnight.

PR #3455, opened 2017-06-06
WT-2596 Update the documentation to detail what occurs when WT crashe…

daveh86:
…s and a backup cursor is open

### DIFF
--- a/src/docs/backup.dox
+++ b/src/docs/backup.dox
@@ -59,9 +59,10 @@ During the period the backup cursor is open, database checkpoints can
 be created, but no checkpoints can be deleted.  This may result in
 significant file growth.
 
-Additionally, if a crash occurs during the period the backup cursor is open,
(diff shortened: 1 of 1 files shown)


daveh86:

src/docs/backup.dox
```diff
@@ -59,6 +59,10 @@ During the period the backup cursor is open, database checkpoints can
 be created, but no checkpoints can be deleted.  This may result in
 significant file growth.
 
+Additionally, if a crash occurs during the period the backup cursor is open,
+then the system will be restored to the most recent checkpoint prior to the
+opening of the backup cursor, even if later database checkpoints were created.
+
```
michaelcahill: I think we need the further qualification here "if a crash occurs during the period the backup cursor is open **and logging is disabled**, then ..."

michaelcahill:

PR #3456, opened 2017-06-06
WT-3361 Resolve Windows build warnings, build more test programs on Windows

keithbostic:
@markbenvenuto, there's only a minor change to the WiredTiger library itself, the rest is boilerplate to deal with test program portability.

@michaelcahill, with these changes, Windows builds without warnings.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -420,7 +420,6 @@ env.Append(CPPPATH=["test/utility"])
 t = env.Program("t_bloom",
     "test/bloom/test_bloom.c",
     LIBS=[wtlib, testutil])
-env.Alias("check", env.SmokeTest(t))
 Default(t)
 
 t = env.Program("t_checkpoint",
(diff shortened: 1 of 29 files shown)


michaelcahill:
FWIW, these changes lgtm.  I'll leave them for @markbenvenuto to check.

SConstruct
```diff
@@ -410,41 +410,37 @@ def builder_smoke_test(target, source, env):
...
                 'test/utility/parse_opts.c'
             ])
+env.Append(CPPPATH=["test/utility"])
 
-#Don't test bloom on Windows, its broken
 t = env.Program("t_bloom",
     "test/bloom/test_bloom.c",
-    LIBS=[wtlib, testutil] + wtlibs)
-#env.Alias("check", env.SmokeTest(t))
+    LIBS=[wtlib, testutil])
```
markbenvenuto: Leave the `+ wtlibs`. It allows the user to build with and without zlib for instance.

keithbostic: > Leave the `+ wtlibs`. It allows the user to build with and without zlib for instance.

Fixed in fbd1eda, thanks!

PR #3457, opened 2017-06-07
WT-3349 Add rwlock statistics

sulabhM:


sulabhM:
@michaelcahill , please have a look.
Not sure where to put the macro WT_RWLOCK_INIT_TRACKED, I have put in misc.i for now.

michaelcahill:
One further change I've pushed is to increment the operation counts for `try_readlock` and `try_writelock`, and in the fast path of `readlock` and `writelock`.

PR #3458, opened 2017-06-08
WT-3350 do away with dynamic allocation of WT_CONDVAR

sulabhM:


sulabhM:
autobuild: retest this please

sulabhM:
@sueloverso , please have a look

src/os_posix/os_mtx_cond.c
```diff
@@ -9,30 +9,22 @@
...
  */
 int
-__wt_cond_alloc(WT_SESSION_IMPL *session, const char *name, WT_CONDVAR **condp)
+__wt_cond_init(WT_SESSION_IMPL *session, WT_CONDVAR *cond, const char *name)
 {
-	WT_CONDVAR *cond;
-	WT_DECL_RET;
-
-	WT_RET(__wt_calloc_one(session, &cond));
+	WT_UNUSED(session);
```
agorrod: I'd feel more comfortable if we cleared the content of `WT_CONDVAR` via memset in here.

src/log/log.c
```diff
@@ -2227,30 +2226,21 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
...
-		 *
-		 * XXX I've seen times when conditions are NULL.
-		 */
-		if (conn->log_cond != NULL) {
-			__wt_cond_signal(session, conn->log_cond);
-			__wt_yield();
-		} else
-			WT_ERR(__wt_log_force_write(session, 1, NULL));
+		__wt_cond_signal(session, &conn->log_cond);
+		__wt_yield();
```
agorrod: This change makes me nervous - if a call comes through here and ``conn->log_cond`` has not been initialized, I suspect it'd crash, and the existing comment inferred that could happen. Do you think it can't any longer?

src/async/async_api.c
```diff
@@ -240,7 +240,7 @@ __async_start(WT_SESSION_IMPL *session)
 	async = conn->async;
 	TAILQ_INIT(&async->formatqh);
 	WT_RET(__wt_spin_init(session, &async->ops_lock, "ops"));
-	WT_RET(__wt_cond_alloc(session, "async flush", &async->flush_cond));
+	WT_RET(__wt_cond_init(session, &async->flush_cond, "async flush"));
```
sueloverso: Was there a reason to swap the order of the args?  I don't have strong feelings either way, it just seems gratuitous.

src/log/log.c
sueloverso: The `log_cond` is used with the internal log server thread.  So it was allocated and initialized when we created the `log_server` thread in `wt_logmgr_open`.  It is in there (instead of `wt_logmgr_create`) because the user may get into this code via reconfigure.  So I do think we can come through here still without initialization because recovery/checkpoint writes log records before `logmgr_open`.  That is we've opened the underlying logging subsystem but haven't started all the worker threads.  Recovery and writing a log record does not need all the worker threads.  Or refactor to always initialize the condition var in `logmgr_create` but reconfigure needs consideration.

src/os_win/os_mtx_cond.c
```diff
@@ -164,15 +160,11 @@ __wt_cond_signal(WT_SESSION_IMPL *session, WT_CONDVAR *cond)
...
 {
-	WT_CONDVAR *cond;
-
-	cond = *condp;
 	if (cond == NULL)
 		return;
 
 	/* Do nothing to delete Condition Variable */
 	DeleteCriticalSection(&cond->mtx);
-	__wt_free(session, *condp);
```
sueloverso: We've removed the one place session is used.  I think we need an `WT_UNUSED` here now.

src/async/async_api.c
michaelcahill: Yes, the convention is that when we allocate and return memory, we do that in the last argument.  When we are initializing a structure, that is the left-most argument (apart from possibly `session`).

src/os_posix/os_mtx_cond.c
michaelcahill: I'd recommend `WT_CLEAR(*cond);`

src/log/log.c
michaelcahill: I sat with @sulabhM and tried to work through this code.  We could add a boolean to track whether `__wt_logmgr_create` has been called, but it seems weird that recovery can call into here without the logging being initialized.

While on the topic, having `reconfigure` call `__wt_logmgr_create` seems kind of backwards to me.  Elsewhere we call `*_create` during `wiredtiger_open` and `*_destroy` during `WT_CONNECTION::close`.

I don't see any harm in unconditionally initializing the condition variable and keeping it live until the connection is closed.  In other words, I don't think there's any reason to destroy and recreate the condition variable as part of a `reconfigure`.

Would it be reasonable to separate out "initializing" logging (done early, before any calls through here) from the current `__wt_logmgr_open`, assuming we can't move all of open before recovery?

sueloverso: `wt_logmgr_create` is always called before recovery is run.  So logging is initialized.  It goes:
* `wt_logmgr_create` to initialize the subsystem (including
* Run recovery
* `wt_logmgr_open` to start worker threads

I misspoke - reconfigure does not do anything with the cond variables and does not come through this path.  They already get created/destroyed as you suggest, created during open and destroyed on close.  I wonder if this is leftover from the earlier logging code because when I add an fatal assertion in the original else clause it never triggers, even running test/recovery and the recovery tests in the test suite.

sueloverso: Given that it appears that the original `else` clause is not used AFAICT, I'm good with this change.

src/async/async_api.c
sueloverso: Ok.  I'm good with that.  The nature of the diff is what prompted that.

src/os_win/os_mtx_cond.c
sueloverso: I think this might be the only outstanding minor change that is needed.

sueloverso:
Other than one minor thing in os_win/os_mtx_cond.c this change LGTM.

src/os_win/os_mtx_cond.c
sulabhM: Looks like our windows pull request tester doesn't look for unused variables.

sulabhM:
` 	} else if (force) {
 -		/*
 -		 * If we are going to wait for this slot to get written,
 -		 * signal the wrlsn thread.
 -		 *
 -		 * XXX I've seen times when conditions are NULL.
 -		 */
 -		if (conn->log_cond != NULL) {
 -			__wt_cond_signal(session, conn->log_cond);
 -			__wt_yield();
 -		} else
 -			WT_ERR(__wt_log_force_write(session, 1, NULL));
 +		__wt_cond_signal(session, &conn->log_cond);
 +		__wt_yield();`

To summarise the review comments surrounding the above change:
There was a possibility to reach this portion of the code without having initialised the CV, which appears to have been left over from a change in the past and is no longer true today. Hence, as per the discussion in this review, it appears safe to assume that the CV will be initialised when executing this code portion.

sulabhM:
@sueloverso , test/csuite/test_wt2719_reconfig would hang every once in a while. I have pushed a change to fix that. Please have a look.

sueloverso:
@sulabhM I assume the fixes you refer to are the ones that call `wt_cond_destroy` only if the PID is set.  There are three similar changes: `conn_stat.c, conn_sweep.c, conn_ckpt.c`.   I'm having difficulty understand those changes.  The condition variable is initialized unconditionally before starting the thread.  So if the later thread creation returns an error so that the pid is not set, it seems the original code is still correct that we want to destroy the condition variable unconditionally as well.  Can you explain the problem and how this solves it?

sulabhM:
@sueloverso : Destroy for those three CVs get called by `__wt_connection_close` and

sulabhM:
@sueloverso : I am trying to conditionally destroy CV, the reason being, these destroy functions are called both by shutdown and by reconfigure. Imagine the case where reconfigure is called to turn off stat, reconfigure in that case unconditionally destroys the stat CV and the thread, and unsets the pid. When the shutdown gets called, the stat CV has already been destroyed, and calling destroy on the CV unconditionally leads to a hang. (It was okay to do so before as `__wt_cond_destroy` used to check the CV being not NULL before a destroy)

You are right, the change I made, in the case thread creation fails will leave the CV initialized. I have pushed another change, which destroys the CV if thread creation failed and then only destroys once if pid is set. This fixes both 1. reconfigure calling destroy, followed by shutdown calling destroy 2. Calling destroy if thread creation fails.

Another look at the code and I see this change is not needed for sweep server as reconfigure doesn't call for a destroy of the `sweep_cond` CV.

sueloverso:
@sulabhM thanks for the explanation.  With that and looking at the code I came up with a minor change that I pushed in changeset 5553401.  See if you like it.  Basically we can destroy the CV at the same place we close the internal session for each of these.  The session code for each thread already does what is needed.

sulabhM:
@sueloverso ex_encrypt aborts on OSX because in that test `pthread_cond_destroy` gets called over an uninitialized CV, pthread implementation on OSX fails, causing WT_PANIC to be called. For some reason it succeeds on Linux. I googled, but couldn't find anything on expected behavior if destroy is called without initializing a CV.

`__btree_conf` has the following portion to setup encryptor configuration:

```
        /*                                                                          
         * We do not use __wt_config_gets_none here because "none" and the empty 
         * string have different meanings. The empty string means inherit the       
         * system encryption setting and "none" means this table is in the clear 
         * even if the database is encrypted.                                       
         */                                                                         
        WT_RET(__wt_config_gets(session, cfg, "encryption.name", &cval));           
        if (cval.len == 0)                                                          
                btree->kencryptor = conn->kencryptor;                               
        else if (WT_STRING_MATCH("none", cval.str, cval.len))                       
                btree->kencryptor = NULL;                                           
        else {                                                                      
                WT_RET(__wt_config_gets_none(                                       
                    session, cfg, "encryption.keyid", &keyid));                     
                WT_RET(__wt_config_gets(session, cfg, "encryption", &enc));         
                if (enc.len != 0)                                                   
                        WT_RET(__wt_strndup(session, enc.str, enc.len,              
                            &enc_cfg[0]));                                          
                ret = __wt_encryptor_config(session, &cval, &keyid,                 
                    (WT_CONFIG_ARG *)enc_cfg, &btree->kencryptor);                  
                __wt_free(session, enc_cfg[0]);                                     
                WT_RET(ret);                                                        
        }                                                                           
                                                                                    
        /* Initialize locks. */                                                                       
        WT_RET(__wt_rwlock_init(session, &btree->ovfl_lock));                       
        WT_RET(__wt_spin_init(session, &btree->flush_lock, "btree flush"));         
                                                                                    
        btree->modified = false;                        /* Clean */                 
```

In that test for a btree, `file:cryptobad.wt` the `__wt_encryptor_config` function call fails, skipping rwlock_init for `btree->ovfl_lock`, but when shutting down `__btree_clear` calls rwlock_destroy on that lock. rwlock contains CVs that in the above process gets destroyed without being initialized. Before the change I made, CV would have been NULL and destroy would be a no-op, not anymore.

Now, `__wt_encryptor_config` for the test ends up calling `rotate_customize` in `examples/c/ex_encrypt.c`, which fails returning EINVAL because the id provided is wrong:

```
	if (ITEM_MATCHES(keyid, "system")) {
		if (my_crypto->password == NULL ||
		    strcmp(my_crypto->password, SYS_PW) != 0) {
			ret = EPERM;
			goto err;
		}
		my_crypto->rot_N = 13;
	} else if (ITEM_MATCHES(keyid, USER1_KEYID))
		my_crypto->rot_N = 4;
	else if (ITEM_MATCHES(keyid, USER2_KEYID))
		my_crypto->rot_N = 19;
	else {
		ret = EINVAL;
		goto err;
	}
```

```
#define	USERBAD_KEYID	"userbad"
```

```
(gdb) p keyid
$1 = {str = 0x6d4037 "userbad,name=rotn)", len = 7, val = 1, type = WT_CONFIG_ITEM_ID}
```

By the looks of it, the encryptor and test are doing the right thing. So maybe `__btree_clear` shouldn't have been called on this btree when `__btree_conf` failed for it.

I am working on figuring out:
1. why was `__btree_clear` called for a btree for which `__btree_conf` failed
2. Is there anything we need to un-do when `__btree_conf` fails with the failure of  `__wt_encryptor_config`.

sueloverso:
Thanks for tracking this problem I mentioned on the call @sulabhM.

sulabhM:
Closing this pull request as per discussion on the jira ticket.

PR #3459, opened 2017-06-08
WT-3362 Checkpoints shouldn't block drops.

michaelcahill:
Testing has uncovered another case where drops can spin trying to lock a checkpoint handle until a checkpoint completes.  This change fixes that in two ways: attempting to lock (but not open) a handle won't spin, and drop will always attempt to lock the live tree before locking any checkpoint handles.

### DIFF
--- a/src/session/session_dhandle.c
+++ b/src/session/session_dhandle.c
@@ -229,7 +229,8 @@ __wt_session_lock_dhandle(
 			WT_ASSERT(session, !F_ISSET(dhandle, WT_DHANDLE_DEAD));
 			return (0);
 		}
-		if (ret != EBUSY || (is_open && want_exclusive))
+		if (ret != EBUSY || (is_open && want_exclusive) ||
+		    LF_ISSET(WT_DHANDLE_LOCK_ONLY))
 			return (ret);
 		lock_busy = true;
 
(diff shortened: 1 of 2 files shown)


src/session/session_dhandle.c
```diff
@@ -229,7 +229,8 @@ __wt_session_lock_dhandle(
 			WT_ASSERT(session, !F_ISSET(dhandle, WT_DHANDLE_DEAD));
 			return (0);
 		}
-		if (ret != EBUSY || (is_open && want_exclusive))
+		if (ret != EBUSY || (is_open && want_exclusive) ||
+		    LF_ISSET(WT_DHANDLE_LOCK_ONLY))
```
agorrod: Looks to me as though ``WT_DHANDLE_LOCK_ONLY`` is set when calling ``__wt_session_lock_dhandle`` for checkpoints as well - is changing the semantic for checkpoints an expected consequence.

Answering my own question.. checkpoint is asking for exclusive access, so would already take this path.

agorrod:
Thanks @michaelcahill this change lgtm. I understand from the ticket that there is also a test change coming to make this complete, but I'd like to get this merged ASAP so we can get more test coverage.
I'm going to merge this one and open a new ticket for the test case.

PR #3460, opened 2017-06-09
SERVER-29439 WiredTiger turtle file "MoveFileExW: Access is denied." error.

keithbostic:
I believe the problem was that MongoDB reads the metadata file, and that potentially races with checkpoints. I never caught it in the act, but it fits all the symptoms.

### DIFF
--- a/src/meta/meta_table.c
+++ b/src/meta/meta_table.c
@@ -267,7 +267,9 @@ __wt_metadata_search(WT_SESSION_IMPL *session, const char *key, char **valuep)
 		 * that Coverity complains a lot, add an error check to get some
 		 * peace and quiet.
 		 */
-		if ((ret = __wt_turtle_read(session, key, valuep)) != 0)
+		WT_WITH_TURTLE_LOCK(session,
+		    ret = __wt_turtle_read(session, key, valuep));
+		if (ret != 0)
 			__wt_free(session, *valuep);
 		return (ret);
 	}
--- a/src/meta/meta_turtle.c
+++ b/src/meta/meta_turtle.c
@@ -246,6 +246,9 @@ __wt_turtle_read(WT_SESSION_IMPL *session, const char *key, char **valuep)
 
 	*valuep = NULL;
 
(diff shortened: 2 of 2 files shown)


agorrod:
Thanks @keithbostic lgtm. I wonder why we never reproduced this on Linux - is there a different semantic?

Is it something we could add a test case for now that you understand it? You could possibly even craft something in Python that uses the checkpoint thread along with metadata cursor reads.

sueloverso:
It seems like something we could add to test/format, a periodic walk through the metadata.  That might have triggered @keithbostic's assertions.  Nice work finding this one.

keithbostic:
> I wonder why we never reproduced this on Linux - is there a different semantic? Eric pointed out that Linux FS semantics don't block access, whereas Windows does - since we rely on this being single writer multiple readers (I think), it'd be nice to enforce that explicitly.

As Eric said: Linux/POSIX generally supports file remove/rename when the file being removed, or the source or destination files being renamed, have open handles. Windows doesn't, which is why we saw it there.

This patch acquires spinlocks on all paths to read/update the turtle file, so we're moving to single reader, single writer plan. We only access the turtle file to open a metadata cursor, and we generally cache metadata cursors, so I don't think single-reader will be a performance problem. It would be trivial to switch to read/write locks and have multiple-reader/single-writer if there's any reason to do so.

> Is it something we could add a test case for now that you understand it?

I don't off-hand see a way to add a test: you have to be reading the turtle file at the precise moment a checkpoint comes through and does its rename.

keithbostic:
> It seems like something we could add to test/format, a periodic walk through the metadata.

We don't currently run test/format on Windows.

> Nice work finding this one.

Lots of credit to Dan on this one.

sueloverso:
Yes, but while it wouldn't have triggered the error, a test/format thread walking the metadata would have triggered your assertion with this thread and checkpoints enabled in test/format.

michaelcahill:
Thanks @keithbostic, lgtm.  I'll merge...

PR #3461, opened 2017-06-12
WT-3355 Fix log archiving corner cases around reconfigure

sueloverso:
@agorrod Please review this change.  It fixes the original failure that test/format found.  It does change the way the log file upgrade/downgrade works a little bit though.  If archiving is turned off, it does not archive log files on a downgrade.  Recovery will still work on a downgrade because it forces the checkpoint into the old-version log file and recovery will not attempt to read an earlier, new-version log file.  However someone attempting to run `printlog` with an older release will get an error until archiving is done.

Without this fix, incremental backup does not work as expected in the case of downgrade.  One must restart the incremental backup process with a full backup again on an error.

This seems the better choice since I'm not aware of any customers running with archiving turned off anyway.


agorrod:
Thanks @sueloverso this change lgtm. I hadn't thought about the consequences of downgrades and incremental backup before.

PR #3462, opened 2017-06-13
WT-3363 Add a test to confirm if operations block waiting for checkpoints to complete

daveh86:


daveh86:
@agorrod can you review this please?

daveh86:
Retest this please

agorrod:
Dave the title of this pull request isn't right. The ticket number looks wrong, and it says "block against with" - I don't know what that means.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+ * An array of the number of operations that each of the worker threads has
+ * performed.
+ */
+static uint64_t thread_counters[N_THREADS];
+static const char *uri;
+
+int
+main(int argc, char *argv[])
+{
+	static WT_EVENT_HANDLER event_handler = {
```
agorrod: I'm surprised this is within the ``main`` function - the example I found in ``csuite/wt2719_reconfig/main.c`` looks more correct to me.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+void* do_ops(void *);
+void op_bulk(WT_CONNECTION *);
+void op_bulk_unique(WT_CONNECTION *, int, int);
+void op_create(WT_CONNECTION *);
+void op_create_unique(WT_CONNECTION *, int, int);
+void op_cursor(WT_CONNECTION *);
+void op_drop(WT_CONNECTION *, int);
+
+#define	N_THREADS 10
+#define	RUNTIME 900.0
```
agorrod: Does this test need to run for 15 minutes to catch problems (presuming this is in seconds)?

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+} THREAD_ARGS;
+
+static pthread_rwlock_t single;
+/*
+ * Time delay to introduce into checkpoints in seconds. Should be at-least
+ * double the maximum time that any one of the operations should take. Currently
+ * this is set to 10 seconds and we expect no single operation to take longer
+ * than 5 seconds.
+ */
+static const uint64_t max_execution_time = 10;
```
agorrod: Why isn't this a #define?

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+
+/* Int used to make unique table names */
```
agorrod: You don't need to state the type of the variable here, but if you do you should use a full word; i.e ``Integer`` rather than ``Int``.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+static uint64_t uid = 1;
+
+/*
+ * An array of the number of operations that each of the worker threads has
+ * performed.
+ */
+static uint64_t thread_counters[N_THREADS];
```
agorrod: I'd have thought this would live in the ``THREAD_ARGS`` structures.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+static const char *uri;
```
agorrod: I'd expect this to always be accessed via the THREAD_OPTS structure, rather than shared via a global variable. It's not strictly necessary, but global variables are harder to maintain, and this isn't tied to the lifetime of the string which makes me nervous.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+	testutil_check(testutil_parse_opts(argc, argv, opts));
+	testutil_make_work_dir(opts->home);
+
+	testutil_check(wiredtiger_open(
+	    opts->home, &event_handler, "create,cache_size=1G,", &opts->conn));
+
+	uri = opts->uri;
+
+	testutil_check(pthread_create(
+	    &ckpt_thread, NULL, do_checkpoints, (void *)opts->conn));
```
agorrod: I expected the last parameter to be ``opts`` not ``opts->conn`` - that matches other uses.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+/* WiredTiger error handling function */
+int
+handle_error(WT_EVENT_HANDLER *handler,
+    WT_SESSION *session, int error, const char *errmsg)
+{
+	(void)(handler);
+	(void)(session);
+	(void)(error);
+
+	/* Ignore complaints about missing files. */
```
agorrod: I'd like more context about why it's OK to get missing file errors in this test program in the comment.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+
+	testutil_check(pthread_join(ckpt_thread, NULL));
+
+	printf("Success\n");
+
+	testutil_cleanup(opts);
+	return (EXIT_SUCCESS);
+}
+
+/* WiredTiger error handling function */
```
agorrod: It's good that you added comments for most of the functions, but it's common practice in WiredTiger to use block-style (i.e: multi-line) comments at the function level.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+		if ((ret = conn->open_session(conn, NULL, NULL, &session)) != 0)
+			testutil_die(ret, "conn.session");
+
+		if ((ret = session->checkpoint(session, config)) != 0)
+			if (ret != EBUSY && ret != ENOENT)
+				testutil_die(ret, "session.checkpoint");
+
+		if ((ret = session->close(session, NULL)) != 0)
+			testutil_die(ret, "session.close");
+		(void)time(&now);
```
agorrod: Shouldn't this be after the sleep below?

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+	int i;
+
+	/* Unused */
+	(void)args;
+
+	(void)time(&start);
+	(void)time(&now);
+
+	for (i = 0; i < N_THREADS; i++)
+		last_ops[i] = 0;
```
agorrod: You can replace this with a single ``memset`` call.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+				op_bulk_unique(conn,
+				    __wt_random(&rnd) & 1, arg->threadnum);
+				break;
+			case 5:
+				op_create_unique(conn,
+				    __wt_random(&rnd) & 1, arg->threadnum);
+				break;
+		}
+		/* Increment how many ops this thread has performed */
+		__wt_atomic_add64(&thread_counters[arg->threadnum], 1);
```
agorrod: Why use atomic add here, but not read atomically? I think this code will change around when you take my suggestion about using the THREAD_ARGS instead of a global.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+		(void)time(&now);
+	}
+
+	return (NULL);
+}
+
+/*
+ * There are 6 operations below. These are taken originally from the operations
+ * we do in test/fops and modified somewhat bit to avoid blocking states.
```
agorrod: "somewhat bit"

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,498 @@
...
+ * - op_bulk
+ * - op_bulk_unique
+ * - op_cursor
+ * - op_create
+ * - op_create_unique
+ * - op_drop
+ */
+
+void
+op_bulk(WT_CONNECTION *conn)
```
agorrod: It's sad that these functions are duplicated. Could we share the code better? Maybe the core functionality could move into ``test/utility`` somewhere?

src/txn/txn_ckpt.c
```diff
@@ -676,6 +677,15 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	saved_isolation = session->isolation;
 	full = idle = logging = tracking = false;
 
+#ifdef HAVE_VERBOSE
```
daveh86: This was meant to be HAVE_DIAGNOSTIC. Mistake on my part.

test/csuite/wt3363_checkpoint_op_races/main.c
daveh86: It can take some time for the underlying issues to surface (specifically those in WT-3362) as it was an alignment of operations executing at the right time. The worst case I saw was just over 5 minutes before we saw a reproduction. I made this 15 minutes to honor the "three times" rule of reproducing a bug.

test/csuite/wt3363_checkpoint_op_races/main.c
daveh86: Agree and this was something I considered.

Ultimately, if we were to go down this path I believe we would need to re-write FOPS.

The functions in FOPS all take no arguments and use global variables in a number of places. There are also some changes to the configs used in these functions.

I'm not opposed to making this change, but think it may be better as another ticket.

daveh86:
@agorrod I have addressed your suggestions and added details. Can you take another look?

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,513 @@
...
+	NULL,
+	NULL
+};
+
+static pthread_rwlock_t single;
+
+/* Counter used to make unique table names. */
+static uint64_t uid = 1;
+
+static const char *uri;
```
agorrod: I'd made a comment here:

I'd expect this to always be accessed via the THREAD_OPTS structure, rather than shared via a global variable. It's not strictly necessary, but global variables are harder to maintain, and this isn't tied to the lifetime of the string which makes me nervous.

Was there a blocker to making this change?

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,513 @@
...
+/*
+ * WiredTiger error handling function.
+ */
+int
+handle_error(WT_EVENT_HANDLER *handler,
+    WT_SESSION *session, int error, const char *errmsg)
+{
+	(void)(handler);
+	(void)(session);
+	(void)(error);
```
agorrod: ``error`` is used, so doesn't need to be in this construct.

test/csuite/wt3363_checkpoint_op_races/main.c
daveh86: We discussed this today. Result was that we intend to move these functions into a test util thread function.

daveh86:
@agorrod, have made the changes here as discussed. Quite a major change from what I had before. Would you be able to take another look if you have time?

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,254 @@
...
+		thread_args[i].uid = &uid;
+		thread_args[i].testopts = opts;
+		thread_args[i].thread_counter = 0;
+		thread_args[i].threadnum = i;
+		testutil_check(pthread_create(
+		    &threads[i], NULL, do_ops, (void *)&thread_args[i]));
+	}
+
+	testutil_check(
+	    pthread_create(&mon_thread, NULL, monitor, &thread_args));
```
agorrod: The parameter being passed to the monitor thread deserves a comment. It's odd that it's different to the other threads.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,254 @@
...
+			 * cause of the hang can be investigated.
+			 */
+			if (thread_counter != last_ops[i])
+				last_ops[i] = thread_counter;
+			else {
+				printf("Thread %d had a task running"
+				    " for more than %d seconds\n",
+				    i, MAX_EXECUTION_TIME/2);
+				abort();
+
```
agorrod: There is an extra newline here.

test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -0,0 +1,254 @@
...
+/*
+ * Time delay to introduce into checkpoints in seconds. Should be at-least
+ * double the maximum time that any one of the operations should take. Currently
+ * this is set to 10 seconds and we expect no single operation to take longer
+ * than 5 seconds.
+ */
+#define	MAX_EXECUTION_TIME 10
+#define	N_THREADS 10
+/*
```
agorrod: We generally leave a newline before multi-line comments. There are a number of cases where you don't below.

test/utility/test_util.h
```diff
@@ -183,6 +201,18 @@ void *dmalloc(size_t);
 void *drealloc(void *, size_t);
 void *dstrdup(const void *);
 void *dstrndup(const char *, size_t);
+/*
+ * The op_ functions can generate errors that we wish to ignore. We have handler
```
agorrod: It's better not to reference parameter/variable names in comments. The `op_` above doesn't read nicely it'd be better replaced with `The following operation `

dist/flags.py
```diff
@@ -8,6 +8,9 @@
 ###################################################
 # Internal routine flag declarations
 ###################################################
+    'diagnostic_timing_stress' : [
+        'DIAGNOSTIC_CHECKPOINT_SLOW',
```
agorrod: I think the `DIAGNOSTIC` word is the less useful descriptor for this flag. I'd prefer it be called `TIMING_STRESS_CHECKPOINT_SLOW`. Since this should be considered a "timing stress" mode.

This comment propagates to several different places in the code, they should all be changed.

src/txn/txn_ckpt.c
```diff
@@ -1697,3 +1700,32 @@ __wt_checkpoint_close(WT_SESSION_IMPL *session, bool final)
...
+ * __checkpoint_diagnostic_latency --
+ *	Optionally add a 10 second delay to a checkpoint to simulate a long
+ *	running checkpoint for debug purposes. The reason for this option is
+ *	finding	operations that can block while waiting for a checkpoint to
+ *	complete.
+ */
+static void
+__checkpoint_diagnostic_latency(WT_SESSION_IMPL *session)
+{
+#ifdef HAVE_DIAGNOSTIC
```
agorrod: I'd avoid this ifdef here now, adding a flag check in the code path shouldn't be noticeable since the code isn't performance critical.

agorrod:
Thanks @daveh86 lgtm - I'm going to hold out merging until tomorrow since I'd like the WT-3373 changes to have time to go through stress testing without other changes being merged.

agorrod:
@daveh86 Could you please resolve the merge conflicts in this branch and then merge?

daveh86:
@agorrod done.

PR #3463, opened 2017-06-13
WT-3367 Move to macOS 10.12

markbenvenuto:
Evergreen is moving all OSX machines from 10.10 to 10.12 to support XCode 8.

I updated wiredtiger to use 10.12 but ran into issues with DYLD_LIBRARY_PATH that I needed to use `install_name_tool` to workaround.

### DIFF
--- a/test/mciproject.yml
+++ b/test/mciproject.yml
@@ -95,6 +95,16 @@ tasks:
           script: |
             set -o errexit
             set -o verbose
+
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @markbenvenuto lgtm. It makes me a bit sad that we need to do that here, since I suspect it means developer users of WiredTiger will need to resolve this themselves in another way.

PR #3464, opened 2017-06-14
WT-3369 WT_CURSOR->uri should always match the URI used to open the cursor

ddanderson:


ddanderson:
retest this please

src/cursor/cur_table.c
```diff
@@ -988,6 +988,13 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 		    table->cgroups[0]->source, NULL, cfg, cursorp);
 
 		__wt_schema_release_table(session, table);
+		if (ret == 0) {
+			/* Fix up the public URI to match what was passed in. */
+			cursor = *cursorp;
+			__wt_free(session, cursor->uri);
+			cursor->uri = NULL;
```
michaelcahill: This is redundant -- `__wt_free` is actually a macro that sets the pointer to NULL.

ddanderson: @michaelcahill, thanks.  I've fixed that, and also another issue with LSM ownership of the `WT_CURSOR->uri` field.

Can you take another look please?

michaelcahill:

PR #3465, opened 2017-06-15
WT-3370 Reset metafile after it could be re-allocated.

sueloverso:
@michaelcahill Please review this fix.  I considered getting rid of `metafile` and just using the array each time, but chose to just reset it as well as clear it out when we're done so future code that might use the variable later needs to pay attention.


agorrod:
Thanks @sueloverso lgtm

PR #3467, opened 2017-06-17
WT-1939 Improve error handling in example code

keithbostic:
@agorrod, @sueloverso: I took a run at cleaning up the example code error handling.

I hope this resolves WT-3364 and WT-1939.

I ended up using Alex's technique from the test/utility code, adding two macros: first, a macro `error_check` that checks for a non-zero return from a function call, and second, a macro `scan_end_check` which checks that `ret == WT_NOTFOUND` after a cursor scan completes.

I decided against adding explicit error checking with messages because I doubt our ability to ever exercise those error message paths and the error handling messages clutters the documentation more than I wanted.

I chose the names `error_check` and `scan_end_check` to hopefully make it clear in the documentation what's going on, happy to switch to something else if anybody has a better idea.

I found and fixed a few places in the example code where we weren't checking error returns and the example code fails, but there were two I didn't fix:

* @sueloverso, there's a `session->transaction_sync` call in `ex_sync.c` that returns failure, can you please take a look and fix it?
* @ddanderson, there's a `join_cursor->next` call in `ex_stat.c` that returns failure, can you please take a look and fix it?

Both of those calls are currently commented out so the tests pass, just search for `KEITH` in the file to find the call.


examples/c/ex_backup.c
```diff
@@ -205,126 +187,117 @@ take_full_backup(WT_SESSION *session, int i)
...
 	WT_SESSION *session;
-	int i, ret;
+	int i;
 	char cmd_buf[256];
 
+	(void)argc;					/* Unused variable */
+	(void)testutil_set_progname(argv);
+
 	(void)snprintf(cmd_buf, sizeof(cmd_buf),
 	    "rm -rf %s && mkdir %s", home, home);
```
ddanderson: Could this be replaced by `home = example_setup(argc, argv);` like other examples?

keithbostic: I think so, but I'll leave that up to @sueloverso: `ex_backup.c` has its own set of 3 databases, and the naming is clearly intended to align between the 3, so I was hesitant to change it.

agorrod:
Thanks @keithbostic I really like the direction you went with this change.

FYI: Sue is on vacation this week - so this is likely to sit until she gets back.

examples/c/ex_backup.c
sueloverso: That test does a lot of directory manipulation.  It actually uses a larger number as it creates a differently numbered incremental database directory for each successive pass.  I suggest leaving it alone and it is likely a one-off and its usage is not worth making into a higher level function.

sueloverso:
@keithbostic I pushed a fix for the `ex_sync` failure.

examples/c/ex_call_center.c
```diff
@@ -95,84 +89,71 @@ main(void)
...
-		home = NULL;
-
-	if ((ret = wiredtiger_open(home, NULL, "create", &conn)) != 0) {
-		fprintf(stderr, "Error connecting to %s: %s\n",
-		    home == NULL ? "." : home, wiredtiger_strerror(ret));
-		return (EXIT_FAILURE);
-	}
-	/* Note: further error checking omitted for clarity. */
+	home = example_setup(argc, argv);
+
```
sueloverso: very minor nit: since `home` is used in the next line, I'd remove the blank line.

keithbostic: Fixed.

PR #3470, opened 2017-06-21
WT-3336 Enable statistics by default in wtperf

daveh86:


daveh86:
@michaelcahill would you mind taking a look at this? 

I've added statistics_log to wtperf by default and removed statistics_log from all the runner files.

agorrod:
@daveh86 Did you check the perf jobs to see if this change causes any variation?

daveh86:
This does have a small impact on performance in those tests. To the tune of about 5% in the worst case.

We could change the scope of the initial change to make the stats 5 seconds by default, which would likely alleviate the worst case. @agorrod do you think that the 5% hit is too much?

daveh86:
retest this please

agorrod:
I've moved this conversation to https://jira.mongodb.org/browse/WT-3336

daveh86:
I'm closing this PR based on the discussion in Jira.

PR #3471, opened 2017-06-22
WT-3379 Avoid a performance regression on secondaries.

michaelcahill:
Continue with update/restore eviction unless we are going to fall back
to using the lookaside table.


michaelcahill:
Thanks for the review @keithbostic, merging...

PR #3472, opened 2017-06-22
WT-3379 Avoid a performance regression on secondaries.

keithbostic:
@michaelcahill, I wrote up what I think is going on, and ordered the tests to try and group tests with the same purpose together.

Feel free to discard this -- it's really, really wordy. (But, on the other hand, I think I really understand it now.)


michaelcahill:
Thanks @keithbostic, I edited slightly but lgtm.  I'll merge...

PR #3473, opened 2017-06-23
WT-3181 Fix config string for timestamps in test/checkpoint.

michaelcahill:


michaelcahill:
There is more work to to for timestamps in `test/checkpoint`, but I'm going to go ahead and merge this simple fix now to unblock testing.

PR #3474, opened 2017-06-23
WT-3140 Add output of handle name to per-handle stats

daveh86:


daveh86:

dist/flags.py
```diff
@@ -142,6 +142,7 @@
     'stat' : [
         'STAT_CLEAR',
         'STAT_JSON',
+        'STAT_JSON_PRINT_TABLES',
```
agorrod: This new flag feels a bit out of place. The others correspond 1-1 with options available via the public statistics configuration options, wheras this tracks internal state during a single stats pass.

I'm kind of concerned that the code is setting and clearing the flag that's in the global `WT_CONNECTION` handle, when it's relevant for a single pass of statistics gathering which is at the per-cursor granularity. i.e: I think it's possible to have two cursors gathering stats at the same time, and I think their setting/clearing of the new flag in this set of flags would interfere with each other.

src/conn/conn_stat.c
```diff
@@ -221,6 +221,101 @@ err:	__stat_sources_free(session, &sources);
...
+	WT_CONNECTION_IMPL *conn;
+
+	conn = S2C(session);
+
+	/*
+	 * If printing the connection stats, write that header and we are done.
+	 */
+	if (conn_stats) {
+		WT_RET(__wt_fprintf(
+		    session, conn->stat_fs, ",\"wiredTiger\":{"));
```
agorrod: I know that `wiredTiger` is camel case, but I prefer `WiredTiger` here and `WiredTigerTables` below. Remembering to start the storage engine with a lower case `w` always seems wrong.

dist/flags.py
ddanderson: @daveh86, I'd suggest instead of creating the `STAT_JSON_PRINT_TABLES` flag, adding a `bool` field `WT_CONNECTION.stat_json_tables` where all the `stat_` fields are defined.  It's functionally the same as the flag, but moving it connects it to those fields.  To partially answer concern raised by @agorrod, there would be a problem if we ever made this stat collection multithreaded, but then `WT_CONNECTION.stat_fs` could not be used as is, if it were multithreaded.  I think grouping together the offenders makes it more obvious.

test/suite/test_stat_log02.py
```diff
@@ -0,0 +1,100 @@
...
+        return None
+
+    def setUpSessionOpen(self, conn):
+        return None
+
+    def test_stats_log_json(self):
+        self.conn = self.wiredtiger_open(
+            None, "create,statistics=(fast),statistics_log=(wait=1,json)")
+        # Wait for the default interval, to ensure stats have been written.
+        time.sleep(2)
```
agorrod: I'm concerned that 2 seconds won't be long enough on all our test machines. Could you poll the file once per second for up to 30 seconds instead? Or alternatively there is the `on_close` configuration that should guarantee something is written after connection close.

test/suite/test_stat_log02.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # Sleep to let stats generate
+        time.sleep(2)
+
+        self.close_conn()
+        self.check_stats_file(".")
+        self.check_file_contains_tables(".")
+
+    def check_stats_file(self, dir):
+        files = glob.glob(dir + '/' + 'WiredTigerStat.[0-9]*')
```
agorrod: Is this going to work if multiple sets of stats get into the file? Will it still look like a single valid json doc?

src/include/connection.h
```diff
@@ -302,6 +302,8 @@ struct __wt_connection_impl {
 	WT_CONDVAR	*stat_cond;	/* Statistics log wait mutex */
 	const char	*stat_format;	/* Statistics log timestamp format */
 	WT_FSTREAM	*stat_fs;	/* Statistics log stream */
+	/* Statistics log print json formatted tables */
```
agorrod: Can you reword this comment - it makes it sound like somewhere I'd check to see if we are printing json formatted tables, wheras it's recording state during printing of json formatted tables.

test/suite/test_stat_log02.py
daveh86: The test is designed to work when we have multiple lines of stats in one file.

agorrod:
Thanks @daveh86 lgtm

PR #3476, opened 2017-06-23
WT-3181 Keep tests compatible with Python 2.6.

michaelcahill:


michaelcahill:
This fix is trivial (converting a Python 2.7 expression in a test to the Python 2.6 equivalent): doing a self review to unblock testing.

PR #3478, opened 2017-06-25
WT-3373 Access violation due to a bug in internal page splitting

keithbostic:
@agorrod, @michaelcahill: I believe this is a minimal fix for WT-3373.


agorrod:
Thanks @keithbostic lgtm, and the comment is excellent. I've run local testing and can no longer reproduce the issue.

PR #3479, opened 2017-06-25
WT-3389 Restructure split code to hold a split generation for the entire operation.

keithbostic:
@agorrod, @michaelcahill, this branch is a different approach to WT-3373 (PR #3478).

Basically, I changed things so that we hold a split generation across the complete insert operation, I think it's going to be safer/cleaner to do that, than wrap the single function the way I did in #3478.

As the underlying code acquires a split generation anyway, there shouldn't be any performance cost (other than potentially failing to evict internal pages as soon as would otherwise be possible). We can also simplify some of the underlying code this way.

There's another change in here: I don't think we need to hold a hazard pointer on the parent page anymore, I believe our reconciliation lock should be sufficient. (If I'm wrong about that, and it's not sufficient, I think we can get the same effect by hitting the parent page's split generation, which is going to be a lot cheaper/cleaner than dealing with hazard pointers here).

Anyway, for your consideration. Of course, this isn't a fix for WT-3373, we'd need a long bake-in period for this change.


keithbostic:
Something else that occurred to me while I was looking at this one: when doing a right-split (leaving the original page as a child of the original parent page), we don't set the split generation of the original page, as we do for the created pages. I'm pretty sure that's correct, but I'd like to hear someone else agree.

sulabhM:
@keithbostic , We merged WT-2309 (add delays during split to encourage races) into develop. I have updated this branch with a merge from develop. Please have a look and let me know if this looks good to you. I can move the delays around or add more of them if you suggest so.

michaelcahill:
@keithbostic, this branch lgtm.  I'll merge.

FTR, we started using a hazard pointer to protect the parent page of a split back in WT-960.  I think the reconciliation lock should be sufficient because the page is guaranteed to be dirty, so it cannot be evicted without reconciliation, which will wait to acquire the lock.

PR #3483, opened 2017-06-26
WT-3393 Missing barrier when a WT_UPDATE structure is appended to list.

keithbostic:


agorrod:
Thanks @keithbostic I agree that this change seems necessary, I'm going to merge.

@michaelcahill It's always worthwhile having another set of eyes look at barrier changes - would you mind taking a look when you are back?

PR #3485, opened 2017-06-27
WT-3363 Change how we check for the long tests flag. Disable the chec…

daveh86:
…kpoint races test unless explicitly requested


daveh86:
This has the intended impact on the check-long task.

http://build.wiredtiger.com:8080/job/wiredtiger-test-check-long/545/console

agorrod:
Thanks @daveh86 lgtm.

PR #3489, opened 2017-06-28
WT-3039 Add uniqueness to python list of log files.

sueloverso:
This accounts for case insensitivity.
@keithbostic Please review.  The new `test_compat01.py` is the first test to need to know how many log files there were and *Log* is case insensitive on Windows and was also grabbing the pre-allocated log files causing the test to fail on Windows.


sueloverso:
self-reviewing the test changes so that jobs work again.

keithbostic:
lgtm.

PR #3490, opened 2017-06-29
WT-3398 in-memory configurations don't support schema_worker operations

keithbostic:
@sueloverso, I changed in-memory configurations to ignore `WT_SESSION.rebalance`.

I suspect you'll need to do the same thing for `WT_SESSION.alter` in WT-3310 branch. I don't see a simple alternative, and absent a reason to make a `__wt_schema_worker` based operation work in-memory, I'm inclined to leave it alone.

### DIFF
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -1049,10 +1049,7 @@ __session_salvage(WT_SESSION *wt_session, const char *uri, const char *config)
 
 	SESSION_API_CALL(session, salvage, config, cfg);
 
-	if (F_ISSET(S2C(session), WT_CONN_IN_MEMORY))
-		WT_ERR_MSG(session, ENOTSUP,
-		    "WT_SESSION.salvage not supported for in-memory "
-		    "configurations");
+	WT_ERR(__wt_inmem_unsupported_op(session, NULL));
 
 	/* Block out checkpoints to avoid spurious EBUSY errors. */
 	WT_WITH_CHECKPOINT_LOCK(session,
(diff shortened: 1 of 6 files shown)


src/session/session_api.c
```diff
@@ -700,6 +700,10 @@ __session_rebalance(WT_SESSION *wt_session, const char *uri, const char *config)
 
 	SESSION_API_CALL(session, rebalance, config, cfg);
 
+	/* In-memory doesn't support rebalance operations. */
+	if (F_ISSET(S2C(session), WT_CONN_IN_MEMORY))
+		goto err;
+
```
sueloverso: Should this go down into `wt_schema_worker` instead of faulting it into each caller?  There are several callers of it in `session_api.c` that require exclusive access, rebalance, salvage, upgrade, verify and alter (on my branch).  Salvage has an error check.  The others do not.  Isn't it also needed on upgrade and verify? It should also be documented in the docs that some functions are invalid with in-memory in docs/in_memory.dox.

keithbostic: I don't think so. I think it's possible that at some point in the future we might need schema-worker to work on in-memory configurations, so I'm hesitant to have it enforce that level of policy.

I reviewed where schema-worker is called from:
* rebalance, ignored in `__session_rebalance`
* salvage, failed in `__session_salvage`
* upgrade, failed in `__session_upgrade` (added in 3a6dc62) 
* verify, failed in `__session_verify`
* compact, ignored in `__wt_session_compact`
* checkpoint, failed in `__session_checkpoint`
* backup cursors, failed in `__backup_start` (added in c5e9978)

PR #3493, opened 2017-07-02
WT-3402 Move cached overflow records to the update list.

keithbostic:
One of the problems I'm running into in [WT-3338](https://jira.mongodb.org/browse/WT-3338) (storing delta changes in WT_UPDATE lists as part of the partial-update optimizations), is dealing with delta records that modify overflow values that were removed, but had to cached for older readers. This change simplifies that partial update problem by moving cached overflow values into the WT_UPDATE list instead of maintaining them in a separate list.

I'm pushing it separately because it's a relatively isolated change with its own tradeoffs.

This should be an overall improvement:
* we're no longer acquiring any transactional state in order to track a cached overflow value,
* we're allocating less memory to store cached overflow values,
* we don't have to search a separate cache list to find the values. (In fact, I've never actually seen a lookaside into the cached list fire in testing. Because of the operational ordering, it's really unlikely a reading thread will race with the thread removing an overflow value.)

One downside is I added an array of paired pointers in the `WT_PAGE_MODIFY` structure that's never truncated, and it can theoretically grow to the number of overflow values on the page that have been deleted. I don't think that's likely to be a problem -- if reviewers disagree, we could add that size calculation into the page's memory footprint so the page would be evicted if the list grew too large.


keithbostic:
One other review note: the change in 452deba is scary and deserves some thought.

dist/stat_data.py
```diff
@@ -244,7 +244,6 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_inmem_splittable', 'in-memory page passed criteria to be split'),
     CacheStat('cache_lookaside_insert', 'lookaside table insert calls'),
     CacheStat('cache_lookaside_remove', 'lookaside table remove calls'),
-    CacheStat('cache_overflow_value', 'overflow values cached in memory', 'no_scale'),
```
agorrod: Please add a section to the upgrading docs, explaining that we removed these stats.

src/btree/bt_ovfl.c
```diff
@@ -109,51 +121,100 @@ __ovfl_cache_col_visible(
...
+__ovfl_cache_row_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 {
-	WT_UPDATE *upd;
-
 	/* Check to see if there's a globally visible update. */
-	for (upd = WT_ROW_UPDATE(page, rip); upd != NULL; upd = upd->next)
-		if (__wt_txn_upd_visible_all(session, upd))
+	for (; upd != NULL; upd = upd->next)
+		 if ((upd->type == WT_UPDATE_STANDARD ||
+		     upd->type == WT_UPDATE_DELETED) &&
```
agorrod: I'm nervous that this check is fragile if we add more update states, I'm also surprised that this change appears to be excluding `WT_UPDATE_RESERVED` - i.e: it's a behavior change for reserved updates, not required for overflow. Could the check be extracted out into a macro, or switched to `!WT_UPDATE_RESERVED`?

src/btree/bt_ovfl.c
```diff
@@ -109,51 +121,100 @@ __ovfl_cache_col_visible(
...
-	addr = unpack->data;
-	addr_size = unpack->size;
+	*updp = NULL;
 
+	/* Read the overflow value. */
 	WT_RET(__wt_scr_alloc(session, 1024, &tmp));
+	WT_ERR(__ovfl_read(session, unpack->data, unpack->size, tmp));
+
+	/*
+	 * Create an update entry with an impossibly low transaction ID to
```
agorrod: I prefer "no transaction ID" to "impossibly low" - that's how I think about it.

src/reconcile/rec_write.c
```diff
@@ -4905,16 +4862,12 @@ __rec_col_var(WT_SESSION_IMPL *session,
...
 		 * discard the underlying blocks, they're no longer useful.
-		 *
-		 * One complication: we must cache a copy before discarding the
-		 * on-disk version if there's a transaction in the system that
-		 * might read the original value.
 		 */
 		if (ovfl_state == OVFL_UNUSED &&
 		    vpack->raw != WT_CELL_VALUE_OVFL_RM)
-			WT_ERR(__wt_ovfl_cache(session, page, upd, vpack));
+			WT_ERR(__wt_ovfl_remove(session, page, upd, vpack));
```
agorrod: Do the two code paths that call `__wt_ovfl_remove` from reconciliation guarantee that the reconciliation is an update-restore? Otherwise I'm missing something about this change.

src/reconcile/rec_track.c
```diff
@@ -557,311 +557,12 @@ __wt_ovfl_reuse_free(WT_SESSION_IMPL *session, WT_PAGE *page)
 }
 
 /*
- * __ovfl_txnc_verbose --
```
agorrod: Seeing these functions removed made me do a happy dance...

src/btree/bt_ovfl.c
```diff
@@ -109,51 +121,100 @@ __ovfl_cache_col_visible(
...
+	 * ensure global visibility, append it to the update list.
+	 *
+	 * We don't need locks or barriers in this function: any thread reading
```
agorrod: I thought we only came through here via reconciliation, so the page should be in `WT_REF_LOCKED`, am I missing a path?

dist/stat_data.py
keithbostic: Done, pushed in 429b8e8.

src/btree/bt_ovfl.c
keithbostic: We have to ignore reserved records here because they can't be returned to the application, and so don't mean we won't have to cache the overflow value. I should have called it out in the commit, it's a bug fix -- we'll get into trouble if there's a reserve update here and we don't cache an overflow value because of it.


I've pushed a macro in 7fbb00af9 so adding more update states won't be fragile.

src/btree/bt_ovfl.c
keithbostic: Sure, pushed in dfe2ccb1e.

src/btree/bt_ovfl.c
keithbostic: Pages being checkpointed won't be `WT_REF_LOCKED`?

src/reconcile/rec_write.c
keithbostic: This doesn't just apply to update-restore reconciliation, it applies to any overflow value that's being removed by any reconciliation.

agorrod:
Thanks @keithbostic lgtm.

@michaelcahill This deserves a look after you return - I may have missed something.

@sulabhM Please proceed with the review - it's a good change to explore in terms of understanding another part of the reconciliation system.

PR #3494, opened 2017-07-03
WT-3409 WiredTiger generations can silently self-deadlock.

keithbostic:
Panic if we self-deadlock in the generation code.


agorrod:
Thanks @keithbostic this looks like a useful condition to catch.

PR #3495, opened 2017-07-03
WT-3406 Reconciliation is choosing reserved records for writing.

keithbostic:
@agorrod, when I wrote this code I assumed that the reserved records were updated to an aborted state before the committed transaction state can roll forward, but that's apparently not the case.

The update happens in `__wt_txn_commit()`:
```
    686         /* Process and free updates. */
    687         for (i = 0, op = txn->mod; i < txn->mod_count; i++, op++) {
    688                 switch (op->type) {
    689                 case WT_TXN_OP_BASIC:
    690                 case WT_TXN_OP_BASIC_TS:
    691                 case WT_TXN_OP_INMEM:
    692                         /*
    693                          * Switch reserved operations to abort to
    694                          * simplify obsolete update list truncation.
    695                          */
    696                         if (op->u.upd->type == WT_UPDATE_RESERVED) {
    697                                 op->u.upd->txnid = WT_TXN_ABORTED;
    698                                 break;
    699                         }
```
I'm not seeing where the committed transaction state is allowed to roll forward, do you know where that happens?


keithbostic:
@agorrod, I updated this fix: we can't simply ignore reserved updates, they have to be "skipped" (and a quick skim makes me think we probably want to include them in tracking the max/min transaction and timestamp values as well).

michaelcahill:
As discussed in the ticket, I don't think this fix is sufficient to fix the general problem.  For example, I think a similar race could allow aborted updates to be written to a page during lookaside eviction.  I'm going to close this pull request and push an alternative fix.

PR #3499, opened 2017-07-06
WT-3394 Fix compilation warnings for GCC-7

agorrod:
This was merged once before, but reverted due to segfaults in wtperf:
http://build.wiredtiger.com:8080/job/wiredtiger-perf-lsm/2792

And a build failure on Windows:
https://evergreen.mongodb.com/task/wiredtiger_windows_64_compile_windows_alt_d5f82a43f1e0c8aafd38f0098bc9349a6de335e6_17_07_06_04_52_16

@sulabhM Is looking at the wtperf failure, and I pushed a fix for the Windows issue.


keithbostic:
@agorrod, sorry for the breakage!

@sulabhM, I reproduced the failure and took a look at the core, and the problem with wtperf is aebd867, so I just reverted that change. (I started out changing loops that gcc 7's `-Wunsafe-loop-optimizations warning` complained about, but it turned out to be too noisy, so I eventually just turned that warning off.)

PR #3501, opened 2017-07-06
WT-3310 Fix test/format to tolerate EBUSY with LSM.

sueloverso:
@keithbostic Please take a look at this.  This fixes the several test/format failures when alter, LSM and compact are configured.  In that case, it is normal that EBUSY could be returned and test/format should not die in that case.

### DIFF
--- a/test/format/util.c
+++ b/test/format/util.c
@@ -480,6 +480,7 @@ WT_THREAD_RET
 alter(void *arg)
 {
 	WT_CONNECTION *conn;
+	WT_DECL_RET;
 	WT_SESSION *session;
 	u_int period;
 	bool access_value;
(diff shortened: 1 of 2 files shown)


test/format/util.c
```diff
@@ -505,8 +506,12 @@ alter(void *arg)
...
 		    access_value ? "random" : "none"));
 		access_value = !access_value;
-		if (session->alter(session, g.uri, buf) != 0)
-			break;
+		/*
+		 * Alter can return EBUSY if concurrent with other operations.
+		 */
+		while ((ret = session->alter(session, g.uri, buf)) != 0 &&
+		    ret != EBUSY)
+			testutil_die(ret, "session.alter");
```
keithbostic: Maybe the `__wt_yield` calls for compact, backup & alter should be sleep? With yield (and we don't even have a yield in the alter call), we're just pounding on the system.

PR #3503, opened 2017-07-06
WT-3387 Add use_timestamp option for checkpoint

sueloverso:
@agorrod This is not ready for merge but I wanted to open a PR in case anyone needs to see this code while I'm away.


michaelcahill:
@sueloverso, what I'm expecting with checkpoint as-of a timestamp is that we will add a `stable_timestamp` that can be set via `WT_CONNECTION::set_timestamp` -- it can be no older than `oldest_timestamp` and probably in the MongoDB case they will usually be the same, but they could conceivably diverge.

Checkpoints will be taken as-of the stable_timestamp, if one is set when the checkpoint transaction starts.

Could we clear `stable_timestamp` to get the effect of `use_timestamp=false`?

sueloverso:
@michaelcahill I have added a framework and some usage for a `stable_timestamp` but it isn't fully complete yet.  It only sets the value in the structure and queries it for the checkpoint.  I've expanded the test to test that however and it is working.  (The test also has a workaround until WT-3440 is merged.). You should review that this is going in the right direction.

src/txn/txn_ckpt.c
```diff
@@ -580,11 +582,30 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, const char *cfg[])
...
+	 * Someone giving us a specific timestamp overrides the general
+	 * use_timestamp.
+	 */
 	WT_RET(__wt_config_gets(session, cfg, "read_timestamp", &cval));
 	if (cval.len > 0) {
 		WT_RET(__wt_snprintf(timestamp_config, sizeof(timestamp_config),
 		    "read_timestamp=%.*s", (int)cval.len, cval.str));
 		txn_cfg[2] = timestamp_config;
+	} else if (txn_global->has_oldest_timestamp) {
+		WT_RET(__wt_config_gets(session, cfg, "use_timestamp", &cval));
```
agorrod: I'm concerned about the order of checking conditions in here. If an application wants to create a checkpoint at a timestamp, but there is no oldest_timestamp set (maybe it's creating a checkpoint shortly after startup?), then we'll create a checkpoint that includes all committed changes. Is that a non-issue because as soon as we make an update that has an associated timestamp, we are guaranteed that `has_oldest_timestamp` is true? I don't think that's true, since setting a commit timestamp doesn't force refreshing the global transaction state which is required to set the global timestamps.

sueloverso: While not directly related to your question, this code has a bug, in that it should be using `has_stable_timestamp` now.  And you're right about needing to figure out what that use case should be where the user asks for a timestamp and none has been set.  At a minimum here I'll document what happens on a mismatch.

sueloverso:
@agorrod you should look at this again as I made a number of changes today, mostly error checking but enough to have another look.

sueloverso:
@michaelcahill and @agorrod I do want to point out that the new `use_timestamp` configuration defaults to `true` which is a change over what current develop does today with timestamps.  If a stable timestamp is in use, the user will automatically get that.  Since this hasn't been released yet it doesn't affect anyone but I wanted to point it out.

agorrod:
lgtm

michaelcahill:
Thanks @sueloverso, merging.  I've opened WT-3358 for the followup work to have checkpoint ignore `stable_timestamp` for logged tables.

PR #3504, opened 2017-07-07
WT-3417 Drain transactions during upgrade/downgrade.

keithbostic:
@sueloverso, for your consideration.

I'm not sure where we want to go with testing on this one, let's talk that over.


sueloverso:
@keithbostic I took a first look at this and I think it will work for you.  I don't have time to do a full/formal review before I leave.  As for testing, there is `test_compat*.py` and a `compatibility` setting in `test/format`.  Perhaps one of those could be expanded to verify the right thing is happening (maybe add a temporary or real stat to indicate we hit the need-to-drain condition).  Also once the new partial update log record is added, then making a test with multiple threads doing that, and another thread downgrading and making sure no V1 log file ends up with that V2 log record in it would be useful.

michaelcahill:
After discussion, closing in favor of #3583.

PR #3505, opened 2017-07-10
WT-3421 Fix unreachable code error exposed when diagnostic build is off.

ddanderson:


test/csuite/wt3363_checkpoint_op_races/main.c
```diff
@@ -84,7 +84,8 @@ main(int argc, char *argv[])
 	 * diagnostic mode is not enable and runs for 15 minutes.
 	 */
 #if !defined(HAVE_DIAGNOSTIC)
-	return (EXIT_SUCCESS);
+	if (true)
+		return (EXIT_SUCCESS);
```
keithbostic: This is going to need a comment, otherwise someone is going to remove the test.

PR #3509, opened 2017-07-13
WT-3380 Make 8-byte timestamps a special case

sulabhM:


src/include/misc.h
```diff
@@ -257,15 +257,22 @@
 #endif
 
 #ifdef HAVE_TIMESTAMPS
-#define	WT_TIMESTAMP(x) (x)
-typedef uint8_t wt_timestamp_t[WT_TIMESTAMP_SIZE];
+#define	WT_TIMESTAMP_PTR(x) (&x)
```
agorrod: @keithbostic We touched on this code change last night when we talked. I wanted to say why I pushed @sulabhM towards the structure we have here. What I want to avoid is having to use

```
#if WT_TIMESATMP_SIZE == 8
#elif HAVE_TIMESTAMPS
#endif
```

For every place that deals with timestamps. What I really want to be able to do is to only care that the size is 8 in cases where it's performance critical. Sulabh suggested just casting the array, but I don't believe that is going to  guarantee well enough an atomic read/write, which is what we need.

It would be great if you can let us know your thoughts on how you'd approach this change.

keithbostic: @agorrod, I took a run at this one yesterday.

First, I agree with you and @sulabhM, I couldn't find any way to do this without switching to passing around pointers, everything I tried generated a total mess.

After some thought, I think the simplest approach is to not try and use a union, but instead to use separate typedefs and hide the magic of acquiring a pointer to the timestamp in the `WT_TIMESTAMP` macro.

I've pushed a new branch, PR #3510, for your consideration.

sulabhM:
Note to self: python test test_timestamp01 doesn't handle large timestamp sizes.

src/btree/bt_debug.c
```diff
@@ -1008,22 +1008,17 @@ __debug_update(WT_DBG *ds, WT_UPDATE *upd, bool hexbyte)
 		WT_RET(ds->f(ds, "\t" "txn id %" PRIu64, upd->txnid));
 
 #ifdef HAVE_TIMESTAMPS
-		if (!__wt_timestamp_iszero(upd->timestamp)) {
+		if (!__wt_timestamp_iszero(WT_GET_TIMESTAMP_PTR(upd))) {
```
keithbostic: FWIW, I suggest we get rid of `WT_GET_TIMESTAMP_PTR` (it's only used in 7 places), and just spell out the `->timestamp` part.

I'd also prefer a different name for `WT_TIMESTAMP_PTR` (basically, we're conflating a bunch of stuff: adding `->timestamp`, adding `&`, and setting the argument to NULL, and that's hard to think about.

I'd suggest spelling out `->timestamp` and `&`, and then using something like `WT_TIMESTAMP_NULL` for setting the argument to NULL.

src/include/misc.h
```diff
@@ -251,21 +251,28 @@
...
 #ifdef HAVE_TIMESTAMPS
-#define	WT_TIMESTAMP(x) (x)
-typedef uint8_t wt_timestamp_t[WT_TIMESTAMP_SIZE];
-#define	WT_DECL_TIMESTAMP(x) wt_timestamp_t x;
+struct __wt_timestamp_t {
+	union {
+		uint64_t val;
+		uint8_t ts[WT_TIMESTAMP_SIZE];
+	};
+};
```
keithbostic: Do we need both the `struct` and the `union`, or would just the `union` suffice?

src/include/txn.i
```diff
@@ -10,58 +10,111 @@ static inline int __wt_txn_id_check(WT_SESSION_IMPL *session);
 static inline void __wt_txn_read_last(WT_SESSION_IMPL *session);
 
 #ifdef HAVE_TIMESTAMPS
+#if WT_TIMESTAMP_SIZE == 8
```
keithbostic: I added an additional macro `WT_TIMESTAMP_UINT64` because it made it easier to do testing, that is,  I could keep `WT_TIMESTAMP_SIZE == 8` (which our test programs expect), but switch to using the `uint8_t` array instead of the `uint64_t` value. It was just a testing hack, usefulness depends on your plans for testing.

src/txn/txn_timestamp.c
```diff
@@ -138,34 +170,57 @@ __wt_txn_global_query_timestamp(
...
+#endif
 	return (0);
 #else
 	WT_UNUSED(session);
 	WT_UNUSED(hex_timestamp);
 	WT_UNUSED(cfg);
 
-	return (ENOTSUP);
+	WT_RET_MSG(session, ENOTSUP, "WT_CONNECTION.query_timestamp "
+	    "requires a version of WiredTiger built with timestamp support");
```
keithbostic: I introduced a bug here, `session` is no longer unused.

src/include/txn.i
sulabhM: I am building with new size to test it out. But then I have only tested the python test suite as of now.

src/txn/txn_timestamp.c
sulabhM: fixed.

sulabhM:
@agorrod , Please have a look.

agorrod:
Thanks @sulabhM This is looking good to me now - could you please merge with develop?

@michaelcahill This one deserves your review please - it's a starting point for optimizing 8 byte timestamps. We've deliberately not included any of the non-necessary optimizations here. This change just contains the structural shuffling necessary to unlock the optimizations.

sulabhM:
autobuild: retest this please

michaelcahill:
@sulabhM, I'm pretty happy with this branch.

I made a suggestion in #3522 to avoid the warning about taking the address of a member of a packed structure (by removing the packed attribute).  @keithbostic will need to look at that as well, but once we decide which way to go I think this branch can be merged.

michaelcahill:
@sulabhM, you mentioned `test_timestamp01` not working with large timestamps above.  Can you please open a ticket on the backlog with the repro?

sulabhM:
autobuild: retest this please

PR #3510, opened 2017-07-15
WT-3380 Special case 8-byte timestamps

keithbostic:
Special case 8-byte timestamps.

### DIFF
--- a/build_posix/aclocal/strict.m4
+++ b/build_posix/aclocal/strict.m4
@@ -82,6 +82,7 @@ AC_DEFUN([AM_GCC_WARNINGS], [
 AC_DEFUN([AM_CLANG_WARNINGS], [
 	w="-Weverything -Werror"
 
+	w="$w -Wno-address-of-packed-member"
 	w="$w -Wno-cast-align"
 	w="$w -Wno-documentation-unknown-command"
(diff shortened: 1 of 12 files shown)


keithbostic:
@agorrod, @sulabhM, I should mention, I did not modify the locking in anyway.

I agree with @michaelcahill's comment in the JIRA ticket that we should be able to turn off some amount of locking in the 8-byte cases. My suggestion (without having actually tried it), would be to move the locking (and barrier instructions), down into the underlying timestamp functions.

keithbostic:
@agorrod, @sulabhM, another comment: we might want to enhance the Jenkins compile job to build without timestamps (`--with-timestamp-size=0`) and with a non-standard number of bytes (`--with-timestamp-size=37`).

But that only tells us there's a successful compile, it doesn't do any testing. I just pushed a change I've been using to make it easy to run the standard 8B case through both the `uint64_t` code and the `uint8_t` array code: if you `#undef WT_TIMESTAMP_UINT64` in `misc.h`, you're running the array code. We could surface something in the API to make that testing easier, if we decide to go with this branch.

agorrod:
I'm going to close this branch out - the changes in here have been merged into the alternative implementation.

PR #3511, opened 2017-07-17
WT-3381 Improve timestamp concurrency.

michaelcahill:
Removes timestamps from `WT_TXN_STATE` as a step towards merging `WT_TXN_STATE` with `WT_TXN`.

This change has transactions add themselves to global, sorted lists (one ordered by commit timestamp, the other ordered by read timestamp).  Each list has its own rwlock, and scans have been replaced by peeking at the first (oldest) transaction in the list.  The list's lock protects the
relevant fields in `WT_TXN` (i.e., no thread will read `txn->commit_timestamp` unless the transaction is in the `commit_timestamp` list).

This should reduce contention for `txn_global->rwlock`, which can be further decomposed in future (e.g., eliminating the scan for `__wt_txn_update_oldest` and replacing `__wt_txn_get_snapshot` with a loop that makes a copy of running transaction IDs in order).

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -80,9 +80,8 @@ __txn_global_query_timestamp(
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_CONFIG_ITEM cval;
+	WT_TXN *txn;
 	WT_TXN_GLOBAL *txn_global;
-	WT_TXN_STATE *txn_state;
-	uint32_t i, session_cnt;
 
 	conn = S2C(session);
 	txn_global = &conn->txn_global;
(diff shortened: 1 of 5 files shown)


michaelcahill:
@agorrod, sys-perf tests against MongoDB master show that this branch addresses the regression introduced by #3500.  Can you please take a look?

src/txn/txn_timestamp.c
```diff
@@ -306,3 +297,109 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+	    prev->commit_timestamp, txn->commit_timestamp) > 0;
+	    prev = TAILQ_PREV(prev, __wt_txn_cts_qh, commit_timestampq))
+		;
+	if (prev == NULL)
+		TAILQ_INSERT_HEAD(
+		    &txn_global->commit_timestamph, txn, commit_timestampq);
+	else
+		TAILQ_INSERT_AFTER(&txn_global->commit_timestamph,
+		    prev, txn, commit_timestampq);
+	__wt_writeunlock(session, &txn_global->commit_timestamp_rwlock);
```
agorrod: I think some statistics in here would be interesting - knowing how the queues are being used could give us a way to validate intuition. I'm happy for that to be another ticket.

PR #3512, opened 2017-07-17
WT-3418 Fix a block manager race in tree close/open

keithbostic:
@sueloverso, for your consideration.

I don't see any reason not to delay setting the `WT_DHANDLE_DEAD` flag until after the block object is discarded, it's a cheap operation.

I don't think this helps with WT-3232 because my reading of that ticket was the cache blocks themselves need to be discarded (although I don't understand why that's the case).

### DIFF
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -188,18 +188,20 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
 	if (!F_ISSET(btree,
 	    WT_BTREE_SALVAGE | WT_BTREE_UPGRADE | WT_BTREE_VERIFY)) {
 		/*
-		 * If the object is already marked dead, we're just here to
+		 * If the handle is already marked dead, we're just here to
 		 * discard it.
 		 */
 		if (F_ISSET(dhandle, WT_DHANDLE_DEAD))
 			discard = true;
 
 		/*
-		 * If a standard object and it's OK to mark the handle dead
-		 * (letting the tree be discarded later), do so. Don't mark
-		 * memory-mapped trees dead (we close the file handle to allow
-		 * the file to be removed, but mapped trees contain pointers
-		 * into memory that become invalid if the mapping is closed).
+		 * Mark the handle dead (letting the tree be discarded later) if
+		 * it's not already marked dead, our caller allows it, it's not
+		 * a final close, and it's not a memory-mapped tree. (We can't
+		 * mark memory-mapped tree handles dead because we close the
+		 * underlying file handle to allow the file to be removed and
+		 * memory-mapped trees contain pointers into memory that become
+		 * invalid if the mapping is closed.)
 		 */
 		if (!discard && force && !final &&
 		    (bm == NULL || !bm->is_mapped(bm, session)))
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -144,10 +144,12 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
 {
 	WT_BM *bm;
 	WT_BTREE *btree;
+	WT_CONNECTION_IMPL *conn;
 	WT_DATA_HANDLE *dhandle;
 	WT_DECL_RET;
-	bool marked_dead, no_schema_lock;
+	bool discard, marked_dead, no_schema_lock;
 
+	conn = S2C(session);
 	btree = S2BT(session);
 	bm = btree->bm;
 	dhandle = session->dhandle;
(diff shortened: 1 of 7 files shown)


keithbostic:
@sueloverso, I think PR #3512 (WT-3418) is ready for you.

sueloverso:
@keithbostic can you explain the race where we ended up with two dhandles and how this fixes it (i.e. where was the bug exactly)?

keithbostic:
@sueloverso 
> can you explain the race where we ended up with two dhandles and how this fixes it (i.e. where was the bug exactly)?

My apologies for not being clear. This is the problem you described in WT-3232:

> The problem is with the meaning of WT_DHANDLE_DEAD. We set that flag at the beginning of the process of making a handle dead. So, in WT-2990, without the lock being held for the entire process of making a handle dead, one thread saw the WT_DHANDLE_DEAD flag set before the underlying btree and blocks were done being discarded/closed, etc and went to open another handle of the same file and found an in-use block manager. 

The heart of the change is to finish closing the btree/block objects before setting `WT_DHANDLE_DEAD`.

It gets messy because:
* `__wt_conn_btree_sync_and_close` is called more than once when used by sweep (and `WT_DHANDLE_DEAD` is already set the second time it's called),
* `__wt_conn_btree_sync_and_close` is called a single time to really close things down when the database is closed,
* code we call to clear out the cache and discard pages expects `WT_DHANDLE_DEAD` to be set in order to discard modified pages, and
* `__wt_checkpoint_close` "owned" converting non-durable handles into dead handles.

I think what I've done here is better than what we had before, but I'm still not very happy with it. It may be that I don't understand the handle code well enough to see the better solution (cc: @michaelcahill).

src/conn/conn_dhandle.c
```diff
@@ -180,45 +181,87 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
...
+	/* Reset the tree's eviction priority (if any). */
+	__wt_evict_priority_clear(session);
+
+	discard = marked_dead = false;
 	if (!F_ISSET(btree,
 	    WT_BTREE_SALVAGE | WT_BTREE_UPGRADE | WT_BTREE_VERIFY)) {
-		if (force && (bm == NULL || !bm->is_mapped(bm, session))) {
-			F_SET(session->dhandle, WT_DHANDLE_DEAD);
+		/*
+		 * If the object is already marked dead, we're just here to
```
sueloverso: While 'object' is not incorrect, we tend to call them handles in this code.

src/conn/conn_dhandle.c
```diff
@@ -180,45 +181,87 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
...
+		 * discard it.
+		 */
+		if (F_ISSET(dhandle, WT_DHANDLE_DEAD))
+			discard = true;
+
+		/*
+		 * If a standard object and it's OK to mark the handle dead
```
sueloverso: I find the phrase "If a standard object" awkward.  The meaning of "standard object" isn't clear nor is it obvious in the conditional below what that would mean. (And it probably should say "If this is a...").  The "it's OK to mark..." refers to a 5-part conditional.  I'd prefer something more obvious like "It is okay to make the handle dead only when..."

src/conn/conn_dhandle.c
```diff
@@ -180,45 +181,87 @@ __wt_conn_btree_sync_and_close(WT_SESSION_IMPL *session, bool final, bool force)
...
-			__wt_evict_priority_clear(session);
-		}
-		if (!marked_dead || final) {
-			if ((ret = __wt_checkpoint_close(
-			    session, final)) == EBUSY)
-				WT_ERR(ret);
-			else
-				WT_TRET(ret);
+		/*
+		 * Get rid of any durable objects we couldn't mark dead. Durable
```
sueloverso: Again just my preference object -> handle

src/txn/txn_ckpt.c
```diff
@@ -1665,18 +1665,6 @@ __wt_checkpoint_close(WT_SESSION_IMPL *session, bool final)
...
-	 * If the handle is already dead or the file isn't durable, force the
-	 * discard.
-	 *
-	 * If the file isn't durable, mark the handle dead, there are asserts
-	 * later on that only dead handles can have modified pages.
-	 */
-	if (F_ISSET(btree, WT_BTREE_NO_CHECKPOINT))
-		F_SET(session->dhandle, WT_DHANDLE_DEAD);
-	if (F_ISSET(session->dhandle, WT_DHANDLE_DEAD))
-		return (__wt_cache_op(session, WT_SYNC_DISCARD));
```
sueloverso: I assume this functionality is now taken care of in the new call to wt_cache_op in wt_conn_btree_sync_and_close.

src/conn/conn_dhandle.c
keithbostic: Fixed in f64c96448.

src/conn/conn_dhandle.c
keithbostic: Fixed in f64c96448.

src/conn/conn_dhandle.c
keithbostic: We're actually not discarding the handle here, we're discarding the underlying object.

src/txn/txn_ckpt.c
keithbostic: Yes.

keithbostic:
@michaelcahill, even if you don't do a detailed review on this one, can I get an lgtm? I don't know the handle code very well.

One specific question: the `force` argument to `__wt_conn_btree_sync_and_close` confuses me, it's not clear to me what exactly we're "forcing".

It seems to me the flag indicates if we're willing to mark a handle dead or not, which makes me think we could merge `final` and `force` into a single flag indicating if it's OK to mark the handle dead or not. That said, I'm sure I'm just missing something...

src/conn/conn_dhandle.c
michaelcahill: I agree with Sue that "object" is confusing. How about "Discard from cache any trees that aren't marked dead"?

src/session/session_api.c
```diff
@@ -166,8 +166,7 @@ __session_alter(WT_SESSION *wt_session, const char *uri, const char *config)
 	WT_WITH_CHECKPOINT_LOCK(session,
 	    WT_WITH_SCHEMA_LOCK(session,
 		ret = __wt_schema_worker(session, uri, __wt_alter, NULL, cfg,
-		WT_BTREE_ALTER |
-		WT_DHANDLE_DISCARD_FORCE | WT_DHANDLE_EXCLUSIVE)));
```
michaelcahill: @sueloverso, please stare at this.

The handle will be discarded regardless because `WT_BTREE_ALTER` is in `WT_BTREE_SPECIAL_FLAGS`.  I can't see any reason why mark the handle dead is preferable to ordinary (full) discard.  Please let me know if I've missed something.

michaelcahill: I should add that there's one remaining use of `WT_DHANDLE_DISCARD_FORCE`, which is in LSM.  I'm inclined to try to eliminate it and find some cleaner way for LSM to get handles purged from cache.

src/conn/conn_dhandle.c
keithbostic: Fixed in d0fd03923.

keithbostic:
@sueloverso, I've asked for a final review, based on Michael's comment.

keithbostic:
@michaelcahill 
> It is not possible to merge "mark_dead" with "final", AFAICT

I took another look and I agree. However:

* `__wt_conn_dhandle_discard` sets `final = true` and `mark_dead = F_ISSET(conn, WT_CONN_IN_MEMORY)`, which looks  wrong to me, `final` overrides `mark_dead`. I pushed a change for that in c46764f.

* `__wt_schema_drop` calls `__drop_file` calls `__wt_conn_dhandle_close_all` calls `__wt_conn_btree_sync_and_close` with `mark_dead = config("force")`, that is, if "force" is configured for drop, `mark_dead = true`. Why is "force" correlated to marking the handle dead?

src/session/session_api.c
sueloverso: Removing the DISCARD_FORCE is fine here.  When doing this I added ALTER to the special flags much later in the timeline and did not remember/realize and go back to delete the other flag.  I've been running alter tests for hours and all is well.  So this LGTM.

michaelcahill:
Forced drops were the reason we added the "marked dead" mode for data handles.  The intent of forced drops is that the data is never going to be needed because we're going to delete the file and we want to do that as quickly as possible (without waiting to evict the file from cache).

That's why the parameter name "force" was plumbed through from the drop code to handle close, but since it caused confusion in the handle management code, I changed it to "mark_dead" there.  The public API for forced drops won't be updated because the intent is still the same: the application wants to force a drop to happen ASAP.

src/conn/conn_dhandle.c
```diff
@@ -727,7 +727,7 @@ __wt_conn_dhandle_discard(WT_SESSION_IMPL *session)
 
 		WT_WITH_DHANDLE(session, dhandle,
 		    WT_TRET(__wt_conn_dhandle_discard_single(
-		    session, true, F_ISSET(conn, WT_CONN_IN_MEMORY))));
+		    session, true, false)));
```
michaelcahill: FTR, now that `__wt_conn_btree_sync_and_close` is checking `WT_CONN_IN_MEMORY` at https://github.com/wiredtiger/wiredtiger/pull/3512/files#diff-1724c58a241ea44b64a43d04fb861d5bR223, this change lgtm.

PR #3516, opened 2017-07-19
WT-3406 Reconciliation should ignore concurrent updates.

michaelcahill:


agorrod:
Thanks @michaelcahill. This change makes sense to me. @keithbostic would you mind doing a review as well?

src/reconcile/rec_write.c
```diff
@@ -1290,8 +1309,8 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 		}
```
keithbostic: I think we have the same problem in the checkpoint loop we have in the eviction loop: `__wt_txn_upd_visible` returns true and then `upd->txnid` is updated, leading to assertion failure at line 1312.

This one is a little trickier because `__wt_txn_upd_visible` takes a `upd` argument, again potentially leading to an invented load.

keithbostic: I think there's an additional problem in the checkpoint loop. The `__wt_txn_upd_visible` check doesn't include any check for `upd->txnid == WT_UPDATE_RESERVED`, which I think means we can legitimately return visibility for a reserved update, again leading to assertion failure at line 1312.

We could do:
```
if (upd->type == WT_UPDATE_RESERVED || txnid == WT_TXN_ABORTED)
```
or:
```
if (upd->type != WT_UPDATE_RESERVED && __wt_txn_upd_visible(session, upd))
```
so it looks like `__wt_txn_read()`.

src/reconcile/rec_write.c
```diff
@@ -1231,36 +1244,42 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 			if (WT_TXNID_LT(txnid, min_txn))
 				min_txn = txnid;
```
michaelcahill: We don't do this elsewhere and I don't think it is necessary here because there are rules around when `upd->txnid` can change that make it unnecessary.  Specifically, it is set to an active transaction ID before an update becomes visible, and can only change to `WT_TXN_ABORTED` when a transaction completes.

So, even if the code were to re-read `upd->txnid`, that is not the cause of the original problem (in fact, it would have made it less likely to fire).

The problem was that reconciliation's test for an "active transaction" for lookaside eviction was based on `txn_global->last_running`, which could change underneath reconciliation.  That could mean a transaction ID that was running concurrent with reconciliation could be cached by reconciliation and appear "committed" when the global `last_running` ID moved forward.

src/reconcile/rec_write.c
michaelcahill: Similarly, I disagree: since reconciliation has a transaction snapshot which is fixed while reconciliation is running, there is no chance of ambiguity when reading a transaction ID -- if it is visible as-of the snapshot, it is immutable (because committed transaction IDs are never rolled back).  If it is not visible as-of the snapshot, it doesn't matter whether reconciliation reads the original transaction ID or `WT_TXN_ABORTED` -- it was not be visible either way.

michaelcahill: Also, unless we were to leave a reserved update with a transaction ID after that transaction commits, or checkpoint were to start calling `WT_CURSOR::reserve`, it should never see a reserved update visible.

src/reconcile/rec_write.c
michaelcahill: To your specific point about `max_txn == UINT64_MAX` -- I think you're suggesting that local variables have no semantic meaning, so the compiler can completely ignore them and do multiple loads from memory.

I don't think is the case (and would break other code, e.g., I remember a bug in `__wt_random` that was resolved by reading the shared state into a local variable).

Even if we do need to worry about it, making `WT_UPDATE->txnid` into a `volatile` should be sufficient, right?  We don't need to order subsequent reads with respect to the read of `upd->txnid`, just make sure it is read once.

michaelcahill: I've pushed the `volatile` change in 0898a8907a19faf08bdd04033cd55d9196af490d.

keithbostic:
Noting discussion off-line, lgtm with the volatile change.

michaelcahill:

PR #3518, opened 2017-07-19
WT-3047 To uncover race conditions in page split in diagnostics mode

bvpvamsikrishna:


bvpvamsikrishna:
@sulabhM @daveh86 

dist/api_data.py
```diff
@@ -413,9 +413,10 @@ def __cmp__(self, other):
         blocking. This option is intended for use with internal stress
         testing of WiredTiger. Only available if WiredTiger is configured
         with --enable-diagnostic. Options are given as a list, such as
-        <code>"diagnostic_timing_stress=[checkpoint_slow]"</code>''',
+        <code>"diagnostic_timing_stress=[checkpoint_slow,
+                                        page_split_slow]"</code>''',
```
agorrod: I'd prefer if this were called `page_split_race` (and that filters through the change)

src/btree/bt_split.c
```diff
@@ -30,6 +30,8 @@ typedef enum {
 	WT_ERR_RETURN				/* Clean up and return error */
 } WT_SPLIT_ERROR_PHASE;
 
+static void __page_split_timing_stress(WT_SESSION_IMPL *);
```
sulabhM: The function definition can be moved before the first function call, and you won't need a separate declaration.

src/btree/bt_split.c
```diff
@@ -2184,3 +2187,25 @@ __wt_split_rewrite(WT_SESSION_IMPL *session, WT_REF *ref, WT_MULTI *multi)
 err:	__split_multi_inmem_fail(session, page, new);
 	return (ret);
 }
+
+/*
+ * __page_split_timing_stress --
+ *  Optionally add a 1000 micro seconds delay to simulate the race
```
sulabhM: I will prefer not to call out the duration of the delay in the comments. That way if we change the duration in the future, we do not have to worry about updating the comment.

dist/api_data.py
```diff
@@ -413,9 +413,10 @@ def __cmp__(self, other):
         blocking. This option is intended for use with internal stress
         testing of WiredTiger. Only available if WiredTiger is configured
         with --enable-diagnostic. Options are given as a list, such as
-        <code>"diagnostic_timing_stress=[checkpoint_slow]"</code>''',
+        <code>"diagnostic_timing_stress=[checkpoint_slow,
```
agorrod: I'd like you to take out the dependency on diagnostic mode and update the option name and general description. When we are chasing race conditions, they sometimes only occur in non-diagnostic builds, so it's not an ideal gating flag.

src/btree/bt_split.c
```diff
@@ -2184,3 +2187,25 @@ __wt_split_rewrite(WT_SESSION_IMPL *session, WT_REF *ref, WT_MULTI *multi)
...
+ *  conditions in page split for debug purposes. The purpose of
+ *  this is to uncover the race conditions in page split.
+ */
+static void
+__page_split_timing_stress(WT_SESSION_IMPL *session)
+{
+       WT_CONNECTION_IMPL *conn;
```
sulabhM: alignment in not correct here. There are 7 spaces followed by the text.
We use tabs to align different levels. and 4 spaces in addition to the tabs if the line wraps over.
Works best if tabs are set to 8 spaces in the text editor.

src/btree/bt_split.c
```diff
@@ -2184,3 +2187,25 @@ __wt_split_rewrite(WT_SESSION_IMPL *session, WT_REF *ref, WT_MULTI *multi)
...
+
+       conn = S2C(session);
+
+       /*
+	* We only want to sleep if the diagnostics flag is set
+	*/
+       if (FLD_ISSET(conn->timing_stress_flags,
+	   WT_TIMING_STRESS_PAGE_SPLIT_SLOW))
+	       __wt_sleep(0, 1000);
```
sulabhM: can use WT_THOUSAND instead of 1000

bvpvamsikrishna:
@sulabhM @daveh86 @agorrod 
Request to review this.

dist/api_data.py
```diff
@@ -411,11 +411,11 @@ def __cmp__(self, other):
         enable insertion of code that interrupts the usual timing of
         operations with a goal of uncovering race conditions and unexpected
         blocking. This option is intended for use with internal stress
-        testing of WiredTiger. Only available if WiredTiger is configured
-        with --enable-diagnostic. Options are given as a list, such as
-        <code>"diagnostic_timing_stress=[checkpoint_slow]"</code>''',
+        testing of WiredTiger. Options are given as a list, such as
+        <code>"diagnostic_timing_stress=[checkpoint_slow,
```
agorrod: Please also change the name of the option to not reference diagnostic. `timing_stress_for_test` seems like a good choice to me.

src/btree/bt_split.c
```diff
@@ -31,6 +31,28 @@ typedef enum {
...
+ */
+static void
+__page_split_timing_stress(WT_SESSION_IMPL *session)
+{
+	WT_CONNECTION_IMPL *conn;
+
+	conn = S2C(session);
+
+	/*
+	 * We only want to sleep if the diagnostics flag is set
```
agorrod: This comment is out of date - and we also use single line form for comments that fit onto a single line (unless it's for a function header).

dist/api_data.py
```diff
@@ -411,11 +411,11 @@ def __cmp__(self, other):
...
+                                        page_split_race]"</code>''',
```
agorrod: The re-indenting here isn't standard - it should match the `verbose` option.

agorrod:
Thanks @bvpvamsikrishna lgtm

PR #3519, opened 2017-07-19
WT-3438 Don't tune eviction thread count when the count is fixed

daveh86:


daveh86:
@agorrod can you have a quick look at this?

agorrod:
Thanks @daveh86 The change looks good to me - I moved it into a different ticket, since the change doesn't match the conversation in WT-2705

PR #3520, opened 2017-07-19
WT-3439 lint cleanup

keithbostic:
@agorrod, there's a bug in the new timestamp code (b06db73), so I did "full-build Friday" a few days early.


michaelcahill:

PR #3521, opened 2017-07-20
WT-3316 Add a developer documentation section starting with schema create

sulabhM:


sulabhM:
@agorrod , please have a look.

src/docs/devdoc-schema.dox
```diff
@@ -0,0 +1,116 @@
+/*! @page devdoc-schema Schema operations
+
+A schema defines the format of the application data in WiredTiger. WiredTiger
+supports various types of schemas (See @ref schema for more
+information), operated upon through a WT_SESSION reference. This section details
+the internals of these various schema operations.
+
+Schema operations cause an update to the metadata and are performed under a
+schema lock to avoid concurrent operations on the same object. Following
+sequence of steps define a generic schema operation:
```
agorrod: I think you should sit down with Amar and discuss the content of these two paragraphs - I feel like they could be clearer, but someone less involved with WiredTiger probably has a better perspective.

src/docs/devdoc-schema.dox
```diff
@@ -0,0 +1,116 @@
...
+partition with-schema-lock {
+    partition with-table-lock {
+        :turn on meta tracking;
+        :check uri}
+
+        split
+            :uri matches "colgroup:";
+            :__create_colgroup();
+        split again
+            :uri matches "file:";
```
agorrod: I'd restructure this a little bit. It's generally not expected that users operate directly on _file:_ URIs - those are pseudo internal, and used as building blocks for other data types. I'd create a picture with the four other URI types you identified, and have them lead into the file block. Some of them should have a many-to-one line to the file block (as in a table can be made up of many files).

agorrod: I'd either change this to _table:_ or add another one that is _table:_ - that's the most common URI, so it should be here.

agorrod:
autobuild: retest this please

agorrod:
@sulabhM Thanks - lgtm. Please merge once it has passed testing

sulabhM:
autobuild: retest this please

PR #3522, opened 2017-07-20
WT-3380 Remove packed attribute from WT_UPDATE.

michaelcahill:
We still check that the compiler doesn't add padding between fields, so the structure is the expected size, but deal with any padding added at the end.

This should allow compilers to generate more efficient code when reading fields from WT_UPDATE because they can assume fields are aligned appropriately.


michaelcahill:
autobuild: retest this please

michaelcahill:
@keithbostic, what do you think?

I don't love that we need to use `WT_UPDATE_SIZE` over `sizeof(WT_UPDATE)`, but I think that warning message is telling us something real so I'm reluctant to suppress it.

michaelcahill:
Closing in favor of #3523.

PR #3523, opened 2017-07-20
WT-3380 Remove packed attribute from WT_UPDATE.

keithbostic:
@michaelcahill, @sulabhM, here's a different change.

The fundamental change is adding `uint8_t data[3]` at the end of the `WT_UPDATE` structure: that means we can allocate `sizeof(WT_UPDATE)` without wasting bytes and it gives us a field name to use, we can replace `WT_UPDATE_DATA(upd);` with the more natural `upd->data;`

The only downside I see is we potentially waste 3 bytes on `WT_UPDATE` structures with no data package (so, `WT_UPDATE_DELETED`, `WT_UPDATE_RESERVED`, or zero-length values), but because we calculate the memory use of a `WT_UPDATE` structure as aligned to 32B, I don't think there's any actual loss.

Also, I tucked in a separate change in 14d3d0a, which locates the declaration of the `WT_UPDATE` timestamp. If you don't do that, using a non-8B timestamp (for example, `--with-timestamp-size=37`) will either fail the build because padding will get inserted in the middle of the structure, or fields like `upd->next` will be unaligned.

### DIFF
--- a/build_posix/aclocal/strict.m4
+++ b/build_posix/aclocal/strict.m4
@@ -82,7 +82,6 @@ AC_DEFUN([AM_GCC_WARNINGS], [
 AC_DEFUN([AM_CLANG_WARNINGS], [
 	w="-Weverything -Werror"
 
-	w="$w -Wno-address-of-packed-member"
 	w="$w -Wno-cast-align"
 	w="$w -Wno-documentation-unknown-command"
 	w="$w -Wno-format-nonliteral"
--- a/build_posix/configure.ac.in
+++ b/build_posix/configure.ac.in
@@ -75,7 +75,7 @@ AC_SUBST(AM_LIBTOOLFLAGS)
 # WiredTiger uses anonymous unions to pad structures. It's part of C11, but
 # some compilers require -std=c11 to support them. Turn on that flag for any
 # compiler that supports it, except for Solaris, where gcc -std=c11 makes
-# some none-C11 prototypes unavailable.
+# some non-C11 prototypes unavailable.
 if test "$wt_cv_solaris" = "no"; then
 	AX_CHECK_COMPILE_FLAG([-std=c11], [AM_CFLAGS="$AM_CFLAGS -std=c11"])
 fi
--- a/dist/s_style
+++ b/dist/s_style
@@ -56,6 +56,11 @@ else
 		cat $t
 	fi
 
+	if grep 'sizeof(WT_UPDATE)' $f > $t; then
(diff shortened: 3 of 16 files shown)


keithbostic:
@michaelcahill, @sulabhM, @agorrod, I think we have a problem.

If we removed the packed attribute as both Michael and my branches do, then build with a non-8B timestamp (`--with-timestamp-size=12`), the union declaration:
```
union __wt_timestamp_t {
        uint64_t val;
        uint8_t ts[WT_TIMESTAMP_SIZE];
};
```
causes padding to be inserted in the middle of the `WT_UPDATE` structure, even after locating the timestamp in the unaligned section of the `WT_UPDATE` structure as I did in 14d3d0a. Here's a list of the actual structure offsets next to the expected structure offset:
```
sizeof(WT_UPDATE): 48            <<< sizeof is wrong
define WT_UPDATE_SIZE: 36
offset txnid: 0 @ 0
offset next: 8 @ 8
offset size: 16 @ 16
offset type: 20 @ 20
offset timestamp: 24 @ 21        <<< timestamp got aligned
offset data: 40 @ 37
```
If we get rid of the union with something like this:
```
diff --git a/src/include/misc.h b/src/include/misc.h
index 5a47621b1..bf7d36e19 100644
--- a/src/include/misc.h
+++ b/src/include/misc.h
@@ -257,11 +257,14 @@
 #endif
 
 #ifdef HAVE_TIMESTAMPS
-union __wt_timestamp_t {
+struct __wt_timestamp_t {
+#if WT_TIMESTAMP_SIZE == 8
        uint64_t val;
+#else
        uint8_t ts[WT_TIMESTAMP_SIZE];
+#endif
 };
-typedef union __wt_timestamp_t wt_timestamp_t;
+typedef struct __wt_timestamp_t wt_timestamp_t;
 #define        WT_DECL_TIMESTAMP(x)    wt_timestamp_t x;
 #define        WT_TIMESTAMP_NULL(x)    (x)
 #else
```
Then the padding in the middle of the `WT_UPDATE` structure goes away, but the sizeof calculation is still wrong.
```
sizeof(WT_UPDATE): 40            <<< sizeof is still wrong
define WT_UPDATE_SIZE: 36
offset txnid: 0 @ 0
offset next: 8 @ 8
offset size: 16 @ 16
offset type: 20 @ 20
offset timestamp: 21 @ 21        <<< timestamp no longer aligned
offset data: 33 @ 33
```
If we don't want to pack the structure, I think we'll need some kind of solution like @michaelcahill's `WT_UPDATE_SIZE` replacing `sizeof(WT_UPDATE)`.

I'll push that branch and then quit, we should probably talk this over.

PR #3525, opened 2017-07-20
WT-3440 Checkpoint start should add a log record.

sueloverso:
@michaelcahill and @agorrod Here's a fix for the problem where the checkpoint LSN refers to the last real record and then recovery applies that record.  Here are points of note:
* The test has a lot of enhancements over the original in WT-3434.
* This pull request is based off develop as of 46975e5.  It will not pass on current develop due to the issues described in the ticket about different behavior.
* I use a log message/printf record.  Should any other information be added to the message?


sueloverso:
Ok, the pull request is based on that parent changeset, the pull request tester jobs don't retain that and apply it against the tip of develop.

sueloverso:
This is now ready for review.  With the merge of the checkpoint timestamp fix this version of the test now passes.  Once this is in I can move forward on the checkpoint work in WT-3387.

michaelcahill:
Thanks @sueloverso.  As discussed, I added a new record type for "checkpoint start" and switched from a `log_printf` (my main motivation is if we ever need to do something based on that record it shouldn't be something applications write to the log).  I'm happy with where this ended up, merging.

agorrod:
Does using a new log record type mean we need to handle upgrade/downgrade considerations?

michaelcahill:
@agorrod, this log record will be ignored by earlier versions of WiredTiger, so I don't think this introduces any upgrade/downgrade issues.  Does that sound right to you @sueloverso?

sueloverso:
I think the answer is yes and no.  @michaelcahill is correct that recovery ignores any record it isn't interested in, and printlog likewise prints out the LSN and header stuff and ignores content it doesn't recognize.  This includes running an older binary on a V1 version of the log file containing this record.  So neither fails with a new log record in a V1 log file.  However, it feels semi-wrong/unclean to write a new record into a V1 log file.  I have several thoughts:
* We could use the `system` record already added.  That just notes an LSN in the log file and has an `unused` WT_ITEM for future use without adding a new record.  It is fine for it to appear elsewhere in the log file.  It is simply an internal record with an LSN inside it.  It could be expanded or modified to include some other information.
* We could conditionally write this new record only if writing a V2 log file.
* We could combine and do both of the options above.
* It comes down to a stylistic decision about what level we want to add new information to the log.  I definitely agree that a printf is more hacky.  But using the existing system record is viable.

sueloverso:
@michaelcahill  I pushed two changes.  One is the fix for the log record definition.  The second is a change to only write the new record into V2 log files.  However, it occurs to me that the new log record is necessary for timestamps to work properly, which then ties timestamps to not running in compatibility mode.

agorrod:
Discussion about this recorded here:
https://jira.mongodb.org/browse/WT-3440?focusedCommentId=1630739&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1630739

agorrod:
Thanks @sueloverso and @michaelcahill lgtm - I'm going to go ahead and merge.

PR #3526, opened 2017-07-20
WT-3433 Add support for alter and readonly.  Add test case.

sueloverso:
@agorrod Please review this small change to correctly handle alter in a readonly connection.


agorrod:
Thanks @sueloverso lgtm

PR #3527, opened 2017-07-21
WT-2705 Add statistics to track latency of cursor operations and of disk operations

daveh86:


daveh86:
retest this please

daveh86:
retest this please

agorrod:
@daveh86 Is this ready for review? Could you switch the Jenkins perf testing onto this branch, to watch for any performance consequences?

daveh86:
@sulabhM can you have a look at this please?

src/include/os_fhandle.i
```diff
@@ -156,6 +160,7 @@ static inline int
 __wt_write(WT_SESSION_IMPL *session,
     WT_FH *fh, wt_off_t offset, size_t len, const void *buf)
 {
+	struct timespec start, stop;
```
sulabhM: I suppose struct timespec goes after WT_DECL_RET and other such macro types. Same holds good at other places.

src/cursor/cur_file.c
```diff
@@ -249,6 +260,8 @@ __curfile_insert(WT_CURSOR *cursor)
 	    (!F_ISSET(cursor, WT_CURSTD_APPEND) &&
 	    F_MASK(cursor, WT_CURSTD_KEY_SET) == 0));
 
+	__wt_stat_write_op_histogram(session, WT_TIMEDIFF_US(stop, start));
```
sulabhM: like at other places, lets update the statistics just after we save end time in the stop variable.

src/cursor/cur_file.c
```diff
@@ -290,13 +304,17 @@ __curfile_update(WT_CURSOR *cursor)
...
+	__wt_epoch(session, &start);
 	WT_ERR(__wt_btcur_update(cbt));
+	__wt_epoch(session, &stop);
 
 	/* Update maintains a position, key and value. */
 	WT_ASSERT(session,
 	    F_MASK(cursor, WT_CURSTD_KEY_SET) == WT_CURSTD_KEY_INT &&
 	    F_MASK(cursor, WT_CURSTD_VALUE_SET) == WT_CURSTD_VALUE_INT);
 
+	__wt_stat_write_op_histogram(session, WT_TIMEDIFF_US(stop, start));
```
sulabhM: can move this statement after getting stop time

src/include/os_fhandle.i
daveh86: I'm not certain thats true. I see plenty of code where we declare the timespec's before the macro types. Example src/txn/txn_ckpt.c, src/btree/bt_io.c-, src/btree/bt_read.c and others. In functions in these files timespec is the first variable declared, when it is declared. Perhaps we let this one go through to "the keeper"?

daveh86:

agorrod:
Did you check the automated perf tests for negative consequences?

daveh86:
The perf branch is running with these now. We should have details of their tomorrow.

daveh86:
Even with the change to not increment statistics without explicitly having a config entry, there was still a regression caused by this change. @agorrod i think we should have a chat about this.

agorrod:
@daveh86 I made a suggestion in the [JIRA ticket](https://jira.mongodb.org/browse/WT-2705?focusedCommentId=1631940&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1631940)

sulabhM:
autobuild: retest this please

src/support/time.c
```diff
@@ -67,3 +70,99 @@ __wt_seconds(WT_SESSION_IMPL *session, time_t *timep)
...
+			    1000000000 - (TIMER_PRECISION_MS * WT_THOUSAND)) {
+				conn->epoch_clock.tv_sec++;
+				conn->epoch_clock.tv_nsec = 0;
+			} else {
+				conn->epoch_clock.tv_nsec
+				    += TIMER_PRECISION_MS * WT_THOUSAND;
+			}
+		}
+
+		__wt_sleep(0, TIMER_PRECISION_MS);
```
sulabhM: Wouldn't we context switch and that itself in most cases be several micro secs. Lets assume we want 1 microsec precision, and we reset every 25 calls. If a single context switch takes say 2 microsecs (totally making this value up after googling context switch times), we can be off by as much as 50 microsecs, not too good if we are aiming for a 1 microsec precision. Need to think more about it. Initially I thought minimum precision is a millisec and context switch can likely be ignored. Not with microsecs I suppose.

daveh86:
@sulabhM that was 100% a concern. And after some thought and discussion, I've changed the whole thing over to a single `uint64_t` for timekeeping. Given this is low precision it should suffice.

daveh86:
@agorrod can you have a look at this?

src/include/connection.h
```diff
@@ -343,6 +343,18 @@ struct __wt_connection_impl {
 
 	WT_SESSION_IMPL *meta_ckpt_session;/* Metadata checkpoint session */
 
+	/* Macro to set a current time to be the same as the clock server */
+#define	WT_CLOCK_GET_TIME(session, time) time = S2C(session)->server_clock
+
+	/* Macro to set the clock servers time */
+#define	WT_CLOCK_SET_TIME(session, time)				\
+	__wt_atomic_store64(&S2C(session)->server_clock,		\
+	(uint64_t)(time.tv_sec * 1000000) + (uint64_t)(time.tv_nsec / 1000))
```
sulabhM: WT_MILLION instead of 1000000 and WT_THOUSAND instead of 1000

src/support/time.c
```diff
@@ -67,3 +70,99 @@ __wt_seconds(WT_SESSION_IMPL *session, time_t *timep)
...
+
+	/* Set the initial time value before starting the thread */
+	__wt_epoch(NULL, &time);
+	WT_CLOCK_SET_TIME(session, time);
+
+	F_SET(conn, WT_CONN_SERVER_CLOCK);
+
+	WT_RET(__wt_thread_create(
+	    session, &conn->clock_tid, __clock_server, session));
+	conn->clock_tid_set = 1;
```
sulabhM: true instead of 1, clock_tid_set being bool

src/support/time.c
```diff
@@ -67,3 +70,99 @@ __wt_seconds(WT_SESSION_IMPL *session, time_t *timep)
...
+	/*
+	 * Clear the clock server flag and then sleep for 2 clock sleeps to
+	 * allow the clock server to finish.
+	 */
+	F_CLR(conn, WT_CONN_SERVER_CLOCK);
+	__wt_sleep(0, 2 * TIMER_PRECISION_US);
+
+	if (conn->sweep_tid_set) {
+		WT_RET(__wt_thread_join(session, conn->clock_tid));
+		conn->clock_tid_set = 0;
```
sulabhM: false instead of 0

src/cursor/cur_file.c
```diff
@@ -185,9 +185,9 @@ __curfile_search(WT_CURSOR *cursor)
 	CURSOR_API_CALL(cursor, session, search, cbt->btree);
 	WT_ERR(__cursor_checkkey(cursor));
 
-	WT_CLOCK_GET_TIME(session, start);
+	start = __wt_rdtsc(session);
 	WT_ERR(__wt_btcur_search(cbt));
-	WT_CLOCK_GET_TIME(session, stop);
+	stop = __wt_rdtsc(session);
 	 __wt_stat_read_op_histogram(session, (stop - start));
```
sueloverso: Just FYI, until WT-3790 is merged these numbers are not meaningful. The TSC is only a count of clock ticks and that varies from machine to machine as to how many ticks per nanosecond.

sulabhM:
@sueloverso , this pull request is ready to be reviewed now.
I pointed perf to this branch and didn't see any regressions. I also executed couple of wtperf workloads, the statistics look neat and meaningful.

dist/stat_data.py
```diff
@@ -389,6 +394,32 @@ def __init__(self, name, desc, flags=''):
...
+    PerfHistStat('perf_hist_writefs_latency_lt100', 'file system write latency histogram - 50-99ms'),
+    PerfHistStat('perf_hist_writefs_latency_lt250', 'file system write latency histogram - 100-249ms'),
+    PerfHistStat('perf_hist_writefs_latency_lt500', 'file system write latency histogram - 250-499ms'),
+    PerfHistStat('perf_hist_writefs_latency_lt1000', 'file system write latency histogram - 500-999ms'),
+    PerfHistStat('perf_hist_writeop_latency_gt10000', 'operation write latency histogram - 10000us+'),
+    PerfHistStat('perf_hist_writeop_latency_lt250', 'operation write latency histogram - 100-249us'),
+    PerfHistStat('perf_hist_writeop_latency_lt500', 'operation write latency histogram - 250-499us'),
+    PerfHistStat('perf_hist_writeop_latency_lt1000', 'operation write latency histogram - 500-999us'),
+    PerfHistStat('perf_hist_writeop_latency_lt10000', 'operation write latency histogram - 1000-9999us'),
+
```
sueloverso: I have a preference for the variables to be called `perf_hist_fsread*` and `perf_hist_fswrite*` and `perf_hist_opread*` and `perf_hist_opwrite*` only so that the ones that go together are grouped together.

src/cursor/cur_file.c
```diff
@@ -179,12 +179,17 @@ __curfile_search(WT_CURSOR *cursor)
 	WT_CURSOR_BTREE *cbt;
 	WT_DECL_RET;
 	WT_SESSION_IMPL *session;
+	uint64_t start, stop;
```
sueloverso: There was a serious effort in WT-3790 to name all timings as `time_start` and `time_stop`. Please adjust these variables.  (This applies to the 4 or 5 other places this appears too)

src/include/btree.i
```diff
@@ -1707,3 +1707,63 @@ __wt_ref_state_yield_sleep(uint64_t *yield_count, uint64_t *sleep_count)
...
+	 * Ignore any operation that takes less than 100us to execute. This
+	 * floor value keeps us from having an excessively large smallest
+	 * execution time bucket. The 100us floor along with the other
+	 * histogram values were chosen here based on testing but performance
+	 * will vary considerably.
+	 */
+	if (usecs  < 100)
+		return;
+
+	if (usecs < 250) {
```
sueloverso: KNF: All of the curly braces here and in the other functions should be removed. These are single statement conditionals.

src/include/os_fhandle.i
```diff
@@ -6,6 +6,68 @@
...
+	/*
+	 * Ignore any operation that takes less than 10ms to execute. This
+	 * floor value keeps us from having an excessively large smallest
+	 * execution time bucket. The 10ms value and the other histogram buckets
+	 * were chosen based on expectations of disk response times and testing.
+	 */
+	if (msecs < 10)
+		return;
+
+	if (msecs < 50) {
```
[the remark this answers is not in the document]
sueloverso: I am also heavily in favor of this suggestion. I think you could create macros for them to incorporate the statistic field name via `##name` in a manner similar to the atomic functions in `gcc.h`. However where I disagree with @agorrod is that this PR has been open for 5+ months, it is worth spending a day more to get the function structure right and be done with it rather than postpone in another ticket.

src/include/os_fhandle.i
```diff
@@ -6,6 +6,68 @@
...
```
sueloverso: Again, KNF to remove all the curly braces.

sulabhM:
@agorrod please have another look at this review.

src/cursor/cur_file.c
```diff
@@ -8,6 +8,13 @@
 
 #include "wt_internal.h"
 
+/*
+ * Define functions that increment histogram statistics for cursor read and
+ * write operations latency.
+ */
+WT_STAT_USECS_HIST_INCR_FUNC(opread, perf_hist_opread_latency, 100)
+WT_STAT_USECS_HIST_INCR_FUNC(opwrite, perf_hist_opwrite_latency, 100)
```
agorrod: @sulabhM this looks OK to me, but it's a new pattern in WiredTiger source base, I think so I'm going to ask @sueloverso to review as well.

sueloverso: Hmm, I was hoping we'd get 4 functions down to one macro. I didn't realize the very slightly different ranges. Would we want to reconsider those ranges to be able to collapse them?  While I know the units are different, usecs versus msecs, I do think that looking at the stats in t2 it will be easier to only keep a consistent set of values in our head, or add an arg to the macro that lets them adjust some?

agorrod: I talked to Sulabh about this today, and I'd talked to Dave and the buckets when he was making the change originally. I think having two sets of buckets is the best solution. One for slower ops and one for faster ops. Attempting to share the same set across all histogram stats leads to having a lot more buckets per histogram which has a real space cost in our data structures.
We couldn't figure out a clean way to keep two sets of buckets with a single macro (we pretty much ended up passing in a set of buckets to each macro call).
Medium term I'd like to have this code auto generated by dist/stat.py,  but I think we are more likely to get that right if we do it for the next set of histogram stats we add..
If you had a concrete change in mind here it would be great if you can push a change.

sueloverso: Thanks for the info on the histogram ranges. Go ahead and merge.

sueloverso:
autobuild: retest this please

PR #3533, opened 2017-07-24
WT-3446 Temporarily disabled checkpoint timestamp testing.

bvpvamsikrishna:
Also modified checkpoint testing to return error if timestamp testing
is initiated


agorrod:
Thanks @bvpvamsikrishna lgtm - I'll merge.

PR #3535, opened 2017-07-25
WT-3336 Enable statistics by default in wtperf

daveh86:
Statistics gathering enabled on a 1s basis and output in JSON.


daveh86:
@bvpvamsikrishna can you review this?

bvpvamsikrishna:
@daveh86 changes are fine with me. No comments from my side.

agorrod:
@daveh86 Is there a reason why you chose not to pull the `statistics=(fast)` configuration option out of the individual runners as well?

daveh86:
@agorrod This was an oversight on my part due to WT-3456. I incorrectly did not believe we would need a mode set.

agorrod:
Can you take the explicit `statistics=` lines out of the wtperf runners as well?

daveh86:
Certainly. I had thought it best to leave those in to keep those tests explicit. Thinking further, this is wtperf, so we do not depend on the statistics configuration for anything within the run. Removing

agorrod:
Thanks @daveh86 lgtm

PR #3536, opened 2017-07-27
WT-3410 Add activity diagram in the documentation for schema rename

sulabhM:


sulabhM:
@agorrod , please have a look

agorrod:
Thanks @sulabhM lgtm.

PR #3537, opened 2017-07-27
WT-3461 Use CLOCK_MONOTONIC for pthread_cond_timedwait if possible.

michaelcahill:
Regardless, don't adjust the realtime clock before calculating when a timed sleep should end.  Otherwise, we can sleep for longer than expected by however much the clock changed.

### DIFF
--- a/build_posix/configure.ac.in
+++ b/build_posix/configure.ac.in
@@ -170,6 +170,44 @@ AS_CASE([$host_os], [darwin*], [], [AC_CHECK_FUNCS([fdatasync])])
 # the generic declaration in AC_CHECK_FUNCS is incompatible.
 AX_FUNC_POSIX_MEMALIGN
 
+# Check for POSIX condition variables with monotonic clock support
+AC_CACHE_CHECK([for condition waits with monotonic clock support],
+  [wt_cv_pthread_cond_monotonic],
+  [AC_RUN_IFELSE([AC_LANG_SOURCE([[
+#include <errno.h>
+#include <pthread.h>
+#include <stdlib.h>
+#include <time.h>
+
+int main()
+{
+  int ret;
+  pthread_condattr_t condattr;
+  pthread_cond_t cond;
+  pthread_mutex_t mtx;
+  struct timespec ts;
+
+  if ((ret = pthread_condattr_init(&condattr)) != 0) exit(1);
+  if ((ret = pthread_condattr_setclock(&condattr, CLOCK_MONOTONIC)) != 0) exit(1);
+  if ((ret = pthread_cond_init(&cond, &condattr)) != 0) exit(1);
+  if ((ret = pthread_mutex_init(&mtx, NULL)) != 0) exit(1);
+  if ((ret = clock_gettime(CLOCK_MONOTONIC, &ts)) != 0) exit(1);
+  ts.tv_sec += 1;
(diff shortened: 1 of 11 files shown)


michaelcahill:
@keithbostic and @sueloverso, can you please take a look at this change?  The basic issue is that if do our monotonic adjustment on clocks when calculating when a `pthread_cond_timedwait` should timeout, then we can sleep for longer than expected (up to the amount the clock changed).  If the clock were to change by 5 minutes, that might be indistinguishable from a hang.

If we decide to go in this direction, we probably want to pull the test case out of `configure.ac.in` and put it in its own file under `aclocal`.  Demonstrating the effect isn't straightforward: if you have a virtual machine you don't care about, run `test_rwlock` repeatedly and force the clock backwards periodically.  @bvpvamsikrishna is working on a better approach suggested by Max Hirschhorn using `libfaketime`.

keithbostic:
@michaelcahill, I pulled out the higher-level time functions into a new file (`support/time.c`), in commit 6581edb because `__wt_epoch()` had diverged and shouldn't. Feel free to discard that change without further discussion.

keithbostic:
@sueloverso, a comment you added in 2c6aec8 bothers me:
```
/*
 * Read into a local variable so that we're comparing the correct
 * value when we check for monotonic increasing time.  There are
 * many places we read into an unlocked global variable.
 */
```

If there are places where multiple threads are reading/setting an unlocked time value, I don't see how reading into a local variable here helps. We can race when doing the assignment from the local variable to the global, too which is presumably just as bad?

sueloverso:
@keithbostic The issue is that some location has time, say, T10.  Time moves backward, say to T9.  The code now wants time to stand still.  If we read T9 directly into that location, get descheduled for a while, and then set the location to T10 so it stands still, there is a window where time has moved backward in that location.  We don't want T9 to appear in that location temporarily.

sueloverso:
This LGTM.

keithbostic:
@sueloverso, it seems to me that if we race with a reader when doing the assignment into the global from the temporary, the exact same thing can happen. So, for example, if the code:
```
__wt_time_check_monotonic(session, &tmp);
tsp->tv_sec = tmp.tv_sec;
tsp->tv_nsec = tmp.tv_nsec;
return;
```
excutes as:
```
__wt_time_check_monotonic(session, &tmp);
tsp->tv_nsec = 0;
[thread descheduled]
tsp->tv_sec = (previous tsp->tv_sec + 1);
return;
```
Time can move backward in that location.

sueloverso:
True that `*tsp = tmp;` could generate instructions and execute as you say but without a temp variable I was easily able to see it in my testing.  With the tmp variable the window is very small.

keithbostic:
> True that *tsp = tmp; could generate instructions and execute as you say but without a temp variable I was easily able to see it in my testing. With the tmp variable the window is very small.

I'm a little uncomfortable with that.

If it matters (that is, if there are threads that asynchronously read the time value as it's being set, and which could be negatively affected by time moving backward), we should eliminate the problem. (I'm happy to do that work, and it should be a separate ticket.)

If it doesn't matter, then we should remove the comment in the source, at least.

michaelcahill:
Thanks for the reviews -- I'm going to merge this change and leave any changes related to protecting global timespecs for a separate ticket.

PR #3539, opened 2017-07-27
WT-3463 Add test phase to take backup without a checkpoint.

sueloverso:
@agorrod This case didn't seem covered in any of the other tickets.  I added a phase to the timestamp test that makes a backup without first doing a checkpoint, recovering from the previous checkpoint.  Please review.


agorrod:
Thanks @sueloverso lgtm

PR #3540, opened 2017-07-28
WT-3466 Track the first commit_timestamp set in a transaction.

michaelcahill:
In particular, if a commit_timestamp is set via `WT_SESSION::timestamp_transaction` and again later via `WT_SESSION::commit_transaction`, there was a window where `session->txn.commit_timestamp` was zero, which caused `WT_CONNECTION::query_timestamp` to return incorrect results.

Similarly, the shared list of transactions with commit timestamps should be sorted by the first commit timestamp: we don't (and shouldn't) re-sort that list if a transaction's commit_timestamp is updated.

### DIFF
--- a/src/include/txn.h
+++ b/src/include/txn.h
@@ -205,7 +205,21 @@ struct __wt_txn {
 	uint32_t snapshot_count;
 	uint32_t txn_logsync;	/* Log sync configuration */
 
+	/*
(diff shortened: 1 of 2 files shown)


agorrod:
Thanks @michaelcahill lgtm

PR #3542, opened 2017-07-30
WT-3470 table open/drop always open a metadata cursor

keithbostic:
@agorrod, in short, `__schema_open_table` holds the metadata cursor pinned across a call to `__wt_schema_open_colgroups`, which calls `__wt_metadata_search`, and so is forced to open an additional metadata cursor.

There's a similar case when dropping a table: `__wt_metadata_remove` opens a metadata cursor, and then calls `__wt_meta_track_update` which also opens a metadata cursor, forcing drop to open a second metadata cursor. I'm less comfortable with this change, but I think it's safe.


agorrod:
Thanks @keithbostic lgtm - I'm going to merge

PR #3543, opened 2017-07-30
WT-3471 the session table cache is never cleared leading to apparent memory leaks

keithbostic:
@agorrod, I'm suspicious this is the problem we're seeing in HELP-4534.

Note that this fix isn't sufficient in the case of a session never being reset.

### DIFF
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -836,6 +836,8 @@ __session_reset(WT_SESSION *wt_session)
 
 	WT_TRET(__wt_session_reset_cursors(session, true));
 
+	WT_TRET(__wt_schema_close_tables(session));
+
 	/* Release common session resources. */
 	WT_TRET(__wt_session_release_resources(session));
 


michaelcahill:
This change has the potential for a large performance regression, since tables will need to be reopened for every operation in MongoDB (since MongoDB calls `WT_SESSION::reset` in between every operation).  What performance testing has been done?

keithbostic:
> This change has the potential for a large performance regression.

I've done no performance testing.

An alternative might be this, because we'll ignore anything that's "current"?

```
index 7c9806788..fa7c7b2ed 100644
--- a/src/schema/schema_list.c
+++ b/src/schema/schema_list.c
@@ -237,6 +236,24 @@ __wt_schema_remove_table(WT_SESSION_IMPL *session, WT_TABLE *table)
 }
 
 /*
+ * __wt_schema_close_tables_gen --
+ *     Close all of the tables in a session that are no longer useful
+ */
+int
+__wt_schema_close_tables_gen(WT_SESSION_IMPL *session)
+{
+       WT_DECL_RET;
+       WT_TABLE *table;
+
+       TAILQ_FOREACH(table, &session->tables, q) {
+               if (table->schema_gen != __wt_gen(session, WT_GEN_SCHEMA))
+                       WT_TRET(__wt_schema_remove_table(session, table));
+       }
+
+       return (ret);
+}
+
+/*
  * __wt_schema_close_tables --
  *     Close all of the tables in a session.
  */
diff --git a/src/session/session_api.c b/src/session/session_api.c
index 6da7f48e7..bcf3a3f28 100644
@@ -836,7 +836,7 @@ __session_reset(WT_SESSION *wt_session)
 
        WT_TRET(__wt_session_reset_cursors(session, true));
 
-       WT_TRET(__wt_schema_close_tables(session));
+       WT_TRET(__wt_schema_close_tables_gen(session));
 
        /* Release common session resources. */
        WT_TRET(__wt_session_release_resources(session));

michaelcahill:
As discussed, I'm concerned that doing any pass through all cached handles during `WT_SESSION::reset` will cause a noticeable performance regression.  Looking at `table->schema_gen` helps a little, but still means we're invalidating every session's table cache after a drop.  I'm going to close this branch and find another approach.

PR #3544, opened 2017-07-30
WT-3456 Ensure stats are disabled when statistics is set to none

daveh86:


agorrod:
Thanks @daveh86 - nice catch and thanks for including a test case. lgtm.

PR #3545, opened 2017-07-31
WT-3476 Make tables first class data handles.

michaelcahill:
Keep a single handle for each table (rather than one per session) and reuse existing sweep logic so that idle tables are closed.


michaelcahill:
This branch currently has the change to make `WT_TABLE` an implementation of `WT_DATA_HANDLE`, and to use the existing session / connection cache code including the sweep server to make sure old table handles get closed.

It is lightly tested so far, and as I was making this change I noticed some problems -- e.g., I think we got away with some things previously because each session owned its own table handles.  In particular, when we're getting a column group or index, we have to keep its table pinned until we are finished.  There appears to be code that releases the table too early in some places.  I haven't tried to fix that yet.

It would also be nice to get ride of the global table lock and instead rely on locking the relevant table handle as required.  For example, if an operation gets a table handle in exclusive mode, it should be safe to make schema changes to that table.  One stumbling block there is that we re-enter `__wt_schema_get_table` when getting a column group or index and that conflicts with an operation holding the table exclusive.  There is machinery in `dhandle->excl_session` and friends for dealing with this, but it wasn't straightforward to get it working.

michaelcahill:
Next steps for this branch:

* clean up the places that make calls into btree vs table code.
* check how table handles are discarded: e.g., on `drop` and `rename`.  Do they go through a `DEAD` phase?  When will they be evicted from session caches?

src/conn/conn_dhandle.c
```diff
@@ -17,7 +17,11 @@ __conn_dhandle_destroy(WT_SESSION_IMPL *session, WT_DATA_HANDLE *dhandle)
 {
 	WT_DECL_RET;
 
-	WT_WITH_DHANDLE(session, dhandle, ret = __wt_btree_discard(session));
+	if (WT_PREFIX_MATCH(dhandle->name, "file:"))
```
agorrod: It would be nice if this wasn't a string comparison, and there were an abstraction. Maybe a construct like:

```
WT_RET(__wt_get_handle_type(session, &type));
if (type == WT_DHANDLE_TYPE_BTREE)
    ....
else if (type == WT_DHANDLE_TYPE_TABLE)
    ....
```

Maybe with a default type of none? I guess it could even be a type field that is accessed directly?

src/session/session_dhandle.c
```diff
@@ -269,7 +269,7 @@ __wt_session_release_btree(WT_SESSION_IMPL *session)
 			__session_discard_dhandle(session, dhandle_cache);
 	}
 
-	if (F_ISSET(btree, WT_BTREE_BULK)) {
+	if (btree != NULL && F_ISSET(btree, WT_BTREE_BULK)) {
```
agorrod: Layering wise I think it'd be nice if the btree specific code here were pulled into a separate function here.

src/session/session_dhandle.c
```diff
@@ -306,6 +306,16 @@ __wt_session_release_btree(WT_SESSION_IMPL *session)
 }
 
 /*
+ * __wt_session_release_btree --
+ *	Unlock a btree handle.
+ */
+int
+__wt_session_release_btree(WT_SESSION_IMPL *session)
```
agorrod: Would `__wt_session_release_btree` and `__wt_session_get_btree` ideally go away?

src/conn/conn_dhandle.c
michaelcahill: We could do that for built-in data handle types, but data sources should (eventually) be able to use the same handle caching.  I guess we could have a generic type to cover external data sources?

agorrod: That would be fine with me. It feels like an implementation showing itself in the interface, but I'd rather that was explicit than peeking under the covers like the string comparison is doing.

src/schema/schema_create.c
```diff
@@ -573,13 +573,14 @@ __create_table(WT_SESSION_IMPL *session,
...
-	if ((ret = __wt_schema_get_table(session,
-	    tablename, strlen(tablename), false, &table)) == 0) {
+	if ((ret = __wt_schema_get_table(
+	    session, tablename, strlen(tablename), false, 0, &table)) == 0) {
 		if (exclusive)
 			WT_ERR(EEXIST);
 		exists = true;
 	}
-	WT_ERR_NOTFOUND_OK(ret);
+	if (ret != ENOENT)
```
michaelcahill: I've tried to explain inline: if the metadata for a table doesn't exist, the `WT_NOTFOUND` from the metadata search causes `ENOENT` to be returned.  That's inherited from the btree code, and we do the same conversion at a higher level, but I'm not sure it's okay to change the low-level code without more thought so I'm just dealing with it here.

michaelcahill:
autobuild: retest this please

michaelcahill:
This branch failed `test/fops` -- attempting to open a cursor on a table found it created but without column groups.  The table lock is supposed to prevent this, and as far as I can tell, we are holding the table lock in the expected mode whenever setting or getting `table->cg_complete`.  I'll keep investigating, but this change should not be merged until this is fixed.

agorrod:
Thanks @michaelcahill this change looks good to me. I'd prefer to hold off merging until after the next master bump, if that's alright with you?

src/schema/schema_list.c
```diff
@@ -9,115 +9,56 @@
...
-		    session, name, namelen, ok_incomplete, &table);
+	saved_dhandle = session->dhandle;
 
-	if (ret == 0) {
-		++table->refcnt;
-		*tablep = table;
+	*tablep = NULL;
+	WT_RET(__wt_scr_alloc(session, namelen + 1, &namebuf));
+	WT_ERR(__wt_buf_fmt(
+	    session, namebuf, "table:%.*s", (int)namelen, name));
```
michaelcahill: In light of WT-3516, I think we should consider whether always allocating and copying a string here will add too much overhead to a common path.  Maybe we need another version of `__wt_schema_get_table` for the case where the caller has a `table:T` URI already...

agorrod:
autobuild: retest this please

agorrod:
@michaelcahill I believe the latest test failure is showing a real issue with this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-pull-request-linux/3099/console

agorrod:
Thanks @michaelcahill lgtm

PR #3549, opened 2017-08-01
WT-3480 Don't set the async flag durings async stats

daveh86:


daveh86:
retest this please

michaelcahill:
@sueloverso I looked at this with @daveh86: we couldn't see any reason why this code should be touching `conn->flags`.  Can you please take a look and see if we're missing something?

The case Dave was chasing involved a close where a connection flag had just been cleared with `F_CLR`, but was still set, leading to async threads not exiting and a hang during `WT_CONNECTION::close`.  We went looking for places that could touch `conn->flags` concurrent with a close, and this is one place we found.

PR #3550, opened 2017-08-01
WT-3378 Add workload to maintain dirty cache around 5% during checkpoint scrub

sulabhM:


sulabhM:
@agorrod , please have a look.

agorrod:
Thanks @sulabhM lgtm

PR #3551, opened 2017-08-02
WT-3471 Sweep the table cache after schema changes.

michaelcahill:
During `WT_SESSION::reset`, if there has been a schema change (such as a `WT_SESSION::drop` operation) since the last sweep, do a pass through the table cache and remove any obsolete table handles.


michaelcahill:
Note: this change is intended for the `mongodb-3.4` branch.  WT-3476 is intended as the forward-looking fix for this issue but is too disruptive to backport.

agorrod:
Thanks @michaelcahill - this change looks good to me.

michaelcahill:
@agorrod thanks for the review, I'll merge.  I've followed up in jira about testing.

PR #3552, opened 2017-08-02
WT-3479 resolving modify updates before writing them to the lookaside file causes snapshot isolation failures.

keithbostic:
Not for review.


agorrod:
Thanks @keithbostic. It took me a while to understand the change in `__wt_col_modify` - I didn't understand until I read your commit message. This change looks good to me.

@michaelcahill This deserves your review as well. If you are happy with it, I think we should go ahead and merge.

michaelcahill:
Thanks @keithbostic, lgtm.  Nice catch!  Merging.

PR #3553, opened 2017-08-02
WT-3472 WT_CURSOR.modify has incorrect API wrapping macros

keithbostic:
Instead of using CURSOR_UPDATE_API_CALL/CURSOR_UPDATE_API_END, use CURSOR_API_CALL/API_END_RET, we're not doing an update, we're only calling an update function.

Replace an API_END/return (ret) pair with API_END_RET().


agorrod:
Thanks @keithbostic lgtm

PR #3554, opened 2017-08-02
WT-3452 Add timestamp-based recovery test.

sueloverso:
@agorrod I added a new test in `test/recovery` for testing timestamps, aborts and recovery.  It is a new program instead of modifying the other one because it adds a new checkpoint thread.  There is a section that is currently ifdef'd out because it can fail until the code for WT-3458 is completed.  Once this is merged I'll add a reminder to that ticket.


sueloverso:
@agorrod I believe this is ready for review.  The code is no longer ifdef'ed now that WT-3458 is merged.  I have seen occasional failures, maybe about 1 out of 100 and I am debugging them.  But I think they are more likely a bug in the library code rather than the test code (i.e. something real).  I am investigating, and will open other tickets based on what I find.

agorrod:
@sueloverso Could this be added as a new program in `test/csuite` instead of as a new program in `test/recovery`?

sueloverso:
It could, but it was copied from and is heavily based off `test/recovery/random-abort`.  There are a few tests in there with similar structure.  I think converting all of them to `test/csuite` style could be a separate ticket and they should all be done together.  Given the similarities I'd rather they all stay together.

PR #3556, opened 2017-08-03
WT-3458 Checkpoint timestamps only apply to non-logged tables.

michaelcahill:
When logging is enabled for a table, that indicates that updates should become durable immediately, since they will be unconditionally rolled forward by recovery after a crash.  Since checkpoints with a timestamp are intended to create a version on disk that does not include newer changes, it doesn't make sense with immediately durable tables.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -1257,7 +1257,7 @@ methods = {
         timestamp. Supplied values must be monotonically increasing.
         See @ref transaction_timestamps'''),
     Config('stable_timestamp', '', r'''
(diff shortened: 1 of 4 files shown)


michaelcahill:
@sueloverso, can you please review this one?

It's a fairly simple code change in the end.  I've tried to make sure the documentation is sensible, but if you find other places that need updating, please let me know.

I talked to @agorrod about abstracting out a function instead of scattering checks of the  `WT_BTREE_NO_LOGGING` flag in multiple places.  If that lands in WT-3388 before this change, I'll update to use it.

PR #3557, opened 2017-08-03
WT-3486 Discard obsolete updates as part of re-instantiating updates based on the lookaside table.

keithbostic:
@agorrod, a minor change that occurred to me that might be helpful in some workloads.

Feel free to simply discard the branch without discussion if you don't think it's worth doing.


agorrod:
Thanks @keithbostic lgtm - I'm not sure how much it'll help, but it definitely makes sense not to keep pointless updates around.

PR #3558, opened 2017-08-04
WT-3471 Sweep the table cache after schema changes.

michaelcahill:
During WT_SESSION::reset, if there has been a schema change (such as a WT_SESSION::drop operation) since the last sweep, do a pass through the table cache and remove any obsolete table handles.

(cherry picked from commit 74510af)


michaelcahill:
@agorrod, as discussed, here is the version of this change for `develop`.  I expect it to become obsolete once WT-3476 is merged, but this way it will appear in all branches and ensure the backported change gets some stress testing.

agorrod:
Thanks @michaelcahill I'm happy for this to be merged as soon as tests pass.

PR #3559, opened 2017-08-04
WT-3481 Fix an LSM bug where expected chunks size could be overestimated

agorrod:
when choosing chunks for a merge.


michaelcahill:

PR #3562, opened 2017-08-04
WT-3489 Add an API to set the tracked commit_timestamp.

michaelcahill:
This is required during rollback, because the commit_timestamp determines how much of the oplog is visible.


michaelcahill:
Note: the initial commit is the code change only.  Tests that set `commit_timestamp` are required before this is ready for review.

michaelcahill:
autobuild: retest this please

agorrod:
Thanks @michaelcahill - I agree that this deserves a test, but the code change looks OK to me from inspection.

agorrod:
Thanks @michaelcahill lgtm

PR #3564, opened 2017-08-07
WT-3492 Cleanup ex_all.c.

michaelcahill:
In particular, execute as much of the code as possible (the backup calls use an absolute path for documentation purposes so can't be run).  Make all functions static so we have to deliberately avoid calling them.


michaelcahill:
(note that I screwed up the ticket number in the branch name and initial commit message: please double check before merging).

michaelcahill:
@bvpvamsikrishna, can you please review this change?

The `ex_all.c` example is special: it contains all of the code "snippets" that appear in the documentation.  The idea is to have them in real code that compiles and runs to make sure the code in our documentation works as expected.

agorrod:
Thanks @michaelcahill lgtm

examples/c/ex_all.c
```diff
@@ -765,20 +812,40 @@ session_ops(WT_SESSION *session)
 	error_check(session->verify(session, "table:mytable", NULL));
 	/*! [Verify a table] */
 
-	/*! [Drop a table] */
-	error_check(session->drop(session, "table:mytable", NULL));
-	/*! [Drop a table] */
-	}
+	/* Call other functions, where possible. */
+	(void)backup;
```
bvpvamsikrishna: I am not sure whether not calling backup is intentional ?

michaelcahill: I'll add a better comment: we can't call `backup` because it has absolute paths in the calls, so won't work unless certain directories exist.  We need to do something with the function now it is `static` or the build will fail, so that's what is happening here.

PR #3565, opened 2017-08-07
WT-3494 __cursor_needvalue() calls __cursor_localkey(), not __cursor_localvalue()

keithbostic:
Found as a bug in the Helium port.


agorrod:
Thanks @keithbostic lgtm.

PR #3566, opened 2017-08-07
WT-3449 Get the Helium software running again.

keithbostic:


keithbostic:
@michaelcahill, would you please review 089b0c7 and 76d3eb1?

089b0c7 is a change to deal with collators, and we base configuration for `collator` and `app_metadata` isn't available in the datasource cursor layer, and we may not find it in the metadata configuration string, and I'm not sure if that's correct or not.

76d3eb1 is a change to start a transaction before calling into the underlying datasource code, that was removed and modified a couple of different times, so I'm not sure what's the correct function to call.

Thanks!

ext/datasources/helium/helium.c
```diff
@@ -75,7 +84,7 @@ static int verbose = 0;					/* Verbose messages */
 		 * If we're set to a Helium error at the end of the day,\
 		 * switch to a generic WiredTiger error.		\
 		 */							\
-		if (ret < 0 && ret > -31,800)				\
+		if (ret < 0 && (ret > -31800 || ret < -31999))		\
```
michaelcahill: This is clearly a bug fix, but it took me a couple of reads to convince myself that it's correct.  How about this (which I think is equivalent):

```
/* If we get an error that is not a system error (indicated by a negative number), and outside the range reserved for WiredTiger errors, map to a generic WiredTiger error. */
if (ret < -31999 || (ret > -31800 && ret < 0))
        ...
```

ext/datasources/helium/helium.c
```diff
@@ -1644,6 +1627,30 @@ helium_cursor_update(WT_CURSOR *wtcursor)
...
+#ifdef XXX_BROKEN_KEITH
+	/*
+	 * XXX
+	 * We don't currently support reserve, this will require some work.
+	 * The test programs don't currently detect it, so return success.
+	 */
+	return (0);
+#else
+	(void)cursor;
+	return (0);
```
michaelcahill: I understand that it isn't tested, but can we make this `ENOTSUP` until it works?

ext/datasources/helium/helium.c
```diff
@@ -2440,6 +2430,13 @@ helium_session_truncate(WT_DATA_SOURCE *wtds,
 
 	ESET(unlock(wt_api, session, &ws->lock));
 	return (ret);
+#else
+	(void)wtds;
+	(void)session;
+	(void)uri;
+	(void)config;
+	return (0);
```
michaelcahill: Likewise, can we start with `ENOTSUP`?

ext/datasources/helium/helium.c
keithbostic: Fixed in 6ad4f35.

ext/datasources/helium/helium.c
keithbostic: No, because `format` doesn't have any way to avoid `reserve` operations. (If you want, I could add code to `format` to make `reserve` operations configurable.)

ext/datasources/helium/helium.c
keithbostic: This is the truncate code, I expect to fix it in one way or another before this branch is merged.

keithbostic:
@michaelcahill:

> w.r.t. the CURSOR_UPDATE_* macros, we should talk them through at some point -- we've certainly gotten lazier about when transaction IDs and snapshots are allocated, which might be what's causing problems here, but I'd like to understand what breaks without the explicit call to __wt_txn_autocommit_check for every function.

Here's the stack:
```
[1502381310:881036][898:0x7ffff7fdd740], t, WT_CURSOR.insert: ./src/include/txn.i, 603: F_ISSET(txn, WT_TXN_RUNNING)
[1502381310:881055][898:0x7ffff7fdd740], t, WT_CURSOR.insert: aborting WiredTiger library

#3  0x0000000000453309 in __wt_abort (session=0x7f7d80)
    at src/os_common/os_abort.c:22
#4  0x00000000004900ad in __wt_assert (session=0x7f7d80, error=0, 
    file_name=0x57a960 "./src/include/txn.i", line_number=603, 
    fmt=0x57a95d "%s") at src/support/err.c:501
#5  0x00000000004a8022 in __wt_txn_id_check (session=0x7f7d80)
    at ./src/include/txn.i:603
#6  0x00000000004a80dd in __wt_ext_transaction_id (wt_api=0x7dcd58, 
    wt_session=0x7f7d80) at src/txn/txn_ext.c:23
#7  0x00007ffff0b1adb9 in cache_value_append (wtcursor=0x1b663e0, remove_op=0)
    at helium.c:674
#8  0x00007ffff0b1c355 in helium_cursor_insert (wtcursor=0x1b663e0)
    at helium.c:1492
#9  0x00000000005022fd in __curds_insert (cursor=0x1318410)
    at src/cursor/cur_ds.c:337
```
In summary, we're doing an insert and the data source is trying to acquire the transaction ID (we associate transaction IDs with updates in the Helium cache so we can do visibility checks). We've not started a transaction and so the lookup eventually fails.

Obviously, it's almost certain we'll need a transaction, once we're in a datasource cursor function that attempts to modify the underlying store.

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill: I've removed the remaining `he_truncate()` calls, the Helium branch should be ready for review and merge.

michaelcahill:

PR #3569, opened 2017-08-08
WT-3450 Add verbose=[timestamp] functionality and related verbose messages

sulabhM:


src/txn/txn_ckpt.c
```diff
@@ -606,6 +609,9 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, const char *cfg[])
 			    sizeof(timestamp_config),
 			    "read_timestamp=%s", timestamp_buf));
 			txn_cfg[2] = timestamp_config;
+			__wt_verbose(session, WT_VERB_TIMESTAMP,
+			    "Checkpoint requested at stable timestamp: %s",
+			    timestamp_config);
```
sueloverso: Minor nit: Why not have this use the `wt_verbose_timestamp` function as well?  It may be less convenient because it is already embedded in a string but the consistency would be nice.

sulabhM: @sueloverso , I thought and spent sometime trying to make __wt_verbose_timestamp more generic but it proved to be more work than the time permitted. We can work on that as and when needed. In this function here, we do not have timestamps, but have strings instead and it would be quite a lot of work to fit into how __wt_verbose_timestamp is written. I am inclined to use __wt_verbose here and change the actual message to match that output by __wt_verbose_timestamp.

agorrod: That makes sense to me Sulabh. I feel like the best alternative is to make this function take varargs - and that seems more complex than is worthwhile yet. If we come across a broader range of verbose timestamp messages in the future we can make the change then.

agorrod:
Thanks @sulabhM lgtm. I'm going to merge since this fixes a recurring build failure.

PR #3570, opened 2017-08-08
WT-3483 Keep timestamps out of metadata.

sueloverso:
@milkie Let me know if this fixes your issue as well.  I've repro'ed with @michaelcahill 's test and reviewed the proposed fix.  If it works for you then I'll get this merged.


milkie:
I tried applying this patch to develop branch and running my unit test with that code.  Unfortunately, it did not fix my problem, and I still get WT_ROLLBACK returned from checkpoint() -- presumably still due to a conflict when checkpoint writes metadata.

michaelcahill:
@sueloverso, I'm inclined to merge this change (or something similar), even though it doesn't fix the issue @milkie is seeing.  The test cases in `test_timestamp05.py` demonstrate problems if updates with timestamps make it into the metadata.

Can you please review?

sueloverso:
LGTM

PR #3572, opened 2017-08-10
WT-3399 Change the checkpoint conflict test

daveh86:


daveh86:
@agorrod can you look at this? It changes the flag for WT-3399 and I can update the test suite.

agorrod:
Thanks @daveh86 lgtm

PR #3573, opened 2017-08-10
WT-3492 Fix for ex_all.c with timestamps disabled.

michaelcahill:


agorrod:
lgtm and passed configure combinations.

PR #3574, opened 2017-08-10
WT-3499 Have checkpoint wait for in-flight commits to finish.

michaelcahill:
After checkpoint calculates an LSN that recovery should roll forward from, it should wait for any commits that are in progress (i.e., transactions that have written to the log but not yet become visible) to complete.  This is done by having all transactions hold a read lock from before the log write until they become visible.  Checkpoint waits for a write lock which ensures all in-flight transactions have drained before checkpoint starts its transaction.

### DIFF
--- a/src/include/txn.h
+++ b/src/include/txn.h
@@ -129,6 +129,13 @@ struct __wt_txn_global {
 	WT_TXN_STATE	  checkpoint_state;	/* Checkpoint's txn state */
 	WT_TXN           *checkpoint_txn;	/* Checkpoint's txn structure */
 
+	/*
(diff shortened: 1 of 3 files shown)


michaelcahill:
@sueloverso, here is one of the approaches we talked about to addressing WT-3499.  I didn't have a chance to run the repro on this branch today to see whether it is sufficient to fix the bug.

sueloverso:
Ha ha!  We were doing this at exactly the same time.  I have a branch PR too!  The only difference is naming and acquiring/releasing the write lock in checkpoint.  I'm grabbing it around the code that gives us our LSN.

michaelcahill:
Closed in favor of #3575.

PR #3575, opened 2017-08-10
WT-3499 Add a visibility rwlock between transactions and checkpoints.

sueloverso:
@michaelcahill and @agorrod here's a branch that solves the visibility and checkpoint issue that my test uncovered.  Please review.  It solves the problem even with my aggressive sleep repro case.  I am going to point the perf jobs at this branch for the day too.

### DIFF
--- a/src/include/txn.h
+++ b/src/include/txn.h
@@ -105,7 +105,7 @@ struct __wt_txn_global {
 	/* Protects the active transaction states. */
 	WT_RWLOCK rwlock;
 
-	/* Protects logging, checkpoints and transaction visibilty. */
+	/* Protects logging, checkpoints and transaction visibility. */
 	WT_RWLOCK visibility_rwlock;
 
 	/* List of transactions sorted by commit timestamp. */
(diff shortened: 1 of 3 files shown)


sueloverso:
@michaelcahill This branch has your usage of the write lock and a bunch of comments explaining this.  I have tested with with the repro.  After one full Jenkins run of the perf tests the plots do not show any change.  We can probably close your branch down.

michaelcahill:
@sueloverso, thanks, I'm happy with this if you're ready to merge.  There was one failure of the stress job, here: http://build.wiredtiger.com:8080/job/wiredtiger-test-recovery-stress/36/console

That was my only reason for not merging, but given that @agorrod and I can see similar looking symptoms when running `test_timestamp04` with 10x the number of records, I'm wondering whether that failure was something else (such as incorrect data after lookaside eviction / restore).

sueloverso:
I took a look last night at the failure and it appears to be completely unrelated to these changes.  It happened on the file that is not involved in logging so it should not be affected by recovery.

PR #3576, opened 2017-08-10
WT-3502 Only keep 10 modify records between full updates

keithbostic:


keithbostic:
@agorrod, I tucked an additional, somewhat related change in here.

We weren't sanity checking `WT_CURSOR.modify` operations, that is, if an application passed in a 32B value with 5 single-byte updates to it, we'd end up creating a bigger update than if we just updated the full value. I don't think this will fire much and it's a trivial test, what I want to catch is applications using `WT_CURSOR.modify` everywhere, regardless of when it's appropriate.

The test is very simplistic (never do a modify update for a value less than 64B). I could be more clever (check if the complete update payload is N bytes greater than the modified update payload), but I don't see much point.

The 64B was picked at random: we might want to switch to something like 100B (because our log records are 128B and so writing a modified update saves us no logging space unless it's at least around 100B). Is saving 40B in the cache on an update worth any effort?

cc: @michaelcahill

michaelcahill:
@keithbostic, this looks okay to me, I'll merge.  We discussed making this limit configurable or auto-tuning, so my concern is that we'll find some other workload where 10 is too small but we can worry about that if it happens.

PR #3579, opened 2017-08-12
WT-3497 Improve logging message when hitting the WT session limits

keithbostic:


sulabhM:
@agorrod , please have a look

agorrod:
Thanks @keithbostic and @sulabhM lgtm

PR #3580, opened 2017-08-12
WT-3493 wt_verbose_dump_txn should display timestamp information

keithbostic:
@michaelcahill, there's one minor change here that might be of interest. I changed `__wt_timestamp_to_hex_string()` to return the string "0" in the case of a zero timestamp.

That's potentially interesting if `__wt_txn_global_query_timestamp()` calls that function with a zero timestamp as we'll return "0" instead of the empty string. I think that's a bug fix, if anything (unless we document somewhere that an empty string is the same as a "0" string), but wanted to call it to your attention.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -1050,7 +1050,9 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
 	const char *iso_tag;
 	uint64_t id;
 	uint32_t i, session_cnt;
-
+#ifdef HAVE_TIMESTAMPS
+	char hex_timestamp[3][2 * WT_TIMESTAMP_SIZE + 1];
+#endif
 	conn = S2C(session);
 	txn_global = &conn->txn_global;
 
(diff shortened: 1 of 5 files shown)


src/txn/txn.c
```diff
@@ -1061,10 +1063,35 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
...
 	    "last running ID: %" PRIu64, txn_global->last_running));
 	WT_RET(__wt_msg(session, "oldest ID: %" PRIu64, txn_global->oldest_id));
-	WT_RET(__wt_msg(session,
-	    "oldest named snapshot ID: %" PRIu64, txn_global->nsnap_oldest_id));
 
-	WT_RET(__wt_msg(session, "checkpoint running? %s",
+#ifdef HAVE_TIMESTAMPS
+	WT_RET(__wt_timestamp_to_hex_string(
+	    session, hex_timestamp[0], &txn_global->commit_timestamp));
+	WT_RET(__wt_msg(session, "commit timestamp: %s", hex_timestamp[0]));
```
sulabhM: These will be output when WT_VERB_EVICT_STUCK is set and won't care about the newly added WT_VERB_TIMESTAMP. If WT_VERB_TIMESTAMP is set and WT_VERB_EVICT_STUCK not, we won't see these. I think that is okay, thinking out loud here.

michaelcahill:
@keithbostic, I think the change to return a "0" string is okay -- queries should get `WT_NOTFOUND` rather than returning zero timestamps, so this should not be visible to applications.  Thanks, I'll merge...

src/txn/txn.c
```diff
@@ -1061,10 +1063,35 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
...
+	WT_RET(__wt_msg(session, "has_oldest_timestamp: %s",
+	    txn_global->has_oldest_timestamp ? "yes" : "no"));
+	WT_RET(__wt_msg(session, "has_pinned_timestamp: %s",
+	    txn_global->has_pinned_timestamp ? "yes" : "no"));
+	WT_RET(__wt_msg(session, "has_stable_timestamp: %s",
+	    txn_global->has_stable_timestamp ? "yes" : "no"));
+	WT_RET(__wt_msg(session, "oldest_is_pinned: %s",
+	    txn_global->oldest_is_pinned ? "yes" : "no"));
+	WT_RET(__wt_msg(session, "stable_is_pinned: %s",
+	    txn_global->stable_is_pinned ? "yes" : "no"));
```
sueloverso: A minor nit:  Did you see how it looked with related fields displayed near each other rather than types of fields grouped?  I.e. if I'm debugging something related to the oldest timestamp, it might be more useful to have `oldest_timestamp`, `has_oldest_timestamp` and `oldest_is_pinned` all printed together.

keithbostic: @sueloverso, please don't hesitate to re-order to make this more useful. (I haven't done any debugging in the timestamp code, so I used the field ordering from the include file as a first approximation of what should be displayed together, and of course alignment issues might will skew how things are grouped there.)

sueloverso: To wrap up, I spoke with @milkie who motivated the original ticket and he thinks that the current ordering is most useful to start.

PR #3582, opened 2017-08-12
WT-3358 LSM will hang if the manager fails to start

keithbostic:


agorrod:
Thanks @keithbostic lgtm, did you reproduce the original failure?

keithbostic:
@agorrod 
> did you reproduce the original failure?

Yes -- I've not reproduced @sulabhM's framework, but you can reproduce this one by returning an error out of a specific call to `__wt_thread_create` during the test.

PR #3583, opened 2017-08-14
WT-3417 Check for active transactions during upgrade/downgrade.

michaelcahill:
Fail upgrade / downgrade if the system isn't quiescent.  Document that applications must wait for an upgrade / downgrade to complete before starting any new operations.  Handle WiredTiger's checkpoint thread carefully: make sure no checkpoint is in progress when the compatibility version changes, and the checkpoint thread is already restarted by reconfigure.

Split out the reconfigure functionality into its own source file.

### DIFF
--- a/src/conn/conn_reconfig.c
+++ b/src/conn/conn_reconfig.c
@@ -1,0 +1,212 @@
+/*-
+ * Copyright (c) 2014-2017 MongoDB, Inc.
+ * Copyright (c) 2008-2014 WiredTiger, Inc.
+ *	All rights reserved.
+ *
+ * See the file LICENSE for redistribution information.
+ */
+
+#include "wt_internal.h"
+
+/*
+ * __wt_conn_compat_config --
+ *	Configure compatibility version.
+ */
+int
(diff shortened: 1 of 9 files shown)


src/conn/conn_reconfig.c
```diff
@@ -0,0 +1,212 @@
...
+		WT_RET_MSG(session, EINVAL, "unknown major version");
+	if (conn->compat_major == WIREDTIGER_VERSION_MAJOR &&
+	    conn->compat_minor > WIREDTIGER_VERSION_MINOR)
+		WT_RET_MSG(session,
+		    EINVAL, "illegal compatibility version");
+
+	/*
+	 * We're doing an upgrade or downgrade, check whether transactions are
+	 * active.  Wait for any in-progress checkpoints to complete.
+	 */
```
sueloverso: This comment is inconsistent with the code here.  There is no waiting for checkpoints in this code.

michaelcahill:
Thanks for the review @sueloverso -- I fixed the comment, it was left over from a version where the checkpoint lock was taken there instead.  I'll merge...

PR #3584, opened 2017-08-14
WT-3460 Implement col store rollback to stable timestamp

sulabhM:


keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

agorrod:
Thanks @sulabhM I pushed up one further nit - please merge once testing passes.

@keithbostic You mentioned:

> test_timestamp04.py doesn't conform to 80-character line rules.

I'm pretty sure we switched Python to using a 100 character line limit, I based the original test case of another timestamp test that also had lines over 80 characters long.

keithbostic:
@agorrod 
> I'm pretty sure we switched Python to using a 100 character line limit, I based the original test case of another timestamp test that also had lines over 80 characters long.
 
I must have forgotten that decision! Anyway, fbm, I don't care about the limit as long as we're consistent.

PR #3585, opened 2017-08-15
WT-3455 Make eviction timestamp aware.

sueloverso:
NOT READY FOR REVIEW.
@michaelcahill Here's the first pass at adding some timestamp awareness to eviction.  This isn't correct because the (tiny, fits in cache) iteration of the test even fails, with EBUSY and memory left in cache.  I'm sure the visibility checks aren't quite right, but they do eliminate the failure where the records are evicted.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1906,8 +1906,7 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		mod = page->modify;
 		if (modified && txn_global->current != txn_global->oldest_id &&
 		    (mod->last_eviction_id == __wt_txn_oldest_id(session) ||
-		    !__wt_txn_visible_all(session, mod->update_txn,
-		    WT_TIMESTAMP_NULL(&mod->update_ts))))
+		    !__wt_txn_visible_all(session, mod->update_txn, NULL)))
 			continue;
 
 fast:		/* If the page can't be evicted, give up. */
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1906,7 +1906,8 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		mod = page->modify;
 		if (modified && txn_global->current != txn_global->oldest_id &&
 		    (mod->last_eviction_id == __wt_txn_oldest_id(session) ||
-		    !__wt_txn_visible_all(session, mod->update_txn, NULL)))
+		    !__wt_txn_visible_all(session, mod->update_txn,
+		    WT_TIMESTAMP_NULL(&mod->update_ts))))
 			continue;
 
 fast:		/* If the page can't be evicted, give up. */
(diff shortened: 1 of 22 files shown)


src/include/txn.i
```diff
@@ -308,9 +308,28 @@ __wt_txn_visible_all(
...
+                        return (false);
+	        __wt_readlock(session, &txn_global->rwlock);
+	        stable_cmp = __wt_timestamp_cmp(timestamp,
+                    &txn_global->stable_timestamp);
+	        cmp = __wt_timestamp_cmp(timestamp,
+                    &txn_global->oldest_timestamp);
+	        __wt_readunlock(session, &txn_global->rwlock);
+                if (stable_cmp > 0 || cmp < 0)
+                        return (false);
+        }
```
michaelcahill: This seems overly complicated because the `pinned_timestamp` has to be less than either the stable or oldest timestamp and we're already checking for that below.

I *think* all we need here is:

```
if (timestamp == NULL)
        return (true);
else if (!txn_global->has_pinned_timestamp)
        return (false);
```

Does that make sense?

src/evict/evict_lru.c
```diff
@@ -1906,7 +1906,8 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 		mod = page->modify;
 		if (modified && txn_global->current != txn_global->oldest_id &&
 		    (mod->last_eviction_id == __wt_txn_oldest_id(session) ||
-		    !__wt_txn_visible_all(session, mod->update_txn, NULL)))
+		    !__wt_txn_visible_all(session, mod->update_txn,
+		    WT_TIMESTAMP_NULL(&mod->update_ts))))
```
michaelcahill: I thought this check was only supposed to be advisory -- i.e., it's only supposed to avoid wasted work.  Is it necessary for correctness?

There is a check that `mod->rec_max_timestamp` is `visible_all` in `__wt_page_can_evict` that I think of as the correctness check...

agorrod:
autobuild: retest this please

michaelcahill:
@sueloverso, I meant to talk these changes over with you: sorry!

I think this branch is sufficient -- i.e., no specific changes for `stable_timestamp` are required outside of fixing lookaside.  Can you please review and merge if it looks good?

sueloverso:
@michaelcahill The code change looks good.  I need to clean up the test changes and then will merge.

sueloverso:
@michaelcahill I created a new test, `test_timestamp07.py` that is a variant of timestamp03 with a focus on eviction, confirming eviction is in play or not as expected.  For the initial 500 keys scenario it is now working.  I have the 1000 keys scenario commented out because it is failing in one scenario and dumping core in another.

agorrod:
@sueloverso I took a look at `test_timestamp07` with 1000 keys, and it's using lookaside eviction, so I believe the failure is probably related to WT-3435. I'll put a comment into that ticket about testing with the change to `test_timestamp07`. I don't think that should block this change being merged.
@michaelcahill if you agree please go ahead and merge.

michaelcahill:
Thanks @sueloverso, lgtm.  Merging...

PR #3586, opened 2017-08-15
WT-3417 Fix compiler warning when timestamps are disabled.

agorrod:


agorrod:
@michaelcahill I'm going to self review on this one - it's a simple change to fix a compilation problem when timestamps are disabled.

PR #3587, opened 2017-08-15
WT-3509 __wt_illegal_value doesn't always provide a failure location

keithbostic:


agorrod:
Thanks @keithbostic lgtm. I'm not a huge fan of macros with function style names, but I understand why it's justifiable here.

keithbostic:
@agorrod 
> I'm not a huge fan of macros with function style names, but I understand why it's justifiable here.

Yeah, me either -- as I think you assumed, I wasn't willing to do the big naming change to make it "better".

PR #3588, opened 2017-08-15
WT-3508 Fix some minor bugs in timestamp-abort.

sueloverso:


agorrod:
Thanks @sueloverso looks good to us.

sueloverso:
Thanks!  I was going to self-review this actually but never got back to it.

PR #3589, opened 2017-08-16
WT-3465 Avoid locks when reading 8-byte timestamps.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm

PR #3591, opened 2017-08-16
WT-3498 snapshot isolation failure with overflow records

keithbostic:
Rework reconciliation to cache overflow records required by readers as part of selecting the WT_UPDATE to be written. That way we know the cached record is associated with the appropriate update.

Instead of caching a single copy of an on-page overflow value that's being removed, and referencing it from both the WT_OVFL_TRACK structure (for existing reader threads), and from all of the WT_UPDATE affected chains, cache a complete copy of the on-disk overflow value itself everywhere it might be needed. This requires 2x the memory when updating row-store overflow tables, and potentially more for column-store tables with RLEs greater than 1, but it has the virtue of working. Additionally, the memory used should be discarded relatively quickly if there's cache pressure: from both the update chains and from the WT_OVFL_TRACK structure whenever we attempt to evict the page.


src/reconcile/rec_write.c
```diff
@@ -1521,7 +1529,23 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+	 * image is rewritten), or any reconciliation of a backing overflow
+	 * record that will be physically removed once it's no longer needed.
+	 */
+	if (*updp != NULL &&
+	    (F_ISSET(r, WT_EVICT_LOOKASIDE) ||
+	    (vpack != NULL &&
+	    vpack->ovfl && vpack->raw != WT_CELL_VALUE_OVFL_RM)))
+		WT_RET(__rec_append_orig_value(session, page, *updp, vpack));
+
+	return (0);
```
keithbostic: @michaelcahill, this change loses the check against `min_timestamp`:
```
if (!__wt_txn_visible_all(
    session, min_txn, WT_TIMESTAMP_NULL(&min_timestamp)))
```
I thought about trying to keep it, but I don't think it's worth complicating the original loop going through the updates to handle `WT_UPDATE_MODIFIED` records.

keithbostic:
@michaelcahill, the change is reasonably straight-forward:
* change `bt_debug.c` to avoid unaligned accesses and to correctly print out `WT_UPDATE_MODIFIED` records,
* change `bt_ovfl.c` to cache the actual overflow records instead of a reference to a WT_UPDATE structure, and to not bother caching anything unless it's a checkpoint,
* change `rec_write.c` to cache removed overflow records in the `WT_UPDATE` chains.

src/reconcile/rec_write.c
michaelcahill: I think this is fine because we're now walking the list a second time instead, in `__rec_append_orig_value`.  I've pushed a followup change: there's no need to track `min_txn` or `min_timestamp` at all.

src/btree/bt_ovfl.c
```diff
@@ -91,134 +89,56 @@ __wt_ovfl_read(WT_SESSION_IMPL *session,
...
 
-	/* Add the value's information to the update list. */
-	WT_RET(__wt_realloc_def(session,
+	/* Copy the overflow item into place. */
+	WT_ERR(__wt_realloc_def(session,
 	    &track->remove_allocated, track->remove_next + 1, &track->remove));
 	track->remove[track->remove_next].cell = unpack->cell;
-	track->remove[track->remove_next].upd = upd;
+	WT_ERR(__wt_strndup(session,
+	    tmp->data, tmp->size, &track->remove[track->remove_next].data));
```
michaelcahill: I know there are a couple of places that make copies of blocks of non-string data with `__wt_strndup` already, and there is a comment to that effect in the implementation.  It still looks weird to me, and forces me to go and remind myself that it isn't doing a `strcpy`.  What if we called it `__wt_memdup`?

src/include/btmem.h
```diff
@@ -899,6 +900,7 @@ struct __wt_update {
 #define	WT_UPDATE_MODIFIED	1	/* partial-update modify value */
 #define	WT_UPDATE_RESERVED	2	/* reserved */
 #define	WT_UPDATE_STANDARD	3	/* complete value */
+#define	WT_UPDATE_IMPOSSIBLE	255
```
michaelcahill: I can see how you ended up with `WT_UPDATE_IMPOSSIBLE`, but I think it's telling us that we need to refactor `__wt_col_modify` and `__wt_row_modify`.  As far as I can see, all we're really doing with this unused type is checking that when we call those functions to apply an existing update list, they don't allocate any updates.

Could you please have a look at restructuring the code so this isn't necessary?  Can the common path allocate updates first, then call a function to apply them (called directly in the places that are passing `WT_UPDATE_IMPOSSIBLE`)?

If there is no way around it, a better name might be `WT_UPDATE_INVALID`, but it still feels wrong to define a value here that can never appear in the struct.

src/btree/bt_ovfl.c
keithbostic: > What if we called it `__wt_memdup`?

Yes, that's better. I'll do that as a separate change outside this ticket, there are 70+ instances of `__wt_strndup`, it'll make this hard to review.

src/include/btmem.h
keithbostic: > Could you please have a look at restructuring the code so this isn't necessary?

I tried to refactor `__wt_col_modify` and `__wt_row_modify` a couple of weeks ago, and didn't get anywhere better.

> If there is no way around it, a better name might be `WT_UPDATE_INVALID`, but it still feels wrong to define a value here that can never appear in the struct.

Agreed.

In 09c2012, I changed `WT_UPDATE.type` to an `int8_t` so I can pass (-1) as the out-of-band value. That's a more traditional approach, and avoids `WT_UPDATE_IMPOSSIBLE`.

keithbostic:
@michaelcahill, this change ran overnight -- I'll  leave it running today, but I think this one is ready to merge if you approve 09c2012.

src/btree/row_modify.c
```diff
@@ -255,14 +255,21 @@ __wt_row_insert_alloc(WT_SESSION_IMPL *session,
...
 	WT_UPDATE *upd;
 
 	*updp = NULL;
 
 	/*
+	 * The code paths leading here are convoluted: assert we never attempt
+	 * to allocate an update structure if only intending to insert one we
+	 * already have.
+	 */
+	WT_ASSERT(session, modify_type != -1);
```
sueloverso: I'm going to drop in here to say that I would definitely prefer some defined constant to literal `-1`s sprinkled throughout.  Also if we changed it to an `int` so it can be negative, does it make more sense to just assert `modify_type > 0`?

keithbostic: Isn't it relatively common to use negative numbers as out-of-band values for integers?

Anyway, I'll let you & Michael fight it out, I'm fine with any of the above.

sueloverso: I don't feel super strongly, but seeing a whole bunch of `-1` sprinkled around felt weird to me.  Looking through our code for that string, the vast, vast majority of them are in the capacity of a `cmp` of some sort which makes sense.  Most non-`cmp` uses seem to be in the `os_*` directories.

src/btree/bt_ovfl.c
michaelcahill: To be clear, when we're duplicating strings, we should use `__wt_strndup`.  Only in the handful of cases that make copies of non-string data should we change to `__wt_memdup`.

michaelcahill:
autobuild: retest this please

michaelcahill:
I've reinstated a named mode, `WT_UPDATE_INVALID`, matching other places where we reserve a type.  I'll merge.

PR #3592, opened 2017-08-17
WT-3365 Fix LSM tree visibility rules when timestamps are used.

agorrod:


agorrod:
@michaelcahill could you review this for me please?

michaelcahill:
Thanks @agorrod, this turned out nice and clean in the end.

I have a feeling there may be some oddness on startup -- e.g., the chunk could appear visible, then when the first timestamp is set, we would query them and it could temporarily become invisible again.  Maybe we should set `WT_LSM_CHUNK_HAS_TIMESTAMP` if we get through here and the global timestamps aren't set, so we never go back?

agorrod:
Thanks for the review @michaelcahill I've made the change you suggested, as well as returning true immediately for stable and on disk chunks - since they are required to only contain visible updates.

agorrod:
autobuild: retest this please

michaelcahill:

PR #3594, opened 2017-08-17
WT-3328 wiredtiger-test-format-stress-ppc failed due to EBUSY during verify

keithbostic:


src/docs/error-handling.dox
```diff
@@ -36,8 +36,8 @@ an object is not available for exclusive access. For example, the
 WT_SESSION::drop or WT_SESSION::verify methods will fail if the object
 has open cursors. Note that internal WiredTiger threads may temporarily
 open cursors on objects (for example, threads performing operations like
-handle sweep or statistics logging), and operations may temporarily fail
-and return \c EBUSY when there are no application cursors open on the object.
+statistics logging), and operations may temporarily fail and return \c EBUSY
+when there are no application cursors open on the object.
```
keithbostic: @agorrod, sweep acquires `dhandle->rwlock` (in `__sweep_expire_one`), why can't that collide with acquiring the same lock in `__wt_session_lock_dhandle`? Just curious what I missed, thanks.

agorrod: The comment in the sweep code says:

> We don't set WT_DHANDLE_EXCLUSIVE deliberately, we want opens to block on us and then retry rather than returning an EBUSY error to the application.

I was thinking that is why sweep doesn't block verify.

michaelcahill: I've been back over WT-3362 and I think @keithbostic is correct that internal threads holding `dhandle->rwlock` in any mode can now cause `WT_SESSION::drop` and `WT_SESSION::verify` to return `EBUSY`.  I've opened WT-3520 to discuss making it configurable whether they wait or not.

PR #3595, opened 2017-08-18
WT-3517 WT_SESSION::reset doesn't need to call out EBUSY specially.

michaelcahill:


michaelcahill:
Self-review: trivial doc change, merging.

PR #3597, opened 2017-08-18
WT-3514 Fix checkpoint's use of stable_timestamp.

michaelcahill:
Prior to this change, there was a window in between checkpoint reading the stable timestamp and using it to start a transaction.  In that window, the application could update timestamps, making it invalid for checkpoint's transaction to start.

Remove the undocumented "read_timestamp" config to WT_SESSION::checkpoint, and instead copy the stable timestamp directly into checkpoint's transaction while holding the global transaction state locked.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -1149,9 +1149,6 @@ methods = {
     Config('name', '', r'''
         if set, specify a name for the checkpoint (note that checkpoints
         including LSM trees may not be named)'''),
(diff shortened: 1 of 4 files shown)


michaelcahill:
@sueloverso, can you please review this one?

sueloverso:
LGTM.  I pushed a fix for the `test_timestamp03` to remove the test case for `read_timestamp` that should fix the PR failure.  I'll merge when it's ready.

PR #3599, opened 2017-08-18
WT-3516 Optimize schema operations when we are certain of URI prefixes.

ddanderson:
These changes speed up a focused benchmark by 13%.  The benchmark is a single threaded application that does cursor_open in a loop.


michaelcahill:

I've made a comment on in #3545 about a follow up change so we don't lose this improvement when moving to table handles as data handles.

PR #3601, opened 2017-08-21
WT-3525 Remove eviction stats check.

sueloverso:
@keithbostic Please review this small test change.  I hope it is sufficient.  Instead of checking the stats after segments of the test I only left in the one at the end of the test.  The test is supposed to be testing eviction so the original intent was to make sure eviction happened (or not based on a boolean).


sueloverso:
@keithbostic Please give this one more look.  I totally removed the statistics checks since they're unreliable.

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

PR #3602, opened 2017-08-21
WT-3527 Use power of two hash table size.

ddanderson:
A good hash function should give us good distribution no matter what modulo we use, and we use a good hash function.

### DIFF
--- a/src/include/connection.h
+++ b/src/include/connection.h
@@ -7,10 +7,10 @@
  */
 
 /*
- * Default hash table size; use a prime number of buckets rather than assuming
- * a good hash (Reference Sedgewick, Algorithms in C, "Hash Functions").
+ * Default hash table size; we don't need a prime number of buckets
+ * because we always use a good hash function.
  */
-#define	WT_HASH_ARRAY_SIZE	509
+#define	WT_HASH_ARRAY_SIZE	512
 
 /*******************************************
  * Global per-process structure.


michaelcahill:
autobuild: retest this please

keithbostic:
autobuild: retest this please

ddanderson:
@keithbostic, can you review, please?  From my reading about modern hash functions and the smhasher test suite, it seems like we'll get a pretty good distribution so the modulo shouldn't matter.  I see about a 1% improvement in my open_cursor test with this change.  If you like, I could put together a distribution test based on the names that mongodb generates.

keithbostic:
It honestly surprises me this matters.

If you compile just the hash function at high optimization, do you see a significant difference in how fast the hash can be calculated?

Did you drill down and try to figure out the operation where the hash matters?

Doing a fast review of where `WT_HASH_ARRAY_SIZE is used, I see only a few uses:

1. the block manager on file open/close
2. the encryptor list
3. connection-level dhandle allocate, find, close, remove
4. underlying file open/close
5. in-memory file open/close
6. schema add-, find- and remove-table
7. session-level dhandle allocate, find, discard

and of all of those, I can't imagine anything other than #3 and #7 being interesting, and I'm skeptical we spend 1% of our time in cursor open doing a hash operation in any of those functions.

If this path matters, did you consider switching to the bitwise AND operation instead of relying on the compiler? That is, something like:
```
#define WT_BUCKET_HASH(n)        ((n) & (WT_HASH_ARRAY_SIZE - 1)
```
should guarantee a single instruction to calculate the bucket.

If this path matters and the issue is finding a bucket from a data handle, we could cache the bucket in the `dhandle`? (That's what we do for statistics buckets, in sessions.)

keithbostic:
Looked at the output of `clang 4.0 -O3`, and using a power-of-two turns 8 relatively slow instructions into a single `and`.

And, it's about 100x faster to calculate a power-of-two modulo.

michaelcahill:
Thanks @ddanderson and @keithbostic, sounds like this is a worthwhile change and lgtm.  I'll merge.

PR #3609, opened 2017-08-23
WT-3538 Misaligned reads when using the lookaside table and timestamps

keithbostic:


michaelcahill:
Thanks @keithbostic, I made a couple of cosmetic changes, lgtm.  I'll merge.

PR #3610, opened 2017-08-23
WT-3529 Add WT_CONNECTION::debug_info method.

sueloverso:
Not yet complete so it's not ready for review.


sueloverso:
@agorrod this is ready for review.  Are there any other subsystems to add?  While I wrote a python unit test for my own testing, all it really does is write stuff to stdout so I'm not sure it is worthwhile.

agorrod:
Thanks @sueloverso this change looks good to me. Please merge if you are ready - I'm only not merging in case you want another iteration on desired output with DanG before pulling the trigger.

PR #3612, opened 2017-08-24
WT-3539 Document API requirements for exclusive access.

michaelcahill:
Some WiredTiger methods require exclusive access to a table or file (e.g., alter, drop, verify, salvage, rebalance).  Add a shared paragraph to each of these methods explaining that they will fail if any cursors are open or the data source is otherwise busy.

While in the area, move "alter" into the "Table operations" section of WT_SESSION, clean up wording around methods that require a transaction to be active/inactive.


michaelcahill:

There is some shuffling in here because the `WT_SESSION::alter` documentation was in the wrong part of the documentation.  Otherwise I think it should be uncontroversial.

sueloverso:
LGTM

PR #3614, opened 2017-08-24
WT-3542 Python test_stats_log_on_json_with_tables timeout on the PPC

keithbostic:


sueloverso:
I'm not sure how this solves the problem.  I think this is another situation similar to WT-3541 where the PPC is very slow.  Maybe change the 30 to 100 like `reconfig02`?  Make the 30 a variable instead of hard-coded in code and comments while you're in there please!

keithbostic:
> Maybe change the 30 to 100 like `reconfig02`?  Make the 30 a variable instead of hard-coded in code and comments while you're in there please!

The problem here was that (although the comments described a requirement to sleep), there was no sleep within the loop.

keithbostic:
autobuild: retest this please

sueloverso:
Ah, I see the addition of the second sleep now, not just a line movement.  LGTM

PR #3618, opened 2017-08-24
WT-3542 Python test_stats_log_on_json_with_tables timeout on the PPC

keithbostic:
Still seeing timeout failures on the PPC: rewrite to not use timers at all.

### DIFF
--- a/test/suite/test_stat_log02.py
+++ b/test/suite/test_stat_log02.py
@@ -26,63 +26,19 @@
 # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 # OTHER DEALINGS IN THE SOFTWARE.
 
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic I haven't seen any failures since the last merge for WT_3542 (http://build.wiredtiger.com:8080/job/wiredtiger-test-unit-ppc/840/) - could you point me at it?

I'd also like to note that the PPC machine we use for testing is going to be offline Friday.

keithbostic:
> I haven't seen any failures since the last merge for WT_3542 

I've been running them outside of Jenkins, the last failure was after 4K runs. The current PR is up to 10K runs, so let's merge it if you're happy with it.

agorrod:
@keithbostic lgtm

PR #3621, opened 2017-08-27
WT-3537 Split pages when nothing can be written

keithbostic:
I tried a few different approaches, I liked this one the best.

In summary, the change is to copy the insert/append list key into the created disk image with a zero-length value, when reconciling row-store insert/append lists in the evict/restore path. That means we'll split the page (once the total memory required by the keys crosses the split boundary), and the restore of the page will split large insert/append lists into multiple update lists.

We're not splitting as aggressively as we could because we're not taking into account the update list's memory requirements (if values are larger than keys, that's where all the memory is). In other words, if the total memory required by the keys alone isn't significant, we're not going to split the page. That works if the in-memory max page size is significantly larger than the split size, not so much if they're about the same.

I tried a couple of times to take update list memory into account in choosing to split, but it gets ugly fast when you lie to reconciliation about how much memory you're using for the disk image. I suspect we could make progress here by configuring the split boundary differently for evict/restore when there's a large page memory footprint, or even placing a maximum count of items we'll put on a page during an evict/restore reconciliation.


keithbostic:
autobuild: retest this please

keithbostic:
One other note: this change could theoretically replace our in-memory split code, it accomplishes roughly the same thing. That would mean we'd no longer need a a separate path for in-memory splits, and we could remove `__wt_leaf_page_can_split()` and `__split_insert()`.

That said, it would likely be a little slower: `__split_insert()` is a pretty fast path (just carving off the last couple insert items in the list onto a new page), and my inclination is to leave it in place as an optimization for append-only workloads.

src/reconcile/rec_write.c
```diff
@@ -5824,32 +5826,50 @@ __rec_row_leaf_insert(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins)
...
-			break;
-		case WT_UPDATE_STANDARD:
-			if (upd->size == 0)
-				val->len = 0;
-			else
+			/*
+			 * When doing evict/restore, move the insert key to the
+			 * page, with an empty value (this allows us to split
+			 * the page if there's a huge, pinned insert list). The
+			 * on-page key must never be read, make sure there is a
```
agorrod: I believe this is the problem. What rollback_to_stable is doing, is moving the visibility point backward in time, so it will mark the UPDATE as WT_TXN_ABORTED. Though.. I'm surprised I see the particular symptoms I'm seeing, which are that the page has no WT_UPDATE structures, not WT_UPDATE structures with WT_TXN_ABORT. So there is something else happening here as well, I'm suspicious that there are two update-restores happening, the second one is happening on a page that only contained WT_TXN_ABORTED updates, and no valid on-disk entries - I didn't get far enough in to confirm that, but I used the following code changes that helped my understanding:

```
index eb8a258..080c4f3 100644
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -298,6 +298,7 @@ __cursor_row_next(WT_CURSOR_BTREE *cbt, bool newpage)
                goto new_insert;
        }

+       WT_RET(__wt_debug_tree_all(session, (WT_BTREE*)session->dhandle->handle, NULL, "tmp2.txt"));
        /* Move to the next entry and return the item. */
        for (;;) {
                /*
diff --git a/src/btree/bt_debug.c b/src/btree/bt_debug.c
index d91ac02..72d651e 100644
--- a/src/btree/bt_debug.c
+++ b/src/btree/bt_debug.c
@@ -207,7 +207,7 @@ __debug_config(WT_SESSION_IMPL *session, WT_DBG *ds, const char *ofile)
                WT_RET(__wt_scr_alloc(session, 512, &ds->msg));
                ds->f = __dmsg_event;
        } else {
-               if ((ds->fp = fopen(ofile, "w")) == NULL)
+               if ((ds->fp = fopen(ofile, "a")) == NULL)
                        return (EIO);
                __wt_stream_set_line_buffer(ds->fp);
                ds->f = __dmsg_file;
diff --git a/src/evict/evict_page.c b/src/evict/evict_page.c
index 03d6e9a..402e2fe 100644
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -583,6 +583,11 @@ __evict_review(
        *flagsp = flags;
        WT_RET(ret);

+       if (LF_ISSET(WT_EVICT_UPDATE_RESTORE)) {
+               fprintf(stderr, "Did an update restore eviction on %p\n", page);
+               WT_RET(__wt_debug_tree_all(session, (WT_BTREE*)session->dhandle->handle, NULL, "tmp.txt"));
+       }
+
        /*
         * Success: assert the page is clean or reconciliation was configured
         * for update/restore. If the page is clean, assert that reconciliation
diff --git a/src/txn/txn_rollback_to_stable.c b/src/txn/txn_rollback_to_stable.c
index e19bbc7..5b8f71f 100644
--- a/src/txn/txn_rollback_to_stable.c
+++ b/src/txn/txn_rollback_to_stable.c
@@ -415,6 +415,7 @@ __wt_txn_rollback_to_stable(WT_SESSION_IMPL *session, const char *cfg[])
        conn = S2C(session);
        WT_RET(__txn_rollback_to_stable_check(session));

+       fprintf(stderr, "Rollback to stable starting\n");
        /* Allocate a non-durable btree bitstring */
        WT_RET(__bit_alloc(session,
            conn->next_file_id, &conn->stable_rollback_bitstring));
@@ -428,6 +429,7 @@ __wt_txn_rollback_to_stable(WT_SESSION_IMPL *session, const char *cfg[])
         * lookaside records should be removed.
         */
        WT_ERR(__txn_rollback_to_stable_lookaside_fixup(session));
+       fprintf(stderr, "Rollback to stable finished\n");
 err:   __wt_free(session, conn->stable_rollback_bitstring);
        return (ret);
 #endif
diff --git a/test/suite/test_timestamp04.py b/test/suite/test_timestamp04.py
index 3af0fee..32f3deb 100644
--- a/test/suite/test_timestamp04.py
+++ b/test/suite/test_timestamp04.py
@@ -64,6 +64,9 @@ class test_timestamp04(wttest.WiredTigerTestCase, suite_subprocess):
             session.begin_transaction(txn_config)
         c = session.open_cursor(self.uri, None)
         if missing == False:
+            while c.next() == 0:
+                print 'K: ' + str(c.get_key()) + ' V: ' + str(c.get_value())
+            c.reset()
             actual = dict((k, v) for k, v in c if v != 0)
             #print expected
             #print actual
@@ -95,7 +98,7 @@ class test_timestamp04(wttest.WiredTigerTestCase, suite_subprocess):
         c = self.session.open_cursor(self.uri)

         # Insert keys each with timestamp=key, in some order
-        key_range = 10000
+        key_range = 1000
         keys = range(1, key_range + 1)

         for k in keys:
@@ -112,10 +115,10 @@ class test_timestamp04(wttest.WiredTigerTestCase, suite_subprocess):

         # Now check that we see the expected state when reading at each
         # timestamp
-        self.check(self.session, 'read_timestamp=' + timestamp_str(key_range / 2),
-            dict((k, 1) for k in keys[:(key_range / 2)]))
-        self.check(self.session, 'read_timestamp=' + timestamp_str(key_range / 2),
-            dict((k, 1) for k in keys[(key_range / 2 + 1):]), missing=True)
+#        self.check(self.session, 'read_timestamp=' + timestamp_str(key_range / 2),
+#            dict((k, 1) for k in keys[:(key_range / 2)]))
+#        self.check(self.session, 'read_timestamp=' + timestamp_str(key_range / 2),
+#            dict((k, 1) for k in keys[(key_range / 2 + 1):]), missing=True)

         # Bump the oldest timestamp, we're not going back...
         self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(key_range / 2))
@@ -127,8 +130,8 @@ class test_timestamp04(wttest.WiredTigerTestCase, suite_subprocess):
             self.session.commit_transaction('commit_timestamp=' + timestamp_str(k + key_range))

         # Now we should have: keys 1-100 with value 2
-        self.check(self.session, 'read_timestamp=' + timestamp_str(2 * key_range),
-            dict((k, 2) for k in keys[:]))
+#        self.check(self.session, 'read_timestamp=' + timestamp_str(2 * key_range),
+#            dict((k, 2) for k in keys[:]))

         # Rollback a quater of the new commits
         self.conn.set_timestamp('stable_timestamp=' + timestamp_str(1 + key_range + key_range / 4))
```

The output to console is:

```
test_timestamp04.test_timestamp04.test_basic -s 2 (row) ... 
Rollback to stable starting
Rollback to stable finished
Did an update restore eviction on 0x1dc8e80
Did an update restore eviction on 0x1dc3a00
Rollback to stable starting
Rollback to stable finished
Did an update restore eviction on 0x1f27b40
ERROR
```

There are also two files created with dumps of the pages from cache. `build_posix/WT_TEST/test_timestamp04.0/tmp.txt` contains the content in cache immediately after an eviction (I don't think this includes the page that has just been evicted) , and `build_posix/WT_TEST/test_timestamp04.0/tmp2.txt` contains the state of the pages when starting the cursor traversal.  That shows a page with records like:

```
K {#c1#b3}
        V {#81}
        K {#c1#b4}
        V {#81}
        K {#c2:}
        V {}
        K {#c2;}
        V {}
        K {#c2<}
        V {}
```

Which isn't expected in this workload - there shouldn't be 0 length values.

keithbostic: @agorrod, if you have a few minutes, let's talk this over. What's happening here is that we're inserting 0-length values on the pages, but at the same time adding globally visible updates in front of those 0-length values, so no cursor retrieval should ever see them (although page-dump will display them of course).

keithbostic:
@agorrod,  f87b5ae fixes up some comments in `test_timestamp04`.

The last change seems wrong because it doesn't match the code we're about to execute to check the table: `dict((k, (2 if j <= (key_range / 4) else 1)`.

But, if you print out the actual values in the table at the interesting range:
```
2495,2
2496,2
2497,2
2498,2
2499,2
2500,2
2501,2
2502,1
2503,1
2504,1
```
It shows we switch to a value of 1 at `key_range / 4 + 2`, not `key_range / 4 + 1`. I'm sure I'm missing something obvious...

keithbostic:
autobuild: retest this please

michaelcahill:
Thanks @keithbostic, it still takes some staring to realize that this use of `__rec_append_orig_value` will in fact append a tombstone, but this change is a definite improvement and I'd like to get it into regular testing.

PR #3622, opened 2017-08-28
WT-3545 Fix array bounds warnings with gcc 7.

michaelcahill:


michaelcahill:
GCC 7.0 appears to need some padding to be confident that WiredTiger won't read past the end of an array (AFAICT, this warning is coming from the unpack calls).  We check that we don't read past the end, so this appears spurious.  I've added some padding and asserted that we don't go past the expected end of the array.

agorrod:
lgtm

PR #3625, opened 2017-08-28
WT-3544 Fix memory stomping during rollback_to_stable.

agorrod:
This was caused by an off-by-one bug calculating the size of a
bitstring, leading to writes past the end of the array when the largest
active file ID was a multiple of 8.


agorrod:
I'm re-merging this, since it was committed with the wrong JIRA ticket tag first time around (sorry)

PR #3626, opened 2017-08-28
WT-3546 Coverity 1380021 dereference before NULL check, lint

keithbostic:


agorrod:
Thanks @keithbostic lgtm

PR #3627, opened 2017-08-29
WT-3547 Fix a deadlock dropping LSM table handles.

michaelcahill:


michaelcahill:
@agorrod, I don't love this but it avoids the deadlock for now.  Can you please review?

FTR, the problem is holding an exclusive "table:" handle lock across a call to `WT_WITHOUT_LOCKS`, which drops locks while waiting for an LSM tree to quiesce.

I'd like to fix this properly, by taking and holding table locks for the duration of operations.  To get there, I think we'd probably have to:
(a) convert LSM to handle locks
(b) add support for nested read locks on data handles while holding an exclusive lock
(c) add support for waiting for exclusive access to an LSM handle (i.e., some way to avoid deadlock that doesn't involve `WT_WITHOUT_LOCKS`).
(d) acquire and hold exclusive table locks across create, drop, rename, etc.

That would be most of the way to completing WT-1598.

agorrod:
Thanks @michaelcahill lgtm - I don't love the compromise either, but agree that pulling the piece of string for the better solution isn't worthwhile.

PR #3628, opened 2017-08-29
WT-3512 Add timestamp API python test

sulabhM:


sulabhM:
Pull request test fails with test/format. test/format is committing a transaction with commit_timestamp older than the oldest timestamp.

sueloverso:
autobuild: retest this please

sulabhM:
@sueloverso , please have a look

src/txn/txn.c
```diff
@@ -592,8 +590,29 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+		/*
+		 * commit timestamp being set should be newer than the oldest
+		 * timestamp and should move forward in a transaction
+		 */
+		__wt_readlock(session, &txn_global->rwlock);
+		if (txn_global->has_oldest_timestamp && __wt_timestamp_cmp(&ts,
+		    &txn_global->oldest_timestamp) < 0) {
+			__wt_readunlock(session, &txn_global->rwlock);
+			WT_ERR_MSG(session, EINVAL,
+			    "commit timestamp %.*s older than oldest timestamp",
```
agorrod: Do we use `%.*s` to print timestamps anywhere else? I know this is from a configuration string, so it's probably OK, I'm just curious.

src/txn/txn.c
```diff
@@ -592,8 +590,29 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+
+		WT_ERR(__wt_txn_parse_timestamp(session, "commit", &ts, &cval));
+		/*
+		 * commit timestamp being set should be newer than the oldest
+		 * timestamp and should move forward in a transaction
+		 */
+		__wt_readlock(session, &txn_global->rwlock);
+		if (txn_global->has_oldest_timestamp && __wt_timestamp_cmp(&ts,
+		    &txn_global->oldest_timestamp) < 0) {
+			__wt_readunlock(session, &txn_global->rwlock);
```
agorrod: It would be better if the code could be structures so there aren't two calls to __wt_readunlock here, since releasing resources in return paths from the middle of a function can be prone to introducing bugs in the future. Something like the structure here https://github.com/wiredtiger/wiredtiger/blob/develop/src/reconcile/rec_write.c#L1811 might work, but the message output means it would require an `if (ret != 0)` as well.

src/txn/txn.c
```diff
@@ -592,8 +590,29 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	    __wt_config_gets_def(session, cfg, "commit_timestamp", 0, &cval));
 	if (cval.len != 0) {
 #ifdef HAVE_TIMESTAMPS
-		WT_ERR(__wt_txn_parse_timestamp(
-		    session, "commit", &txn->commit_timestamp, &cval));
+		wt_timestamp_t ts;
```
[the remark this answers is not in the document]
sulabhM: done

src/txn/txn.c
```diff
@@ -592,8 +590,29 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+
+		WT_ERR(__wt_txn_parse_timestamp(session, "commit", &ts, &cval));
+		/*
+		 * commit timestamp being set should be newer than the oldest
+		 * timestamp and should move forward in a transaction
```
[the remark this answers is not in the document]
sulabhM: done

src/txn/txn.c
sulabhM: done

src/txn/txn.c
sulabhM: Yeah, I think we do this everywhere we have config string, I copied from elsewhere.

agorrod:
Thanks @sulabhM lgtm - @sueloverso or @michaelcahill  please do a review as well

src/txn/txn.c
```diff
@@ -592,8 +590,28 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+			    "commit timestamp %.*s older than oldest timestamp",
+			    (int)cval.len, cval.str);
+
+		if (F_ISSET(txn, WT_TXN_HAS_TS_COMMIT) &&
+		    __wt_timestamp_cmp(&ts, &txn->commit_timestamp) < 0)
+			WT_ERR_MSG(session, EINVAL,
+			    "commit timestamp %.*s older than the current "
+			    "commit timestamp set for this transaction",
+			    (int)cval.len, cval.str);
+		__wt_timestamp_set(&txn->commit_timestamp, &ts);
```
sueloverso: This code change and the `wt_txn_commit` change above are essentially identical.  It would be good to factor that out.  I agree with Alex that having to take the read lock on every `wt_txn_commit` makes me sad.

test/suite/test_timestamp08.py
```diff
@@ -0,0 +1,187 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# test_timestamp08.py
```
sueloverso: This is the same test name/file name as Vamsi added in https://github.com/wiredtiger/wiredtiger/pull/3641/files.

test/suite/test_timestamp08.py
```diff
@@ -0,0 +1,187 @@
...
+    uri = 'table:' + tablename
+
+    def test_timestamp_api(self):
+        if not wiredtiger.timestamp_build():
+            self.skipTest('requires a timestamp build')
+
+        self.session.create(self.uri, 'key_format=i,value_format=i')
+        c = self.session.open_cursor(self.uri)
+
+        # Begin by adding some data
```
sueloverso: For all the comments in here: Capitalize the beginning of the sentence and end with a period '.'.

test/suite/test_timestamp08.py
```diff
@@ -0,0 +1,187 @@
...
+        # commit timestamp >= stable timestamp, stable timestamp is set to 5 at
+        # the moment.
+        # Also, when explicitly set, commit timestamp for a transaction can be
+        # earlier than the commit timestamp of an earlier transaction
+        self.session.begin_transaction()
+        c[2] = 2
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+            lambda: self.session.commit_transaction(
+                'commit_timestamp=' + timestamp_str(2)),
+                '/older than oldest timestamp/')
```
sueloverso: The comment discusses the stable timestamp and the code/error message use oldest timestamp.  One or the other is probably incorrect.

test/suite/test_txn17.py
```diff
@@ -0,0 +1,81 @@
...
+
+from suite_subprocess import suite_subprocess
+import wiredtiger, wttest
+
+def timestamp_str(t):
+    return '%x' % t
+
+class test_txn17(wttest.WiredTigerTestCase, suite_subprocess):
+    def test_txn_api(self):
+        # Test API functionality tagged as requires_transaction
```
sueloverso: Again, periods at the end of the sentences.

src/txn/txn.c
```diff
@@ -592,8 +590,28 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+		WT_ERR(__wt_txn_parse_timestamp(session, "commit", &ts, &cval));
+		/*
+		 * Commit timestamp being set should be newer than the oldest
+		 * timestamp and should move forward in a transaction.
+		 */
+		__wt_readlock(session, &txn_global->rwlock);
+		if (txn_global->has_oldest_timestamp && __wt_timestamp_cmp(&ts,
+		    &txn_global->oldest_timestamp) < 0)
+			ret = EINVAL;
+		__wt_readunlock(session, &txn_global->rwlock);
```
michaelcahill: I think this would be better as:

```c
WT_WITH_TIMESTAMP_READLOCK(session, &txn_global->rwlock,
   ret = (txn_global->has_oldest_timestamp && __wt_timestamp_cmp(&ts,
        &txn_global->oldest_timestamp) < 0) ? EINVAL : 0)
```

test/suite/test_timestamp08.py
sulabhM: good catch.
This isn't right, I wanted to check if commit ts >= stable ts and I haven't checked that correctly. I will have to put the test case in, as well as the code to catch it through timestamp_transaction and commit_transaction

src/txn/txn_timestamp.c
```diff
@@ -459,6 +464,50 @@ __wt_txn_global_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
 	return (0);
 }
 
+#ifdef HAVE_TIMESTAMPS
+/*
+ * __wt_txn_commit_ts_validate --
+ *	Validate commit timestamp being set
+ */
+int
+__wt_txn_commit_ts_validate(WT_SESSION_IMPL *session,
```
agorrod: The other timestamp APIs all use `timestamp` not `ts` in their names. i.e: `__wt_txn_commit_timestamp_validate`

sueloverso: Can this be extended to include validation of other timestamps?  How about `wt_timestamp_validate`?  I can see that validating read timestamps is similar.  Maybe pass in a flag such as read or commit for the non-overlapping parts.

agorrod:
Thanks @sulabhM lgtm - I agree that the three bools aren't ideal, but I agree it's the best solution for now.

PR #3629, opened 2017-08-29
WT-3431 Replace WT_TXN_STATE with sorted lists.

michaelcahill:
Each list is protected by a separate rwlock, allowing us to create
snapshots without sorting and avoid any transaction operations having to
scan sessions.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -276,7 +276,7 @@ __wt_txn_oldest_id(WT_SESSION_IMPL *session)
 	 * the active checkpoint then it's safe to ignore the checkpoint ID in
 	 * the visibility check.
 	 */
-	checkpoint_pinned = txn_global->checkpoint_state.pinned_id;
+	checkpoint_pinned = txn_global->checkpoint_pinned_id;
 	if (checkpoint_pinned == WT_TXN_NONE ||
 	    WT_TXNID_LT(oldest_id, checkpoint_pinned))
 		return (oldest_id);
@@ -490,6 +490,8 @@ __wt_txn_begin(WT_SESSION_IMPL *session, const char *cfg[])
 		if (session->ncursors > 0)
 			WT_RET(__wt_session_copy_values(session));
 
+		__wt_txn_release_snapshot(session);
+
 		/*
 		 * We're about to allocate a snapshot: if we need to block for
 		 * eviction, it's better to do it beforehand.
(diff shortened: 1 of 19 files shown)


michaelcahill:
I'm going to do another pass through this branch, please hold off on review until I've done that: some more cleanup should be possible.

src/include/txn.i
```diff
@@ -532,17 +534,15 @@ static inline int
...
 	txn = &session->txn;
-	txn_state = WT_SESSION_TXN_STATE(session);
 
 	/*
 	 * Check the published snap_min because read-uncommitted never sets
 	 * WT_TXN_HAS_SNAPSHOT.
 	 */
 	if (F_ISSET(txn, WT_TXN_RUNNING) &&
-	    !F_ISSET(txn, WT_TXN_HAS_ID) && txn_state->pinned_id == WT_TXN_NONE)
+	    !F_ISSET(txn, WT_TXN_PUBLIC_ID | WT_TXN_PUBLIC_PINNED_ID))
```
agorrod: I don't think this is right. My understanding is that this will return true if WT_TXN_PUBLIC_ID and/or WT_TXN_PUBLIC_PINNED_ID is set. The previous test was ensuring they are both set.

src/txn/txn_global.c
```diff
@@ -0,0 +1,627 @@
...
+	    prev = TAILQ_PREV(prev, __wt_txn_mp_qh, metadata_pinnedq))
+		;
+	if (prev == NULL)
+		TAILQ_INSERT_HEAD(
+		    &txn_global->metadata_pinnedh, txn, metadata_pinnedq);
+	else
+		TAILQ_INSERT_AFTER(
+		    &txn_global->metadata_pinnedh, prev, txn, metadata_pinnedq);
+	__wt_writeunlock(session, &txn_global->metadata_pinned_rwlock);
+	F_SET(txn, WT_TXN_PUBLIC_METADATA_PINNED);
```
agorrod: I'm surprised setting this happens after the lock is released - could you add a comment explaining why?

src/txn/txn_global.c
```diff
@@ -0,0 +1,627 @@
...
+	WT_TXN_GLOBAL *txn_global;
+
+	txn = &session->txn;
+	txn_global = &S2C(session)->txn_global;
+
+	if (!F_ISSET(txn, WT_TXN_PUBLIC_PINNED_ID))
+		return;
+
+	__wt_writelock(session, &txn_global->pinned_id_rwlock);
+	TAILQ_REMOVE(&txn_global->pinned_idh, txn, pinned_idq);
```
agorrod: Since the F_CLR happens outside the lock - should this re-check that there is an ID pinned before removing? Or does TAILQ_REMOVE handle the case where the item doesn't exist?

src/txn/txn_global.c
```diff
@@ -0,0 +1,627 @@
...
+
+	conn = S2C(session);
+	txn_global = &conn->txn_global;
+	oldest_session = NULL;
+
+	last_running = oldest_id = txn_global->current;
+	if ((metadata_pinned = txn_global->checkpoint_txn_id) == WT_TXN_NONE)
+		metadata_pinned = oldest_id;
+
+	__wt_readlock(session, &txn_global->id_rwlock);
```
agorrod: I'm a bit surprised that these three IDs are retrieved under different locks - looking further it seems as though callers can hold the broader txn lock if they need to see a consistent view. Maybe adding a comment about that here would be worthwhile?

michaelcahill:
This had a dramatic, negative effect on performance, see http://build.wiredtiger.com:8080/job/wiredtiger-perf-btree/plot/ for example.  More work will be needed to get this ready to merge.

agorrod:
We aren't going to pull this one in now - I've attached a diff to the ticket, and we can resuscitate the branch if required.

PR #3630, opened 2017-08-30
WT-3549 Fix a deadlock sweeping table handles.

michaelcahill:
We need to acquire the global table lock before the table handle lock.


michaelcahill:
@agorrod, if you can review this fix, that would be great.  I could repro as described in the ticket running `test_schema03` with aggressive sweep settings.  With this fix, I get 100+ runs with no stall.

I think this is fairly straightforward: we need to always acquire the global table lock before locking a table.

michaelcahill:
cc @sueloverso, thanks again for chasing this down.

agorrod:
Thanks @michaelcahill lgtm

PR #3631, opened 2017-08-30
WT-3540 For debugging only copy log files.

sueloverso:
Tables can be so large that they fill test machine file systems.


sueloverso:
Self-reviewing.

PR #3632, opened 2017-08-30
WT-3495 Don't ftruncate if log cursors are open

sueloverso:
@agorrod Please review this.  It avoids the ftruncate if there are log cursors open.  Other than test programs log cursors are very rarely used so I don't think it has much user-visible impact.


agorrod:
Thanks @sueloverso lgtm

PR #3634, opened 2017-08-31
WT-3557 format should ignore incompatible keywords

keithbostic:


michaelcahill:
I took a look @keithbostic, lgtm.  I'll merge.

PR #3635, opened 2017-08-31
WT-3468 Avoid shared timespec usage in LSM

sueloverso:
@agorrod Please review.  I implemented a form of @michaelcahill 's suggestions.  I have an atomic work unit counter and only the server thread looks for idleness and keeps track of how long a tree has been idle.

I've run `make check` and all the `run.py lsm` and `test/format` with `data_source=lsm` for over 100 iterations.  But I do not know for sure whether or not the lsm manager idle settings have kicked in or not.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -785,6 +785,7 @@ iS
 iSh
 ibackup
 icount
+idlems
 idx
 ifdef
 ifdef's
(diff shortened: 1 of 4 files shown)


michaelcahill:
Thanks @sueloverso, these changes lgtm.  I'll merge.

PR #3636, opened 2017-08-31
WT-3524 Test format likely to pin cache full when enabling timestamps

keithbostic:


agorrod:
@keithbostic this change overlaps with work @sulabhM has done in WT-3512  - I'd like to see how those changes fall out before merging this one.

keithbostic:
@sulabhM, one note, this fixes a bug in 770ecb1: in the function `test/format/ops.c: commit_transaction()` the current develop branch code releases a lock we don't hold, if `g.c_txn_timestamps` isn't set.

PR #3637, opened 2017-08-31
WT-3555 Streamline open_cursor for simple tables.

ddanderson:
This change results in a consistent 20% improvement for my focused benchmark that repeatedly calls WT_SESSION->open_cursor/close_cursor in a single thread.

### DIFF
--- a/src/cursor/cur_std.c
+++ b/src/cursor/cur_std.c
@@ -570,7 +570,8 @@ __wt_cursor_close(WT_CURSOR *cursor)
 	__wt_buf_free(session, &cursor->value);
 
 	__wt_free(session, cursor->internal_uri);
-	__wt_free(session, cursor->uri);
+	if (!F_ISSET(cursor, WT_CURSTD_URI_SHARED))
+		__wt_free(session, cursor->uri);
 	__wt_overwrite_and_free(session, cursor);
 	return (0);
 }
--- a/src/cursor/cur_table.c
+++ b/src/cursor/cur_table.c
@@ -998,8 +998,11 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 		if (ret == 0) {
 			/* Fix up the public URI to match what was passed in. */
 			cursor = *cursorp;
-			__wt_free(session, cursor->uri);
-			WT_TRET(__wt_strdup(session, uri, &cursor->uri));
+			if (!F_ISSET(cursor, WT_CURSTD_URI_SHARED))
+				__wt_free(session, cursor->uri);
+			cursor->uri = table->iface.name;
+			WT_ASSERT(session, strcmp(uri, cursor->uri) == 0);
+			F_SET(cursor, WT_CURSTD_URI_SHARED);
 		}
 		return (ret);
 	}
--- a/src/cursor/cur_table.c
+++ b/src/cursor/cur_table.c
@@ -987,8 +987,12 @@ __wt_curtable_open(WT_SESSION_IMPL *session,
 
 	if (table->is_simple) {
 		/* Just return a cursor on the underlying data source. */
-		ret = __wt_open_cursor(session,
-		    table->cgroups[0]->source, NULL, cfg, cursorp);
+		if (table->is_simple_file)
+			ret = __wt_curfile_open(session,
+			    table->cgroups[0]->source, NULL, cfg, cursorp);
+		else
+			ret = __wt_open_cursor(session,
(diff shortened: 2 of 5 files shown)


ddanderson:
retest this please

ddanderson:
@michaelcahill, could you review this please?

michaelcahill:
@ddanderson, I was surprised by the 20% figure, since `__wt_open_cursor` doesn't do much.

One thing it does do, which would be lost by this change, is get the correct URI in the returned cursor: https://github.com/wiredtiger/wiredtiger/blob/develop/src/session/session_api.c#L370-L379

Does that `strdup` account for the performance difference here?  Could we get the same effect some other way (i.e., point `cursor->uri` at `table->name` and set a flag so it isn't freed when the cursor is closed)?

michaelcahill:
@ddanderson, I posted too soon.

As far as I can tell, the `strdup` I mention above shouldn't be necessary: cursors should always have a non-NULL `uri`.  Does taking out that code help in the workload you're measuring (without this branch)?

My comment about avoiding `strdup` does apply to this code: https://github.com/wiredtiger/wiredtiger/blob/develop/src/cursor/cur_table.c#L995-L998 -- does that appear in profiling?  If so, we can probably find a way to avoid it and have `cursor->uri` point at  `table->iface.name` instead.

ddanderson:
@michaelcahill, 

> One thing it does do, which would be lost by this change, is get the correct URI in the returned cursor: https://github.com/wiredtiger/wiredtiger/blob/develop/src/session/session_api.c#L370-L379

In this case, I don't think we care about it.  It's true that with this bypass, we don't have the current URI in the cursor when it's returned after _this_ call to `__wt_curfile_open`.  It would have been filled in with `"file:...."` anyway, not what we we'll need eventually.  When the call stack gets back to the original call to `__session_open_cursor_int`, we will fill `cursor->uri` with the original uri, that is, `"table:..."`.  Does that make sense?  Yes, avoiding that extra `strdup` is probably the biggest gain with this shortcut, we also avoid having to parse `"file:"`, that's a minor win.

> My comment about avoiding strdup does apply to this code: https://github.com/wiredtiger/wiredtiger/blob/develop/src/cursor/cur_table.c#L995-L998 -- does that appear in profiling?

Great idea, I'll try it tomorrow and see.

ddanderson:
@michaelcahill, here are some results for various changes and combinations.  For each change, I made two runs, and the first two numbers are the opens per second (averaged over 60 seconds).  Next to that is the calculated nanoseconds per open (from the two runs averaged), and the nanosecond improvement over the baseline.  FTR, the test is using a table name that looks like ones that mongodb is using, since the length may be relevant in doing `strcmp`, etc. Also, I'm using`tcmalloc` with `LD_PRELOAD`.

- ### Baseline   (develop branch)
   1295898, 1290325   773ns per open

- ### shared table uri (Michael's suggested sharing of the uri for tables)
   1445968, 1461024   668ns per open  -105ns

- ### shared file uri (Michael's suggested sharing of the uri for files)
   1409769, 1404583   710ns per open  -63ns

- ### shortcut file:  (the change proposed on this branch)
   1554967, 1560085   642ns per open  -131ns

- ### shared table uri & shared file uri  (combining two changes from above)
  1665350, 1685993   596ns per open  -177ns

- ### shared table uri & shortcut file:  (combining two changes from above)
  1816898, 1794468   553ns per open  -220ns

michaelcahill:
Baseline (develop branch)

1295898, 1290325 773ns per open

shared table uri (Michael's suggested sharing of the uri for tables)

1445968, 1461024 668ns per open -105ns

shared file uri (Michael's suggested sharing of the uri for files)

1409769, 1404583 710ns per open -63ns

shortcut file: (the change proposed on this branch)

1554967, 1560085 642ns per open -131ns

shared table uri & shared file uri (combining two changes from above)

1665350, 1685993 596ns per open -177ns

shared table uri & shortcut file: (combining two changes from above)

1816898, 1794468 553ns per open -220ns

michaelcahill:
@ddanderson, given the nearly 30% speedup, we should go with the final version you measured.  Did you end up with a new flag indicating whether the URI should be freed when the cursor is closed?

ddanderson:
@michaelcahill, ready to review.  I used a flag indicating sharing, and just pushed my changes.  If you prefer I can make the flag the inverse.  `WT_CURSTD_URI_OWNED` ?

michaelcahill:
Thanks, @ddanderson -- lgtm and the perf results are compelling.  I'll merge.

PR #3638, opened 2017-09-01
WT-3558 Disable metadata visibility assertion in checkpoints

agorrod:
It is better to return a WT_ROLLBACK so the checkpoint can be retried.


agorrod:
@michaelcahill Would you mind reviewing this change please?

michaelcahill:
Thanks @agorrod, lgtm for now until WT-3559 fixes this properly.  I'll merge.

PR #3639, opened 2017-09-01
WT-3537 Split pages when nothing can be written

keithbostic:


sueloverso:
This LGTM as the cache size isn't integral to the intent of the test.

PR #3644, opened 2017-09-05
WT-3435 Fix lookaside eviction with timestamps.

michaelcahill:
Previously we disabled lookaside eviction with timestamps more recent than the stable timestamp, which has the effect of disabling lookaside eviction.  This branch changes which updates are used to create page images so that checkpoints as-of a stable timestamp are correct.


michaelcahill:
!!! WIP only, not ready for review / merge.

In particular, as described in the ticket, we need to track which WT_REFs have lookaside records associated so that subsequent checkpoints can update the versions of data on disk.

michaelcahill:
autobuild: retest this please

src/btree/bt_sync.c
```diff
@@ -62,6 +62,90 @@ __sync_checkpoint_can_skip(WT_SESSION_IMPL *session, WT_PAGE *page)
...
+	*skip_walkp = false;
+
+	/*
+	 * Hedge our bets: save a ref on either side of the page we're
+	 * evicting.
+	 *
+	 * !!! Otherwise, the page could split in the meantime, so backing up
+	 * later is not safe.  We allow in-memory splits during checkpoints,
+	 * for example.
+	 */
```
agorrod: I'm glad you put a big comment here, but this seems a bit alarming to me. I don't have another suggestion, and I see that it's not much different to what was happening before - it's just more explicit.

src/cache/cache_las.c
```diff
@@ -280,129 +280,61 @@ __wt_las_cursor_close(
 }
 
 /*
- * __wt_las_sweep --
- *	Sweep the lookaside table.
+ * __wt_las_remove_block --
+ *	Remove all records matching a key prefix from the lookaside store.
  */
 int
-__wt_las_sweep(WT_SESSION_IMPL *session)
```
agorrod: Getting rid of lookaside sweep is a nice simplification

michaelcahill:
TODO:

- [ ] stress testing shows that this branch is leaking blocks -- I suspect some path setting a ref to 
lookaside probably needs to free the old address.
- [ ] implement the optimization of tracking the minimum timestamp in lookaside for a page, and skipping reading pages back in for a checkpoint if the checkpoint's `read_timestamp` < `page->modify->page_las->min_timestamp`

src/btree/bt_read.c
```diff
@@ -281,22 +204,18 @@ __las_page_instantiate(WT_SESSION_IMPL *session,
 		WT_ILLEGAL_VALUE_ERR(session);
 		}
 
+	/* The page is instantiated, we no longer need the lookaside entries. */
+	WT_ERR(__wt_las_remove_block(
+	    session, cursor, read_id, ref->page_las->las_pageid));
+
```
keithbostic: `__last_page_instantiate()` can still fail without instantiating the page. Is it OK to remove the LAS entries before we know the page will be successfully instantiated?

A similar question about the caller, `__page_read()`. Currently, the last thing `__page_read()` does is call `__las_page_instantiate()`, so it can't fail after the call, however, it would be easy to add new code to that function that could fail without noticing the LAS entries had already been removed. If this is an issue, maybe a comment?

Maybe move the call to `__wt_las_remove_block()` into `__page_read()`, done after the page has been successfully instantiated.

src/evict/evict_lru.c
```diff
@@ -1870,9 +1872,27 @@ __evict_walk_file(WT_SESSION_IMPL *session,
 
 		/* Pages that are empty or from dead trees are fast-tracked. */
 		if (__wt_page_is_empty(page) ||
-		    F_ISSET(session->dhandle, WT_DHANDLE_DEAD))
+		    F_ISSET(session->dhandle, WT_DHANDLE_DEAD) ||
+		    F_ISSET(btree, WT_BTREE_LOOKASIDE))
```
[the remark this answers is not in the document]
keithbostic: This test will skip the `if (WT_PAGE_IS_INTERNAL(page))` test, which may not be what you wanted.

src/evict/evict_lru.c
```diff
@@ -1870,9 +1872,27 @@ __evict_walk_file(WT_SESSION_IMPL *session,
...
+		    WT_CACHE_EVICT_DIRTY_HARD) &&
+		    !F_ISSET(conn, WT_CONN_EVICTION_NO_LOOKASIDE) &&
+		    F_ISSET(cache, WT_CACHE_EVICT_DIRTY) &&
+		    !modified && page->modify != NULL &&
+		    !__wt_txn_visible_all(session, page->modify->rec_max_txn,
+		    WT_TIMESTAMP_NULL(&page->modify->rec_max_timestamp))) {
+			__wt_page_only_modify_set(session, page);
+			modified = true;
+		}
+
```
keithbostic: This clause is pretty obscure.

What does it mean to be "blocked by eviction"?

I think the `F_ISSET(cache, WT_CACHE_EVICT_DIRTY)` test is intended to mean "if we mark this page dirty will it be chosen?" but that test is only a few lines further down in the same function. Does it have some other meaning, or couldn't we just jump to "fast"?

Instead of calling `__wt_page_only_modify_set()` here, would it be better to enhance the final clause of `__wt_page_can_evict()` with some version of this test?

Finally, it seems like this test will potentially fail the subsequent check: `If the oldest transaction hasn't changed since the last time this page was written, it's unlikely we can make progress`. Doesn't that minimize the usefulness of this check? (If we shuffled the test order a little in this function, I think we could just jump to the "fast" label if this clause tests true.)

src/evict/evict_page.c
```diff
@@ -353,19 +352,32 @@ __evict_page_dirty_update(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
...
-		ref->addr = addr;
+		if (mod->mod_replace.addr == NULL)
+			ref->addr = NULL;
+		else {
+			WT_RET(__wt_calloc_one(session, &addr));
+			*addr = mod->mod_replace;
+			mod->mod_replace.addr = NULL;
+			mod->mod_replace.size = 0;
+			ref->addr = addr;
+		}
```
keithbostic: I don't understand this change. I wouldn't expect to see a `WT_PM_REC_REPLACE` update with `mod->mod_replace.addr == NULL` unless it was a root page (which wouldn't get here).

src/evict/evict_page.c
```diff
@@ -545,22 +559,26 @@ __evict_review(
...
 
 			/*
 			 * Check if reconciliation suggests trying the
 			 * lookaside table.
 			 */
-			lookaside_retryp = &lookaside_retry;
+			if (F_ISSET(cache, WT_CACHE_EVICT_DIRTY_HARD |
+			    WT_CACHE_EVICT_DIRTY_HARD) &&
+			    !F_ISSET(conn, WT_CONN_EVICTION_NO_LOOKASIDE))
+				lookaside_retryp = &lookaside_retry;
```
keithbostic: The `WT_CACHE_EVICT_DIRTY_HARD` test is being repeated.

src/evict/evict_page.c
```diff
@@ -574,9 +592,9 @@ __evict_review(
 	 * table, allowing the eviction of pages we'd otherwise have to retain
 	 * in cache to support older readers.
 	 */
-	if (ret == EBUSY && lookaside_retry && __wt_cache_stuck(session)) {
-		LF_CLR(WT_EVICT_SCRUB | WT_EVICT_UPDATE_RESTORE);
-		LF_SET(WT_EVICT_LOOKASIDE);
+	if (ret == EBUSY && lookaside_retry) {
+		LF_CLR(WT_REC_SCRUB | WT_REC_UPDATE_RESTORE);
+		LF_SET(WT_REC_LOOKASIDE);
```
keithbostic: Why did we stop checking `cache->evict_aggressive_score`?

src/reconcile/rec_write.c
```diff
@@ -569,17 +568,19 @@ __rec_write_check_complete(
...
+	 * likely get to write at least one of the blocks.  If we've created a
+	 * page image for a page that previously didn't have one, or we had a
+	 * page image and it is now empty, that's also progress.
 	 */
-	if (lookaside_retryp == NULL)
+	if (r->multi_next > 1)
+		return (0);
+	if (r->multi_next == 0 && r->page->dsk != NULL)
+		return (0);
+	if (r->multi_next == 1 && r->page->dsk == NULL)
```
keithbostic: It would be useful to enhance this comment to say more about the paths being detected here -- I can't off-hand think of how we might have had a disk image and it would be gone without the page being discarded?

keithbostic:
@michaelcahill, in `__rec_split_write()`, the move of the `copy_image:` label to include `__rec_update_las` is interesting. If we match the previous block based on its checksum, we jump to that label, and my current belief is there's no reason to update the LAS table if we don't need to rewrite the previous block?

src/reconcile/rec_write.c
```diff
@@ -1280,131 +1257,127 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+				if (prev_upd != NULL)
+					prev_upd->next = upd->next;
+				else {
+					first_upd = upd->next;
+					if (ins != NULL)
+						ins->upd = first_upd;
+					else
+						page->modify->mod_row_update[
+						    WT_ROW_SLOT(page, ripcip)] =
+						    first_upd;
```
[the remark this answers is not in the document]
michaelcahill: `WT_ROW_UPDATE` isn't an lvalue.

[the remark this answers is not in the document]
michaelcahill: To be clear, this attempt to discard aborts is orthogonal to the lookaside change and I don't have a workload where it is critical.  I'm inclined to split this out into a separate ticket.

src/evict/evict_page.c
michaelcahill: This is the case where no updates were visible (so there is no page image), but updates were saved to the lookaside table.

src/evict/evict_page.c
michaelcahill: It's checked above via the `_HARD` flags, before setting `lookaside_retryp`.

michaelcahill:
@keithbostic, I fixed a few bugs today, but still haven't got everything working.  (*Then* I planned to work through your review comments).

If you have time to work through any of the remaining issues, that would be great.  There is (at least) a weird failure in `test_cursor_random02.py`, and `test/format` stress runs will fail fairly quickly with snapshot isolation failures or worse.

keithbostic:
@michaelcahill, what's happening with `test_cursor_random02.py` is:

* we're checkpointing `file:random-000002.lsm` as part of closing the handle,
* it's a single leaf page with a single internal/parent root page,
* we checkpoint the leaf page,
* `page.read_gen == WT_READGEN_OLDEST` so we try and evict it,
* because there's only a single leaf page, both `prev` and `next` are the parent/root page,
* page eviction is successful and the state is `WT_PM_REC_MULTIBLOCK`,
* we try to do the split, which fails (because we can't lock the parent page during a checkpoint), returning `EBUSY`,
* we switch to the saved `prev` page, then do a walk-next, which, since we're in the root page, returns the walk is complete, and we've never checkpointed the internal page,
* the subsequent `__wt_evict_file` fails because there's a dirty page in the tree.

I'll stare at this some tomorrow, but I'm pessimistic about trying to move back and forth in the tree while guaranteeing a visit to every page.

keithbostic:
@michaelcahill, a random thought. Rather than pass the `checkpointing` flag, would it be reasonable to have a session checkpoint thread flag? (We already have `WT_SESSION_LOCKED_CHECKPOINT` which would probably work, and we could hide the check behind a more appropriately named macro.)

Anyway, I'm anticipating future changes in reconciliation/eviction/splitting based on checkpoint eviction, and that might be simpler than driving the `checkpointing` flag all the way through the system.

keithbostic:
@michaelcahill, I'm also see `test_timestamp04` timing out.

What's happening there is that in the `row-smallcache` test, all of the pages are in the LAS file and none of them can be evicted because they fail the `__rec_txn_read` visibility tests:
```
cache dump
cache full: yes
cache clean check: yes (201%)
cache dirty check: yes (198%)
file:ts04_nots_nologged.wt(<live>):
internal: 1 pages, 0MB, 0/1 clean/dirty pages, 0/0 clean/dirty MB, 0MB max page, 0MB max dirty page
file:ts04_nots_logged.wt(<live>):
internal: 1 pages, 0MB, 0/1 clean/dirty pages, 0/0 clean/dirty MB, 0MB max page, 0MB max dirty page
file:ts04_ts_nologged.wt(<live>):
internal: 1 pages, 0MB, 0/1 clean/dirty pages, 0/0 clean/dirty MB, 0MB max page, 0MB max dirty page
file:ts04_ts_logged.wt(<live>):
internal: 1 pages, 0MB, 0/1 clean/dirty pages, 0/0 clean/dirty MB, 0MB max page, 0MB max dirty page
file:WiredTigerLAS.wt(<live>):
internal: 1 pages, 0MB, 0/1 clean/dirty pages, 0/0 clean/dirty MB, 0MB max page, 0MB max dirty page
leaf: 23 pages, 3MB, 0/23 clean/dirty pages, 0/3 clean/dirty MB, 2MB max page, 2MB max dirty page
file:WiredTiger.wt(<live>):
internal: 1 pages, 0MB, 1/0 clean/dirty pages, 0/0 clean/dirty MB, 0MB max page, 0MB max dirty page
cache dump: total found: 4MB vs tracked inuse 4MB
total dirty bytes: 3MB
```

And the application threads are just spinning in `__wt_cache_eviction_worker`.

I didn't take a look at the visibility tests (I'm assuming they're correct), but it occurred to me that when we update the LAS table in `__rec_update_las`, we give the update the current session's transaction ID (or an allocated transaction ID if the session doesn't have one), which is at least going to delay our ability to evict those pages.

Would it be better to use the update transaction already associated with the update record itself when adding an update to the LAS table?

I don't see any obvious way to drill that hole, but it might be a way to attack this problem.

keithbostic:
@michaelcahill, note, this branch sets but no longer uses `WT_SAVE_UPDATE.onpage_upd`.

src/reconcile/rec_write.c
```diff
@@ -1192,138 +1169,109 @@ err:	__wt_scr_free(session, &tmp);
...
-				max_txn = txnid;
+		if (WT_TXNID_LT(max_txn, txnid))
+			max_txn = txnid;
 
+		/*
+		 * Find the first update we can use.
+		 *
+		 * Update/restore eviction can handle any update (including
+		 * uncommitted updates).  Lookaside eviction can save any
+		 * committed update.  Regular eviction eviction checks that the
```
keithbostic: Typo: "eviction eviction".

keithbostic:
@michaelcahill, I've been chasing the format isolation failures. I don't yet understand the problem, but I have additional information.

First, there's a CONFIG below that fires in a few seconds for me.

Second, the problem is a record with a set of WT_MODIFY records, that's being routed through the update/restore split code (you can turn off lookaside entirely and the problem still happens).

Third, this code change in 504b246 is related to the problem:
```
@@ -1477,24 +1477,6 @@ __split_multi_inmem(
 			break;
 		WT_ILLEGAL_VALUE_ERR(session);
 		}
-
-		/*
-		 * If an update from the list was saved into the page image, it
-		 * must have been globally visible and we don't need two copies.
-		 */
-		if (supd->onpage_upd != NULL) {
-			for (prev_upd = upd; prev_upd != NULL &&
-			    prev_upd->next != supd->onpage_upd;
-			    prev_upd = prev_upd->next)
-				;
-			if (prev_upd != NULL) {
-				WT_ASSERT(session,
-				    prev_upd->next == supd->onpage_upd);
-				__wt_update_obsolete_free(
-				    session, page, prev_upd->next);
-				prev_upd->next = NULL;
-			}
-		}
```
if you put that code back in, the test starts passing again.

Here's the CONFIG:
```
#  RUN PARAMETERS
abort=0
alter=0
auto_throttle=1
backups=0
bitcnt=3
bloom=1
bloom_bit_count=45
bloom_hash_count=13
bloom_oldest=0
cache=2
checkpoints=1
checksum=uncompressed
chunk_size=9
compaction=0
compression=none
data_extend=0
data_source=table
delete_pct=15
dictionary=0
direct_io=0
encryption=none
evict_max=2
file_type=row-store
firstfit=0
huffman_key=0
huffman_value=0
in_memory=0
insert_pct=1
internal_key_truncation=1
internal_page_max=9
isolation=snapshot
key_gap=5
key_max=64
key_min=32
leaf_page_max=9
leak_memory=0
logging=0
logging_archive=1
logging_compression=none
logging_prealloc=1
long_running_txn=0
lsm_worker_threads=3
merge_max=13
mmap=1
modify_pct=20
ops=100000
prefix_compression=1
prefix_compression_min=4
read_pct=32
rebalance=0
repeat_data_pct=68
reverse=0
rows=100000
salvage=0
split_pct=83
statistics=0
statistics_server=0
threads=4
timer=360
transaction_timestamps=0
transaction-frequency=94
value_max=128
verify=0
wiredtiger_config=
write_pct=32
```

src/btree/bt_sync.c
```diff
@@ -221,8 +313,40 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+			 * visit.  We want to avoid this code being too special
+			 * purpose, so try to reuse the ordinary eviction path.
+			 *
+			 * If eviction succeeded, it steps to the next ref, so
+			 * we have to skip the next walk.  If eviction fails,
+			 * remember so we don't retry it.
+			 */
+			if (!WT_PAGE_IS_INTERNAL(page) &&
+			    page->read_gen == WT_READGEN_OLDEST &&
+			    !evict_failed) {
```
keithbostic: One additional test here might be to test `page->modify->rec_result != WT_PM_REC_MULTIBLOCK`, because we can't evict any page that might require a split, there's no point in trying.

src/reconcile/rec_write.c
```diff
@@ -1334,161 +1302,122 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+		for (upd = first_upd; upd->next != NULL; upd = upd->next)
+			;
+		if (__wt_timestamp_cmp(
+		    &r->min_saved_timestamp, &upd->timestamp) > 0)
+			__wt_timestamp_set(
+			    &r->min_saved_timestamp, &upd->timestamp);
+	}
+#else
+	WT_UNUSED(saved);
+#endif
```
keithbostic: @michaelcahill: if we append the original on-page value to the update list, we set the append update item's txnid/timestamp to 0. Is it correct to then set `min_saved_timestamp` to 0 as well?

test/format/wts.c
```diff
@@ -365,14 +365,30 @@ wts_init(void)
 	 * Configure the maximum key/value sizes, but leave it as the default
 	 * if we come up with something crazy.
 	 */
+	/* XXX disable overflow records.  Test-only. */
+#define	__XXX_NO_OVERFLOW 0
```
michaelcahill: @keithbostic you'll want to change this to 1 to repro the overflow issues.  This will all disappear before this branch is ready for final review.

src/cache/cache_las.c
```diff
@@ -280,129 +280,61 @@ __wt_las_cursor_close(
...
-	if (remove_cnt > conn->las_record_cnt)
-		conn->las_record_cnt = 0;
+	if (remove_cnt > S2C(session)->las_record_cnt)
+		S2C(session)->las_record_cnt = 0;
 	else if (remove_cnt > 0)
-		(void)__wt_atomic_sub64(&conn->las_record_cnt, remove_cnt);
-
-	F_CLR(session, WT_SESSION_NO_CACHE);
+		(void)__wt_atomic_sub64(
+		    &S2C(session)->las_record_cnt, remove_cnt);
```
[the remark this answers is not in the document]
michaelcahill: @keithbostic, even though we're not sweeping the lookaside table, multiple threads can be reading different pages with lookaside entries simultaneously.  That means we still need to take care when updating the global record count.

src/include/connection.h
```diff
@@ -367,7 +367,6 @@ struct __wt_connection_impl {
 	WT_SESSION_IMPL *las_session;	/* Lookaside table session */
 	bool		 las_written;	/* Lookaside table has been written */
 
-	WT_ITEM		 las_sweep_key;	/* Sweep server's saved key */
 	uint64_t	 las_record_cnt;/* Count of lookaside records */
```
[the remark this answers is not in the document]
michaelcahill: Agreed, done.

src/btree/bt_read.c
michaelcahill: Sure, done.

src/btree/bt_split.c
```diff
@@ -1474,6 +1476,25 @@ __split_multi_inmem(
...
+				;
+			if (prev_upd != NULL) {
+				WT_ASSERT(session,
+				    prev_upd->next == supd->onpage_upd);
+				__wt_update_obsolete_free(
+				    session, page, prev_upd->next);
+				prev_upd->next = NULL;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
michaelcahill: We've followed up on this separately: it is only okay to strip updates after the on-page update when reconciliation with `WT_REC_VISIBLE_ALL`, which is now required for update/restore eviction.  Otherwise, since modify operations are not idempotent, we can't re-read the update that was chosen as the on-page version.

I'll look into stripping out updates from `__rec_txn_read` again in other cases: I had a more comprehensive version of that code earlier in this branch but dropped it to avoid complicating review / debugging.

src/evict/evict_lru.c
```diff
@@ -75,7 +75,8 @@ __evict_entry_priority(WT_SESSION_IMPL *session, WT_REF *ref)
 		return (WT_READGEN_OLDEST);
 
 	/* Any page from a dead tree is a great choice. */
-	if (F_ISSET(btree->dhandle, WT_DHANDLE_DEAD))
+	if (F_ISSET(btree->dhandle, WT_DHANDLE_DEAD) ||
+	    F_ISSET(btree, WT_BTREE_LOOKASIDE))
```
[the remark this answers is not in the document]
michaelcahill: We want lookaside pages out of cache as soon as possible.  There is no fast path here -- the application doesn't want to access lookaside pages, it wants the cache to not be full so it can make progress with real work.  That change has been split out into a separate ticket.

src/evict/evict_lru.c
michaelcahill: I've added the `goto fast` and tried to clarify the comment.  FWIW, this change was necessary to avoid hangs in testing.

michaelcahill: Also note that we have to mark the page dirty: it isn't just about the eviction checks, we have to call reconciliation on the page for this to be useful.

michaelcahill:
@keithbostic, above you asked:

> Rather than pass the checkpointing flag, would it be reasonable to have a session checkpoint thread flag?

We already have `WT_SESSION_IS_CHECKPOINT`, but that seems less general than passing a flag here -- we may want to multi-thread checkpoint writes at some point, so eviction threads could be getting involved.  I don't feel super strongly: we're both talking about the effect on potential future work...

I'll take a look at what that change would look like.

src/btree/bt_sync.c
michaelcahill: I'm inclined to leave that to eviction to figure out -- we should be able to allow splits in this path, so I'd prefer not to have extra checks preventing them.

michaelcahill:
@keithbostic, I believe this branch is ready for a final review.  Can you please take a look?

keithbostic:
@michaelcahill, final review.

I've pushed some changes in #3691 so we don't have to do another round and we can get this merged today, feel free to discard and/or cherry-pick from that branch without further discussion. In the comments below, I've flagged when there's a commit in #3691 with relevant changes.

* In `__rec_split_write()`, we're still jumping to `skip_write` (which updates the LAS table), when we've matched the previous version of the block based on checksum. There are current 3 jumps to skip_write: in-memory, update/restore and checksum-match. At least two of them don't need to update the LAS table?

* When we update the LAS table in `__rec_update_las()`, the update gets the current session's transaction ID (or an allocated transaction ID if the session doesn't have one), which I think potentially delays our ability to evict those pages. Couldn't we instead use the transaction ID of the update we're storing in the LAS table?

* The sweep server is still opening an LAS cursor but no longer needs it, and the checkpoint server doesn't have an LAS cursor, but probably should (otherwise it single-threads against application threads reading LAS table rows). I've pushed possible changes as part of #3691.

* I was briefly confused by the fact that `mod_replace_las` and `mod_replace_min_timestamp` are the same as `lookaside_pageid` and `lookaside_min_timestamp` but the names are dissimilar. I'd suggest `mod_replace_las_pageid`, `mod_replace_las_min_timestamp`, `las_pageid` and `las_min_timestamp` for consistency, but I wouldn't fight about it.

* The comments in `__sync_evict_page()` are out-of-date. I'm not sure what you were trying to say here, so I didn't attempt to rewrite it.
```
* Hedge our bets: we have a ref on either side of the page we're
* evicting.  Otherwise, the page could split in the meantime, so
* backing up later is not safe.  We allow in-memory splits during
* checkpoints, for example.
...
* Try to evict the original ref: after this, we must update the walk
* point to either the previous or next ref so our caller can continue.
```
* I don't believe the code in `__sync_file()` to loop around `EBUSY` returns from `__sync_dup_walk()` is still needed, since `__sync_dup_walk()` has its own loop.  I've pushed possible changes as part of #3691.

With respect to your comments:

> One additional test here might be to test `page->modify->rec_result != WT_PM_REC_MULTIBLOCK`, because we can't evict any page that might require a split, there's no point in trying.

It's a lot of work, and isn't it a common case for some workloads to see a large page here?

>> Why did we stop checking `cache->evict_aggressive_score`?

> It's checked above via the _HARD flags, before setting lookaside_retryp.

I don't think it's the same thing: the XXX_HARD flags can be set before `__wt_cache_stuck()` will return true, that is, before `cache->evict_aggressive_score == WT_EVICT_SCORE_MAX`.

src/reconcile/rec_write.c
```diff
@@ -433,10 +446,18 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
...
 		ret = __rec_write_check_complete(session, r, lookaside_retryp);
 
+	/*
+	 * XXX special case to fall back to lookaside eviction during
+	 * checkpoints if a page can't be evicted.
+	 */
+	if (ret == EBUSY && lookaside_retryp != NULL &&
+	    !F_ISSET(r, WT_REC_UPDATE_RESTORE) && !r->update_uncommitted)
+		*lookaside_retryp = true;
+
```
keithbostic: Is the `lookaside_retryp` handling split between `__wt_reconcile()` and `__rec_write_check_complete()` because it makes it easier to catch the `EBUSY` case? I'd suggest moving all of the code into `__rec_write_check_complete()`, but I wouldn't argue about it.

src/reconcile/rec_write.c
```diff
@@ -576,17 +600,29 @@ __rec_write_check_complete(
...
+	 * will attempt to allocate a zero-length array).
+	 */
+	if (r->page->type != WT_PAGE_ROW_LEAF) {
+		if (r->entries == 0 && r->supd_next != 0)
+			return (EBUSY);
+		for (multi = r->multi, i = 0; i < r->multi_next; ++multi, ++i)
+			if (multi->disk_image == NULL &&
+			    multi->supd_entries != 0)
+				return (EBUSY);
+	}
```
keithbostic: The test against `r->entries == 0` concerns me. I'm not sure it's wrong, but...

Do you remember the test case that required this change?

My inclination would be to move this check into `__rec_split_write()`, rather than waiting until the final check. (It then becomes clear the check would be against `WT_CHUNK.entries`, too, which resolves my first question.)

src/reconcile/rec_write.c
```diff
@@ -910,49 +928,52 @@ __rec_init(WT_SESSION_IMPL *session,
 #endif
 
 	/*
+	 * When operating on the lookaside table, we can't save updates and
+	 * there is no point trying update/restore eviction: just evict pages
+	 * as soon as we can.
+	 */
+	if (F_ISSET(btree, WT_BTREE_LOOKASIDE))
+		LF_CLR(WT_REC_LOOKASIDE | WT_REC_UPDATE_RESTORE);
```
keithbostic: I don't believe these flags can be set because our callers never set them (and `__evict_review()` explicitly checks). We could use an assert instead, so we don't imply it's a configuration we expect.

src/reconcile/rec_write.c
```diff
@@ -1201,138 +1216,116 @@ err:	__wt_scr_free(session, &tmp);
...
+		return (0);
 
-	skipped = false;
-	update_mem = 0;
-	max_txn = WT_TXN_NONE;
-#ifdef HAVE_TIMESTAMPS
-	__wt_timestamp_set_zero(&max_timestamp);
-#endif
+	for (upd = first_upd; upd != NULL; upd = next_upd) {
+		next_upd = upd->next;
```
keithbostic: I think the `next_upd = upd->next ... upd = next_upd` work is a holdover from a previous version where the list was modified? I don't see any need for it at this point.

src/reconcile/rec_write.c
```diff
@@ -1343,161 +1336,123 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
 	 */
-	if (*updp != NULL &&
-	    (F_ISSET(r, WT_EVICT_LOOKASIDE) ||
-	    (vpack != NULL &&
-	    vpack->ovfl && vpack->raw != WT_CELL_VALUE_OVFL_RM)))
-		WT_RET(__rec_append_orig_value(session, page, *updp, vpack));
+	if (F_ISSET(r, WT_REC_LOOKASIDE) || (*updp != NULL && vpack != NULL &&
+	    vpack->ovfl && vpack->raw != WT_CELL_VALUE_OVFL_RM))
+		WT_RET(
+		    __rec_append_orig_value(session, page, first_upd, vpack));
```
keithbostic: I'm lost, here. You removed the comment about lookaside eviction, but left the test in a different form. I'm specifically confused why `WT_REC_LOOKASIDE` would require overflow item removal if `*updp == NULL`?

src/reconcile/rec_write.c
```diff
@@ -2082,7 +2048,8 @@ __rec_split_page_size_from_pct(
 /*
  * __wt_split_page_size --
  *	Split page size calculation: we don't want to repeatedly split every
- * time a new entry is added, so we split to a smaller-than-maximum page size.
+ *	time a new entry is added, so we split to a smaller-than-maximum page
+ *	size.
```
keithbostic: And people say I'm picky.  :)

michaelcahill:
Thanks for the review @keithbostic.  Here's my todo list:

- [ ] In __rec_split_write(), we're still jumping to skip_write (which updates the LAS table), when we've matched the previous version of the block based on checksum.
- [ ] When we update the LAS table in __rec_update_las(), the update gets the current session's transaction ID (or an allocated transaction ID if the session doesn't have one), which I think potentially delays our ability to evict those pages. Couldn't we instead use the transaction ID of the update we're storing in the LAS table?
- [ ] `mod_replace_las_pageid`, `mod_replace_las_min_timestamp`, `las_pageid` and `las_min_timestamp for consistency`
- [ ] The comments in __sync_evict_page() are out-of-date.
- [ ] Checking `page->modify->rec_result != WT_PM_REC_MULTIBLOCK` before attempting eviction during checkpoints: it's a lot of work, and isn't it a common case for some workloads to see a large page here?
- [ ] Why did we stop checking `__wt_cache_stuck()` before attempting lookaside eviction?
- [ ] Suggest moving all of the `lookaside_retryp` code into __rec_write_check_complete()
- [ ] The test against `r->entries == 0` in `rec_write.c` concerns me...  move this check into `__rec_split_write()`
- [ ] Don't believe `WT_REC_LOOKASIDE | WT_REC_UPDATE_RESTORE` can be set when evicting from the lookaside table... could use an assert instead.
- [ ] `next_upd = upd->next ... upd = next_upd` work is a holdover from a previous version
- [ ] why does `WT_REC_LOOKASIDE` require a stable item in the update list if *updp == NULL?

PR #3646, opened 2017-09-10
WT-3543 Ensure that failed checkpoints cleanup properly

keithbostic:


agorrod:
Thanks @keithbostic I like this change, I'll go ahead and merge.

PR #3648, opened 2017-09-11
WT-3556 remove wtstats support from WiredTiger

keithbostic:
@agorrod, this branch removes wtstats support. I left the `stat_data.py` file in place for now, until we figure out whether or not t2 needs it.


agorrod:
@keithbostic I think you should remove `tools/stat_data.py` as well - I've added a comment to the JIRA ticket explaining why it's OK.

PR #3649, opened 2017-09-11
WT-3574 No need for atomic add.

sueloverso:
@agorrod and @sulabhM please review.  I may be missing some motivation that requires this operation to be atomic.  While the test isn't a performance related test, performing more operations means we test more code when it is killed.

### DIFF
--- a/test/csuite/timestamp_abort/main.c
+++ b/test/csuite/timestamp_abort/main.c
@@ -224,7 +224,7 @@ thread_run(void *arg)
 			 * by another thread.
 			 */
 			testutil_check(pthread_rwlock_rdlock(&commit_ts_lock));
-			stable_ts = __wt_atomic_addv64(&global_ts, 1);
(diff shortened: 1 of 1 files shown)


sueloverso:
Hold off on merging this. @keithbostic has some ideas about removing the need for the rwlock in both `test/format` and `timestamp_abort`.

keithbostic:
I pushed a branch in #3650 that avoids the need for locking in `format` by storing the last-committed timestamp in per-thread memory, and having the thread that updates the oldest timestamp sweep the per-thread memory for the lowest committed value.

We could do something similar in `timestamp_abort`, if people like this approach.

sueloverso:
I had thought about that too.  I'll take a run at the `timestamp_abort` version.

sueloverso:
Closing this in favor of https://github.com/wiredtiger/wiredtiger/pull/3650

PR #3652, opened 2017-09-12
WT-3575 ASCII statistics log printing

keithbostic:


test/format/wts.c
```diff
@@ -224,11 +224,12 @@ wts_open(const char *home, bool set_api, WT_CONNECTION **connp)
 		if (mmrand(NULL, 0, 5) == 1 &&
 		    memcmp(g.uri, "file:", strlen("file:")) == 0)
 			CONFIG_APPEND(p,
-			    ",statistics=(fast)"
-			    ",statistics_log=(wait=5,sources=(\"file:\"))");
+			    ",statistics=(fast),statistics_log="
+			    "(json,on_close,wait=5,sources=(\"file:\"))");
```
sueloverso: Is it intentional to change `test/format`'s use of statistics in this ticket as it is not really related to the problem of the ticket?  Or is this a leftover from your testing?

keithbostic: Yes, that was intentional.

PR #3654, opened 2017-09-12
WT-3430 Add timestamp queue statistics.

sueloverso:
@agorrod Please review.  I added the timestamp queue statistics.  I tested it with the `timestamp_abort` program (yielding additional tickets yesterday).  The statistics indicate that that program can insert at the head of the queue about 23-28% of the time in my couple of runs.

The ticket mentions "how many entries need to be walked to update IDs".  I did not add in a statistic for that because the ordered queues means we only have to look at the head to know our minimum timestamp value.  If you have something else in mind let me know.


sueloverso:
Good catch @sulabhM.  I will fix that and merge tomorrow.

PR #3655, opened 2017-09-12
WT-3581 format: unexpected checkpoint resolution

keithbostic:
@agorrod, this was an error I didn't expect to happen, obviously it does. This is a fix to ensure we don't panic in this case (there's no reason to panic as far as the block manager is concerned). I'm chasing the root cause, as I said, I didn't expect this failure.

Update: the root cause is discussed in the [JIRA ticket](https://jira.mongodb.org/browse/WT-3581?focusedCommentId=1671385&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1671385).

I believe this change is correct and ready for merge.


agorrod:
Thanks @keithbostic lgtm

PR #3656, opened 2017-09-13
WT-3576 convert timestamps to integers before comparing in python test

sulabhM:


sulabhM:
autobuild: test this please

PR #3657, opened 2017-09-13
WT-3561 Remove test/recovery from Make.subdirs.

michaelcahill:


michaelcahill:
@sulabhM, I think this should have happened as part of WT-3561: that's why the original change caused some unexpected build issues.  Can you please review once testing completes?

michaelcahill:
autobuild: retest this please

PR #3658, opened 2017-09-13
WT-3295 Fix LSM merge span calculations

squalfof:
```
chunk->generation > max_gen 
// could be expanded as
chunk->generation - youngest->generation > max_gap
// which is always FALSE, given the predicate above
if (oldest_gen - youngest_gen > max_gap) break;
```
thus this predicate is useless. besides,
```
previous->generation <= max_gen
// could be expanded as
previous->generation - youngest->generation <= max_gap
// which is always TRUE. otherwise it'll jump out in the previous loop
```

Thought what this part actually intended is
```
if (nchunks >= merge_min &&
    chunk->generation - youngest_gen > max_level) {
    break;
}
```
1.  place it near another generation predicate, a little more cohesive;
2. replace youngest->generation with youngest_gen, thought youngest_gen is intended for generation compare; 
3. doubt should 'youngest' be renamed ? it's the chunk at the end, not 'youngest';


wtbuild:

michaelcahill:
autobuild: ok to test

michaelcahill:
autobuild: retest this please

squalfof:
Some Questions related:
```
/* Track chunk generations seen while looking for a merge */
if (chunk->generation < youngest_gen)
    youngest_gen = chunk->generation;
else if (chunk->generation > oldest_gen)
    oldest_gen = chunk->generation;

if (oldest_gen - youngest_gen > max_gap)
    break;

F_SET(chunk, WT_LSM_CHUNK_MERGING);
record_count += chunk->count;
--start_chunk;
```
If oldest_gen & youngest_gen are within range [start_chunk, end_chunk], then should reorder code, make ```F_SET(), record_count, start_chunk``` executed first ? Otherwise, when ```break``` happens, oldest_gen may represent ```chunk->generation```

squalfof:
Question2:
```
/*
 * If the merge would be too big, or we have a full window
 * and we could include an older chunk if the window wasn't
 * full, remove the youngest chunk.
 */
if (chunk_size > lsm_tree->chunk_max ||
    (nchunks == merge_max && start_chunk > 0 &&
     chunk->generation == lsm_tree->chunk[start_chunk - 1]->generation)) {
```
what the last generation comparison does is to make sure window-slide will not break the max_gap & max_level assumption?

squalfof:
@michaelcahill @agorrod could you take a look at the code change ?

src/lsm/lsm_merge.c
```diff
@@ -235,6 +235,14 @@ __lsm_merge_span(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree,
 			break;
 
 		/*
+		 * If we have enough chunks for a merge and the next chunk is
+		 * in too high a generation, stop.
+		 */
+		if (nchunks >= merge_min &&
+		    chunk->generation - youngest_gen > max_gap)
```
agorrod: I don't think it's safe to do this comparison until `youngest_gen` has been updated  - otherwise there is a possibility that `chunk->generation` is smaller than `youngest_gen` and there is an underflow.

src/lsm/lsm_merge.c
```diff
@@ -283,9 +279,9 @@ __lsm_merge_span(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree,
 			WT_ASSERT(session,
 			    F_ISSET(youngest, WT_LSM_CHUNK_MERGING));
 			F_CLR(youngest, WT_LSM_CHUNK_MERGING);
-			record_count -= youngest->count;
-			chunk_size -= youngest->size;
+			/* try again with smaller range */
 			--end_chunk;
+			goto retry_find;
```
agorrod: Using this goto path here makes me nervous - in the other case where we restart the search any altered state has been reset.

squalfof: i don't like such update either. However, ```youngest_gen``` and ```oldest_gen``` would represent obsolete value without go back to the original ```retry_find ``` place. it's not that easy to restore them like what you did on ```record_count ``` and ```chunk_size```.

src/lsm/lsm_merge.c
squalfof: eee yes, an underflow bug would be introduced 👍 
By the point of this predicate it's guaranteed that ```old_gen - yon_gen <= max_gen```, otherwise it couldn't reach here. Therefore, if this predicate hits and jumps out, we'll get an valid start & end pair.
However, if we place this predicate after  ```youngest_gen & oldest_gen``` is updated then break out,
it's possible that ``` old_gen - yon_gen > max_gen ``` and will fall into [code](https://github.com/squalfof/wiredtiger/blob/f3770bf512d40a3071d402c39d0faaac4a0ec1ea/src/lsm/lsm_merge.c#L306), an valid start & end will be ignored.
how about keep this predication here with ?

``` (int32_t)chunk->generation - youngest_gen > (int32_t)max_gap```

Comment (agorrod) [reply src/lsm/lsm_merge.c]: Your suggestion should be correct, but we try to avoid casting in our code base. How about changing it to:

```
chunk->generation > youngest_gen + max_gap
```

File: src/lsm/lsm_merge.c (hunk 2 above)
Comment (agorrod) [reply src/lsm/lsm_merge.c]: I'd probably pull the six lines here:
https://github.com/wiredtiger/wiredtiger/blob/develop/src/lsm/lsm_merge.c#L311
out into a function called `__lsm_merge_clear(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree, u_int start, u_int nchunks)` and call that before the `goto retry_find` as well as in the location below.

My fear is that a bug in this code will leave a chunk with the `WT_LSM_CHUNK_MERGING` flag set, which would lead it to never be merged.

File: src/lsm/lsm_merge.c [hunk 3]
```diff
@@ -255,21 +282,10 @@ __lsm_merge_span(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree,
...
-			max_gen = youngest->generation + max_gap;
-			if (previous->generation <= max_gen &&
-			    chunk->generation > max_gen)
-				break;
-		}
-
 		F_SET(chunk, WT_LSM_CHUNK_MERGING);
 		record_count += chunk->count;
 		--start_chunk;
+		++nchunks;
```
squalfof: Initially, ```nchunks = (end_chunk + 1) - start_chunk```. Here start_chunk is updated, hence nchunks should be updated as well. Later nchunks will be used, it should reflect the most recent data.

squalfof:
@agorrod could you please help to check why there is the **failing check**, seems it's about style ?
Besides, i've updated the code as well.
Thanks.

keithbostic:
autobuild: retest please

keithbostic:
autobuild: retest this please

agorrod:
Thanks @squalfof sorry for the delay here - I'm going to merge this into the other branch you have been working from.

PR #3659, opened 2017-09-15
WT-3223 Progress messages for checkpoint operation.

nehakhatri5:
While checkpointing a table (tree), log verbose message every 20 seconds
indicating the amount of work done in that duration. It was also observed that
the step of deleting the checkpoints, that are no longer required, is consuming
considerable amount of time while checkpointing a tree. Hence added a log
message at this step too.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -633,7 +633,7 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 						conn->ckpt_tree_verb_time);
 			if (time_diff > 15)
 				__wt_verbose(session, WT_VERB_CHECKPOINT,
-			            "delete-checkpoint took : %" PRIu64 "sec",
+				    "delete-checkpoint took : %" PRIu64 "sec",
 				    time_diff);
 			conn->ckpt_tree_verb_time = current_time;
 		}
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -389,14 +389,18 @@ static int
 __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 {
 	WT_BLOCK_CKPT *a, *b, *ci;
+	WT_CONNECTION_IMPL *conn;
 	WT_CKPT *ckpt, *next_ckpt;
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	uint64_t ckpt_size;
+	uint64_t time_diff;
+	struct timespec current_time;
 	bool deleting, fatal, locked;
 
 	ci = &block->live;
 	fatal = locked = false;
+	conn = S2C(session);
 
 #ifdef HAVE_DIAGNOSTIC
 	WT_RET(__ckpt_verify(session, ckptbase));
@@ -622,6 +626,19 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 		 * the live system's checkpoint available list.
 		 */
 		WT_ERR(__wt_block_extlist_overlap(session, block, b));
+
+#ifdef HAVE_VERBOSE
+		if (WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT)) {
+			__wt_epoch(session, &current_time);
+			time_diff = WT_TIMEDIFF_SEC(current_time,
+						conn->ckpt_tree_verb_time);
+			if (time_diff > 15)
+				__wt_verbose(session, WT_VERB_CHECKPOINT,
+			            "delete-checkpoint took : %" PRIu64 "sec",
+				    time_diff);
+			conn->ckpt_tree_verb_time = current_time;
+		}
+#endif
 
 		/*
 		 * If we're updating the live system's information, we're done.
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -627,7 +627,6 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 		 */
 		WT_ERR(__wt_block_extlist_overlap(session, block, b));
 
-#ifdef HAVE_VERBOSE
 		if (WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT)) {
 			__wt_epoch(session, &current_time);
 			time_diff = WT_TIMEDIFF_SEC(current_time,
@@ -638,7 +637,6 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 				    time_diff);
 			conn->ckpt_tree_verb_time = current_time;
 		}
-#endif
 
 		/*
 		 * If we're updating the live system's information, we're done.
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -388,14 +388,14 @@ __ckpt_verify(WT_SESSION_IMPL *session, WT_CKPT *ckptbase)
 static int
 __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 {
+	struct timespec current_time;
 	WT_BLOCK_CKPT *a, *b, *ci;
-	WT_CONNECTION_IMPL *conn;
 	WT_CKPT *ckpt, *next_ckpt;
+	WT_CONNECTION_IMPL *conn;
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	uint64_t ckpt_size;
 	uint64_t time_diff;
-	struct timespec current_time;
 	bool deleting, fatal, locked;
 
 	ci = &block->live;
@@ -549,6 +549,9 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 		    !F_ISSET(ckpt, WT_CKPT_DELETE))
 			continue;
 
+		/* Set the checkpoint deletion start time. */
+		__wt_epoch(session, &conn->ckpt_tree_verb_time);
+
 #ifdef HAVE_VERBOSE
 		if (WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT)) {
 			if (tmp == NULL)
@@ -630,7 +633,7 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 		if (WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT)) {
 			__wt_epoch(session, &current_time);
 			time_diff = WT_TIMEDIFF_SEC(current_time,
-						conn->ckpt_tree_verb_time);
+			    conn->ckpt_tree_verb_time);
 			if (time_diff > 15)
 				__wt_verbose(session, WT_VERB_CHECKPOINT,
 				    "delete-checkpoint took : %" PRIu64 "sec",
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -75,8 +75,7 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 	WT_PAGE *page;
 	WT_REF *walk;
 	WT_TXN *txn;
-	uint64_t int_bytes_delta, int_page_delta, leaf_bytes_delta, leaf_delta;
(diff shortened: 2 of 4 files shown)


nehakhatri5:
Not ready fro review.

sueloverso:
I took a quick look at this.  If this is just extra information for users interested in checkpoints this is a fine direction.  Do we expect that MongoDB wants to have this information all the time in general?  If so, currently the only verbose setting turned on for MongoDB is `recovery_progress`.  If MongoDB wants this, then `recovery_progress` should be renamed to something more general like `progress` and these messages should be put under that heading.

nehakhatri5:
Form the discussions that I had with Alex, we want to add these messages only for checkpoints. If it is required for MongoDB logs, then I can further change the code.

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
 			}
+			++page_counter;
+
 			WT_ERR(__wt_reconcile(
 			    session, walk, NULL, WT_CHECKPOINTING, NULL));
+
+			if (timer && page_counter > 5000) {
+				__wt_epoch(session, &cur_time);
+				time_diff = WT_TIMEDIFF_SEC(cur_time,
+						conn->ckpt_tree_verb_time);
```
sulabhM: same as before, indent 4 spaces

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+
+				if (time_diff > 20) {
+
```
sulabhM: Done need new line here.

src/btree/bt_sync.c
```diff
@@ -68,15 +68,18 @@ __sync_checkpoint_can_skip(WT_SESSION_IMPL *session, WT_PAGE *page)
...
 {
-	struct timespec end, start;
+	struct timespec end, start, cur_time;
 	WT_BTREE *btree;
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
 	WT_PAGE *page;
 	WT_REF *walk;
 	WT_TXN *txn;
+	uint64_t int_bytes_svpt, int_page_svpt, leaf_bytes_svpt, leaf_svpt;
```
sulabhM: maybe variable names should be renamed. I am not getting a hint on what this variable is by "svpt".
leaf_svpt should be leaf_page_svpt to be consistent

src/btree/bt_sync.c
```diff
@@ -89,6 +92,9 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 
 	internal_bytes = leaf_bytes = 0;
 	internal_pages = leaf_pages = 0;
+	int_bytes_svpt = leaf_bytes_svpt = 0;
```
sulabhM: convention is to initialize variables closest to the first use, maybe these can move just above the for loop where these variables get used.

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 				leaf_bytes += page->memory_footprint;
 				++leaf_pages;
 			}
+			++page_counter;
+
 			WT_ERR(__wt_reconcile(
```
sulabhM: Please note: __wt_reconcile will end up calling __ckpt_process which will reset conn->ckpt_tree_verb_time

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+					    "%" PRIu64 " internal pages (%"
+					    PRIu64 "B)", time_diff, leaf_svpt,
+					    leaf_bytes_svpt, int_page_svpt,
+					    int_bytes_svpt);
+
+					int_bytes_svpt = internal_bytes;
+					int_page_svpt = internal_pages;
+					leaf_bytes_svpt = leaf_bytes;
+					leaf_svpt = leaf_pages;
+					conn->ckpt_tree_verb_time = cur_time;
```
sulabhM: I don't like the idea of having ckpt_tree_verb_time initialized somewhere else. I understand what you are trying to do, but makes me uneasy and I am unable to figure out what the time diff really is: 

__checkpoint_tree() initialises ckpt_tree_verb_time. Then __sync_file() and __ckpt_process get time locally and subtract ckpt_tree_verb_time to obtain time elapsed since last reset of ckpt_tree_verb_time.

__wt_reconcile in __sync_file ends up calling __ckpt_process which will reset conn->ckpt_tree_verb_time if checkpoint delete is involved? So does that mean "time_diff > 20" is from the time checkpoint was deleted ?

Why not have start_time and end_time locally around delete checkpoint and the for loop in sync_file that iterates through pages are writes them to disk. It will be cleaner to understand. You have looked at this code more than I have, probably I am missing something ?

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 				leaf_bytes += page->memory_footprint;
 				++leaf_pages;
 			}
+			++page_counter;
```
sulabhM: The other page counters are uint64, is it possible that we can overflow this one being uint32. It doesn't get reset to 0 after every 5000, if timer is false.

src/block/block_ckpt.c
```diff
@@ -389,14 +389,18 @@ static int
 __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 {
 	WT_BLOCK_CKPT *a, *b, *ci;
+	WT_CONNECTION_IMPL *conn;
 	WT_CKPT *ckpt, *next_ckpt;
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	uint64_t ckpt_size;
+	uint64_t time_diff;
```
bvpvamsikrishna: This declaration can be along with the previous one.

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+					    int_bytes_svpt);
+
+					int_bytes_svpt = internal_bytes;
+					int_page_svpt = internal_pages;
+					leaf_bytes_svpt = leaf_bytes;
+					leaf_svpt = leaf_pages;
+					conn->ckpt_tree_verb_time = cur_time;
+				}
+
+				page_counter = 0;
```
sulabhM: I see, this makes sure that page counter gets reset to 0 when it reaches 5000. When timer is false we do not reset it and can possibly overflow, though we wont be using it. Why not increment page counter only when timer is true. Need to put a comment stating what page_counter does.

src/block/block_ckpt.c
```diff
@@ -623,6 +627,17 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 		 */
 		WT_ERR(__wt_block_extlist_overlap(session, block, b));
 
+		if (WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT)) {
+			__wt_epoch(session, &current_time);
+			time_diff = WT_TIMEDIFF_SEC(current_time,
+						conn->ckpt_tree_verb_time);
```
bvpvamsikrishna: This indentation is only 4 more than previous one.

src/block/block_ckpt.c
```diff
@@ -623,6 +627,17 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
...
+			if (time_diff > 15)
```
bvpvamsikrishna: instead of hardcoding 15, we can use macro so that later change will be easier.

src/btree/bt_sync.c
```diff
@@ -68,15 +68,18 @@ __sync_checkpoint_can_skip(WT_SESSION_IMPL *session, WT_PAGE *page)
 static int
 __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 {
-	struct timespec end, start;
+	struct timespec end, start, cur_time;
```
bvpvamsikrishna: variable names to maintain the alphabetical order.

src/btree/bt_sync.c
```diff
@@ -68,15 +68,18 @@ __sync_checkpoint_can_skip(WT_SESSION_IMPL *session, WT_PAGE *page)
...
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
 	WT_PAGE *page;
 	WT_REF *walk;
 	WT_TXN *txn;
+	uint64_t int_bytes_svpt, int_page_svpt, leaf_bytes_svpt, leaf_svpt;
 	uint64_t internal_bytes, internal_pages, leaf_bytes, leaf_pages;
 	uint64_t oldest_id, saved_pinned_id;
+	uint32_t page_counter;
+	uint64_t time_diff;
```
bvpvamsikrishna: we can add this new variable to existing one, instead of declaring on a new line.

src/btree/bt_sync.c
bvpvamsikrishna: instead of leaf_svpt, we can have the name as leaf_pages_svft to increase readability.

src/btree/bt_sync.c
bvpvamsikrishna: indentation is not proper.

src/btree/bt_sync.c
```diff
@@ -221,8 +227,46 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+
 			WT_ERR(__wt_reconcile(
 			    session, walk, NULL, WT_CHECKPOINTING, NULL));
+
+			if (timer && page_counter > 5000) {
+				__wt_epoch(session, &cur_time);
+				time_diff = WT_TIMEDIFF_SEC(cur_time,
+						conn->ckpt_tree_verb_time);
+
+				if (time_diff > 20) {
```
bvpvamsikrishna: use a macro instead of hardcoding a number

src/btree/bt_sync.c
nehakhatri5: The idea here is to start a timer at the time when we start checkpointing a tree(file), and then  log progress message for the steps that are taking considerable time , while checkpointing that file. Hence the timer is started as we enter the function __checkpoint_tree. 
In the course of checkpointing the tree, we also delete the checkpoints that are no longer required. It was noticed in certain instances that deleting checkpoint was taking considerable time while checkpointing a tree. So it is worth logging a message at this step and then log another message for sync operation it that is taking more than 20 sec after this step.

src/btree/bt_sync.c
nehakhatri5: 'svpt' is for savepoint, the value saved in a svpt gets used in the next log message when the stats has to be displayed in that message.

src/block/block_ckpt.c
```diff
@@ -389,14 +389,18 @@ static int
 __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
 {
 	WT_BLOCK_CKPT *a, *b, *ci;
+	WT_CONNECTION_IMPL *conn;
 	WT_CKPT *ckpt, *next_ckpt;
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	uint64_t ckpt_size;
+	uint64_t time_diff;
+	struct timespec current_time;
```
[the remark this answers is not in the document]
sueloverso: The `struct timespec` should be listed before any of the `WT_*` structures.  See `__wt_bt_write` in `bt_io.c:174` or `__log_server` in `conn_log.c:868` for examples.

src/btree/bt_sync.c
sueloverso: I had no idea what `svpt` meant either.  My suggestion would be `_save`.

nehakhatri5: I think I used this term because of my RDBMS background. It's used there. I will change it.

src/block/block_ckpt.c
nehakhatri5: The two are not associated , hence kept seperate.

src/btree/bt_sync.c
nehakhatri5: Moved the page counter increment statement under if(timer) condition.

src/btree/bt_sync.c
nehakhatri5: Since the number is specific to this specific logic, not created a macro for it.

src/block/block_ckpt.c
sueloverso: Our coding standard puts them together, being the same type, on the declaration line anyway.

nehakhatri5:
This will be handled differently in another pull request.

PR #3660, opened 2017-09-15
WT-3457 skip scrubbing if very little data to scrub

sulabhM:


sulabhM:
autobuild: retest this please

src/txn/txn_ckpt.c
```diff
@@ -405,6 +405,20 @@ __checkpoint_reduce_dirty_cache(WT_SESSION_IMPL *session)
...
 	if (cache_size < 10 * WT_MEGABYTE)
 		return;
+
+	/*
+	 * Skip scrubbing if it won't perform at-least some minimum amount of
+	 * work. Scrubbing is supposed to bring down the dirty data to eviction
+	 * checkpoint target before the actual checkpoint starts. Do not perform
+	 * scrubbing if the dirty data to scrub is less than a pre-configured
+	 * size. This size is to an extent based on the configured cache size
+	 * without being too large or too small for large cache sizes.
```
sueloverso: This comment does not help me understand the calculation below.  There are a lot of magic numbers in there.  Please comment on what the calculations mean and how/why it was chosen.  Is the 0.2 for 20% of the cache_size related to the default eviction_dirty_trigger setting?  Should it use that instead?

src/txn/txn_ckpt.c
```diff
@@ -405,6 +405,20 @@ __checkpoint_reduce_dirty_cache(WT_SESSION_IMPL *session)
...
+	 */
+	if ((__wt_cache_dirty_leaf_inuse(cache) +
+	    WT_MIN((0.2 * conn->cache_size) / 100, 500 * WT_MEGABYTE)) <
+	    ((cache->eviction_checkpoint_target * conn->cache_size) / 100))
```
sueloverso: In parsing this, wouldn't we skip scrubbing in this case anyway?  The second half of the computation is the (default) 5% of cache for the checkpoint target.  So if the dirty cache is already less than the eviction_checkpoint_target we would (today) break out of the for loop a few lines below.

I thought the intent was to skip scrubbing if we're close to the 5%.  So shouldn't the calculation be something like:
if (wt_cache_dirty_leaf_inuse(cache) <
((eviction_checkpoint_target * cache_size)/100 + <calculation of meaning of "close to scrub target" >)

sulabhM: You are right. I fixed the logic

src/txn/txn_ckpt.c
sulabhM: The numbers are arbitrary to some extent. I talked to the support folks to have an idea of what the installation sizes, IO capacity people usually run with (onAWS, 100+ GB cache is common, 50 MBPS would be median IO bandwidth and tops at 320 MBPS, is what I heard). Then discussed with @agorrod and we decided to have a minimum of 500 MB or 0.2% of cache size as the delta that scrub should clean. There is no concrete reason behind that value though. I updated comment to reflect this. I am not too happy with the comment and open to suggestions on improving it.

sulabhM:
@sueloverso , please have another look.

src/txn/txn_ckpt.c
```diff
@@ -405,6 +405,24 @@ __checkpoint_reduce_dirty_cache(WT_SESSION_IMPL *session)
...
+	/*
+	 * Skip scrubbing if it won't perform at-least some minimum amount of
+	 * work. Scrubbing is supposed to bring down the dirty data to eviction
+	 * checkpoint target before the actual checkpoint starts. Do not perform
+	 * scrubbing if the dirty data to scrub is less than a pre-configured
+	 * size. This size is to an extent based on the configured cache size
+	 * without being too large or too small for large cache sizes. For the
+	 * values chosen, for instance, 100 GB cache will require at-least
+	 * 200 MB of dirty data above eviction checkpoint target, which should
+	 * equate to a scrub phase a few seconds long. Said that, the value of
```
sueloverso: "Said that," should be "That said,".   But thank you for adding the clarifying sentences.

src/txn/txn_ckpt.c
```diff
@@ -405,6 +405,24 @@ __checkpoint_reduce_dirty_cache(WT_SESSION_IMPL *session)
...
+	 * 0.2% and 500 MB are still somewhat arbitrary.
+	 */
+	if (__wt_cache_dirty_leaf_inuse(cache) <
+	    ((cache->eviction_checkpoint_target * conn->cache_size) / 100) +
+	    WT_MIN((0.2 * conn->cache_size) / 100, 500 * WT_MEGABYTE))
```
sueloverso: This makes more sense now.  I think it would be a bit more readable if it was something like this:
scrub_min = WT_MIN(...);
if (dirty < ((checkpoint+target * cache_size) / 100) + scrub_min)

It makes it easier to digest and see where we may want to tweak it in the future.

sulabhM:
@sueloverso thanks, I will merge.

PR #3666, opened 2017-09-18
WT-3590 close error changes

sueloverso:
@michaelcahill I separated the commits because the more I think about it, the more it seems like just checking `ret` before writing the clean close checkpoint record will solve the problem.  I think it is a change we should take even if you still want the checkpoint.  Hopefully that is the only change needed.  I separately committed adding in the checkpoint call.  But if we don't need that, it'd be great!

### DIFF
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1116,6 +1116,12 @@ err:	/*
 		}
 	}
 
+	if (ret != 0) {
+		__wt_err(session, ret,
+		    "failure during close, disabling further writes");
+		F_SET(conn, WT_CONN_PANIC);
+	}
+
 	WT_TRET(__wt_connection_close(conn));
 
 	/* We no longer have a session, don't try to update it. */
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1044,6 +1044,7 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
 	WT_SESSION *wt_session;
 	WT_SESSION_IMPL *s, *session;
 	uint32_t i;
+	const char *checkpoint_cfg[] = { NULL, NULL, NULL };
 
 	conn = (WT_CONNECTION_IMPL *)wt_conn;
 
@@ -1083,6 +1084,27 @@ err:	/*
 				    s->event_handler, wt_session, NULL));
 			WT_TRET(wt_session->close(wt_session, config));
 		}
+
+	/*
+	 * Perform a system-wide checkpoint so that all tables are consistent
+	 * with each other.  All transactions are resolved but ignore
+	 * timestamps to make sure all data gets to disk.  Do this before
+	 * shutting down all the subsystems.  We have shut down all user
+	 * sessions, but send in true for waiting for internal races.
+	 */
+	if (!F_ISSET(conn, WT_CONN_IN_MEMORY | WT_CONN_READONLY)) {
+		s = NULL;
+		WT_TRET(__wt_open_internal_session(
+		    conn, "close_ckpt", true, 0, &s));
+		if (s != NULL) {
+			checkpoint_cfg[0] =
+			    WT_CONFIG_BASE(session, WT_SESSION_checkpoint);
+			checkpoint_cfg[1] = "force=1,use_timestamp=false";
+			wt_session = &s->iface;
+			WT_TRET(__wt_txn_checkpoint(s, checkpoint_cfg, true));
+			WT_TRET(wt_session->close(wt_session, config));
+		}
+	}
 
 	WT_TRET(__wt_connection_close(conn));
 
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1044,7 +1044,6 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
 	WT_SESSION *wt_session;
 	WT_SESSION_IMPL *s, *session;
 	uint32_t i;
-	const char *checkpoint_cfg[] = { NULL, NULL, NULL };
 
 	conn = (WT_CONNECTION_IMPL *)wt_conn;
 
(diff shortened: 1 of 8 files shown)


sueloverso:
The PR tester failure is because, for `test_readonly01`, the checkpoint log records are now:
{noformat}
LSN X: checkpoint start
LSN X+1: commit modifying metadata
LSN X+2: checkpoint record
LSN X+3: checkpoint record (from end of close)
{noformat}
The LSN that get written into the turtle file is X.  Then it finds a commit record after the checkpoint LSN and determines recovery is needed.  That determination fails for a read-only database.  This only fails on Linux, the test does not fail on my Mac.  On there I see:
{noformat}
LSN X: checkpoint start
LSN X+1: commit modifying metadata
LSN X+2: checkpoint record
LSN X+3: file_sync metadata start
LSN X+4: file_sync metadata end
LSN X+5: checkpoint record from end of close
{noformat}
On my Mac the LSN that ends up in the turtle file is X+3.  The `dhandle_discard` on the metadata file on the Mac results in the sync records and does not on Linux.  I'm still investigating why.

If avoiding that checkpoint and the change to `conn_open.c` is sufficient, it avoids this increasingly complex change to `conn_api.c`

michaelcahill:
@sueloverso, I've pushed some more changes in here -- I've split them into commits so hopefully it's easy to throw away anything we don't want.

First up, I added a Linux-only test, `test_close_writefail.py` that simulates write failure by reaching underneath WiredTiger and closing a file descriptor before calling connection close.  That fails on `develop` -- the files are out of sync after reopening.  You should double-check the details -- I think it is going on to update the metadata with a checkpoint LSN after the failure, so recovery skips log records that should be applied.

I made one of the changes we discuss, marking the metadata dirty during successful clean shutdown so the existing recovery code correctly detects when recovery is required.

Lastly, I added a "soft panic" during close if there is an error return after the checkpoint completes.  That guarantees nothing else will be written.  This change involved some plumbing, which I'll understand if you don't like but the test case fails on the branch without this change.  With all of the changes, the test completes as expected (the connection close fails, but the trees are consistent when it is reopened).

Please take a look, and particularly check that the test failure on `develop` makes sense.  It is making a backup before recovery runs, so hopefully it will be easy to see what's going on.  Thanks!

sueloverso:
@michaelcahill Thanks for that amazing test!  So compact yet so powerful.  I did modify it slightly - I renamed it into the `test_bug*` category of tests for one-off.  I added a bunch of comments to it also and used `copy_wiredtiger_home` for making the backup.

I investigated the failure on develop.  The key piece, you're correct, is that closing the metadata file results in writing to the turtle file, which then causes us to skip recovery.  So your changes to `conn_dhandle.c` for avoiding that are the crux for not getting the `file_sync` records into the log and then that LSN into the turtle file.

I have a couple comments in specific areas for you to think about but then this is ready for merge.

src/conn/conn_api.c
```diff
@@ -1084,6 +1084,44 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
...
 
+	/*
+	 * Perform a system-wide checkpoint so that all tables are consistent
+	 * with each other.  All transactions are resolved but ignore
+	 * timestamps to make sure all data gets to disk.  Do this before
+	 * shutting down all the subsystems.  We have shut down all user
+	 * sessions, but send in true for waiting for internal races.
+	 */
+	if (!F_ISSET(conn, WT_CONN_IN_MEMORY | WT_CONN_READONLY) &&
+	    conn->ever_modified) {
```
sueloverso: While I can reason about `ever_modified` and not wanting to mark the metadata dirty if nothing ever changed, if I remove that condition, I can successfully run `make check` and the entire Python test suite.  Is it just a minor optimization or is there some other failure you're thinking about here?

src/conn/conn_dhandle.c
```diff
@@ -223,15 +223,14 @@ __wt_conn_dhandle_close(
...
-		 * invalid if the mapping is closed.)
+		 * it's not already marked dead, and it's not a memory-mapped
+		 * tree. (We can't mark memory-mapped tree handles dead because
+		 * we close the underlying file handle to allow the file to be
+		 * removed and memory-mapped trees contain pointers into memory
+		 * that become invalid if the mapping is closed.)
 		 */
 		bm = btree->bm;
-		if (!discard && mark_dead && !final &&
+		if (!discard && mark_dead &&
```
sueloverso: This is the key piece for avoiding putting the `file_sync` records into the log, and the LSN into the turtle file and therefore, allowing recovery to run.

src/conn/conn_open.c
```diff
@@ -127,7 +127,7 @@ __wt_connection_close(WT_CONNECTION_IMPL *conn)
 	 * conditional because we allocate the log path so that printlog can
 	 * run without running logging or recovery.
 	 */
-	if (FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED) &&
+	if (ret == 0 && FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED) &&
```
sueloverso: While this makes intuitive sense, it actually doesn't help the situation at hand.  On develop with this change and running the test, yes we avoid writing that final checkpoint record, but recovery still doesn't run because the true determination is that the close of the metadata handle wrote the `file_sync` records already, and updated the LSN in the turtle file.  So skipping writing the final checkpoint record here doesn't actually matter.  See my `conn_dhandle.c` comment.  Should we keep this change?  It doesn't feel wrong, it just don't solve the problem necessarily either.

test/suite/test_bug018.py
```diff
@@ -74,11 +84,12 @@ def test_close_writefail(self):
 
         # Make a backup for forensics in case something goes wrong.
         backup_dir = 'BACKUP'
-        shutil.rmtree(backup_dir, ignore_errors=True)
-        shutil.copytree('.', backup_dir, lambda src, names: (n for n in names if n != backup_dir))
+        copy_wiredtiger_home('.', backup_dir, True)
```
michaelcahill: Thanks for this one @sueloverso, I knew `wttest.backup` was wrong because it uses a backup cursor, but forgot about `helper`.

src/conn/conn_open.c
michaelcahill: @sueloverso, as discussed, I'm inclined to keep this one.  It doesn't make sense to write the log record if there has been a failure...

src/conn/conn_api.c
michaelcahill: I had the idea that we didn't write new files if the application never did any updates -- clearly I was mistaken, so I'll pull that code out.

michaelcahill:
Thanks @sueloverso, I've made your suggested change and I'm happy with this now.  I'll merge.

PR #3667, opened 2017-09-20
WT-3593 Add an API to enforce consitent use of timestamps

sueloverso:
Not ready for review.


sueloverso:
@agorrod this is ready for review.  In particular, this API is fully available, not dependent on `HAVE_DIAGNOSTICS`.  I can do that if you feel it should be.  But since it is something written into the metadata for a table it was not a clean path.

sueloverso:
@dgottlieb here are the changes.  See if they meet what you had in mind.

dgottlieb:
I played with this a bit and it looks good. Just to highlight one of the useful choices you made, as I suspect it wasn't obvious, was the choice to return `EINVAL` is very useful. Specifically, returning `WT_ROLLBACK` would be problematic. With this the MongoDB `invariantWTOK` checks will crash right when this happens (directly after WT itself logs the error message). A `WT_ROLLBACK` gets treated special to work with our retry-loops and I can see the log lines being missed in that scenario.

sueloverso:
@dgottlieb Thanks!  Although I didn't think about `EINVAL` in the way you did, detecting this certainly seemed like an error.  And the point was to detect errors.  I did consider calling `wt_panic` but it seemed an error was sufficient.  Retrying definitely didn't seem like it would help.  I'm glad we're in sync on that.

src/include/api.h
```diff
@@ -39,9 +39,11 @@
 } while (0)
 
 /* An API call wrapped in a transaction if necessary. */
+#ifdef HAVE_TIMESTAMPS
```
agorrod: I prefer not to duplicate the `TXN_API_CALL` macros, would you consider instead adding:
```
#define WT_TXN_TIMESTAMP_FLAG_CHECK(s)                                 \
       __wt_txn_timestamp_flags((s));
#else
#define WT_TXN_TIMESTAMP_FLAG_CHECK(s)
#endif
```
And adding a line to the existing macros?

src/include/btree.h
```diff
@@ -97,6 +97,12 @@ struct __wt_btree {
 	uint64_t maxmempage;		/* In-memory page max size */
 	uint64_t splitmempage;		/* In-memory split trigger size */
 
+#define	BTREE_DBG_COMMITTS_ALWAYS	0x0001
+#define	BTREE_DBG_COMMITTS_NEVER	0x0002
+#define	BTREE_DBG_READTS_ALWAYS		0x0004
+#define	BTREE_DBG_READTS_NEVER		0x0008
```
agorrod: I think these names could be better:
 * The should have a `WT_` prefix
 * Could you use `DEBUG` instead of `DBG`? I find that easier to read.
 * Having the `TS` without a separator made me think you'd misspelt `commits` at first

I know there is a risk of them becoming very long, so don't mind if you'd like to push back.

I'm also a fan of having the default represented with an explicit flag, which would mean adding `WT_BTREE_DEBUG_COMMIT_TS_NONE` and `WT_BTREE_DEBUG_READ_TS_NONE` (or whichever form you end up choosing for the common prefix). I don't think that adds to the overhead - we are adding a flag check to common paths regardless.

src/include/btree.h
```diff
@@ -97,6 +97,12 @@ struct __wt_btree {
...
+	uint32_t debug;			/* Debugging information */
```
agorrod: Could you call this `debug_flags`?

src/include/api.h
sueloverso: Fixed

src/include/btree.h
sueloverso: I did not add in the `_NONE` settings because there is no place to use them and having and setting flags that are never used in code is confusing.  And adding code to check it when that isn't necessary also didn't appeal to me.  I took all the other suggestions.

There are a few other non-flag locations that don't prepend the structure name, such as for direct_io.  So to prevent unpleasantly long names, I went with `WT_DEBUG_COMMIT_TS_ALWAYS`, etc.

src/include/btree.h
sueloverso: Fixed.

sueloverso:
I am happy to make them undocumented.  I'm also happy with any of the three proposed names; all are equally intuitive to me.  I'll pick one of yours if they're ones you prefer.

sueloverso:
These changes are ready.  They are now undocumented and I chose `assert`.

michaelcahill:

PR #3668, opened 2017-09-21
WT-3599 reconciliation calculates block matching checksums too frequently.

keithbostic:
We can ignore this branch until WT-3435/#3644 is merged, it should be an easy fix-up afterward.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -220,7 +220,14 @@ typedef struct {
 	 * There's some trickiness here, see the code for comments on how
 	 * these fields work.
 	 */
-	bool	 cell_zero;		/* Row-store internal page 0th key */
+	bool	cell_zero;		/* Row-store internal page 0th key */
+
+	/*
+	 * We calculate checksums to find previously written identical blocks,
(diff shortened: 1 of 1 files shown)


src/reconcile/rec_write.c
```diff
@@ -3345,6 +3354,91 @@ __rec_split_write_header(WT_SESSION_IMPL *session,
...
 /*
+ * __rec_split_write_reuse --
+ *	Check if a previously written block can be reused.
+ */
+static bool
+__rec_split_write_reuse(WT_SESSION_IMPL *session,
+    WT_RECONCILE *r, WT_MULTI *multi, WT_ITEM *image, bool last_block)
+{
+	WT_PAGE_MODIFY *mod;
+	WT_MULTI *multi_match;
```
sulabhM: in alphabetic order, WT_MULTI before WT_PAGE_MODIFY

keithbostic: Thank you @sulabhM, fixed in 06634ac.

michaelcahill:
Thanks @keithbostic, this lgtm -- it's clearer than before and should avoid some calls to `__wt_checksum`.  I'll merge.

PR #3669, opened 2017-09-22
WT-3453 Enhance lookaside table coverage

bvpvamsikrishna:
Added scenarios for old reader and old timestamp
as well as to check the values after recovery
This branch will pass only after wt-3435 is merged.


agorrod:
@bvpvamsikrishna I'm surprised that this branch passed testing - I thought it relied on changes coming in WT-3435 - could you please confirm that the test is working as expected?

michaelcahill:
@agorrod, note that `test_las.py` is a "long" test, i.e., only runs with `run.py --long ...`

A consequence is that it isn't tested by the pull request tester jobs.

michaelcahill:
Thanks @bvpvamsikrishna, lgtm.  I've re-tested with the current tip of develop.  I'll merge.

PR #3670, opened 2017-09-22
WT-3601 Coverity 1381365: bt_split.c:__split_ref_move() resource leak

keithbostic:


michaelcahill:
Thanks @keithbostic, I'll take this one since @nehakhatri5 is in transit.

lgtm, I'll merge.

PR #3671, opened 2017-09-23
WT-3435 Fix lookaside eviction with timestamps.

keithbostic:
@michaelcahill, here's a possible change for checkpoint eviction.

Instead of trying to maintain our position in the tree while evicting, we accumulate pages we'd like to evict, and once we step past an internal page, we process the list and evict the pages we've accumulated. It's obviously not as efficient with respect to the cache, but it avoids trying to maintain our position in the tree.

### DIFF
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -58,91 +58,9 @@ __sync_checkpoint_can_skip(WT_SESSION_IMPL *session, WT_PAGE *page)
 		    i = 0; i < mod->mod_multi_entries; ++multi, ++i)
 			if (multi->addr.addr == NULL)
 				return (false);
+
+	WT_ASSERT(session, mod->mod_replace.addr != NULL);
 	return (true);
-}
-
-/*
- * __sync_dup_walk --
- *	Duplicate a tree walk point.
- */
-static inline int
-__sync_dup_walk(WT_SESSION_IMPL *session, WT_REF *walk)
-{
-	bool busy;
-
-	/*
-	 * We already have a hazard pointer, we should generally be able to get
-	 * another one.  That said, we could race with some thread attempting
-	 * to get exclusive access: just continue on to the ordinary write path
-	 * in that case.
-	 */
-#ifdef HAVE_DIAGNOSTIC
-	WT_RET(__wt_hazard_set(session, walk, &busy, __func__, __LINE__));
-#else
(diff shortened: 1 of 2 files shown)


michaelcahill:
@keithbostic, thanks -- I see the appeal.

That said, will this approach cope with the case where there is more data in lookaside than fits in cache, and we're checkpointing with an old timestamp?  If there is replication lag in MongoDB (i.e., some nodes in a replicaset can't keep up), then that will be a common case.

What I think will happen with this approach is that the checkpoint will leave pages dirty, then eviction will skip them because we don't allow dirrty eviction during a checkpoint.  Can you see some way around that?

keithbostic:
> That said, will this approach cope with the case where there is more data in lookaside than fits in cache, and we're checkpointing with an old timestamp?

I believe this approach evicts the exact same pages as the previous approach, except it evicts the set of the pages found so far after each internal node, rather than each page as it's encountered.

The technique is the same and the evictor is the same, I don't see why there would be a difference, aside from the bundling through each internal node.

keithbostic:
@michaelcahill, one additional thing that's maybe worth mentioning.

In the previous approach, we can't evict any page that results in a split because we can't modify its parent. (This change explicitly tests for `page->modify->rec_result != WT_PM_REC_MULTIBLOCK` before attempting eviction, for that reason.)

However, in this approach, checkpoint should be able to evict pages that split into their parent, that's why I mentioned drilling a hole for the `checkpointing` flag into the split code. We can't climb the tree I don't think, but we should be able to evict pages or parts of pages that require a split.

michaelcahill:
As discussed, the thing that makes me doubt this approach is the management of free space in a file during checkpoint.

If we only evict read-only pages and never allocate any blocks, then I think this approach works, but that seems like it won't cope with some of the cases we're concerned about where checkpoint needs to visit more data in the lookaside file than fits in cache.  But if we evict a dirty page, that will allocate space in the file but not appear in the tree, so verify will detect (correctly) that we have leaked disk blocks.

If we can solve the problem of evicting dirty pages behind a checkpoint, that would have some major benefits for avoiding stalls during write-heavy workloads where a single tree contains most of the dirty data in cache.  That seems hard, but I think it would be necessary to make this approach work.

keithbostic:
Agreed, closing in favor of the existing work in #3644.

PR #3672, opened 2017-09-24
WT-3603 format threads always traverse the same RNG space.

keithbostic:
While looking at one of the WT-3435/#3644 failures, I realized all of the format threads were pounding on the same key/value pairs. Interestingly enough, if you change that, the failure goes away.

I think we generally want the format threads to work different parts of the RNG space, but clearly we also want to stress the same key/value pairs as well.

### DIFF
--- a/test/format/config.h
+++ b/test/format/config.h
@@ -165,9 +165,9 @@ static CONFIG c[] = {
 	  "if values are huffman encoded",			/* 20% */
 	  C_BOOL, 20, 0, 0, &g.c_huffman_value, NULL },
 
-	{ "independent_rng",
+	{ "independent_thread_rng",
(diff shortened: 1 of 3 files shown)


agorrod:
Actually @keithbostic I was on the verge of merging this, but then thought the option may be critical for reproducing failures - do you think it should be exposed via the configuration file?

keithbostic:
> I was on the verge of merging this, but then thought the option may be critical for reproducing failures - do you think it should be exposed via the configuration file?

I think that's a good point (and it's not like configuration variables are expensive to add).

I've pushed that change for your consideration.

agorrod:
Thanks @keithbostic lgtm

PR #3673, opened 2017-09-25
WT-3602 Write the compatibility version given into the basecfg file.

sueloverso:
@michaelcahill or @agorrod Please review this change.  It only addresses fixing the compatibility in the `basecfg` file on database creation.  It does not address whether we should modify the `basecfg` file on a `conn->reconfigure()` call.  Currently the `basecfg` file is static after creation and that change would be a new path to modifying it later, unless we come up with some other solution.


agorrod:
Thanks @sueloverso this change looks good to me. I think we should address the reconfigure issue in a separate ticket, so I'm going to go ahead and merge this one now.

PR #3674, opened 2017-09-25
WT-3606 Set connection fields after error checking.

sueloverso:
@agorrod Please review.  I found this by inspection while looking at the compatibility code for another ticket.


agorrod:
Thanks @sueloverso lgtm

PR #3676, opened 2017-09-26
WT-3605 LSM doesn't support type 'r'

keithbostic:
@agorrod, I took a look at WT-3605 (because it was a user question), and here's a branch that I hope makes things better.

It looks to me like LSM doesn't currently support anything other than 'S' or 'u' because the underlying file format is always 'u' (and 'S' is the only other format than  'u' that naturally looks like 'u' when it's stored in a WT_ITEM structure).

That said, it seems LSM could support other formats? (For example, couldn't a record-number be packed and stored in a WT_ITEM as format 'u'?)

I'm also unclear on how this connects to more complex tables schemas, that can presumably be configured to store in LSM tables.

Anyway, I hope this branch is moving in the right direction. Happy to wrap this up if you can give me a little guidance, or happy to have it handed off to someone else.

### DIFF
--- a/src/docs/lsm.dox
+++ b/src/docs/lsm.dox
@@ -111,9 +111,12 @@ session.create("table:T", "type=lsm");
 @m_endif
 @endcode
 
-The default type for all schema objects will continue to be btree.
-
 @section lsm_caveats Caveats
+
+@subsection lsm_key_format Key_format configuration
+
+LSM only supports key formats of \c S or \c u, that is, \c key_format
+must be configured to one of those two column types.
 
 @subsection lsm_hazard Hazard configuration
 
--- a/src/docs/lsm.dox
+++ b/src/docs/lsm.dox
@@ -115,8 +115,8 @@ session.create("table:T", "type=lsm");
 
 @subsection lsm_key_format Key_format configuration
 
-LSM only supports key formats of \c S or \c u, that is, \c key_format
-must be configured to one of those two column types.
+LSM does not support column-store trees, \c key_format may not be set
+to type \c 'r'.
 
 @subsection lsm_checkpoints Named checkpoints
 
--- a/src/docs/lsm.dox
+++ b/src/docs/lsm.dox
@@ -118,15 +118,6 @@ session.create("table:T", "type=lsm");
 LSM only supports key formats of \c S or \c u, that is, \c key_format
 must be configured to one of those two column types.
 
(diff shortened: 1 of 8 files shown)


michaelcahill:
@keithbostic, LSM supports anything but 'r' -- e.g., `ComplexLSMDataSet` in the test suite.  All row store keys are passed to chunk cursors in raw mode.

The primary reason we don't support column stores is that I didn't implement `append`, which can't defer to the btree implementation or record numbers would restart each time the primary LSM chunk switches.

keithbostic:
@michaelcahill, I've updated the WT-3605/#3676 branch, I think it should be ready to merge.

src/lsm/lsm_tree.c
```diff
@@ -345,6 +346,14 @@ __wt_lsm_tree_create(WT_SESSION_IMPL *session,
 	WT_RET_NOTFOUND_OK(ret);
 
 	if (!F_ISSET(S2C(session), WT_CONN_READONLY)) {
+		/*
+		 * LSM only supports S or u key formats.
```
sulabhM: Should the comment be single lined and be updated to "supports everything other than 'r'"

michaelcahill:
@keithbostic I pushed a change to filter out LSM / recno combinations from scenarios rather than filtering.  Please take a look.

src/docs/lsm.dox
```diff
@@ -111,18 +111,12 @@ session.create("table:T", "type=lsm");
...
 
-@subsection lsm_hazard Hazard configuration
+@subsection lsm_key_format Key_format configuration
 
-Reads from an LSM cursor may need to position a cursor in each active chunk.
-The number of chunks depends on the chunk size, and how many chunks have
-been merged.  There must be at least as many hazard pointers available as
-there are chunks in the tree for each cursor that is open on the LSM tree.
-The number of hazard pointers is configured with the \c "hazard_max"
-configuration key to ::wiredtiger_open.
```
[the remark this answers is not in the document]
michaelcahill: I see you called that out in a commit, scratch it.

src/lsm/lsm_tree.c
keithbostic: Thank you, @sulabhM, fixed in 15e392549.

keithbostic:
>  I pushed a change to filter out LSM / recno combinations from scenarios rather than checking in each test.

Nice!

lgtm.

PR #3677, opened 2017-09-27
WT-3230 Add automatic style checking scripts

keithbostic:
@michaelcahill, @agorrod: this was one of my scripts that I had lying around, it complained if I was about to commit a file with unexpected local function declarations.

I've cleaned up to work on the file in-place, rather than just complaining.

The idea is to make it easy for people to match the existing function declaration style -- let's face it, it's hard for people not steeped in the KNF/BSD/whatever tradition to get this right, even if we could write it up in a coherent way.


keithbostic:
I pushed this one the wrong branch. Closing.

PR #3678, opened 2017-09-27
WT-3230 automatic style checking: function local variable declarations

keithbostic:
@michaelcahill, @agorrod: this was one of my scripts that I had lying around, it complained if I was about to commit a file with unexpected local function declarations.

I've cleaned up to work on the file in-place, rather than just complaining.

The idea is to make it easy for people to match the existing function declaration style -- let's face it, it's hard for people not steeped in the KNF/BSD/whatever tradition to get this right, even if we could write it up in a coherent way.


sueloverso:
@keithbostic I am a big fan of this!  I'm good with it as-is.  One extension or new script would be to flag if local vars on a single line are out of alphabetical order.  I happened to notice two lines where that is an issue in the diff.

keithbostic:
> One extension or new script would be to flag if local vars on a single line are out of alphabetical order. I happened to notice two lines where that is an issue in the diff.

I futzed with that once, but it's tricky. To get it "right" you almost have to have a real parser (for example, coping with something like `int (*bbb) (int x, int y), ccc, *aaa[3];`.

Thinking about it, we could probably get to 95% or better by sorting based on the first string in each comma-separated unit (where parenthesis protect commas from counting).

I'd be hesitant to merge lines:
```
int aaa;
int bbb;
```
should probably be merged, but I can imagine other constructs where it enhances readability to put declarations on different lines from other declarations.

sueloverso:
I'm definitely good with it not merging lines.  I'm also good getting close just to catch mistakes.  I'd be fine if we completely skip complex lines like you have above.

dist/function.py
```diff
@@ -0,0 +1,131 @@
...
+    line = line.strip()
+    line = re.sub("^const ", "", line)
+    line = re.sub("^volatile ", "", line)
+
+    # Let WT_UNUSED terminate the parse. It often appears at the beginning
+    # of the function and looks like a WT_XXX variable declaration.
+    if re.search('^WT_UNUSED', line):
+        return False,0
+
+    # Let lines not terminated with a semicolon terminate the parse, if means
```
sulabhM: typo "if" --> "it"

dist/function.py
```diff
@@ -0,0 +1,131 @@
...
+        # the local declarations. Quit tracking declarations as soon as we
+        # find anything we don't understand, leaving it untouched.
+        with open(name, 'r') as f:
+            tfile = open(tmp_file, 'w')
+            tracking = False
+            for line in f:
+                if not tracking:
+                    tfile.write(line)
+                    if re.search('^{$', line):
+                        r = [[] for i in range(30)]
```
sulabhM: I am wondering if we can get rid of "30" here. How about having a variable containing the count of illegal types, say illegal_type_count = 6, declared with "types". Then instead of "30" we can have "len(types) - illegal_type_count". That said, I am fine with how it is at the moment too, it is very readable.

dist/function.py
```diff
@@ -0,0 +1,131 @@
...
+                    continue
+
+                found,value = function_args(line)
+                if value == -1:
+                    print >>sys.stderr, \
+                        name + ": illegal declaration: " + line.strip()
+                    sys.exit(1)
+                if value == 0 or not found:
+                    for arg in filter(None, r):
+                        for p in sorted(arg, key=function_args_alpha):
```
sulabhM: Are we sorting individual variable on a single line here ?
A comment will help

dist/function.py
```diff
@@ -0,0 +1,131 @@
...
+                    tfile.write(line)
+                    tracking = False
+                    continue
+                r[value].append(line)
+
+            tfile.close()
+            compare_srcfile(tmp_file, name)
+
+# Report missing function comments.
+#missing_comment()
```
sulabhM: Since we removed it from style.py, should we enable it here

dist/function.py
```diff
@@ -0,0 +1,131 @@
...
+        skip_re = re.compile(r'DO NOT EDIT: automatically built')
+        func_re = re.compile(
+            r'(/\*(?:[^\*]|\*[^/])*\*/)?\n\w[\w \*]+\n(\w+)', re.DOTALL)
+        s = open(f, 'r').read()
+        if skip_re.search(s):
+            continue
+        for m in func_re.finditer(s):
+            if not m.group(1) or \
+               not m.group(1).startswith('/*\n * %s --\n' % m.group(2)):
+                   print "%s:%d: missing comment for %s" % \
```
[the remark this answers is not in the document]
keithbostic: Agreed, fixed in 4b7f599.

dist/function.py
keithbostic: Thanks! Fixed in 4b7f599.

dist/function.py
keithbostic: > I am wondering if we can get rid of "30" here. How about having a variable containing the count of illegal types, say illegal_type_count = 6, declared with "types". Then instead of "30" we can have "len(types) - illegal_type_count". That said, I am fine with how it is at the moment too, it is very readable.

Agreed, fixed in 4b7f599.

I ended up using enumerate and getting rid of the specific indexes, but my Python is weak, this one probably deserves a second look.

dist/function.py
keithbostic: > Are we sorting individual variable on a single line here ?

I added a comment in 4b7f599.

dist/function.py
keithbostic: > Since we removed it from style.py, should we enable it here
 
Thanks!  Turned it off for debugging, forgot to turn in back on.

Fixed in 4b7f599.

agorrod:
Thanks @keithbostic lgtm

PR #3680, opened 2017-09-27
WT-3609 Change __wt_panic to handle a NULL session argument

keithbostic:


michaelcahill:

PR #3681, opened 2017-09-28
WT-3610 Only require that the first commit_timestamp is oldest.

michaelcahill:
We don't actually need commit timestamps to move forward within a transaction as long as the first timestamp used is the oldest one (it determines the transaction's order in a list sorted by timestamp).


michaelcahill:
@sulabhM, can you please review?  The commit_timestamp order check was causing failures in MongoDB integration testing.

sulabhM:
LGTM, I will merge.

PR #3683, opened 2017-09-29
WT-3614 Full-build Friday lint

keithbostic:
@nehakhatri5, since you're on the East Coast, can I ask you for review, please?

Thanks!

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -386,14 +386,9 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 	/* Evict pages from the cache as needed. */
 	WT_RET(__evict_pass(session));
 
-	if (!F_ISSET(conn, WT_CONN_EVICTION_RUN) ||
-	    cache->pass_intr != 0)
+	if (!F_ISSET(conn, WT_CONN_EVICTION_RUN) || cache->pass_intr != 0)
 		return (0);
 
-	/*
-	 * Clear the walks so we don't pin pages while asleep,
-	 * otherwise we can block applications evicting large pages.
-	 */
 	if (!__wt_cache_stuck(session)) {
 		/*
 		 * Try to get the handle list lock: if we give up, that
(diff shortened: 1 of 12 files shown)


test/format/format.h
```diff
@@ -104,7 +104,7 @@ typedef struct {
 	WT_CONNECTION	 *wts_conn;
 	WT_EXTENSION_API *wt_api;
 
-	int   rand_log_stop;			/* Logging turned off */
+	bool  rand_log_stop;			/* Logging turned off */
```
nehakhatri5: Just one clarification required. There are other places in code where it is possible to switch the datatype from int to bool e.g. in function nop_compress, the argument compression_failed. I am assuming that the remaining such variables are for another pull request, not this one.

src/evict/evict_lru.c
```diff
@@ -404,7 +399,13 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 		if ((ret = __evict_lock_handle_list(session)) == EBUSY)
 			return (0);
 		WT_RET(ret);
+
+		/*
+		 * Clear the walks so we don't pin pages while asleep,
+		 * otherwise we can block applications evicting large pages.
+		 */
```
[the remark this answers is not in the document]
keithbostic: I think eviction is one of the areas of the system that's been reworked so many times it tends to suffer from fragmentation. (Pretty much everybody in the WiredTiger team has worked on it more than a few times.)

Anyway, any clarity you can add to it would be welcome -- I'd suggest a new ticket though, and @agorrod as your reviewer.

test/format/format.h
keithbostic: > There are other places in code where it is possible to switch the datatype from int to bool e.g. in function nop_compress, the argument compression_failed.

We converted to`bool` types two years ago (September, 2015). We never found a tool to help, we did it by hand (mostly by searching for integer variables that were set to 0 or 1), and I'm sure we missed some.

Any you find should be updated, if you create the ticket I'm happy to be your reviewer!

That said, `nop_compress` is an external API ([a WT_COMPRESSOR method](http://source.wiredtiger.com/2.9.3/struct_w_t___c_o_m_p_r_e_s_s_o_r.html)), so the `int` wasn't changed to `bool`.

nehakhatri5:
Thanks for the clarification, LGTM.

PR #3685, opened 2017-10-03
WT-3613 Enable eviction in LSM chunks after they are flushed.

michaelcahill:


src/include/schema.h
```diff
@@ -296,7 +296,9 @@ struct __wt_table {
 		F_CLR(session, WT_SESSION_LOCKED_CHECKPOINT);		\
 		__wt_spin_unlock(session, &__conn->checkpoint_lock);	\
 	}								\
+	__wt_yield();							\
 	op;								\
+	__wt_yield();							\
```
michaelcahill: This change is not related to the main fix here but was something I noticed during testing -- yielding here avoid operations like `verify` getting "stuck" temporarily while trying to close an LSM tree because the closing thread was doing 1000 yields while holding locks.

src/lsm/lsm_tree.c
```diff
@@ -133,10 +130,8 @@ __lsm_tree_close(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree, bool final)
 		 * other schema level operations will return EBUSY, even though
 		 * we're dropping the schema lock here.
 		 */
-		if (i % WT_THOUSAND == 0)
-			WT_WITHOUT_LOCKS(session,
-			    __wt_lsm_manager_clear_tree(session, lsm_tree));
-		__wt_yield();
+		WT_WITHOUT_LOCKS(session,
+		    __wt_lsm_manager_clear_tree(session, lsm_tree));
```
michaelcahill: This is the other half of the "it's better to yield without locks" change when closing LSM trees.

src/lsm/lsm_work_unit.c
```diff
@@ -429,6 +427,19 @@ __wt_lsm_checkpoint_chunk(WT_SESSION_IMPL *session,
...
 
+	/*
+	 * Enable eviction on the live chunk so it doesn't block the cache.
+	 * Future reads should direct to the on-disk chunk anyway.
+	 */
+	btree = session->dhandle->handle;
+	if (btree->evict_disabled_open) {
+		btree->evict_disabled_open = false;
+		__wt_evict_file_exclusive_off(session);
+	}
```
michaelcahill: This is the real fix.  I know we've oscillated on when to allow eviction in LSM chunks, but there is a special case that avoids forced eviction when `btree->lsm_primary` is set, and it stays set after this, so I'm inclined to fix the test failures now and chase any fallout in LSM stability later.

sueloverso:
Just FYI, I ran this branch on one of the CONFIGs from WT-3200 that failed nearly 100% of the time with a stuck cache and it is over 10 iterations.  So it is looking good and I'll let it keep running overnight.

agorrod:
Thanks @michaelcahill I agree that there may be some performance fallout to LSM users in this change, but pinning content in cache is really a bug so takes precedence.

@sueloverso thanks for confirming.

PR #3686, opened 2017-10-03
WT-3617 Fail update/restore with uncommitted updates after stable.

michaelcahill:
This can't happen at snapshot isolation (where only one update can be uncommitted per record), but can happen at lower isolation levels.  Since update/restore has to discard updates from the stable (on-page) update and older, it can't cope with an uncommitted update after the stable update.


agorrod:
Thanks @michaelcahill lgtm

PR #3688, opened 2017-10-03
WT-3620 pthread_condattr_init() allocates memory on FreeBSD

keithbostic:


michaelcahill:
@keithbostic, thanks for this.  Presumably the same is at least potentially true of `pthread_mutexattr_init` -- can you please also make the analogous change to `__wt_spin_init`?

keithbostic:
> Presumably the same is at least potentially true of pthread_mutexattr_init

I've pushed the related change for `pthread_mutexattr_init`.

michaelcahill:

PR #3693, opened 2017-10-04
WT-3621 Add a new test for a full backup followed by a table create.

sueloverso:
@agorrod Please review this new test.  It was a scenario we didn't test that came out of discussions about backup and schema operations.  The good news is that we pass gracefully and we don't have a bug that would cause recovery to fail.  We already test drop and rename in another test but not explicitly create like this.


agorrod:
Thanks @sueloverso lgtm

PR #3695, opened 2017-10-05
WT-3600 validate timestamps to be lowercase hex

sulabhM:


sulabhM:
@agorrod , please have a look. wiredtiger-test-unit-ppc is running over this branch and passes compilation.

agorrod:
Thanks @sulabhM

PR #3700, opened 2017-10-06
WT-3631 strip leading zeroes from timestamps before comparing

sulabhM:


michaelcahill:
@sulabhM, I pushed #3701 with the changes I was expecting here -- can you please take a look?

michaelcahill:
Thanks @sulabhM, lgtm.  I'll merge.

PR #3703, opened 2017-10-06
WT-3636 Account for page image sizes in cache consistently

keithbostic:
@agorrod, @michaelcahill, I did some review today based on talking to Bruce Lucas about HELP-5066, and found a couple of problems in our page-image accounting. I don't believe this would lead to eviction problems, but in the presence of mapped files, I think our statistics for page-image bytes would grow without bound.


michaelcahill:
Thanks @keithbostic, I agree that at worst this could lead to some confusing statistics, but don't believe we've seen any evidence of that in the field.  I'll merge.

PR #3704, opened 2017-10-06
WT-3629 cache underflow and logging

keithbostic:
@michaelcahill, this branch is ready for review, but mostly just to give us something concrete to talk about.

There are 4 changes:
* 3fc6a3c: `__wt_cache_bytes_other` needs to protect against underflow and it wasn't doing that. I fixed it by changing the reads to use `volatile`. I know you prefer setting variables to volatile instead of casting, but that will be a more invasive change because there are general routines that take `WT_CACHE` fields for update, and we'd have to create volatile versions of those functions and/or change many (most?) of the `WT_CACHE` fields to volatile. (For the record, I don't think setting more `WT_CACHE` fields to volatile is necessarily a bad idea.)

* 1aae727: the `__wt_cache_decr_XXX` functions were detecting overflow by doing an atomic decrement and checking for underflow. AFAICT, `__sync_sub_and_fetch` isn't going to be slower than `__sync_val_compare_and_swap` on platforms we care about, and I think it's a safer/cleaner approach.

* 1aae727: the `__wt_cache_decr_XXX` functions were only logging failure in `HAVE_DIAGNOSTIC` mode, and the HELP-5066 ticket made me think we should log the error in all cases. The previous code tried to avoid logging field underflow more than once, and I removed that -- in this code, we'll log every underflow of a value. If that's a bad idea, we can switch it to once a second or something. (Although the previous approach doesn't work; it used a static in the inline function, and static variables in inline functions always refer to the same object.)

* c0b0cd2: there was a special case function `__wt_cache_decr_zero_uint64` that reset values to 0 after any underflow. As this branch won't underflow the variable, I don't think that function is needed, I think we can use `__wt_cache_decr_check_uint64` instead.


michaelcahill:
@keithbostic, this change generally looks okay to me.  If we start doing more casts to volatile, we may want some version of this Linux macro:
```c
#define ACCESS_ONCE(x) (*(volatile typeof(x) *)&(x))
```
I'm not worried about switching from an atomic subtract to a compare and swap, but I am a little concerned about the loop.  In a hot cache where many threads are busy updating and evicting, it seems like looping here could become expensive.

Have you ever thought about using the same approach we have for statistics values to maintain cache accounting?  On the one hand, that should reduce cache line contention and make operations faster, but on the other it would make it more expensive to check for underflow...

Another consideration is that underflow in these calculations used to be a real problem but afaik we haven't seen it in the past ~2 years.  Maybe we can stop worrying about it unless we see it again (i.e., just assert that the result of each atomic subtract is less than 1<<62)?

keithbostic:
> If we start doing more casts to volatile, we may want some version of this Linux macro:

I agree, I'd be good to add that now, I think there is one other place we'd use it, `__wt_log_cmp()`.

> I'm not worried about switching from an atomic subtract to a compare and swap, but I am a little concerned about the loop.  In a hot cache where many threads are busy updating and evicting, it seems like looping here could become expensive.

FTR (so I can find it again), the best paper I found was: https://htor.inf.ethz.ch/publications/img/atomic-bench.pdf. I think the summary is that CAS shouldn't be significantly slower than FAA in most cases.

I agree about the loop.

> Have you ever thought about using the same approach we have for statistics values to maintain cache accounting?  On the one hand, that should reduce cache line contention and make operations faster, but on the other it would make it more expensive to check for underflow...

No, I hadn't thought about that, it's a good question. We could do that for values where we don't need a quiescent/correct total, but what about the values where we do need a correct total?

The places where we call `__wt_cache_decr_zero_uint64` are places we thought we needed correct totals, or, at least, values we didn't want to underflow. Getting those values wrong would mean run-time problems, for example, letting `WT_CACHE.bytes_dirty_intl` underflow means eviction is going to go nuts because it thinks there's an infinite number of dirty bytes in the cache.

> Another consideration is that underflow in these calculations used to be a real problem but afaik we haven't seen it in the past ~2 years.  Maybe we can stop worrying about it unless we see it again (i.e., just assert that the result of each atomic subtract is less than 1<<62)?

I agree we haven't seen problems in awhile. But, if we underflow in the field, there's no reason to believe we will continue to do useful work (see above, eviction goes nuts).

How about if we did the underflow check and panicked the system if it failed? I could live with that.

keithbostic:
autobuild: retest this please

michaelcahill:
@keithbostic, as discussed, let's try the approach where we do an atomic decrement and panic if the result is "less than zero" (i.e. unreasonably large).  That will need to go through a round of MongoDB testing before we can merge it into master, given where we are in the cycle...

keithbostic:
> let's try the approach where we do an atomic decrement and panic if the result is "less than zero" (i.e. unreasonably large).

Pushed in 58a8de8.

agorrod:
@keithbostic I pushed a change in here - I'd be happy for this to be merged now. I'm also happy to have a conversation next week if you prefer your solution.

michaelcahill:
@keithbostic, I talked this through with @agorrod and we chickened out of having a panic in this path.  I think this version addresses the original concerns about silent failures in the field while giving some chance that we won't fall in a heap if some bug triggers this.

keithbostic:
Thanks, @agorrod, @michaelcahill, I'm good to go with this version, I'll merge as soon as it builds.

PR #3707, opened 2017-10-08
WT-3619 compaction interactions with checkpoints and eviction.

keithbostic:
@michaelcahill, a possible approach for your consideration that makes the test case run.

* Of the changes, it might not be necessary to quit compaction if eviction is stuck. I didn't test it extensively, but not tapping the compaction thread to do eviction plus allowing checkpoints to interrupt compaction, appeared sufficient to fix the specific test case (at least, it didn't fail as quickly). Also note, the code in this branch allows eviction to starve compaction. That said, I think there are reasons compaction is a bad idea if there's eviction pressure and I can live with it if you and @agorrod can.

* This branch changes compaction to take/release the WT_BTREE handle flush-lock on every page it reviews. I don't expect that lock to be a contentious operation since checkpoint is the only other thread that takes it  (and if it is contentious, the first collision will stop compaction from trying again). If it's a problem, we could switch to read/write locks instead and/or we could acquire the lock on every N pages that compaction reviews.

* I used `__wt_eviction_needed` instead of `__wt_cache_stuck` to measure eviction's state because I didn't want to wait until eviction was totally screwed -- but that might not be the right thing to do.

* This doesn't help LSM, it has completely different code (using LSM merges) for compaction. I didn't look at whether or not LSM has similar problems to Btree.


keithbostic:
Update: using `__wt_eviction_needed` allows compaction to essentially kill itself if the cache is small enough, see `test_compact02`.

Switch to `__wt_cache_stuck` so the test suite passes, I don't expect to merge this branch as-is anyway.

michaelcahill:
@keithbostic, I think I'm generally okay with these changes.

The thing I'd like to talk through is whether the `WT_SESSION_NO_EVICTION` flag means that `compact` will keep filling the cache, blocking other operations until it gets stuck.  In other words, should we add at a call to `__wt_cache_eviction_check` to the compact loop when it isn't holding the `flush_lock`?

keithbostic:
> The thing I'd like to talk through is whether the `WT_SESSION_NO_EVICTION` flag means that `compact` will keep filling the cache, blocking other operations until it gets stuck.

I agree, I anticipate compact will fill the cache because it's reading faster than the eviction plus application threads are likely to be writing. One of the early versions I did included a pause to let eviction catch up, but I got rid of that at some point.

> In other words, should we add at a call to `__wt_cache_eviction_check` to the compact loop when it isn't holding the `flush_lock`?

That makes sense to me. I don't think that compact should continue if application threads are becoming involved with eviction because I don't want compaction to mess with query latency. That said, I don't want compact to spin forever, either, if the system is suffering eviction pressure anyway. How should we walk that line?

agorrod:
Thanks @keithbostic and @michaelcahill these changes look good to me - at first I thought some of the logic was at cross purposes - but then I realized they serve two different purposes: stopping a compact on cache stuck vs helping with eviction when there is cache pressure.

I'll go ahead and merge once this passes testing - it'll be nice to have one less stress test failure mode.

agorrod:
I can't spot the particular failure on Windows here - I suspect it was because I changed the Windows build to support multiple jobs in parallel.

agorrod:
autobuild: retest this please

PR #3709, opened 2017-10-08
WT-3622 Allow upper case hexadecimal timestamps

michaelcahill:


michaelcahill:
@sulabhM, this became necessary because I kicked off integration tests over the weekend and it turns out the MongoDB does generate upper case hex timestamps.  Can you please review?

PR #3710, opened 2017-10-08
WT-3641 Track maximum timestamp used in each btree.

michaelcahill:


agorrod:
Thanks @michaelcahill this change looks good to me. I'll merge when testing passes.

PR #3711, opened 2017-10-09
WT-3642 Avoid lookaside reads for dead trees.

michaelcahill:


michaelcahill:
@agorrod, the context is in the ticket -- MongoDB tests hit this case.  Can you please take a look?

src/evict/evict_file.c
```diff
@@ -37,6 +41,21 @@ __wt_evict_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
 		return (0);
 
+	walk_flags = WT_READ_CACHE | WT_READ_NO_EVICT;
+
+	/* If discarding a dead tree, remove any lookaside entries. */
+	if (F_ISSET(dhandle, WT_DHANDLE_DEAD) &&
+	    F_ISSET(S2C(session), WT_CONN_LAS_OPEN) &&
+	    !WT_IS_METADATA(dhandle) && !F_ISSET(btree, WT_BTREE_LOOKASIDE)) {
+		__wt_las_cursor(session, &las_cursor, &session_flags);
+		WT_TRET(__wt_las_remove_block(
```
agorrod: A comment here saying that '0' is special would be nice.

michaelcahill:
Thanks @agorrod, I've added detail to the comment and tried to clarify that this should never apply to metadata (if for no other reason, because the metadata cannot be dropped).  I'll go ahead and merge as discussed.

PR #3713, opened 2017-10-09
WT-3638 When restoring updates, don't discard on-page tombstones.

michaelcahill:
Reconciliation selects an update as the "on-page" version that is correct as-of the reconciliation.  If that update is a tombstone, no value will appear on the page.  Our API permits a modify of a deleted record, so we can't discard a tombstone if it was chosen as the on-page update, or a later modify update will have no starting value.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -1478,10 +1478,13 @@ __split_multi_inmem(
 		}
 
 		/*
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, for you -- found this after adding some assertions that updates lists don't consist of a single modify update.

PR #3714, opened 2017-10-09
WT-3640 Fix bytes-read and bytes-written statistics.

keithbostic:


michaelcahill:

PR #3715, opened 2017-10-09
WT-3611 Backup comment doesn't match the code.

keithbostic:


src/cursor/cur_backup.c
```diff
@@ -253,16 +253,11 @@ __backup_start(
...
 	    WT_FS_OPEN_CREATE, WT_STREAM_WRITE, &cb->bfs));
 	/*
-	 * If a list of targets was specified, work our way through them.
-	 * Else, generate a list of all database objects.
-	 *
-	 * Include log files if doing a full backup, and copy them before
-	 * copying data files to avoid rolling the metadata forward across
-	 * a checkpoint that completes during the backup.
+	 * If targets were specified, add them to the list, else add all of the
+	 * database objects to the list. Include log files if a full backup.
```
sueloverso: I see how you ended up there, but it feels awkwardly worded.  The second sentence makes it seem like it is different than the 'else' clause of the first sentence.  Maybe something like:
If targets were specified, add them to the list.  Otherwise it is a full backup so add all database objects, including log files, to the list.

keithbostic: I like your change, done.

PR #3717, opened 2017-10-10
WT-3642 Use calculated walk flags consistently when closing a tree.

michaelcahill:


michaelcahill:
The original change in #3711 was incomplete: it calculated `walk_flags` but didn't use the calculated value in all the places it should have.

agorrod:
lgtm

PR #3718, opened 2017-10-10
WT-3257 Add test for log=recover settings

sueloverso:
@agorrod Please review this new test.  It uncovered a bug, tracked in WT-3643. That line is commented out in this test for now.  This is ready to merge.


sueloverso:
This PR is in a race with WT-3643.  We can hold off on this one until that is merged and then I can push the fix for the test here.

agorrod:
Thanks @sueloverso the test case lgtm - I'm inclined to wait until WT-3643 is merged before merging this PR.

PR #3719, opened 2017-10-10
WT-3644 Port to FreeBSD release 11.1

keithbostic:


michaelcahill:
Thanks @keithbostic, lgtm and I tested with the versions of clang I have handy and they were happy to accept the flag.  I'll merge.

PR #3720, opened 2017-10-10
WT-3643 Set panic on error path if recovery needed.

sueloverso:
@michaelcahill Please review this.  There is explanation in the ticket as well.  Basically on the error path out of wiredtiger_open, we were checkpointing the metadata file, updating the checkpoint LSN causing us to ignore and never recover the data.  This is one approach, because having panic set means we discard the dhandle.  Given it is the error path of `wiredtiger_open` it seemed cleaner than drilling more special case connection checking down into the dhandle code.

### DIFF
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -2645,8 +2645,15 @@ err:	/* Discard the scratch buffers. */
 		__wt_scr_discard(session);
 	__wt_scr_discard(&conn->dummy_session);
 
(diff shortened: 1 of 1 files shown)


michaelcahill:
Thanks @sueloverso, this matches what we're now doing if there is an error during close.  It's a little subtle, but if it causes confusion we can wrap this "soft panic" to make it clearer.  I'll merge.

PR #3721, opened 2017-10-11
WT-3585 round up read timestamp to oldest

bvpvamsikrishna:
Added test in test_timestamp03.py


bvpvamsikrishna:
autobuild:retest this please

bvpvamsikrishna:
@sulabhM @nehakhatri5 

src/txn/txn.c
```diff
@@ -440,11 +440,37 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
 		wt_timestamp_t ts;
+		WT_TXN_GLOBAL *txn_global;
 
+		txn_global = &S2C(session)->txn_global;
 		WT_RET(__wt_txn_parse_timestamp(session, "read", &ts, &cval));
-		WT_RET(__wt_timestamp_validate(session,
-		    &ts, &cval, true, false, false));
+		WT_WITH_TIMESTAMP_READLOCK(session, &txn_global->rwlock,
+		    cmp = __wt_timestamp_cmp(&ts,
+		    &txn_global->oldest_timestamp));
```
sulabhM: You should check if txn_global->has_oldest_timestamp is set before comparing with txn_global->oldest_timestamp. For example like what you have in __wt_timestamp_validate, you can do:

```
WT_WITH_TIMESTAMP_READLOCK(session, &txn_global->rwlock,
    older_than_oldest_ts = (txn_global->has_oldest_timestamp &&
        __wt_timestamp_cmp(&ts, &txn_global->oldest_timestamp) < 0));
if (older_than_oldest_ts) {
    // blah blah
}
```

src/txn/txn.c
```diff
@@ -440,11 +440,37 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+			 * If given read timestamp is earlier than oldest
+			 * timestamp then round the read timestamp to
+			 * oldest timestamp based on round_to_oldest option.
+			 */
+			WT_RET(__wt_config_gets_def(session, cfg,
+			    "round_to_oldest", 0, &cval));
+			round_to_oldest = cval.val;
+			if (round_to_oldest) {
+				__wt_verbose_timestamp(session, &ts,
+				"Rounded to oldest timestamp");
```
sulabhM: Rounded what?
Would "Rounded read timestamp to oldest timestamp" be more descriptive ?

src/txn/txn.c
```diff
@@ -440,11 +440,37 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+				WT_WITH_TIMESTAMP_READLOCK(session,
```
sulabhM: We take lock twice to obtain the oldest ts, once to compare, next time to set the read ts equal to the oldest.
Can we do this by taking lock once and caching the oldest ts in a local variable.

src/txn/txn.c
```diff
@@ -440,11 +440,37 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
 	if (cval.len > 0) {
 #ifdef HAVE_TIMESTAMPS
+		bool round_to_oldest;
+		int cmp;
 		wt_timestamp_t ts;
+		WT_TXN_GLOBAL *txn_global;
 
+		txn_global = &S2C(session)->txn_global;
 		WT_RET(__wt_txn_parse_timestamp(session, "read", &ts, &cval));
-		WT_RET(__wt_timestamp_validate(session,
```
sulabhM: With this change we should call __wt_timestamp_validate something as __wt_commit_timestamp_validate. That function was used to validate read_timestamp, but even if validation failed it would have given output saying commit_timestamp older than oldest .. . It was a result of trying to generalise that function for all timestamp validation but I guess we were not very successful at that. That function really only validates commit_timestamp.

sulabhM:
I had a chat with Alex, it looks like we have to guarantee that if the flag "round_to_oldest" is set, we make sure that read_timestamp doesn't fall behind the oldest. The problem with the current change is that oldest timestamp can be updated between the validation and putting that read_timestamp in the "txn->read_timestamp" queue. I don't have a solution to this issue, probably we need to have a chat around the problem.

sulabhM:
I suggest adding an assert after taking the read queue lock that the read_timestamp we are adding to the queue is not older than the oldest timestamp. The function that does that is __wt_txn_set_read_timestamp. There is a possibility of this happening, when the application is not being diligent in synchronising its read transactions with setting the oldest timestamp and ends up calling set_timestamp oldest while trying to open a transaction.

src/txn/txn.c
bvpvamsikrishna: The output will be time Timestamp : <timestamp val> : Rounded to oldest timestamp

sulabhM: Let's say timestamp value is 56, the message will be for instance "Timestamp 56 : Rounded to oldest timestamp".
Would it make more sense to say something like:
"Timestamp 56 : Rounded read timestamp to the oldest timestamp"

It conveys that a "read timestamp" got rounded and points to coming from the configure transaction semantics.

src/txn/txn.c
```diff
@@ -440,11 +440,37 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+		    &txn_global->oldest_timestamp));
+		if (cmp < 0) {
+			/*
+			 * If given read timestamp is earlier than oldest
+			 * timestamp then round the read timestamp to
+			 * oldest timestamp based on round_to_oldest option.
+			 */
+			WT_RET(__wt_config_gets_def(session, cfg,
+			    "round_to_oldest", 0, &cval));
+			round_to_oldest = cval.val;
```
nehakhatri5: A new variable 'round_to_oldest' is not requires in the function. In the if condition below you can directly use the 'cval.val', like this "if (cval.val) {". This way an additional variable and an assignment can be avoided. That would  also be consistent with other cval usages in the code.

src/txn/txn.c
nehakhatri5: The alignment can be corrected by adding 4 spaces.

test/suite/test_timestamp03.py
```diff
@@ -269,6 +269,24 @@ def test_timestamp03(self):
         self.check(self.session, 'read_timestamp=' + old_ts,
             self.table_nots_nolog, dict((k, self.value2) for k in orig_keys))
 
+        # Scenario: 4a
+        # This scenario is same as earlier one with read_timestamp earlier than
+         # oldest_timestamp and using the option of round_to_oldest
+        earlier_ts = timestamp_str(90)
+        self.check(self.session,
```
nehakhatri5: We should also test the case where round_to_oldest = false and read_timestamp is older than oldest timestamp.

sulabhM: read timestamp older than the oldest timestamp with default (false) round_to_oldest is already tested in test_timestamp09

bvpvamsikrishna:
@sulabhM  @nehakhatri5 
Request to check updates, suggest alternatives for line 457.

src/txn/txn.c
```diff
@@ -441,20 +441,61 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+		WT_TXN_GLOBAL *txn_global;
+		char timestamp_buf[2 * WT_TIMESTAMP_SIZE + 1];
+		bool round_to_oldest;
 
+		txn_global = &S2C(session)->txn_global;
 		WT_RET(__wt_txn_parse_timestamp(session, "read", &ts, &cval));
-		WT_RET(__wt_timestamp_validate(session,
-		    &ts, &cval, true, false, false));
-		__wt_timestamp_set(&txn->read_timestamp, &ts);
+		WT_RET(__wt_config_gets_def(session, cfg, "round_to_oldest", 0,
```
sulabhM: The convention followed is that if you need to wrap to multiple lines, the later lines get more arguments. So in this case : 
```
WT_RET(__wt_config_gets_def(session,
    cfg, "round_to_oldest", 0, &cval));
```

src/txn/txn.c
```diff
@@ -441,20 +441,61 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+		    &cval));
+		round_to_oldest = false;
+		__wt_readlock(session, &txn_global->rwlock);
+		if (__wt_timestamp_cmp(&ts, &txn_global->oldest_timestamp) < 0)
+		{
+			/*
+			 * To circumvent error: unused-result
```
sulabhM: Can fit this comment in a single line: 
/* To circumvent error: unused-result */

src/txn/txn.c
```diff
@@ -441,20 +441,61 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+			 */
+			(void)(__wt_timestamp_to_hex_string(session,
+			    timestamp_buf, &ts) + 1);
+			round_to_oldest = cval.val;
```
sulabhM: Why not move this outside the if block, where you parse the value passed for round_to_oldest. We are initializing to false there, why not just set to the value user passed there itself.

src/txn/txn.c
```diff
@@ -441,20 +441,61 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+		WT_RET(__wt_config_gets_def(session, cfg, "round_to_oldest", 0,
+		    &cval));
+		round_to_oldest = false;
+		__wt_readlock(session, &txn_global->rwlock);
+		if (__wt_timestamp_cmp(&ts, &txn_global->oldest_timestamp) < 0)
+		{
+			/*
+			 * To circumvent error: unused-result
+			 */
+			(void)(__wt_timestamp_to_hex_string(session,
```
sulabhM: You should do WT_RET(__wt_timestamp_to_hex_string(..));

That will get rid of compiler complaining about unused return. Also if this function fails the error will be handled gracefully.

src/txn/txn.c
bvpvamsikrishna: because this block is not required, when if is evaluated to false.

bvpvamsikrishna:
@agorrod 
Just though of making my concern explicit about this change. The call to  __wt_txn_set_read_timestamp is with in a critical section with txn_global->rwlock. The function __wt_txn_set_read_timestamp also has a critical section inside.

src/txn/txn.c
sulabhM: That looks a bit inconsistent to me. Either of the following two look consistent:

1. parse value for "round_to_oldest" config parameter outside the if block and set local variable round_to_oldest to that value there itself.
2. set round_to_oldest to false outside the if block and then inside if block parse value for "round_to_oldest" config parameter and set round_to_oldest.

I like option 1 better, but that is my opinion. The code currently mixes the above two approaches.
I would let second reviewer look at this, as this is more of code layout than a correctness thing

sulabhM:
Also evaluate if there is a merit in adding the assert that timestamp being added to the read queue in __wt_txn_set_read_timestamp is not older than the oldest.

src/txn/txn.c
```diff
@@ -441,20 +441,58 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+			    timestamp_buf);
+		}
 #else
 		WT_RET_MSG(session, EINVAL, "read_timestamp requires a "
 		    "version of WiredTiger built with timestamp support");
 #endif
 	}
 
 	return (0);
+
```
agorrod: This whitespace change is wrong

src/txn/txn.c
```diff
@@ -441,20 +441,58 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
+		bool round_to_oldest;
 
+		txn_global = &S2C(session)->txn_global;
 		WT_RET(__wt_txn_parse_timestamp(session, "read", &ts, &cval));
-		WT_RET(__wt_timestamp_validate(session,
-		    &ts, &cval, true, false, false));
-		__wt_timestamp_set(&txn->read_timestamp, &ts);
+		WT_RET(__wt_config_gets_def(session,
+		    cfg, "round_to_oldest", 0, &cval));
+		round_to_oldest = false;
```
agorrod: This is a weird semantic - we pull the configuration value into cval, then assign an unconditional value to `round_to_oldest`. I can see that it's done so that `round_to_oldest` is always initialized for the verbose block below, but the code would be clearer if this initialization was done above in a local variable initialization block (like `txn_global`). Then the `__wt_config_gets_def` call deserves a comment explaining that we are parsing the configuration value now to avoid needing to do it while holding the `txn_global` lock.
It's a bit sad that all WT_TRANSACTION::begin calls will now have to take this lock and do some timestamp comparison - does this change introduce performance regressions? The check is really only for enforcing our API contract - maybe that could happen only in diagnostic mode.

agorrod: I talked to @bvpvamsikrishna directly about this. My performance concerns were unfounded - this change is replacing a call to validate timestamps that did the same locking.
My remaining concern is around the usage of `round_to_oldest` - and we discussed a resolution

PR #3722, opened 2017-10-11
WT-3646 Only use lookaside when operations are blocked waiting for cache

michaelcahill:
We recently changed eviction to try using the lookaside table sooner. For some workloads, this can lead to poor performance and runaway growth in the lookaside table size.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1406,14 +1406,14 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 
 #ifdef HAVE_TIMESTAMPS
 	/* Track the oldest saved timestamp for lookaside. */
-	if (F_ISSET(r, WT_REC_LOOKASIDE)) {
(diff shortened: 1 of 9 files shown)


michaelcahill:
Please hold off on reviewing for now: I'm going to try a few more changes here to see if I can resolve some other issues with lookaside eviction that I'm seeing while testing.

agorrod:
Thanks @michaelcahill lgtm - go ahead and merge after tests pass.

src/reconcile/rec_write.c
```diff
@@ -1982,6 +1994,29 @@ __rec_leaf_page_max(WT_SESSION_IMPL *session, WT_RECONCILE *r)
 	return (page_size * 2);
 }
 
+#define	WT_REC_MAX_SAVED_UPDATES	100
```
keithbostic: @michaelcahill, an alternative might be to track the memory in the saved updates and use that value plus the argument `len` as the comparison against `r->space_avail`.

PR #3723, opened 2017-10-11
WT-3646 Improve performance when lookaside eviction is required.

michaelcahill:
In particular,
* fix the optimization for checkpoints so most lookaside pages can be skipped when the stable timestamp is lagging.
* allow row store leaf pages to split when evicting and no values are visible (currently using a dumb heuristic of splitting as soon as 100 records need updates saved).
* allow use of lookaside and splits when checkpoints do eviction.

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -344,6 +344,9 @@ __evict_page_dirty_update(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
 		/*
 		 * Update the parent to reference the replacement page.
 		 *
+		 * A page evicted with lookaside entries may not have an
+		 * address, if no updates were visible to reconciliation.
+		 *
 		 * Publish: a barrier to ensure the structure fields are set
 		 * before the state change makes the page available to readers.
 		 */
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -574,10 +574,14 @@ __evict_review(
 			}
 
 			/*
-			 * Check if reconciliation suggests trying the
-			 * lookaside table.
+			 * We usually only consider lookaside eviction once
+			 * eviction is struggling.  However, eviction in
+			 * service of a checkpoint (which is only done when
+			 * urgent eviction is required) will try lookaside
+			 * eviction immediately.
 			 */
-			if (__wt_cache_aggressive(session) &&
+			if ((__wt_cache_aggressive(session) ||
+			    WT_SESSION_IS_CHECKPOINT(session)) &&
 			    F_ISSET(cache, WT_CACHE_EVICT_CLEAN |
 				WT_CACHE_EVICT_DIRTY_HARD) &&
 			    !F_ISSET(conn, WT_CONN_EVICTION_NO_LOOKASIDE))
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1293,7 +1293,7 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 			*updp = upd;
 
 #ifdef HAVE_TIMESTAMPS
-		/* Track the first update with non-zero timestamp. */
+		/* Track the first/last update with non-zero timestamp. */
 		if (first_ts_upd == NULL &&
 		    !__wt_timestamp_iszero(&upd->timestamp))
 			first_ts_upd = upd;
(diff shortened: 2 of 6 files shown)


src/evict/evict_page.c
```diff
@@ -574,10 +574,14 @@ __evict_review(
...
-			 * lookaside table.
+			 * We usually only consider lookaside eviction once
+			 * eviction is struggling.  However, eviction in
+			 * service of a checkpoint (which is only done when
+			 * urgent eviction is required) will try lookaside
+			 * eviction immediately.
 			 */
-			if (__wt_cache_aggressive(session) &&
+			if ((__wt_cache_aggressive(session) ||
+			    WT_SESSION_IS_CHECKPOINT(session)) &&
```
keithbostic: I'm not sure why we'd want this change. Is there reason to believe the CPU is struggling, what's the additional cost of a second reconciliation attempt? Or are we seeing consistent failure on the first reconciliation?

src/reconcile/rec_write.c
```diff
@@ -1660,6 +1660,11 @@ __rec_child_modify(WT_SESSION_IMPL *session,
 			if (F_ISSET(r, WT_REC_EVICT))
 				return (EBUSY);
 
+			if (ref->addr == NULL) {
+				*statep = WT_CHILD_IGNORE;
+				WT_CHILD_RELEASE(session, *hazardp, ref);
+			}
+
```
keithbostic: I had to stare at this again, I think we need a comment in `__evict_page_dirty_update()` to flag that lookaside pages may not have addresses? A comment here wouldn't be bad, either.

src/reconcile/rec_write.c
```diff
@@ -1997,6 +2002,25 @@ __rec_leaf_page_max(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+__rec_need_split(WT_RECONCILE *r, size_t next_len)
+{
+	if (r->page->type == WT_PAGE_ROW_LEAF &&
+	    r->supd_next >= WT_REC_MAX_SAVED_UPDATES)
+		return (true);
+
+	if (r->raw_compression)
+		return (next_len > r->space_avail);
+	else
+		return (WT_CHECK_CROSSING_BND(r, next_len));
```
keithbostic: You're going to get complaints from the static analyzers about an else after a return, I'd suggest:
```
return (r->raw_compression ?
    next_len > r->space_avail : WT_CHECK_CROSSING_BND(r, next_len));
```

src/reconcile/rec_write.c
```diff
@@ -2567,7 +2592,8 @@ __rec_split_crossing_bnd(
 	 * large enough, just split at this point.
 	 */
 	if (WT_CROSSING_MIN_BND(r, next_len) &&
-	    !WT_CROSSING_SPLIT_BND(r, next_len)) {
+	    !WT_CROSSING_SPLIT_BND(r, next_len) &&
+	    !__rec_need_split(r, 0)) {
```
keithbostic: You can merge these two lines.

src/reconcile/rec_write.c
```diff
@@ -2558,7 +2582,8 @@ __rec_split_crossing_bnd(
 	WT_BTREE *btree;
 	size_t min_offset;
 
-	WT_ASSERT(session, WT_CHECK_CROSSING_BND(r, next_len));
+	WT_ASSERT(session, WT_CHECK_CROSSING_BND(r, next_len) ||
+	    __rec_need_split(r, 0));
```
keithbostic: I think the `WT_CHECK_CROSSING_BND` test in this assertion is redundant.

src/reconcile/rec_write.c
```diff
@@ -1997,6 +2002,25 @@ __rec_leaf_page_max(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+{
+	if (r->page->type == WT_PAGE_ROW_LEAF &&
+	    r->supd_next >= WT_REC_MAX_SAVED_UPDATES)
+		return (true);
+
+	if (r->raw_compression)
+		return (next_len > r->space_avail);
+	else
+		return (WT_CHECK_CROSSING_BND(r, next_len));
+}
```
keithbostic: I understand what `__rec_need_split()` does, and I think it will work, but nothing in the ticket, commit or code explains why we're doing it?

test/suite/test_timestamp07.py
```diff
@@ -135,6 +163,15 @@ def ckpt_backup(self, check_value, valcnt, valcnt2, valcnt3):
         self.session.checkpoint(ckptcfg)
         self.backup_check(check_value, valcnt, valcnt2, valcnt3)
 
+    def check_stable(self, check_value, valcnt, valcnt2, valcnt3):
+        self.ckpt_backup(check_value, valcnt, valcnt2, valcnt3)
+
+        # When reading as-of a timestamps, tables 1 and 3 should match (both
+        # use timestamps and we're not running recovery, so logging behavior
+        # should be irrelevant).
+        self.check_reads(self.session, 'read_timestamp=' + self.stablets,
```
keithbostic: Typo, `as-of a timestamps`.

src/reconcile/rec_write.c
```diff
@@ -1997,6 +2002,25 @@ __rec_leaf_page_max(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
 }
 
+#define	WT_REC_MAX_SAVED_UPDATES	100
+
+/*
+ * __rec_need_split --
+ *	Check whether adding some bytes to the page requires a split.
+ */
+static bool
+__rec_need_split(WT_RECONCILE *r, size_t next_len)
```
keithbostic: I'd rename `next_len` to be `len`, here.

src/reconcile/rec_write.c
keithbostic: I'm confused why this isn't an existing bug, that is, I'm not sure why this change wasn't needed before today. Couldn't we get here with an empty lookaside page before?

michaelcahill: This is an existing bug, but because pages with lookaside entries weren't splitting, our existing tests that do checkpoints with lookaside entries always ended up with a non-empty on-disk page image.

michaelcahill:
Thanks for the review @keithbostic, I've made the changes you recommended.  Merging this into the other WT-3646 branch...

PR #3724, opened 2017-10-12
WT-3649 Disable lookaside eviction during close.

michaelcahill:


agorrod:
Thanks @michaelcahill this change seems correct - should I go ahead and merge?

PR #3725, opened 2017-10-12
WT-3645 Allow eviction of lookaside pages as soon as writes commit.

michaelcahill:


agorrod:
Thanks @michaelcahill I really like the idea of this change and the implementation makes sense to me. I'll go ahead and merge.

PR #3726, opened 2017-10-12
WT-3646 Don't trigger eviction from checkpoints in write heavy workloads

michaelcahill:
Previously, checkpoint attempted to push out any page requiring forced eviction.  With this change, pages read into cache by checkpoint are marked with a different read generation, so that checkpoint can distinguish between pages it read (from lookaside) and big / hot pages that happen to be in cache when a checkpoint runs.


agorrod:
lgtm

PR #3727, opened 2017-10-12
WT-3646 Revert a change to rec_write.c that broke lookaside testing.

agorrod:
Which is setting a timestamp for lookaside evictions.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -1392,14 +1392,14 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 
 #ifdef HAVE_TIMESTAMPS
 	/* Track the oldest saved timestamp for lookaside. */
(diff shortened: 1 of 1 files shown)


agorrod:
@michaelcahill and @keithbostic I did some digging into the cause of the [long test suite](http://build.wiredtiger.com:8080/job/wiredtiger-test-unit-long/2659/) failures - and it turned out to be this part of the original WT-3646 change - I didn't fully understand the intent of this in the original change - and my attempt at reverting it obviously re-introduced the original problem it was addressing.

I'll keep digging for now, but it'd be great if one of you could pick this up as well - since I'm unlikely to figure it out. If you pick this up I recommend using the changes in the PR for [WT-3651](https://github.com/wiredtiger/wiredtiger/pull/3728) which makes test_las run in about 10 seconds with those changes

michaelcahill:
This should be fixed by #3729.

PR #3730, opened 2017-10-13
WT-3223 Checkpoint progress messages

nehakhatri5:
During a database checkpoint, output verbose progress messages indicating the the cumulative amount of pages synced every 20 seconds


agorrod:
autobuild: retest this please

I think the Windows failure was probably due to a change I made in Jenkins rather than a problem here.

src/btree/bt_sync.c
```diff
@@ -300,12 +352,27 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+					__sync_checkpoint_update_internal(
+					    conn, page->memory_footprint);
+				}
 			} else {
 				leaf_bytes += page->memory_footprint;
 				++leaf_pages;
+
+				/* Update the checkpoint progress data */
+				if (track_ckpt_progress) {
+					__sync_checkpoint_update_leaf(
```
sulabhM: both __sync_checkpoint_update_internal and __sync_checkpoint_update_leaf do very similar thing, the size argument is actually the same for both. I wonder if we can replace two functions with a single function or even a macro.

src/btree/bt_sync.c
```diff
@@ -340,6 +407,13 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 			evict_failed = false;
 			WT_ERR(__wt_reconcile(
 			    session, walk, NULL, WT_REC_CHECKPOINT, NULL));
+
+			/* Track checkpoint progress. */
+			if (++i > 5000 ) {
```
sulabhM: replace 5000 with 5 * WT_THOUSAND. Can add a comment explaining why "5000"

sulabhM: If we are not tracking checkpoint progress why increment i and set it to 0, ie we can replace the code with:

```
if (track_ckpt_progress && (++i > 5000)) {
     __sync_checkpoint_progress(session);
    i = 0;
}
```

We will also end up doing less comparisons and increments.

src/include/connection.h
```diff
@@ -268,6 +268,19 @@ struct __wt_connection_impl {
 	uint64_t  ckpt_time_recent;	/* Checkpoint time recent/total */
 	uint64_t  ckpt_time_total;
 
+	/* Checkpoint stats and verbosity timers */
+	struct timespec ckpt_start_time;/* Checkpoint start/stop time */
+	struct timespec ckpt_stop_time;
+					/* Checkpoint verbosity start time */
+	struct timespec ckpt_verb_start_time;
```
sulabhM: ckpt_start_time vs ckpt_verb_start_time, we can rename variables in such a way that I can tell which one includes the scrub time and which one doesn't.

src/include/connection.h
```diff
@@ -268,6 +268,19 @@ struct __wt_connection_impl {
 	uint64_t  ckpt_time_recent;	/* Checkpoint time recent/total */
 	uint64_t  ckpt_time_total;
 
+	/* Checkpoint stats and verbosity timers */
+	struct timespec ckpt_start_time;/* Checkpoint start/stop time */
+	struct timespec ckpt_stop_time;
+					/* Checkpoint verbosity start time */
```
sulabhM: We start this comment from the start of the line.

src/txn/txn_ckpt.c
```diff
@@ -536,33 +536,32 @@ __checkpoint_stats(
...
 #ifdef HAVE_VERBOSE
+	WT_CONNECTION_IMPL *conn;
 	struct timespec stop;
 	uint64_t msec;
 
 	if (!WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT))
 		return;
 
+	conn = S2C(session);
+
```
sulabhM: probably dont need this new line

src/btree/bt_sync.c
```diff
@@ -300,12 +352,27 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 				continue;
 			}
 
+			track_ckpt_progress =
+			    conn->ckpt_verb_start_time.tv_sec > 0;
```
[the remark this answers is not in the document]
nehakhatri5: This code is under the switch case WT_SYNC_CHECKPOINT and whenever the control is in this case that means it is here for checkpoint operation and will have a ckpt_verb_start_time set at the beginning of the checkpoint. Hence it is not required to reset this time.

__sync_file gets called only from __wt_cache_op. __wt_cache_op calls __sync_file only under following two cases:
        case WT_SYNC_CHECKPOINT:
        case WT_SYNC_WRITE_LEAVES:

The code that tracks the progress is under the case WT_SYNC_CHECKPOINT only, so the progress tracking will not happen in any other __sync_file call that does not belong to a checkpoint i.e. the progress data will not be updated outside of a checkpoint operation.

src/btree/bt_sync.c
nehakhatri5: But the two functions are operating on two different variables. So I would leave that with 2 inline functions.

src/btree/bt_sync.c
nehakhatri5: If track_ckpt_progress is true, then it would lead to two evaluations in the if statements all the time. This if condition is here to periodically check the checkpoint progress. In my opinion there would not be significant gain by changing it.

src/include/connection.h
nehakhatri5: This code comment is specific for the member variable, but as it won't fit in the same line of declaration I have put it in a line above. e.g. see https://github.com/wiredtiger/wiredtiger/blob/b284b2ff3768b05a30ee72eb14741947ea3fa09a/src/include/connection.h#L207

src/btree/bt_sync.c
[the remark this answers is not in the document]
agorrod: This deserves a comment explaining that the code path is only reached when doing checkpoints. I believe @sulabhM is right as well though - if the verbose setting is disabled via a reconfigure call, not resetting the timer at the end of checkpoint will mean this code doesn't work as expected.

src/btree/bt_sync.c
agorrod: Sulabh was right here - the code is clearer when explicitly checking the variable, and this isn't enough of a performance path to justify removing a comparison.

src/include/connection.h
agorrod: @nehakhatri5 there are some examples of having parameter comments indented on separate lines, but we try to avoid doing it because it is ambiguous which line they are associated with. In this case I feel like a single block comment along with well named variables is enough.

nehakhatri5:
@sulabhM, @bvpvamsikrishna and @agorrod , I have addressed the review issue. Please have a look.

src/btree/bt_sync.c
```diff
@@ -137,6 +137,57 @@ err:	WT_TRET(__wt_page_release(session, next, flags));
...
+ */
+static void
+ __sync_checkpoint_progress(WT_SESSION_IMPL *session)
+{
+	struct timespec cur_time;
+	WT_CONNECTION_IMPL *conn;
+	uint64_t time_diff;
+
+	conn = S2C(session);
+	time_diff = 0;
```
sulabhM: Don't need to initialise to zero, we overwrite that value couple of lines below.

src/include/connection.h
```diff
@@ -268,6 +268,18 @@ struct __wt_connection_impl {
 	uint64_t  ckpt_time_recent;	/* Checkpoint time recent/total */
 	uint64_t  ckpt_time_total;
 
+	/* Checkpoint stats and verbosity timers */
+	struct timespec ckpt_start_time_after_scrub;
+	struct timespec ckpt_stop_time;
+	struct timespec ckpt_verb_start_time;
```
sulabhM: maybe we should call it ckpt_start_time_before_scrub. When I am reading code where we use ckpt_verb_start_time, I have to scroll down to these declarations to figure if this one starts before or after scrub phase.

src/btree/bt_sync.c
```diff
@@ -137,6 +137,57 @@ err:	WT_TRET(__wt_page_release(session, next, flags));
...
+	__wt_epoch(session, &cur_time);
+
+	/* Time since the full database checkpoint started */
+	time_diff = WT_TIMEDIFF_SEC(cur_time, conn->ckpt_verb_start_time);
```
sulabhM: I can't tell without scrolling to declarations if ckpt_verb_start_time started before or after scrub phase.

src/btree/bt_sync.c
```diff
@@ -137,6 +137,57 @@ err:	WT_TRET(__wt_page_release(session, next, flags));
...
+
+	if ((time_diff / 20) > conn->ckpt_progress_msg_count) {
+		__wt_verbose(session, WT_VERB_CHECKPOINT_PROGRESS, "Checkpoint"
```
sulabhM: @agorrod I am not sure if there are strong guidelines on how to structure such multiline statements. Your point of view will be appreciated

src/txn/txn_ckpt.c
```diff
@@ -941,6 +946,8 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	cache->eviction_scrub_limit = 0.0;
 	WT_STAT_CONN_SET(session, txn_checkpoint_scrub_target, 0);
 
+	conn->ckpt_verb_start_time.tv_sec = 0;
```
sulabhM: This timer doesn't need to be reset twice, once before "err:" label and once inside "err:" label.
We can reset once at the start of the "err:" label and that should suffice.

src/include/connection.h
nehakhatri5: The purpose of this time is to display time delta from checkpoint start, in verbose messages. Renaming it to ckpt_start_time_before_scrub might not imply the reason behind using this time. I can rename it to ckpt_verb_start_time_before_scrub if that sounds OK.

src/btree/bt_sync.c
nehakhatri5: Good point.

nehakhatri5:
@agorrod The code seems good for your look now. Please review.

agorrod:
@nehakhatri5 Please take a look at the change I pushed last night which was a continuation of our pair programming review and resolve the merge conflict.

nehakhatri5:
@agorrod As discussed, further updated the code and resolved the merge conflict. Please have look.

agorrod:
Thanks @nehakhatri5 - I've pushed a tiny change, this lgtm now. I'll merge once it passes testing.

PR #3731, opened 2017-10-13
WT-3553 Change testutil_die to abort rather than exit

bvpvamsikrishna:


keithbostic:
autobuild: retest this please

PR #3732, opened 2017-10-13
WT-3655 Don't dirty pages to induce lookaside eviction.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm - I'll merge once it passes testing.

PR #3733, opened 2017-10-13
WT-3598 Open cursor not to set txn error

bvpvamsikrishna:


bvpvamsikrishna:
@sulabhM @nehakhatri5 

src/session/session_api.c
```diff
@@ -444,18 +444,11 @@ err:		if (cursor != NULL)
 	}
 
 	/*
-	 * Opening a cursor on a non-existent data source will set ret to
```
nehakhatri5: This block of comment should still be present as it explains that it is possible to get either ENOENT or WT_NOTFOUND when opening a cursor. And hence it would be required to map the WT_NOTFOUND to ENOENT in CURSOR_OPEN_API_END.

bvpvamsikrishna:

src/include/api.h
```diff
@@ -89,6 +89,15 @@
 #define	TXN_API_END(s, ret)	TXN_API_END_RETRY(s, ret, 1)
 
 /*
+ * In cursor open call we should not change the transaction state.
+ * If method is about to return WT_NOTFOUND (some underlying object was
+ * not found), map it to ENOENT.
+ */
+#define	CURSOR_OPEN_API_END(s, ret)					\
```
agorrod: It would be better if this had a generic name, something like: `API_END_RET_NO_TXN_ERROR`. That matches our other conventions:
`API_END` is a standard prefix
`_RET` is for macros that return directly
`_NO_TXN_ERROR` describes why it's special

It took me quite a while to spot the `0` being passed to `API_END` - I think a better name would make that clearer. There would be a corresponding reworking of the comment to match the generic macro.

test/fops/file.c
```diff
@@ -55,16 +55,7 @@ obj_bulk(void)
 	}
 
 	if (use_txn) {
-		/*
-		 * As the operations are being performed concurrently,
-		 * return value can be ENOENT, EBUSY or EINVAL will set
-		 * error to transaction opened by session. In these
-		 * cases the transaction has to be aborted.
-		 */
-		if (ret != ENOENT && ret != EBUSY && ret != EINVAL)
```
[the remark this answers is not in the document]
bvpvamsikrishna: Even if 'session->create' returns error, it will not set the transaction error. So, will not be forced to rollback.

bvpvamsikrishna:
@agorrod, I have updated for review comments. Request to check.

agorrod:
@bvpvamsikrishna This looks good to me - I pushed some changes to the comments, so they make complete sentences and don't reference local variable names. If it passes testing I'll go ahead and merge.

PR #3734, opened 2017-10-13
WT-3657 Revert transaction changes for lookaside updates.

michaelcahill:
Specifically, reverts most of WT-3645 (#3725).

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -490,7 +490,7 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
 				    &session->txn, WT_TXN_HAS_TS_READ) &&
 				    __wt_timestamp_cmp(
 				    &ref->page_las->min_timestamp,
-				    &session->txn.read_timestamp) > 0)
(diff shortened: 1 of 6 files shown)


michaelcahill:
@keithbostic, for the cases I can reproduce, reverting this change seems to help.

That is unfortunate, because the original change allowed lookaside pages to be evicted from cache when there is a long running transaction (causing `r->last_running` to be old in reconciliation so many updates appear uncommitted).  Let's point some tests at this branch, and if it helps, merge this change and find some other solution to allowing lookaside pages to be evicted from cache.

keithbostic:
@michaelcahill, the original change looks correct to me, I'm missing something. What is the problem in the original change that you think is causing failures? I'm hesitant to merge this without understanding what I'm changing.

michaelcahill:
Closing in favor of WT-3669: the issue isn't `WT_TXN_FIRST` vs `WT_TXN_NONE` (either should work, and if either doesn't, it's going to cause other problems.  What I'm concerned about is assumptions in code that transaction IDs only transition from some set value to `WT_TXN_ABORTED` (a large value).  There could be problems caused by having a transaction ID move backwards (from a set value to zero).

PR #3735, opened 2017-10-13
WT-3657 Use saved update size for splits, don't grow raw buffer.

michaelcahill:
Previously, we would split pages once reconciliation saw 100 saved updates.  Apart from suffering from the same problems as other naive split approaches (e.g., creating a new page with a single saved update each time a new insert plus eviction happens), the previous approach also doubled the size of the buffer passed to raw compression each time splits were triggered by the saved update count, leading to attempts to allocate unexpectedly large amounts of memory.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -2633,11 +2633,11 @@ __rec_split_crossing_bnd(
 }
 
 /*
- * __rec_split_raw_worker --
- *	Handle the raw compression page reconciliation bookkeeping.
+ * __rec_split_raw --
+ *	Raw compression.
  */
 static int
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, another change for your consideration: this one needs some careful thought, particularly with regards to the `split_grow` label change.  There are some cases in raw compression I'm not sure about so this may not be quite right.

OTOH, this does seem to avoid the large memory allocations we've seen in some format failures.

keithbostic:
@michaelcahill, I don't see this: in `__rec_split_raw_worker`, we don't jump to `split_grow` unless `__rec_need_split` returns false, so if a split was triggered by the saved update count, I don't see how we'd end up jumping to `split_grow`? Or is it we're ending up in the "we  need more rows" path?

PR #3738, opened 2017-10-13
WT-3662 Lookaside records can remain until the file is evicted.

keithbostic:


keithbostic:
@michaelcahill, the 3b38059 change is a minor standalone fix/cleanup based on reviewing WT-3663, I'm including it here because I'd already changed `__rec_split_write()`.

michaelcahill:
@keithbostic, I ran some workloads today against `develop` and this branch, and while I can reproduce some failure modes on develop, I didn't figure out why they aren't reproducing on this branch.

However, I've reviewed the change and I'm happy with it -- we don't want to "leak" overflow records if the eviction doesn't succeed.  I think the remaining issues on develop are connected somehow with eviction giving up but haven't yet nailed that down.

I'm going to merge so that this gets regular testing on develop and keep chasing why this change helps over the next day or two.

PR #3739, opened 2017-10-14
WT-3663 lookaside records ignored unless a backing disk block written

keithbostic:


michaelcahill:

For the record, while this path tried to cope with lookaside eviction, when WT-3435 first went in I don't think it could actually end up here (or at least it was very unlikely).  We only attempted lookaside when there was a single page image, which should have been handled in `evict_page.c` as a `WT_PM_REC_REPLACE`.  This fix is necessary, and I think more recent changes to splits make this path far more likely.

PR #3740, opened 2017-10-15
WT-3665 change format to configure the WiredTiger checkpoint thread

keithbostic:
@sueloverso, the reason I wanted a review from you is the library checkpoint configuration.

When configuring the library checkpoint thread, I'm waiting between 1 and 20 MB of log file between checkpoints (max 1024MB), and between 5 and 100 seconds (max 3600). Does that sound reasonable to you?

### DIFF
--- a/test/format/config.c
+++ b/test/format/config.c
@@ -29,6 +29,7 @@
 #include "format.h"
 #include "config.h"
 
+static void	   config_checkpoint(void);
 static void	   config_checksum(void);
 static void	   config_compression(const char *);
 static void	   config_encryption(void);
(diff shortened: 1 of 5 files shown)


keithbostic:
autobuild: retest this please

sueloverso:
Hmm, some of those values may be off but it depends on what you're trying to test.  The log file size is not modified by test/format, so it is using the default of 100Mb log files.  Checkpoints written more often than that do not help log archiving.  Having multiple checkpoints in a single log file don't allow us to archive sooner.  But there is no harm in it either.  I'd probably suggest 20-200 for checkpoint_log_size.  The wait range seems reasonable although I have seen configurations where the entire test finishes regularly in < 5 seconds.

keithbostic:
> I'd probably suggest 20-200 for checkpoint_log_size. 

Done in 57db489.

> The wait range seems reasonable although I have seen configurations where the entire test finishes regularly in < 5 seconds.

Now that format has `timer=`, I generally set `ops=0` and run a configuration for 3-5 minutes, I find that gives me better testing and skips runs that are simply too fast to do anything other than check the configuration.

keithbostic:
@sueloverso, we should randomly select a log file size, too, that way we'll test file archiving/rotation/whatever with a variety of sizes.

Added for your consideration in 7651bb2.

PR #3741, opened 2017-10-17
WT-3667 fix timestamp tests to backup with log archive set to false

sulabhM:


agorrod:
lgtm

agorrod:
Holding off merging - I had another thought that I added to the ticket.

sueloverso:
LGTM.

PR #3743, opened 2017-10-17
WT-3662 Write lookaside after reconciliation has succeeded.

keithbostic:


keithbostic:
@michaelcahill, I added an assert that the on-page update would be found in the list when instantiating the page (when I was debugging WT-3662 a few days ago), but apparently that's not always the case.

I'll revert that change so testing can proceed, but I don't see the path: in what case wouldn't we find the on-page update when re-instantiating the page?

PR #3744, opened 2017-10-18
WT-3652 Skip unnecessary lookaside reads / writes.

michaelcahill:
Also fix reads during checkpoint so the expected read generation is set, and checkpoint can attempt to clean up after itself.

Add a "lookaside_score" measuring the proportion of unstable updates in cache (those required for historic reads), use that to determine when to use the lookaside table rather than waiting for the cache to become stuck.


michaelcahill:
Not ready for review: in particular, we need more careful tracking than the current `r->max_txn` -- that just happens to be sufficient for the test case where no updates are visible.

In addition, I'll add the code as described in the ticket to skew lookaside eviction to use the most recent versions of updates when `txn_global->has_stable_timestamp` is false.  Checkpoints can figure out if eviction has been too eager when choosing a recent version to evict.

michaelcahill:
autobuild: retest this please

agorrod:
autobuild: retest this please

The last failure was a timeout on Windows - I'd like to know if it was a one-off.

agorrod:
@michaelcahill and @keithbostic  I merged the develop branch in here, because I wanted to try out an Evergreen patch build in preparation for maybe updating MongoDB master. The merge wasn't clean - there was a conflict with the fix you pushed in WT-3674. That block of code has been completely removed from this branch.
It would be great if you could confirm the merge was the right thing to do.
Please let me know if there is something I can do to push this change ahead during my day tomorrow.

agorrod:
@michaelcahill I made a little change in here - the changes look good to me. I'm going to go ahead and merge to get this into testing.

PR #3745, opened 2017-10-18
WT-3616 format failed to report a stuck cache

keithbostic:
@agorrod, there are two changes here.

48cc54c sets a flag when a page is read on behalf of checkpoint, and then doesn't count evicting that page towards eviction progress. I am hesitant to not count "normal" eviction of metadata or lookaside pages towards eviction progress, it's not impossible that those pages could consume a reasonable piece of the cache and evicting them could be progress. So instead, I don't count the eviction of pages read on behalf of a checkpoint as eviction progress. (To be clear, I'm not excited about this approach, and it's a special case I'd rather not have, I just couldn't think of anything I liked better.)

4d8c187 restructures `__evict_server()` to remove the `WT_CACHE.pages_evicted` field.  In summary, I think `WT_CACHE.pages_evicted` can be more replaced with comparing the value of `WT_CACHE.pages_evict` when `__evict_server()` is entered, against its current value when evaluating if work has been done. The original code was relatively tricky though, I might have missed something.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -304,7 +304,7 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref)
 	WT_ITEM tmp;
 	WT_PAGE *page;
 	size_t addr_size;
-	uint32_t new_state, previous_state;
+	uint32_t flags, new_state, previous_state;
 	const uint8_t *addr;
 	bool timer;
 
@@ -372,9 +372,11 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * the allocated copy of the disk image on return, the in-memory object
 	 * steals it.
 	 */
-	WT_ERR(__wt_page_inmem(session, ref, tmp.data,
-	    WT_DATA_IN_ITEM(&tmp) ?
-	    WT_PAGE_DISK_ALLOC : WT_PAGE_DISK_MAPPED, &page));
+	flags =
+	    WT_DATA_IN_ITEM(&tmp) ? WT_PAGE_DISK_ALLOC : WT_PAGE_DISK_MAPPED;
+	if (WT_SESSION_IS_CHECKPOINT(session))
+		LF_SET(WT_PAGE_CHECKPOINT_READ);
+	WT_ERR(__wt_page_inmem(session, ref, tmp.data, flags, &page));
 	tmp.mem = NULL;
 
 skip_read:
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -296,7 +296,7 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_REF *ref)
  *	Read a page from the file.
  */
 static int
-__page_read(WT_SESSION_IMPL *session, WT_REF *ref)
+__page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
 {
 	struct timespec start, stop;
 	WT_BTREE *btree;
@@ -304,7 +304,7 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref)
 	WT_ITEM tmp;
 	WT_PAGE *page;
 	size_t addr_size;
-	uint32_t flags, new_state, previous_state;
+	uint32_t page_flags, new_state, previous_state;
 	const uint8_t *addr;
 	bool timer;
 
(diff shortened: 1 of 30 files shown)


src/btree/bt_read.c
```diff
@@ -372,9 +372,11 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * the allocated copy of the disk image on return, the in-memory object
 	 * steals it.
 	 */
-	WT_ERR(__wt_page_inmem(session, ref, tmp.data,
-	    WT_DATA_IN_ITEM(&tmp) ?
-	    WT_PAGE_DISK_ALLOC : WT_PAGE_DISK_MAPPED, &page));
+	flags =
+	    WT_DATA_IN_ITEM(&tmp) ? WT_PAGE_DISK_ALLOC : WT_PAGE_DISK_MAPPED;
+	if (WT_SESSION_IS_CHECKPOINT(session))
+		LF_SET(WT_PAGE_CHECKPOINT_READ);
```
michaelcahill: This part doesn't work -- pages can be read as part of a checkpoint and stay in cache, so marking pages here isn't valid.  We could use `WT_SESSION_IS_CHECKPOINT` during the eviction if we want to exclude eviction of these pages for some reason.

That said, what was reported yesterday on the ticket was eviction of metadata triggering "progress" from the point of view of the eviction server, when nothing else could be evicted.  The problem with that is that we disable eviction checks when reading metadata, so metadata pages can be read back into cache when it is full.  That is why we need special treatment of metadata eviction (and any other eviction where reads disable ordinary eviction).

src/evict/evict_lru.c
```diff
@@ -411,46 +411,54 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 		__wt_readunlock(session, &conn->dhandle_lock);
 		WT_RET(ret);
 
-		cache->pages_evicted = 0;
```
agorrod: I'm concerned we lose something dropping this part of the change. I *believe* that this call to reset pages_evicted meant that we tracked if progress was being made between calls to the `__evict_server` function. That is valuable because the eviction server does this "gather candidates" pass, and then often pauses for a while or does some eviction between passes - which gives other threads a chance to do eviction.

keithbostic: Yes, I think you're correct -- I missed that.

My change would mean we only decide we "did work" if pages were pushed out (by any thread), during the time the thread executing `__evict_server()` started, and called `__evict_pass()`.

michaelcahill:
I also reinstated the tracking of eviction progress across calls to (the poorly named) `__evict_server` function.

michaelcahill:
@keithbostic, I think this branch is ready to go, if you agree, can you please merge?

keithbostic:
autobuild: retest this please

src/evict/evict_lru.c
```diff
@@ -377,13 +377,10 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 	WT_CACHE *cache;
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
-	uint64_t orig_pages_evicted;
 
 	conn = S2C(session);
 	cache = conn->cache;
-	WT_ASSERT(session, did_work != NULL);
 	*did_work = false;
```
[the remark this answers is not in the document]
michaelcahill: I've put it back the other way and done a survey: it was about 50/50 through the code whether we did local variable initialization or cleared output parameters first.

src/evict/evict_lru.c
```diff
@@ -411,46 +408,56 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 		__wt_readunlock(session, &conn->dhandle_lock);
 		WT_RET(ret);
 
-		cache->pages_evicted = 0;
-	} else if (cache->pages_evicted != cache->pages_evict) {
-		cache->pages_evicted = cache->pages_evict;
+		/* Make sure we'll notice next time we're stuck. */
+		cache->last_eviction_progress = 0;
+		return (0);
+	}
```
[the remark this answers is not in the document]
michaelcahill: Good catch, I've put `*did_work = true` back in there.  I see an obviously better test: this is the only code that tracks whether we are making long-term progress.

michaelcahill:
@keithbostic thanks for the review, I made those changes.  Merging...

PR #3746, opened 2017-10-18
WT-3673 opening the lookaside table can race with the statistics server

keithbostic:
@michaelcahill, @agorrod, I was thinking we might want to do something to avoid introducing this kind of race in the future. 

I was wondering about something similar to what we do on WiredTiger close with the `WT_CONN_CLOSING` flag. We could add a `WT_CONN_OPENING` flag that's initially set,
and change the servers to not execute their work loops until that flag is cleared?


keithbostic:
autobuild: retest this please

michaelcahill:
I've been leaving this alone because @agorrod was looking at it.  Neither of us has had much time for reviews or typing this week, so if this is still waiting for a review I'll take a closer look on Monday.

agorrod:
I preferred the idea of getting rid of the `CONN_LAS_CONFIGURED` flag - but when I started making that change I got to the code in `__wt_open_internal_session` and stopped making progress. The semantic in there seems very strange to me: if it happens to arrive before the lookaside session is opened the session will use the connection wide shared cursor, otherwise it will open a cursor in the session - I remember how we got there, and don't want to bite off a big change in ordering at startup now..
@keithbostic I'd be happy if you could take a run at removing the flag - if you can't get find a neat solution feel free to merge this and then open a new ticket for making this better in the future.

keithbostic:
@agorrod, @michaelcahill, I took another run at this one.

We can't get rid of `WT_SESSION_LOOKASIDE_CURSOR` because we have to have a way for checkpoint to turn off using a private lookaside cursor. (As an aside Michael, maybe we should open a separate checkpoint-only lookaside cursor so checkpoint doesn't compete with readers for the spin lock? I don't have a sense of how useful that would be, though.)

We could get rid of `WT_CONN_LOOKASIDE_OPEN`, but we'd have to order/publish when `WT_CONNECTION.las_session` gets set, and we'd want a macro to hide our check of that variable anyway, so I just don't see the point.

To simplify the whole thing and get rid of the magic in `__wt_open_internal_session`, I separated the creation of the lookaside cursor from the creation of the internal session handle, that way we don't have to set `WT_CONN_LOOKASIDE_OPEN` until everything is correctly configured, and that avoids racing with any other thread trying to get to the shared lookaside cursor.

Alex, back to you.

michaelcahill:
We briefly had a lookaside cursor in the checkpoint session: that turns out to be bad.  The reason is that the writes to the lookaside table then have the same transaction ID as the checkpoint, and hence can't be evicted until the checkpoint completes.

keithbostic:
> The reason is that the writes to the lookaside table then have the same transaction ID as the checkpoint, and hence can't be evicted until the checkpoint completes.

Even simpler, we could open two lookaside table sessions/cursors at startup, and assign one of them to the checkpoint thread whenever it runs. 

Again, this could be totally wasted effort, I don't have a sense of the contention we see for the shared lookaside spin lock. Do you?

One other not-well-thought-out thought: when we insert into the lookaside table, why couldn't we notice it is the lookaside table (WT_BTREE.flags), and set the transaction ID to anything we want? I was thinking about drilling a hole from reconciliation to the cursor insert code to do that a few weeks ago, when it occurred to me we could simply set the transaction ID in the row-modify code based on the underlying table information.

agorrod:
@keithbostic this change lgtm now - I'm going to go ahead and merge. Should we open a new ticket to capture the conversation about potential lookaside cursor management improvements, or wait for a perf issue to trigger the optmization?

keithbostic:
> Should we open a new ticket to capture the conversation about potential lookaside cursor management improvements, or wait for a perf issue to trigger the optimization?

Let's leave that to @michaelcahill: I'm happy to gather data (I can try to find a workload where giving checkpoint its own lookaside cursor matters, or we could instead add a statistic that tracks if/when the shared lookaside cursor isn't available to a thread).

Regardless, I'm relatively sure it's an easy change if Michael thinks any of these changes are worth doing.

michaelcahill:
@keithbostic, I'd like to get to a state where stress tests are running reliably, with lookaside and checkpoints behaving the way we want, before worrying about contention for that spin lock.  I haven't seen it as an issue in any of my testing: until WT-3652 is merged, it would be lost in the noise of constantly waiting because the cache is stuck.

As far as magic transaction IDs goes: we're already setting the transaction ID for lookaside writes on commit.  I think that's as far as we can go safely: if there is any chance of a rollback (and I think there is), then we can't use arbitrary transaction IDs -- we have to use something that is invisible until commit, or the update could be evicted in the meantime.

PR #3747, opened 2017-10-18
WT-3672 Test format failure with commit timestamp older than oldest timestamp

keithbostic:
FTR, I can't reproduce this one, but I think there is a way format can race.


sulabhM:
autobuild: retest this please

PR #3752, opened 2017-10-23
WT-3680 metadata unroll should discard in-process checkpoints

keithbostic:
Alex, I don't know what happened in BF-6928, but my guess is an operation did a checkpoint and then subsequently failed, and we don't clean up checkpoints when unrolling tracked operations.

I've added code to do that, and I've also added code to panic if we fail to successfully either apply or unroll any set of tracked operations.


agorrod:
Thanks @keithbostic lgtm.

PR #3753, opened 2017-10-24
WT-3681 Only truncate a log if we're less than the trunc_lsn.

sueloverso:
@michaelcahill Please review this change.  The real fix is checking if `rd_lsn` is less than the `trunc_lsn`.  I also fixed verbose messages.


sueloverso:
@michaelcahill or @agorrod - whoever gets to it first, Bruce has confirmed that this fixes the problem he saw.

michaelcahill:

PR #3754, opened 2017-10-24
WT-3683 Allow eviction of clean pages with history when cache is stuck.

michaelcahill:


michaelcahill:
@keithbostic, this is an opening gambit.  Let's talk about what we can live with...

michaelcahill:
Thanks @keithbostic.

The Windows test timed out but I don't believe that's related to this change.

I'm going to go ahead and merge to unblock stress tests.

PR #3755, opened 2017-10-24
WT-3677 test/format compaction doesn't handle timeout error return

keithbostic:


agorrod:
Thanks @keithbostic lgtm

PR #3758, opened 2017-10-25
WT-3693 Add a callback for when the all_committed timestamp changes.

michaelcahill:


michaelcahill:
Does not yet need review: being used to help diagnose test failures initially.

agorrod:
@michaelcahill where are we at with this branch? Should I go ahead and review?

michaelcahill:
That's really a question for @dgottlieb -- is there any value in this change for diagnosing issues with MongoDB's use of timestamps?  Otherwise, I'm happy to toss it.

dgottlieb:
I think we can toss it.

PR #3759, opened 2017-10-25
WT-3666 Don't write uncommitted updates during eviction for checkpoints.

michaelcahill:


michaelcahill:
Hold off reviewing these changes: I've seen `test_timestamp07` fail on this branch.

michaelcahill:
I think all of these changes are important, but the last commit (29d2e95) should fix the repeatable read failures we have been seeing (particularly in {{test_timestamp04}} and {{test_timestamp07}}).  Some commentary is in the jira ticket.

src/btree/bt_read.c
```diff
@@ -114,12 +114,25 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t btree_id)
 	 * updates for the next key, and so on.
 	 *
 	 * Search for the block's unique prefix, stepping through any matching
-	 * records.
+	 * records.  Because of the special visibility rules for lookaside, a
+	 * new block can appear in between our search and the block of
+	 * interest.  Keep trying until we find it.
```
keithbostic: I'd suggest updating both versions of this comment to note the lookaside block may not appear at all, and why.

If you're willing, I'd even abstract this code out to a `__wt_las_cursor_position` call, given that it's relatively obscure what's happening.

src/reconcile/rec_write.c
```diff
@@ -378,9 +378,16 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
 	 * Otherwise we would need to keep updates in memory that go back older
 	 * than the version in the disk image, and since modify operations
 	 * aren't idempotent, that is problematic.
+	 *
+	 * If we try to do eviction using transaction visibility, we had better
+	 * have a snapshot.  This doesn't apply to reconciliation: there are
+	 * (rare) cases where we write data at read-uncommitted isolation.
```
keithbostic: Should "reconciliation" in the comment be "checkpoint"?

keithbostic:
I'm still bothered by the fact we're using `WT_TXN_NONE` as both "there's no transaction here" and "there's a transaction here but it's globally visible".

I can't find anything that looks wrong, but I'd suggest defining `WT_TXN_VISIBLE` equal to `WT_TXN_FIRST` (or even give it a unique value), and use `WT_TXN_VISIBLE` instead in the places where we're assigning a transaction ID intended to mean globally visibility.

For example, we set lookaside transaction updates in `txn_commit()` to `WT_TXN_NONE`, but then in the reconciliation code, we have:
```
/* Track the oldest saved timestamp for lookaside. */
if (F_ISSET(r, WT_REC_LOOKASIDE))
        for (upd = first_upd; upd != NULL; upd = upd->next)
                if (upd->txnid != WT_TXN_ABORTED &&
                    upd->txnid != WT_TXN_NONE &&
                    __wt_timestamp_cmp(
                    &upd->timestamp, &r->min_saved_timestamp) < 0)
                        __wt_timestamp_set(
                            &r->min_saved_timestamp, &upd->timestamp);
```
I understand we can't do lookaside eviction on the lookaside file, but the intent here isn't to skip "globally visible updates".

michaelcahill:
@keithbostic, I'd be happy renaming `WT_TXN_NONE`, but I don't see any difference between "not owned by a transaction" and "globally visible".  If there is a difference, it's probably a bug.

In the case you mention in `__rec_txn_read`, I'm not sure how necessary it is to skip updates with `WT_TXN_NONE`: we could probably assert instead that we never see an update with a timestamp but no transaction ID.

keithbostic:
> but I don't see any difference between "not owned by a transaction" and "globally visible".

This is probably my unfamiliarity with the transaction code, just ignore the suggestion if you don't like it.

To restate: when I see things tests like `(*updp)->txnid == WT_TXN_NONE` it bothers me because I know some updates can legitimately have transaction IDs of `WT_TXN_NONE` that determine their visibility, and I start to try and figure out if/when the two cases overlap.

src/btree/bt_read.c
michaelcahill: Thanks, agreed -- done.

michaelcahill:
@keithbostic, 8f8f777 is the key here: in `test_las` we do a checkpoint with `stable_timestamp=1`, then follow up with another checkpoint with `stable_timestamp=10000`.  The first checkpoint skips over everything, so it marks the tree clean, then the second checkpoint never visits it.

I still couldn't catch where a page was becoming dirty without the tree (it's hard to catch because we allow that temporarily during checkpoints).  That change fixes things in my testing, though, and I've made changes based on your feedback.  I'll remove or explain that `txnid == WT_TXN_NONE` check as part of WT-3652.

I think this branch is ready for a final review.

michaelcahill:
Talk to @keithbostic offline, got hit lgtm.  Merging...

PR #3760, opened 2017-10-26
WT-3696 Add checks to ensure session usage is single threaded.

agorrod:
This currently fails with at least test/fops - that will need to be fixed before the branch is merged.


keithbostic:
We document sessions cannot be used **concurrently** by multiple threads, not that multiple threads can't share sessions. Won't this trigger for users like MongoDB, that cache sessions and hand them out to threads on demand?

Wouldn't it be enough to check if any session crossing the API boundary is currently inside the API?

agorrod:
bq. Won't this trigger for users like MongoDB, that cache sessions and hand them out to threads on demand?

I don't believe so - the thread ID is cleared on exit, so a different thread can be the next one to call the API.

bq. Wouldn't it be enough to check if any session crossing the API boundary is currently inside the API?

The difficulty with that is that we call API methods internally - so we don't currently have a clean way to determine what the API boundary is. Ensuring all nested API calls are from the same thread ID was my attempt at dealing with that.

I'll add some documentation if there isn't  push back against the general idea.

keithbostic:
bq. I don't believe so - the thread ID is cleared on exit, so a different thread can be the next one to call the API.

I see now -- agreed, lgtm.

dist/s_define.list
```diff
@@ -2,6 +2,8 @@
 API_CALL
 API_CALL_NOCONF
 API_SESSION_INIT
+DIAG_SINGLE_THREAD_CHECK_START
+DIAG_SINGLE_THREAD_CHECK_STOP
```
keithbostic: This is outside of our documented namespace (so are the API_XXX and TXN_XXX macros, but we shouldn't make the problem worse).

Might be worth #undef of the API_XXX and TXN_XXX macros while you're there.

src/include/session.h
```diff
@@ -96,6 +96,12 @@ struct __wt_session_impl {
 	size_t	  scratch_cached;	/* Scratch bytes cached */
 #ifdef HAVE_DIAGNOSTIC
 	/*
+	 * Variables used to look for violations of the contract that a
+	 * session is only used by a single session at once.
+	 */
+	volatile size_t api_tid;
+	volatile size_t api_enter_refcnt;
```
keithbostic: The thread-ID should probably be `uintmax_t`, not `size_t`, and the reference count should be a `u_int`?

(Neither of them contain the size of an object in memory.)

keithbostic: I don't think `volatile` is necessary.

If you want to be sure you catch threads racing, my guess is you want a publish when the thread-ID is set, does that sound correct to you?

dist/s_define.list
sueloverso: I don't see any compelling reason this is outside the `WT_` namespace.  I suggest `WT_DIAG_THREAD_CHECK_*`

src/os_posix/os_thread.c
```diff
@@ -67,10 +67,32 @@ __wt_thread_join(WT_SESSION_IMPL *session, wt_thread_t tid)
...
+#endif
+}
+
+/*
+ * __wt_thread_descriptor --
  *	Fill in a printable version of the process and thread IDs.
  */
 int
-__wt_thread_id(char *buf, size_t buflen)
+__wt_thread_descriptor(char *buf, size_t buflen)
```
sueloverso: Very minor nit - I'd prefer `__wt_thread_str` because to me the word `descriptor` makes me think of a resource like a file descriptor that we need to keep track of or release, etc.

src/os_posix/os_thread.c
```diff
@@ -67,10 +67,32 @@ __wt_thread_join(WT_SESSION_IMPL *session, wt_thread_t tid)
...
+	/*
+	 * POSIX 1003.1 allows pthread_t to be an opaque type; on systems where
+	 * it's a pointer, print the pointer to match gdb output.
+	 */
+	self = pthread_self();
+#ifdef __sun
+	*id = (size_t)self;
+#else
+	*id = (size_t)(void *)self;
+#endif
```
sueloverso: Are we sure this is going to work for all systems?  I recall the entire reason and history of why we ended up with `__wt_thread_id` as a string was because this sort of thing wasn't valid everywhere (maybe Mac?).  If you haven't I suggest looking into the historical ticket to make sure this is okay.

src/include/session.h
agorrod: I added volatile because I didn't want the compiler to optimize out any code related to the variables. I think that's valid - I don't know specifically how the compiler could avoid re-reading a variable (which is the canonical example where volatile is useful), but my understanding is that it's best to mark any variable shared across threads as volatile.

I think a publish might be necessary - without it the macro might trigger a false positive, I didn't add it because I figured that any application that shares session structures amongst threads should have an explicit protocol in place to hand out sessions, so not protecting this assertion could help uncover application usage problems as well.

dist/s_define.list
agorrod: I've created WT-3706 for the more general change here, and fixed my particular violation.

agorrod:
@sueloverso Would you mind taking another look at this?

src/include/session.h
```diff
@@ -96,6 +96,12 @@ struct __wt_session_impl {
 	size_t	  scratch_cached;	/* Scratch bytes cached */
 #ifdef HAVE_DIAGNOSTIC
 	/*
+	 * Variables used to look for violations of the contract that a
+	 * session is only used by a single session at once.
+	 */
+	volatile uintmax_t api_tid;
+	volatile u_int32_t api_enter_refcnt;
```
keithbostic: The `u_int32_t` should be a `uint32_t`?

dist/s_define.list
```diff
@@ -22,6 +22,8 @@ WT_CACHE_LINE_PAD_END
 WT_CONN_CHECK_PANIC
 WT_DEADLOCK
 WT_DEBUG_BYTE
+WT_DIAG_SINGLE_THREAD_CHECK_START
+WT_DIAG_SINGLE_THREAD_CHECK_STOP
```
sueloverso: Minor nit - this is a really long name and `DIAG` and `CHECK` are somewhat redundant in this context.

src/include/api.h
```diff
@@ -6,12 +6,38 @@
...
+	WT_ASSERT(session, (s)->id == 0 || (s)->api_tid == 0 ||		\
+	    (s)->api_tid == __tmp_api_tid);				\
+	if ((s)->api_tid == 0)						\
+		(s)->api_tid = __tmp_api_tid;				\
+	++(s)->api_enter_refcnt;					\
+	}
+
+#define	WT_DIAG_SINGLE_THREAD_CHECK_STOP(s)				\
+	if (--(s)->api_enter_refcnt == 0)				\
+		(s)->api_tid = 0;
```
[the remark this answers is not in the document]
sueloverso: What could be reordered with this conditional?  Would it reorder as in check if `api_enter_refcnt` is 1, set `api_tid` and then decrement `api_enter_refcnt`?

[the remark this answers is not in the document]
keithbostic: It's not the specific reordering in the conditional I'm worried about -- we don't want to do any of the API work until after the session's ID has been updated, and absent a barrier instruction there's no guarantee that will happen. Ditto exiting, we want to make sure all API work has been done before we clear the session's ID, which requires a barrier.

agorrod:
Thanks for the feedback. @keithbostic sorry I forgot to add barriers when I made the review changes - I wasn't trying to be sneaky. I didn't get back to this today, but I'll plan to do it tomorrow.

agorrod:
@sueloverso Would you mind taking another look at this one?

sueloverso:
LGTM, merged.

PR #3761, opened 2017-10-26
WT-3695 format failed to report a stuck cache

keithbostic:
@michaelcahill, a branch for your consideration, to catch another case where we're timing out rather than reporting cache stuck.

The workload in WT-3695 fails to report eviction is stuck because we're managing to write root pages and lookaside pages created during checkpoints, neither of which are actual progress, those pages are created in-memory.

Because there are a bunch of pages we create in memory (root or internal page splits, insert splits, root page reconciliations, and rebalance or salvage operations), I flipped the way we check for eviction progress. Instead of flagging if page eviction **doesn't** count as eviction progress, I flag if page eviction **should** count as eviction progress. It seems to me that if eviction progress isn't tied to page counts, it's safer and simpler to do it that way.

I'm only counting eviction of pages we read without eviction disabled, or pages we created from those pages, as eviction progress. That seems right: if we haven't evicted a page we've read in 15 minutes, we're either not reading pages at all, or we're kind of done.

But... what about insert-only workloads, where all we're doing is appending new items and internally splitting and evicting those pages? I could set "eviction progress" on in-memory split pages as a special case (but exclude lookaside table inserts?), as long as there aren't a bunch more special cases in the future.

Thoughts?


keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill, I don't much like 87d68a0, can you think of a better way to spot lookaside pages we're creating in memory and then evicting?

agorrod:
Thanks @keithbostic these changes do look better to me now - though I know @michaelcahill has been thinking about this code a lot more than me recently, so it's better if he does a final review.

src/evict/evict_page.c
```diff
@@ -296,6 +296,15 @@ __evict_page_dirty_update(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
...
 	WT_ASSERT(session, ref->addr == NULL);
 
+	/*
+	 * Don't count discarding a modified lookaside table page as eviction
+	 * making progress. This isn't precise, but what we're trying to spot
+	 * is active lookaside pages created in reconciliation, as opposed to
+	 * lookaside pages read to satisfy normal cursor operations.
+	 */
+	if (F_ISSET(S2BT(session), WT_BTREE_LOOKASIDE))
+		F_SET_ATOMIC(ref->page, WT_PAGE_EVICT_NO_PROGRESS);
```
agorrod: @keithbostic @michaelcahill and I talked about this branch, and if we take ignoring lookaside pages out we would be happy to merge the branch. It seems to us as though lookaside pages take up space in cache, so evicting them reclaims that space..

Do you remember if this part of the change was important in avoiding undetected stalled caches?

keithbostic: >  if we take ignoring lookaside pages out we would be happy to merge the branch. It seems to us as though lookaside pages take up space in cache, so evicting them reclaims that space. Do you remember if this part of the change was important in avoiding undetected stalled caches?

I recall that it was important to avoiding stalled caches, but lookaside has changed a lot since then, I'm happy to merge without that change and see where we end up. Pushed in a0f76dd.

PR #3762, opened 2017-10-26
WT-3699 support minimum cache sizes in format

keithbostic:


test/format/config.h
```diff
@@ -101,6 +101,10 @@ static CONFIG c[] = {
 	  "size of the cache in MB",
 	  0x0, 1, 100, 100 * 1024, &g.c_cache, NULL },
 
+	{ "cache_minimum",
+	  "minimum size of the cache in MB",
+	  C_IGNORE, 1, 100, 100 * 1024, &g.c_cache_minimum, NULL },
```
michaelcahill: Could we fix this to 20 rather than rolling the dice from 1-100?

keithbostic: > Could we fix this to 20 rather than rolling the dice from 1-100?

The `C_IGNORE` means `cache_minimum` is ignored unless it's set, that is, format won't roll dice at all. The "1" means you can't set `cache_minimum` to less than 1MB, and I've pushed a change of the "100" to "0" in 8292e2a, for clarity.

keithbostic:
autobuild: retest this please

agorrod:
Thanks @keithbostic lgtm

PR #3763, opened 2017-10-27
Revert "WT-3652 Skip unnecessary lookaside reads / writes."

agorrod:
Reverts wiredtiger/wiredtiger#3744


agorrod:
I'm going to revert this change - it's causing test failures.

PR #3765, opened 2017-10-28
WT-3705 Full build Friday and lint

keithbostic:
@agorrod, this is all random cleanup along with clang-tidy and clang50 warnings.

There are no bug fixes at all, the branch can be ignored as long as you like.


agorrod:
lgtm

PR #3766, opened 2017-10-30
WT-3630 Send the test environment to "make check".

michaelcahill:
(It also invokes Python to sanitize the build).


ddanderson:
lgtm.

PR #3769, opened 2017-10-30
WT-3707 Install signal handler to detect child unexpectedly failing.

sueloverso:
@keithbostic Please review this change to `timestamp_abort` to install a signal handler.

### DIFF
--- a/test/csuite/timestamp_abort/main.c
+++ b/test/csuite/timestamp_abort/main.c
@@ -84,6 +84,8 @@ static uint64_t th_ts[MAX_TH];
     "transaction_sync=(enabled,method=none)"
 #define	ENV_CONFIG_REC "log=(archive=false,recover=on)"
 
+static void handler(int)
+    WT_GCC_FUNC_DECL_ATTRIBUTE((noreturn));
 static void usage(void)
     WT_GCC_FUNC_DECL_ATTRIBUTE((noreturn));
 static void
@@ -434,9 +436,23 @@ timestamp_build(void)
 extern int __wt_optind;
 extern char *__wt_optarg;
 
+static void
+handler(int sig)
+{
+	pid_t pid;
+
+	WT_UNUSED(sig);
+	pid = wait(NULL);
+	/*
+	 * The core file will indicate why the child exited. Choose EINVAL here.
+	 */
+	testutil_die(EINVAL, "Child process %d abnormally exited", pid);
+}
+
 int
 main(int argc, char *argv[])
 {
+	struct sigaction sa;
 	struct stat sb;
 	FILE *fp;
 	WT_CONNECTION *conn;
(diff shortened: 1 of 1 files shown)


test/csuite/timestamp_abort/main.c
```diff
@@ -536,6 +552,9 @@ main(int argc, char *argv[])
 		 * kill the child, run recovery and make sure all items we wrote
 		 * exist after recovery runs.
 		 */
+		memset(&sa, 0, sizeof(sa));
+		sa.sa_handler = handler;
+		sigaction(SIGCHLD, &sa, NULL);
```
keithbostic: Why did we need sigaction(2), wasn't signal(3) sufficient?

test/csuite/timestamp_abort/main.c
```diff
@@ -434,9 +436,23 @@ timestamp_build(void)
...
+handler(int sig)
+{
+	pid_t pid;
+
+	WT_UNUSED(sig);
+	pid = wait(NULL);
+	/*
+	 * The core file will indicate why the child exited. Choose EINVAL here.
+	 */
+	testutil_die(EINVAL, "Child process %d abnormally exited", pid);
```
keithbostic: `pid` isn't necessarily an int, so at least it needs casting, and I'd go with a  `u_int` (and `uintmax_t` is for the rigorous).

keithbostic: Why bother waiting, do we care about the child pid, given that it just exited?

test/csuite/timestamp_abort/main.c
sueloverso: Every man page for `signal` says that to be portable you should use `sigaction` instead.  Here's what it says on margay:
```
 The  behavior  of  signal() varies across UNIX versions, and has also varied historically across different versions of Linux.  Avoid its use: use sigaction(2) instead.  See Portability below.
```

test/csuite/timestamp_abort/main.c
sueloverso: If the parent doesn't call `wait` then the child process remains a zombie.  I'll fix the output message.

sueloverso:
I added a similar fix from WT-3672 to publish the timestamp after the commit so that it isn't re-ordered.  This branch now solves both issues in the ticket.

test/csuite/timestamp_abort/main.c
keithbostic: Huh... learn something every day...

test/csuite/timestamp_abort/main.c
keithbostic: > then the child process remains a zombie. I'll fix the output message.

How is that possible, doesn't the init(8) process (or whatever passes for init these days) eventually do the wait?

sueloverso: The init process would become the parent if the parent process exits.  But the child process remains a zombie upon exit so that the parent can call wait to get its exit status.  If this parent process knows the child has exited, I think it should call wait/waitpid so the child can be properly cleaned up.

PR #3770, opened 2017-10-31
WT-3705 Full build Friday and lint

agorrod:
Reverts wiredtiger/wiredtiger#3765


agorrod:
This change introduced a significant performance regression - reverting it until the root cause is identified.

Note: follow on work in WT-3708 was removed due to the revert. It should be re-instated when this branch is revived.

PR #3774, opened 2017-11-01
WT-3717 Add a verbose lookaside mode

agorrod:
This required removing the temporary lookaside option, since we are about to run out of bits in verbose flags.


agorrod:
For some reason pull request testing is running slowly on this branch. We should do performance testing before we decide to merge.

sueloverso:
@agorrod I'd bet good money that PR testing is running slowly because of the change to `timestamp_abort`.  There already is code just a couple lines above your change that copies just the log files.  You added code to copy the entire database (copying the log files a second time).  The tables that get generated a often very large and slow to copy, which is why I only copied the log files in the debug code I added a while ago.

FYI, I for one use the `VERB_TEMPORARY` all the time in my debugging.  It is intentional that it looks "unused" in the code base.

test/csuite/timestamp_abort/main.c
```diff
@@ -605,6 +605,11 @@ main(int argc, char *argv[])
 	    "cp -p WiredTigerLog.* ../%s.SAVE",
 	     home, home, home));
 	(void)system(buf);
+	testutil_check(__wt_snprintf(buf, sizeof(buf),
+	    "rm -rf ../%s.SAVE_FULL && mkdir ../%s.SAVE_FULL && "
+	    "cp -p * ../%s.SAVE_FULL",
+	     home, home, home));
+	(void)system(buf);
```
sueloverso: Before merging this should be commented or ifdefed out.  I'm good leaving the code in, with a comment stating that it will be slow, tables are very, very large and many non-archived log files are created.  It will cause PR testing to go very slowly. That way if/when the next problem surfaces someone can just reinstate this for debugging rather than redoing it every time.  But it should not be part of the normal code.

agorrod:
@sueloverso I've gone through and cleaned this change up. I took a quick pass at including WT-3716 here, which would have let me keep the WT_VERB_TEMPORARY but it grew bigger than I was comfortable with. I've added a comment to WT-3716, which I'd like to get resolved ASAP.

Would you be OK merging this in the mean time, otherwise there are going to be some merge conflicts to resolve.

PR #3775, opened 2017-11-01
WT-3713 Make error output more concise.  Some refactoring.

sueloverso:
@michaelcahill Here's a fix that makes the output from timestamp_abort more concise and readable.


michaelcahill:

PR #3777, opened 2017-11-02
WT-3715 Lookaside eviction tuning.

michaelcahill:
Multiple changes aimed at improving performance and decreasing stalls
when applications keep more history than fits in cache.

* Change the order of fields in lookaside keys so that writes to lookaside are mostly appends.

* Allow splits when writing to lookaside.  This involved breaking a "NO_EVICTION" flag into two parts: (1) ignore the cache size, and (2) avoid splits.  There is a further "NO_RECONCILE" flag to prevent re-entry into reconciliation if writing to the lookaside table triggers eviction.

* Support multiple lookaside sessions / cursors simultaneously (initially 5).

* Don't count lookaside pages as part of the dirty content in cache. This means lookaside eviction unblocks operations more quickly (rather than moving dirty data from one page to another).  Prefer to evict lookaside pages as soon as possible unless operations are blocked by the dirty limit in cache.

* Only sweep the lookaside table when dropping a table that had lookaside entries.  There can still be significant pauses during drop, so we should consider moving this work into the sweep thread and making it incremental.

* Add statistics that indicate the range of pinned timestamps.


michaelcahill:
@agorrod, as discussed, here are a bunch of changes I've been testing.  There probably isn't much point doing a detailed review at this point, but if/when we get good results, we can sift through these changes to decide what to keep.

sueloverso:
FYI, I went to test your latest change, 2f5b368, against the insert test. Mongod hangs when I run `rs.initiate()`.  If I run with the previous commit, 533fb0b, it all works fine.

michaelcahill:
Thanks @sueloverso, unfortunately commit 2f5b368 contained a bug, since fixed.  I've been running more tests and can reproduce the performance regression.  I've made a little more progress but I've got more work to do, so I don't think there's anything else to do here for now.

michaelcahill:
@sueloverso, I think these changes are working as expected now.  Can you please take a look?

src/cache/cache_las.c
```diff
@@ -62,13 +88,15 @@ __wt_las_stats_update(WT_SESSION_IMPL *session)
 int
 __wt_las_create(WT_SESSION_IMPL *session)
 {
+	WT_CACHE *cache;
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
-	uint32_t session_flags;
 	const char *drop_cfg[] = {
 	    WT_CONFIG_BASE(session, WT_SESSION_drop), "force=true", NULL };
+	int i;
```
sueloverso: I think that is supposed to be listed above 'const char *'.

src/cache/cache_las.c
```diff
@@ -185,9 +225,10 @@ __wt_las_cursor_open(WT_SESSION_IMPL *session)
  */
 void
 __wt_las_cursor(
-    WT_SESSION_IMPL *session, WT_CURSOR **cursorp, uint32_t *session_flags)
+	WT_SESSION_IMPL *session, WT_CURSOR **cursorp, uint32_t *session_flags)
```
sueloverso: KNF.  I think the original was properly 4 spaces indented.

src/cache/cache_las.c
```diff
@@ -336,12 +405,14 @@ __las_insert_block_verbose(WT_SESSION_IMPL *session, WT_MULTI *multi)
  *	Copy one set of saved updates into the database's lookaside buffer.
  */
 int
-__wt_las_insert_block(WT_SESSION_IMPL *session,
-    WT_PAGE *page, WT_CURSOR *cursor, WT_MULTI *multi, WT_ITEM *key)
+__wt_las_insert_block(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
+	WT_PAGE *page, WT_MULTI *multi, WT_ITEM *key)
```
sueloverso: KNF, this also looks like it is a tab instead of 4 spaces?

src/cache/cache_las.c
```diff
@@ -536,26 +642,232 @@ __wt_las_remove_block(WT_SESSION_IMPL *session,
...
+	WT_ASSERT(session, cursor->session == &session->iface);
+
+	/*
+	 * If we're starting a new sweep, gather the list of trees to sweep.
+	 *
+	 * Otherwise, when continuing a sweep, position the cursor using the
+	 * key from the last call (we don't care if we're before or after the
+	 * key, either side is fine).
+	 */
+	if (key->size == 0) {
```
sueloverso: The awkward nested if could be re-written as:
/*
 * If we are continuing a sweep, position the cursor...
 * Otherwise start a new sweep and gather the list of trees to sweep.
 */
if (key->size != 0) {
 <continue sweep code>
} else if ((ret == __las_sweep_init(session)) != 0)
   goto srch_notfound;

src/evict/evict_lru.c
```diff
@@ -606,6 +607,23 @@ __evict_update_work(WT_SESSION_IMPL *session)
 		F_SET(cache, WT_CACHE_EVICT_SCRUB);
 
 	/*
+	 * Try lookaside evict when:
+	 * (0) not scrubbing (writing to lookaside during a checkpoint is
+	 * likely to slow the checkpoint down).
+	 * (1) the cache is stuck; OR
+	 * (2) the cache is more than half way from the dirty target to the
+	 *     dirty trigger; and
+	 * (3) the lookaside score goes over 80.
```
sueloverso: Nit: Swap (2) and (3) since they're in that order in the conditional.

src/btree/bt_read.c
```diff
@@ -684,9 +696,16 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
...
+			 * a blanket ban on checking the cache size in
+			 * sessions, but still require a transaction (e.g.,
+			 * when updating metadata or lookaside).  Make sure we
+			 * start a transaction if the session flag is set.
 			 */
-			return (LF_ISSET(WT_READ_NO_EVICT) ? 0 :
-			    __wt_txn_autocommit_check(session));
+			return (LF_ISSET(WT_READ_IGNORE_CACHE_SIZE) &&
+			    !F_ISSET(session, WT_SESSION_IGNORE_CACHE_SIZE) ?
+			    0 : __wt_txn_autocommit_check(session));
```
sueloverso: I think the comment needs more. It took me a while to think through this.  My interpretation ended up something like:
If WT_READ_IGNORE_CACHE_SIZE was passed in explicitly we're done.  If we set WT_READ_IGNORE_CACHE_SIZE because it was set in the session (e.g. ...) then make sure we start a transaction.

michaelcahill:
Thanks @sueloverso, all done in 994712f.

dist/flags.py
```diff
@@ -136,12 +138,12 @@
...
         'SESSION_LOGGING_INMEM',
         'SESSION_LOOKASIDE_CURSOR',
-        'SESSION_NO_CACHE',
         'SESSION_NO_DATA_HANDLES',
-        'SESSION_NO_EVICTION',
+        'SESSION_NO_RECONCILE',
         'SESSION_NO_LOGGING',
         'SESSION_NO_SCHEMA_LOCK',
         'SESSION_QUIET_CORRUPT_FILE',
+        'SESSION_READ_WONT_NEED',
```
michaelcahill: I'm happy to talk through flag naming: we had `NO_EVICTION` sprinkled liberally through the codebase, leading to many paths ignoring the cache size and unable to split trees normally.  This renaming attempts to clarify what happens when `NO_EVICTION` is set.

src/cache/cache_las.c
```diff
@@ -86,16 +114,17 @@ __wt_las_create(WT_SESSION_IMPL *session)
...
 	 */
-	session_flags = WT_SESSION_NO_EVICTION;
-	WT_RET(__wt_open_internal_session(
-	    conn, "lookaside table", true, session_flags, &conn->las_session));
-	WT_RET(__wt_las_cursor_open(conn->las_session));
+	for (i = 0; i < WT_LAS_NUM_SESSIONS; i++) {
+		WT_RET(__wt_open_internal_session(conn, "lookaside table",
+		    true, WT_LAS_SESSION_FLAGS, &cache->las_session[i]));
+		WT_RET(__wt_las_cursor_open(cache->las_session[i]));
+	}
```
michaelcahill: This now opens 5 sessions for lookaside access rather than the 1 we had before.

We could open these lazily, I think (the first time lookaside is required).  Worth doing?

src/evict/evict_lru.c
```diff
@@ -1632,6 +1650,28 @@ __evict_walk_file(WT_SESSION_IMPL *session,
...
+	 * on the dirty limit.  Once application threads are stalled by the
+	 * dirty limit, don't take any lookaside pages unless we're also up
+	 * against the total cache size limit.
+	 */
+	if (F_ISSET(btree, WT_BTREE_LOOKASIDE)) {
+		if (!F_ISSET(cache, WT_CACHE_EVICT_DIRTY_HARD))
+			target_pages = remaining_slots;
+		else if (!F_ISSET(cache, WT_CACHE_EVICT_CLEAN_HARD))
+			target_pages = 0;
+	}
```
michaelcahill: This is subtle and took a few attempts to get to this version.  It might need discussion and/or a longer comment.

src/include/btree.i
```diff
@@ -189,7 +190,7 @@ __wt_cache_decr_check_size(
  */
 static inline void
 __wt_cache_decr_check_uint64(
-    WT_SESSION_IMPL *session, uint64_t *vp, size_t v, const char *fld)
+    WT_SESSION_IMPL *session, uint64_t *vp, uint64_t v, const char *fld)
```
michaelcahill: This was a weird lurking bug I bumped into.  Please check that this was the intent of this function.

src/include/btree.i
```diff
@@ -261,7 +262,7 @@ __wt_cache_page_byte_dirty_decr(
 		    decr, "WT_BTREE.bytes_dirty_intl");
 		__wt_cache_decr_check_uint64(session, &cache->bytes_dirty_intl,
 		    decr, "WT_CACHE.bytes_dirty_intl");
-	} else if (!btree->lsm_primary) {
+	} else if (!btree->lsm_primary && !F_ISSET(btree, WT_BTREE_LOOKASIDE)) {
```
michaelcahill: This needs careful thought.  With this change, lookaside pages no longer contribute to the "dirty bytes in cache".  My thinking:

1. lookaside isn't involved in checkpoints, so from that point of view doesn't contribute to the work checkpoints have to do; and
2. evicting a (dirty) page by writing entries to lookaside should contribute to unblocking operations, rather than requiring a second eviction of the lookaside page(s) before the dirty bytes in cache goes down.

That said, this goes against some of the purpose of the 20% limit on dirty bytes in cache, which was to limit the impact of memory allocation fragmentation from lots of small allocations.

src/cache/cache_las.c
```diff
@@ -212,12 +252,30 @@ __wt_las_cursor(
...
+			}
+			__wt_spin_unlock(session, &cache->las_lock);
+			if (*cursorp != NULL)
+				break;
+			/*
+			 * If all the lookaside sessions are busy, stall.
+			 *
+			 * XXX better as a condition variable.
+			 */
+			__wt_sleep(0, 1000);
```
michaelcahill: I haven't seen threads waiting here in my testing so far, but it's workload-dependent and fixed sleeps are always bad for some case.  Maybe a followup ticket?

src/cache/cache_las.c
```diff
@@ -536,26 +642,232 @@ __wt_las_remove_block(WT_SESSION_IMPL *session,
...
+		 * from live files that have aged out.  If we track for each
+		 * entry whether it was the on-page value chosen by
+		 * reconciliation, we can safely remove entries from that point
+		 * on (for the given key) that are visible to all readers.
+		 */
+		if (__bit_test(cache->las_sweep_dropmap,
+		    las_id - cache->las_sweep_dropmin)) {
+			WT_ERR(cursor->remove(cursor));
+			++remove_cnt;
+		}
```
michaelcahill: Note the TODO above: I'd like a followup ticket to follow close after this.

agorrod: Done: WT-3737

dist/stat_data.py
```diff
@@ -453,6 +453,8 @@ def __init__(self, name, desc, flags=''):
     TxnStat('txn_pinned_checkpoint_range', 'transaction range of IDs currently pinned by a checkpoint', 'no_clear,no_scale'),
     TxnStat('txn_pinned_range', 'transaction range of IDs currently pinned', 'no_clear,no_scale'),
     TxnStat('txn_pinned_snapshot_range', 'transaction range of IDs currently pinned by named snapshots', 'no_clear,no_scale'),
+    TxnStat('txn_pinned_timestamp', 'transaction range of timestamps currently pinned', 'no_clear,no_scale'),
+    TxnStat('txn_pinned_timestamp_oldest', 'transaction range of timestamps pinned by the oldest timestamp', 'no_clear,no_scale'),
```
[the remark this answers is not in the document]
michaelcahill: They're in "timestamp" units -- i.e., however the application updates timestamps.  With MongoDB, the top 32-bits is a time and the low 32-bits are a counter, so when there is lag, we typically see 2**32 units per second, which `t2` interprets as steps of 4 `G`.

That isn't super helpful, but I don't know how to calculate something better here.

src/cache/cache_las.c
sueloverso: Do we need to think about adjusting `WT_EXTRA_INTERNAL_SESSIONS`?  That is currently 20, but with potentially many LSM threads, eviction threads, several logging threads, and others and now 5 of these, we could be encroaching on that.

agorrod: Good question - I've opened WT-3738 to address it.

PR #3779, opened 2017-11-03
WT-3716 Out of bits for verbose flags

keithbostic:
Make 64 bits available for verbose flags.


src/include/connection.h
```diff
@@ -417,7 +417,40 @@ struct __wt_connection_impl {
...
+#define	WT_VERB_SHARED_CACHE		0x0000000000800000
+#define	WT_VERB_SPLIT			0x0000000001000000
+#define	WT_VERB_TEMPORARY		0x0000000002000000
+#define	WT_VERB_THREAD_GROUP		0x0000000004000000
+#define	WT_VERB_TIMESTAMP		0x0000000008000000
+#define	WT_VERB_TRANSACTION		0x0000000010000000
+#define	WT_VERB_VERIFY			0x0000000020000000
+#define	WT_VERB_VERSION			0x0000000040000000
+#define	WT_VERB_WRITE			0x0000000080000000
+	uint64_t verbose;
```
sueloverso: You're missing the change in `misc.h` to `WT_VERBOSE_ISSET` to use `FLD64_ISSET`.

keithbostic:
I'm not sure where this work will end up, but we're done with this branch.

PR #3780, opened 2017-11-07
WT-3730 For simple tables, do not use table dhandle after it is released.

ddanderson:
Reverts "WT-3555 Streamline open_cursor for simple tables. (#3637)"
This reverts commit 982121f68b336eb58edfbae26300ef46935277b8.

### DIFF
--- a/src/cursor/cur_std.c
+++ b/src/cursor/cur_std.c
@@ -570,8 +570,7 @@ __wt_cursor_close(WT_CURSOR *cursor)
 	__wt_buf_free(session, &cursor->value);
 
 	__wt_free(session, cursor->internal_uri);
-	if (!F_ISSET(cursor, WT_CURSTD_URI_SHARED))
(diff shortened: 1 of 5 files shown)


ddanderson:
@michaelcahill or @agorrod , could you review this?  I think it may be the cause of HELP-5263.

ddanderson:
@agorrod, @michaelcahill, I backed out the change because I didn’t see a straightforward fix that preserves the optimization.  I see ways to preserve the optimization (by holding onto the table DHANDLE), but it is either a little messy (putting a `dhandle_to_release` in the file cursor) or too involved (perhaps creating a new "simpletable:" uri that handles this simple case).  Maybe you have some ideas? According to my previous testing, this change gained about ~ 20% of the time spent in opening a cursor. But if we end up going all in for cursor caching, this should be a lot less important.

michaelcahill:

PR #3781, opened 2017-11-08
WT-3648 add debug information

sueloverso:
@agorrod Here's a debug branch for the timestamp problem.  I have not been able to reproduce it (with or without this change). Since you were able to reproduce it locally, can you try this branch?  I added a couple asserts and record what value we set as the oldest and what thread had it on each iteration.  Next step might be to copy the th_ts array into the debug structure so we know what it looked like at the time we set the oldest.

Not for merge.

### DIFF
--- a/test/csuite/timestamp_abort/main.c
+++ b/test/csuite/timestamp_abort/main.c
@@ -28,6 +28,7 @@
 
 #include "test_util.h"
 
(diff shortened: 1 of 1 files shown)


sueloverso:
Never mind.  I believe I figured out the problem.

PR #3782, opened 2017-11-08
WT-3648 Read the thread's timestamp information once.

sueloverso:
@agorrod Please review.  This should fix the failure seen in `timestamp_abort` and fixes the similar usage in `test/format`.

### DIFF
--- a/test/csuite/timestamp_abort/main.c
+++ b/test/csuite/timestamp_abort/main.c
@@ -151,7 +151,7 @@ thread_ts_run(void *arg)
 			this_ts = th_ts[i];
 			if (this_ts == 0)
 				goto ts_wait;
-			if (this_ts != 0 && this_ts < oldest_ts)
+			else if (this_ts < oldest_ts)
 				oldest_ts = this_ts;
 		}
 
--- a/test/csuite/timestamp_abort/main.c
+++ b/test/csuite/timestamp_abort/main.c
@@ -121,7 +121,7 @@ thread_ts_run(void *arg)
 	WT_CURSOR *cur_stable;
 	WT_SESSION *session;
 	THREAD_DATA *td;
-	uint64_t i, last_ts, oldest_ts;
+	uint64_t i, last_ts, oldest_ts, this_ts;
 	char tscfg[64];
 
 	td = (THREAD_DATA *)arg;
@@ -148,10 +148,11 @@ thread_ts_run(void *arg)
 			 * any thread still with a zero timestamp we go to
 			 * sleep.
 			 */
-			if (th_ts[i] == 0)
+			this_ts = th_ts[i];
+			if (this_ts == 0)
 				goto ts_wait;
-			if (th_ts[i] != 0 && th_ts[i] < oldest_ts)
-				oldest_ts = th_ts[i];
+			if (this_ts != 0 && this_ts < oldest_ts)
+				oldest_ts = this_ts;
 		}
 
 		if (oldest_ts != UINT64_MAX &&
--- a/test/csuite/timestamp_abort/main.c
+++ b/test/csuite/timestamp_abort/main.c
@@ -75,7 +75,7 @@ static const char * const ckpt_file = "checkpoint_done";
 
 static bool compat, inmem, use_ts;
 static volatile uint64_t global_ts = 1;
-static uint64_t th_ts[MAX_TH];
+static volatile uint64_t th_ts[MAX_TH];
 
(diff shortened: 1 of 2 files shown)


test/csuite/timestamp_abort/main.c
```diff
@@ -148,10 +148,11 @@ thread_ts_run(void *arg)
 			 * any thread still with a zero timestamp we go to
 			 * sleep.
 			 */
-			if (th_ts[i] == 0)
+			this_ts = th_ts[i];
+			if (this_ts == 0)
 				goto ts_wait;
-			if (th_ts[i] != 0 && th_ts[i] < oldest_ts)
-				oldest_ts = th_ts[i];
+			if (this_ts != 0 && this_ts < oldest_ts)
```
agorrod: I don't think there is any need to re-check `this_ts` for 0 here, is there?

I'm concerned that the compiler will just get rid of the local variable.

I also don't really see how re-reading `this_ts` could lead to oldest moving further forward than expected. This code isn't relying on seeing a stable view of the values and once it's published the thread has finished committing their transaction. Could you describe the scenario that triggers a failure?

sueloverso: I put the description in the ticket itself. I think you're right that there's no need to recheck for 0.  I'll fix that.

agorrod:
Thanks @sueloverso - sorry I missed the description in the ticket. I see the window now - well done figuring that one out!

I'm still concerned the compiler could get rid of the local variable - essentially optimizing away this fix.

sueloverso:
First, I didn't add the description to the ticket until you asked here.  It just seemed a better place to put it.  So you didn't miss it because it wasn't there at the time!  Your concern exceeds my compiler expertise, unfortunately.

test/csuite/timestamp_abort/main.c
```diff
@@ -121,7 +121,7 @@ thread_ts_run(void *arg)
 	WT_CURSOR *cur_stable;
 	WT_SESSION *session;
 	THREAD_DATA *td;
-	uint64_t i, last_ts, oldest_ts;
+	uint64_t i, last_ts, oldest_ts, this_ts;
```
agorrod: I think this_ts needs to be `volatile` to avoid the compiler optimizing out the local variable. @keithbostic probably has an opinion about that, since we are still working towards being on the same page about the purpose of volatile.

sueloverso: I thought `volatile` was for globals so that it wasn't cached/optimized between threads.  We `WT_PUBLISH` every modification to the array so that isn't the problem.  I don't know what `volatile` does to a local stack variable, if anything.  It probably isn't harmful.  But I'll let @keithbostic have the say here.

michaelcahill: @sueloverso, the `WT_PUBLISH` only guarantees that all writes by the publishing before the publish are visible before the published write.  It doesn't have any effect on the code generated by compilers or scheduling within CPUs by reading threads.

The concern here is that setting a local variable `this_ts = th_ts[i]` has no semantic meaning: the compiler can (and usually does) implement that as an ordinary load from memory, but (absent some constraint like `volatile` or a barrier in the reading loop), it could equally well cache a value in a register and keep reading it (never seeing a published update), or read from memory every time the source code uses a local variable.

In this case, I think the best solution is to mark the global `th_ts` array `volatile`, so that the compiler has to generate a read each time it accessed, and cannot assume that the value stays the same after it reads into a local variable.

sueloverso: I think we need to talk because what you describe is what I'm trying to avoid and a description of the actual problem.  Marking the `th_ts` array volatile would likely make the original bug happen more frequently.  I want to be guaranteed that the value the reader first reads from `th_ts[i]` is the only value it sees for the duration of that loop.  If it changes on the writing CPU, I do not want the reader to see that updated value between the statements, but the next time through the walk of the array.  The sleep at the bottom should cause it to see updated values the next time around.

michaelcahill: Hi @sueloverso, sorry if it wasn't clear, but I'm suggesting we add `volatile` to `th_ts` in addition to your changes.  Without it, the local variable has no semantic meaning: the compiler could substitute re-reading the global value where it sees uses of the local variable.  Marking the variable (or the read) `volatile` prevents that.

sueloverso: Thanks for that clarification.  I was totally misinterpreting at first.  That makes sense and I have pushed that change.  This should be ready for merge.

michaelcahill:

PR #3783, opened 2017-11-09
WT-3637 Heap use after free failure

bvpvamsikrishna:


src/reconcile/rec_write.c
```diff
@@ -407,6 +407,17 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
 	 */
 	WT_PAGE_LOCK(session, page);
 
+	/*
+	 * The state of the page can change after the check is made and between
+	 * the lock, check again to make sure the state is not changed in
```
sulabhM: Can improve upon the comment, see if you like something like this:

"Though the caller checks for the page to be evict-able, there is a window between that check and the page lock where the page can become ineligible for eviction. We continue with the eviction of this page only if the page remains evict-able after taking the page lock."

src/reconcile/rec_write.c
```diff
@@ -407,6 +407,17 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
 	 */
 	WT_PAGE_LOCK(session, page);
 
+	/*
+	 * The check, is page evict-able is done again, even though the caller
+	 * checks, as there is a window between the check and page lock which
+	 * can result in page becoming ineligible for eviction.
```
michaelcahill: Sorry @bvpvamsikrishna, that comment still seems difficult to read and understand.  How about:

> Now that the page is locked, if attempting to evict it, check again whether eviction is permitted.  The page's state could have changed while we were waiting to acquire the lock (e.g., the page could have split).

michaelcahill:
Thanks @bvpvamsikrishna, lgtm.  I'll merge (since it already passed and the only changes since then are to the comment).

PR #3784, opened 2017-11-09
WT-3710 Get page-level lock to ensure single threaded page-split

nehakhatri5:
When a page split is in progress it is possible to get another split on this
page through its child in another thread. Acquire a page lock on the page
undergoing the split to prevent concurrent split on same page.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -345,6 +345,9 @@ __split_ref_prepare(
 	 * ascend into the created children, but eventually fail as that parent
 	 * page won't yet know about the created children pages. That's OK, we
 	 * spin there until the parent's page index is updated.
+	 *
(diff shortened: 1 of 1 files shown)


src/btree/bt_split.c
```diff
@@ -345,17 +345,22 @@ __split_ref_prepare(
 	 * ascend into the created children, but eventually fail as that parent
 	 * page won't yet know about the created children pages. That's OK, we
 	 * spin there until the parent's page index is updated.
+	 *
+	 * Get a page-level lock on the newly created parent page to
+	 * single-thread the split in this page.
```
michaelcahill: I'd suggest:

> Lock the newly created page to ensure it doesn't split until all child pages have been updated.

michaelcahill:

PR #3786, opened 2017-11-13
WT-3658 Fix the string length calculation when size is given

nehakhatri5:
When the format 'S' is preceded by a size, that indicates the maximum number of
bytes the string can store. This does not include the terminating NULL byte.
Hence strlen cannot be used to determine the length of the string. Instead use
custom version of strnlen.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -1165,6 +1165,7 @@ strget
 strlen
 strncpy
 strndup
+strnlen
 strtok
 strtoll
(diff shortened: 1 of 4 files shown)


nehakhatri5:
@sulabhM, @bvpvamsikrishna Please review this code change.

nehakhatri5:
@agorrod Please review this change.

michaelcahill:

PR #3787, opened 2017-11-13
WT-3444 Don't try to flush LSM if already in compact.

sueloverso:
@agorrod Here are fixes for the LSM issue where an update sees an LSM cursor with read-only settings.  There really are three different but related fixes in here.  Please review this.  The fixes are:
1.  In `wt_lsm_compact` we should hold the `lsm_tree` writelock until after we set the `WT_LSM_TREE_COMPACTING` flag. There could be a race with another compact and the existing code could end up with two threads in compact.
2. The LSM alter code should do a full discard on the tree.  The code that was there was a "mostly discard" and that wasn't clean and left things in a weird state.  I modified that to discard to force a tree open after an alter.  The alter was forcing the tree to be fully flushed to disk.
3. The LSM compact code should only push a flush work unit if the primary chunk isn't already stable.  This change is what really fixes the issue because if compact follows an alter, reopening the tree ends up with the primary chunk on disk and we want to proceed to the compact (i.e. the else clause).  We do not want to start a txn and force out a flush.

But you know LSM better than I do so please review this carefully.  The CONFIG in the ticket could reliably reproduce the bug in under 50 iterations and this ran for over 6000 over the weekend on PPC.

### DIFF
--- a/src/lsm/lsm_tree.c
+++ b/src/lsm/lsm_tree.c
@@ -16,7 +16,7 @@ static int __lsm_tree_set_name(WT_SESSION_IMPL *, WT_LSM_TREE *, const char *);
 
 /*
  * __lsm_tree_discard_state --
- *	Free the metadata configuration state  related LSM tree pointers.
+ *	Free the metadata configuration state-related LSM tree pointers.
  */
 static void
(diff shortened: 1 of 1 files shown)


agorrod:
Thanks @sueloverso - this change lgtm and LSM perf testing didn't show any regressions. I'm going to go ahead and merge.

PR #3788, opened 2017-11-14
WT-3745 Favor oplog reads when under cache pressure.

michaelcahill:
There are several behavior changes here to avoid situations where oplog
reads can block when a primary is under severe cache update pressure.
In particular:
* don't block if a scan encounters a large page;
* don't stall reads when the dirty limit of cache is reached;
* mark pages read from lookaside clean if it is safe to do so.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -206,10 +206,12 @@ __wt_las_cursor_open(WT_SESSION_IMPL *session)
 	 * opens (the first update is safe because it's single-threaded from
 	 * wiredtiger_open).
 	 */
-#define	WT_BTREE_LAS_FLAGS						\
-	(WT_BTREE_LOOKASIDE | WT_BTREE_NO_CHECKPOINT | WT_BTREE_NO_LOGGING)
-	if (F_MASK(btree, WT_BTREE_LAS_FLAGS) != WT_BTREE_LAS_FLAGS)
-		F_SET(btree, WT_BTREE_LAS_FLAGS);
+	if (!F_ISSET(btree, WT_BTREE_LOOKASIDE))
+		F_SET(btree, WT_BTREE_LOOKASIDE);
+	if (!F_ISSET(btree, WT_BTREE_NO_CHECKPOINT))
+		F_SET(btree, WT_BTREE_NO_CHECKPOINT);
+	if (!F_ISSET(btree, WT_BTREE_NO_LOGGING))
+		F_SET(btree, WT_BTREE_NO_LOGGING);
 
 	session->las_cursor = cursor;
 	F_SET(session, WT_SESSION_LOOKASIDE_CURSOR);
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -206,12 +206,10 @@ __wt_las_cursor_open(WT_SESSION_IMPL *session)
 	 * opens (the first update is safe because it's single-threaded from
 	 * wiredtiger_open).
 	 */
-	if (!F_ISSET(btree, WT_BTREE_LOOKASIDE))
-		F_SET(btree, WT_BTREE_LOOKASIDE);
-	if (!F_ISSET(btree, WT_BTREE_NO_CHECKPOINT))
-		F_SET(btree, WT_BTREE_NO_CHECKPOINT);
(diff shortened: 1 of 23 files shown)


michaelcahill:
@sueloverso, these changes aren't ready for review, but if you are looking at MongoDB performance issues, they might be worth considering.  I've been testing them locally and in Evergreen, cf https://evergreen.mongodb.com/version/5a0a6b5de3c331553d015944 -- so far this branch seems to avoid elections, but at the cost of slower writes.

More testing (and probably further tweaking) required.

sueloverso:
I will test this with the `insert_ttl.js` test I was running yesterday.  My version was only using a single node replica set, so I only have a primary.  I'll give it a try.

src/txn/txn.c
```diff
@@ -804,7 +811,7 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	update_timestamp = F_ISSET(txn, WT_TXN_HAS_TS_COMMIT);
 #endif
 
-	__wt_txn_release(session);
+	WT_ERR(__wt_txn_release(session));
```
sueloverso: I don't think this really is valid here.  We are well beyond the point in {{wt_txn_commit}} where it states "Note: we're going to commit: nothing can fail after this point.".  What is a user going to do with an error return from {{wt_txn_release}}?  Should they assume their transaction/data is not committed? It is committed and visible yet the commit returned error.

src/txn/txn.c
```diff
@@ -541,18 +541,20 @@ __wt_txn_reconfigure(WT_SESSION_IMPL *session, const char *config)
...
 	WT_TXN_GLOBAL *txn_global;
 	WT_TXN_STATE *txn_state;
+	bool readonly;
 
 	txn = &session->txn;
 	txn_global = &S2C(session)->txn_global;
 	txn_state = WT_SESSION_TXN_STATE(session);
 
-	WT_ASSERT(session, txn->mod_count == 0);
+	readonly = txn->mod_count != 0;
```
sueloverso: Is this backwards?  Wouldn't a readonly txn have a `txn->mod_count == 0`? If it is correct, it needs a comment because it is not intuitive.

src/txn/txn.c
```diff
@@ -921,6 +930,12 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
 	txn->mod_count = 0;
 
 	__wt_txn_release(session);
+	/*
+	 * We're between transactions, if we need to block for eviction, it's
+	 * a good time to do so.  Note that we must ignore any error return
+	 * because the user's data is committed.
+	 */
+	(void)__wt_cache_eviction_check(session, false, readonly, NULL);
```
sueloverso: We don't actually have to ignore the error in the call from `wt_txn_rollback` and the comment mentions commit, which is incorrect (just copied from `wt_txn_commit`).

src/cache/cache_las.c
```diff
@@ -206,12 +206,10 @@ __wt_las_cursor_open(WT_SESSION_IMPL *session)
...
 	 */
-	if (!F_ISSET(btree, WT_BTREE_LOOKASIDE))
-		F_SET(btree, WT_BTREE_LOOKASIDE);
-	if (!F_ISSET(btree, WT_BTREE_NO_CHECKPOINT))
-		F_SET(btree, WT_BTREE_NO_CHECKPOINT);
-	if (!F_ISSET(btree, WT_BTREE_NO_LOGGING))
-		F_SET(btree, WT_BTREE_NO_LOGGING);
+#define	WT_BTREE_LAS_FLAGS						\
+	(WT_BTREE_LOOKASIDE | WT_BTREE_NO_CHECKPOINT | WT_BTREE_NO_LOGGING)
+	if (F_MASK(btree, WT_BTREE_LAS_FLAGS) != WT_BTREE_LAS_FLAGS)
```
agorrod: Why do we care about checking first?
I also thought we were moving away from LAS as an abbreviation?

agorrod:
Thanks @michaelcahill - this looks good and passed all testing except Windows. I'm going to go ahead and merge into develop. If issues arise overnight we should follow up in other tickets.

PR #3789, opened 2017-11-14
WT-3720 flags macros cast flags to unsigned values, hiding warnings.

keithbostic:
@agorrod, @michaelcahill, based on the discussion in WT-3716, here's the first step.

If you agree this is a reasonable approach, I'll go ahead and do the rest of the work required for WT-3716 and WT-3720.


agorrod:
@keithbostic this seems like a reasonable approach. I think it would be neater if the blocks of flags only had a starting marker, but  it's safer the way you implemented it.

Could you outline your plan here. Specifically:
 * Will we switch all flags defines to use this form?
 * Will there be some way to mark flags that require 64 bits vs 32 bits?

keithbostic:
> Will we switch all flags defines to use this form?

I think that’s an open question, I don’t feel strongly, I could be talked into doing them all or just where it’s needed. Since we have to declare them unsigned for clang, perhaps it’s easier to do them all?

> Will there be some way to mark flags that require 64 bits vs. 32 bits?

I don’t think one is necessary. The Python script just uses however many bits it needs, and if we get the field declaration wrong and try to assign a 64 bit flag to a 32 bit field, the compiler will complain.

agorrod:
@keithbostic I've moved the discussion into the JIRA ticket.

keithbostic:
@agorrod, I've pushed the next step in this change, automatically generating most of the other flags in the source tree.

I skipped the logging slot bit masks because they're relatively tricky.

I didn't explicitly call out the places where a flag is placed on disk so we can't automatically generate it: let me know if you think that would be worth doing.

Assuming you approve this change, I think we can merge it, and then I'll do the remaining WT-3716 fixes in a separate branch. That said, I'd rather not merge this change until after we're done with the 3.6 branch, I think it's likely safe, but I'm not interested in finding out if I'm wrong.

src/include/log.h
```diff
@@ -92,6 +118,12 @@ union __wt_lsn {
  * a few special states, reserve the top few bits for state.  That makes
  * the maximum size less than 32 bits for both joined and released.
  */
+/*
+ * XXX
```
agorrod: I'd prefer if this had a more obvious name like `FLD_LOG_SLOT_ISSET` since it is doing something a bit odd.

keithbostic: > I'd prefer if this had a more obvious name like `FLD_LOG_SLOT_ISSET` since it is doing something a bit odd.

Agreed, pushed in a3d72e4.

agorrod:
Thanks @keithbostic lgtm.

PR #3791, opened 2017-11-15
WT-3746 Add sleep if we need to wait for writes to complete.

sueloverso:
@agorrod please review this small change.  With this 1 msec sleep it eliminates nearly all extraneous calls to `log_force_write`.  The tests I have run show no impact on performance and in MongoDB there is an independent thread calling log flush.


agorrod:
Thanks @sueloverso lgtm

PR #3792, opened 2017-11-16
WT-3725 Add timestamp related statistics

sueloverso:
@agorrod and @keithbostic I'm choosing you both because you're often have different views on how many statistics we should have. I went pretty far on these (but could go farther) or I could scale back.  Please take a look.  Here are some points to consider:
* I added statistics around the 3 timestamp and transaction related rwlocks. I added contention statistics around only the write locks. I did add the 6 more stats for the fail/success of the read locks.
* I committed the fail/success changes separately so we can choose to take/discard those entirely.
* If someone adds another code path that acquires one of these locks, they need to remember to add the stats line too.  I could, but did not, add a macro for acquiring these rwlocks and bumping the stat. But even with the macro, someone would need to remember to use the macro instead of just `__wt_read/writelock`.


dist/stat_data.py
```diff
@@ -446,19 +446,41 @@ def __init__(self, name, desc, flags=''):
...
+    TxnStat('txn_read_queue_try_success', 'transactions read timestamp queue write locks succeeded'),
+    TxnStat('txn_read_queue_wrlock', 'transactions read timestamp queue write locks'),
     TxnStat('txn_rollback', 'transactions rolled back'),
+    TxnStat('txn_set_ts', 'transactions set timestamp calls'),
+    TxnStat('txn_setts_commit', 'transactions set timestamp commit calls'),
+    TxnStat('txn_setts_commit_upd', 'transactions set timestamp commit updates'),
+    TxnStat('txn_setts_oldest', 'transactions set timestamp oldest calls'),
+    TxnStat('txn_setts_oldest_upd', 'transactions set timestamp oldest updates'),
+    TxnStat('txn_setts_stable', 'transactions set timestamp stable calls'),
+    TxnStat('txn_setts_stable_upd', 'transactions set timestamp stable updates'),
```
agorrod: I'd really like these statistics, but I found the names a bit confusing at the start. I'd prefer "transaction global .." rather than "transactions" - when I read transactions it made me think these statistics were tracking live transactions with timestamps associated with them, rather than calls to the set_timestamp API.

keithbostic: I'm not a fan of `setts`, I strongly prefer `set_ts`.

src/include/txn.i
```diff
@@ -31,7 +31,7 @@ __wt_txn_timestamp_flags(WT_SESSION_IMPL *session)
 }
 
 #if WT_TIMESTAMP_SIZE == 8
-#define	WT_WITH_TIMESTAMP_READLOCK(session, l, e)       e
+#define	WT_WITH_TIMESTAMP_READLOCK(session, l, stat, e)       e
```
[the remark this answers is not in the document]
keithbostic: I second this proposal, assuming there's no performance reason not to use the heavier-weight `WT_RWLOCK_INIT_TRACKED` (the existing macro has more statistics, including calculating a wait time).

src/include/txn.i
```diff
@@ -85,8 +85,9 @@ __wt_timestamp_set_zero(wt_timestamp_t *ts)
 
 #else /* WT_TIMESTAMP_SIZE != 8 */
 
-#define	WT_WITH_TIMESTAMP_READLOCK(s, l, e)	do {                    \
+#define	WT_WITH_TIMESTAMP_READLOCK(s, l, stat, e) do {                  \
 	__wt_readlock((s), (l));                                        \
+	WT_STAT_CONN_INCR((s), stat);                                   \
 	e;                                                              \
 	__wt_readunlock((s), (l));                                      \
 } while (0)
```
keithbostic: I'd move the statistics set outside of the lock, there's no reason to do it inside the lock, is there?

Probably doesn't matter, but it feels odd.

sueloverso:
@agorrod This is now ready for final review. Using `WT_RWLOCK_INIT_TRACKED` made the diff quite a bit shorter than it was. There are a couple whitespace/KNF edits thrown in there that I saw in my travels.

agorrod:
Thanks @sueloverso these changes lgtm now, but I'm going to hold off merging until the 3.6 RC candidate has been cut to avoid any conflicts.

agorrod:
This introduced a performance regression on some MongoDB workloads - I added a comment to the JIRA ticket.

michaelcahill:
@sueloverso, I've switched `$PERF_BRANCH` to point here so we can confirm that using `__wt_rdtsc` resolves the perf regression.

That said, how confident can we be that the calculated times we get from `rdtsc` are accurate (or at least comparable between runs / machines)?  Maybe we only care at the high level to detect "slow" events, and the units or skew from clock scaling isn't going to be an issue.

@fedorova, do you have any advice for how to interpret the statistics we gather from this branch?

sueloverso:
@michaelcahill This ticket is waiting onWT-3790 which asks some of those questions because we need a translation from tsc ticks to nanoseconds.

agorrod:
I switched the branch back - there was a regression on the LSM jobs, but that is due to this branch not including a change that is on develop that improved perf on those jobs.

The flagged regression was in Evergreen perf testing, we should kick off a patch build there too.

sueloverso:
@agorrod I put perf comparisons in the ticket. This is ready for merge now I believe.

agorrod:
Thanks @sueloverso lgtm.

PR #3793, opened 2017-11-17
WT-3751 Allow splits when no data is visible.

michaelcahill:
In particular, when reconciling insert lists on row store leaf pages, we also need to check for splits.  Now that we are handling this case correctly, we are seeing splits with empty page images, which requires further changes in reconciliation to track the last key associated with a page (even if that key wasn't written to the page image).

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -264,6 +264,9 @@ typedef struct {
 	WT_ITEM *cur, _cur;		/* Key/Value being built */
 	WT_ITEM *last, _last;		/* Last key/value built */
 
+	WT_ITEM *skipkey, _skipkey;	/* Last skipped key */
+	bool last_key_skipped;		/* Last key was skipped. */
+
 	bool key_pfx_compress;		/* If can prefix-compress next key */
 	bool key_pfx_compress_conf;	/* If prefix compression configured */
 	bool key_sfx_compress;		/* If can suffix-compress next key */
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, please take a look.  This change allows reconciliation to split pages, and deals with places I could find where reconciliation expects *something* to be written before a split.

I suspect some cleanup is possible ;-)

I could not see a path that involves writing keys to pages: empty values give incorrect results (e.g., we end up with checkpoints containing invalid data), and row store reads are allergic to `WT_CELL_DEL`.  Even if we made the latter work now, we'd have issues with data files not being readable by earlier versions.

michaelcahill:
@keithbostic, this branch is failing in some MongoDB workloads with errors like this:
```
132976. [MongoDFixture:job0] 2017-11-20T12:10:34.381+0000 E STORAGE [thread1680] 
WiredTiger error (0) [1511179834:381081][1820:0x7f258a247700], file:index-10653-
4737883234740453942.wt, eviction-server: the 1 key on page at [reconcile-image] is the first non-
overflow key on the page and has a non-zero prefix compression value
132977. [MongoDFixture:job0] 2017-11-20T12:10:34.381+0000 E STORAGE [thread1680] 
WiredTiger error (0) [1511179834:381893][1820:0x7f258a247700], file:index-10653-
4737883234740453942.wt, eviction-server: __rec_split_write, 3707: verify_image == false || 
__wt_verify_dsk_image(session, "[reconcile-image]", chunk->image.data, 0, true) == 0
```
I really only want to track keys for the case where nothing at all is stored on page, but this indicates that what's here at the moment is interacting with prefix compression.  If you can figure this out (or see a better way to get the same effect), please take a run at it today.

michaelcahill:
Thanks @keithbostic, this has been reviewed and has been through stress and perf testing during the day today.  I'm going to merge.

PR #3794, opened 2017-11-20
WT-3752 Allow trimming of obsolete modify updates.

michaelcahill:
Obsolete update discards look for a contiguous block of obsolete updates at the end of an update chain.  Previously, we reset the search each time we saw a modify update.  That is not what we want: it is fine to discard obsolete modify updates as long as there is a complete value somewhere after them in the list.


michaelcahill:
Thank @sulabhM, will you please go ahead and merge?

PR #3795, opened 2017-11-20
WT-3751 Allow splits when no data is visible.

keithbostic:
A slightly different approach, routing all of the built keys through the
"current" buffer.


michaelcahill:
Thanks @keithbostic, lgtm.  I'll merge this into the other branch.

PR #3796, opened 2017-11-21
WT-3754 serial updates to the oldest to consider pinned ts for visibility

sulabhM:


sulabhM:
@agorrod please review

src/btree/row_modify.c
```diff
@@ -339,9 +342,23 @@ __wt_update_obsolete_check(
...
-		    S2C(session)->txn_global.last_running;
+	if (count > 20 && page->modify != NULL) {
+		page->modify->obsolete_check_txn = txn_global->last_running;
+#ifdef HAVE_TIMESTAMPS
+		if (txn_global->has_pinned_timestamp) {
+			WT_WITH_TIMESTAMP_READLOCK(session, &txn_global->rwlock,
+			    __wt_timestamp_set(
+				&page->modify->obsolete_check_timestamp,
+				&txn_global->pinned_timestamp));
+			page->modify->has_obsolete_check_timestamp = true;
```
michaelcahill: @sulabhM, can you explain why we need this boolean?

Without it, `page->modify->obsolete_check_timestamp` should be zero, which means it should always pass the `__wt_txn_visible_all` check below.

src/include/serial.i
```diff
@@ -309,11 +310,20 @@ __wt_update_serial(WT_SESSION_IMPL *session, WT_PAGE *page,
...
 	 */
 	if ((txn = page->modify->obsolete_check_txn) != WT_TXN_NONE) {
-		if (!__wt_txn_visible_all(session, txn, NULL)) {
+#ifdef HAVE_TIMESTAMPS
+		obsolete_timestamp =
+		    page->modify->has_obsolete_check_timestamp ?
+		    &page->modify->obsolete_check_timestamp : NULL;
+#else
+		WT_UNUSED(obsolete_timestamp);
+#endif
```
michaelcahill: I think this block can be simplified to `obsolete_timestamp = WT_TIMESTAMP_NULL(&page->modify->obsolete_check_timestamp);`, then pass it unconditionally below.

michaelcahill:
Thanks @sulabhM, lgtm.  I'm going to kick off a patch build to see what impact this has on MongoDB performance before merging.

agorrod:
lgtm - I'm going to merge

PR #3797, opened 2017-11-21
WT-3758 Turn on snappy by default for LAS

bvpvamsikrishna:
To turn on snappy by default for LAS.
This improves the performance of insert-heavy workloads that end up disk-bound writing lookaside records.


michaelcahill:
Status: we're waiting on performance test results before deciding whether to merge this change.

agorrod:
Thanks @bvpvamsikrishna the testing looked good - I'll go ahead and merge.

PR #3798, opened 2017-11-22
WT-3732 For index add/drop, temporarily get the table dhandle exclusively

ddanderson:
This ensures that there is a time during the add/drop that all cursors
on the table are closed and must be reopened.

Fix some tests that leave cursors open during add/drop.


ddanderson:
@michaelcahill, can you take a look at this? I've written up some notes in https://jira.mongodb.org/browse/WT-3732 about this. I don't see a simple solution that totally solves the problem, perhaps you do? I do see complex solutions that I think will solve it...

agorrod:
@ddanderson would it make sense to hold the handle exclusively while creating the index instead of just getting and releasing it?

ddanderson:
@agorrod, conceptually, holding the handle exclusively would be better, and I believe it used to do that.  However, apparently this didn't work out, the change in https://github.com/wiredtiger/wiredtiger/pull/3627/files removed it (WT-3547 is the ticket).

agorrod:
@sueloverso You made the change in WT-3547 - could you help @ddanderson verify that this change doesn't re-introduce the hang? If it doesn't I'm inclined to merge this for now to unblock cursor caching, and work towards the complete solution outlined in WT-1598 over time.

sueloverso:
@ddanderson I can run your branch against the tests in WT-3547 that were failing. There were several and none on AWS. A few failed on MacOS (both my laptop and Evergreen) and PPC. I helped investigate failures but I was not actually involved in the code fix nor review so I had to read up on them today.

sueloverso:
All tests have run successfully on my laptop and PPC. I will leave tests running all weekend on the PPC to make sure it gets enough runtime to confirm this branch is okay.

sueloverso:
@ddanderson This ran well on PPC over the weekend. Please get it rebased with the current version of develop since there were a lot of merges over the past few days and then I'm good for it to be merged.

ddanderson:
@sueloverso, thanks for the tests!  I'll merge it.

PR #3799, opened 2017-11-22
WT-3760 Avoid writing overflow values into the lookaside file

bvpvamsikrishna:
Avoid writing overflow values into the lookaside file


agorrod:
Thanks @bvpvamsikrishna lgtm.

PR #3801, opened 2017-11-22
WT-3607 compaction skips all in-cache blocks with associated disk images

keithbostic:


michaelcahill:
Hey @keithbostic, nice catch!

Wouldn't this bug have made compact skip all pages?  In other words, won't all on-disk pages have disk images at this point, so compact would skip over them?

I'm asking (1) to understand and (2) to check whether our tests needs work the way we want.

keithbostic:
> Wouldn't this bug have made compact skip all pages? In other words, won't all on-disk pages have disk images at this point, so compact would skip over them?

The test I changed is only done in the case of compacting a page found in the cache, where there had been a reconciliation that had a `WT_PM_REC_MULTIBLOCK` result, so an on-disk page or a one-for-one block replacement of an in-cache page wouldn't trigger the problem. That's why the failure was rare: you have to compact a file where there are blocks at the end of the file that are the result of the reconciliation of a page that's still in-memory and which has associated disk images.

michaelcahill:

PR #3803, opened 2017-11-23
WT-3761 Queue page for urgent eviction on page release

nehakhatri5:
Queue the page for urgent eviction when application thread is not supposed to do
it. This is done to allow splitting the pages to maintain the performance of
write operations.


michaelcahill:
@nehakhatri5, I've taken a look -- one issue with the change was it was only attempting eviction here if an in-memory split was possible.  I didn't expect that as part of the change.  I've also suggested that we check whether the page is clean, because we should always be able to evict a clean page (e.g., when doing a scan with `WT_SESSION_READ_WONT_NEED`.

If this version looks okay to you and performs as well in tests, then I think we should merge it in the morning.

michaelcahill:

PR #3804, opened 2017-11-23
WT-3763 Tune eviction for various MongoDB workloads.

michaelcahill:
In particular, balance primary inserts, overflowing the cache to use the lookaside table, secondary inserts and secondary reads of the oplog (assuming the oplog is at least partially stored in the lookaside table).


michaelcahill:
autobuild: retest this please

(timed out, not clear what the cause was)

agorrod:
I'm going to go ahead and merge this branch. It makes some MongoDB workloads significantly faster, and the wtperf Jenkins jobs are no longer showing a regression.

PR #3805, opened 2017-11-23
WT-3761 Don't immediately evict pages even if they look clean

agorrod:
It is an optimization that tickles a corner case with checkpoints and the fast truncate path.


agorrod:
lgtm (self review)

PR #3807, opened 2017-11-24
WT-3763 Decide whether to retry eviction based on eviction activity

agorrod:
Rather than on whether transaction visibility is moving forward, this should help since transaction visibility isn't as simple as it was before timestamps.


agorrod:
I'm going to pull this into the other 3763 branch - so there is just one place to work.

PR #3809, opened 2017-11-24
WT-3765 Prevent eviction of pages being truncated.

michaelcahill:
When an application performs a truncate operation, WiredTiger marks pages deleted.  If such a page is subsequently read with a view earlier than the truncate, the page is reinstantiated and all records deleted (as if truncate had taken the slow path).

Such a page cannot be evicted: if the truncate is rolled back, it expects to find the page and any tombstones so it can roll them all back.  If the page is evicted or split, the rollback will fail.

This change takes two approaches: don't allow checkpoints to queue pages for urgent eviction, since checkpoints use special rules to determine whether eviction is permitted.  In addition, check for uncommitted truncate operations before allowing any page to be evicted.

### DIFF
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -1336,6 +1336,14 @@ __wt_page_can_evict(WT_SESSION_IMPL *session, WT_REF *ref, bool *inmem_splitp)
 		return (false);
 
 	/*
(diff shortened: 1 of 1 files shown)


agorrod:

PR #3810, opened 2017-11-24
WT-3763 Retry eviction if we're using lookaside.

michaelcahill:
When we're not under cache pressure, retrying eviction of pages can waste effort if the global transaction state hasn't changed.  However, lookaside eviction allows pages to be evicted even when a timestamp or transaction ID is stuck, so we should retry in that case.


agorrod:
I'm going to pull this into the other 3763 branch - so there is just one place to work.

PR #3811, opened 2017-11-25
WT-3753 Building on Windows --enable-java

frasator:
SConstruct update to support building wiredtiger_java.dll.

-> wiredtiger.jar is created inside lang/java/**wiredtiger.jar**
-> **wiredtiger_java.dll** is created in root repo folder.

Usage example from cmd:
c:\Python27\python.exe c:\Python27\Scripts\scons.py --enable-java=c:/Users/paco/Downloads/swigwin-3.0.12/swig.exe,"C:/Program Files/Java/jdk1.8.0_151/"

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -31,6 +31,9 @@ AddOption("--enable-lz4", dest="lz4", type="string", nargs=1, action="store",
 AddOption("--enable-python", dest="lang-python", type="string", nargs=1, action="store",
           help="Build Python extension, specify location of swig.exe binary")
 
+AddOption("--enable-java", dest="lang-java", type="string", nargs=1, action="store",
+          help="Build java extension, specify location of swig.exe binary and Java JDK dir separated by comma")
+
 AddOption("--enable-snappy", dest="snappy", type="string", nargs=1, action="store",
           help="Use snappy compression")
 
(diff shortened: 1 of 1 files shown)


wtbuild:

agorrod:
autobuild: test this please

agorrod:
Thanks @frasator for the contribution - well done getting something running!
@ddanderson would you mind taking a look at this proposed change?

frasator:
You are welcome

It also works with the current master (tag 2.9.3) branch.

SConstruct
```diff
@@ -272,6 +275,63 @@ if useLz4:
...
 wt_objs = [env.Object(a) for a in wtsources]
 
+#
+# wiredtiger_java.dll
+## Usage example
+## c:\Python27\python.exe c:\Python27\Scripts\scons.py --enable-java=c:/Users/paco/Downloads/swigwin-3.0.12/swig.exe,"C:/Program Files/Java/jdk1.8.0_151/"
+useJava = GetOption("lang-java")
+if useJava:
+    useJavaSplit = useJava.split(',')
+    if len(useJavaSplit) == 2:
```
ddanderson: It would be good to have an error reported if the option value doesn't have exactly one comma. For example, with `--enable-java=true` I think as written, it will silently continue and not build java.

SConstruct
```diff
@@ -272,6 +275,63 @@ if useLz4:
...
+            target="wiredtiger_java",
+            source=wt_objs + [objectJavaWrap] + ['build_win/wiredtiger.def'], LIBS=wtlibs)
+
+        env.Depends(wtjavadll, [filelistfile, version_file])
+        Default(wtjavadll)
+        # wiredtiger.jar
+        env['JAVAC'] = '"' + javaPath + '/bin/javac.exe"'
+        env['JAR'] = '"' + javaPath + '/bin/jar.exe"'
+        wtClasses = env.Java('lang/java/build', 'lang/java/src/')
+        env.Depends(wtClasses, swigJavaFiles)
```
ddanderson: I'd leave out editorializing in the comment.

SConstruct
```diff
@@ -272,6 +275,63 @@ if useLz4:
...
+        "lang/java/src/com/wiredtiger/db/SearchStatus.java",
+        "lang/java/src/com/wiredtiger/db/Session.java",
+        "lang/java/src/com/wiredtiger/db/WT_ITEM_HOLD.java",
+        "lang/java/src/com/wiredtiger/db/WT_MODIFY_LIST.java",
+        "lang/java/src/com/wiredtiger/db/wiredtiger.java",
+        "lang/java/src/com/wiredtiger/db/wiredtigerConstants.java",
+        "lang/java/src/com/wiredtiger/db/wiredtigerJNI.java"]
+
+        swigCFile = "wiredtiger_wrap.c"
+
```
ddanderson: I prefer the way the section that runs SWIG for Python organizes this. Using a cloned copy of the environment (as the Python code does) is also more flexible, as we may have to turn off certain compiler warnings only for SWIG generated code.

ddanderson:
Hello @frasator, can you also sign the MongoDB contributor agreement here: https://www.mongodb.com/legal/contributor-agreement . Thanks!

SConstruct
frasator: @ddanderson,  I moved the java section after the python section, added a check  for  comma (",") and updated some comments, also added an error message with an example of usage.
Didn't fixed the `SWIGFLAGS`  because is the first time I use Scons and I'm not sure how to do it the way you want, so please feel free to make the necessary modifications to fit your needs.

SConstruct
```diff
@@ -355,6 +358,61 @@ if GetOption("lang-python"):
 
     Default(swiginstall, copySwig)
 
+# Javap SWIG wrapper for WiredTiger
+enableJava = GetOption("lang-java")
+if enableJava and "," in enableJava:
+    enableJavaPaths = enableJava.split(',')
+    if len(enableJavaPaths) == 2:
+        swigExe = enableJavaPaths[0]
```
ddanderson: @frasator, I don't think this catches the error with `lang-java=x,y,z`, won't that be silently accepted?
Other than that, I like the changes you've made.

frasator:
Updated it again,I think is more simple with count.

ddanderson:
@frasator, that looks much better, thanks.  Did you get a chance to sign the contributor agreement?

agorrod:
@ddanderson I can see that the contributor agreement has been signed - thanks @frasator !

ddanderson:
Thanks, @frasator!

PR #3812, opened 2017-11-27
WT-3768 Add an update type to version on-page values.

michaelcahill:
Previously, we stored full copies of on-disk values in the lookaside table as well as in the disk image when we stored values that were not globally visible.

This branch uses a different approach, and stores a new update type (called WT_UPDATE_BIRTHMARK) whose purpose is to give the on-page value a transaction time when it becomes visible.  Readers that see a birthmark should read the on-disk value.  Readers that skip over a birthmark because it is not visible to them should not see the on-page value (equivalent to reading a tombstone at the end of the update list).


michaelcahill:
@keithbostic, this branch is not yet working, but it would be great if you can take a look anyway.

I think the idea is basically sound, and works for the common case in MongoDB of insert-only tables where there is only ever a single value for each key and the only thing we need to figure out is whether the value should be visible to a reader.  There is more of a writeup in the ticket.

Where this change is currently falling down is in interactions between lookaside and update/restore eviction.  We're ending up with an extra copy of the on-page data in the update list in some cases (with `txnid == WT_TXN_NONE`), so readers see a value when they should get `WT_NOTFOUND`.  Running the Python `timestamp07` test is a fairly reliable reproducer.

michaelcahill:
@keithbostic, I merged `develop` into this branch today because there were some conflicts, but I suspect I've screwed something up.  It's now failing in a different way, asserting that we're trying to attach a birthmark without finding an on-page value.

Feel free to revert back to yesterday's version if this wastes your time.

michaelcahill:
@keithbostic, just FYI, I merged develop in here to resolve a conflict.  No real change, just hoping I to avoid treading on your if you have changes in flight.

keithbostic:
@michaelcahill, @bvpvamsikrishna: the change in ff4c226 fixes the problem in `test_timestamp07`, without that change we're setting the wrong `onpage_timestamp` value and so this test:
```
if (ref->page_las->las_skew_newest &&
    !S2C(session)->txn_global.has_stable_timestamp &&
    __wt_txn_visible_all(session, ref->page_las->las_max_txn,
    WT_TIMESTAMP_NULL(&ref->page_las->onpage_timestamp))) {
        page->modify->rec_max_txn = ref->page_las->las_max_txn;
        __wt_timestamp_set(&page->modify->rec_max_timestamp,
            &ref->page_las->onpage_timestamp);
}
```
in `__las_page_instantiate()` incorrectly marks the page clean.

Michael, please review the code movement in ff4c226: a related question is if we should be clearing `*updp` before setting `r->update_used` and/or `all_visible`. I'd say "yes", but I'm not positive about `all_visible`.

Second, this change also fixes the failures in `test_timestamp04`, but I don't completely understand why. I removed the existing birthmark related code in `__txn_rollback_to_stable_lookaside_fixup()` because I'm sure it's wrong, but I'm not sure exactly what we should replace it with. I halfway convinced myself we're safe because there must be a tombstone in the existing, in-memory update list, but again, I'm not positive of that.

Third, in the build for ff4c226, the `format` smoke test failed:
```
t: process 34745
[1513094782:685712][34890:0x7f6129a74700], t, file:wt.wt, eviction-server: __split_multi_inmem, 1484: cbt.compare == 0 || supd->ins != NULL
```
This is an assert I added, that is, unless we have an insert record, we should expect an exact match with an existing key on the page.

michaelcahill:
Thanks @keithbostic, this branch is generally looking good to me.

It makes sense that we have to take any existing birthmark into account when calculating `onpage_timestamp`, so rearranging the code seems fine from that point of view.  I think we only need to track that timestamp when there are saved updates, so it doesn't have to be done unconditionally at the end of the function.

@agorrod has kicked off a MongoDB [patch build](https://evergreen.mongodb.com/version/5a30a73ce3c33178530013d9), so we'll see how that ends up.  I've also pointed the [stress-sanitizer job](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/) at the branch to get some more testing overnight.  If that is looking good, I'll try to get this merged tomorrow.

michaelcahill:
Unfortunately there was a failure fairly quickly in [stress testing](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/17933/console):

```
[1513145000:740259][8203:0x7fd13f204700], t, file:wt, WT_CURSOR.reserve: __rec_append_orig_value, 1216: unpack != NULL && unpack->type != WT_CELL_DEL
```

michaelcahill:
Thanks @keithbostic!

I've been putting this branch through its paces today and I'm happy to merge it now.  That will give us more coverage over the weekend.

PR #3813, opened 2017-11-27
WT-3767 Avoid lookaside reads when history is not required for queries.

michaelcahill:


michaelcahill:
@keithbostic, here is another change I think will help by avoiding lookaside reads when the on-page versions are correct for queries.  This is a common case when reading the oplog during secondary lag (i.e., it is on the critical path for MongoDB replication performance).

Please take a look -- adding new ref states always turns into more work than expected but so far I think this is hanging together.  The nastiest part is making sure we load the history before attempting an update.

agorrod:
@michaelcahill this looks like a real issue with the last commit you pushed. I can reproduce it easily on linux, but my timestamp filtering knowledge isn't strong enough to be confident about solving it.

src/include/btmem.h
```diff
@@ -757,12 +763,13 @@ struct __wt_ref {
 	volatile uint32_t pindex_hint;	/* Reference page index hint */
 
 #define	WT_REF_DISK	 0		/* Page is on disk */
-#define	WT_REF_DELETED	 1		/* Page is on disk, but deleted */
-#define	WT_REF_LOCKED	 2		/* Page locked for exclusive access */
-#define	WT_REF_LOOKASIDE 3		/* Page is on disk with lookaside */
-#define	WT_REF_MEM	 4		/* Page is in cache and valid */
-#define	WT_REF_READING	 5		/* Page being read */
-#define	WT_REF_SPLIT	 6		/* Parent page split (WT_REF dead) */
+#define	WT_REF_AMNESIA	 1		/* Page is in cache without history */
```
agorrod: My vote is for calling this new state `WT_REF_MEM_PARTIAL` - or `WT_REF_MEM_INCOMPLETE` if you have a taste for three extra characters.

michaelcahill: What I'm trying to capture with `WT_REF_AMNESIA` is the idea that we have a snapshot of the page without history.  We need to keep checking that the snapshot is valid for each reader of the page: if some reader needs the history, it has to instantiate it (and flip `ref->state` to `WT_REF_MEM`).

I don't love `MEM_PARTIAL` or `MEM_INCOMPLETE` because they suggest something is missing but don't have any connection to missing history.  They're almost negative flags ("REF_MEM" without something), which has gotten us confused in the past.

I spent ~too much~some time looking at a thesaurus: how about `WT_REF_LATENT` or `WT_REF_REPRESSED` or `WT_REF_BLANK_SLATE`?

Maybe less loaded, how about `WT_REF_IMAGE`, meaning "the disk image has been loaded without any history"?

michaelcahill:
@sueloverso, I believe this branch is working now.  At least it runs WiredTiger tests and I haven't seen any failures from MongoDB tests.  Can you please take a look, think about the naming issue @agorrod raised, and about whether we should extend testing to improve coverage?

src/include/btmem.h
sueloverso: I think `WT_REF_IMAGE` is too generic. Although I don't like `WT_REF_AMNESIA` it is unique enough to catch one's attention.  I too spent a bit of time with the thesaurus, how about `WT_REF_LIMBO`? Defined at dictionary.com as `an intermediate, transitional, or midway state` which is a pretty accurate description of this, although it can have some connotations of being "incomplete". I see how you got to `AMNESIA` too.

sueloverso:
Neither the ticket nor this pull request have any testing information in them.  I pushed a sub-branch with one potential naming change. I also pushed to this branch a couple statistics.

sueloverso:
autobuild: retest this please

michaelcahill:
There was a failure on this branch in stress testing: http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/17931/console

```
[1513059547:487103][10010:0x7fa3f62c7700], t, file:wt.wt, WT_CURSOR.remove: __cursor_page_pinned, 72: cbt->ref->state != WT_REF_LIMBO || !F_ISSET(&session->txn, WT_TXN_UPDATE) || eviction_ok
[1513059547:487146][10010:0x7fa3f62c7700], t, file:wt.wt, WT_CURSOR.remove: aborting WiredTiger library
```

@sueloverso, can you please take a look?

src/btree/bt_cursor.c
```diff
@@ -57,13 +57,36 @@ __cursor_state_restore(WT_CURSOR *cursor, WT_CURFILE_STATE *state)
...
+
+	if (!F_ISSET(cbt, WT_CBT_ACTIVE))
+		return (false);
+
+	/*
+	 * If we get here during an update with a page lacking history, we'd
+	 * better be able to evict it.
+	 */
+	WT_ASSERT(session, cbt->ref->state != WT_REF_LIMBO ||
+	    !F_ISSET(&session->txn, WT_TXN_UPDATE) || eviction_ok);
```
michaelcahill: @sueloverso, @keithbostic this is the assertion that fired.  In other words, we get here with a "limbo" page trying to do a `WT_CURSOR::remove` operation and `eviction_ok == false`.

This relates to WT-3211 and WT-3776.  We have to evict in order to get a page that remove can update.  Will that violate the recent bug fix?

keithbostic: The original problem was caused by a cursor-remove call positioned to an already removed item, that is, it's possible:

* to have a page populated entirely by deleted items,
* where there's a cursor positioned on one of those items,
* call WT_CURSOR.remove on the item with WT_CURSTD_OVERWRITE set,
* the page is evicted,
* which results in a discarded page (because all the items are deleted),
* the cursor remove then re-instantiates an empty page to do its remove,
* the remove fails, setting the return value to not-found,
* the WT_CURSTD_OVERWRITE configuration resets the return value to 0,
* success implies the cursor should return positioned, 
* there's nothing to position the cursor to,
* core dump.

If there's a chance that a WT_REF_LIMBO page can be full of deleted items during the eviction, there's a problem. Can that happen? I took a look at the branch, but wasn't sure of the answer to that question.

If that can't happen, I think we're OK, and WT_REF_LIMBO can figure into the question of whether or not eviction is permitted.

If that can happen, the alternative I thought about in WT-3776 was to always allow eviction in the cursor-remove path, and define "the cursor is positioned" as "the cursor's key remains set", which scares me some. I don't think there's a lot of functionality at risk here that a user would care about, but it's the kind of edge case we'll get caught on a couple of years from now.

We could also fix this problem by not allowing deletion of a page during a cursor-remove eviction, that is, we won't create an empty page if we're evicting on behalf of cursor-remove and WT_CURSTD_OVERWRITE is configured. It's a nasty little side-case, but I think we could drill the hole without getting too ugly.

sueloverso: I've been running the CONFIG and have not yet reproduced the failure.

Do we ever write out a page full of deleted items? If we can then a LIMBO page can be full of deleted items, otherwise it cannot and we're OK.  A LIMBO page is read-only and instantiated from what we read from disk. However `positioned` would be true after a `search`.

sueloverso: (Update - it took 100 iterations to reproduce so we can test a fix.)

keithbostic: Once an item is deleted, it won't ever be written out, so I don't believe it's possible for row-store to write out a page full of deleted items.

sueloverso:
I have pushed a fix for this and it has run well over 300 iterations on the failing CONFIG.

src/btree/bt_cursor.c
```diff
@@ -952,7 +952,8 @@ __wt_btcur_remove(WT_CURSOR_BTREE *cbt)
 	 * that's all a positioned cursor implies), but it's probably safer to
 	 * avoid page eviction entirely in the positioned case.
 	 */
-	if (__cursor_page_pinned(cbt, !positioned) &&
+	if (__cursor_page_pinned(cbt,
+	    (!positioned || cbt->ref->state != WT_REF_LIMBO)) &&
```
keithbostic: I'd suggest a big comment explaining why it's OK to evict these pages, why we're guaranteed they won't end up empty.

michaelcahill:
@sueloverso, sorry I didn't get much time to look at this today.  There's still an issue here -- although a `WT_REF_LIMBO` page must have had some entries at some point, I think it's possible for the page to have tombstones on all of the records -- i.e., the latest version of the page is empty, but there is some historical data being kept for snapshot reads.

If it helps, we could probably skip eviction for `remove` for cases other than `LIMBO`.  Once there are enough tombstones, the page will be marked with `WT_READGEN_OLDEST`, so eviction will be attempted when the page is released (e.g., if the cursor is reset or walks off the page).

michaelcahill:
@sueloverso, I should have been more explicit -- I merged #3841 knowing about the hang: it isn't reproducing for me, so I'm inclined to chase it in this branch (which now also includes WT-3768) rather than keeping an extra pull request open for a change that looks right to me.

michaelcahill:
The current failure in this branch relates to evicting a tree during close.  I think we should read all `WT_REF_LOOKASIDE` and `WT_REF_LIMBO` pages in that case but on this branch we're skipping some, leading to eviction of the parent page failing because it treats a lookaside ref as "active".

Alternatively, we could be correctly visiting a `WT_REF_LIMBO` page as we walk the tree but evicting the page and leaving it as `WT_REF_LOOKASIDE`.  That would also explain this -- in `bt_read.c` we have to distinguish between a checkpoint (which sets `WT_READ_CACHE | WT_READ_LOOKASIDE`) and the walk to evict a tree, which sets the same flags.  We shouldn't call `__las_page_skip` for the eviction case.

```
#3  0x00007f23ee69187b in __wt_assert (session=0x55afe7d0dfc0, error=0,
    file_name=0x7f23ee6d5e80 <__func__.35087> "__rec_child_modify",
    line_number=1806, fmt=0x7f23ee6d51fa "%s") at ../src/support/err.c:501
#4  0x00007f23ee65fb62 in __rec_child_modify (session=0x55afe7d0dfc0,
    r=0x55afe8a03790, ref=0x7f23dc005470, hazardp=0x7fffde67eb1f,
    statep=0x7fffde67eb2c) at ../src/reconcile/rec_write.c:1806
#5  0x00007f23ee666934 in __rec_row_int (session=0x55afe7d0dfc0,
    r=0x55afe8a03790, page=0x55afeab4bd80) at ../src/reconcile/rec_write.c:5084
#6  0x00007f23ee65d2a3 in __wt_reconcile (session=0x55afe7d0dfc0,
    ref=0x55afeac3e908, salvage=0x0, flags=130, lookaside_retryp=0x0)
    at ../src/reconcile/rec_write.c:470
#7  0x00007f23ee60894a in __wt_evict_file (session=0x55afe7d0dfc0,
    syncop=WT_SYNC_CLOSE) at ../src/evict/evict_file.c:96
#8  0x00007f23ee58fd74 in __wt_cache_op (session=0x55afe7d0dfc0,
    op=WT_SYNC_CLOSE) at ../src/btree/bt_sync.c:405
#9  0x00007f23ee6a9e1c in __checkpoint_tree (session=0x55afe7d0dfc0,
    is_checkpoint=false, cfg=0x0) at ../src/txn/txn_ckpt.c:1569
#10 0x00007f23ee6aa5e4 in __wt_checkpoint_close (session=0x55afe7d0dfc0,
    final=true) at ../src/txn/txn_ckpt.c:1812
#11 0x00007f23ee5bd607 in __wt_conn_dhandle_close (session=0x55afe7d0dfc0,
    final=true, mark_dead=false) at ../src/conn/conn_dhandle.c:329
#12 0x00007f23ee5be6eb in __wt_conn_dhandle_discard_single (
    session=0x55afe7d0dfc0, final=true, mark_dead=false)
    at ../src/conn/conn_dhandle.c:720
#13 0x00007f23ee5bea35 in __wt_conn_dhandle_discard (session=0x55afe7d0dfc0)
    at ../src/conn/conn_dhandle.c:787
#14 0x00007f23ee5c3324 in __wt_connection_close (conn=0x55afe7e89a80)
    at ../src/conn/conn_open.c:112
#15 0x00007f23ee5b2f8b in __conn_close (wt_conn=0x55afe7e89a80, config=0x0)
    at ../src/conn/conn_api.c:1136
```

src/btree/bt_cursor.c
```diff
@@ -55,15 +55,31 @@ __cursor_state_restore(WT_CURSOR *cursor, WT_CURFILE_STATE *state)
...
+	 * release the page so we get it again with history in that case.
 	 */
-	return (F_ISSET(cbt, WT_CBT_ACTIVE) &&
-	    (!eviction_ok || cbt->ref->page->read_gen != WT_READGEN_OLDEST));
+	if (cbt->ref->page->read_gen == WT_READGEN_OLDEST)
+		return (false);
+
+	if (cbt->ref->state == WT_REF_LIMBO &&
+	    F_ISSET(&session->txn, WT_TXN_UPDATE))
+		return (false);
```
michaelcahill: @sueloverso, if this is changed to:

```
	if (F_ISSET(&session->txn, WT_TXN_UPDATE) &&
            cbt->ref->state != WT_REF_MEM)
		return (false);
```

Does that address the issue with attempting to update a LIMBO page due to a transient change in the ref state?

keithbostic:
@sueloverso, e0fdd44 cleans up a bunch of failures for me (when doing a one-for-one page replacement, we weren't discarding the previous WT_REF lookaside information).

@michaelcahill, I think we can just blindly free `WT_REF.page_las` in that path, because there's no possibility that `WT_REF.page_del` can't be set. I'd appreciate your review of that, however.

sueloverso:
Thanks @keithbostic I will pull these changes and run all the different configs while gone this weekend.

keithbostic:
@sueloverso, to be clear, we're not there yet, I'm still seeing some failures. It would be great to know which CONFIGs still fail, though.

keithbostic:
autobuild: retest this please

keithbostic:
@michaelcahill, @sueloverso figured out the `timestamp04` failure (fix committed in c5c639e), but we were left with a couple of questions:

* Could we merge `WT_PAGE_LOOKASIDE.eviction_to_lookaside` and `WT_PAGE_LOOKASIDE.invalid`? I thought I could set `invalid = true` instead of freeing the lookaside structure in page-read or adding an additional variable, but that didn't appear to work.

* Shouldn't we be testing `WT_PAGE_LOOKASIDE.invalid` in the `WT_REF_LIMBO` case of `__page_read()`? That is, why do lookaside page instantiation, if `invalid == true`?

* And, finally, why isn't `__txn_rollback_to_stable_btree_walk()` testing for `WT_REF_LOOKASIDE as well as `WT_REF_LIMBO`?

I'll try and figure this out on my own tomorrow, but if you're around and feel like answering, that'd be spiffy.

src/evict/evict_page.c
```diff
@@ -268,9 +268,16 @@ __evict_page_clean_update(WT_SESSION_IMPL *session, WT_REF *ref, bool closing)
...
 	 * an address, it's a disk page; if it has no address, it's a deleted
 	 * page re-instantiated (for example, by searching) and never written.
+	 *
+	 * If evicting a WT_REF_LIMBO reference, we get to here and transition
+	 * back to WT_REF_LOOKASIDE, else discarding any lookaside information.
 	 */
 	__wt_ref_out(session, ref);
-	if (ref->addr == NULL) {
+	if (!closing && ref->page_las != NULL &&
+	    ref->page_las->eviction_to_lookaside == true) {
```
michaelcahill: Style nit: should just be `&& ref->page_las->eviction_to_lookaside`: avoid equality comparison with `true` or `false`.

src/txn/txn_rollback_to_stable.c
```diff
@@ -275,22 +259,24 @@ __txn_rollback_to_stable_btree_walk(
...
 
 	/* Walk the tree, marking commits aborted where appropriate. */
 	ref = NULL;
-	while ((ret = __wt_tree_walk_custom_skip(session, &ref,
-	    __txn_rollback_to_stable_custom_skip,
-	    NULL, WT_READ_NO_EVICT)) == 0 && ref != NULL) {
-		page = ref->page;
+	while ((ret = __wt_tree_walk(session, &ref,
+	    WT_READ_CACHE | WT_READ_LOOKASIDE | WT_READ_NO_EVICT)) == 0 &&
+	    ref != NULL) {
```
michaelcahill: I don't have all the details of rollback in my head at the moment, but won't this walk no longer visit refs with the `WT_REF_DELETED` state?  It looks like the code tried pretty hard to cover that case previously and AFAICT it no longer does.

I don't know if we have a test that combines `truncate` with a `rollback_to_stable` -- MongoDB's oplog truncates (the only use of truncate in MongoDB) should never be rolled back by reconcilation.

src/include/btmem.h
```diff
@@ -820,16 +827,14 @@ struct __wt_ref {
 #undef	ref_ikey
 #define	ref_ikey	key.ikey
 
-	union {
-		WT_PAGE_DELETED	*page_del;	/* Deleted page information */
-		WT_PAGE_LOOKASIDE *page_las;	/* Lookaside information */
-	};
+	WT_PAGE_DELETED	  *page_del;	/* Deleted page information */
+	WT_PAGE_LOOKASIDE *page_las;	/* Lookaside information */
 };
```
michaelcahill: As noted, it isn't great that `WT_REF` has grown, but I think this change or something very similar is going to be necessary for WT-3805, so I can live with it.

keithbostic:
autobuild: retest this please

src/txn/txn_rollback_to_stable.c
keithbostic: @michaelcahill 
>  but won't this walk no longer visit refs with the WT_REF_DELETED state?

I don't think so: the `WT_READ_CACHE` tests in the walk code appear before the `WT_REF_DELETED` checks, so I believe they will be visited.

There's also a question of what the underlying read code itself does, but again I think they will be visited.

keithbostic:
The e7d5c26 build failure is real:
```

sueloverso:
autobuild: retest this please

sueloverso:
There is no joy reproducing the assertion failure above. It is the only currently known outstanding issue on this branch. I've attempted over 700 iterations with the CONFIG in the ticket (which admittedly leaves most items to random). Do we hold this branch for that to repro or merge into develop and deal with it when it reappears?

keithbostic:
> I'd like to stare at it before it goes in, but if you/Alex/Michael disagree, happy for it to go in immediately.

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

sueloverso:
@michaelcahill I believe this is ready to merge. I've pulled it up with current develop also. All known failures are gone. Please merge if you agree.

michaelcahill:
Thanks @sueloverso and @keithbostic!  It's been a big effort to get this change over the line but it's looking good to me and in our testing.  I'll go ahead and merge...

PR #3814, opened 2017-11-28
WT-3763 Disable suffix compression on key with saved updates.

michaelcahill:


michaelcahill:
@keithbostic, running `timestamp04` and `test/csuite/test_timestamp_abort` both hit the assertion that we don't write empty pages during splits.  After some investigation, I noticed that the keys we were promoting had been suffix compressed and didn't contain enough bytes to correctly assign updates to `WT_MULTI` structures.

If this passes automated testing, I'm going to merge it because it at least avoids the issue in my local testing.  Can you please decide if this is the right fix?

michaelcahill:
This has past the relevant tests, merging to get wider coverage.

keithbostic:
>  `timestamp04` and `test/csuite/test_timestamp_abort` both hit the assertion that we don't write empty pages during splits.  After some investigation, I noticed that the keys we were promoting had been suffix compressed and didn't contain enough bytes to correctly assign updates to `WT_MULTI` structures.

@michaelcahill, I don't understand why this change is needed -- can you please walk me through the failure case?

(Alternatively, maybe it's time to just remove the suffix compression code? It used to be a pretty harmless chunk of code, but now that it requires a walk through (and comparison with), the saved update list, maybe it's not worth doing?)

agorrod:
@keithbostic What was happening was that two keys were ending up in the parent page with only a single byte of information after suffix compression. When we were going through the saved update list applying them to (eventually empty on disk) pages, there was an empty page that didn't have any saved updates either. Turned out that is because the previous page had already consumed the updates destined for this page.

I don't think this problem is enough to motivate removing suffix compression - it's clearly working in at least some workloads or we wouldn't have encountered the problem :)

keithbostic:
> What was happening was that two keys were ending up in the parent page with only a single byte of information after suffix compression.

Yeah, that really shouldn't be possible, and honestly, I didn't see the path where it can happen (which worries me). Maybe a different way to ask this question: what changed that allowed this to happen?

The reason I pushed for a6901a6 to work the way it does in [lines 5671-5680](https://github.com/wiredtiger/wiredtiger/blob/develop/src/reconcile/rec_write.c#L5671-L5680) was so this kind of change wouldn't be necessary, and I would like to understand what I missed.

This isn't urgent, of course, but I am concerned about it.

I'm also not feeling much love for suffix compression at this point: it only works on the first internal level, it means a search through saved updates (which we do a lot more of, these days), and it's always been a pain to make correct.

michaelcahill:
@keithbostic, if you revert the change, things were failing pretty quickly, e.g. https://jira.mongodb.org/browse/WT-3771 references http://build.wiredtiger.com:8080/job/wiredtiger/6319/console where running `timestamp_abort` resulted in:

```
[1511834284:944993][18875:0x7f1050ccf700], file:collection.wt, eviction-server: row-store leaf page at [write-check] has no entries
--
[1511834284:945035][18875:0x7f1050ccf700], eviction-server: cache eviction thread error: WT_ERROR: non-specific WiredTiger error
[1511834284:945043][18875:0x7f1050ccf700], eviction-server: the process must exit and restart: WT_PANIC: WiredTiger library panic
[1511834284:945048][18875:0x7f1050ccf700], eviction-server: aborting WiredTiger library
```

I also reproduced it occasionally running `timestamp04` and `timestamp07`.

PR #3816, opened 2017-11-29
WT-3763 Revert part of the change that made reconciliation more likely to use lookaside

nehakhatri5:


agorrod:
Thanks @nehakhatri5 lgtm - I'm going to merge.

PR #3817, opened 2017-11-29
WT-3773 Fix a bug calculating on-disk images for modify updates.

michaelcahill:
Previously, when evicting a page with modify updates, some of the modifications could be skipped when calculating the complete value to write to the new page.  This could lead to updates being lost.

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -40,9 +40,9 @@ typedef struct {
 
 	/*
 	 * Track the oldest running transaction and whether to skew lookaside
-	 * to the newest or oldest update.
+	 * to the newest update.
 	 */
-	bool las_skew_oldest;
+	bool las_skew_newest;
 	uint64_t last_running;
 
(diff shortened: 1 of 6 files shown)


src/reconcile/rec_write.c
```diff
@@ -40,9 +40,9 @@ typedef struct {
 
 	/*
 	 * Track the oldest running transaction and whether to skew lookaside
-	 * to the newest or oldest update.
+	 * to the newest update.
 	 */
-	bool las_skew_oldest;
```
michaelcahill: This may appear gratuitous, but I believe it fixes a real bug when checkpoint does eviction, where the resulting pages don't choose the oldest values but also don't choose the newest.  I made this change here because I thought this was involved in the `modify` issue.  lmk if it should be separated out.

agorrod:
Thanks @michaelcahill this change looks good to me. I'm going to merge once Windows testing finishes.

PR #3818, opened 2017-11-29
WT-3774 Add a test case that covers lookaside and modify

agorrod:


agorrod:
@sulabhM or @bvpvamsikrishna please review this test change.

sulabhM:
I will merge this change in.

PR #3821, opened 2017-12-01
WT-3587 Remove HAVE_VERBOSE conditional compilation

bvpvamsikrishna:
Not ready for review yet.
Need to check the performance implications.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -492,7 +492,7 @@ __wt_evict_create(WT_SESSION_IMPL *session)
 	 */
 #if !defined(HAVE_DIAGNOSTIC)
 	/* Need verbose check only if not in diagnostic build */
-	if (WT_VERBOSE_ISSET(session, WT_VERB_EVICTSERVER)
+	if (WT_VERBOSE_ISSET(session, WT_VERB_EVICTSERVER))
 #endif
 		__wt_epoch(session, &conn->cache->stuck_time);
 
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -425,6 +425,11 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 		return (0);
 	}
 
+#if !defined(HAVE_DIAGNOSTIC)
+	/* Need verbose check only if not in diagnostic build */
+	if (!WT_VERBOSE_ISSET(session, WT_VERB_EVICT_STUCK))
+		return (0);
+#endif
 	/*
 	 * If we're stuck for 5 minutes in diagnostic mode, or the verbose
 	 * evict_stuck flag is configured, log the cache and transaction state.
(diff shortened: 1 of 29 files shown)


michaelcahill:
@bvpvamsikrishna do you have performance results for this branch?

bvpvamsikrishna:
@sulabhM @nehakhatri5 request to review this pull request.

src/txn/txn_timestamp.c
```diff
@@ -65,25 +65,20 @@ __wt_timestamp_to_hex_string(
...
  */
 void
 __wt_verbose_timestamp(WT_SESSION_IMPL *session,
     const wt_timestamp_t *ts, const char *msg)
 {
-#ifdef HAVE_VERBOSE
 	char timestamp_buf[2 * WT_TIMESTAMP_SIZE + 1];
 
-	if (__wt_timestamp_to_hex_string(session, timestamp_buf, ts) != 0)
+	if (!WT_VERBOSE_ISSET(session, WT_VERB_TIMESTAMP) ||
```
sulabhM: this check is already done by __wt_verbose() called next. We will effectively be doing this test twice now.

bvpvamsikrishna: The reason why I added this is reduce performance impact by avoiding the execution of "if" statement before the __wt_verbose call.

bvpvamsikrishna:
@sulabhM Request to check the updates done.

src/evict/evict_lru.c
```diff
@@ -421,13 +417,14 @@ __evict_server(WT_SESSION_IMPL *session, bool *did_work)
 
 	/* Eviction is stuck, check if we have made progress. */
 	if (*did_work) {
-#if defined(HAVE_DIAGNOSTIC) || defined(HAVE_VERBOSE)
-		__wt_epoch(session, &cache->stuck_time);
+#if !defined(HAVE_DIAGNOSTIC)
+		/* Need verbose check only if not in diagnostic build */
+		if (WT_VERBOSE_ISSET(session, WT_VERB_EVICT_STUCK))
 #endif
+			__wt_epoch(session, &cache->stuck_time);
```
nehakhatri5: The function __evict_server gets called repeatedly for eviction pass. So having the __wt_epoch outside of diagnostic mode or in non-verbose mode might have performance impact on eviction thread. Has the eviction thread's performance been compared before and after this change?

bvpvamsikrishna:
@nehakhatri5 Request to check the updates.

src/txn/txn_timestamp.c
sulabhM: I don't have a good solution to this issue, I think it should be okay to check twice if WT_VERB_TIMESTAMP is set.
@agorrod can you have a look at this change here, what do you think.

bvpvamsikrishna:
@agorrod Request to review this pull request.

src/evict/evict_lru.c
agorrod: @nehakhatri5 We ran the Jenkins performance tests, which include a number that stress eviction and there was no noticeable change, so I think this one is safe.

src/txn/txn_timestamp.c
agorrod: I don't have a strong opinion on this. I think the code is more obvious with the explicit check, so I'm happy for it to stay the way it is here.

PR #3822, opened 2017-12-01
WT-3079 Resume eviction walks per tree.

michaelcahill:
Record how many pages we want and how many pages we have queued so far in a tree, then resume the walk next iteration.  This avoids a single tree with a target larger than the queue size being walked completely before the eviction server moves on to the next tree.


agorrod:
Thanks @michaelcahill this change looks good to me - and the code is clearer now so I'm a fan. The perf plots are looking good - I'm going to go ahead and merge.

PR #3823, opened 2017-12-01
WT-3783 Fix transaction isolation to use the correct enum

keithbostic:
@michaelcahill, can you please look at dfd2c6f?

I think we're setting `las_session->txn.isolation = WT_ISO_SNAPSHOT`, because `WT_TXN_ISO_READ_UNCOMMITTED` == 2.

### DIFF
--- a/dist/s_style
+++ b/dist/s_style
@@ -61,6 +61,15 @@ else
 		cat $t
 	fi
 
+	if ! expr "$f" : 'examples/c/.*' > /dev/null &&
+	   ! expr "$f" : 'ext/.*' > /dev/null &&
+	   ! expr "$f" : 'src/include/wiredtiger_ext\.h' > /dev/null &&
+	   ! expr "$f" : 'src/txn/txn_ext\.c' > /dev/null &&
+	   grep WT_TXN_ISO_ $f; then
+		echo "$f: WT_TXN_ISO_XXX constants only for the extension API"
+		cat $t
+	fi
+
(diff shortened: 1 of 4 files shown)


michaelcahill:
@keithbostic, your change is correct -- I think we need something like this as well:

```
diff --git a/dist/s_style b/dist/s_style
index 3ae14949f..78810dc95 100755
--- a/dist/s_style
+++ b/dist/s_style
@@ -61,6 +61,15 @@ else
                cat $t
        fi

+       if ! expr "$f" : 'examples/c/.*' > /dev/null &&
+          ! expr "$f" : 'ext/.*' > /dev/null &&
+          ! expr "$f" : 'src/include/wiredtiger_ext\.h' > /dev/null &&
+          ! expr "$f" : 'src/txn/txn_ext\.c' > /dev/null &&
+           grep WT_TXN_ISO_ $f; then
+               echo "$f: WT_TXN_ISO_* constants are only for the extension API"
+               cat $t
+       fi
+
        if ! expr "$f" : 'src/include/queue\.h' > /dev/null &&
            egrep 'STAILQ_|SLIST_|\bLIST_' $f ; then
                echo "$f: use TAILQ for all lists"
```

There is sane checking of enum usage in C++ but AFAICT, no combination of warning flags will catch this kind of thing in C.

keithbostic:
@michaelcahill, 
>  I think we need something like this as well

Agreed, pushed into the branch.

PR #3825, opened 2017-12-02
WT-3776 Segmentation fault during cursor remove operation

keithbostic:


src/btree/bt_cursor.c
```diff
@@ -57,10 +57,10 @@ __cursor_state_restore(WT_CURSOR *cursor, WT_CURFILE_STATE *state)
  * grown too large).
  */
 static inline bool
-__cursor_page_pinned(WT_CURSOR_BTREE *cbt)
+__cursor_page_pinned(WT_CURSOR_BTREE *cbt, bool eviction_ok)
```
nehakhatri5: With this change,  this function could say that a page is pinned even if it is not flagged for forced eviction.  The function header comment currently does not indicate that.
Is it because the comment in the function __wt_btcur_remove is sufficient to explain that situation and it is fine to leave the function header like that?

keithbostic: > The function header comment currently does not indicate that.

Thank you, @nehakhatri5, I've tried to fix this in d99810f.

I personally prefer simple comments at the function declaration, and then more detailed comments in the function body so that comments are close to the behavior they describe.

nehakhatri5: Thank you. Looks good.

agorrod:
Thanks @keithbostic this change lgtm as well.

PR #3826, opened 2017-12-05
WT-3786 Transactions should read their writes regardless of timestamps.

michaelcahill:


agorrod:
This change lgtm @michaelcahill - I'm happy for it to be merged if @milkie confirms it fixes the problem he sees.

michaelcahill:
Thank @agorrod -- I wanted to add @dgottlieb as a reviewer but first he has to accept an invitation to join the wiredtiger group...

PR #3827, opened 2017-12-05
WT-3597 Timestamp diagnostic when inserting older key

sueloverso:
This is not yet ready for review.


sueloverso:
@agorrod This is ready for review.

agorrod:
Thanks @sueloverso lgtm

PR #3829, opened 2017-12-07
WT-3787 test_compact02 failed as compaction halted due to eviction pressure

keithbostic:
@ddanderson, is there a way to change this code to only ignore failure in the case of `EBUSY`, rather than ignoring all errors?


ddanderson:
@keithbostic, I added checking for 'optional' exceptions that match a certain string, and added a convenience function for handling the EBUSY case - seems like it may come up elsewhere.  You can merge if it passes tests and it works for you.

keithbostic:
Thank you, @ddanderson!

PR #3830, opened 2017-12-08
WT-3793 WiredTiger page debug dump functions should unpack integer keys

keithbostic:
Change page debug dump functions for integer keys so we see:
```
0x801c518d0: row-store leaf
        disk 0x807480000, dsk_mem_size 1066, entries 99, clean, memory_size 1938, disk-alloc
        K {1}
        V {VALUE-H}
```
instead of:
```
0x801c518d0: row-store leaf
        disk 0x807480000, dsk_mem_size 1066, entries 99, clean, memory_size 1938, disk-alloc
        K {#81}
        V {VALUE-H}
```

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -162,6 +162,7 @@ HFS
 HHHH
 HHHHLL
 HHHLL
+HILQr
 Hendrik
 HyperLevelDB
 ID's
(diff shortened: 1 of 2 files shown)


agorrod:
This looks like a good change. @bvpvamsikrishna could you review as well? You were using this code recently.

bvpvamsikrishna:
lgtm
I think it will be great if we can extend debug dump functions to print even composite keys in more readable form.

PR #3832, opened 2017-12-08
WT-3795 lint cleanups for the op-tracking software, reduce record write size.

keithbostic:
@agorrod, @fedorova, these are the changes to get the static analysis software to quit complaining about how we were using `__func__`. Sorry about the mess. There is a bit of carrot here, though, I think I've reduced the size of the op-tracking write record from 24B to 16B.

The only thing I don't like about this change is the need for `WT_TRACK_OP_DECL` (because `WT_TRACK_OP_INIT` can't always immediately follow the function declarations), but I don't see any alternative.


agorrod:
autobuild: retest this please

PR #3833, opened 2017-12-09
WT-3716 Restore the WT_VERB_TEMPORARY verbose flag.

keithbostic:
@agorrod, here's the rest of the WT-3720/WT-3716 change, putting the temporary verbose flag back in, and moving the verbose flags from 32 to 64 bits

Just for clarity, I redid the change from scratch by adding the verbose flag and then looking for compiler/lint complaints. The compiler/lint checks found everything I needed to change, so I think there's some evidence that we made the correct design choices here.


agorrod:
Thanks @keithbostic lgtm

PR #3837, opened 2017-12-12
WT-3790 Add translation from rdtsc ticks to nanoseconds.

sueloverso:
@agorrod, here is one implementation to convert rdtsc to nanoseconds to make the time comparisons useful. Please review. In addition I replaced places where we do intra-function timings for stats in several locations. There are other replacements in the WT-3725 branch that are disjoint from these. 

Here I searched for "timespec start, stop" and in WT-3725 I searched for "timespec enter, leave". Once this is merged, in WT-3725 I'll make the names consistent.


src/conn/conn_api.c
```diff
@@ -1285,6 +1285,35 @@ __conn_rollback_to_stable(WT_CONNECTION *wt_conn, const char *config)
...
+{
+#if (defined __i386) || (defined __amd64)
+	struct timespec start, stop;
+	uint64_t diff_nsec, diff_tsc;
+	uint64_t i, tsc_start, tsc_stop;
+
+	__wt_epoch(session, &start);
+	tsc_start = __wt_rdtsc(session);
+	/*
+	 * This needs to be CPU intensive and large enough.
```
agorrod: Is the loop for a million operations enough? I suspect the compiler would just remove that entirely unless we do something more involved. It's neat to have this in a single function call, but maybe we could wrap it around some real work done by `wiredtiger_open`? Maybe starting https://github.com/wiredtiger/wiredtiger/blob/develop/src/conn/conn_api.c#L2352 and ending around https://github.com/wiredtiger/wiredtiger/blob/develop/src/conn/conn_api.c#L2599

src/conn/conn_api.c
```diff
@@ -1285,6 +1285,35 @@ __conn_rollback_to_stable(WT_CONNECTION *wt_conn, const char *config)
...
 }
 
+/*
+ * __conn_calibrate_ticks --
+ *	Calibrate a ratio from rdtsc ticks to nanoseconds.
+ */
+static void
+__conn_calibrate_ticks(WT_SESSION_IMPL *session)
+{
+#if (defined __i386) || (defined __amd64)
```
agorrod: Does it make sense to add a HAVE_RDTSC instead of using this pair of flags?

src/btree/bt_io.c
```diff
@@ -183,13 +182,15 @@ __wt_bt_write(WT_SESSION_IMPL *session, WT_ITEM *buf,
 	WT_KEYED_ENCRYPTOR *kencryptor;
 	WT_PAGE_HEADER *dsk;
 	size_t dst_len, len, result_len, size, src_len;
+	uint64_t start, stop;
```
agorrod: How did you choose these as a use case to switch over to `rdtsc`?

sueloverso: As mentioned in the initial comment in the PR. I looked for ones that are intra-function timings and that I had not already done in WT-3725. I think I'll make this ticket changes complete on their own and remove the changes from the other PR.

src/conn/conn_api.c
sueloverso: I took the defines from `__wt_rdtsc()`. That is a possibility. I'd like @fedorova's opinion. In particular, my investigation makes me wonder/think if we should be using `rdtscp` instead.

src/conn/conn_api.c
sueloverso: My investigation in the ticket indicates that we really want something self-contained and clearly CPU intensive, with no option of I/O or other interference that can lead to inconsistent measurements. A million seemed sufficient in my experiments giving a large enough number of ticks to compute the ratio.

sueloverso: You are correct, good catch. If I declare `i` as `volatile` however it works as expected and that loop does not get optimized away.  I tested various values for the loop, from 1M to 1B and they all calculated the same ratio out to 4 decimal places. But there was a noticeable delay with 1B.  I think for our level of accuracy, 1M will be sufficient.

src/conn/conn_api.c
fedorova: @sueloverso, rdtscp prevents the hardware from reordering the rdtscp instruction with respect to other instructions, surrounding it. This would be important if we relied on capturing very fine-grained nanosecond-lasting events, but in our case we are interested in events that last milliseconds, or perhaps microseconds. So using rdtscp is not that important. There are two downsides for using rdtscp: (1) it has a much higher overhead in comparison to rdtsc, (2) it is available on fewer processors. That's why I chose to use rdtsc in the timing function.

sueloverso:
@agorrod I ran the Jenkins perf jobs with this branch today and they are either the same or some are 5-10% improved (particularly read throughput on both evict-lsm and evict-btree).

fedorova:
@sueloverso, I see that you start and stop the timer around certain functions like locks and I/O. This seems equivalent to using WT_TRACK_OP_INIT and WT_TRACK_OP_END inside those functions, just like this is currently done in the evict_lru_pages, for example. The benefit of using the WT_TRACK_OP* macros is that soon we will have the tools to conveniently view the data generated by them.

src/conn/conn_api.c
```diff
@@ -1285,6 +1285,35 @@ __conn_rollback_to_stable(WT_CONNECTION *wt_conn, const char *config)
...
+	tsc_stop = __wt_rdtsc(session);
+	__wt_epoch(session, &stop);
+	diff_nsec = WT_TIMEDIFF_NS(stop, start);
+	diff_tsc = tsc_stop - tsc_start;
+	S2C(session)->tsc_nsec_ratio = (double)diff_tsc/(double)diff_nsec;
+#else
+	S2C(session)->tsc_nsec_ratio = 1.0;
+#endif
+}
+
```
fedorova: @sueloverso, I like the idea of using a loop for determining the ratio of clock ticks to nanoseconds. Based on my investigation, there is no cross-platform way to do this conversion reliably: for example the method suggested for OS X is broken and even the stack overflow community had no answer.

I do not know if one million iterations is sufficient though.

src/conn/conn_api.c
sueloverso: Thanks for those comments. I agree we do not need the precision of very fine-grained accuracy and higher overhead is definitely something we want to avoid. So I am good staying with rdtsc.

src/conn/conn_api.c
sueloverso: I ran with 1M, 10M and 1B and they all computed the same ratio out to the 4th decimal place.  There was a noticeable delay in `wiredtiger_open` with 1B.

sueloverso:
@fedorova while I agree there exists some overlap in the op tracking and timing around locks and I/O, this change is more rote - it is only modifying existing `wt_epoch` calls to use the faster call (and renaming the variables). I think it should be a separate project/ticket to find and fix overlap at a future time.

sueloverso:
@agorrod one thing that occurs to me is that I did not modify long-term timers that exist inside other structures. Some examples are `cache->stuck_time` and `lsm_tree->last_flush_time` and `conn->ckpt_timer_start`. Do we want to move everything not requiring date/time information (such as statistics logging) to this?

src/conn/conn_api.c
agorrod: I've created WT-3802 to track moving this to HAVE_RDTSC in WT-3802

PR #3840, opened 2017-12-14
WT-3806 Make sure rdtsc values move forward in time

sueloverso:
Not quite ready for review. Needs additional testing.


sueloverso:
This is ready for review.  I tested it in conjunction with WT-3725 to make sure it didn't add overhead. There was a very slight decline over not having it at all (about 1%), which is nearly statistically insignificant. But I think it is necessary as rdtsc can be reset.

src/include/misc.i
```diff
@@ -38,17 +54,16 @@ __wt_rdtsc(WT_SESSION_IMPL *session) {
...
-	uint64_t a, d;
-
-	WT_UNUSED(session);
+	uint64_t a, d, x;
 
 	__asm__ volatile ("rdtsc" : "=a" (a), "=d" (d));
-	return ((d << 32) | a);
+	x = (d << 32) | a;
+	__tsc_check_monotonic(session, &x);
+	return (x);
```
keithbostic: We're not checking the `__wt_optrack_get_expensive_timestamp()` path in `__wt_rdtsc()` for time-travel, is that not possible?

If time travel is possible, I'd suggest changing `__wt_optrack_get_expensive_timestamp()` to call `__wt_epoch()` instead of `__wt_epoch_raw()`.

src/include/os.h
```diff
@@ -66,7 +66,7 @@
 	(WT_TIMEDIFF_NS((end), (begin)) / WT_BILLION)
 
 #define	WT_TSCDIFF_NS(s, end, begin)					\
-	(__wt_tsc_to_nsec((s), (uint64_t)(end) - (uint64_t)(begin)))
+	(__wt_tsc_to_nsec((s), (end), (begin)))
```
keithbostic: I don't think we need to parenthesize `s`, `end` or `begin`.

There are other, close-by time macros with similar unnecessary parenthesis.

src/include/misc.i
keithbostic: I noticed `__wt_rdtsc()` has compiler-specific assembly code, which has historically been put into `include/gcc.h` or `include/msvc.h`, but this particular assembly code is shared between gcc/clang and MSVC.

There's also x86 #ifdef's in `__conn_calibrate_ticks()` in `conn_api.c`, which bothers me, we don't want architecture specific code climbing up into "portable" code.

@michaelcahill, @agorrod, any thoughts on where x86-specific code like this should land?

keithbostic:
> But I think it is necessary as rdtsc can be reset.

@sueloverso, I thought rdtsc can only be reset when the processor is reset -- is there some other way?

sueloverso:
> @sueloverso, I thought rdtsc can only be reset when the processor is reset -- is there some other way?

Anytime the processor hibernates or sleeps the rdtsc is reset. I can easily cause wildly different tsc values to happen by simply closing my laptop at the right time while running an example program.  While I've pushed the KNF related changes, I need to come up with a different way for calculations. It doesn't feel like stalling the tsc reading until it catches up is a good way because a reset value is not related to the original value in the same way a clock time adjustment is.

The infrastructure here is still needed. But I think the mechanism needs rethinking.

sueloverso:
@agorrod I took a different approach to this. See what you think.

agorrod:
Thanks @sueloverso lgtm - I'll go ahead and merge.

PR #3841, opened 2017-12-14
WT-3767 Avoid lookaside reads when history is not required for queries.

keithbostic:


sueloverso:
@keithbostic and I chatted about this issue today. I have restarted the original CONFIG that showed the assertion failure (which is now gone in this branch).

sueloverso:
Looks like the test_timestamp07(table-cg.nolog.1000keys) hung (at least the one on tinderbox). I am not able to repro any hang in timestamp07 on my AWS box though.

michaelcahill:
Thanks @keithbostic, lgtm.  Building the special case directly into the remove code seems better to me than overloading what it means for a cursor to be positioned.  I'll merge this into the parent.

PR #3820, opened 2017-11-30
WT-3769 Enable prefix compression by default for LAS

bvpvamsikrishna:

PR #3843, opened 2017-12-17
WT-3769 Fix a bug in reverse cursor walks with prefix compression enabled

keithbostic:
@michaelcahill, please review this one carefully, I've repeatedly gotten it wrong.

Once it's merged, I'll update the WT-3767 (#3813) branch, they aren't going to merge cleanly.

There is a related branch that includes tests, see #3820.


keithbostic:
@michaelcahill, I think this one is ready to go, assuming it passes the pull-request tests.

We can probably figure out a simpler approach to making lookaside table prefix compression work, let me know if I should do that.

michaelcahill:
@keithbostic, sorry I didn't get to this today -- it's first on my list tomorrow.

agorrod:
@keithbostic this change lgtm to me, but Michael should definitely do a review as well.

I'd like to split this bug fix out from enabling prefix compression on the lookaside table, since the fix isn't really related to lookaside. Would you mind repurposing WT-3769 to reflect this bug fix, and creating a new ticket to track enabling prefix compression on the lookaside table (or vice versa)?

keithbostic:
> I'd like to split this bug fix out from enabling prefix compression on the lookaside table, since the fix isn't really related to lookaside. Would you mind repurposing WT-3769 to reflect this bug fix, and creating a new ticket to track enabling prefix compression on the lookaside table (or vice versa)?

I was going down this path myself, but looking at it again today and to figure out the minimal change, I think we might want to take the whole change, to minimize future backport effort.

* The changes to `__cursor_page_pinned()` are just moving the test for `WT_CURSTD_KEY_INT` into the underlying function instead of doing it at a higher level, and adding comments to explain what we're doing.
* There's a switch from `WT_CURSTD_KEY_INT` to `WT_CURSTD_KEY_EXT`, but that's limited to column-store, so it's safe.
* Then there's the change in cursor remove positioning, in `__wt_btcur_remove()`, and that needs to be part of making prefix-compression work.
* And then there's changes in the upper-level cursor code, but that's just adding asserts that we see what we expect to see on cursor function return.

In other words, all the dangerous stuff has to come into the new ticket anyway. Thoughts?

agorrod:
Sorry @keithbostic - I've made the change I was thinking, which is to move the changes in btmem.h into a new ticket - WT-3816. This branch looks good to me now, once it passes testing I'll go ahead and merge.

PR #3845, opened 2017-12-18
WT-3809 Fix a bug in lookaside related to birthmarks

keithbostic:
@michaelcahill, my guess is this can happen but we're not seeing the failure because the compiler is reversing the tests as they don't contain any side-effects?

The comment leading into the return if/else pattern could use some work as well.


src/btree/bt_ret.c
```diff
@@ -180,8 +180,10 @@ __wt_value_return_upd(WT_SESSION_IMPL *session,
 			continue;
 
 		if (!ignore_visibility && !__wt_txn_upd_visible(session, upd)) {
-			if (upd->type == WT_UPDATE_BIRTHMARK)
+			if (upd->type == WT_UPDATE_BIRTHMARK) {
 				skipped_birthmark = true;
+				break;
```
michaelcahill: I don't believe `break` is correct, here.  If we subsequently see an update, we should take it.

src/btree/bt_ret.c
```diff
@@ -216,7 +218,9 @@ __wt_value_return_upd(WT_SESSION_IMPL *session,
 	 * If we hit the end of the chain, roll forward from the update item we
 	 * found, otherwise, from the original page's value.
 	 */
-	if (upd == NULL && !skipped_birthmark) {
+	if (skipped_birthmark)
+		WT_ERR(__wt_buf_set(session, &cursor->value, "", 0));
```
michaelcahill: This case is already worrying for a tombstone (we should return `WT_NOTFOUND` so if we get here it's probably masking an error).

I understand the bug fix, but would it be reasonable to assert that neither case ever happens?

michaelcahill: Also, as mentioned above, if we find a visible update, we should take it, and only check whether we have skipped a birthmark if `upd == NULL`.

src/btree/bt_ret.c
keithbostic: > I don't believe `break` is correct, here.  If we subsequently see an update, we should take it.

That's not what the existing code does (unless I'm missing something).

If `skipped_birthmark` is set, we will always take the second if/else branch:
```
/*
 * If we hit the end of the chain, roll forward from the update item we
 * found, otherwise, from the original page's value.
 */
if (upd == NULL && !skipped_birthmark) {
        /*
         * Callers of this function set the cursor slot to an impossible
         * value to check we're not trying to return on-page values when
         * the update list should have been sufficient (which happens,
         * for example, if an update list was truncated, deleting some
         * standard update required by a previous modify update). Assert
         * the case.
         */
        WT_ASSERT(session, cbt->slot != UINT32_MAX);

        WT_ERR(__value_return(session, cbt));
} else if (upd->type == WT_UPDATE_TOMBSTONE || skipped_birthmark)
        WT_ERR(__wt_buf_set(session, &cursor->value, "", 0));
else
        ...
```

src/btree/bt_ret.c
keithbostic: > This case is already worrying for a tombstone (we should return WT_NOTFOUND so if we get here it's probably masking an error).
> I understand the bug fix, but would it be reasonable to assert that neither case ever happens?

I don't understand what you're trying to say, here.

keithbostic:
@michaelcahill, I reworked the change to reflect we should ignore skipped birthmarks unless we don't find a visible update -- can you take a run at the additional restrictions? The tighter we tie this down, the better.

It's a little worrisome that we didn't see a test failure from the skipped birthmarks bug.

michaelcahill:
Thank @keithbostic, lgtm -- I'll merge.

I tried to say this earlier but don't think I was very clear.  The case touched by this change sets a cursor's value to empty if no value is visible.  However, my understanding is that we should be catching that at a higher level (i.e., `__wt_cursor_valid` should be called before we get here).

In other words, this code should only exist as a backstop / paranoia case and should not be executed unless there is a bug somewhere in WiredTiger (because returning an empty value to the application in this case is not okay).  That's why I'm not losing sleep over this not being caught by testing.

PR #3846, opened 2017-12-18
WT-3810 Add repeated tries to calibrate and avoid divide by zero.

sueloverso:
@agorrod This branch hopefully fixes the divide by zero build failure. In that, I could only come up with two scenarios where we could divide by zero: 1. the loop of only 1 million was so short it didn't register and 2. somehow even with volatile, the loop was optimized away.

I've addressed the first one by increasing the loop to 100M, and added an outer loop to calibrate three times, to avoid one iteration getting a scheduling issue. The idea is that weird scheduling issues won't happen every single iteration through this tight loop and we'll get at least one good reading. I cannot reproduce nor verify whether the second situation is happening. If you have other suggestions instead of `volatile` for not allowing optimization, let me know because that is supposed to work.


sueloverso:
autobuild: please retest this

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

src/conn/conn_api.c
```diff
@@ -1294,21 +1294,41 @@ __conn_calibrate_ticks(WT_SESSION_IMPL *session)
...
 #if (defined __i386) || (defined __amd64)
 	struct timespec start, stop;
-	uint64_t diff_nsec, diff_tsc, tsc_start, tsc_stop;
+	uint64_t diff_nsec, diff_tsc, min_nsec, min_tsc;
+	uint64_t tries, tsc_start, tsc_stop;
 	volatile uint64_t i;
 
-	__wt_epoch(session, &start);
-	tsc_start = __wt_rdtsc(session);
+	min_nsec = min_tsc = 0;
```
agorrod: It would make me feel better if `min_nsec` was initialized to 1 - then we could never end up with a divide by 0 from this code.

keithbostic:
> I don't have a better suggestion than using volatile.

Volatile should work -- if we think it's not working we could use a memory barrier, possibly with a no-op function call.

src/conn/conn_api.c
```diff
@@ -1294,21 +1294,41 @@ __conn_calibrate_ticks(WT_SESSION_IMPL *session)
...
+			min_nsec = diff_nsec;
+		else
+			min_nsec = WT_MIN(min_nsec, diff_nsec);
+		if (min_tsc == 0)
+			min_tsc = diff_tsc;
+		else
+			min_tsc = WT_MIN(min_tsc, diff_tsc);
+	}
+	WT_ASSERT(session, min_nsec != 0 && min_tsc != 0);
+	S2C(session)->tsc_nsec_ratio = (double)min_tsc/(double)min_nsec;
```
keithbostic: The operator `/` needs a space.

src/conn/conn_api.c
sueloverso: I came up with another potential cause for the original problem, or really a form of the first one: the system's clock has insufficient granularity to register nanoseconds. Hopefully increasing the loop by 100x alleviates that. I have mixed feelings on your suggestion because although it avoids the divide by zero we have no idea that the algorithm is not effective.  We'll just compute inaccurate values/time differences later. But having a divide by zero in the field is not a good thing.

sueloverso:
@agorrod and @keithbostic I made new changes. If the calibrate function cannot get a good reading then we set a flag and `wt_rdtsc` falls back to using `wt_epoch`.

agorrod:
Thanks @sueloverso lgtm, I'm going to merge.

PR #3847, opened 2017-12-20
WT-3632 Increase how granularly cache usage settings can be configured

nehakhatri5:
Now on the cache config param value greater than 100 is considered an absolute
value for that parameter. Hence more granular values can be configured for cache 
usage settings.


nehakhatri5:
@sulabhM @bvpvamsikrishna Please review the code change here.

src/config/config_def.c
```diff
@@ -1090,16 +1098,18 @@ static const WT_CONFIG_CHECK confchk_wiredtiger_open_usercfg[] = {
...
 	    NULL, 0 },
 	{ "eviction_dirty_target", "int",
-	    NULL, "min=1,max=99",
+	    NULL, "min=1,max=10TB",
 	    NULL, 0 },
 	{ "eviction_dirty_trigger", "int",
-	    NULL, "min=1,max=99",
+	    NULL, "min=1,max=10TB",
+	    NULL, 0 },
+	{ "eviction_target", "int", NULL, "min=10,max=10TB", NULL, 0 },
```
bvpvamsikrishna: can make eviction_target also same as other parameters, in multiple lines.

dist/api_data.py
```diff
@@ -454,29 +454,35 @@ def __cmp__(self, other):
...
     Config('eviction_checkpoint_target', '5', r'''
         perform eviction at the beginning of checkpoints to bring the dirty
-        content in cache to this level, expressed as a percentage of the total
-        cache size.  Ignored if set to zero or \c in_memory is \c true''',
-        min=0, max=99),
+        content in cache to this level, absolute when greater than 100,
+        otherwise expressed as a percentage of the total cache size, should not
+        exceed the \c cache_size. Ignored if set to zero or \c in_memory is \c
+        true''',
+        min=0, max='10TB'),
```
bvpvamsikrishna: previously this parameter can max be specified as 99% of cache size, but now if user specifies as 100% error will not be returned, I think this is a change in behaviour, we need to return same error if users specifies as 100%, subsequent logic is also not handling this, I think this has to be taken care.

dist/api_data.py
```diff
@@ -454,29 +454,35 @@ def __cmp__(self, other):
...
     Config('eviction_dirty_target', '5', r'''
         perform eviction in worker threads when the cache contains at least
-        this much dirty content, expressed as a percentage of the total cache
-        size.''',
-        min=1, max=99),
+        this much dirty content, absolute when greater than 100, otherwise
+        expressed as a percentage of the total cache size, should not exceed the
+        \c cache_size''',
+        min=1, max='10TB'),
```
bvpvamsikrishna: above comment applies to this as well as subsequent one's also.

dist/api_data.py
sulabhM: maybe we can change wording and the behaviour to say "absolute when greater than or equal to 100" . so If you provide 100, it gets treated as 100 bytes? If I provide "1000", what unit is it treated? I guess that will be bytes.

src/config/config_def.c
sulabhM: this file is auto generated .

src/conn/conn_cache.c
```diff
@@ -8,6 +8,32 @@
...
+	conn = S2C(session);
+
+	/*
+	 * Anything above 100 is an absolute value; convert it to percentage.
+	 * An absolute value should not exceed the cache size.
+	 */
+	if (*param > 100.0) {
+		if (*param > conn->cache_size)
+			WT_RET_MSG(session, EINVAL,
+			    "%s should be lower than cache size!", param_name);
```
sulabhM: We don't use exclamation mark in any other message. Can drop it, I do not feel this one strongly.

src/conn/conn_cache.c
```diff
@@ -8,6 +8,32 @@
...
+ *	Convert the absolute value in cache config to percentage.
+ */
+static inline int
+__cache_config_abs_to_pct(WT_SESSION_IMPL *session,
+    double *param, const char *param_name)
+{
+	WT_CONNECTION_IMPL *conn;
+
+	conn = S2C(session);
+
```
sulabhM: should we check for param to be not NULL before dereferencing.

src/conn/conn_cache.c
```diff
@@ -8,6 +8,32 @@
...
+	/*
+	 * Anything above 100 is an absolute value; convert it to percentage.
+	 * An absolute value should not exceed the cache size.
+	 */
+	if (*param > 100.0) {
+		if (*param > conn->cache_size)
+			WT_RET_MSG(session, EINVAL,
+			    "%s should be lower than cache size!", param_name);
+		*param = (100.0 * *param) / (conn->cache_size + 1);
```
sulabhM: will "*param * 100.0" be more readable. Can we reach here before setting the cache size, I don't think that is possible. cache_size shouldn't be zero here. Maybe we can assert to that instead of adding 1 to cache_size. If planning to keep it this way, please add a comment stating why adding 1 to cache size here.

src/evict/evict_lru.c
```diff
@@ -595,7 +595,8 @@ __evict_update_work(WT_SESSION_IMPL *session)
 	dirty_inuse = __wt_cache_dirty_leaf_inuse(cache);
 	if (__wt_eviction_dirty_needed(session, NULL))
 		F_SET(cache, WT_CACHE_EVICT_DIRTY | WT_CACHE_EVICT_DIRTY_HARD);
-	else if (dirty_inuse > (cache->eviction_dirty_target * bytes_max) / 100)
+	else if (dirty_inuse > (uint64_t)
```
sulabhM: Not sure what coding conventions say about indentation here. @agorrod should take a final look at it.

src/include/cache.i
```diff
@@ -339,9 +339,9 @@ __wt_eviction_needed(
 	 * we involve the application thread.
 	 */
 	if (pct_fullp != NULL)
-		*pct_fullp = (u_int)WT_MAX(0, 100 - WT_MIN(
-		    (int)cache->eviction_trigger - (int)pct_full,
-		    (int)cache->eviction_dirty_trigger - (int)pct_dirty));
+		*pct_fullp = (double)WT_MAX(0.0, 100.0 - WT_MIN(
```
sulabhM: "(double)WT_MAX" that casting seems like redundant, since both arguments to WT_MAX are double. The arguments to WT_MIN, involve subtraction where one operand is double, so the other should implicitly cast to double, and you shouldn't need an explicit casting.

src/conn/conn_cache.c
nehakhatri5: WiredTiger has a provision to provide a cache pool that can be shared between multiple connections. In that scenario the the cache is dynamically allocated for each connection and there are chances of having cache_size 0 for a specific connection. For such cases we want to avoid divide by 0. I will add the comment to indicate that.

dist/api_data.py
nehakhatri5: @bvpvamsikrishna  Thanks for bringing up. In my discussion about this with Alex, we thought that a value above 100 should be considered an absolute value. So I considered that 100 should be treated as percentage. I agree that with this change, if 100 is specified that would mean 100% and it won't give an error like earlier. 
@agorrod What would be your recommendation on this.
@sulabhM Yes, 1000 will have a unit of bytes but in current shape 100 would be 100%. I may make a change based on the final decision here.

bvpvamsikrishna: @nehakhatri5  May be I am not clear in my comment.  What I mean is the value of these parameters should not be same as cache size, instead they should be less than that, but I think in this logic we accept the value of eviction_target to be same as cache size.  for example, as a user if I specify cache_size and eviction target as "1 TB", both the values as same we will process and continue, but previously user will not be able to do this as max value accepted by eviction_target is 99 only.

bvpvamsikrishna: also if absolute value of eviction_target is greater than cache_size, how to detect and return error, for example if cache_size = 1 TB and eviction_target = 2 TB, do we return error ?

nehakhatri5:
@sulabhM @bvpvamsikrishna Please have another look at the updated code and let me know if further changes are required.

src/conn/conn_cache_pool.c
```diff
@@ -674,8 +674,9 @@ __cache_pool_adjust(WT_SESSION_IMPL *session,
 			 * potentially a negative feedback loop in the
 			 * balance algorithm.
 			 */
-			smallest = (100 * __wt_cache_bytes_inuse(cache)) /
-			    cache->eviction_trigger;
+			smallest = (uint64_t)
```
sulabhM: is this code structuring okay, please confirm with Alex

src/docs/upgrading.dox
```diff
@@ -28,6 +28,15 @@ The performance visualization tool \c wtstats has been removed and is
 no longer supported.
 </dd>
 
+<dt>::wiredtiger_open cache configuration changes</dt>
```
sulabhM: Should we also say that existing configurations/setups do not need to do anything or worry about it.

src/evict/evict_lru.c
sulabhM: Review this indentation with Alex again

src/evict/evict_lru.c
```diff
@@ -610,10 +617,9 @@ __evict_update_work(WT_SESSION_IMPL *session)
 	 * Scrub dirty pages and keep them in cache if we are less than half
 	 * way to the clean or dirty trigger.
 	 */
-	if (bytes_inuse < ((cache->eviction_target + cache->eviction_trigger) *
-	    bytes_max) / 200 && dirty_inuse < (uint64_t)
-	    ((cache->eviction_dirty_target + cache->eviction_dirty_trigger) *
-	    bytes_max) / 200)
+	if (bytes_inuse < (uint64_t)((target + trigger) * bytes_max) / 200 &&
+	    dirty_inuse <
```
sulabhM: Review this indentation with Alex again

src/evict/evict_lru.c
```diff
@@ -626,9 +632,8 @@ __evict_update_work(WT_SESSION_IMPL *session)
 	if (!F_ISSET(conn, WT_CONN_EVICTION_NO_LOOKASIDE) &&
 	    (__wt_cache_stuck(session) ||
 	    (__wt_cache_lookaside_score(cache) > 80 &&
-	    dirty_inuse > (uint64_t)
-	    ((cache->eviction_dirty_target + cache->eviction_dirty_trigger) *
-	    bytes_max) / 200)))
+	    dirty_inuse >
```
sulabhM: review this indentation with Alex again

src/conn/conn_cache_pool.c
sueloverso: The cast should be on the same line as the thing it is casting. If needed, break the line after the `=` instead.

src/evict/evict_lru.c
sueloverso: Again the cast should be with the item. Break after the `>` if needed.

src/evict/evict_lru.c
```diff
@@ -557,11 +557,17 @@ __evict_update_work(WT_SESSION_IMPL *session)
...
+	double dirty_target, dirty_trigger, target, trigger;
 	uint64_t bytes_inuse, bytes_max, dirty_inuse;
 
 	conn = S2C(session);
 	cache = conn->cache;
 
+	dirty_target = cache->eviction_dirty_target;
+	dirty_trigger = cache->eviction_dirty_trigger;
+	target = cache->eviction_target;
+	trigger = cache->eviction_trigger;
```
sueloverso: This is really the one function that uses these values. My initial thought when I read the config changes would be that instead of computing a `double` as a percentage, why not just convert everything to an absolute value and then I think we can get rid of the multiplications and divisions below and just do comparisons?

src/conn/conn_cache.c
```diff
@@ -8,6 +8,38 @@
 
 #include "wt_internal.h"
 
+/*
+ * __cache_config_abs_to_pct --
+ *	Convert the absolute value in cache config to percentage.
+ */
+static inline int
+__cache_config_abs_to_pct(WT_SESSION_IMPL *session,
+    double *param, const char *param_name)
```
sueloverso: If you take my suggestion below this becomes `__cache_config_pct_to_abs`.

src/evict/evict_lru.c
nehakhatri5: In case of shared cache the cache size is decided dynamically per connection. In that situation we would want to provide a % input value per connection for these parameters. An absolute value would not be able to adjust like % value per connection. Hence we chose this path of converting absolute to percentage.

src/docs/upgrading.dox
nehakhatri5: I think it is important to mention if the change is not backwards compatible, but in this case compatibility is maintained so it is ok to not mention that.

src/evict/evict_lru.c
bvpvamsikrishna: my understanding is wiredtiger should not take the same % when the cache size is changed. for eg:-
1. wiredtiger open (cache size = 2 TB, eviction target = 1TB)
now eviction target is 50%.
2. reconfigure wiredtiger with cache size = 3 TB, now what will be the eviction target set, will it be 50% of new cache size ( which will be 1.5 TB) or still we consider it as 1 TB ?

my understanding is we have to consider the value provided by user which is 1 TB, and hence when re configured, it should be only 33%  should not be 50%.

nehakhatri5: @bvpvamsikrishna Yes, when a reconfigure is done with a different cache size a new percentage would be calculated based on the absolute value in the input.

But here I am specifically referring the case of shared cache. WiredTiger has a provision to specify a cache pool that can be shared between multiple connections. With shared cache, the cache size is dynamically decided per connection. For a cache size generated dynamically like this, these parameters' (absolute)values should be converted to percentage values, which can be applied to every connection.

agorrod: @sueloverso the reason to keep these as percentages is that otherwise it becomes very difficult to keep the values consistent when the size changes, and in a cache pool configuration it changes regularly. Which made me think about cache pools, and I think we should disallow absolute values here when the cache pool is in use, especially given the special case `don't divide by zero` in the new code for cache pool configurations. There is precedent for similar checks here:
https://github.com/wiredtiger/wiredtiger/blob/develop/src/conn/conn_cache_pool.c#L80
We should add something asserting that the 5 configuration options need to be less than or equal to 100

src/conn/conn_cache.c
```diff
@@ -8,6 +8,38 @@
...
+	/*
+	 * Anything above 100 is an absolute value; convert it to percentage.
+	 * An absolute value can't exceed the cache size.
+	 */
+	if (*param > 100.0) {
+		if (*param > conn->cache_size)
+			WT_RET_MSG(session, EINVAL,
+			    "%s should not exceed cache size", param_name);
+		/*
+		 * Avoid division by zero if the cache size has not yet been set
```
agorrod: Once you take my advice about disallowing absolute values and shared cache it should be possible to replace this with an assertion that the shared cache isn't configured (unless there is an ordering problem).

src/evict/evict_lru.c
sueloverso: Thanks for the details everyone.

sueloverso:
I definitely want to second Alex's comment about adding cases to the Python test suite. It is important to try mix-n-match for this new API for both proper usage and error behavior.

nehakhatri5:
Please hold on review, some more tests coming.

nehakhatri5:
@sueloverso @bvpvamsikrishna @agorrod Thanks for the inputs. Please have a look at the updated pull request.

dist/api_data.py
```diff
@@ -525,7 +532,9 @@ def __cmp__(self, other):
     Config('shared_cache', '', r'''
         shared cache configuration options. A database should configure
         either a cache_size or a shared_cache not both. Enabling a
-        shared cache uses a session from the configured session_max''',
+        shared cache uses a session from the configured session_max. shared
+        cache can not have absolute values configured for cache eviction
```
agorrod: Need to capitalize on the new sentence.

src/include/cache.h
```diff
@@ -105,12 +105,12 @@ struct __wt_cache {
 	WT_CONDVAR *evict_cond;		/* Eviction server condition */
 	WT_SPINLOCK evict_walk_lock;	/* Eviction walk location */
 
-	u_int eviction_dirty_target;    /* Percent to allow dirty */
-	u_int eviction_dirty_trigger;	/* Percent to trigger dirty eviction */
-	u_int eviction_trigger;		/* Percent to trigger eviction */
-	u_int eviction_target;		/* Percent to end eviction */
+	double eviction_dirty_target;    /* Percent to allow dirty */
```
agorrod: Adding a comment above this block explaining why we are storing percentages in a double type would be worthwhile.

src/include/cache.i
```diff
@@ -312,7 +312,7 @@ __wt_eviction_dirty_needed(WT_SESSION_IMPL *session, u_int *pct_fullp)
  */
 static inline bool
 __wt_eviction_needed(
-    WT_SESSION_IMPL *session, bool busy, bool readonly, u_int *pct_fullp)
+    WT_SESSION_IMPL *session, bool busy, bool readonly, double *pct_fullp)
 {
 	WT_CACHE *cache;
 	u_int pct_dirty, pct_full;
```
agorrod: Why didn't you push the `double` change down into `__wt_eviction_clean_needed` and `__wt_eviction_dirty_needed` functions?

test/suite/test_config04.py
```diff
@@ -154,6 +153,47 @@ def test_eviction_bad2(self):
...
+    def test_eviction_absolute(self):
+        self.common_test('eviction_target=50MB,eviction_trigger=60MB,'
+            'eviction_dirty_target=20MB,eviction_dirty_trigger=15MB,'
+            'eviction_checkpoint_target=13MB')
+
+    def test_eviction_absolute_bad(self):
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError, lambda:
+            self.wiredtiger_open('.','create,eviction_target=70MB,'
+                                 'eviction_trigger=60MB'),
+            '/eviction target must be lower than the eviction trigger/')
```
agorrod: It'd be nice if there were a few more failure cases in here where inputs exceed other inputs that they aren't allowed to.
It would also be interesting to set a cache size to less than 1% via the absolute size, and verify that the setting is as expected.

dist/api_data.py
sueloverso: Also it should be "A shared cache..."

src/conn/conn_cache.c
```diff
@@ -8,6 +8,44 @@
...
+	conn = S2C(session);
+
+	WT_ASSERT(session, param != NULL);
+
+	/*
+	 * Anything above 100 is an absolute value; convert it to percentage.
+	 */
+	if (*param > 100.0) {
+		/*
+		 * In shared cache configuration the cache size changes
```
sueloverso: "In a shared cache..."

src/conn/conn_cache.c
```diff
@@ -8,6 +8,44 @@
...
+		 * regularly; an absolute value for the configuration parameter
```
sueloverso: I don't think a semicolon is necessary here. The two statements are not very tightly tied together. I suggest:
In a shared cache configuration the cache size changes regulary. Therefore, we require a percentage setting and do not allow an absolute size setting.

test/suite/test_config04.py
sueloverso: I would also like to see something similar to `common_cache_size_test` to verify that the settings are correct.

agorrod:
Thanks @nehakhatri5 lgtm, please merge once testing finishes

PR #3848, opened 2017-12-20
WT-3812 debugging page output should handle complex key/value items.

keithbostic:
@michaelcahill, @bvpvamsikrishna suggested a couple of weeks ago we should crack complex key/value pairs when dumping pages for debugging, and it's a good idea. I ended up implementing it when I was working on WT-3768, it made it a lot easier to figure out what the lookaside table was doing.

First, I ended up using the streaming pack/unpack APIs, is there anything that would be simpler?

Second, it mostly works, but there's a problem with zero-length `u` formats. I'm  using the pack stream APIs you added, and there's some magic around zero-length `U/u` formats that I don't understand.

The lookaside `__wt_las_insert_block()` code has no problem inserting zero-length items, and obviously we can subsequently read them, but when I try and call `wiredtiger_unpack_item()` on the "QuBu" format it fails because the saved byte string isn't long enough -- there's no stored length, as near as I can figure out, and `wiredtiger_unpack_item()` fails here:
```
/* Lower-level packing routines treat a length of zero as unchecked. */
if (ps->p >= ps->end)
        return (ENOMEM);
```
Don't bother chasing this if you don't know the answer off-hand, but I've been staring at it for a little while now, and I'm hoping you just know what's going on.

### DIFF
--- a/src/support/scratch.c
+++ b/src/support/scratch.c
@@ -162,74 +162,70 @@ __wt_buf_set_printable(
  */
 const char *
 __wt_buf_set_printable_format(WT_SESSION_IMPL *session,
-    const void *p, size_t size, const char *format, WT_ITEM *buf)
+    const void *buffer, size_t size, const char *format, WT_ITEM *buf)
 {
 	WT_DECL_ITEM(tmp);
 	WT_DECL_PACK_VALUE(pv);
 	WT_DECL_RET;
-	WT_PACK_STREAM *ps;
-	WT_SESSION *wt_session;
+	WT_PACK pack;
+	const uint8_t *p, *end;
 	const char *retp, *sep;
 
-	ps = NULL;
-	wt_session = (WT_SESSION *)session;
+	p = (const uint8_t *)buffer;
+	end = p + size;
 
-	WT_ERR(__wt_buf_init(session, buf, 32));
+	WT_ERR(__wt_buf_init(session, buf, 0));
 
-	WT_ERR(wiredtiger_unpack_start(wt_session, format, p, size, &ps));
-
-	for (sep = ""; *format != '\0'; ++format)
-		switch (*format) {
-		case '.':
-		case '>':
-		case '<':
-		case '@':
+	WT_ERR(__pack_init(session, &pack, format));
+	for (sep = ""; (ret = __pack_next(&pack, &pv)) == 0;) {
+		WT_ERR(__unpack_read(session, &pv, &p, (size_t)(end - p)));
+		switch (pv.type) {
 		case 'x':
+			break;
+		case 's':
+		case 'S':
+			WT_ERR(__wt_buf_catfmt(
+			    session, buf, "%s%s", sep,  pv.u.s));
+			sep = ",";
+			break;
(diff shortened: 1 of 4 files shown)


src/support/scratch.c
```diff
@@ -155,6 +155,89 @@ __wt_buf_set_printable(
...
+	const char *retp, *sep;
+
+	ps = NULL;
+	wt_session = (WT_SESSION *)session;
+
+	WT_ERR(__wt_buf_init(session, buf, 32));
+
+	WT_ERR(wiredtiger_unpack_start(wt_session, format, p, size, &ps));
+
+	for (sep = ""; *format != '\0'; ++format)
```
michaelcahill: This loop should use `__pack_init`, `__pack_next` rather than attempting to parse the string.  The current version will fail with a format like "3qS", for example.

keithbostic: Pushed in 3a4d6ec.

agorrod:
@keithbostic did your latest push address your original question here? i.e: is this one ready for final review now?

keithbostic:
@agorrod 

> did your latest push address your original question here? i.e: is this one ready for final review now?

I'm no longer using the streaming unpack APis at all, at @michaelcahill's suggestion. I assume he suggested the change because I was using the external APIs incorrectly, and there's no problem there.

AFAIK, it's ready for final review.

michaelcahill:
FTR, the streaming unpack routines are fine to use.  They can also be combined with `__pack_init` and `__pack_next`.  The problem was the new code that attempted to parse a format string by walking a character at a time.

src/support/scratch.c
```diff
@@ -155,6 +155,85 @@ __wt_buf_set_printable(
...
+	WT_ERR(__pack_init(session, &pack, format));
+	for (sep = ""; (ret = __pack_next(&pack, &pv)) == 0;) {
+		WT_ERR(__unpack_read(session, &pv, &p, (size_t)(end - p)));
+		switch (pv.type) {
+		case 'x':
+			break;
+		case 's':
+		case 'S':
+			WT_ERR(__wt_buf_catfmt(
+			    session, buf, "%s%s", sep,  pv.u.s));
```
agorrod: Adding quotes around strings might be better so it's possible to know whether a `,` is a separator or part of a string. Though I guess that'll mean strings containing quote characters will be difficult to interpret. Another option might be to print out the string length as part of the output, but that will be difficult to separate from other fields. Maybe keeping it simple is best?

keithbostic: @agorrod 

> I went through the exact same thought process, and futzed with it a bit (and we already put `{}` around the key/value pairs, it's not just commas).

I'd be inclined to either leave it alone or insert backslashes in front of `{}` and commas, but don't feel strongly.

agorrod: Cool - this gets an lgtm from me then.

keithbostic: @agorrod 

Which do you prefer? Leave alone, or backslash escape?

agorrod: Leave it alone

PR #3850, opened 2017-12-21
WT-3818 __rec_txn_read() code order cleanup

keithbostic:
@michaelcahill, I made some changes in WT-3768 that made a not-great situation worse, so I've tried to clean it up.

There shouldn't be any real changes here, just code ordering for clarity.

Feel free to discard without further discussion if you don't want this change.


keithbostic:
@michaelcahill, follow-up question:

The code in `__rec_txn_read()` explicitly updates the maximum on-page timestamp before discarding the chosen update (discarded because the update is a `WT_UPDATE_BIRTHMARK` type, or we're doing update/restore eviction and we skipped a birthmark).

That change was in ff4c226 and I remember it fixed a failure in `test_timestamp07`, but logically I don't know why it's required. If we're not putting that update on the page, why would it be necessary to track that timestamp?

src/reconcile/rec_write.c
```diff
@@ -1463,40 +1479,20 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 #else
 	timestampp = NULL;
 #endif
-	all_visible = *updp == first_txn_upd && !uncommitted &&
+	all_visible = upd == first_txn_upd && !uncommitted &&
```
michaelcahill: This is a little subtle -- `*updp` may have diverged from `upd` at this point.  I'm open to adding a boolean local variable at [line 1427](https://github.com/wiredtiger/wiredtiger/pull/3850/files#diff-a6c8d99189e5865e35b893d6792f1165R1427) if you think that would be clearer.

src/reconcile/rec_write.c
```diff
@@ -1409,17 +1409,32 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
 	/* If all of the updates were aborted, quit. */
 	if (first_txn_upd == NULL) {
-		WT_ASSERT(session, *updp == NULL);
+		WT_ASSERT(session, upd == NULL);
 		return (0);
 	}
 
+	/* If no updates were skipped, record that we're making progress. */
+	if (upd == first_txn_upd)
+		r->update_used = true;
```
michaelcahill: I changed this back to checking the update before resolving birthmarks.  It may not be a big deal because this is only used as a heuristic, but I think of this as tracking whether reconciliation selected the most recent update in some list.

Maybe it would be better to track whether `all_visible` is ever true during reconciliation (so some update chain will go away during update/restore).  That's a real behavior change, so probably deserves a separate ticket and a round of performance tests.

PR #3851, opened 2017-12-24
WT-3779 Add support for string formats with WT_CURSOR::modify

keithbostic:


agorrod:
Thanks @keithbostic lgtm - I'm excited about this change, it'll make it much easier to add modify support to our test programs.

PR #3855, opened 2017-12-27
WT-3800 Removed incorrect NOTREACHED comment

ddanderson:


ddanderson:
@sueloverso, a fix to the error you spotted.

sueloverso:
@ddanderson before merging this branch into develop, I pushed a sub-branch with some debugging in it for the test to catch its failure.

ddanderson:
@sueloverso, please take another look.  I added a course way to track progress for these tests. A couple of observations, WT-2853 original failure mode was a hang. Perhaps the problem is back under the right circumstances. I don't know how WT-2834 would be involved though.  While investigating WT-2853 with different parameters, I saw some failures -- some assertions about what data would be turned out to be incorrect if the writer thread hit one of the records at just the right time. It should result in an abort. Can't see how that gets us to one or more hung processes though. Anyway, I removed the errant assertions.

sueloverso:
Thanks Don. LGTM.

PR #3856, opened 2017-12-27
WT-3800 Add debugging to retain database directories.

sueloverso:
I created this branch against the fix in the WT-3800 branch for debugging the hang and having artifacts after a failure.


sueloverso:
autobuild: retest this please

PR #3858, opened 2017-12-29
WT-3819 clang improvements

keithbostic:
@agorrod, these are the last set of changes to move the clang static analysis tools, scan-build and clang-tidy, into the source tree.

To run the tests, you clone a tree and do:
```
% sh dist/clang-scan/s_clang-scan
% sh dist/clang-tidy/s_clang-tidy
```

The output is minimal, none for scan-build and maybe 20 lines for clang-tidy (which means we could incorporate scan-build into the Jenkins tests, I suppose).

The scripts are unfortunately a little fiddly, and we don't run cleanly under clang-tidy, but this should allow people other than me to run the tests.

Also, I have not parallelized clang-tidy and it takes 10-15 minutes to run.

cc: @markbenvenuto, @acmorrow


agorrod:
Thanks @keithbostic this is looking good to me. Why did you decide to put the scripts in sub-directories of dist?

I'm also not a huge fan of renaming wt_session, but I guess API checking makes it worthwhile. Are other places in the API not picked up because we use function pointers?

keithbostic:
@agorrod 

> Why did you decide to put the scripts in sub-directories of dist?

I thought I'd end up needing to diff the output of the scripts against "known failures", so I'd need more infrastructure files, but then I worked around pretty much everything and that wasn't necessary. I've moved the scripts to the top-level.

> I'm also not a huge fan of renaming wt_session, but I guess API checking makes it worthwhile. Are other places in the API not picked up because we use function pointers?

Yeah, me either -- this last set of changes are all the things I really didn't want to change, but finally decided to make clang happy.

I didn't figure out the root cause, there are other external functions that clang doesn't notice and that aren't method pointers.

We could switch our source tree and documentation to fully use `wt_session` instead of `session` for all `WT_SESSION` parameters. That's a bigger change by a lot, but not in areas likely to affect our ability to easily do backports. And, we might have to do that anyway, as future clang releases figure out our documentation?

If you support that approach, you've got my vote.

sueloverso:
bq. Yeah, me either -- this last set of changes are all the things I really didn't want to change, but finally decided to make clang happy.

Add me to the list of not-a-fan of the renaming. It certainly diminishes the consistency within our own code in a non-obvious way. To me that is more important from our local perspective. Is there a way to avoid that particular piece of clang checking? If you really think this is only a stopgap until clang figures out documentation, and we'll have to bite the bullet eventually, I guess we should do that now and keep our consistency.

keithbostic:
Chatting with @sueloverso, some additional comments.

We could switch just the documentation and example code from `WT_SESSION *session;` to use `wt_session` (I think the example code needs to match the documentation), and ignore everything else.

In total, there are maybe 170 files with 750 uses that would require review: 100 `example/`, 140 `lang/` or `ext/`, 120 documentation, 70 `utilities/`, and the rest is test code.

While it's a big change, it's easy to do and is unlikely to introduce errors, the compiler will catch any editing mistakes.

And, of course, alternatively we could decide to turn off this clang warning, and back out the changes to the library code.

I OK with that, although I think the complaint isn't unreasonable: if there's a different variable name in the documentation vs. the code, one of them *must* better than the other, and should have been used in both places, and it's a minor aid to understanding the code.

I don't feel terribly strongly about this one, myself.

agorrod:
Since none of the options are particularly appealing, I'm inclined to disable the parameter name warning and back out that subset of changes.

keithbostic:
@agorrod 
> Since none of the options are particularly appealing, I'm inclined to disable the parameter name warning and back out that subset of changes.

Pushed and verified, Alex, let me know if there are any other changes you want.

agorrod:

PR #3859, opened 2017-12-31
WT-3820 Add a WT_SESSION.breakpoint method for Python debugging.

keithbostic:
@agorrod, I was hacking this into the tree again today, and thought I'd clean it up enough that maybe we could leave it in permanently?

It's the obvious hack allowing you to set a gdb breakpoint in `__wt_session_breakpoint` and then do something like:
```
if key == 5:
     self.session.breakpoint()
```
in Python.

I do that kind of thing a fair amount because it let's me pick a point in a Python test and then squirrel around in the in-memory tree. (I'm guessing @michaelcahill does that a fair bit as well, maybe he's got a better way of doing it?)


ddanderson:
@keithbostic, LGTM.  I'll merge.

PR #3860, opened 2018-01-01
WT-3775 Improve set timestamp error message

bvpvamsikrishna:
1. txn_global->rwlock was not acquired as access of txn_global->oldest_timestamp & txn_global->stable_timestamp is for printing in error msg purpose only.

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -556,8 +556,8 @@ __wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 	bool has_oldest_ts, has_stable_ts;
 
 	/*
- 	 * Added this redundant initialization to circumvent build failure.
- 	 */
+	 * Added this redundant initialization to circumvent build failure.
+	 */
 	__wt_timestamp_set_zero(&oldest_ts);
 	__wt_timestamp_set_zero(&stable_ts);
 	/*
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -563,7 +563,7 @@ __wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 	    if ((has_oldest_ts = txn_global->has_oldest_timestamp))
 		__wt_timestamp_set(&oldest_ts, &txn_global->oldest_timestamp);
 	    if ((has_stable_ts = txn_global->has_stable_timestamp))
-		__wt_timestamp_set(&stable_ts, &txn_global->stable_timestamp););
+		__wt_timestamp_set(&stable_ts, &txn_global->stable_timestamp));
 
 	if (cmp_oldest && has_oldest_ts &&
 	    __wt_timestamp_cmp(ts, &oldest_ts) < 0) {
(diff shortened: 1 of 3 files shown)


bvpvamsikrishna:
@sulabhM @nehakhatri5 Request for a review of this pull request.

src/txn/txn_timestamp.c
```diff
@@ -432,17 +434,25 @@ __wt_txn_global_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
 	    __wt_timestamp_cmp(&oldest_ts, &commit_ts) > 0) {
 		__wt_readunlock(session, &txn_global->rwlock);
+		WT_RET(__wt_timestamp_to_hex_string(
+		    session, hex_timestamp1, &oldest_ts));
+		WT_RET(__wt_timestamp_to_hex_string(
+		    session, hex_timestamp2, &commit_ts));
 		WT_RET_MSG(session, EINVAL,
-		    "set_timestamp: oldest timestamp must not be later than "
-		    "commit timestamp");
+		    "set_timestamp: oldest timestamp %s must not be later than"
```
sulabhM: I think the convention is to put the space at the end of the previous line and not start the next line with a space. Same at the other places.

src/txn/txn_timestamp.c
```diff
@@ -553,14 +567,28 @@ __wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 	    txn_global->has_stable_timestamp &&
 	    __wt_timestamp_cmp(ts, &txn_global->stable_timestamp) < 0));
 
-	if (older_than_oldest_ts)
+	if (older_than_oldest_ts) {
+	/*
```
sulabhM: This is not aligned right, same for other comments below.

src/txn/txn_timestamp.c
```diff
@@ -553,14 +567,28 @@ __wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 	    txn_global->has_stable_timestamp &&
 	    __wt_timestamp_cmp(ts, &txn_global->stable_timestamp) < 0));
 
-	if (older_than_oldest_ts)
+	if (older_than_oldest_ts) {
+		/*
+		 * As global oldest timestamp access is for error message only,
```
agorrod: I think you should acquire a lock in this case - it's not in a performance path so doesn't warrant an exception, and without a lock we can see invalid timestamps (i.e: a timestamp might be half way through being updated). It would actually be nice if this value was identical to the one used in the comparison (i.e: we got a local copy) otherwise the error message could be confusing, since the state may have changed.

src/txn/txn_timestamp.c
```diff
@@ -553,14 +567,24 @@ __wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 	    txn_global->has_stable_timestamp &&
 	    __wt_timestamp_cmp(ts, &txn_global->stable_timestamp) < 0));
 
-	if (older_than_oldest_ts)
+	if (older_than_oldest_ts) {
+		WT_WITH_TIMESTAMP_READLOCK(session, &txn_global->rwlock,
+		    ret = __wt_timestamp_to_hex_string(
+		    session, hex_timestamp, &txn_global->oldest_timestamp));
```
nehakhatri5: Here, it is possible that txn_global->oldest_timestamp gets updated by another thread. So it would be different from from what it might have been originally at line#565. If this happens the error message will show an incorrect value of the oldest timestamp. 
Can we save the oldest_timestamp in line#565 into a local variable while we have the lock and use that saved value to display in the error message?

bvpvamsikrishna:
autobuild: retest this please

bvpvamsikrishna:
@nehakhatri5 @agorrod , request to check the updates.

agorrod:
Thanks @bvpvamsikrishna lgtm - I'll go ahead and merge.

PR #3861, opened 2018-01-02
WT-3822 Update WiredTiger copyrights to 2018

keithbostic:
It's that time of the year.


agorrod:
Thanks @keithbostic lgtm.

PR #3862, opened 2018-01-04
WT-3824 Calibrate the ticks much earlier. Check for too small value.

sueloverso:
@agorrod and @sulabhM Please review these changes. The stack and assertion showed that we were using the time function very early in startup. I moved the calibration to the first thing after creating and initializing the connection structure. I also reorganized the calibration to default to using epoch time and only overriding that if we have good values.


sueloverso:
@keithbostic suggested moving this into the `WT_PROCESS` and computing it once for all databases that may be managed rather than per-connection. That makes sense and I'll code that up instead.

PR #3863, opened 2018-01-04
WT-1228 Implementation of cursor caching.

ddanderson:
This branch/pull request is not yet ready for merge.  It's mainly for comments and pre-review.

There are some basic tests for caching, including some with drops.  More testing and performance work remain to be done.


ddanderson:
With the latest commit, if I enable `cache_cursors=true` in the session config in `wttest.py`, and run the test suite, I get about dozen failures that fall into three classes: random cursors, json dump cursors and regular dump cursors. I suspect these will be pretty easy to fix: we're not currently paying proper attention to the random and dump bits in the cursor flags. The easiest solution will be to not have these specialized cursors participate in caching.

ddanderson:
@michaelcahill and @agorrod, if/when you have time to review the current state of this branch it would be appreciated.

dist/api_data.py
```diff
@@ -708,6 +708,15 @@ def __cmp__(self, other):
...
 
 session_config = [
+    Config('cache_cursors', '', r'''
+        enable caching of cursors for reuse. Any calls to WT_CURSOR::close
+        for a cursor created in this session will mark the cursor
+        as cached and keep it available to be reused for later calls
+        to WT_SESSION::open_cursor. Cached cursors may be eventually
+        closed. This flag requires
+        ::wiredtiger_open or WT_CONNECTION::reconfigure to be called
+        with \c cache_cursors enabled.''',
```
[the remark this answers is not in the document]
ddanderson: In the previous incarnation, I had one: the connection-wide flag was required in order to enable the session flag.  That made sense because there was some overhead to tracking/managing the caching. Now there isn't any extra overhead.  Would you want to see a connection-wide flag used in that same way, or rather a connection-wide flag that means "turn on caching in all sessions unless explicitly disabled"?

dist/stat_data.py
```diff
@@ -282,11 +282,13 @@ def __init__(self, name, desc, flags=''):
     ##########################################
     # Cursor operations
     ##########################################
+    CursorStat('cursor_cache', 'cursors cached on close'),
```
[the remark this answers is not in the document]
ddanderson: Sure, good idea.

dist/api_data.py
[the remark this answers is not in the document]
michaelcahill: A couple of points: the documentation string above still refers to a flag in `wiredtiger_open`.  I'd like less surface area rather than more, so suggest removing that from the docstring.

To the other point, I guess I'd ask why not have this behavior always enabled?  Is there any downside other than extra memory associated with session handles?

src/session/session_api.c
```diff
@@ -385,6 +400,9 @@ __session_open_cursor_int(WT_SESSION_IMPL *session, const char *uri,
 	if (*cursorp == NULL)
 		return (__wt_bad_object_type(session, uri));
 
+	if (owner != NULL)
+		F_CLR(*cursorp, WT_CURSTD_CACHEABLE);
+
```
michaelcahill: It's probably worth a comment inline to explain what's going on here: i.e., dependent cursors remain referenced by their owning cursor and are cached as part of caching the owning cursor.

src/cursor/cur_std.c
```diff
@@ -556,6 +569,180 @@ err:		cursor->saved_err = ret;
...
+	WT_CONFIG_ITEM cval;
+	WT_CURSOR *cursor;
+	WT_DECL_RET;
+	const char *tmp_cfg;
+
+	if (cfg == NULL) {
+		tmp_cfg = NULL;
+		cfg = &tmp_cfg;
+	}
+	WT_RET(__wt_config_gets(session, cfg, "readonly", &cval));
```
michaelcahill: I suggest using `__wt_config_gets_def` consistently (and checking the defaults), to avoid repeated parsing of config string defaults on every cursor (re)open.

src/cursor/cur_std.c
```diff
@@ -556,6 +569,180 @@ err:		cursor->saved_err = ret;
...
+
+#define	CHECKPOINT_MATCH(s, cval)					\
+	((s == NULL && cval.len == 0) ||				\
+	    (s != NULL && WT_STRING_MATCH(s, cval.str, cval.len)))
+
+	/*
+	 * Walk through all cursors, if there is a cached
+	 * cursor that matches uri and configuration, use it.
+	 */
+	TAILQ_FOREACH(cursor, &session->cursors, q) {
```
michaelcahill: Would it be reasonable to switch to using a hash table here (not necessarily before merging, just eventually)?

src/include/api.h
```diff
@@ -171,10 +171,32 @@
...
+
+#define	CURSOR_OPEN_CACHE(s, uri, cfg, cursorp) do {		\
+	int __cret;							\
+	if (F_ISSET(session, WT_SESSION_CACHE_CURSORS)) {		\
+		if ((__cret = __wt_cursor_open_cache(			\
+		    s, uri, cfg, cursorp)) == 0)			\
+			return (0);					\
+		WT_RET_NOTFOUND_OK(__cret);				\
+	}								\
+} while (0)
```
michaelcahill: Can this logic just be folded into the functions, rather than using a macro?

Also, naming: I read `__wt_cursor_open_cache` as "open the cursor cache" -- maybe `__wt_cursor_cache_get` and `__wt_cursor_cache_release`?

src/cursor/cur_std.c
ddanderson: Yes. I found this to be a hot spot with profiling.

src/cursor/cur_std.c
ddanderson: I think so.  As the number of cached cursors builds up, the sequential search will start to be a problem, just as it is in MongoDB if there's a more random access pattern between collections.

src/include/api.h
ddanderson: Thanks, I will do that.

dist/api_data.py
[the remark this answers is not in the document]
ddanderson: @michaelcahill I'm open to enabling caching globally. I suppose I've been waiting to see if performance was comparable on a variety of use cases before taking that step.

lukech:
autobuild: retest this please

lukech:
autobuild: retest this please

lukech:
autobuild: retest this please

ddanderson:
@agorrod, can you do a review when you get a chance?

With these latest commits, WT-3882 and WT-3883 should be code complete, and this is ready for a preliminary review.  There are a handful of items left to do before I consider this ready for a final review:

- WT-3884: support reconfigure (not much to be done)
- I'm currently saving the hash bucket number in the cursor, I think I should really save the hash value, to make comparisons faster when traversing large hash lists.
- Need stats for sweeping
- More test cases + some ad hoc testing to force error paths.

I'll be doing some more performance testing and tuning, so there may be some other fixes due to that.

agorrod:
Thanks @ddanderson this is looking pretty good to me after a first pass.

@sueloverso if you get a chance it would be great if you can do a review of this change.

I've switched the perf tests over to this branch, though I don't expect them to show any difference, since they aren't generally cursor open/close heavy.

I'd like to hold off merging until we have WT-3805 merged, which is in code review at the moment as well.

src/cursor/cur_std.c
```diff
@@ -556,6 +581,215 @@ err:		cursor->saved_err = ret;
...
+	WT_STAT_DATA_DECR(session, session_cursor_cached);
+
+	TAILQ_REMOVE(&session->cursor_cache[cursor->cache_bucket], cursor, q);
+	TAILQ_INSERT_HEAD(&session->cursors, cursor, q);
+	F_CLR(cursor, WT_CURSTD_CACHED);
+}
+
+/*
+ * __wt_cursor_cache_release --
+ *	Put the cursor into a cached state, called during close operations.
```
sueloverso: Minor nit, add "called during cursor close operations". Not closing the session or connection.

src/cursor/cur_std.c
```diff
@@ -556,6 +581,215 @@ err:		cursor->saved_err = ret;
...
+		    CHECKPOINT_MATCH(cursor->checkpoint)) {
+			if ((ret = cursor->reopen(cursor, false)) != 0) {
+				F_CLR(cursor, WT_CURSTD_CACHEABLE);
+				session->dhandle = NULL;
+				(void)cursor->close(cursor);
+				return (ret);
+			}
+
+			F_CLR(cursor, WT_CURSTD_APPEND | WT_CURSTD_OVERWRITE |
+			    WT_CURSTD_RAW);
```
sueloverso: This line deserves a comment.

src/cursor/cur_std.c
```diff
@@ -556,6 +581,215 @@ err:		cursor->saved_err = ret;
...
+	WT_CURSOR *cursor;
+	WT_DECL_RET;
+	uint64_t bucket;
+	bool have_config;
+
+	if (owner != NULL && F_ISSET(owner, WT_CURSTD_CACHEABLE))
+		return (WT_NOTFOUND);
+	have_config = (cfg != NULL && cfg[0] != NULL && cfg[1] != NULL);
+	if (have_config) {
+		WT_RET(__wt_config_gets_def(
```
sueloverso: All of these that do the same thing should be alphabetized on the config string name I think. Also a small comment about not ever using the cache for specially configured cursors would be helpful.

src/session/session_api.c
```diff
@@ -49,6 +49,61 @@ __wt_session_reset_cursors(WT_SESSION_IMPL *session, bool free_buffers)
...
+int
+__wt_session_cursor_cache_sweep(WT_SESSION_IMPL *session)
+{
+	WT_CURSOR *cursor, *cursor_tmp;
+	WT_CURSOR_LIST *cached_list;
+	WT_DECL_RET;
+	uint32_t position;
+	int i, t_ret, nbuckets, nchecked, ncleaned;
+	bool caching;
+	bool productive;
```
sueloverso: Join the two bool variables onto a single line.

src/session/session_api.c
```diff
@@ -49,6 +49,61 @@ __wt_session_reset_cursors(WT_SESSION_IMPL *session, bool free_buffers)
...
+
+	position = session->cursor_sweep_position;
+	productive = true;
+	nbuckets = nchecked = ncleaned = 0;
+
+	caching = F_ISSET(session, WT_SESSION_CACHE_CURSORS);
```
sueloverso: This needs a comment because I expected that if `caching` is false we'd return immediately.  A comment about why we cannot would be helpful here. My guess is that the session could have been reconfigured.

src/session/session_api.c
```diff
@@ -167,6 +222,41 @@ __session_clear(WT_SESSION_IMPL *session)
...
+__session_close_cursors(WT_SESSION_IMPL *session, WT_CURSOR_LIST *cursors)
+{
+	WT_CURSOR *cursor, *cursor_tmp;
+	WT_DECL_RET;
+
+	/* Close all open cursors. */
+	WT_TAILQ_SAFE_REMOVE_BEGIN(cursor, cursors, q, cursor_tmp) {
+		if (F_ISSET(cursor, WT_CURSTD_CACHED))
+			/*
+			 * Put the cached cursor is an open state
```
sueloverso: nit: 'is' should be 'in'.

src/session/session_api.c
```diff
@@ -167,6 +222,41 @@ __session_clear(WT_SESSION_IMPL *session)
...
+			 * that allows it to be closed.
+			 */
+			WT_TRET_NOTFOUND_OK(cursor->reopen(cursor, false));
+		else
+			/*
+			 * Notify the user that we are closing the cursor handle
+			 * via the registered close callback.
+			 */
+			if (session->event_handler->handle_close != NULL &&
```
sueloverso: The `if` should be joined with the `else` above and then you can move the comment inside the conditional.

src/session/session_api.c
```diff
@@ -1947,16 +2070,19 @@ __open_session(WT_CONNECTION_IMPL *conn,
...
 		WT_ERR(__wt_calloc(session, WT_HASH_ARRAY_SIZE,
 		    sizeof(struct __dhandles_hash), &session_ret->dhhash));
 	for (i = 0; i < WT_HASH_ARRAY_SIZE; i++)
 		TAILQ_INIT(&session_ret->dhhash[i]);
 
+	/* Initialize the cursor cache hash buckets and sweep trigger. */
+	for (i = 0; i < WT_HASH_ARRAY_SIZE; i++)
+		TAILQ_INIT(&session_ret->cursor_cache[i]);
+	session_ret->cursor_sweep_countdown = WT_SESSION_CURSOR_SWEEP_COUNTDOWN;
+
```
sueloverso: Unless I'm not seeing it, I think we're leaking the cursor cache array. The `dhhash` array is freed in `conn_open.c:175` and maybe this should be too?

agorrod:
> make it the default. Should I do that?

Yes please.

src/session/session_api.c
ddanderson: You're right, we should return immediately. I've simplified the code.

dist/api_data.py
```diff
@@ -719,6 +719,13 @@ def __cmp__(self, other):
 ]
 
 session_config = [
+    Config('cache_cursors', '', r'''
```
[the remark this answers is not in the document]
ddanderson: We're going to make cursor caching the default, so it will be true.

dist/stat_data.py
```diff
@@ -670,6 +673,7 @@ def __init__(self, name, desc, flags=''):
     CursorStat('cursor_prev', 'prev calls'),
     CursorStat('cursor_remove', 'remove calls'),
     CursorStat('cursor_remove_bytes', 'cursor-remove key bytes removed', 'size'),
+    CursorStat('cursor_reopen', 'reuses from cache'),
```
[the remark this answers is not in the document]
ddanderson: Fixed, thanks.

src/cursor/cur_std.c
```diff
@@ -134,6 +147,18 @@ __wt_cursor_reconfigure_notsup(WT_CURSOR *cursor, const char *config)
 	return (__wt_cursor_notsup(cursor));
 }
 
+/*
+ * __wt_cursor_reopen_notsup --
+ *	Unsupported cursor reopen.
+ */
```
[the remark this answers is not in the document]
ddanderson: @sueloverso, I agree with your sentiment, but there are already a half dozen or so `wt_*_notsup` functions. If you feel strongly enough, I'll open a JIRA to address it more completely, okay?

src/cursor/cur_std.c
```diff
@@ -134,6 +147,18 @@ __wt_cursor_reconfigure_notsup(WT_CURSOR *cursor, const char *config)
...
+int
+__wt_cursor_reopen_notsup(WT_CURSOR *cursor, bool check_only)
```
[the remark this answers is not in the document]
ddanderson: See my reply above.

src/cursor/cur_std.c
ddanderson: Thanks, fixed.

src/cursor/cur_std.c
ddanderson: Yes, I've clarified with some comments.

src/cursor/cur_std.c
ddanderson: Thanks, done.

src/session/session_api.c
ddanderson: Thanks, I've simplified this code, so `caching` is no longer needed.

src/session/session_api.c
ddanderson: Thanks, done.

dist/api_data.py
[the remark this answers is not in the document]
sueloverso: What I meant is instead of `''` have it explicitly say `'true'` to make it clear to the reader what the default is.

src/cursor/cur_std.c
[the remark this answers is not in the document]
sueloverso: Ah, so there are. Thanks for that information, then it is fine the way it is.

src/session/session_api.c
ddanderson: @sueloverso, `cursor_cache` is statically sized in the `__wt_session_impl` struct (session session.h: line 82).

test/suite/test_cursor13.py
```diff
@@ -0,0 +1,342 @@
...
+
+class test_cursor13_drops(test_cursor13_base):
+    def open_and_drop(self, uri, cursor_session, drop_session, nopens, ntrials):
+        for i in range(0, ntrials):
+            cursor_session.create(uri, 'key_format=S,value_format=S')
+            for i in range(0, nopens):
+                c = cursor_session.open_cursor(uri)
+                c.close()
+            drop_session.drop(uri)
+            confirm_does_not_exist(self, uri)
```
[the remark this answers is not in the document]
ddanderson: @sueloverso, dropping a URI doesn't change anything in the cache, so there's nothing really to check. The proof is that the drop succeeds at all.  I've added a comment to clarify.

ddanderson:
@agorrod and @sueloverso, I've resolved the comments from Sue's review.  I've made cursor caching the default.  I've done the first two items in the "remains to be done" list above, including WT-3884 (reconfigure support).  Now, remaining to be done:

- Need stats for sweeping
- More test cases + some ad hoc testing to force error paths.

dist/api_data.py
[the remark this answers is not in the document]
ddanderson: I have committed a change with an explicit `'true'`.

agorrod:
Thanks @ddanderson this is looking good to me apart from the test failure - which I believe was a crash while running the new test case.

I was thinking of merging this today, but with the test failure I'll leave it up to you and @sueloverso to decide if it should be merged tomorrow.

src/session/session_api.c
ddanderson: I agree, I've modified dhhash to be statically allocated in the structure.  That matches similar uses in `connection.h`.

ddanderson:
@sueloverso, this is ready for merge if you agree.  I've backed out the new test that tests sweep statistics. The test works, but apparently not in a parallel testing environment. My simple fixes didn't work, I'll need to rework that.

src/cursor/cur_std.c
```diff
@@ -556,6 +581,237 @@ err:		cursor->saved_err = ret;
...
+		if (cval.val != 0)
+			return (WT_NOTFOUND);
+
+		WT_RET(__wt_config_gets_def(
+		    session, cfg, "readonly", 0, &cval));
+		if (cval.val)
+			return (WT_NOTFOUND);
+
+		/*
+		 * Look for checkpoint last, the value will stay in 'cfg'.
```
sueloverso: Minor nit: do you mean 'cval' here instead of cfg?

ddanderson: Thanks!  Fixed.  I'll merge after it passes the PR tests.

PR #3864, opened 2018-01-04
WT-3824 Move tick calibration early - into global process structure.

sueloverso:
@keithbostic and @agorrod here's a change that moves the tsc calibration into the `__wt_process` global instead of the connection.

### DIFF
--- a/src/support/global.c
+++ b/src/support/global.c
@@ -41,6 +41,67 @@ __wt_endian_check(void)
 }
 
 /*
+ * __global_calibrate_ticks --
+ *	Calibrate a ratio from rdtsc ticks to nanoseconds.
+ */
+static void
+__global_calibrate_ticks()
+{
+#if defined (__i386) || defined (__amd64)
+	struct timespec start, stop;
+	double ratio;
+	uint64_t diff_nsec, diff_tsc, min_nsec, min_tsc;
+	uint64_t tries, tsc_start, tsc_stop;
+	volatile uint64_t i;
+
+	min_nsec = min_tsc = 0;
+	/*
+	 * Run this calibration loop a few times to make sure we get a
+	 * reading that does not have a potential scheduling shift in it.
+	 * The inner loop is CPU intensive but a scheduling change in the
+	 * middle could throw off calculations. Take the minimum amount
+	 * of time and compute the ratio.
+	 */
+	for (tries = 0; tries < 3; ++tries) {
+		__wt_epoch(NULL, &start);
+		tsc_start = __wt_rdtsc(NULL);
+		/*
+		 * This needs to be CPU intensive and large enough.
+		 */
+		for (i = 0; i < WT_MILLION * 100; i++)
+			;
+		tsc_stop = __wt_rdtsc(NULL);
+		__wt_epoch(NULL, &stop);
+		diff_nsec = WT_TIMEDIFF_NS(stop, start);
+		diff_tsc = tsc_stop - tsc_start;
+		if (min_nsec == 0)
+			min_nsec = diff_nsec;
+		else
+			min_nsec = WT_MIN(min_nsec, diff_nsec);
+		if (min_tsc == 0)
+			min_tsc = diff_tsc;
+		else
+			min_tsc = WT_MIN(min_tsc, diff_tsc);
+	}
+
+	/*
+	 * If we couldn't get a good reading then fallback to getting
+	 * time with wt_epoch. One possible reason is that the system's
+	 * clock granularity is not fine-grained enough.
+	 */
+	__wt_process.tsc_nsec_ratio = WT_TSC_DEFAULT_RATIO;
+	if (min_nsec != 0) {
+		ratio =
+		    (double)min_tsc / (double)min_nsec;
+		if (ratio > DBL_EPSILON)
+			__wt_process.tsc_nsec_ratio = ratio;
+	}
+#else
+	__wt_process.tsc_nsec_ratio = WT_TSC_DEFAULT_RATIO;
+#endif
+}
+
+/*
  * __wt_global_once --
  *	Global initialization, run once.
  */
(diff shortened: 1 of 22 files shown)


src/support/global.c
```diff
@@ -40,6 +40,72 @@ __wt_endian_check(void)
...
+
+	min_nsec = min_tsc = 0;
+	/*
+	 * Run this calibration loop a few times to make sure we get a
+	 * reading that does not have a potential scheduling shift in it.
+	 * The inner loop is CPU intensive but a scheduling change in the
+	 * middle could throw off calculations. Take the minimum amount
+	 * of time and compute the ratio.
+	 */
+	__wt_process.use_epochtime = false;
```
keithbostic: I don't think this value is ever read, my bet is clang will complain.

src/support/global.c
```diff
@@ -40,6 +40,72 @@ __wt_endian_check(void)
...
+__global_calibrate_ticks(void)
+{
+#if defined (__i386) || defined (__amd64)
+	struct timespec start, stop;
+	double ratio;
+	uint64_t diff_nsec, diff_tsc, min_nsec, min_tsc;
+	uint64_t tries, tsc_start, tsc_stop;
+	volatile uint64_t i;
+
+	min_nsec = min_tsc = 0;
```
keithbostic: This initialization seems like it would go better near the loop in which it's used, setting it to `UINT64_MAX` means you don't have to explicitly test for 0, `WT_MIN` always works (shrug).

src/support/global.c
```diff
@@ -40,6 +40,72 @@ __wt_endian_check(void)
...
+		ratio =
+		    (double)min_tsc / (double)min_nsec;
+		if (ratio > DBL_EPSILON) {
+			__wt_process.tsc_nsec_ratio = ratio;
+			__wt_process.use_epochtime = false;
+		}
+	}
+#else
+	__wt_process.use_epochtime = true;
+	__wt_process.tsc_nsec_ratio = WT_TSC_DEFAULT_RATIO;
```
keithbostic: I'd flip the order, that is, set `ratio` then `epochtime` in all 3 cases -- not really necessary, but I did spend a second wondering why it was different in the `#else` case.

src/support/time.c
```diff
@@ -81,6 +81,7 @@ __wt_tsc_to_nsec(WT_SESSION_IMPL *session, uint64_t end, uint64_t begin)
 {
 	double tsc_diff;
 
+	WT_UNUSED(session);
```
keithbostic: Is there any reason to pass `session` as an argument? Could probably not pass it through the `WT_TSCDIFF_NS` macro, if there's a reason that macro should take it as an argument.

src/support/global.c
```diff
@@ -40,6 +40,72 @@ __wt_endian_check(void)
...
+	/*
+	 * If we couldn't get a good reading then fallback to getting
+	 * time with wt_epoch. One possible reason is that the system's
+	 * clock granularity is not fine-grained enough.
+	 */
+	__wt_process.tsc_nsec_ratio = WT_TSC_DEFAULT_RATIO;
+	__wt_process.use_epochtime = true;
+	if (min_nsec != 0) {
+		ratio =
+		    (double)min_tsc / (double)min_nsec;
```
keithbostic: Whitespace, this could be one line.

src/support/global.c
sueloverso: It is used in `wt_rdtsc` to determine if we should always use `wt_epoch` so it needs to be false at this point. Probably a comment would be good. This is where a flag would be better because a `USE_EPOCH_TIME` flag would not yet be set and the code checks if it is.

michaelcahill:
In the interests of resolving test failures, I've pushed a change (c9c040f) implementing the review comments.  It's passing tests, so I'll merge.

src/support/global.c
```diff
@@ -47,62 +47,58 @@ __wt_endian_check(void)
 static void
 __global_calibrate_ticks(void)
 {
+	/*
+	 * Default to using __wt_epoch until we have a good value for the ratio.
+	 */
+	__wt_process.tsc_nsec_ratio = WT_TSC_DEFAULT_RATIO;
+	__wt_process.use_epochtime = true;
```
sueloverso: @michaelcahill Did you read my reply on this particular line? This line turns off all rdtsc forever. We will never compute or otherwise use the tsc once this is true.

src/support/global.c
```diff
@@ -47,62 +47,58 @@ __wt_endian_check(void)
...
 
-	min_nsec = min_tsc = 0;
 	/*
 	 * Run this calibration loop a few times to make sure we get a
 	 * reading that does not have a potential scheduling shift in it.
 	 * The inner loop is CPU intensive but a scheduling change in the
 	 * middle could throw off calculations. Take the minimum amount
 	 * of time and compute the ratio.
 	 */
-	__wt_process.use_epochtime = false;
```
sueloverso: Removing this line and setting it to true above makes all of this a moot point and we will always use wt_epoch forever.

src/support/global.c
michaelcahill: @sueloverso, sorry if I misunderstood -- I changed it to start `true` but be set to `false` later in the function if we get a good sample.  Is there some reason why the function will exit in between?

michaelcahill: Oh, sorry, I see it now :-(

It's super confusing to have the `__wt_rdtsc` function not use `rdtsc`.  It would be much clearer to have the generic function called `__wt_gettime` or similar.

sueloverso: The `wt_rdtsc` looks at the boolean to decide if it should use `wt_epoch` or use `rdtsc`. The calibrate function is calling `wt_rdtsc` so if the boolean is set, `wt_rdtsc` assumes we already decided to avoid rdtsc and calls . If we couldn't get good readings and compute a ratio we were falling back to setting the boolean and using `wt_epoch`.

sueloverso: I'll fix the naming and possibly split the functions in WT-3825. And yes, you're correct that it will compute an incorrect ratio really close to 1.0 (my laptop shows 0.999995) but later uses will use rdtsc and compute incorrect times.

PR #3865, opened 2018-01-05
WT-3825 Fix calculation of CPU ticks per unit time.

michaelcahill:


michaelcahill:
@sueloverso, since I broke it, here is an attempt at fixing it...

Feel free to throw this out if you disagree with the naming, or edit as you see fit.

PR #3867, opened 2018-01-05
WT-3828 Add a function reference to resolve a link error on some systems

ddanderson:
This resolves the problem on my laptop OS/X.


ddanderson:
@keithbostic, could you review this please?  It seems that (at least w/ my compiler chain), a reference to a function (text symbol) is needed for the linker to pull in a .o file, in this case global.o.  Without this, I get undefined references to the __wt_process data symbol, also defined in global.o .

PR #3868, opened 2018-01-08
WT-3832 Add string to dictionary for lint

agorrod:


agorrod:
@bvpvamsikrishna could you please review once it passes testing?

PR #3869, opened 2018-01-08
WT-3826 Consider anything other than fscanf success to be EOF.

sueloverso:
@keithbostic Please review this defensive change for the file system issue. Interestingly enough, `timestamp-abort` already had a similar change in it for when `ret == 0`.

### DIFF
--- a/test/csuite/random_abort/main.c
+++ b/test/csuite/random_abort/main.c
@@ -391,9 +391,13 @@ main(int argc, char *argv[])
 		 */
 		for (last_key = UINT64_MAX;; ++count, last_key = key) {
 			ret = fscanf(fp, "%" SCNu64 "\n", &key);
-			if (ret != EOF && ret != 1)
-				testutil_die(errno, "fscanf");
-			if (ret == EOF)
+			/*
+			 * Consider anything other than clear success in
+			 * getting the key to be EOF. We've seen file system
+			 * issues where the file ends with zeroes on a 4K
+			 * boundary and does not return EOF but a ret of zero.
+			 */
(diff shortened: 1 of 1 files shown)


test/csuite/random_abort/main.c
```diff
@@ -391,9 +391,13 @@ main(int argc, char *argv[])
...
-			if (ret != EOF && ret != 1)
-				testutil_die(errno, "fscanf");
-			if (ret == EOF)
+			/*
+			 * Consider anything other than clear success in
+			 * getting the key to be EOF. We've seen file system
+			 * issues where the file ends with zeroes on a 4K
+			 * boundary and does not return EOF but a ret of zero.
+			 */
+			if (ret != 1)
```
keithbostic: Won't this change mean that we'll ignore corruption in the file? That is, if we see random garbage, fscanf will fail to convert it to an integer, it will return 0 and we'll quit thinking we've hit EOF.

sueloverso: Yes, that is exactly what WT-3826 is giving us, garbage and `ret == 0`. We "quit" meaning we'll stop verifying the records from this thread and move on to the next file and verify those records.

keithbostic: And we are sure to identify gaps in that case?

sueloverso: I believe the answer is yes. There is a larger loop that resets the expected keyspace for the next file.

keithbostic: @sueloverso has explained this to me, I'm merging.

PR #3870, opened 2018-01-09
WT-3079 Further backport fix.

michaelcahill:
Remove some old code that changed the target_pages calculation.


michaelcahill:
Note that the code removed here could set `target_pages` to something larger than the calculated value (specifically for DEAD trees).  That in turn could push the `walk_progress > walk_target`.

PR #3872, opened 2018-01-11
WT-3565 Add test with mixed timestamp/no timestamp usage.

sueloverso:


agorrod:
Thanks @sueloverso lgtm

PR #3873, opened 2018-01-11
WT-3831 uninitialized buffer value in statlog server path comparison

keithbostic:
@sueloverso,  here's the valgrind failure: http://build.wiredtiger.com:8080/job/wiredtiger-valgrind/2664/valgrindResult/pid=34127,0x0/


agorrod:
Thanks @keithbostic lgtm - I've switched the valgrind job over to this branch, once testing is done I'm happy to merge.

PR #3874, opened 2018-01-11
WT-3792 LSM version 1 metadata incompatibility

keithbostic:


keithbostic:
I've confirmed this branch fixes the Jenkins `wiredtiger-compat-develop` test.

michaelcahill:

FTR, the change that introduced this bug, WT-3295, is not yet part of a public release of WiredTiger, so both that change and this fix should be in the next release.

PR #3878, opened 2018-01-12
WT-3796 Report a better error message if transaction commit fails

keithbostic:


src/txn/txn.c
```diff
@@ -968,6 +970,17 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
 	return (ret);
 }
 
+/*
+ * __wt_txn_rollback_required --
+ *	Rollback is required, prepare to log a reason.
```
agorrod: Could you extend this comment a bit? Maybe: `Prepare to log a reason if the user attempts to use the transaction to do anything other than rollback.`

keithbostic: > Could you extend this comment a bit?

Pushed in 358b305.

PR #3880, opened 2018-01-15
WT-3799 Idle format threads should not pin timestamps.

michaelcahill:


michaelcahill:
As a starting point for discussion...

test/format/ops.c
```diff
@@ -500,12 +505,10 @@ commit_transaction(TINFO *tinfo, WT_SESSION *session)
...
 		/*
-		 * Update the thread's last-committed timestamp. Don't let the
-		 * compiler re-order this statement, if we were to race with
-		 * the timestamp thread, it might see our thread update before
-		 * the transaction commit.
+		 * Clear the threads' timestamp to prevent anything being
+		 * pinned.
 		 */
-		WT_PUBLISH(tinfo->timestamp, ts);
+		WT_PUBLISH(tinfo->timestamp, 0);
```
agorrod: Note that this is safe because nothing is later attempting to read from the timestamp, so it's OK to move the oldest timestamp forward once the commit is done.

keithbostic: I don't understand why `WT_PUBLISH` is needed at line 499: I think the `__wt_atomic_addv64` guarantees the ordering of the `g.timestamp` read and inserting a write-barrier before reading `g.timestamp` doesn't add anything.

michaelcahill: @keithbostic, fair enough -- since the timestamp thread waits 15 seconds before `oldest_timestamp` can catch up to current, I don't think barriers are necessary here.  I've pushed the change to remove them.

michaelcahill:
Thanks for the reviews @agorrod and @keithbostic: merging.

PR #3881, opened 2018-01-15
WT-3740 race in page dirty-byte decrement.

keithbostic:


michaelcahill:

PR #3882, opened 2018-01-16
WT-3844 Avoid a checkpoint deadlock after lookaside eviction.

michaelcahill:
Make sure the eviction server thread doesn't stop on a limbo page (blocking exclusive access to the page if eviction is no longer necessary).


agorrod:
Thanks @michaelcahill lgtm - I'm going to go ahead and merge.

PR #3883, opened 2018-01-16
WT-3597 Change WT_UPDATE ordering check to be part of verbose timestamp.

agorrod:
Rather than gating it unconditionally on HAVE_DIAGNOSTIC, the condition
won't flag an error to the application, but can generate a log message,
which should be seen in the MongoDB log if verbose=timestamp is enabled.


agorrod:
I'm going to go ahead and merge this change, to unblock bumping MongoDB. @sueloverso please take a look as well.

PR #3885, opened 2018-01-17
WT-3799 Don't have format set oldest_timestamp=0.

michaelcahill:


michaelcahill:
@agorrod, I think this is obvious and straightforward: can you please take a look?

agorrod:
Thanks @michaelcahill lgtm

PR #3889, opened 2018-01-18
WT-3835 cursor remove tries to return a searched-for key when no such key exists

keithbostic:


michaelcahill:

PR #3890, opened 2018-01-19
WT-3829 Make schema operations atomic.

michaelcahill:
For example, if creating "table:T" consisting of "colgoup:T" and "file:T.wt", all of the metadata updates should commit atomically rather than doing multiple auto-committed writes.

Also, when scanning the metadata to see created objects, if a partially created object is found, skip over it rather than making the scan fail.


michaelcahill:
NOT READY FOR REVIEW: only manually tested (by inspection of `wt printlog` output), not tested with MongoDB.

michaelcahill:
@sueloverso, I believe this branch is now ready for review.  Can you please take a look?

I have added tests for the "metadata:create" cursor behavior change (skip over incomplete metadata entries rather than treating them as an error).  This allows MongoDB's startup test that reconciles `_mdb_catalog` with `WiredTiger.wt` to see all of the valid entries, rather than stopping part way through if it hits a bad one and deleting the rest.

I have *not* added tests that calling `session->create("table:T")` now writes a single log record rather than having 3 separate commits, but I have confirmed that manually via `wt printlog`.  If you have thoughts on how to write that test reliably, please let me know.

sueloverso:
@michaelcahill I have added a new test confirming log records. Once PR testing completes, I'll merge.

sueloverso:
I should note that I ran the new test in a develop tree and it fails in the expected ways counting the log records.

PR #3891, opened 2018-01-19
WT-3860 lint

keithbostic:


src/utilities/util_main.c
```diff
@@ -41,7 +41,7 @@ usage(void)
 	    "\t" "compact\t  compact an object\n"
 	    "\t" "copyright copyright information\n"
 	    "\t" "create\t  create an object\n"
-	    "\t" "downgrade\t  downgrade an database\n"
+	    "\t" "downgrade downgrade an object\n"
```
agorrod: I'm surprised the explicit tab character is removed. Oh I see - downgrade is the longest work, should we just use spaces or a more explicit separator?
Is it possible to downgrade anything other than an entire database?

keithbostic: Thank you, @agorrod, you're correct, and I've backed out the "database" to "object" part of the change.

michaelcahill:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

PR #3892, opened 2018-01-21
WT-3846 Added a navigator chart to the top of every HTML page showing an inte…

fedorova:
…rval of execution across all threads (trace files). Using the navigator chart, we can easily navigate to another interval (by clicking on that interval) without having to go back to the main page. Unified the legend, so we have just one for all trace files, and it appears on top of the HTML file for that interval.


fedorova:
autobuild: retest this please

agorrod:
Thanks @fedorova lgtm - I'm going to go ahead and merge.

PR #3893, opened 2018-01-22
WT-3847 Add stub api for prepared transaction

bvpvamsikrishna:
Added stub api for prepared transaction.
Updated begin_transaction to take an additional argument ignore_prepare.
Added a python test test_prepare01.py to check that invoking prepare_transaction will return error, for test_prepare01.py, copied test-txn01.py and added prepare_transaction call before commit_transaction.


lukech:
autobuild: retest this please

bvpvamsikrishna:
@sulabhM  @nehakhatri5 
Request for a review of this pull request.

test/suite/test_prepare01.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def check(self, cursor, committed, total):
+        # The cursor itself should see all of the records.
+        if cursor != None:
+            cursor.reset()
+            self.assertEqual(self.cursor_count(cursor), total)
+
+        # Read-uncommitted should see all of the records.
+        # Snapshot and read-committed should see only committed records.
+        self.check_txn_cursor('isolation=read-uncommitted', total)
```
nehakhatri5: If you agree with the comment above about stubbing ignore_prepare option, then it would be good to test begin_transaction with an input of ignore_prepare option.

test/suite/test_prepare01.py
```diff
@@ -0,0 +1,177 @@
...
+    # Open a session with snapshot isolation, and assert the number of records
+    # visible to the cursor matches the expected value.
+    def check_txn_session(self, level, expected):
+        s = self.conn.open_session(level)
+        cursor = s.open_cursor(self.uri, None)
+        s.begin_transaction()
+        self.assertEqual(self.cursor_count(cursor), expected)
+        s.close()
+
+    def check(self, cursor, committed, total):
```
nehakhatri5: In this test function check() are we testing anything additional to what is already tested in test_txn01.py?

dist/api_data.py
```diff
@@ -1180,6 +1184,13 @@ def __cmp__(self, other):
         choices=['background', 'off', 'on']),
 ]),
 
+'WT_SESSION.prepare_transaction' : Method([
```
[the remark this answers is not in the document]
bvpvamsikrishna: I think we can add this info after implementing the functionality, instead of for stub.

test/suite/test_prepare01.py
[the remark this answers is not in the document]
bvpvamsikrishna: I think the comment is enough, it is to check that same cursor should see all records irrespective of isolation level. The keyword is 'itself'.

bvpvamsikrishna: This check is to test that all are visible from with in the same cursor different from others and as per comment.

test/suite/test_prepare01.py
```diff
@@ -0,0 +1,177 @@
...
+        for i in xrange(self.nentries):
+            if i > 0 and i % (self.nentries / 37) == 0:
+                self.check(cursor, committed, i)
+                self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+                    lambda: self.session.prepare_transaction(), msg)
+                self.session.commit_transaction()
+                committed = i
+                self.session.begin_transaction()
+
+            if self.key_format == 'S':
```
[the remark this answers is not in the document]
bvpvamsikrishna: Frankly I am not sure of the logic either. I copied test_txn01.py, so that the prepare_transaction can be tested that it is not breaking any basic transaction properties.

dist/api_data.py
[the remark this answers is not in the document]
bvpvamsikrishna: I have added this scope to https://jira.mongodb.org/browse/WT-3850

lukech:
autobuild: retest this please

bvpvamsikrishna:
@sulabhM @nehakhatri5, request to check the updates.

src/txn/txn.c
```diff
@@ -448,6 +448,15 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
 		txn_global = &S2C(session)->txn_global;
 		WT_RET(__wt_txn_parse_timestamp(session, "read", &ts, &cval));
 
+		/*
+		 * Prepare transactions are supported only in timestamp build,
+		 * as it is to support mongodb exclusively.
+		 */
+		WT_RET(__wt_config_gets_def(session,
```
nehakhatri5: I think that ignore_prepare option does not have a dependency on read_timestamp. So parsing of ignore_prepare should be independent/outside of read_timestamp parsing block.

bvpvamsikrishna: Prepare is supported only with timestamp builds, as prepare_timestamp is mandatory in prepare_transaction.  Hence all logic related to prepare is being kept with in timestamp macro inclusion.

examples/c/ex_all.c
```diff
@@ -900,6 +900,23 @@ transaction_ops(WT_SESSION *session_arg)
 	error_check(session->commit_transaction(session, NULL));
 	/*! [transaction isolation] */
 
+	/*! [transaction prepare] */
+	/*
+	 * An active transaction can be prepared, if successful will guarantee
+	 * a commit. No operations, apart from commit/rollback operation is
+	 * allowed on a prepared transaction, including session/cursor
+	 * operations. At the moment, prepare transaction is designed to
+	 * support mongodb exclusively.
```
sulabhM: Maybe we should say "MongoDB".

test/suite/test_prepare01.py
```diff
@@ -0,0 +1,177 @@
...
+        ('fix-f', dict(uri='file:text_txn01',key_format='r',value_format='8t')),
+        ('fix-t', dict(uri='table:text_txn01',key_format='r',value_format='8t')),
+        ('row-f', dict(uri='file:text_txn01',key_format='S',value_format='S')),
+        ('row-t', dict(uri='table:text_txn01',key_format='S',value_format='S')),
+    ])
+
+    # Return the number of records visible to the cursor.
+    def cursor_count(self, cursor):
+        count = 0
+        # Column-store appends result in phantoms, ignore records unless they
```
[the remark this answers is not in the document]
sulabhM: I didn't understand this part " self.value_format == 'S' or cursor.get_value() == 0xab ".
Why do we count in if value_format is 'S' irrespective of the value.

bvpvamsikrishna:
@agorrod, request to review this pull request.

dist/s_string.ok
```diff
@@ -772,6 +772,7 @@ gostring
 gostruct
 goutf
 gt
+guarantee
```
agorrod: I'm surprised this is required?

dist/s_string.ok
```diff
@@ -945,6 +946,7 @@ mkdir
 mmap
 mmrand
 mnt
+mongodb
```
agorrod: I agree with Sulabh's comment further down - you should use MongoDB - which already has an entry in this file.

examples/c/ex_all.c
agorrod: A comment with this information would be useful in wiredtiger.in.

dist/api_data.py
```diff
@@ -1182,6 +1186,13 @@ def __cmp__(self, other):
...
+    Config('prepare_timestamp', '', r'''
+        set the prepare timestamp for the updates of the current transaction.
+        The supplied value should not be older than any active read timestamps.
+        See @ref transaction_timestamps'''),
```
agorrod: I'd like it if the description explains that the configuration option is required.

examples/c/ex_all.c
```diff
@@ -900,6 +900,23 @@ transaction_ops(WT_SESSION *session_arg)
...
+	 */
+	error_check(session->open_cursor(
+	    session, "table:mytable", NULL, NULL, &cursor));
+	error_check(session->begin_transaction(session, NULL));
+	cursor->set_key(cursor, "key");
+	cursor->set_value(cursor, "value");
+	session->prepare_transaction(session, "prepare_timestamp=2a");
```
agorrod: I think you need to error check the `prepare_transaction` call.

src/txn/txn.c
agorrod: It's generally better to avoid mentioning MongoDB in cases like this - you could drop the second half of this sentence.

examples/c/ex_all.c
agorrod: I prefer:
```
Prepare a transaction which guarantees a subsequent commit will succeed.
Only commit and rollback are allowed on a transaction after it has been prepared.
```

examples/c/ex_all.c
bvpvamsikrishna: If error check is done, build will fail as currently prepare_transaction will return error. Will update this example as part of wt-3848.

bvpvamsikrishna:
@agorrod, Request to review the updates done.

agorrod:
lgtm - I'm going to merge.

PR #3895, opened 2018-01-22
WT-3865 avoid cache misses in WiredTiger tree-walk and hazard functions

keithbostic:
WT_REF.state is a volatile, avoid pounding on memory in the heavily-used tree-walk and hazard pointer functions.

@michaelcahill, feel free to discard this branch without further discussion if you don't think it's worth doing. (I made a note to myself based on some debugging in WT-3767 to review if WT_REF.state should really be declared non-volatile -- after reviewing that today, I think it's better being volatile, but we are forcing a bunch of cache misses in some heavily-used code.)

### DIFF
--- a/src/btree/bt_walk.c
+++ b/src/btree/bt_walk.c
@@ -293,7 +293,7 @@ __tree_walk_internal(WT_SESSION_IMPL *session,
 	WT_DECL_RET;
 	WT_PAGE_INDEX *pindex;
 	WT_REF *couple, *couple_orig, *ref;
-	uint32_t slot;
+	uint32_t current_state, slot;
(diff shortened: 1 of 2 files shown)


michaelcahill:
autobuild: retest this please

keithbostic:
autobuild: retest this please

agorrod:
@keithbostic Can you a performance testing against this change to see if it helps?

keithbostic:
@agorrod 
> Can you a performance testing against this change to see if it helps?

Probably not worth the effort; I'll discard the branch.

agorrod:
but I like the idea..

PR #3896, opened 2018-01-22
WT-3871 format should read from a timestamp

keithbostic:
@michaelcahill, in thinking about the upcoming work in WT-3766 and WT-3805, I thought it would be useful to explicitly read from a timestamp and verify we're seeing the record we expect to see.

### DIFF
--- a/test/format/ops.c
+++ b/test/format/ops.c
@@ -456,8 +456,8 @@ snap_check(WT_CURSOR *cursor,
  * begin_transaction --
  *	Choose an isolation configuration and begin a transaction.
  */
-static u_int
-begin_transaction(TINFO *tinfo, WT_SESSION *session)
+static void
+begin_transaction(TINFO *tinfo, WT_SESSION *session, u_int *iso_configp)
 {
 	u_int v;
 	char *config, config_buf[64];
@@ -491,9 +491,9 @@ begin_transaction(TINFO *tinfo, WT_SESSION *session)
 		}
 		break;
 	}
-	testutil_check(session->begin_transaction(session, config));
+	*iso_configp = v;
 
-	return (v);
+	testutil_check(session->begin_transaction(session, config));
 }
 
 /*
@@ -667,7 +667,7 @@ ops(void *arg)
 		 */
 		if (!SINGLETHREADED &&
 		    !intxn && mmrand(&tinfo->rnd, 1, 100) >= g.c_txn_freq) {
-			iso_config = begin_transaction(tinfo, session);
+			begin_transaction(tinfo, session, &iso_config);
 			snap =
 			    iso_config == ISOLATION_SNAPSHOT ? snap_list : NULL;
 			intxn = true;
--- a/test/format/ops.c
+++ b/test/format/ops.c
@@ -478,9 +478,10 @@ begin_transaction(TINFO *tinfo, WT_SESSION *session, u_int *iso_configp)
 		config = "isolation=snapshot";
 		if (g.c_txn_timestamps) {
 			/*
-			 * Update the thread's read timestamp with the current
-			 * value to prevent the oldest timestamp moving past our
-			 * allocated timestamp before the commit completes.
+			 * Set the thread's read timestamp to the current value
+			 * before allocating a new read timestamp. This
+			 * guarantees the oldest timestamp won't move past the
+			 * allocated timestamp before the transaction begins.
 			 */
 			tinfo->read_timestamp = g.timestamp;
 			tinfo->read_timestamp =
@@ -495,6 +496,12 @@ begin_transaction(TINFO *tinfo, WT_SESSION *session, u_int *iso_configp)
 	*iso_configp = v;
 
 	testutil_check(session->begin_transaction(session, config));
+
+	/*
+	 * It's OK for the oldest timestamp to move past a running query, clear
+	 * the thread's read timestamp, it no longer needs to be pinned.
+	 */
+	tinfo->read_timestamp = 0;
 }
 
 /*
@@ -926,12 +933,6 @@ deadlock:			++tinfo->deadlock;
 			++tinfo->rollback;
 			break;
 		}
-
-		/*
-		 * Clear the thread's read timestamp: it no longer needs to
-		 * be pinned.
-		 */
-		tinfo->read_timestamp = 0;
 
 		intxn = false;
 		snap = NULL;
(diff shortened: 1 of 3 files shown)


test/format/ops.c
```diff
@@ -481,6 +452,50 @@ snap_check(WT_CURSOR *cursor,
...
+	default:
+		v = ISOLATION_SNAPSHOT;
+		config = "isolation=snapshot";
+		if (g.c_txn_timestamps) {
+			/*
+			 * Update the thread's read timestamp with the current
+			 * value to prevent the oldest timestamp moving past our
+			 * allocated timestamp before the commit completes.
+			 */
+			tinfo->read_timestamp = g.timestamp;
```
sueloverso: Remove this line since we overwrite it on the next line.

test/format/util.c
```diff
@@ -614,9 +614,11 @@ timestamp(void *arg)
...
 		for (i = 0; i < g.c_threads; ++i) {
 			tinfo = tinfo_list[i];
-			this_ts = tinfo->timestamp;
-			if (this_ts != 0 &&
-			    this_ts < oldest_timestamp)
+			this_ts = tinfo->commit_timestamp;
+			if (this_ts != 0 && this_ts < oldest_timestamp)
+				oldest_timestamp = this_ts;
+			this_ts = tinfo->read_timestamp;
+			if (this_ts != 0 && this_ts < oldest_timestamp)
```
sueloverso: Minor nit: the comment above the for loop needs to be updated too.

test/format/ops.c
```diff
@@ -652,10 +667,7 @@ ops(void *arg)
 		 */
 		if (!SINGLETHREADED &&
 		    !intxn && mmrand(&tinfo->rnd, 1, 100) >= g.c_txn_freq) {
-			iso_config = isolation_config(&tinfo->rnd, session);
-			testutil_check(
-			    session->begin_transaction(session, NULL));
-
+			iso_config = begin_transaction(tinfo, session);
```
sueloverso: I'd prefer if `begin_transaction` was slightly renamed because I totally read it as `session->begin_transaction` in my head. Anything like `begin_transaction_iso` or `begin_transaction_config` or `start_transaction`.

test/format/ops.c
keithbostic: That's correct, I think, notice there's an atomic operation between the two assignments. Basically, we set the thread's timestamp (so the timestamp thread won't go past the "current" value), then update the current value (which is an atomic operation so we're guaranteed the first assignment completes before the current value changes), then we assign the result of the atomic operation because that's the timestamp we're really going to use for our reads.

test/format/ops.c
keithbostic: I think reading it as `WT_SESSION.begin_transaction` is OK, that's actually what it does (and the name matches the function `commit_transaction` that we use to commit).

I'll make setting `iso_config` a side-effect rather than a return value, that should help.

test/format/util.c
keithbostic: Thanks, fixed.

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

lukech:
autobuild: retest this please

test/format/ops.c
```diff
@@ -914,6 +927,12 @@ deadlock:			++tinfo->deadlock;
 			break;
 		}
 
+		/*
+		 * Clear the thread's read timestamp: it no longer needs to
+		 * be pinned.
+		 */
+		tinfo->commit_timestamp = 0;
```
agorrod: The comment says "read timestamp", but the code says `commit_timestamp` Should this be clearing both?

keithbostic: > The comment says "read timestamp", but the code says `commit_timestamp` Should this be clearing both?

Thanks, Alex -- code was wrong, we should be clearing the read-timestamp here, at the end of the transaction.

I've pushed a fix.

agorrod:
Stress sanitizer job is running on this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/

test/format/ops.c
```diff
@@ -481,6 +452,51 @@ snap_check(WT_CURSOR *cursor,
...
+			testutil_check(__wt_snprintf(
+			    config_buf, sizeof(config_buf),
+			    "read_timestamp=%" PRIx64, tinfo->read_timestamp));
+			config = config_buf;
+		}
+		break;
+	}
+	*iso_configp = v;
+
+	testutil_check(session->begin_transaction(session, config));
```
michaelcahill: It would be cleaner to clear `tinfo->read_timestamp` here, rather than in `ops`.  There's no reason to prevent the application's timestamp moving forward because WT already copes with the `oldest_timestamp` moving past a running query.

In other words `begin_transaction` consumes the `read_timestamp`, just like `commit_transaction` consumes the `commit_timestamp`.

test/format/ops.c
michaelcahill: I think the comment needs something.  Maybe:

```
Set the thread's read timestamp to the current value before allocating a new read timestamp.
This guarantees that the oldest timestamp won't move past the allocated timestamp before
the transaction begins.
```

test/format/ops.c
keithbostic: Both comments fixed in 4b5cf52.

keithbostic:
> That's fine and exercises the timestamp read path but it's different to MongoDB's use case, where read timestamps lag writes on purpose to only return changes that are committed by a majority of replica set nodes.

My goal was to push lookaside eviction and timestamps a little more. I thought this would be sufficient because read timestamps will lag the oldest timestamp (but I agree it's a relatively small lag, depending on thread interleaving). Format threads do fall behind in our testing, including in eviction, especially with artificially small caches -- format found problems in update chain handling when debugging both the cursor reserve and modify operations.

I was hesitant to let read timestamps fall too far behind the oldest timestamp because of the recent problems with format cache timeouts.

Format already has code to read from a checkpoint, and it would be relatively easy to adapt that work to read from timestamps as well, and we could lag it as much as we want.

How would you suggest we configure that?

agorrod:
The stress tests were stable with this change, I'm inclined to take the branch as-is, and then consider extending read support via another mechanism later or in a different test program.

@michaelcahill is that OK with you?

PR #3897, opened 2018-01-23
WT-3766 Remove LAS entries if they're visible and no longer needed.

sueloverso:
@keithbostic and @michaelcahill  here's the code version of our discussion. I don't know if we have a solid test to know we're taking this path. It passes the python LAS test and normal tests.


src/cache/cache_las.c
```diff
@@ -859,17 +866,59 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+			continue;
+		}
+
+		/*
 		 * TODO it would also be good to remove entries in lookaside
 		 * from live files that have aged out.  If we track for each
 		 * entry whether it was the on-page value chosen by
 		 * reconciliation, we can safely remove entries from that point
 		 * on (for the given key) that are visible to all readers.
 		 */
```
keithbostic: This comment isn't necessary any more, is it?

src/cache/cache_las.c
```diff
@@ -859,17 +866,59 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+					    saved_key->data, saved_key->size));
+				/*
+				 * If it is a birthmark record, even the
+				 * oldest key can be removed because it is
+				 * on disk already and now visible. Otherwise
+				 * we don't remove the first entry.
+				 */
+				if (upd_type != WT_UPDATE_BIRTHMARK)
+					continue;
+			}
```
keithbostic: The `__wt_txn_visible_all` call is relatively expensive (including a timestamp read-lock). I think we could avoid repeating that call once we find a visible record with something like this:
```
bool found_visible = false;

/* Save a copy of each new key. */
if (saved_key == NULL || saved_key->size != las_key.size ||
    memcmp(saved_key->data, las_key.data, las_key.size) != 0) {
        found_visible = false;
        WT_ERR(__wt_cursor_get_raw_key(cursor, saved_key));
        if (!WT_DATA_IN_ITEM(saved_key))
                WT_ERR(__wt_buf_set(session, saved_key,
                    saved_key->data, saved_key->size));
}

/*
 * If this entry isn't globally visible we cannot remove it.
 * If it is visible then perform additional checks to see
 * whether it has aged out of a live file.
 */
if (!found_visible) {
        if (!__wt_txn_visible_all(session, txnid, val_ts))
                continue;
        found_visible = true;
         
        /*
         * If it is a birthmark record, even the oldest key
         * can be removed because it is on disk already and
         * now visible. Otherwise we don't remove the first
         * entry.
         */
        if (upd_type != WT_UPDATE_BIRTHMARK)
                continue;
}
++remove_cnt;
```

src/cache/cache_las.c
```diff
@@ -768,15 +768,22 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 	WT_CACHE *cache;
 	WT_CURSOR *cursor;
 	WT_DECL_RET;
-	WT_ITEM *key, las_key;
-	uint64_t cnt, las_counter, las_pageid, remove_cnt;
+	WT_ITEM *key, las_key, las_timestamp, las_value, *saved_key;
```
keithbostic: `saved_key` could probably be a scratch buffer.

src/cache/cache_las.c
sueloverso: fixed.

src/cache/cache_las.c
sueloverso: Hmm, I'm torn on this. Your suggested code is more convoluted than what I have. If it is important to avoid the visibility call for performance it would work. But the question is, do we care? This is the sweep thread, not a critical path. The readlock mentioned is only for odd-sized timestamps and is not used in 8-byte timestamps.

sueloverso: @michaelcahill we're going to let you be the tiebreaker here! @keithbostic falls more on the side of avoiding the visibility call for performance. I fall on the side of keeping the flow more readable (IMO) because it isn't a perf path.

sueloverso:
This is not yet ready for merge. I ran some tests with this branch on mongod and it dropped core.

agorrod:
I've switched one of the stress testers onto this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/configure

michaelcahill:
@sueloverso, I pushed a fix in b2fd059 based on running the `secondary_performance` workload against a replica set, which forces use of lookaside.  I think it's pretty solid if you're happy with these changes.

src/cache/cache_las.c
```diff
@@ -727,10 +729,30 @@ __las_sweep_init(WT_SESSION_IMPL *session)
...
+	 * but possibly better, alternative might be to review all lookaside
+	 * blocks in the cache in order to get rid of them, and slowly review
+	 * lookaside blocks that have already been evicted.
+	 */
+	cache->las_sweep_cnt =
+	    (uint64_t)WT_MAX(100, cache->las_entry_count / 30);
+
 	__wt_spin_lock(session, &cache->las_sweep_lock);
+
 	/* If no files have been dropped, there's nothing to do. */
```
keithbostic: The comment `there's nothing to do` is no longer correct.

src/cache/cache_las.c
```diff
@@ -768,15 +790,23 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
 	uint32_t las_id, session_flags;
+	uint8_t upd_type;
 	int notused;
+	bool newkey;
 
 	cache = S2C(session)->cache;
 	cursor = NULL;
 	key = &cache->las_sweep_key;
 	remove_cnt = 0;
+	WT_CLEAR(saved_key);
```
keithbostic: `saved_key` can be a scratch buffer, there's no need to allocate new memory.

src/cache/cache_las.c
```diff
@@ -859,20 +884,66 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+		} else
+			newkey = false;
+
+		/*
+		 * If this entry isn't globally visible we cannot remove it.
+		 * If it is visible then perform additional checks to see
+		 * whether it has aged out of a live file.
+		 */
+		if (!__wt_txn_visible_all(session, txnid, val_ts))
+			continue;
```
keithbostic: Did we decide the `__wt_txn_visible_all` call isn't expensive enough to worry about? Or is there a case where one visible entry doesn't imply subsequent entries can be discarded?

keithbostic:
@sueloverso, [this failure](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/18388/) hit a Jenkins `Identified problems` check @michaelcahill put in last October, but I think it's real.

This branch hasn't had the recent test/format changes do more timestamp work, so I think it's possible this is a real failure introduced in this branch.

src/cache/cache_las.c
michaelcahill: I'm not worried about it: I haven't seen evidence of it in perf traces.

sueloverso:
I pushed a change that removed the birthmark piece and I have not been able to reproduce a failure since then (all repros were under 8 iterations, I'm up to 25 with the fix).

michaelcahill:
@sueloverso, fair enough but then I think we need to talk some more about what's going on.

If eviction wrote a page using the newest versions of all of the updates, and all the content in lookaside becomes obsolete, it should be possible for sweep to remove it all.

sueloverso:
I agree and will continue to debug it. But I wanted to get the fix into Jenkins testing to make sure it actually fixes and run reliably there.

michaelcahill:
@sueloverso, I've merged the changes from WT-3878 into this branch (via develop).  Can you please take another look at sweeping birthmarks?

michaelcahill:
I looked at this failure a little today.  What I'm seeing when the snapshot test is failing is that format has 4 items in `snap_list` and the transaction has done 3 updates.  The item that is failing is the value that doesn't appear in `txn->mod`.  I think that means either it was a format `READ` operation, or the update failed but format added it to the track list.

It would help (me) a lot to include `op` in `SNAP_OPS` rather than having to interpret the `deleted` and `insert` fields.  It would also be good to catch this error during the search, if possible.  I'll take a run at that tomorrow if nobody beats me to debugging this one.

keithbostic:
@michaelcahill 
> It would help (me) a lot to include op in SNAP_OPS rather than having to interpret the deleted and insert fields.

Pushed in WT-3887 (#3912).

keithbostic:
@michaelcahill 
> The item that is failing is the value that doesn't appear in `txn->mod`.  I think that means either it was a format `READ` operation, or the update failed but format added it to the track list.

keithbostic:
Sorry about the close/re-open, clicked the wrong button.

keithbostic:
@michaelcahill 
> What I'm seeing when the snapshot test is failing is that format has 4 items in `snap_list` and the transaction has done 3 updates.  The item that is failing is the value that doesn't appear in `txn->mod`.  I think that means either it was a format `READ` operation, or the update failed but format added it to the track list.

It's a `READ` operation -- I've now seen it when it's the only thing in `snap_list`, that is, a single operation that was suddenly not repeatable.

It's also consistently `WT_NOTFOUND` for me, a read operation reporting something like:
```
snapshot-isolation 0000004628.00/opqrstuvwxyzabcde search mismatch
        expected {0000004628/LMNOPQRST}
found {deleted}
```

michaelcahill:
I've pushed some more changes in e47b594, including an attempt to have sweep only give up at block boundaries.  I also fixed a bug with `saved_key`: it should be the real key from the original page, but instead we were saving the raw lookaside key.  I've rearranged how we track the sweep key to start the next walk: I think it's clearer now.

My attempt to avoid storing a birthmark followed by a tombstone in the lookaside table got bogged down -- it works apart from `rollback_to_stable`, which relies on the tombstones when removing birthmarks.  I went some way down the path of rewriting birthmarks rather than removing them in the rollback code, but didn't get that working.

That said, I'm still seeing some problems, including:

```
[1517467809:677153][1520:0x7ff0d1cb2740], t, file:wt, WT_SESSION.verify: lookaside entry count went negative with decrement of 1
```

That doesn't surprise me: if a sweep and a call to `__las_page_instantiate` collide, they can't get accurate counts of entries removed from the lookaside table because `cursor->remove` will return zero to both callers if they race.  We could round to zero in this case rather than failing, I guess.

However, I think there's another problem if sweep collides with a page read: sweep could remove a birthmark, then some earlier update in the list could be restored by `__las_page_instantiate` before it is removed by sweep, leading to incorrect reads.

Maybe we should consider a read/write lock, held exclusive by sweep and shared by `__las_page_instantiate`?  I think that would address both of these issues.  We may need to further tune the `cache->las_sweep_cnt` calculation and the delay between calls to `__wt_las_sweep` to avoid long latencies during reads...

michaelcahill:
autobuild: retest this please

lukech:
autobuild: retest this please

lukech:
autobuild: retest this please

PR #3899, opened 2018-01-24
WT-3852 Add per-key timestamp validation checking as an option to assert()

sueloverso:


sueloverso:
@keithbostic your review is mostly because I added a comment to flags.py reminding me and others exactly what a new flag should look like in a list, per our discussion yesterday. @agorrod this essentially undoes your verbose changes from a week or two ago and restores the original checks, except under the assert flags rather than diagnostic. @michaelcahill because you were discussing this with Dan yesterday.

src/txn/txn.c
```diff
@@ -630,39 +630,50 @@ __txn_commit_timestamp_validate(WT_SESSION_IMPL *session)
...
+				continue;
+			/*
+			 * Check for consistent per-key timestamp usage.
+			 * If timestamps are or are not used originally then
+			 * they should be used the same way always. Check
+			 * timestamps are used in order.
+			 */
+			op_used_ts =
+			    __wt_timestamp_iszero(&op->u.upd->timestamp);
+			upd_used_ts = __wt_timestamp_iszero(&upd->timestamp);
```
agorrod: I thought maybe there was a case that went from not using timestamps to using timestamps (once)? At the end of initial sync in MongoDB.

sueloverso: I believe there is a table that goes from not using timestamps to using them for all keys. @dgottlieb is that correct? This config is primarily for the mdb catalog where some keys always have timestamps and some keys never have timestamps so the broader per-table settings are not able to be used.

dgottlieb: I fear I've written more than what you were both asking about. If either of you have further clarifying questions, I'll try to give a more focused response. :)

> I thought maybe there was a case that went from not using timestamps to using timestamps (once)? At the end of initial sync in MongoDB.

That is generally correct. The more nuanced story is that initial sync has a few phases. The first cloning phase I believe will never timestamp any writes. Then there is an oplog application phase which will timestamp writes. After that is an optional "document fetching" phase which will not timestamp writes. For now, my goal for verification is to not check anything during the entirety of initial sync. We don't support reading from a node at a time when it was initial syncing so it's less important. If there was interest, I believe there is a path to appropriate alter the table settings between each phase.

What I want to do when MongoDB comes out of initial sync, is have storage set each table with its appropriate verification setting. I suspect iterating through each table may be a penalty we don't want to pay in general (think customers with thousands of tables), but I think it's important to do this at least in our evergreen debug build tests.

> I believe there is a table that goes from not using timestamps to using them for all keys.

There are two cases I know of that mostly* match this description. The first being the initial sync process @agorrod mentioned. The second is for index builds. Index builds on an existing collection can be updated in two ways. One is the collection scan to initialize the index. The other is when a user performs a concurrent insert/update/delete to a collection while the index is building. The first is not timestamped, the second is. Foreground index builds don't allow concurrent updates during the build process. Those tables should experience a sudden change in timestamps as @sueloverso mentioned. Background index builds do allow concurrent updates and will have a mix of both varieties. Notably this can include a timestamped write to an index entry followed by a non-timestamped write.

Similarly to initial sync, I don't plan on enforcing any timestamp behavior on indexes until they are finished building.

The purpose of this ticket, as @sueloverso described, is for the `_mdb_catalog`. Once again, I won't enforce any timestamping until initial sync completes. After the node is in steady-state replication, this table has the unique property of containing data for replicated and non-replicated collections. To keep parity with existing rollback behavior, we don't want to rollback updates to non-replicated collections. 

Accidentally timestamping some writes to non-replicated metadata documents can cause a `rollback_to_stable` or, more simply, a read transaction at a timestamp to return an incorrect result (that is, if MongoDB's high-level catalog allowed for versioned reads). Likewise, accidentally forgetting to timestamp a write to a replicated collection's metadata document can result in similar corruption.

The work to squash out any remaining (known) untimestamped writes is nearing completion (tickets linked from https://jira.mongodb.org/browse/SERVER-30809 ). All of those tickets have to do with timestamping writes to the `_mdb_catalog`, which is why I'm particularly interested in having independent validation that we got it right.

agorrod: Thanks for the explanation @dgottlieb that all makes sense to me - I'll go ahead and merge.

PR #3900, opened 2018-01-24
WT-3863 Use the tsc/nanosecond ratio when parsing the operation tracking files.

fedorova:
This way, we show the time units in nanoseconds, not in clock cycles.


agorrod:
Thanks @fedorova lgtm - I'll merge.

PR #3901, opened 2018-01-24
WT-3854 Remove write lock from commit path leaving old entries in queue.

sueloverso:
Not quite fully tested or ready for merge.


sueloverso:
@michaelcahill  and @agorrod I think this is ready for review. I posted some WT-specific results in the ticket. The `test_timestamp_abort` was able to get over 12% more operations completed on this branch.

src/txn/txn_timestamp.c
```diff
@@ -699,19 +734,17 @@ void
...
 		return;
+	flags = txn->flags;
+	LF_CLR(WT_TXN_PUBLIC_TS_COMMIT);
 
-	__wt_writelock(session, &txn_global->commit_timestamp_rwlock);
-	TAILQ_REMOVE(&txn_global->commit_timestamph, txn, commit_timestampq);
-	--txn_global->commit_timestampq_len;
-	__wt_writeunlock(session, &txn_global->commit_timestamp_rwlock);
-	F_CLR(txn, WT_TXN_PUBLIC_TS_COMMIT);
+	WT_PUBLISH(txn->clear_ts_queue, true);
```
agorrod: I think a comment above here would be useful. Something like:
Notify other operations that our transaction is inactive, and can be safely cleaned up from the queue. It's not necessary for the entry to be removed prior to returning since there is no guarantee around the order of commit completing.

sueloverso: Fixed. Thanks.

agorrod:
Thanks @sueloverso

PR #3902, opened 2018-01-25
WT-3840 Dump more information on data corruption in diagnostic mode

keithbostic:
@agorrod, I ended up increasing the stack buffer size from 2KB to 4KB, and I added an error complaint, if we truncate a message we should notice it.


agorrod:
@keithbostic sorry for the delay on this one - I'm in transit now, but will get to it when I land tomorrow.

@bvpvamsikrishna It would be good if you can take a look at this PR, if it doesn't interrupt prepared transactions work.

bvpvamsikrishna:
lgtm

agorrod:
Thanks @keithbostic this lgtm - it should help diagnose data corruption errors in the future.

PR #3903, opened 2018-01-25
WT-3876 Added missing operation tracking macros. Those missing macros resulte…

fedorova:
…d in inaccurate log files. Added a version of WT_RET macro that tracks the exit from the function before returning.

### DIFF
--- a/dist/s_function
+++ b/dist/s_function
@@ -73,10 +73,11 @@ done
 for f in `find bench examples ext src test -name '*.[ci]'`; do
 	sed -n -e '/API_CALL.*;$/,/API_END.*;/{=;p;}' \
 	       -e '/LSM_.*ENTER*;$/,/LSM_.*LEAVE*;/{=;p;}' \
+	       -e '/WT_TRACK_OP_INIT/,/WT_TRACK_OP_END/{=;p;}' \
 	       -e '/va_start/,/va_end/{=;p;}' $f | \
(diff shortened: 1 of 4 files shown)


keithbostic:
@fedorova, I simplified one case, but otherwise lgtm, I'll merge this once it passes testing.

I also added `WT_TRACK_OP_INIT/WT_TRACK_OP_END` to the list of start/stop pairs we'll complain about if there's a return in-between them.

Some of these functions are complex enough that I'd suggest wrapping them instead of trying to identify all of the return cases, mostly so we avoid introducing a bug in the future if someone doesn't realize there's tracking being done and adds a random return.

In other words, leave the underlying function alone, but rename it and make it static, and then add something like this that just wraps it. That way we're pretty unlikely to introduce a bug in the future by adding a new return case.

```
/*
 * __wt_cond_wait_signal --
 *      Wait on a mutex, optionally timing out.  If we get it before the time
 * out period expires, let the caller know.
 */
void
__wt_cond_wait_signal(WT_SESSION_IMPL *session, WT_CONDVAR *cond,
    uint64_t usecs, bool (*run_func)(WT_SESSION_IMPL *), bool *signalled)
{
        WT_TRACK_OP_DECL;
        
        WT_TRACK_OP_INIT(session);
        __cond_wait_signal_tracked(session, cond, usecs, run_func, signalled);
        WT_TRACK_OP_END(session);
}
```

PR #3908, opened 2018-01-29
WT-3877 WT_CONN_EVICTION_NO_LOOKASIDE is cleared unconditionally

keithbostic:


agorrod:
Thanks @keithbostic lgtm.

PR #3909, opened 2018-01-29
WT-3881 Skip validating update structures from this txn.

sueloverso:
@agorrod and @dgottlieb here's a PR that solves an issue Dan saw over the weekend when the commit timestamp is set in the middle or the end of the transaction, the new validate code was comparing against updates from earlier in the transaction. I modified the validation to skip updates from its own transaction. The code does not set the timestamp in the update structures until much later in the commit process, long past the time where we're allowed to error. So we need to skip them.


agorrod:
Thanks @sueloverso lgtm.

PR #3910, opened 2018-01-30
WT-3848 enhance post conditions of prepared transactions(approach-1)

bvpvamsikrishna:
WIP : For review of approach. In this approach, we change the function pointers of API such that the new functions will return error, and we reset the pointers back to original form in commit / rollback.
To enhance post conditions of prepared transactions to check validity of operation invoked.
Only session level API's are modified.
Cursor level API's are not updated to return error yet.


michaelcahill:
@bvpvamsikrishna, I prefer #3911 -- restoring those session pointers scares me, and I think it would be *much* more code to handle cursor operations as well with this approach.

bvpvamsikrishna:
Approach is abandoned, becoming too complex for cursor operations

PR #3911, opened 2018-01-30
WT-3848 enhance post conditions of prepared transactions

bvpvamsikrishna:
WIP : For review of approach. In this approach we return error, if transaction state is prepared, we will add this check in all possible API's which should not be supported after a prepare-transaction call.
To enhance post conditions of prepared transactions to check validity of operation invoked.
Only session level API's are modified.
Cursor level API's are not updated to return error yet.


test/suite/test_prepare02.py
```diff
@@ -0,0 +1,88 @@
...
+class test_prepare02(wttest.WiredTigerTestCase, suite_subprocess):
+    def test_timestamp_range(self):
+        if not wiredtiger.timestamp_build():
+            self.skipTest('requires a timestamp build')
+
+        self.session.create("table:mytable", "key_format=S,value_format=S")
+        # Commit after prepare is permitted
+        self.session.begin_transaction()
+        self.session.prepare_transaction()
+        with self.expectedStderrPattern(''):
```
michaelcahill: @bvpvamsikrishna, can you please add some cursor operations in here as well (e.g., `c['key'] = 'value' for some cursor `c` opened in `self.session`)?

michaelcahill: Sorry, I see in the pull request description that cursor operations are not yet checked.

It seems like that wouldn't be too hard to add with this approach (compared to #3910)?

bvpvamsikrishna:
@michaelcahill @agorrod 
Request to check updates made for cursor operations post conditions as well.
I need to add automated tests, I have tested some operations manually.

dist/s_string.ok
```diff
@@ -945,6 +946,7 @@ mkdir
 mmap
 mmrand
 mnt
+mongodb
```
agorrod: The two changes to this file can be removed.

src/cursor/cur_log.c
```diff
@@ -292,8 +292,11 @@ static int
 __curlog_reset(WT_CURSOR *cursor)
 {
 	WT_CURSOR_LOG *cl;
+	WT_SESSION_IMPL *session;
 
 	cl = (WT_CURSOR_LOG *)cursor;
+	session = (WT_SESSION_IMPL *)cursor->session;
+	WT_RET(__wt_txn_context_prepare_check(session, false));
```
agorrod: I'm surprised the log cursor needs a change for prepare.

src/include/misc.i
```diff
@@ -193,15 +193,44 @@ __wt_snprintf_len_incr(
...
+	if (!requires_txn && F_ISSET(&session->txn, WT_TXN_RUNNING))
+		WT_RET_MSG(session, EINVAL,
+		    "%s: not permitted in a running transaction",
+		    session->name);
 	if (requires_txn && !F_ISSET(&session->txn, WT_TXN_RUNNING))
 		WT_RET_MSG(session, EINVAL,
 		    "%s: only permitted in a running transaction",
 		    session->name);
-	if (!requires_txn && F_ISSET(&session->txn, WT_TXN_RUNNING))
+#ifdef HAVE_TIMESTAMPS
```
agorrod: Why didn't you call `__wt_txn_context_prepare_check` rather than duplicating the code?

bvpvamsikrishna:
@agorrod @sueloverso I believe this pull request is ready for review, request to review.
I have added the list of session level API's which can / cannot be invoked in a prepared transaction state in the jira ticket [WT-3848](https://jira.mongodb.org/browse/WT-3848). Request to review the list and provide your opinion.
None of the cursor API's can be invoked in prepared transaction state irrespective of cursor type.  Request to indicate in the review if any of them should be allowed to be invoked in prepared state.

Currently I have added checks in all the session call's to check for the prepared state.
Alternative is to add this check inside SESSION_API_CALL macro so that code changes will be at minimal places as this macro is called by all the session level API's.  If we take this approach, then some session level API's will check for the transaction running state as well. I felt little odd thinking that with in a session API, we check for transaction state twice, once for prepare state and other for running state. Request your thoughts on this.

If this is too much text to read, we can have a small discussion meeting based on your availability

bvpvamsikrishna:
@agorrod @sueloverso  I have committed the changes I was referring to above " moving the prepare check to SESSION_API_CALL". Ready for your review.

test/suite/test_prepare02.py
```diff
@@ -0,0 +1,136 @@
...
+            self.assertRaises(wiredtiger.WiredTigerError,
+                lambda: self.session.verify("table:mytable", None))
+        with self.expectedStderrPattern(" not permitted in a"):
+            self.assertRaises(wiredtiger.WiredTigerError,
+                lambda: self.session.snapshot("name=test"))
+        with self.expectedStderrPattern(" not permitted in a"):
+            self.assertRaises(wiredtiger.WiredTigerError,
+                lambda: self.session.timestamp_transaction(
+                    "commit_timestamp=2a"))
+        self.session.log_printf("Wrote 100 records")
```
sueloverso: It isn't clear to me why `log_printf` and `log_flush` are allowed. There are other, equally non-active-transaction methods that we're forbidding such as `transaction_pinned_range` and some other cursor types, etc. These need comments both in the WT code and test code if they stay.

Some session methods are not listed here, such as `transaction_pinned_range` and `log_flush`. (Related to earlier comment about alphabetizing to make it easier to detect missing methods.)

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,223 @@
...
+            cursor.set_key(self.genkey(i))
+            cursor.set_value(self.genvalue(i))
+            self.session.prepare_transaction()
+            #cursor.set_value(self.genvalue(i))
+            #with self.expectedStderrPattern(preparemsg):
+            #    self.assertRaises(wiredtiger.WiredTigerError,
+            #            lambda: cursor.set_value(self.genvalue(i)))
+            with self.expectedStderrPattern(preparemsg):
+                self.assertRaises(wiredtiger.WiredTigerError,
+                        lambda: cursor.insert())
```
sueloverso: These can be a single line with `assertRaisesWithMessage`.

bvpvamsikrishna:
@sueloverso 
I have addressed your comments. Request to check.

In test_prepare02.py, the session level operations invocation is in the same order as used in function assignment within __open_session method. This will be easy to verify any missing functions.

transaction_pinned_range is the only function that is missing in test_prepare02.py. I have created [WT-3916](https://jira.mongodb.org/browse/WT-3916) to address this as the invocation of transaction_pinned_range is complex.

sueloverso:
@bvpvamsikrishna I did not get to this today. I will review it again first thing tomorrow.

src/session/session_api.c
```diff
@@ -291,7 +291,13 @@ __session_reconfigure(WT_SESSION *wt_session, const char *config)
 	WT_SESSION_IMPL *session;
 
 	session = (WT_SESSION_IMPL *)wt_session;
-	SESSION_API_CALL(session, reconfigure, config, cfg);
+	/*
+	 * Even though this operation is not allowed in prepared transaction
+	 * state, true is passed indicating the operation is allowed in
```
sueloverso: I suggest removing some of the repetitiveness in this comment:
* Even though this operation is not allowed while the transaction is prepared,
* we pass in true so that the running transaction check below takes precedence
* over the prepared transaction check.

src/session/session_api.c
```diff
@@ -1469,7 +1476,13 @@ __session_begin_transaction(WT_SESSION *wt_session, const char *config)
 	WT_SESSION_IMPL *session;
 
 	session = (WT_SESSION_IMPL *)wt_session;
-	SESSION_API_CALL(session, begin_transaction, config, cfg);
+	/*
+	 * Even though this operation is not allowed in prepared transaction
```
sueloverso: Same wording here and the other places below.

test/suite/test_prepare02.py
```diff
@@ -0,0 +1,123 @@
...
+    return '%x' % t
+
+class test_prepare02(wttest.WiredTigerTestCase, suite_subprocess):
+    def test_prepare_session_operations(self):
+        if not wiredtiger.timestamp_build():
+            self.skipTest('requires a timestamp build')
+
+        self.session.create("table:mytable", "key_format=S,value_format=S")
+        cursor = self.session.open_cursor("table:mytable", None)
+        # Session operations not permitted after prepare_transaction
```
sueloverso: I think this comment should be expanded because I thought it specifically referred to the begin/prepare lines directly below it. I think it should:
1. Have a blank line after the open_cursor call.
2. Say something like:
 \# 
\# Test the session methods that are forbidden after the transaction is prepared.
 \#

test/suite/test_prepare02.py
```diff
@@ -0,0 +1,123 @@
...
+        self.session.begin_transaction()
+        self.session.prepare_transaction()
+        msg = "/ not permitted in a/"
+        # Below operations are not supported in prepared state. Operations
+        # are listed in the same order as declared with in session
+        # structure and assigned in open_session. Any function missed below
+        # is to be considered as supported in prepare transaction state.
```
sueloverso: I recommend a bit of a rewording:
 \#
 \# The operations listed below are not supported in the prepared state.
 \#
 \# The operations are listed in the same order as they are declared in the session
 \# structure. Any function missing below is allowed in the prepared state.
 \#

test/suite/test_prepare02.py
```diff
@@ -0,0 +1,123 @@
...
+                "commit_timestamp=2a"), msg)
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+            lambda:self.session.checkpoint(), msg)
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+            lambda: self.session.snapshot("name=test"), msg)
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+            lambda:self.session.transaction_sync(), msg)
+        self.session.rollback_transaction()
+
+        # Commit after prepare is permitted
```
sueloverso: Minor nit - add a period at the end of the sentences in the next three comments.

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        session.create, but report errors more completely
+        """
+        try:
+            self.session.create(name, args)
+        except:
+            print('**** ERROR in session.create("' + name + '","' + args + '") ***** ')
+            raise
+
+    # Create and session and test cursor operations.
```
sueloverso: I think the comment here should be 'Create the table and test cursor operations.'

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+
+    def assertCursorHasNoKeyValue(self, cursor):
+        keymsg = '/requires key be set/'
+        valuemsg = '/requires value be set/'
+        self.assertRaisesWithMessage(
+            wiredtiger.WiredTigerError, cursor.get_key, keymsg)
+        self.assertRaisesWithMessage(
+            wiredtiger.WiredTigerError, cursor.get_value, valuemsg)
+
+    def session_create(self, name, args):
```
sueloverso: Was this just for your debugging purposes? It doesn't seem to be needed in this test since the create is not expected to fail. I think it can be removed now.

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+        # Check insert and reset operations
+        for i in range(0, self.nentries):
+            self.session.begin_transaction()
+            cursor.set_key(self.genkey(i))
+            cursor.set_value(self.genvalue(i))
+            self.session.prepare_transaction()
+            #cursor.set_value(self.genvalue(i))
+            #with self.expectedStderrPattern(preparemsg):
+            #    self.assertRaises(wiredtiger.WiredTigerError,
+            #            lambda: cursor.set_value(self.genvalue(i)))
```
sueloverso: These commented out lines can be removed.

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+        preparemsg = "/ not permitted in a prepared transaction/"
+        if self.tablekind == 'row':
+            keyformat = 'key_format=S'
+        else:
+            keyformat = 'key_format=r'  # record format
+        if self.tablekind == 'fix':
+            valformat = 'value_format=8t'
+        else:
+            valformat = 'value_format=S'
+        create_args = keyformat + ',' + valformat
```
sueloverso: This sort of thing should be part of the scenarios so that you can just access them in one location. You can see an example in `test_timestamp04.py` in the `types` section rather than hardcoded here.

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+
+        self.pr('creating session: ' + create_args)
+        self.session_create(tablearg, create_args)
+        self.pr('creating cursor')
+        cursor = self.session.open_cursor(tablearg, None, None)
+        self.assertCursorHasNoKeyValue(cursor)
+        self.assertEqual(cursor.uri, tablearg)
+
+        # Check insert and reset operations
+        for i in range(0, self.nentries):
```
sueloverso: While 10 is pretty small is there a reason (I see it is used below for next/prev) to check the error case every iteration? If it works the first insert, it should work on all the later ones too.

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+            i -= 1
+
+        self.assertEqual(i, -1)
+        self.assertEqual(prevret, wiredtiger.WT_NOTFOUND)
+        self.assertCursorHasNoKeyValue(cursor)
+
+        # Check search, update, remove, reserve, reconfigure operations.
+        cursor.reset()
+        self.assertCursorHasNoKeyValue(cursor)
+
```
sueloverso: I suggest setting key to `nentries / 2` in case we decide even 10 entries is too many.
A comment here about what you're doing with these keys/values would be helpful.
 \# Search for a specific key.
 \# Verify we get the expected error and then later we can update and remove it.

test/suite/test_prepare03.py
```diff
@@ -0,0 +1,211 @@
...
+        cursor.remove()
+
+        # Check search_near operation
+        cursor.set_key(self.genkey(10))
+        self.session.begin_transaction()
+        self.session.prepare_transaction()
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+            lambda:cursor.search_near(), preparemsg)
+        self.session.commit_transaction()
+        if self.uri == 'lsm':
```
sueloverso: We definitely need a comment here about why LSM needs this handling here.

src/cursor/cur_std.c
```diff
@@ -823,6 +823,7 @@ __wt_cursor_close(WT_CURSOR *cursor)
 
 	session = (WT_SESSION_IMPL *)cursor->session;
 
+	WT_RET(__wt_txn_context_prepare_check(session, false));
```
agorrod: I don't believe WT_CURSOR::close should do a prepare check - we want applications to be able to release resources, and this API call doesn't rely on any prepare semantics.

agorrod:
Thanks @bvpvamsikrishna lgtm - I'm going to merge.

PR #3912, opened 2018-01-31
WT-3887 log operations in format snapshot-isolation tracking

keithbostic:


test/format/ops.c
```diff
@@ -372,11 +375,11 @@ snap_check(WT_CURSOR *cursor,
 				return (ret);
 
 		/* Check for simple matches. */
-		if (ret == 0 && !start->deleted &&
+		if (ret == 0 && start->vset &&
```
sueloverso: This (and the others) has become less readable with `vset` instead of `deleted`. What about checking `start->op != REMOVE`?

keithbostic: > This (and the others) has become less readable with vset instead of deleted. What about checking `start->op != REMOVE`?

Pushed in 187fdd217.

PR #3914, opened 2018-02-01
WT-3890 Clear the flag after removal and add a write barrier.

sueloverso:


src/txn/txn_timestamp.c
```diff
@@ -693,9 +693,10 @@ __wt_txn_set_commit_timestamp(WT_SESSION_IMPL *session)
 		TAILQ_FOREACH_SAFE(qtxn, &txn_global->commit_timestamph,
 		    commit_timestampq, txn_tmp) {
 			if (qtxn->clear_ts_queue) {
-				qtxn->clear_ts_queue = false;
 				TAILQ_REMOVE(&txn_global->commit_timestamph,
 				    qtxn, commit_timestampq);
+				WT_WRITE_BARRIER();
+				qtxn->clear_ts_queue = false;
```
keithbostic: I simpler way to do the same thing is:
```
698,699c698
<                               WT_WRITE_BARRIER();
<                               qtxn->clear_ts_queue = false;
---
>                               WT_PUBLISH(qtxn->clear_ts_queue, false);
```

sueloverso: Fixed

PR #3915, opened 2018-02-01
WT-3888 cursor search-near can return the wrong value for "exact"

keithbostic:


agorrod:
@nehakhatri5 please take a look and review this in your morning.

agorrod:
Thanks @keithbostic lgtm, it would be great if Neha can take a look to. For context - the failure here is that we guarantee when returning from a WT_CURSOR method that the cursor is positioned. If there is an error the cursor should exit being positioned on the same key it was positioned on at the start of the call.

keithbostic:
@nehakhatri5, happy to walk you through this in a bit more detail, I kind of tossed this over the fence w/o a lot of explanation. In summary...

When we do a cursor.search-near, we return an `exact` integer value that's < 0, == 0 or > 0 depending on whether we're returning an exact match to the searched-for key, or a key that's less-than or greater-than the searched-for key. The problem comes because in some fairly rare cases, we call cursor.next and cursor.prev inside of cursor.search-near.

Basically, if we do a search and don't find anything useful (imagine we find the key but it's invisible to us), we'll do a cursor.next to find a close-by key. If that fails (imagine there's nothing after our position in the tree), we'll do a cursor.search to get back to where we were, and then do a cursor.prev to find a close-by key.

The problem was returning `exact`: cursor.next would succeed and then replace the key that the user originally gave us with the next key (the key is in `WT_CURSOR.key` or `WT_CURSOR.recno`), and then everything after that point would be based on the next key, not the original key the user gave us. If cursor.next succeeded and we could return the key it found, everything was fine, but if cursor.next succeeded, but we couldn't return the key it returned, and then we did the `cursor.search + cursor.prev` step, we'd do it based on the wrong key.

The change was to restore the user's original key before taking that final `cursor.search + cursor.prev` step.

I think the way I did it is safe, although it depends on our ability to call `__cursor_state_restore` multiple times without it causing problems.

nehakhatri5:
Thanks Keith, LGTM.

PR #3919, opened 2018-02-02
WT-3891 A couple more places needing the sweep walk rwlock.

sueloverso:
@michaelcahill my run hit a place where I needed to also use the sweepwalk rwlock. @agorrod I also added it to the rollback to stable code. While in there, I noticed that the entry count is not being updated after the removal like it is in all the other remove loops. I fixed that. Please let me know if my change is not correct.


agorrod:
Thanks @sueloverso this change looks good to me - I'm going to go ahead and merge. I've also create WT-3892 to try and get better testing coverage of race conditions in lookaside moving forward.

PR #3920, opened 2018-02-05
WT-3891 Take the sweep walk rwlock whenever removing content from lookaside

keithbostic:


src/cache/cache_las.c
```diff
@@ -18,6 +18,28 @@
 	(WT_SESSION_IGNORE_CACHE_SIZE | WT_SESSION_READ_WONT_NEED |	\
 	WT_SESSION_NO_RECONCILE)
 
+/*
+ * __las_set_read_uncommitted --
+ *	Switch to read-uncommitted.
+ */
```
sueloverso: I'd prefer these be balanced (in naming) and both general even if we only have one usage at the moment. How about `las_set_isolation(session, iso, *saved)` and then `las_restore_isolation(session, saved)`?

src/cache/cache_las.c
```diff
@@ -330,6 +352,51 @@ __wt_las_cursor_close(
...
+	WT_ITEM las_key;
+	uint64_t las_counter, las_pageid;
+	uint32_t las_id;
+
+	*decrp = 0;
+
+	conn = S2C(session);
+
+	__wt_writelock(session, &conn->cache->las_sweepwalk_lock);
+
```
sueloverso: Too much whitespace.

src/cache/cache_las.c
```diff
@@ -330,6 +352,51 @@ __wt_las_cursor_close(
...
+	/*
+	 * Search for the block's unique prefix and step through all matching
```
sueloverso: The term 'unique prefix' brings other things to mind. How about "the block's unique page and btree identifiers"? Or even just "the block's unique identifiers".?

src/cache/cache_las.c
```diff
@@ -330,6 +352,51 @@ __wt_las_cursor_close(
...
+	 * records, removing them.
+	 */
+	for (ret = __wt_las_cursor_position(cursor, btree_id, pageid);
+	    ret == 0; ret = cursor->next(cursor)) {
+		WT_ERR(cursor->get_key(cursor,
+		    &las_pageid, &las_id, &las_counter, &las_key));
+
+		/*
+		 * Confirm the search using the unique prefix; if not a match,
```
sueloverso: Here too.

src/cache/cache_las.c
```diff
@@ -546,19 +632,33 @@ __wt_las_insert_block(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
...
+			WT_TRET(__wt_txn_rollback(las_session, NULL));
+	}
+
+	__las_pop_isolation(las_session, saved_isolation);
+
 	if (insert_cnt > 0) {
-		(void)__wt_atomic_add64(
-		    &S2C(session)->cache->las_entry_count, insert_cnt);
-		WT_ERR(__las_insert_block_verbose(session, multi));
+		if (ret == 0) {
```
sueloverso: While this makes sense with careful reading, a comment would be good here.

src/cache/cache_las.c
```diff
@@ -546,19 +632,33 @@ __wt_las_insert_block(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
...
+			(void)__wt_atomic_add64(
+			    &conn->cache->las_entry_count,
+			    insert_estimate - insert_cnt);
+
```
sueloverso: These are all related, no need for every line to have extra whitespace.

src/cache/cache_las.c
```diff
@@ -627,76 +727,41 @@ __wt_las_cursor_position(WT_CURSOR *cursor, uint32_t btree_id, uint64_t pageid)
...
-		 */
-		if (las_pageid != pageid || las_id != btree_id)
-			break;
+	WT_ERR(__wt_txn_begin(las_session, NULL));
 
-		WT_ERR(cursor->remove(cursor));
-		++remove_cnt;
-	}
-	WT_ERR_NOTFOUND_OK(ret);
+	ret = __las_remove_block(
```
sueloverso: I think a comment here (or at the top) about "We are really a wrapper function setting up the cursor, transaction and needed isolation level. The worker function will acquire needed locks to remove the identified entries." (I don't like that wording exactly, but it's the idea.)

src/cache/cache_las.c
```diff
@@ -958,8 +1030,9 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 		}
 
 		WT_ERR(cursor->remove(cursor));
-		++remove_cnt;
+		++decrement_cnt;
 	}
+
```
sueloverso: Remove whitespace

src/cache/cache_las.c
```diff
@@ -983,15 +1056,24 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 	if (0) {
 err:		__wt_buf_free(session, sweep_key);
 	}
+	if (local_txn) {
+		if (ret == 0)
+			ret = __wt_txn_commit(session, NULL);
+		else
+			WT_TRET(__wt_txn_rollback(session, NULL));
+
```
sueloverso: Remove whitespace

src/cache/cache_las.c
keithbostic: Done in ea154c9.

src/cache/cache_las.c
keithbostic: I disagree. The first line is a variable initialization and the second is code in the function, I think there should be whitespace between them.

src/cache/cache_las.c
keithbostic: I went with:
```
Search for the block's unique btree ID and page ID prefix and step
through all matching records, removing them.
```

src/cache/cache_las.c
keithbostic: I went with:
```
Confirm the record matches; if not a match, we're done
searching for records for this page.
```

src/cache/cache_las.c
keithbostic: I went with:
```
If the transaction successfully committed and we inserted records,
adjust the final entry count. We may have also deleted records, 
but we must have intended to insert records to be in this function
at all, checking the insert count is sufficient.
```

src/cache/cache_las.c
keithbostic: I put the increment/decrement together, but left space before the `__las_insert_block_verbose` call.

src/cache/cache_las.c
keithbostic: I went with:
```
This is an external API for removing records from the lookaside
table, first acquiring a lookaside table cursor and enclosing
transaction, then calling an underlying function to do the work.
```

src/cache/cache_las.c
keithbostic: Why? The locking call isn't related to the `for` loop in any way, and the lock call isn't near at the start of the `for` loop, so think they should be separate.

src/cache/cache_las.c
keithbostic: Done.

PR #3923, opened 2018-02-07
WT-3895 Fixes to manage the timestamp queue.

sueloverso:
@agorrod There are fixes for the commit timestamp queue bugs that the mongodb tests found. First we need to modify the `first_commit_timestamp` under the lock and later than it was before. Also we want to remove ourselves from the queue first because our timestamp may be smaller than it was last time and we could end up with a list loop. The alternative would have been to walk the entire queue removing all to-be-cleared transactions from the list but we want to minimize the walks.


agorrod:
Thanks @sueloverso lgtm

PR #3925, opened 2018-02-08
WT-3901 Fixed the log corruption bug.

fedorova:
Flushing the operation tracking memory buffer to the file works as follows:
1. We remember the old value of the file offset.
2. We flush the buffer, opening the file if it has not been open yet.
3. We add the return value to the remembered offset.

The problem occurs in the situation when we flush to a file that has not been open yet. Here is what happens:
1. We remember the old value. It is 0, because the file has not been opened yet.
2. As we flush the buffer, the file is opened. BUT: opening the file writes some metadata to it and advances the offset to 12. We flush the buffer at offset 12. We return the number of bytes flushed.
Upon return we should add the number of bytes flushed to 12, the previous offset, but:

3. We add the returned value (number of bytes flushed) to the remembered offset, which was 0!
So the offset is now incorrect. Next time we flush the buffer, we corrupt the data.

This fix avoids remembering the old offset value (step 1). It calls the flush function, remembers the number of flushed bytes it returns, then in step 3 it gets the fresh value of the offset and adds to it the returned bytes.


fedorova:
This bug occurs with some compilers, but not with others. That is why I have not seen it occur in my earlier testing.

fedorova:
@keithbostic, this is definitely a more elegant and less error-prone way to handle the offset. I don't know why I had it incremented in optrack.h previously. Seems silly in retrospect. Thanks for noticing this!

keithbostic:
@fedorova, I pushed a version of what we discussed, back to you!

PR #3926, opened 2018-02-11
WT-3805 Avoid reading lookaside pages in truncate fast path.

keithbostic:


keithbostic:
@agorrod, I think this branch is working, but I'd like to point a stress tester at it for a day before we merge it.

keithbostic:
@agorrod, the commits in 9375406,  4f9977f and bb57722 are core library changes, and probably deserve more careful review.

src/btree/bt_read.c
```diff
@@ -571,15 +571,6 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
 	    ref->page->dsk == NULL ||
 	    F_ISSET(ref->page->dsk, WT_PAGE_LAS_UPDATE));
 
-	/*
-	 * If reading for a checkpoint, there's no additional work to do, the
-	 * page on disk is correct as written.
-	 */
-	if (session->dhandle->checkpoint != NULL) {
```
agorrod: @keithbostic I don't understand how this change relates to the comment that went with the commit:

> Checkpoints can read pages based on deleted cells

This code path is dealing with reading content from a checkpoint, not reading pages while creating a checkpoint. It feels as though the code was reasonable as-was, am I missing something?

keithbostic: That code is really just `if we're reading from a checkpoint`, and it (1) asserts the previous state was on-disk and (2) jumps to "done" as we have no further work to do.

When reading in a checkpoint, with respect to (1), `__inmem_row_int()` creates internal pages that contain children with a WT_REF.state of `WT_REF_DELETED` (see the code for `WT_CELL_ADDR_DEL`).

Then, with respect to (2), in the case of the checkpoint then reading one of the child `WT_REF_DELETED` pages, we need to re-instantiate the deleted nature of the page, otherwise we'll return incorrect information, key/value pairs will magically reappear in the checkpoint.

I'm certain about the first point, I'm pretty sure about the second point (although I didn't force it to happen to prove it).

The right things happen if that clause just disappears.

agorrod: I believe you, but I'd like to talk this one through. I'm a bit surprised a page that's part of a completed checkpoint has such refs.

keithbostic: Reconciliation writes `WT_CELL_ADDR_DEL` cells in the case of `WT_CHILD_PROXY` after running through this code in `__rec_child_deleted`:
```
/*
 * If the original page cannot be freed, we need to keep a slot on the
 * page to reference it from the parent page.
 *
 * If the delete is not visible in this checkpoint, write the original
 * address normally.  Otherwise, we have to write a proxy record.
 */
if (__wt_txn_visible(
    session, page_del->txnid, WT_TIMESTAMP_NULL(&page_del->timestamp)))
        *statep = WT_CHILD_PROXY;
```
tbh, I don't really remember how it works, but it doesn't appear to differentiate between checkpoints and eviction.

agorrod:
I've switched http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer-ppc over to run on this branch.

agorrod:
This change lgtm - if we get several hours of stress testing without new failures I'll go ahead and merge.

michaelcahill:
Thanks @keithbostic, lgtm -- I'll merge.

PR #3927, opened 2018-02-12
WT-3912 fast-delete pages don't re-instantiate the delete transaction's timestamp.

keithbostic:


agorrod:
Thanks @keithbostic lgtm - nice catch!

michaelcahill:
Yes, good catch -- thanks, @keithbostic!

PR #3928, opened 2018-02-13
WT-3805 Avoid reading lookaside pages in fast truncate path

keithbostic:


keithbostic:
Here's a CONFIG that will cause this branch to fast-delete pages with lookaside information as well as instantiate fast-delete pages that have lookaside information.
```
#  RUN PARAMETERS
abort=0
alter=0
auto_throttle=1
backups=0
bitcnt=4
bloom=1
bloom_bit_count=46
bloom_hash_count=32
bloom_oldest=0
cache=40
cache_minimum=40
checkpoints=on
checkpoint_log_size=102
checkpoint_wait=98
checksum=uncompressed
chunk_size=10
compaction=0
compression=none
data_extend=0
data_source=file
delete_pct=9
dictionary=0
direct_io=0
encryption=rotn-7
evict_max=0
file_type=row-store
firstfit=0
huffman_key=0
huffman_value=0
independent_thread_rng=0
in_memory=0
insert_pct=13
internal_key_truncation=1
internal_page_max=15
isolation=random
key_gap=11
key_max=36
key_min=28
leaf_page_max=13
leak_memory=0
logging=0
logging_archive=1
logging_compression=none
logging_file_max=467190
logging_prealloc=0
long_running_txn=1
lsm_worker_threads=4
merge_max=4
mmap=1
modify_pct=17
ops=0
prefix_compression=0
prefix_compression_min=3
quiet=0
read_pct=55
rebalance=1
repeat_data_pct=22
reverse=0
rows=1000000
runs=100
salvage=1
split_pct=73
statistics=0
statistics_server=0
threads=4
timer=4
transaction_timestamps=0
transaction-frequency=52
value_max=2477
value_min=2
verify=1
wiredtiger_config=
write_pct=6
```

keithbostic:
autobuild: retest this please

agorrod:
@keithbostic this change generally looks good to me. Given the complexity of the code it definitely deserves another reviewer - @sueloverso would you mind taking a look?

I'd also like for this to get merged directly into develop, we should be able to do that by clicking "edit" next to the title of the pull request and changing the base branch (which will be even easier once WT-3909 is merged).

src/btree/bt_delete.c
```diff
@@ -322,25 +343,43 @@ __wt_delete_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
-	 * to different pages, and we still have to find them all for an abort.
+	 * In the first case, check to see if the transaction has been resolved,
+	 * maybe we don't have to do that much work.
 	 */
+	page_del = ref->page_del;
+	if (page_del != NULL &&
+	    __wt_txn_visible_all(session,
+	    page_del->txnid, WT_TIMESTAMP_NULL(&page_del->timestamp)))
+		__wt_free(session, ref->page_del);
+
```
keithbostic: @michaelcahill, would you please confirm this visibility test change? The idea is that if we're about to instantiate a deleted page, and the transaction that did the delete is now globally visible, we don't need to worry about the transaction aborting and we can skip a bunch of work.

src/cache/cache_las.c
```diff
@@ -353,6 +353,102 @@ __wt_las_cursor_close(
 	return (ret);
 }
 
+/*
+ * __wt_las_page_skip_locked --
```
sueloverso: Are there any changes in these two functions other than movement and adding the function name prefix? The code movement makes detecting any changes and reviewing them hard.

src/btree/bt_delete.c
```diff
@@ -322,25 +343,43 @@ __wt_delete_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * In the first case, we have a page reference structure, in the second
 	 * second, we don't.
```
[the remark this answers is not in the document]
keithbostic: Thanks, fixed in 8a4124b75.

src/cache/cache_las.c
keithbostic: No changes other than movement and function name prefixes.

agorrod:
Thanks @keithbostic I'm good to go with this change now, there was a format test failure due to a timeout, but I suspect that's more related to adding truncate support to format than this change.

I'm going to leave it to you to go ahead and merge in case you know something I don't.

keithbostic:
@michaelcahill 
> I added an assertion that we're only throwing away ref->page_del if it was rolled back -- if we can repeatedly get into the truncate code after an earlier truncate commits, we should talk some more.

I don't believe there's any path without rolling back, and I like the assert. I thought about doing something like setting the ID to abort, but then realized it wouldn't work for timestamps.

Just an idea: but would it be reasonable for `__wt_txn_visible_all` to check `WT_TXN_ABORTED` (similarly to how `__wt_txn_visible` and `__txn_visible_id` do, then this would feel like less of a special case.

michaelcahill:
@keithbostic, I could imagine a `__wt_txn_obsolete` that returns `txnid == WT_TXN_ABORTED || __wt_txn_visible_all(txnid)`, but it doesn't make sense for `__wt_txn_visible_all` to return `true` for an aborted update (which is not visible to anyone).

PR #3929, opened 2018-02-14
WT-3912 reading a truncated page must set timestamp

keithbostic:


src/txn/txn.c
```diff
@@ -870,12 +870,19 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+			if (F_ISSET(txn, WT_TXN_HAS_TS_COMMIT)) {
+				WT_UPDATE **upd;
+
 				__wt_timestamp_set(
 				    &op->u.ref->page_del->timestamp,
 				    &txn->commit_timestamp);
+				for (upd = op->u.ref->page_del->update_list;
+				    *upd != NULL; ++upd)
+					__wt_timestamp_set(&(*upd)->timestamp,
+					    &txn->commit_timestamp);
```
agorrod: It's a corner case of a corner case, but do we need to skip updating the timestamp on any update that already had one set? As for `WT_TXN_OP_BASIC_TS` above.

keithbostic: Great question.

I believe this isn't a possible path, that is, `WT_SESSION_LOGGING_INMEM` must always be set if we're in `__wt_txn_modify_page_delete()`, implying these updates can't be type `WT_TXN_OP_BASIC_TS`. That said, I'm not completely sure of that.

@michaelcahill, would you please weigh in on this question? Specifically, should `__wt_txn_modify_page_delete()` be checking `WT_SESSION_LOGGING_INMEM`, even if just for clarity, and is there a better way to do this work?

agorrod: I feel like that is a bug - I doubt it's relevant at the moment because it will only occur if a timestamp user is calling `WT_SESSION::set_timestamp` and then doing truncate operations, and I don't think MongoDB does that. On the other hand, what this will do is cause an update that is supposed to have a timestamp as configured with `WT_SESSION::set_timestamp` to have a timestamp as it is at `WT_SESSION::commit`. I don't see a simple fix here, since I suspect fixing this would require adding a new WT_TXN_OP_REF_DELETE_TS which causes this to be a no-op.
So - I believe @michaelcahill should confirm my theory and potential solution before we invest the effort.

michaelcahill: I think this case (and indeed `WT_TXN_OP_BASIC_TS`) can be handled by checking if `__wt_timestamp_iszero` on the update before setting the commit timestamp.

At some point during the timestamp work we decided it was okay to reserve zero as a special case: the separate optype precedes that decision, I think.  I'll take a run at this, it doesn't seem like it has to be hard.

michaelcahill: Scratch that, I remember why this is more complicated.  It may need a separate type for the truncate operation...

michaelcahill: I didn't get to this today, my current thinking is to remove the `WT_TXN_OP_BASIC_TS` operation type and instead add a boolean `needs_timestamp` field that tracks whether the update should be assigned a timestamp during commit.  That will work for processing commit timestamps for truncate operations as well.

If someone wants to take this before I get to it, be my guest.

agorrod: To complete the circle on this comment - the reason why there was a separate type is that some tables don't have timestamps applied, so checking for zero isn't enough. At a point in time we thought there might be a category of tables in that bucket, but it is currently only the WiredTiger metadata, so I've pushed a change that uses a zero timestamp as an indicator and does an explicit check for the metadata.

michaelcahill:
Thanks @keithbostic, @agorrod -- lgtm.  I'll merge.

PR #3930, opened 2018-02-14
WT-3905 Save the checkpoint timestamp in metadata. Use it in tests.

sueloverso:
@michaelcahill Here are the changes to store the timestamp that checkpoint is using in the metadata. @agorrod for your review also. And @dgottlieb mostly FYI for API usage. Some considerations:
1. I write a `checkpoint_stable` record into the metadata. This is written for all file records, not just the record in the turtle file.  The record in the turtle file is the only one used to store the timestamp on restart.  I did this for a couple reasons:
- It allows us to see timestamps in other metadata records in a `wt list -v` that may give helpful information in debugging.
- It allows us to extend querying the recovery timestamp in the future to an individual table if we ever wanted.
2. I named it `query_timestamp("get=recovery")`.
3. One benefit of this is that it allowed me to use it in the `timestamp_abort` test with better accuracy on the verification side and removal of an entire table and a bunch of code.
4. Currently, until WT-3906 is complete, this only works on an unclean shutdown. The APIs work and information is correct but the data after restart from a clean shutdown in a collection-like table will not (yet) be as of the stable. The testing will be extended when I do that ticket.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -338,8 +338,7 @@ file_meta = file_config + [
     Config('checkpoint_lsn', '', r'''
         LSN of the last checkpoint'''),
     Config('checkpoint_timestamp', '', r'''
-        stable timestamp of the last checkpoint''',
-        type=string),
+        stable timestamp of the last checkpoint'''),
     Config('id', '', r'''
         the file's ID number'''),
     Config('version', '(major=0,minor=0)', r'''
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -337,6 +337,9 @@ file_meta = file_config + [
         the file checkpoint entries'''),
     Config('checkpoint_lsn', '', r'''
         LSN of the last checkpoint'''),
+    Config('checkpoint_timestamp', '', r'''
+        stable timestamp of the last checkpoint''',
+        type=string),
     Config('id', '', r'''
         the file's ID number'''),
     Config('version', '(major=0,minor=0)', r'''
@@ -1339,7 +1342,7 @@ methods = {
         \c oldest_timestamp and the read timestamps of all active readers, and
         \c stable returns the most recent \c stable_timestamp set with
         WT_CONNECTION::set_timestamp.  See @ref transaction_timestamps''',
-        choices=['all_committed','oldest','pinned','stable']),
+        choices=['all_committed','oldest','pinned','recovery','stable']),
 ]),
 
 'WT_CONNECTION.set_timestamp' : Method([
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -82,12 +82,12 @@ __wt_verbose_timestamp(WT_SESSION_IMPL *session,
 }
 
 /*
- * __wt_txn_parse_timestamp --
- *	Decodes and sets a timestamp.
+ * __wt_txn_parse_timestamp_raw --
+ *	Decodes and sets a timestamp. Don't do any checking.
  */
 int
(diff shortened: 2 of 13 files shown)


src/txn/txn_timestamp.c
```diff
@@ -86,8 +86,8 @@ __wt_verbose_timestamp(WT_SESSION_IMPL *session,
  *	Decodes and sets a timestamp.
  */
 int
-__wt_txn_parse_timestamp(WT_SESSION_IMPL *session,
-     const char *name, wt_timestamp_t *timestamp, WT_CONFIG_ITEM *cval)
+__wt_txn_parse_timestamp(WT_SESSION_IMPL *session, const char *name,
+    wt_timestamp_t *timestamp, WT_CONFIG_ITEM *cval, bool zero_ok)
```
agorrod: I don't love having to change all callers of this function for one case, could you split it up into multiple functions instead? Maybe a new `__wt_txn_parse_timestamp_zero_ok` that is called by `__wt_txn_parse_timestamp`, or a shared internal function?

dist/api_data.py
```diff
@@ -1339,7 +1341,7 @@ def __cmp__(self, other):
         \c oldest_timestamp and the read timestamps of all active readers, and
         \c stable returns the most recent \c stable_timestamp set with
         WT_CONNECTION::set_timestamp.  See @ref transaction_timestamps''',
-        choices=['all_committed','oldest','pinned','stable']),
+        choices=['all_committed','oldest','pinned','recovery','stable']),
```
agorrod: Is the `recovery` timestamp actually the `checkpoint` timestamp? Could we call it that, and have it more generally useful?

dgottlieb: MongoDB only needs to query the value that existed on startup. For an online rollback calling `rollback_to_stable`, MongoDB knows precisely what it has set the stable timestamp to.

This certainly doesn't preclude that there may be some other use case. Perhaps having `recovery` now that is guaranteed to remain constant for the lifetime of the program, and, if it becomes useful, later adding in a `checkpoint` value later that tracks checkpoints as they happen?

dgottlieb:
I imagine this is true, but I wanted to communicate the contract that MongoDB will fulfill and make sure WiredTiger will respond with the behavior I'm anticipating. I don't think I'd be able to tell if the code is doing this (particularly for the second, less-common case):

1. MongoDB starts, the data on disk may be from a crash or a clean shutdown.
2. MongoDB calls `wiredtiger_open`.
3. MongoDB gets the on disk timestamp by querying WiredTiger. So long as MongoDB has taken a stable checkpoint, this value should be meaningful.
4. MongoDB will set the stable timestamp on the `WT_CONNECTION`. Before...
5. Accepting 0+ writes, advancing the stable timestamp 0+ times and taking 0+ checkpoints.
6. Clean shutdown.

Is WiredTiger guaranteed to have a stable timestamp in the most recent checkpoint that can be queried?  Is this still true if MongoDB crashes anywhere in that sequence?

And the less common-case, for example when an administrator restarts a node outside of the replica set to build an index, followed by adding it back in:

1. MongoDB starts, the data on disk (the "original data") is from a clean shutdown.
2. MongoDB calls `wiredtiger_open`.
3. MongoDB **does not** query the on disk checkpoint timestamp.
4. MongoDB **does not** set the stable timestamp on the `WT_CONNECTION`.
5. MongoDB sends writes to WiredTiger, meanwhile taking 0+ "unstable" checkpoints.
6. Clean shutdown (leaving behind the "modified data").

In that case, MongoDB would require the `checkpoint_timestamp` written in the `WiredTiger.wt` file from the "original data" to match what now exists in the "modified data".

PR #3932, opened 2018-02-16
 WT-3867 Bi-weekly WT codebase lint

keithbostic:


michaelcahill:

PR #3934, opened 2018-02-17
WT-3904 Allow log server thread to retry on permission error.

sueloverso:
@agorrod Here is the small change to allow the log server thread to retry on a permission error. We record a message and retry up to 10 times. We reset if it soon succeeds and it if doesn't we will eventually panic. Please review.


agorrod:
Thanks @sueloverso lgtm

PR #3935, opened 2018-02-19
WT-3919 Add Python test cases for cursor cache sweep.

agorrod:


agorrod:
@ddanderson I did a bit of git/JIRA shuffling and re-created your changes. They look fine to me, though the test takes a little while to run - do you think we should consider it for the long suite?

michaelcahill:
autobuild: retest this please

ddanderson:
@agorrod, it looks like it is failing in the same way as before. I find that it runs fine when I run it locally (even with `run.py -j N`) but fails the pull request test on a timeout. I haven't come up with a solution for this.

ddanderson:
This branch was superseded by https://github.com/wiredtiger/wiredtiger/pull/3949 .

PR #3936, opened 2018-02-19
WT-3923 __wt_txn_context_prepare_check() requires API initialization

keithbostic:
@agorrod, I noticed some functions were calling `__wt_txn_context_prepare_check()` explicitly (probably because they didn't call `API_XXX` macros?).

Anyway, that won't work because `__wt_txn_context_prepare_check()` uses `WT_SESSION.name`.

I added `API_XXX` macros in a few places, I don't see any reason that's not safe.

### DIFF
--- a/src/cursor/cur_join.c
+++ b/src/cursor/cur_join.c
@@ -539,17 +539,16 @@ __curjoin_extract_insert(WT_CURSOR *cursor)
 	WT_ITEM ikey;
 	WT_SESSION_IMPL *session;
 
-	cextract = (WT_CURJOIN_EXTRACTOR *)cursor;
 	/*
 	 * This insert method may be called multiple times during a single
 	 * extraction.  If we already have a definitive answer to the
 	 * membership question, exit early.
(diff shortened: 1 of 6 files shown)


keithbostic:
@agorrod:

There are two remaining explicit calls to `__wt_txn_context_prepare_check()`: `__session_log_printf` and `__session_transaction_pinned_range`. Maybe I'm missing something, but I don't see why either of those functions would not be permitted inside of a prepare.

agorrod:
@bvpvamsikrishna please take a look at this one.

bvpvamsikrishna:
@keithbostic 

Currently I have taken a conservative approach for the functions that are allowed to be called in a prepared state. As prepare_transaction support is specific for MongoDB,  no operations are allowed in a prepared state for now, and if we face any issues with this approach, that time we can handle as will be knowing more about the use case.

agorrod:
lgtm to me too @bvpvamsikrishna could you please merge?

PR #3937, opened 2018-02-20
WT-3925 format: salvage dump comparison failure

keithbostic:
@agorrod, I introduced a bug in the last set of format changes, the result is we're doing almost nothing other than remove operations in our format runs.

I'd like to push this as soon as we can, and then I'll watch format testing overnight.


agorrod:
lgtm

PR #3938, opened 2018-02-21
WT-3829 All metadata reads should be at read-uncommitted isolation.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm

PR #3939, opened 2018-02-21
WT-3926 Allow read_timestamp to be set after begin_transaction.

michaelcahill:
Decoupling these means setting a read_timestamp will not be blocked due to a full cache, so is reasonable to do while holding an application lock (which may be necessary to allocate a valid timestamp).

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -1214,6 +1214,10 @@ methods = {
         read using the specified timestamp.  The supplied value must not be
         older than the current oldest timestamp.  This can only be set once
         for a transaction.  @ref transaction_timestamps'''),
+    Config('round_to_oldest', 'false', r'''
+        if read timestamp is earlier than oldest timestamp,
+        read timestamp will be rounded to oldest timestamp''',
+        type='boolean'),
 ]),
 
 'WT_SESSION.rollback_transaction' : Method([]),
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -1161,7 +1161,7 @@ methods = {
         Transactions with higher values are less likely to abort''',
         min='-100', max='100'),
     Config('read_timestamp', '', r'''
-        read using the specified timestamp.  The supplied value should not be
+        read using the specified timestamp.  The supplied value must not be
         older than the current oldest timestamp.  See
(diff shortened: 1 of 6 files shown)


michaelcahill:
@bvpvamsikrishna, can you please review this change?

dist/api_data.py
```diff
@@ -1198,18 +1198,22 @@ def __cmp__(self, other):
...
 'WT_SESSION.timestamp_transaction' : Method([
     Config('commit_timestamp', '', r'''
         set the commit timestamp for the current transaction.  The supplied
-        value should not be older than the first commit timestamp set for the
-        current transaction.  The value should also not be older than the
+        value must not be older than the first commit timestamp set for the
+        current transaction.  The value must also not be older than the
         current oldest and stable timestamps.  See
         @ref transaction_timestamps'''),
+    Config('read_timestamp', '', r'''
```
bvpvamsikrishna: Why we only accept "read_timestamp" but not "round_to_oldest" in set_timestamp ?

sueloverso: I took a look at this PR. The code appears to accept `round_to_oldest` but it is not part of the API definition above. That should be fixed.

src/txn/txn_timestamp.c
```diff
@@ -662,6 +660,108 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+		/*
+		 * This code is not using the timestamp validate function to
+		 * avoid a race between checking and setting transaction
+		 * timestamp.
+		 */
+		WT_RET(__wt_timestamp_to_hex_string(session,
+		    hex_timestamp[0], &ts));
+		__wt_readlock(session, &txn_global->rwlock);
+		if (__wt_timestamp_cmp(&ts, &txn_global->oldest_timestamp) < 0)
+		{
```
sueloverso: KNF - brace at end of previous line.

agorrod:
@michaelcahill I pushed the change @bvpvamsikrishna suggested - does this look OK to you now?

michaelcahill:
Thanks @bvpvamsikrishna, @sueloverso and @agorrod! -- I'll go ahead and merge.

PR #3940, opened 2018-02-21
WT-3906 Add config API to connection close for checkpoint.

sueloverso:
Not ready for merge. Add API and new test for forcing the checkpoint on connection close to write to the stable timestamp or to write out all dirty data.


sueloverso:
@agorrod this is ready for review. I used the same default in `api_data.py` for close as I did for checkpoint.  @dgottlieb FYI.

test/suite/test_timestamp12.py
```diff
@@ -0,0 +1,119 @@
...
+            self.session.begin_transaction()
+            c_op[i] = 1
+            c_coll[i] = 1
+            self.pr("i: " + str(i))
+            self.session.commit_transaction(
+              'commit_timestamp=' + timestamp_str(i))
+
+        # Close and reopen the connection.
+        self.close_conn(self.close_cfg)
+        self.conn = self.setUpConnectionOpen(".")
```
ddanderson: @sueloverso, I'd change these two lines (starting with `self.conn = `) to just:
```
self.open_conn()
```

dist/api_data.py
```diff
@@ -1295,6 +1295,12 @@ def __cmp__(self, other):
     Config('leak_memory', 'false', r'''
         don't free memory during close''',
         type='boolean'),
+    Config('use_timestamp', 'true', r'''
+        by default, create the close checkpoint as of the last stable timestamp
+        if timestamps are in use, or all current updates if there is no
+        stable timestamp set.  If false, this option generates a checkpoint
+        with all updates including those later than the timestamp''',
```
bvpvamsikrishna: In the sentence, "If false", may be better, not to reference timestamp, like "If false, this option generates a checkpoint with all updates"

test/suite/test_timestamp12.py
sueloverso: Fixed.

dist/api_data.py
sueloverso: Fixed. Thanks.

sueloverso:
@agorrod this is ready for merge I believe. I addressed the comments from @ddanderson and @bvpvamsikrishna.

sueloverso:
Merging since Alex is on vacation and it has been reviewed by two others.

PR #3941, opened 2018-02-21
WT-3927 LSM truncate operations are too slow.

keithbostic:


agorrod:
Thanks @keithbostic this seems like the right approach to me.

PR #3943, opened 2018-02-22
WT-3932 Move hash tables out of session handles.

michaelcahill:
In 043f61e, we embedded hash tables for cursors and data handles directly in the WT_SESSION_IMPL structure.  That is problematic because we allocate all configured session handles in a single block of memory on startup, and the hash tables grew the structre from just over 1KB to ~18KB.


agorrod:
@michael.cahill you didn't ask for my review here, but the change lgtm. I'm happy for it to be merged and to follow up with a different solution if necessary, since this issue means we can't bump master at the moment.

PR #3944, opened 2018-02-22
WT-3933 test/format failure illegal WT_REF.state rolling back deleted page

keithbostic:
@michaelcahill, setting `WT_REF.page_del.txnid` to `WT_TXN_ABORTED` allows the page to be evicted and the `WT_REF.page_del` structure to be discarded.

So, making that change before marking the updates are aborted means we can drop core accessing the updates, making the change after we swap in the WT_REF's previous state means we can drop core getting to the `WT_REF.page_del.txnid` field.

### DIFF
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -125,6 +125,7 @@ __wt_delete_page(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
 	 * this point on.
 	 */
 	if (ref->page_del != NULL) {
+		WT_ASSERT(session, ref->page_del->txnid == WT_TXN_ABORTED);
 		__wt_free(session, ref->page_del->update_list);
 		__wt_free(session, ref->page_del);
 	}
@@ -201,7 +202,7 @@ __wt_delete_page_rollback(WT_SESSION_IMPL *session, WT_REF *ref)
 			if (!__wt_atomic_casv32(&ref->state,
 			    WT_REF_DELETED, ref->page_del->previous_state))
 				break;
-			return (0);
+			goto done;
 		case WT_REF_LOCKED:
 			/*
 			 * A possible state, the page is being instantiated.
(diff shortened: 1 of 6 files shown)


michaelcahill:
@keithbostic, I'm not following this part:

> making the change after we swap in the WT_REF's previous state means we can drop core getting to the WT_REF.page_del.txnid field

It shouldn't be possible to evict the page at that point because the truncating transaction is still running, so

michaelcahill:
Apologies for the clicking the wrong button.

keithbostic:
@keithbostic, I'm not following this part:

>> making the change after we swap in the WT_REF's previous state means we can drop core getting to the WT_REF.page_del.txnid field

> It shouldn't be possible to evict the page at that point because the truncating transaction is still running, so the `visible_all` check on `page_del` will fail.  Maybe some other operation could clobber the `page_del`, but that seems scary since the truncating transaction is still running.

The test in `__wt_page_can_evict()` checks `WT_TXN_ABORTED`, not the visibility of `ref.page_del.txnid`, which I think makes this problematical, that is, we've made the page evictable before the transaction has been fully rolled back.

If it's the last thing the routine does, well maybe...

michaelcahill:
@keithbostic, I'm talking about this part of the condition:
```
 	    !__wt_txn_visible_all(session,
 	    ref->page_del->txnid, WT_TIMESTAMP_NULL(&ref->page_del->timestamp)))
```
That can't be true while `ref->page_del->txnid` is the running transaction's ID.

michaelcahill:
@keithbostic, sorry we didn't overlap today -- I've pushed 50308f2 with the change I'd prefer, but haven't had a chance to kick off tests on zSeries to see whether this reintroduces the bug.

keithbostic:
@michaelcahill, your change lgtm, I'll merge into develop so it runs over the weekend.

PR #3945, opened 2018-02-24
WT-3911 Ignore lookaside data on shutdown

keithbostic:
@michaelcahill, I'm not terribly happy with this, but it's the simplest solution I found.

Everything else I tried (mostly, trying to find a cleaner restructuring of `conn_dhandle.c` functions), got ugly fast, so I ended up looking at the `closing` flag at a relatively low level.

### DIFF
--- a/src/btree/bt_discard.c
+++ b/src/btree/bt_discard.c
@@ -76,16 +76,8 @@ __wt_page_out(WT_SESSION_IMPL *session, WT_PAGE **pagep)
 	page = *pagep;
 	*pagep = NULL;
 
-	/*
-	 * Unless we're discarding a handle or closing the database, we should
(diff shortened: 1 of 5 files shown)


keithbostic:
@michaelcahill, 6ed66e1 fixes the build error, I believe (we can now discard dirty pages when closing the database), and b267a4f is a more aggressive version of that change, it generally asserts we should only discard dirty pages in the `__wt_evict_file()` function with `WT_SYNC_DISCARD` set.

I think b267a4f is correct, but if not we can revert it to 6ed66e1.

michaelcahill:
Thanks @keithbostic -- I like the more aggressive approach of asserting that we only discard dirty pages when discarding a whole tree.

lgtm -- I'll merge.

PR #3946, opened 2018-02-25
WT-3935 WT-3919 Enable cursor caching by default, add additional tests

ddanderson:
Enabling this exposed a bug which is also fixed: cursors with owners should not be eligible for caching.
Modified cursor cache test to expect caching by default, and removed
some scenarios that use complex datasets, they won't work.

### DIFF
--- a/src/cursor/cur_file.c
+++ b/src/cursor/cur_file.c
@@ -465,7 +465,7 @@ __curfile_close(WT_CURSOR *cursor)
 	WT_CURSOR_BULK *cbulk;
 	WT_DECL_RET;
 	WT_SESSION_IMPL *session;
-	bool released;
+	bool dead, released;
 
 	cbt = (WT_CURSOR_BTREE *)cursor;
 	CURSOR_API_CALL_PREPARE_ALLOWED(cursor, session, close, cbt->btree);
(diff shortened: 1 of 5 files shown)


ddanderson:
Not ready for review or merge, at least until the PR checks succeed.

ddanderson:
@agorrod, this is ready for review and merge, I don't know if you want to review it or pass to @sueloverso or @michaelcahill.

As WT-3935 describes, cursor caching wasn't really enabled by default. Changing it to the default exposed a few problems that are fixed here:

- "overwrite" option for cursors not handled correctly (since `overwrite=true` is default).
- `*_API_CALL` macros used without always matching `API_END`. When I'm tweaking performance, I might want to do this differently, but this is correct.
- When a data handle is dropped, and the cached cursor for that data handle closes, there is a final unlock of the data handle when there shouldn't be.
- Cursors with owners shouldn't be eligible for caching (at least not in this limited version of cursor caching). If they are, then I'm seeing a crash on close.  I think what was happening was that caching/reopening moves the the cursor to and from the open cursor list, and the nice ordering of owners vs ownees in the list was getting messed up. This might have been what was happening in the tests for sweep.  They are running now, so are part of this commit also.

michaelcahill:
Thanks @ddanderson, I've taken a look and this makes sense to me.  I'll merge so it gets testing in `develop`.  I'll follow up separately to make sure we have a plan for updating MongoDB 3.6 (to keep WT cursor caching disabled, at least for now).

PR #3948, opened 2018-02-26
WT-3934 LSM chunk checkpoints can race with system checkpoints

keithbostic:


michaelcahill:
Thanks @keithbostic, lgtm (I made a minor naming change) -- I'll merge.

PR #3950, opened 2018-02-26
WT-3913 Enhance cursor operations to account for prepare : WIP

bvpvamsikrishna:
Work in progress.
Modify to return WT_ROLLBACK error for updated operations when encounter a prepared updated.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -532,6 +532,7 @@ __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 			do {
 				__wt_yield();
 			} while (FLD_ISSET(upd->state, WT_UPDATE_STATE_LOCKED));
+                        /* FALLTHROUGH */
 		case WT_UPDATE_STATE_READY:
 			return (__wt_txn_visible(session,
 			    upd->txnid, WT_TIMESTAMP_NULL(&upd->timestamp)));
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -514,8 +514,32 @@ __wt_txn_visible(
 static inline bool
 __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 {
-	return (__wt_txn_visible(session,
-	    upd->txnid, WT_TIMESTAMP_NULL(&upd->timestamp)));
+	/*
+	 * If update state is ready, it indicates either prepared update
+	 * has not started or committed.
+	 */
+	if (FLD_ISSET(upd->state, WT_UPDATE_STATE_READY))
+		return (__wt_txn_visible(session,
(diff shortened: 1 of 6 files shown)


src/include/txn.i
```diff
@@ -480,31 +485,59 @@ __txn_visible_id(WT_SESSION_IMPL *session, uint64_t id)
  * __wt_txn_visible --
  *	Can the current transaction see the given ID / timestamp?
  */
-static inline bool
+static inline int
 __wt_txn_visible(
-    WT_SESSION_IMPL *session, uint64_t id, const wt_timestamp_t *timestamp)
+    WT_SESSION_IMPL *session, uint64_t id, const wt_timestamp_t *timestamp,
+    WT_UPDATE *upd)
```
michaelcahill: It seems wrong to me that we now have `__wt_txn_visible` taking an update.  I expected any logic that required an update to stay in `__wt_txn_upd_visible` and this function just deals with the visibility of a transaction ID.

src/include/txn.i
```diff
@@ -480,31 +485,59 @@ __txn_visible_id(WT_SESSION_IMPL *session, uint64_t id)
...
+				__wt_yield();
+			} while (FLD_ISSET(upd->state, WT_UPDATE_STATE_LOCKED));
+			/* FALLTHROUGH */
+		case WT_UPDATE_STATE_READY:
+			return (!F_ISSET(txn, WT_TXN_HAS_TS_READ) ?
+			    WT_VISIBLE_TYPE_TRUE :
+			    (__wt_timestamp_cmp(timestamp,
+				&txn->read_timestamp) <= 0 ?
+				WT_VISIBLE_TYPE_TRUE : WT_VISIBLE_TYPE_FALSE));
+		}
```
michaelcahill: I think this logic is basically correct, just not where I expected it.

src/include/txn.i
```diff
@@ -514,8 +547,21 @@ __wt_txn_visible(
 static inline bool
 __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 {
-	return (__wt_txn_visible(session,
-	    upd->txnid, WT_TIMESTAMP_NULL(&upd->timestamp)));
+	WT_VISIBLE_TYPE visible_type;
+
+	visible_type = __wt_txn_visible(session, upd->txnid,
+	    WT_TIMESTAMP_NULL(&upd->timestamp), upd);
```
michaelcahill: What I expected was for `__wt_txn_visible` to continue a boolean.

Then wait if `upd->state == LOCKED`.  Then save `upd->state`, call `__wt_txn_visible` and make sure the state doesn't change (loop if it does).

bvpvamsikrishna:
Will raise a new pull request for this ticket.

PR #3954, opened 2018-02-28
WT-3913 Enhance cursor operations to account for prepare

bvpvamsikrishna:
Read operations to return prepare conflict is still pending.


bvpvamsikrishna:
@michaelcahill, Request for a quick review of this pull request.
This contains the changes for update operation and for read operation as well for "ignore_prepare=true" case only.
Now I am making changes for read operation to return error for cases of "ignore_prepare=false". 
I am changing the prototype of __wt_txn_read to return an int so that error can be returned. 
I will inform you again, after I finish modifications. 
currently in test_prepare04.py, I have commented the scenario's to be tested with ignore_prepare=false cases.

src/include/txn.i
```diff
@@ -511,11 +517,38 @@ __wt_txn_visible(
  * __wt_txn_upd_visible --
  *	Can the current transaction see the given update.
  */
-static inline bool
+static inline int
```
michaelcahill: Shouldn't the return be a `WT_VISIBLE_TYPE`?

src/include/txn.i
```diff
@@ -9,6 +9,12 @@
 static inline int __wt_txn_id_check(WT_SESSION_IMPL *session);
 static inline void __wt_txn_read_last(WT_SESSION_IMPL *session);
 
+typedef enum {
+	WT_VISIBLE_TYPE_FALSE=0,        /* Not a visible update */
+	WT_VISIBLE_TYPE_PREPARE=1,      /* Prepared update */
+	WT_VISIBLE_TYPE_TRUE=2          /* A visible update */
+} WT_VISIBLE_TYPE;
```
michaelcahill: I'd change the constant names to `WT_VISIBLE_FALSE`, etc. (remove `_TYPE`) to keep them a little shorter.

src/include/txn.i
```diff
@@ -755,10 +788,23 @@ static inline int
...
+	 */
+	if (ignore_prepare_set)
+		F_CLR(txn, WT_TXN_IGNORE_PREPARE);
+
 	if (txn->isolation == WT_ISO_SNAPSHOT)
-		while (upd != NULL && !__wt_txn_upd_visible(session, upd)) {
+		while (upd != NULL && __wt_txn_upd_visible(session, upd) !=
+		    WT_VISIBLE_TYPE_TRUE) {
+			if (ignore_prepare_set)
+				F_SET(txn, WT_TXN_IGNORE_PREPARE);
```
michaelcahill: I don't think we should re-enable the flag until just before the two `return` lines.  As written, we'd have the flag set after the first update.  That should never happen, but it's hard to reason about.

bvpvamsikrishna:
@michaelcahill, Request your review for this pull request.

bvpvamsikrishna:
@agorrod, Request your review for this pull request.

src/include/txn.i
```diff
@@ -511,30 +519,63 @@ __wt_txn_visible(
...
+		}
+
+		upd_visible = __wt_txn_visible(session, upd->txnid,
+		    WT_TIMESTAMP_NULL(&upd->timestamp));
+		/*
+		 * Update state changed from beneath during txn visibility
+		 * check, try again.
+		 */
+		if (upd->state != upd_state)
+			continue;
```
sulabhM: the state can change after the txn visibility check, before line 542, and in that case we will continue. Also the state can change after we have done this check. Can you explain further in the comment, why this is okay.

src/include/txn.i
```diff
@@ -9,6 +9,14 @@
 static inline int __wt_txn_id_check(WT_SESSION_IMPL *session);
 static inline void __wt_txn_read_last(WT_SESSION_IMPL *session);
 
+#if 0
```
sulabhM: can be removed, since you moved this to another file.

agorrod:
I suggested some changes face to face with @bvpvamsikrishna it is worth holding off on further review until those changes are done.

src/btree/bt_curnext.c
```diff
@@ -101,7 +102,14 @@ __cursor_fix_next(WT_CURSOR_BTREE *cbt, bool newpage)
...
 		cbt->ins = NULL;
-	upd = cbt->ins == NULL ? NULL : __wt_txn_read(session, cbt->ins->upd);
+
+	upd = NULL;			/* -Werror=maybe-uninitialized */
+	if (cbt->ins != NULL) {		/* -Werror=dangling-else */
+		if (__wt_txn_read(session, cbt->ins->upd, &upd) ==
+		    WT_VISIBLE_PREPARE)
+			return (WT_PREPARE_CONFLICT);
+	} else
+		upd = NULL;
```
sulabhM: Since we are already initialising upd to NULL, we don't need else statement here.
Same comment for the other instances of this below.

src/btree/bt_curnext.c
```diff
@@ -671,7 +724,14 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
-err:	if (ret != 0)
+	/*
+	 * If prepare conflict occurs, cursor should not be reset, as current
+	 * cursor position will be reused in case of a retry from user.
+	 */
+err:	if (ret == WT_PREPARE_CONFLICT)
+		F_SET(cbt, WT_CBT_RETRY_NEXT);
+	else if (ret != 0)
 		WT_TRET(__cursor_reset(cbt));
+	/* Need to check whether cursor state to be restored, after reset. */
```
sulabhM: Does this comment specify a work item that needs further investigation ?

src/btree/bt_curprev.c
```diff
@@ -247,7 +251,13 @@ __cursor_fix_prev(WT_CURSOR_BTREE *cbt, bool newpage)
 	    cbt->ins_head, cbt->ins_stack, cbt->next_stack, cbt->recno);
 	if (cbt->ins != NULL && cbt->recno != WT_INSERT_RECNO(cbt->ins))
 		cbt->ins = NULL;
-	upd = cbt->ins == NULL ? NULL : __wt_txn_read(session, cbt->ins->upd);
+	upd = NULL;			/* _Werror=maybe-uninitialized */
+	if (cbt->ins != NULL) {		/* -Werror=dangling-else */
+		if (__wt_txn_read(session, cbt->ins->upd, &upd) ==
+		    WT_VISIBLE_PREPARE)
+			return (WT_PREPARE_CONFLICT);
+	} else
```
sulabhM: Same as before, else not needed.

src/btree/bt_curprev.c
```diff
@@ -340,8 +353,13 @@ new_page:	if (cbt->recno < cbt->ref->ref_recno)
...
 		cbt->ins = __col_insert_search_match(cbt->ins_head, cbt->recno);
-		upd = cbt->ins == NULL ?
-		    NULL : __wt_txn_read(session, cbt->ins->upd);
+		upd = NULL;		/* _Werror=maybe-uninitialized */
+		if (cbt->ins != NULL) {		/* -Werror=dangling-else */
+			if (__wt_txn_read(session, cbt->ins->upd, &upd) ==
+			    WT_VISIBLE_PREPARE)
+				return (WT_PREPARE_CONFLICT);
+		} else
+			upd = NULL;
```
sulabhM: Same as before, else not needed.

src/btree/bt_curprev.c
```diff
@@ -625,7 +680,14 @@ __wt_btcur_prev(WT_CURSOR_BTREE *cbt, bool truncating)
...
-err:	if (ret != 0)
+	/*
+	 * If prepare conflict occurs, cursor should not be reset, as current
+	 * cursor position will be reused in case of a retry from user.
+	 */
+err:	if (ret == WT_PREPARE_CONFLICT)
+		F_SET(cbt, WT_CBT_RETRY_PREV);
+	else if (ret != 0)
 		WT_TRET(__cursor_reset(cbt));
+	/* Need to check whether cursor state to be restored, after reset. */
```
sulabhM: Does this comment say that this needs further looking into?

src/btree/bt_cursor.c
```diff
@@ -512,15 +507,27 @@ __wt_btcur_search(WT_CURSOR_BTREE *cbt)
...
 		    __cursor_row_search(session, cbt, cbt->ref, false) :
 		    __cursor_col_search(session, cbt, cbt->ref));
-		valid = cbt->compare == 0 && __wt_cursor_valid(cbt, &upd);
+
+		/* If prepared update, return prepare conflict. */
+		visibility = __wt_cursor_valid(cbt, &upd);
+		if (visibility == WT_VISIBLE_PREPARE)
+			WT_ERR(WT_PREPARE_CONFLICT);
+		valid = cbt->compare == 0 && (visibility == WT_VISIBLE_TRUE ?
+		    true : false);
```
sulabhM: Can this be:
`valid = cbt->compare == 0 && visibility == WT_VISIBLE_TRUE;`

src/btree/bt_cursor.c
```diff
@@ -512,15 +507,27 @@ __wt_btcur_search(WT_CURSOR_BTREE *cbt)
...
 		    __cursor_row_search(session, cbt, NULL, false) :
 		    __cursor_col_search(session, cbt, NULL));
-		valid = cbt->compare == 0 && __wt_cursor_valid(cbt, &upd);
+
+		/* If prepared update, return prepare conflict. */
+		visibility = __wt_cursor_valid(cbt, &upd);
+		if (visibility == WT_VISIBLE_PREPARE)
+			WT_ERR(WT_PREPARE_CONFLICT);
+		valid = cbt->compare == 0 && (visibility == WT_VISIBLE_TRUE ?
+		    true : false);
```
sulabhM: Can this be:
`valid = cbt->compare == 0 && visibility == WT_VISIBLE_TRUE;`

src/btree/bt_cursor.c
```diff
@@ -617,15 +625,25 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
...
 		 */
-		if (cbt->slot != 0 && cbt->slot != cbt->ref->page->entries - 1)
-			valid = __wt_cursor_valid(cbt, &upd);
+		if (cbt->slot != 0 &&
+		    cbt->slot != cbt->ref->page->entries - 1) {
+			visibility = __wt_cursor_valid(cbt, &upd);
+			if (visibility == WT_VISIBLE_PREPARE)
+				WT_ERR(WT_PREPARE_CONFLICT);
+
+			valid = (visibility == WT_VISIBLE_TRUE ? true : false);
```
sulabhM: This can be:
`valid = visibility == WT_VISIBLE_TRUE;`

src/btree/bt_cursor.c
```diff
@@ -617,15 +625,25 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
...
 		WT_ERR(__cursor_func_init(cbt, true));
 		WT_ERR(btree->type == BTREE_ROW ?
 		    __cursor_row_search(session, cbt, NULL, true) :
 		    __cursor_col_search(session, cbt, NULL));
-		valid = __wt_cursor_valid(cbt, &upd);
+		visibility = __wt_cursor_valid(cbt, &upd);
+		if (visibility == WT_VISIBLE_PREPARE)
+			WT_ERR(WT_PREPARE_CONFLICT);
+
+		valid = (visibility == WT_VISIBLE_TRUE ? true : false);
```
sulabhM: This can be:
`valid = visibility == WT_VISIBLE_TRUE;`

src/btree/bt_cursor.c
```diff
@@ -784,8 +814,17 @@ retry:	WT_ERR(__cursor_func_init(cbt, true));
 		 * key/value pair.
 		 */
 		if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE) &&
-		    cbt->compare == 0 && __wt_cursor_valid(cbt, NULL))
-			WT_ERR(WT_DUPLICATE_KEY);
+		    cbt->compare == 0) {
+			visibility = __wt_cursor_valid(cbt, NULL);
+			if (visibility == WT_VISIBLE_TRUE)
+				WT_ERR(WT_DUPLICATE_KEY);
```
sulabhM: These lines can be joined, but I am not fussy about it:

```
if ((visibility = __wt_cursor_valid(cbt, NULL)) == WT_VISIBLE_TRUE)
        WT_ERR(WT_DUPLICATE_KEY);
```

src/btree/bt_cursor.c
```diff
@@ -784,8 +814,17 @@ retry:	WT_ERR(__cursor_func_init(cbt, true));
...
+			/*
+			 * Do we have to return WT_ROLLBACK here ??,
```
sulabhM: I guess this is a work item that needs further looking into.

src/btree/bt_cursor.c
```diff
@@ -805,10 +844,18 @@ retry:	WT_ERR(__cursor_func_init(cbt, true));
 		 * column-store implicitly fills the gap with empty records.
 		 * Fail in that case, the record exists.
 		 */
-		if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE) &&
-		    ((cbt->compare == 0 && __wt_cursor_valid(cbt, NULL)) ||
-		    (cbt->compare != 0 && __cursor_fix_implicit(btree, cbt))))
-			WT_ERR(WT_DUPLICATE_KEY);
+		if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE)) {
+			if (cbt->compare ==0 ) {
+				visibility = __wt_cursor_valid(cbt, NULL);
```
sulabhM: Same comments as the previous block. can combine assignment and comparison inside if block. Also needs looking into if rollback has to be returned.

src/btree/bt_cursor.c
```diff
@@ -1028,8 +1076,13 @@ retry:	if (positioned == POSITIONED)
 		/* Check whether an update would conflict. */
 		WT_ERR(__curfile_update_check(cbt));
 
-		if (cbt->compare != 0 || !__wt_cursor_valid(cbt, NULL))
+		if (cbt->compare != 0)
```
sulabhM: Same comments as earlier block.

src/btree/bt_cursor.c
```diff
@@ -1042,8 +1095,11 @@ retry:	if (positioned == POSITIONED)
 		 */
 		WT_ERR(__curfile_update_check(cbt));
 
+		visibility = __wt_cursor_valid(cbt, NULL);
+		if (visibility == WT_VISIBLE_PREPARE)
+			WT_ERR(WT_PREPARE_CONFLICT);	/* WT_ROLLBACK ?? */
```
sulabhM: Same as before, assignment and comparison can be combined.

src/btree/bt_cursor.c
```diff
@@ -1207,8 +1264,13 @@ retry:	WT_ERR(__cursor_func_init(cbt, true));
 		 */
 		if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE)) {
 			WT_ERR(__curfile_update_check(cbt));
-			if (cbt->compare != 0 || !__wt_cursor_valid(cbt, NULL))
+			if (cbt->compare != 0)
 				WT_ERR(WT_NOTFOUND);
+			visibility = __wt_cursor_valid(cbt, NULL);
```
sulabhM: Same comments as previous block.

src/btree/bt_cursor.c
```diff
@@ -1224,10 +1286,17 @@ retry:	WT_ERR(__cursor_func_init(cbt, true));
 		 */
 		if (!F_ISSET(cursor, WT_CURSTD_OVERWRITE)) {
 			WT_ERR(__curfile_update_check(cbt));
+			/*
```
sulabhM: An item that needs further looking into.

src/btree/bt_random.c
```diff
@@ -421,16 +422,19 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	 * the next entry, if that doesn't work, move to the previous entry.
 	 */
 	WT_ERR(__wt_row_random_leaf(session, cbt));
-	if (__wt_cursor_valid(cbt, &upd)) {
+	visibility = __wt_cursor_valid(cbt, &upd);
+	if (visibility == WT_VISIBLE_TRUE) {
```
sulabhM: Can combine assignment and comparison inside if block.

bvpvamsikrishna:
Fixed the review comments.
@sulabhM, request to check.

src/btree/bt_curnext.c
```diff
@@ -101,7 +105,13 @@ __cursor_fix_next(WT_CURSOR_BTREE *cbt, bool newpage)
 	    cbt->ins_head, cbt->ins_stack, cbt->next_stack, cbt->recno);
 	if (cbt->ins != NULL && cbt->recno != WT_INSERT_RECNO(cbt->ins))
 		cbt->ins = NULL;
-	upd = cbt->ins == NULL ? NULL : __wt_txn_read(session, cbt->ins->upd);
+
+	upd = NULL;			/* -Werror=maybe-uninitialized */
```
agorrod: This initialization should go in the in general initialization block below the declaration block (unless it's in a loop).

src/btree/bt_curnext.c
```diff
@@ -193,8 +208,13 @@ __cursor_var_next(WT_CURSOR_BTREE *cbt, bool newpage)
 		/* Check any insert list for a matching record. */
 		cbt->ins_head = WT_COL_UPDATE_SLOT(page, cbt->slot);
 		cbt->ins = __col_insert_search_match(cbt->ins_head, cbt->recno);
-		upd = cbt->ins == NULL ?
-		    NULL : __wt_txn_read(session, cbt->ins->upd);
+		upd = NULL;		/* -Werror=maybe-uninitialized */
+		if (cbt->ins != NULL) {		/* -Werror=dangling-else */
```
agorrod: Don't need the comment.

src/btree/bt_curnext.c
```diff
@@ -582,6 +610,27 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
+		upd = NULL; 		/* -Werror=maybe-uninitialized */
+		switch (__wt_cursor_valid(cbt, &upd)) {
+		case WT_VISIBLE_FALSE:
+			break;
+		case WT_VISIBLE_PREPARE:
+			return (WT_PREPARE_CONFLICT);
+		case WT_VISIBLE_TRUE:
+			F_CLR(cbt, WT_CBT_RETRY_NEXT);
+			WT_RET(__cursor_kv_return(session, cbt, upd));
+			return (0);
```
agorrod: These two lines can be merged, i.e: `return (__cursor_kv_return...)`

src/btree/bt_curnext.c
```diff
@@ -606,10 +655,14 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 		if (F_ISSET(cbt, WT_CBT_ITERATE_APPEND)) {
 			switch (page->type) {
 			case WT_PAGE_COL_FIX:
-				ret = __cursor_fix_append_next(cbt, newpage);
+				if ((ret = __cursor_fix_append_next(cbt,
+				    newpage)) == WT_PREPARE_CONFLICT)
```
agorrod: This line wrapping looks a bit wrong. Try moving the `cbt` to the second line as well.

src/btree/bt_curnext.c
```diff
@@ -54,8 +55,10 @@ __cursor_fix_append_next(WT_CURSOR_BTREE *cbt, bool newpage)
 	 * insert is aborted, we simply return zero (empty), regardless of
 	 * whether we are at the end of the data.
 	 */
-	if (cbt->recno < WT_INSERT_RECNO(cbt->ins) ||
-	    (upd = __wt_txn_read(session, cbt->ins->upd)) == NULL) {
+	__wt_txn_read(session, cbt->ins->upd, &visibility, &upd);
+	if (visibility == WT_VISIBLE_PREPARE)
+		return (WT_PREPARE_CONFLICT);
+	if (cbt->recno < WT_INSERT_RECNO(cbt->ins) || upd == NULL) {
```
agorrod: Think about whether switching the order of the comparison and visibility check matters for correctness or performance.

src/btree/bt_curprev.c
```diff
@@ -570,34 +600,39 @@ __wt_btcur_prev(WT_CURSOR_BTREE *cbt, bool truncating)
...
-				ret = __cursor_var_append_prev(cbt, newpage);
+				WT_ERR_IF(
+				    __cursor_var_append_prev( cbt, newpage),
+				    ret != WT_NOTFOUND);
 				break;
 			WT_ILLEGAL_VALUE_ERR(session);
 			}
 			if (ret == 0)
 				break;
 			F_CLR(cbt, WT_CBT_ITERATE_APPEND);
```
agorrod: I wonder if this `F_CLR` needs to be added to the error handling code as well?

src/include/btmem.h
```diff
@@ -997,7 +997,7 @@ struct __wt_update {
 						   finalized prepare */
 #define	WT_UPDATE_STATE_LOCKED		1	/* locked */
 #define	WT_UPDATE_STATE_PREPARED	2	/* prepared */
-	uint8_t state;			/* state (one byte : conserve memory) */
+	volatile uint8_t state;
```
agorrod: I don't see how this change fits in - why is it required?

src/include/txn.i
```diff
@@ -542,30 +547,64 @@ __wt_txn_visible(
  * __wt_txn_upd_visible --
  *	Can the current transaction see the given update.
  */
-static inline bool
+static inline WT_VISIBLE_TYPE
 __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
```
agorrod: There are callers of this function that don't seem to have changed - (`rec_write.c` and `bt_ret.c`). That surprises me.

src/include/txn.i
```diff
@@ -542,30 +547,64 @@ __wt_txn_visible(
...
+		if (upd_state == WT_UPDATE_STATE_LOCKED) {
+			/* Commit is in progress, yield and try again. */
+			__wt_yield();
+			continue;
+		}
+
+		upd_visible = __wt_txn_visible(session, upd->txnid,
+		    WT_TIMESTAMP_NULL(&upd->timestamp));
+		/*
+		 * Update should not change during txn visibility check.
```
agorrod: Line wrapping in this comment doesn't look quite right.

src/include/btmem.h
michaelcahill: This was done in response to a comment I made, and Keith said something similar (somewhere I can't find right now).

That is, the locking protocol requires that `state` is switched to `WT_UPDATE_STATE_LOCKED` before overwriting the timestamp as part of committing a prepared transaction.  Readers must see the `state` change so they can be sure they have a valid timestamp.  Making the field volatile is necessary but not sufficient on CPUs with weak memory models.

src/btree/bt_cursor.c
```diff
@@ -1057,9 +1071,16 @@ retry:	if (positioned == POSITIONED)
...
+		} else {
+			WT_ERR(__wt_cursor_valid(cbt, NULL, &valid));
+			if (!valid) {
+				if (!__cursor_fix_implicit(btree, cbt))
+					WT_ERR(WT_NOTFOUND);
+				cbt->recno = cursor->recno;
+			} else
+				ret = __cursor_col_modify(
+				    session, cbt, WT_UPDATE_TOMBSTONE);
+		}
```
michaelcahill: I think it's unfortunate that we have to repeat the `__cursor_fix_implicit` here.

Did you consider calling `__wt_cursor_valid` before checking `cbt->compare != 0`?  Or would that generate a prepare conflict in some case where `WT_NOTFOUND` is expected?

src/include/txn.i
```diff
@@ -538,34 +543,77 @@ __wt_txn_visible(
...
  * __wt_txn_upd_visible --
  *	Can the current transaction see the given update.
  */
 static inline bool
 __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 {
-	return (__wt_txn_visible(session,
-	    upd->txnid, WT_TIMESTAMP_NULL(&upd->timestamp)));
+	return (__wt_txn_upd_visible_type(session, upd) == WT_VISIBLE_TRUE);
 }
```
michaelcahill: I like the way this has turned out: neat!

src/include/cursor.i
```diff
@@ -190,6 +190,7 @@ __cursor_reset(WT_CURSOR_BTREE *cbt)
 		if (!F_ISSET(cbt, WT_CBT_NO_TXN))
 			__cursor_leave(session);
 		F_CLR(cbt, WT_CBT_ACTIVE);
+		F_CLR(cbt, WT_CBT_RETRY_NEXT | WT_CBT_RETRY_PREV);
```
keithbostic: I think a simpler/cleaner approach to this would be to add `WT_CBT_RETRY_NEXT | WT_CBT_RETRY_PREV` to `WT_CBT_POSITION_MASK`.

src/include/cursor.i
```diff
@@ -321,6 +336,8 @@ __cursor_func_init(WT_CURSOR_BTREE *cbt, bool reenter)
 
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
 
+	F_CLR(cbt, WT_CBT_RETRY_NEXT | WT_CBT_RETRY_PREV);
+
```
keithbostic: I think a simpler/cleaner approach to this would be to clear `WT_CBT_RETRY_NEXT` in `__wt_btcur_next` and `WT_CBT_RETRY_PREV` in `__wt_btcur_prev`. Then `__cursor_func_init()` doesn't do anything at all with those flags, (either testing or clearing). Something like:
```
index 82096bb3c..25986779f 100644
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -610,6 +610,7 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
                        F_CLR(cbt, WT_CBT_RETRY_NEXT);
                        return (__cursor_kv_return(session, cbt, upd));
                }
+               F_CLR(cbt, WT_CBT_RETRY_NEXT);
        }
 
        WT_RET(__cursor_func_init(cbt, false));
diff --git a/src/btree/bt_curprev.c b/src/btree/bt_curprev.c
index 7e5783a8e..aa2b8918b 100644
--- a/src/btree/bt_curprev.c
+++ b/src/btree/bt_curprev.c
@@ -565,6 +565,7 @@ __wt_btcur_prev(WT_CURSOR_BTREE *cbt, bool truncating)
                        F_CLR(cbt, WT_CBT_RETRY_PREV);
                        return (__cursor_kv_return(session, cbt, upd));
                }
+               F_CLR(cbt, WT_CBT_RETRY_PREV);
        }
 
        WT_RET(__cursor_func_init(cbt, false));
diff --git a/src/include/cursor.i b/src/include/cursor.i
index 4e4430e4c..a8d739514 100644
--- a/src/include/cursor.i
+++ b/src/include/cursor.i
@@ -336,8 +336,6 @@ __cursor_func_init(WT_CURSOR_BTREE *cbt, bool reenter)
 
-       F_CLR(cbt, WT_CBT_RETRY_NEXT | WT_CBT_RETRY_PREV);
-
        if (reenter) {
 #ifdef HAVE_DIAGNOSTIC
                __wt_cursor_key_order_reset(cbt);
```

src/include/txn.i
```diff
@@ -538,34 +543,77 @@ __wt_txn_visible(
...
+
+		upd_visible = __wt_txn_visible(
+		    session, upd->txnid, WT_TIMESTAMP_NULL(&upd->timestamp));
+
+		/*
+		 * The visibility check is only valid if the update does not
+		 * change state.  If the state does change, recheck visibility.
+		 */
+		if (upd->state != upd_state)
+			continue;
```
keithbostic: It seems odd we yield in the commit case, but not in the `update-changing` case. Why are they different?

src/include/txn.i
```diff
@@ -538,34 +543,77 @@ __wt_txn_visible(
...
+		if (!upd_visible)
+			return (WT_VISIBLE_FALSE);
+
+		if (upd_state == WT_UPDATE_STATE_PREPARED)
+			return (F_ISSET(&session->txn, WT_TXN_IGNORE_PREPARE) ?
+			    WT_VISIBLE_FALSE : WT_VISIBLE_PREPARE);
+
+		return (WT_VISIBLE_TRUE);
+	}
+}
```
keithbostic: I'd recommend a `/* NOTREACHED */` after the for loop, it makes some static analyzers happier.

src/include/txn.i
keithbostic: I agree with @michaelcahill, this turned out quite nicely, good work!

src/include/txn.i
```diff
@@ -786,11 +835,22 @@ static inline int
...
 
 	txn = &session->txn;
+	ignore_prepare_set = F_ISSET(txn, WT_TXN_IGNORE_PREPARE);
+	/*
+	 * Clear the ignore prepare setting of txn, as it is not suppose to
+	 * affect the visibility for update operation
+	 */
+	if (ignore_prepare_set)
+		F_CLR(txn, WT_TXN_IGNORE_PREPARE);
+
```
keithbostic: typo, `not suppose` -> `not supposed`, `operation` -> `operations.`

src/include/txn.i
```diff
@@ -786,11 +835,22 @@ static inline int
 __wt_txn_update_check(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 {
 	WT_TXN *txn;
+	uint32_t ignore_prepare_set;
```
keithbostic: `ignore_prepare_set` should be a boolean.

src/include/txn.i
```diff
@@ -786,11 +835,22 @@ static inline int
...
+		F_CLR(txn, WT_TXN_IGNORE_PREPARE);
+
 	if (txn->isolation == WT_ISO_SNAPSHOT)
 		while (upd != NULL && !__wt_txn_upd_visible(session, upd)) {
 			if (upd->txnid != WT_TXN_ABORTED) {
+				if (ignore_prepare_set)
+					F_SET(txn, WT_TXN_IGNORE_PREPARE);
 				WT_STAT_CONN_INCR(
 				    session, txn_update_conflict);
 				WT_STAT_DATA_INCR(
```
keithbostic: The `WT_STAT_CONN_INCR` and `WT_STAT_DATA_INCR` lines can be joined, they don't need to be split.

src/include/txn.i
```diff
@@ -786,11 +835,22 @@ static inline int
...
+	 * affect the visibility for update operation
+	 */
+	if (ignore_prepare_set)
+		F_CLR(txn, WT_TXN_IGNORE_PREPARE);
+
 	if (txn->isolation == WT_ISO_SNAPSHOT)
 		while (upd != NULL && !__wt_txn_upd_visible(session, upd)) {
 			if (upd->txnid != WT_TXN_ABORTED) {
+				if (ignore_prepare_set)
+					F_SET(txn, WT_TXN_IGNORE_PREPARE);
```
keithbostic: This function has gotten kind of messy, and we're doing work we don't have to do if the transaction isn't a `WT_ISO_SNAPSHOT` configuration. The compiler will probably remove that work, but I'm not sure, given as this is an inline function. Here's a suggested rewrite, discard it if you don't like it.
```
/*
 * __wt_txn_update_check --
 *      Check if the current transaction can update an item.
 */
static inline int
__wt_txn_update_check(WT_SESSION_IMPL *session, WT_UPDATE *upd)
{
        WT_TXN *txn;
        bool ignore_prepare_set;
         
        txn = &session->txn;
        if (txn->isolation != WT_ISO_SNAPSHOT)
                return (0);

        /*
         * Clear the ignore prepare setting of txn, as it is not supposed to
         * affect the visibility for update operations.
         */
        ignore_prepare_set = F_ISSET(txn, WT_TXN_IGNORE_PREPARE);
        F_CLR(txn, WT_TXN_IGNORE_PREPARE);
        for (;
            upd != NULL && !__wt_txn_upd_visible(session, upd);
            upd = upd->next)
                if (upd->txnid != WT_TXN_ABORTED)
                        break;
        if (ignore_prepare_set)
                F_SET(txn, WT_TXN_IGNORE_PREPARE);

        if (upd == NULL)
                return (0);

        WT_STAT_CONN_INCR(session, txn_update_conflict);
        WT_STAT_DATA_INCR(session, txn_update_conflict);
        return (__wt_txn_rollback_required(
            session, "conflict between concurrent operations"));
}
```

keithbostic: It might also be cleaner to pass a boolean `ignore_prepare` into `__wt_txn_upd_visible_type()` instead of checking `F_ISSET(txn, WT_TXN_IGNORE_PREPARE)`. If `__wt_txn_update_check()` called `__wt_txn_upd_visible_type()` directly, then it wouldn't have to  set/clear the `txn` flags at all.

bvpvamsikrishna: I haven't made this change, as this is resulting in adding WT_TXN_IGNORE_PREPARE to rec_write.c and bt_ret.c. 
If you feel strongly about this change, I will do it.

bvpvamsikrishna: This exact version is failing for test_prepare01.py, so I have modified some part. I still need to  figure out what is the problem with this version.

agorrod:
lgtm

PR #3957, opened 2018-02-28
WT-3946 Fix a truncate segfault resetting a NULL start cursor.

michaelcahill:


michaelcahill:
Merging: this is a comment-only change in `develop`, it's mostly interesting for backporting.

PR #3959, opened 2018-02-28
WT-3947 Added wiredtiger_open configuration 'cache_cursors' to disable cursor caching.

ddanderson:
Added and improved tests that check cursor caching.


ddanderson:
@sueloverso, could you review this change please?

michaelcahill:
I'd also be fine with not allowing `cache_cursors` to be reconfigured: it seems like something the application is either prepared to pay for (extra memory per open session for more performance), or it isn't.  I'm struggling to think of when an application would switch it on/off.

michaelcahill:
I reviewed, and we were only checking for `cache_cursors` during `wiredtiger_open`.  In other words, the code didn't allow the setting to be reconfigured.  I pushed a change with the update to `api_data.py` to reflect that.

This otherwise LGTM, so if you're happy, please merge.

ddanderson:
@michaelcahill and @sueloverso, thanks for your review and changes.  LGTM, I'll merge.

PR #3960, opened 2018-03-01
 WT-3935 Fix cursor pointers to btree memory after reopen

michaelcahill:


michaelcahill:
@ddanderson, I hope this change is clear -- the issue was detected by MongoDB tests that validate (i.e., call `WT_SESSION::verify`) between phases.

test/suite/test_cursor13.py
```diff
@@ -201,6 +201,20 @@ def test_reconfig(self):
...
+    # Test we can reopen across a verify.
+    def test_verify(self):
+        if self.dstype != None:
+            self.session.reconfigure('cache_cursors=true')
+            ds = self.dstype(self, self.uri, 100)
+            ds.populate()
+            for loop in range(10):
+                c = self.session.open_cursor(self.uri)
+                ds.check()
+                c.close()
```
michaelcahill: Interestingly, if I didn't open the cursor here, it seems like no cursor was cached or reopened by `ds.check()`.  Any idea why that would be?

ddanderson: It stumped me too, but I found the answer and added a comment.

PR #3961, opened 2018-03-01
WT-3936 Add prepare testing to timestamp_abort and format.

sueloverso:
Here are the changes to test prepare in both `test/format` and `timestamp_abort`. It required some slight gymnastics in `timestamp_abort` because of the assertion in the code that there are no logged records that are part of the transaction when we prepare. It also means we cannot test prepared transactions in `format` unless logging is turned off.

Please review.

### DIFF
--- a/test/format/ops.c
+++ b/test/format/ops.c
@@ -193,6 +193,7 @@ wts_ops(int lastrun)
 			tinfo = tinfo_list[i];
 			total.commit += tinfo->commit;
 			total.deadlock += tinfo->deadlock;
+			total.prepare += tinfo->prepare;
 			total.insert += tinfo->insert;
 			total.remove += tinfo->remove;
 			total.rollback += tinfo->rollback;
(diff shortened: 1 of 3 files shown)


test/format/ops.c
```diff
@@ -575,6 +576,32 @@ commit_transaction(TINFO *tinfo, WT_SESSION *session)
...
+	 * commit timestamp. Set the prepare timestamp to whatever
+	 * the global value is now. The later commit will increment
+	 * it at whatever value it is then.
+	 */
+	testutil_check(__wt_snprintf(
+	    config_buf, sizeof(config_buf),
+	    "prepare_timestamp=%" PRIx64, g.timestamp));
+	ret = session->prepare_transaction(session, config_buf);
+	++tinfo->prepare;
+	return (ret);
```
keithbostic: We don't really need `ret` here, and we should use `WT_DECL_RET` if we declare it. I'd suggest:
```
        ++tinfo->prepare;

        /*
         * Prepare timestamps must be less than or equal to the
         * commit timestamp. Set the prepare timestamp to whatever
         * the global value is now. The later commit will increment
         * it at whatever value it is then.
         */
        testutil_check(__wt_snprintf(config_buf, sizeof(config_buf),
            "prepare_timestamp=%" PRIx64, g.timestamp));
        return (session->prepare_transaction(session, config_buf));
```

test/format/ops.c
```diff
@@ -1027,6 +1054,16 @@ ops(void *arg)
...
+			/*
+			 * Occasionally prepare the transaction too.
+			 */
+			if (rnd == 1) {
+				ret = prepare_transaction(tinfo, session);
+				testutil_assert(
+				    ret == 0 || ret == WT_PREPARE_CONFLICT);
+				if (ret == WT_PREPARE_CONFLICT)
+					goto deadlock;
+			}
```
keithbostic: I'd suggest:
```
/*
 * If we're in a transaction, commit 40% of the time and
 * rollback 10% of the time.
 */
switch (rnd) {
case 1:                                         /* 10% */
        /*
         * Occasionally prepare the transaction, then commit.
         */
        ret = prepare_transaction(tinfo, session);
        testutil_assert(ret == 0 || ret == WT_PREPARE_CONFLICT);
        if (ret == WT_PREPARE_CONFLICT)
                goto deadlock;
        /* FALLTHROUGH */
case 2: case 3: case 4:                         /* 40% */
        commit_transaction(tinfo, session);
        break;
case 5:                                         /* 10% */
```
rather than testing `rnd` explicitly, but I don't feel strongly.

test/format/ops.c
sueloverso: Fixed

test/format/ops.c
sueloverso: Fixed.

PR #3962, opened 2018-03-02
WT-3949 WT_CURSOR::modify buffer overflow for string values.

michaelcahill:


michaelcahill:
@keithbostic, I don't think this is too serious because as far as I know we're only using WT_CURSOR::modify on string values in testing.  That said, it's a bug that we should fix...

PR #3965, opened 2018-03-03
WT-3952 page-delete update list traversed after it has been discarded.

keithbostic:
@michaelcahill, the changes in 2c697cd and c984903 are the only ones that should require careful review, the others are comment clarification and code shuffling to make it easier to make the fix.


michaelcahill:
Thanks @keithbostic, I noticed a few other things in the same area and pushed fixes.  I'll merge.

FTR, this doesn't cover all of WT-3922 -- I'll add notes there with the remaining work.

PR #3966, opened 2018-03-03
WT-3954 prepared operations evicted before commit

keithbostic:
@bvpvamsikrishna, @michaelcahill, for your review.

Commit e4d9ad8 changes format so the oldest timestamp doesn't move past the prepared timestamp. That fix seems right to me, and seems to help, but I may not have it right, since it doesn't avoid the problem entirely (or maybe there's just another failure path I haven't found yet).

Commit 694eeed changes `__wt_txn_rollback()` to allow aborted transactions, `__wt_txn_prepare` sets `WT_UPDATE_RESERVE` updates to aborted. With that change, commit fc4b42d changes format to prepare transactions that are then rolled back.

Commit 6eb70e4 changes the prepare and rollback code to assert no updates are being prepared/rolled back from the lookaside file, and ignores all prepare/rollback updates from the metadata file.  I may have been a little too aggressive here: the original prepare code ignored metadata updates for the purposes of setting the prepare timestamp, but still set the update state to prepared. That looked wrong to me, but the change should be confirmed by one of you.

Edit:
Commit b18dc22 is on a separate commit because I'm suspicious I'm misunderstanding a visibility rule. Anyway, it seems to me like it's unsafe to do a type switch there because the update list can be truncated and freed (that's the whole point of switching the type), which means commit/rollback will update a freed WT_UPDATE.

Edit:
Commit 34bde21 changes prepare to not remove the transaction from the global table, which keeps eviction from discarding pages which include prepared (but not resolved), transactions.
@bvpvamsikrishna, I'd suggest expanding where WT_UPDATE.state field is set so we can assert `__wt_free_update_list()` never frees a WT_UPDATE structure in an unresolved state.

That fixes the last problem I'm seeing, this branch is now passing my testing.

### DIFF
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -117,9 +117,7 @@ __wt_delete_page(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
 	/*
 	 * If this WT_REF was previously part of a fast-delete operation, there
 	 * may be existing page-delete information. The structure is only read
-	 * after a WT_REF_DELETED state is switched to locked: immediately after
-	 * locking (from a state other than WT_REF_DELETED), free the previous
-	 * version.
+	 * while the state is locked, free the previous version.
 	 *
 	 * Note: changes have been made, we must publish any state change from
 	 * this point on.
@@ -294,9 +292,9 @@ __wt_delete_page_skip(WT_SESSION_IMPL *session, WT_REF *ref, bool visible_all)
 
 	/*
 	 * The page_del structure can be freed as soon as the delete is stable:
-	 * it is only read when the ref state is WT_REF_DELETED.  It is worth
-	 * checking every time we come through because once this is freed, we
-	 * no longer need synchronization to check the ref.
+	 * it is only read when the ref state is locked. It is worth checking
+	 * every time we come through because once this is freed, we no longer
+	 * need synchronization to check the ref.
 	 */
 	if (skip && ref->page_del != NULL && (visible_all ||
 	    __wt_txn_visible_all(session, ref->page_del->txnid,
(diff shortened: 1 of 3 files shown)


bvpvamsikrishna:
@keithbostic, Thanks a lot for the changes.

The changes mostly looks good to me except, the changes with respect to visibility (i.e commit  [34bde21](https://github.com/wiredtiger/wiredtiger/pull/3966/commits/34bde2166bb388671cfcdf0035cefdbe83e3c6c7)) is only required now, as [WT-3913](https://jira.mongodb.org/browse/WT-3913) is still in progress. After [WT-3913](https://jira.mongodb.org/browse/WT-3913) is merged, the visibility rules will be proper to compliment these changes.

So, in my opinion, it would be best to disable the prepare_transactions part of test/format for now, till [WT-3913](https://jira.mongodb.org/browse/WT-3913) & [WT-3849 ](https://jira.mongodb.org/browse/WT-3849) are merged. 
Would you mind making these changes to this pull request.

keithbostic:
@bvpvamsikrishna 
> Would you mind making these changes to this pull request.

Done and pushed.

keithbostic:
@bvpvamsikrishna, I'll merge this branch once the PR passes the standard tests, so we get testing back online. I don't think there's anything scary in it at this point, regardless.

Obviously, we need to remember to turn prepare testing back on in format as soon as the visibility test is fixed.

keithbostic:
@bvpvamsikrishna, in the NYC stand up this morning, @GeertBosch said they have a scheduled item to do multithreaded testing that would include transaction prepare. That's potentially going to fail in any WiredTiger drop since I reverted 34bde21. (It can also fail in single-threaded mode, but it's a lot less likely to do so, unless the cache is severely constrained.)

@GeertBosch said NYC is happy to wait on multithreaded prepare testing, but please reach out to him when a future drop can be expected to work for that testing.

PR #3967, opened 2018-03-05
WT-3942 Update test_compact02 to handle being halted by eviction pressure

keithbostic:
@agorrod, I'm pushing a fix for this one so we don't have to keep triaging the failure.

I don't see any evidence there's an underlying bug, it's just taking a very long time for eviction to finish in the small cache case, so I'm increasing the maximum timer from 1 minute to 5.


agorrod:
lgtm

PR #3969, opened 2018-03-07
WT-3963 Add a schema-abort test.

sueloverso:
Not ready for merge but for discussion and debugging.


agorrod:
autobuild: retest this please

agorrod:
@sueloverso I made some changes to get this branch passing tests. How would you feel about getting it merged? Is there more material work to be done?

sueloverso:
@agorrod I have not looked at this in a while. I'll study it and see if it is adequate.

agorrod:
@sueloverso I wasn't sure of the status - you pushed a commit, but I'm not confident that is all the changes you want in here. Please merge if you are happy.

sueloverso:
I need to revisit this one and see what, if anything, more I wanted.

sueloverso:
@agorrod here's the status of this test, I believe:
* It needs a bunch more comments to describe what it is testing. I will add those.
* The threads do some schema work and then also some timestamped CRUD work. After crash/restart, the program only verifies the CRUD work. It would be nice if we could verify the state of the schema/files after crash and recovery. That could be follow-on work.

I will add the comments and we can get this merged and into testing.

sueloverso:
@agorrod Researching this more deeply, the reason for the `#define WT_ENABLE_SCHEMA_TXN 1` change on this branch was to make sure and confirm that the test program reproduced the `WT_ROLLBACK` issue seen in the original failures when that transaction existed. So yes, the test passes by disabling that but it was there to make sure the test itself is valid.

sueloverso:
@agorrod I also see periodic failures in `test_timestamp_abort` on this branch due to the addition of LSM support. If we want this branch we should decide whether to debug and pursue those failures or revert the changes to `timestamp_abort/{main.c,smoke.sh}`

sueloverso:
My one reluctance is that this test (re-enabling the txns) has not reproduced the rollback error so it isn't clear this test is useful yet. I don't think it is harmful either, but not sure of its value-add.

agorrod:
Thanks @sueloverso I'm going to go ahead and merge. I'll also create a follow on ticket to enable LSM testing.

PR #3971, opened 2018-03-07
WT-3938 Allow the session table cache to be disabled.

michaelcahill:
Adds a new config to ::wiredtiger_open, "session_table_cache", which
defaults to true (maintaining existing behavior).  With
"session_table_cache=false", table handles are only held open while in
use, reducing the memory associated with each session handle (at the
expense of additional work next time the table is accessed).


src/schema/schema_stat.c
```diff
@@ -19,14 +19,17 @@ __wt_curstat_colgroup_init(WT_SESSION_IMPL *session,
...
-	WT_RET(__wt_schema_get_colgroup(session, uri, false, NULL, &colgroup));
+	WT_RET(__wt_schema_get_colgroup(
+	    session, uri, false, &table, &colgroup));
 
 	WT_RET(__wt_scr_alloc(session, 0, &buf));
 	WT_ERR(__wt_buf_fmt(session, buf, "statistics:%s", colgroup->source));
 	ret = __wt_curstat_init(session, buf->data, NULL, cfg, cst);
 
 err:	__wt_scr_free(session, &buf);
+	WT_TRET(__wt_schema_release_table(session, table));
```
michaelcahill: @agorrod, check out this change.  It is necessary because otherwise the `table` is released inside `__wt_schema_get_colgroup`, and releasing can now cause the table handle to be destroyed.

I wonder whether this could also be causing the crash we have seen reported when opening statistics cursors?

michaelcahill:
@ddanderson, this is a change for the `mongodb-3.4` branch and earlier, to disable caching of table handles per session.  Can you please take a look?

michaelcahill:
Note that testing always fails for patches against maintenance branches (there are warnings on those branches that we haven't backported the fixes for, so even an empty patch build will fail).

src/schema/schema_stat.c
agorrod: That does seem like a possibility. Should I open a ticket to make a similar change in develop?

agorrod:
This lgtm - if @ddanderson approves I think we should go ahead and merge.

src/schema/schema_open.c
```diff
@@ -551,12 +551,13 @@ __wt_schema_get_colgroup(WT_SESSION_IMPL *session,
 			if (tablep != NULL)
```
ddanderson: The `if` check can be removed, the `else` block can be removed. Per the assertion at the top of this function, `tablep` cannot be NULL.

src/schema/schema_open.c
```diff
@@ -595,7 +596,8 @@ __wt_schema_get_index(WT_SESSION_IMPL *session,
...
```
ddanderson: The `else` clause cannot happen.

src/schema/schema_open.c
```diff
@@ -606,13 +608,11 @@ __wt_schema_get_index(WT_SESSION_IMPL *session,
 	    session, table, tend + 1, strlen(tend + 1), indexp));
```
ddanderson: The `tablep` check can likewise be removed.

michaelcahill:
Thanks for the review @ddanderson, I made those changes.

PR #3972, opened 2018-03-07
WT-3964 Stop wrapping schema operations in a transaction

agorrod:
It causes an issue with recovery after a hard crash.


sueloverso:
I think this needs to be #ifdef not #ifndef because it that is not defined so it will leave them in.

agorrod:
Thanks @sueloverso - fixed.

PR #3973, opened 2018-03-08
WT-3815 Performance changes and code simplification

ddanderson:
Performance changes include:
  - don't sweep for dead cursors too often
  - handle a "" config string in open_cursor efficiently.
  - cursors with checkpoint specified are not cached.


agorrod:
Thanks @ddanderson that is great news!

michaelcahill:
Thanks @ddanderson, lgtm.  Feel free to merge: I'm sorry I don't have time tonight to watch tests.

I also looked at your most recent patch build.  I think I led you slightly astray with `truncateStart` cursor.  We'll need to call `truncateStart->next(truncateStart)` before passing it to the truncate method.

PR #3974, opened 2018-03-08
WT-3938 Ignore the "session_table_cache" wiredtiger_open config.

michaelcahill:
This was added to support older versions of MongoDB but is no longer required because the per-session table cache has been replaced.  Making this change so that customers with this option enabled will have a seamless upgrade.


agorrod:
Thanks @michaelcahill lgtm

PR #3975, opened 2018-03-08
WT-3967 Fix a coding error that was causing cursor sweep not to run.

ddanderson:
Also updated the long sweep test to correspond to the current sweep algorithm,
which doesn't run as often.

### DIFF
--- a/test/suite/test_cursor13.py
+++ b/test/suite/test_cursor13.py
@@ -387,11 +387,14 @@ class test_cursor13_big_base(test_cursor13_base):
         return uri_map
 
     def close_uris(self, uri_map, range_arg):
+        closed = 0
         for i in range_arg:
             cursors = uri_map[self.uriname(i)]
             while len(cursors) > 0:
                 cursors.pop().close()
                 self.closecount += 1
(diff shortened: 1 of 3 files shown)


ddanderson:
@sueloverso, could you do a review of this please?

test/suite/test_cursor13.py
```diff
@@ -465,25 +468,32 @@ def test_cursor_sweep(self):
...
                 bottom_range = half
                 # Let the dhandle sweep run and find the closed cursors.
                 time.sleep(3.0)
             else:
                 bottom_range = 0
 
+            # The session cursor sweep runs at most once a second and
+            # traverses a fraction of the cached cursors.  We'll run for
+            # ten seconds with pauses to make sure we see sweep activity.
+            pause_point = self.opens_per_round / 100
```
sueloverso: While I realize `opens_per_round` is large, we probably should make sure `pause_point` does not end up as 0 to avoid a divide-by-zero later.

ddanderson: Thanks, I'll add that.

PR #3976, opened 2018-03-09
WT-3968 Use compression feedback to improve WiredTiger compression.

keithbostic:
@agorrod, for your consideration, I've written up some notes in the JIRA ticket.


keithbostic:
@agorrod, I ended up using `memory_page_image_max` as the configuration value and having it apply to both internal and leaf pages, let me know if you disagree -- I was hesitant to add both an internal and leaf page version of the max-image, it didn't seem all that useful (although, to be honest, I don't see much utility in separate maximum leaf and internal page sizes either, at this point).

keithbostic:
autobuild: retest this please

src/include/btree.h
```diff
@@ -95,7 +95,11 @@ struct __wt_btree {
 	uint32_t maxleafkey;		/* Leaf page max key size */
 	uint32_t maxleafvalue;		/* Leaf page max value size */
 	uint64_t maxmempage;		/* In-memory page max size */
+	uint32_t maxmempage_image;	/* In-memory page image max size */
 	uint64_t splitmempage;		/* In-memory split trigger size */
+#define	WT_COMPRESS_ADJ		100
+	uint32_t intl_compadjust;	/* Internal page compression adjust */
+	uint32_t leaf_compadjust;	/* Leaf page compression adjust */
```
agorrod: Do we reconcile enough internal pages to justify splitting the calculation up? I'd be inclined to just share a compression ratio.

src/reconcile/rec_write.c
```diff
@@ -2239,16 +2239,34 @@ __wt_split_page_size(WT_BTREE *btree, uint32_t maxpagesize)
...
+
+	/*
+	 * Adjust for compression. Compression can be too effective, bound the
+	 * page split size at the maximum in-memory page image size, so point
+	 * updates don't cause forced eviction.
+	 */
+	compadjust = WT_PAGE_IS_INTERNAL(r->page) ?
+	    btree->intl_compadjust : btree->leaf_compadjust;
+	if (compadjust != 0) {
+		split_size = (split_size * compadjust) / WT_COMPRESS_ADJ;
```
agorrod: I don't really get this calculation. I didn't expect it to be divided by `WT_COMPRESS_ADJ`, oh is that just to avoid using a floating point number? i.e: we are shifting two decimal places?

I'd assumed that `WT_COMPRESS_ADJ` was used to control how much history weighted in when adjusting the compression rate.

src/reconcile/rec_write.c
```diff
@@ -3729,13 +3753,34 @@ __rec_split_write(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+	 */
+	if (!btree->compressor_is_snappy &&
+	    btree->type != BTREE_COL_FIX && compressed_size) {
+		compression_adj = (uint32_t)(((double)
+		    chunk->image.size / compressed_size) * WT_COMPRESS_ADJ);
+		adjustp = WT_PAGE_IS_INTERNAL(r->page) ?
+		    &btree->intl_compadjust : &btree->leaf_compadjust;
+		if (*adjustp == 0 ||
+		    compression_adj < *adjustp - *adjustp / 20 ||
+		    compression_adj > *adjustp + *adjustp / 20)
```
agorrod: I'd expected this to be more like the algorithm in eviction tuning, that takes historical success into account. i.e something like:

```
#define COMPRESSION_ADJUST_HISTORY_WEIGHT 200
copmadjust = ((compadjust * (COMPRESSION_ADJUST_HISTORY_WEIGHT - 1)) + compression_adj) / COMPRESSION_ADJUST_HISTORY_WEIGHT;
```

Do you think that's not necessary?

keithbostic: I thought about considering history when doing this work, but eventually didn't in favor of simplicity. In commit c5a90f4, I made a change so we'll only move 10% (of the shifted value), at a time, that is, outliers will only move the compression ratio by a small amount.

src/reconcile/rec_write.c
keithbostic: `WT_COMPRESS_ADJ` is just to avoid using a floating point number, I've renamed it to `WT_COMPRESS_SHIFT` and added a comment.

src/include/btree.h
keithbostic: Well, every leaf page reconciliation triggers a reconciliation of its parent, so I would expect us to reconcile more internal pages than leaf pages.

That said, I don't disagree with you: arguing internal pages compress significantly differently than leaf pages requires that block-manager address cookie values make a difference in compression (vs. normal leaf-page values), and I have no data to show that's the case.

keithbostic:
@agorrod, I've tried to respond to your comments. To be honest, I don't like the algorithm to slow-walk the changes to the shared memory adjustment value all that much, but I can't think of anything that's better. My worry is we'll get into a cycle of frantically updating that variable (but then again, I guess we're doing I/O so updating shared memory without barriers isn't really going to be the primary issue).

Let me know if you have a better idea!

agorrod:
@keithbostic to close the loop here - we talked and you are going to change the code to operate on floating numbers, and limit the adjustment to a proportion of the actual difference, so the algorithm adjusts gradually.

keithbostic:
@agorrod, another run at dynamic compression adjustment.

The short version is that rather than trying to maintain an "adjustment" percentage, I'm maintaining a "pre-compression" page size, and that's what we use as the number of bytes to accumulate before compression is called.

If you have time, I'd like to walk you through the changes to make sure I'm moving in the right direction.

agorrod:
@keithbostic I've pushed the foundation for a test case in here  - I'd appreciate if you can take a look. It does not help in that it doesn't tell us how compressible our users data will be, but it does at least give us the ability to test with data that compresses differently, to make sure our algorithm works as expected.

@ddanderson please take a look at the workgen changes I made - they aren't very pretty - I'd appreciate your feedback about how to make them better.

keithbostic:
autobuild: retest this please

bench/workgen/workgen.cxx
```diff
@@ -1106,13 +1105,40 @@ void Operation::kv_gen(bool iskey, uint32_t randomizer, uint64_t n,
...
+
+        /* Never overwrite the record number identifier */
+        if (random_len > size - 20)
+            random_len = size - 20;
+
+        for (int i = 0; i < random_len; ++i)
+            /*
+             * TODO: It'd be nice to use workgen_rand here, but this class
+             * is without the context of a runner thread, so it's not easy
+             * to get access to a state.
```
ddanderson: @agorrod, to get the equivalent of calling workgen_rand here, I think maybe passing the ThreadRunner as an argument to kv_gen(), and then have a small method on ThreadRunner that returns a random value.

bench/workgen/workgen.cxx
```diff
@@ -1106,13 +1105,40 @@ void Operation::kv_gen(bool iskey, uint32_t randomizer, uint64_t n,
...
-        randomizer %= 1000;
-        snprintf(result, 6, ":%3.3d:", randomizer);
-        n -= RANDOMIZER_SIZE;
-        result += RANDOMIZER_SIZE;
-    }
+    /* Setup the buffer, defaulting to zero filled. */
     workgen_u64_to_string_zf(n, result, size);
+
+    /*
+     * Compressability is a percentage, 100 is all zeroes, it applies to the
```
ddanderson: Compressibility (spelling typo).

bench/workgen/workgen.cxx
```diff
@@ -1106,13 +1105,40 @@ void Operation::kv_gen(bool iskey, uint32_t randomizer, uint64_t n,
...
+         * 100 - ((100 * 25) / 100) = 75
+         * That means that 75% of the string will be random numbers, and 25
+         * will be easily compressible zero-fill.
+         */
+        uint64_t random_len = size - ((size * compressibility) / 100);
+
+        /* Never overwrite the record number identifier */
+        if (random_len > size - 20)
+            random_len = size - 20;
+
```
ddanderson: We don't have record number identifiers in the value like wtperf, so we don't have to make an adjustment.

bench/workgen/workgen.cxx
```diff
@@ -1106,13 +1105,40 @@ void Operation::kv_gen(bool iskey, uint32_t randomizer, uint64_t n,
     if (n > max)
         THROW((iskey ? "Key" : "Value") << " (" << n
           << ") too large for size (" << size << ")");
-    if (randomizer != 0) {
-        randomizer %= 1000;
-        snprintf(result, 6, ":%3.3d:", randomizer);
-        n -= RANDOMIZER_SIZE;
-        result += RANDOMIZER_SIZE;
-    }
```
ddanderson: I think we these changes, we should get rid of the `RANDOMIZER_SIZE` definition.

ddanderson:
@agorrod, I also noticed that the runners you added have some functions that could be shared with other runners (`operations` and `op_append`), but that issue is not specific to your changes.  I'll open a separate ticket.

bench/workgen/workgen.cxx
agorrod: I made a change like that - it was more invasive than I'd hoped, with a possible layering violation of referencing `ThreadRunner` in `workgen.h`

agorrod:
@michaelcahill would you mind taking a look at this? It doesn't apply to snappy compression, so I think it's pretty safe, but definitely worth considering the magic numbers and the algorithm @keithbostic and I ended up with.

michaelcahill:
@agorrod, I'm happy with where this change has ended up.  Should I go ahead and merge?

agorrod:
Yes please.

michaelcahill:
Thanks for pushing this one through @keithbostic, lgtm.  I'll go ahead and merge.  It will be interesting to keep an eye on perf tests over the next day or so (both [Jenkins](http://source.wiredtiger.com/jenkins/plots/) and [Evergreen](https://evergreen.mongodb.com/waterfall/sys-perf?bv_filter=develop&skip=0)) to see if there is any unexpected fallout.

PR #3978, opened 2018-03-11
WT-3972 Allow more than 64K cursors on a single table.

michaelcahill:


src/support/mtx_rw.c
```diff
@@ -48,9 +48,8 @@
  *		uint8_t current;		// Current ticket
  *		uint8_t next;			// Next available ticket
  *		uint8_t reader;			// Read queue ticket
- *		uint8_t __notused;		// Padding
- *		uint16_t readers_active;	// Count of active readers
- *		uint16_t readers_queued;	// Count of queued readers
+ *		uint8_t readers_queued;		// Count of queued readers
```
agorrod: With this change, I think the comment [here](https://github.com/wiredtiger/wiredtiger/blob/develop/src/support/mtx_rw.c#L209) should be brought up to the general description block - I presume that's the reason why it's OK to use a smaller number for queued readers?

michaelcahill: Agreed, that comment is the reason this change is possible.  I'll update with your suggestion.

test/suite/test_cursor14.py
```diff
@@ -0,0 +1,62 @@
...
+        ('table-r-complex', dict(type='table:', keyfmt='r',
+            dataset=ComplexDataSet)),
+        ('table-S-complex', dict(type='table:', keyfmt='S',
+            dataset=ComplexDataSet)),
+        ('table-S-complex-lsm', dict(type='table:', keyfmt='S',
+            dataset=ComplexLSMDataSet)),
+    ])
+
+    # WT_CURSOR.insert doesn't leave the cursor positioned, verify any
+    # subsequent cursor operation fails with a "key not set" message.
```
sueloverso: This comment seems entirely unrelated to the code below it.

michaelcahill: Yes, you're right -- this was inherited from whatever test I started with.  I've removed it.

agorrod:
lgtm

PR #3980, opened 2018-03-12
WT-3958 Add last checkpoint timestamp to query_timestamp.

sueloverso:
@agorrod or @keithbostic please review.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -899,6 +899,13 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	 * checkpointing the metadata since we know that all files in the
 	 * checkpoint are now in a consistent state.
 	 */
+#ifdef HAVE_TIMESTAMPS
+	/*
+	 * Record the last checkpoint timestamp if we were successful.
+	 */
+	__wt_timestamp_set(
+	    &conn->txn_global.last_ckpt_timestamp, &txn->read_timestamp);
+#endif
 	WT_ERR(__wt_txn_commit(session, NULL));
 
 	/*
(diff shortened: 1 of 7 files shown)


src/txn/txn_ckpt.c
```diff
@@ -899,6 +899,13 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
 	 * checkpointing the metadata since we know that all files in the
 	 * checkpoint are now in a consistent state.
 	 */
+#ifdef HAVE_TIMESTAMPS
+	/*
+	 * Record the last checkpoint timestamp if we were successful.
+	 */
+	__wt_timestamp_set(
```
agorrod: How much does the exact ordering matter here? I would have expected to only set this timestamp:
* When doing a full checkpoint.
* After the metadata has been updated to reflect the finalized checkpoint.

i.e: I think it's preferrable for the value returned via `last_checkpoint` to slightly lag the actual checkpoint creation, than for it to be updated prior to the checkpoint being finalized.

sueloverso: The reason I put it here is that we're reading the timestamp out of the transaction and this is the last point it is valid. Once `wt_txn_commit` returns the `read_timestamp` has been cleared (in `wt_txn_release` we call `wt_txn_clear_read_timestamp`). I could store it in a temp variable to be set later. 

But you're correct that it should only be on a full checkpoint. I will fix that.

sueloverso: I took the temp variable route as that is more correct as you stated. Fixed.

agorrod:
Thanks @sueloverso lgtm, and it passed "configure combinations" Jenkins test - so there aren't any conditional timestamp issues.

PR #3984, opened 2018-03-13
WT-3959 Fix setting recovery timestamp on restart.

sueloverso:
Not yet ready for review.


sueloverso:
@michaelcahill and @agorrod this branch is ready for review. I implemented using the recovery timestamp if we're checkpointing with timestamps but no stable has been set yet. That covers the case where `wt` gets run and when no timestamps are in use at all. @dgottlieb just FYI.

One thing to note, I put in a large comment in `meta_ckpt.c` because I write `checkpoint_timestamps` in the metadata records of all the tables even though only the one in the turtle file is valid and used. Those can be used as a debugging aid via `wt list -v`.

sueloverso:
After thinking about it, I decided to only write the checkpoint timestamp into the turtle file and leave all others as "0". I came up with different scenarios where having information on each table could be confusing and misleading. If there is a need for debugging or ever a need to query an individual table's timestamp, we can add it back in.

sueloverso:
Close in favor of #4038 .

PR #3985, opened 2018-03-14
WT-3849 Prepare timestamp validation.

bvpvamsikrishna:


bvpvamsikrishna:
@sulabhM, @nehakhatri5 
Request your review for this pull request.

src/txn/txn.c
```diff
@@ -990,22 +988,14 @@ __wt_txn_prepare(WT_SESSION_IMPL *session, const char *cfg[])
 	txn = &session->txn;
 
 	WT_ASSERT(session, F_ISSET(txn, WT_TXN_RUNNING));
-	/* Transaction should not have a commit timestamp set. */
-	WT_ASSERT(session, !F_ISSET(txn, WT_TXN_HAS_TS_COMMIT));
```
nehakhatri5: Can a prepared transaction have a commit timestamp already set?

src/txn/txn_timestamp.c
```diff
@@ -668,6 +680,83 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+				    cval.str, hex_timestamp);
+			}
+		 }
+#else
+		WT_UNUSED(timestamp);
+		WT_RET_MSG(session, EINVAL, "prepare_timestamp requires a "
+		    "version of WiredTiger built with timestamp support");
+#endif
+	} else
+		WT_RET_MSG(session, EINVAL, "prepare timestamp is required");
```
nehakhatri5: Is it required to have this error? I see that __wt_txn_parse_read_timestamp does not have a similar error.

src/txn/txn_timestamp.c
```diff
@@ -668,6 +680,83 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+ */
+int
+__wt_txn_parse_prepare_timestamp(
+    WT_SESSION_IMPL *session, const char *cfg[], wt_timestamp_t *timestamp)
+{
+	WT_CONFIG_ITEM cval;
+
+	WT_RET(__wt_config_gets_def(session,
+	    cfg, "prepare_timestamp", 0, &cval));
+	if (cval.len > 0) {
```
nehakhatri5: Here, should we check for ret == 0 also?

src/txn/txn_timestamp.c
```diff
@@ -668,6 +680,83 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+#ifdef HAVE_TIMESTAMPS
+		WT_TXN *prev;
+		WT_TXN_GLOBAL *txn_global;
+		wt_timestamp_t oldest_ts;
+		char hex_timestamp[2 * WT_TIMESTAMP_SIZE + 1];
+
+		txn_global = &S2C(session)->txn_global;
+
+		if (F_ISSET(&session->txn, WT_TXN_HAS_TS_COMMIT))
```
nehakhatri5: Should this be an ASSERT like before?

src/txn/txn_timestamp.c
```diff
@@ -668,6 +680,83 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+		prev = TAILQ_LAST(&txn_global->read_timestamph,
+		    __wt_txn_rts_qh);
+		if (prev != NULL &&
+		    __wt_timestamp_cmp(&prev->read_timestamp, timestamp) >= 0) {
+			__wt_readunlock(session,
+			    &txn_global->read_timestamp_rwlock);
+			WT_RET(__wt_timestamp_to_hex_string(session,
+			    hex_timestamp, &prev->read_timestamp));
+			WT_RET_MSG(session, EINVAL,
+			    "prepare timestamp %.*s not later than an active "
```
nehakhatri5: s/not later than/older than

src/txn/txn_timestamp.c
```diff
@@ -668,6 +680,83 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+			    "read timestamp %s ", (int)cval.len, cval.str,
+			    hex_timestamp);
+		}
+		__wt_readunlock(session, &txn_global->read_timestamp_rwlock);
+
+		/*
+		 * If there are no active readers, prepare timestamp must not
+		 * be older than oldest timestamp.
+		 */
+		if (prev == NULL) {
```
nehakhatri5: If there are active readers, would we allow a prepare ts older that oldest ts?

src/txn/txn_timestamp.c
```diff
@@ -668,6 +680,83 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
+		wt_timestamp_t oldest_ts;
+		char hex_timestamp[2 * WT_TIMESTAMP_SIZE + 1];
+
+		txn_global = &S2C(session)->txn_global;
+
+		if (F_ISSET(&session->txn, WT_TXN_HAS_TS_COMMIT))
+			WT_RET_MSG(session, EINVAL,
+			    "commit timestamp should not have been set before "
+			    "prepare transaction");
+
```
nehakhatri5: What if an application tries to set a prepare ts again for a ts, would we allow that?

test/suite/test_prepare04.py
```diff
@@ -27,7 +27,7 @@
 # OTHER DEALINGS IN THE SOFTWARE.
 #
 # test_prepare04.py
-#   Prepare: prepare conflict with update and read operations
+#   Prepare: Update and read operations generates prepared conflict error.
```
nehakhatri5: s/generates/generate

test/suite/test_prepare05.py
```diff
@@ -0,0 +1,119 @@
...
+            lambda: self.session.prepare_transaction(
+            'prepare_timestamp=' + timestamp_str(1)),
+            "/older than the oldest timestamp/")
+        self.session.commit_transaction('commit_timestamp=' + timestamp_str(3))
+
+        # Check setting the prepare timestamp same as oldest timestamp is valid.
+        self.session.begin_transaction()
+        self.session.prepare_transaction('prepare_timestamp=' + timestamp_str(2))
+        self.session.commit_transaction('commit_timestamp=' + timestamp_str(3))
+
```
nehakhatri5: What about a test setting prepare ts again for a txn.

src/txn/txn_timestamp.c
bvpvamsikrishna: WT_RET at line 693 does that.

src/txn/txn_timestamp.c
bvpvamsikrishna: Assert will be effective only in case of diagnostic build, so I think return error is better.

test/suite/test_prepare05.py
bvpvamsikrishna: already in test_prepare02.py

src/txn/txn_timestamp.c
bvpvamsikrishna: it will through error, please refer SESSION_API_CALL

src/txn/txn_timestamp.c
bvpvamsikrishna: no active reader will be older than oldest ts. so if we validate against active reader, it implies we have already validated against oldest ts.

src/txn/txn_timestamp.c
bvpvamsikrishna: not later than is like greater and if we change it to older than, then we miss equal to condition. So I choose to use not later than.

src/txn/txn_timestamp.c
bvpvamsikrishna: Even for set read timestamp also, this error will be returned, only that check is at different place i.e. in caller.

src/txn/txn_timestamp.c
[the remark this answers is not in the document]
bvpvamsikrishna: Prepared transaction has no restriction on the isolation levels it supports.
We have a check in the read_timestamp, as the isolation level is more associated with read operation and timestamp reading is synonymous with snapshot isolation.

bvpvamsikrishna:

bvpvamsikrishna:
@agorrod, request your review for this pull request.

agorrod:
Thanks @bvpvamsikrishna lgtm

PR #3986, opened 2018-03-15
WT-3980 failure returning a modified update without a backing "real" update

keithbostic:
@agorrod, this should make future failures of this type more obvious.


agorrod:
Thanks @keithbostic lgtm

PR #3989, opened 2018-03-16
WT-3994 Avoid GCC-7 and Clang errors for mongodb-3.4 branch

lukech:
Test old branch PR Jenkins jobs

### DIFF
--- a/build_posix/aclocal/strict.m4
+++ b/build_posix/aclocal/strict.m4
@@ -56,6 +56,8 @@ AC_DEFUN([AM_GCC_WARNINGS], [
 		w="$w -Wno-error=format-truncation"
 		w="$w -Wno-error=format"
 		w="$w -Wno-error=implicit-fallthrough"
+		w="$w -Wno-error=maybe-uninitialized"
+		w="$w -Wno-error=unsafe-loop-optimizations"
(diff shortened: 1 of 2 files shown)


michaelcahill:
@lukech, I like this change (apart from the dummy change to `txn.c`).  Is it ready for review and merge?

lukech:
@michaelcahill I probably need to talk with @agorrod about the solution being taken, before kicking off the code review (and likely converting the BUILD ticket into a WT ticket).

lukech:
@agorrod @keithbostic (cc @michaelcahill), please help to review the changes, which were based on the list of GCC warning option changes in WT-3394. Added a couple of more lines to avoid fatal errors for some warnings. Please note warnings will be retained with the "-Wno-error-xxx" setting (though not treated as errors).

lukech:
Confirmed the same build errors triggered by the default GCC-7 warning settings also existed on mongodb-3.2, but not mongodb-3.6. We may need the same changes to be ported back to mongodb-3.2 later on.

agorrod:
@lukech is this branch ready to merge?

lukech:
@agorrod Yes, will merge.

PR #3991, opened 2018-03-19
WT-3985 Pre-allocated log files accumulate on Windows

keithbostic:
@sueloverso, the problem was the "continue" statements when we hit `.` and `..` were branching to the end of the loop, `single` was true and so we returned immediately.

All the credit goes to @markbenvenuto, he figured it out.


sueloverso:
One question here is why it didn't stop creating new ones? Does that code have a bug or did the non-single path also return a 0 count? Do we have any test that can confirm this fixes the problem?

src/os_win/os_dir.c
```diff
@@ -59,30 +59,41 @@ __directory_list_worker(WT_FILE_SYSTEM *file_system,
...
+			break;
+
+skip:		if (FindNextFileW(findhandle, &finddata) != 0)
+			continue;
+		windows_error = __wt_getlasterror();
+		if (windows_error == ERROR_NO_MORE_FILES)
+			break;
+		__wt_errx(session,
+		    "%s: directory-list: FindNextFileW: %s",
+		    pathbuf->data, __wt_formatmessage(session, windows_error));
```
markbenvenuto: Use `prefix` instead of `pathbuf->data` in the error message. `pathbuf->data` is `wchar_t` in this case.

markbenvenuto: After a closer look, my comment is wrong

test/suite/test_bug019.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # There was a bug where pre-allocated log files accumulated on
+    # Windows systems due to an issue with the directory list code.
+    def test_bug019(self):
+        # Create a table just to write something into the log.
+        self.session.create(self.uri, 'key_format=i,value_format=i')
+        self.populate(self.entries)
+        self.session.checkpoint()
+
+        # Loop, making sure pre-allocation is working and the range is moving.
```
sueloverso: This is not guaranteed to work. The reason is that the ordering of files returned by readdir is not known.  We only use one file returned from readdir that matches our prefix. That is not necessarily the one with the lowest number in its name. We've seen databases where some "old" (i.e. small numbered) pre-allocated log file stayed there for a long time.

One way to do it for this bug in particular would be to:
* Run the first populate. This puts pressure on the file creation and will cause the pre-allocation to decide it needs to allocate a bunch of files.
* Sleep a bit to let the internal thread run and create them.
* Take our initial `listdir` of pre-allocated files.
* In a loop a few times:
** Run populate.
** Take the `listdir` of pre-allocated files and make sure that it is a different set of files from our original list.
** Make that list the new original list.

With the original bug, the list of pre-allocated files after the second populate will be a superset of the original and it should fail.

keithbostic: Thank you, @sueloverso!

Fix pushed in 2c526b3.

michaelcahill:
Thanks @markbenvenuto, @keithbostic and @sueloverso!  lgtm, I'll merge.

sueloverso:
@keithbostic did you or @markbenvenuto verify that the current form of the test detected the original problem on Windows before the fix?

keithbostic:
@sueloverso 
> @keithbostic did you or @markbenvenuto verify that the current form of the test detected the original problem on Windows before the fix?

No, I don't believe either of us did.

markbenvenuto:
I did not run the python test.

PR #3993, opened 2018-03-21
WT-3987 Avoid reading lookaside pages in truncate fast path.

michaelcahill:


michaelcahill:
The key part of this change is not calling `__wt_las_page_skip_locked`, which (among other things) gives up if called as part of an update transaction.  Since truncate will always be part of an update transaction, that defeated the original intent.

I believe the code was also unnecessarily complicated in its handling of limbo pages (ones with a read-only snapshot of data but with history in the lookaside table).  It's simpler to just attempt to evict those pages (transitioning to `WT_REF_DISK`) before marking them deleted.

michaelcahill:
@keithbostic, can you please take a look?

It would also be great if you can turn the test you have into something we can run regularly but I know that may not be straightforward because it isn't obvious how to know how much reading is too much during a truncate call.

keithbostic:
@michaelcahill, lgtm, I'll merge.

For clarity:

 > It's simpler to just attempt to evict those pages (transitioning to WT_REF_DISK) before marking them deleted.

That should be transitioning to WT_REF_LOOKASIDE, correct?

This passes the test I wrote last night, we go from 25K lookaside table reads during truncate to a handful. If we're OK with limiting lookaside reads to 100, I think we could be confident the test would pass unless something has gone wrong.

michaelcahill:
@keithbostic thanks for the review, and you're correct about the transition to `WT_REF_LOOKASIDE` when we evict a limbo page.

PR #3994, opened 2018-03-21
WT-3987 Avoid reading lookaside pages in truncate fast path.

keithbostic:
@michaelcahill, this is [Zseries failure 41508](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-zseries/41508/console).

I believe the fix to only read `WT_REF.addr` once is needed, but I've never seen it fail, I noticed it while looking at the core dump.

### DIFF
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -174,6 +174,8 @@ __wt_delete_page_rollback(WT_SESSION_IMPL *session, WT_REF *ref)
 {
 	WT_UPDATE **updp;
 	uint64_t sleep_count, yield_count;
+	uint32_t current_state;
+	bool locked;
 
 	/*
 	 * If the page is still "deleted", it's as we left it, reset the state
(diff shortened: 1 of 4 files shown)


keithbostic:
@michaelcahill, this change makes things a lot better (and we should merge it), but I think there's another lurking bug:
```
case WT_REF_MEM:
case WT_REF_SPLIT:
        /*      
         * We can't use the normal read path to get a copy of 
         * the page because the session may have closed the
         * cursor, we no longer have the reference to the tree
         * required for a hazard pointer.  We're safe because 
         * with unresolved transactions, the page isn't going
         * anywhere.
         *       
         * The page is in an in-memory state, which means it
         * was instantiated at some point. Walk the list of 
         * update structures and abort them.
         */      
        for (upd =
            ref->page_del->update_list; *upd != NULL; ++upd)
                (*upd)->txnid = WT_TXN_ABORTED;
        goto done;
```
We failed in delete-rollback because `ref.page_del.update_list == NULL`.

I'm looking for that one, but so far nothing.

agorrod:
I switched stress testing onto this branch:

michaelcahill:
Thanks @keithbostic -- these changes look correct to me.

Is cd6ea41 sufficient to fix the `update_list == NULL` failure?

keithbostic:
@michaelcahill 
> Is cd6ea41 sufficient to fix the update_list == NULL failure?

I believe 2f4a8f1 explains and fixes the `WT_REF.page_del.update_list == NULL` core.

Note that 2fadb99 is an unrelated cleanup, I'm removing a bug fix I put in a couple of weeks ago, I think I misunderstood a failure I was seeing.

keithbostic:
@michaelcahill, another one, 0a1a7a9.

I've only seen this once in running all day, so I'm not entirely sure I fixed the right problem, but reconciliation failed in `__rec_child_modify()`, when it found a child in a `WT_REF_LOOKASIDE` state, and this is the only thing I can think of that might cause that.

I'm also not at all sure of the correctness of the statement that we don't require a lock to read the `WT_REF.page_del` structure when the only page we have locked is its parent.

michaelcahill:
@keithbostic, some coordination when committing a truncate operation with a timestamp is required.

Imagine transaction T does a truncate and marks page P deleted, but has not yet committed.  Now some other thread reads page P and instantiates tombstones on the page with txnid T.  If T commits concurrent with the read, it may not see the `update_list` and the tombstones would not get commit timestamps.

Given that, I've reinstated 2fadb99.

agorrod:
Thanks @keithbostic and @michaelcahill this change lgtm, I'm going to go ahead and merge. Let's watch for more fallout in stress testing.

PR #3995, opened 2018-03-22
WT-3982 Fix transaction visibility bugs related to lookaside usage.

michaelcahill:
At low isolation levels, new records can appear in between positioning a cursor and calling next/prev.  For this reason, search_near cannot assume that after calling next (or prev), it will have a key larger (or smaller) than the search key.  Loop and check in this (rare) case instead.

This applies particularly to internal code that searches in the lookaside table, because that access is always done at read-uncommitted isolation.

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -655,30 +655,43 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
 		exact = 0;
 		F_CLR(cursor, WT_CURSTD_KEY_SET | WT_CURSTD_VALUE_SET);
 		F_SET(cursor, WT_CURSTD_KEY_INT | WT_CURSTD_VALUE_INT);
-	} else if ((ret = __wt_btcur_next(cbt, false)) != WT_NOTFOUND) {
-		WT_ERR(ret);
-		exact = 1;
 	} else {
 		/*
-		 * The cursor next call may have overwritten our caller's key,
-		 * restore it to its original value.
+		 * We didn't find an exact match: try after the search key,
+		 * then before.  We have to loop here because at low isolation
+		 * levels, new records could appear as we are stepping through
+		 * the tree.
 		 */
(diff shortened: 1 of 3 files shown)


michaelcahill:
@sueloverso, I reproduced the failure from the ticket and traced it back some way to a `search_near` call returning with `exact == 1` but positioned on a key less than the search key.

I believe this can happen if there are two threads evicting pages with lookaside entries concurrently.  If thread 1 gets lookaside pageid 1 and thread 2 gets pageid 2, but thread 2 finishes the eviction then the page is read back in before thread 1 finishes writing, then the call to `__wt_las_cursor_position` could return with the cursor incorrectly positioned somewhere in the `pageid == 1` block.

That would mean the lookaside entries for pageid 2 would not be restored, leading to the type of failure seen in the test case on the ticket.

src/btree/bt_cursor.c
```diff
@@ -655,30 +655,43 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
...
+		while ((ret = __wt_btcur_next(cbt, false)) != WT_NOTFOUND) {
+			WT_ERR(ret);
+			if (btree->type == BTREE_ROW)
+				WT_ERR(__wt_compare(session, btree->collator,
+				    &cursor->key, &state.key, &exact));
+			else
+				exact = cbt->recno < state.recno ? -1 :
+				    cbt->recno == state.recno ? 0 : 1;
+			if (exact >= 0)
+				goto done;
```
sueloverso: Do we want to add in a stat that would count how many times we hit this situation (here and below)? It might be interesting right now because we're thinking about this specific issue and code but I'm not sure if it is helpful long term.

sueloverso:
This does not fix the problem. All of my overnight runs hit the problem on this branch as well.

agorrod:
Thanks @michaelcahill and @sueloverso these changes look good to me and fix the problem we were reproducing.

PR #3996, opened 2018-03-22
WT-3987 Avoid reading lookaside pages in truncate fast path

keithbostic:
@michaelcahill, @agorrod, for your review.

I split the commits into two parts so we can separately revert them, if I've gone in the wrong direction on page instantiation.

Also, can you think of a reason we didn't see this fail in format stress testing? It seems to me cursors reading from instantiated deleted pages in snapshot isolation should have read the wrong records (because we weren't applying lookaside records in the deleted-page instantiation path). I found the problem by inspection, and I'm worried we're not testing lookaside as well as we should be.

### DIFF
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -311,6 +311,31 @@ __wt_delete_page_skip(WT_SESSION_IMPL *session, WT_REF *ref, bool visible_all)
 }
 
 /*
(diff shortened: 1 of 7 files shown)


agorrod:
Thanks @michaelcahill and @keithbostic these changes lgtm, I'm going to go ahead and merge.

PR #3997, opened 2018-03-22
WT-3979 Doxygen upgrade reports warnings on build.

keithbostic:
This isn't a complete fix, but it gets rid of most of the warnings.


michaelcahill:
Sadly, my change requires 1.8.14 to work correctly.  cf https://github.com/doxygen/doxygen/pull/602

Reverting -- we should do this cleanup after we require 1.8.14 everywhere and upgrade all our test / development systems.

agorrod:
Thanks @keithbostic and @michaelcahill these changes lgtm.

PR #3998, opened 2018-03-23
WT-3990 Coverity tickets

keithbostic:


test/bloom/test_bloom.c
```diff
@@ -67,26 +68,22 @@ main(int argc, char *argv[])
 	g.c_key_max = 100;
 	g.c_k = 8;
 	g.c_factor = 16;
-	g.c_srand = 3233456;
 
 	/* Set values from the command line. */
-	while ((ch = __wt_getopt(progname, argc, argv, "c:f:k:o:s:")) != EOF)
+	while ((ch = __wt_getopt(progname, argc, argv, "c:f:k:o:")) != EOF)
```
agorrod: I'm a bit sad to lose the ability to set a seed - the intent was to allow testers to reproduce exactly the same sequence by specifying a starting point. Having said that.. I've not run the program like that for a long time, so I'm not going to push back hard.

keithbostic: @agorrod 
> I'm a bit sad to lose the ability to set a seed

I'm happy to put it back in if you think it's worth keeping.

PR #3999, opened 2018-03-26
WT-3973 Allow alter to modify app_metadata

sulabhM:


src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+	WT_DECL_RET;
+	char *value;
+	bool is_colgroup;
+
+	value = NULL;
+
+	is_colgroup = WT_PREFIX_MATCH(name, "colgroup:");
+	if (!is_colgroup && !WT_PREFIX_MATCH(name, "index:"))
+		WT_ERR_MSG(session, EINVAL,
+		    "expected a 'colgroup:' or 'index:' source: '%s'", name);
```
agorrod: It feels like alter should use consistent error reporting. i.e: either this or the `__wt_unexepected_object_type` used in `__alter_file`

src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+		    name, cfg, WT_CONFIG_BASE(session, index_meta)));
+
+err:	__wt_scr_free(session, &data_source);
+	__wt_free(session, value);
+	return (ret);
+}
+
+/*
+ * __alter_table --
+ *	WT_SESSION::alter for a table.
```
agorrod: This style of comment with WT_XXX::YYY is generally used for functions that are used as function pointers in handles - i.e it would be more appropriate as the comment for `__wt_schema_alter` than this function.

src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+		WT_ERR_MSG(session, EINVAL,
+		    "expected a 'colgroup:' or 'index:' source: '%s'", name);
+
+	/* Read the schema value. */
+	WT_ERR(__wt_metadata_search(session, name, &value));
+
+	/* Get the data source URI. */
+	if ((ret = __wt_config_getones(session, value, "source", &cval)) != 0)
+		WT_ERR_MSG(session, EINVAL,
+		    "index or column group has no data source: %s", value);
```
agorrod: This error message doesn't have enough context - it should mention that it was encountered while doing an alter.

src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+
+	WT_ERR(__wt_scr_alloc(session, 0, &data_source));
+	WT_ERR(__wt_buf_fmt(session,
+	    data_source, "%.*s", (int)cval.len, cval.str));
+
+	/* Alter the data source */
+	WT_ERR(__wt_schema_alter(session, data_source->data, cfg));
```
agorrod: Is it necessary to retrieve and alter the source?

src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+	/* Alter the indices. */
+	WT_ERR(__wt_schema_open_indices(session, table));
+	for (i = 0; i < table->nindices; i++) {
+		if ((idx = table->indices[i]) == NULL)
+			continue;
+		WT_ERR(__alter_tree(session, idx->name, cfg));
+	}
+
+	/* Alter the table */
+	WT_ERR(__alter_apply(session,
```
nehakhatri5: Unlike rename_table, why alter is done before __wt_meta_track_handle_lock? Just curious.

src/schema/schema_alter.c
nehakhatri5: +1

src/schema/schema_alter.c
sulabhM: It is the same as rename, here:
https://github.com/wiredtiger/wiredtiger/blob/master/src/schema/schema_rename.c#L145

I think if this error is encountered, it will be something like:
WT_SESSION.alter : index or column group has no data source: colgroup:blah

Though it should say it is because of alter, if desired I can modify the message to say "Alter failed because index or column ... ".

src/schema/schema_alter.c
sulabhM: I confirmed with @michaelcahill yesterday, we think it will be good to alter the data source.

src/schema/schema_alter.c
sulabhM: fixed

src/schema/schema_alter.c
sulabhM: fixed.

src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+		WT_TRET(__wt_schema_release_table(session, table));
+	return (ret);
+}
+
+/*
+ * __wt_schema_alter --
+ *	Alter an object.
+ */
+int
+__wt_schema_alter(WT_SESSION_IMPL *session, const char *uri, const char *cfg[])
```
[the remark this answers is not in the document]
sulabhM: fixed.

src/schema/schema_alter.c
```diff
@@ -63,7 +54,164 @@ err:	__wt_free(session, config);
...
+{
```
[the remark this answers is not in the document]
sulabhM: It pre-existed like this, I suppose it is assert just before usage. This function is re-writing the existing one here:
https://github.com/wiredtiger/wiredtiger/blob/master/src/schema/schema_alter.c#L38

test/suite/test_alter03.py
```diff
@@ -0,0 +1,81 @@
...
+#    Check if app_metadata can be altered.
+class test_alter03(wttest.WiredTigerTestCase):
+    name = "alter03"
+
+    def verify_metadata(self, metastr):
+        if metastr == '':
+            return
+        cursor = self.session.open_cursor('metadata:', None, None)
+        #
+        # Walk through all the metadata looking for the entries that are
```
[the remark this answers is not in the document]
sulabhM: This function is kind of common between all alter tests. I mostly just copied it.

We expect to find multiple URIs that match the metastr, hence we do not break out of the loop. The comment here says "... looking for the entries that are the URIs for the named object." I think once you understand that there will be multiple objects matching the name (because, for instance, a table has multiple objects) it is obvious that we want to check them all, and not break out of the loop.

test/suite/test_alter03.py
```diff
@@ -0,0 +1,81 @@
...
+        for k in range(entries):
+            c[k+1] = 1
+        c.close()
+
+        # Verify the string in the metadata
+        self.verify_metadata(app_meta_orig)
+
+        # Alter app metadata and verify
+        self.session.alter(uri, app_meta_new)
+
```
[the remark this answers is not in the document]
sulabhM: good catch.

src/schema/schema_alter.c
agorrod: Makes sense - leave as is

src/schema/schema_alter.c
agorrod: OK

src/schema/schema_alter.c
sulabhM: It benefits from releasing lock early. Rename is multistage, because it has to rename the underlying files/sources and then the metadata. Metadata rename is really search, remove-old and insert-new and can be done without metatracking under lock. Alter is simpler and is mostly just modifying configuration for the tables. Being just a single step, it needs to be done under metatracking.

agorrod:
Thanks @sulabhM lgtm

dist/api_data.py
```diff
@@ -15,10 +15,13 @@ def __init__(self, name, default, desc, subconfig=None, **flags):
...
 
-# Metadata shared by all schema objects
-common_meta = [
+common_runtime_config = [
     Config('app_metadata', '', r'''
         application-owned metadata for this object'''),
+]
+
+# Metadata shared by all schema objects
+common_meta = [
```
michaelcahill: I'd expect to see `common_meta = common_runtime_config + [...` rather than repeating it below.

Is there some case you're trying to exclude where we used to allow `app_metadata` but shouldn't?  Or does that lead to `app_metadata` appearing multiple times somewhere?

src/schema/schema_alter.c
```diff
@@ -63,7 +54,166 @@ err:	__wt_free(session, config);
...
+	/*
+	 * We know that we have exclusive access to the file.  So it will be
+	 * closed after we're done with it and the next open will see the
+	 * updated metadata.
+	 */
+	filename = uri = session->dhandle->name;
+	if (!WT_PREFIX_SKIP(filename, "file:"))
+		return (__wt_unexpected_object_type(session, uri, "file:"));
+
+	WT_RET(__wt_meta_track_on(session));
```
michaelcahill: I would expect the calls to `__wt_meta_track_on/off` to be in the top-level function (`__wt_schema_alter`) rather than just for `file:` types.

In case it isn't clear, this is like "begin transaction" for schema-changing operations: we want the whole operation wrapped in a transaction so that if it fails part-way through, we roll back all of the changes.

src/schema/schema_alter.c
```diff
@@ -63,7 +54,166 @@ err:	__wt_free(session, config);
...
+		if ((idx = table->indices[i]) == NULL)
+			continue;
+		WT_ERR(__alter_tree(session, idx->name, newcfg));
+	}
+
+	/* Alter the table */
+	WT_ERR(__alter_apply(session,
+	    uri, newcfg, WT_CONFIG_BASE(session, table_meta)));
+
+	if (WT_META_TRACKING(session)) {
```
michaelcahill: Associated with the comment above about `__wt_meta_track_on/off`, I think we should be able to:

```
WT_ASSERT, session, WT_META_TRACKING(session));
```

Otherwise, we're saying we can do an alter operation without the possibility of handling a failure.

dist/api_data.py
sulabhM: Yes, `app_metadata` started appearing multiple times at some places. For instance, `app_metadata` is defined over `file_runtime_config`. `file_config` builds upon `file_runtime_config` and `format_meta`, `format_meta` builds upon `common_meta`. If `app_metadata` is part of both `common_meta` and `file_runtime_config`, it shows up twice in `file_config`.

For this reason I had to pull out `app_metadata` from several meta configs and put in back in the final configs as needed.

src/schema/schema_alter.c
sulabhM: I did look at putting meta tracking around all the changes. What happens is once meta tracking is switched ON, when file alter tries to lock the dhandle in fails.
 (`__wt_session_get_dhandle` called by `__wt_session_get_btree_ckpt` called by `__wt_exclusive_handle_operation`).

I will work with @agorrod to figure out a solution to this problem.

agorrod:
@michaelcahill we've made a set of changes here - could you take a last look and merge if you are happy?

dist/api_data.py
michaelcahill: OK, I've taken a closer look and opened WT-4002 to follow up on this.

PR #4001, opened 2018-03-26
WT-3982 Revert a test change that broke in some configurations

agorrod:
Timestamp abort needs to increase the allowed number of session handles, if a high thread count were chosen.


michaelcahill:

PR #4003, opened 2018-03-26
WT-3996 Add a test of truncate with timestamps and lookaside.

michaelcahill:
Makes sure we read the expected data when lookaside eviction is active and data is truncated with a timestamp.  Fixes a bug checking whether truncate operation need the commit timestamp to be applied.  Even with that fix, this test failed before a8fde4b was merged.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -251,8 +251,18 @@ static inline bool
 __wt_txn_update_needs_timestamp(WT_SESSION_IMPL *session, WT_TXN_OP *op)
 {
 	WT_TXN *txn;
+	wt_timestamp_t *timestamp;
 
 	txn = &session->txn;
+
+	/*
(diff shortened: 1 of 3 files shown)


michaelcahill:
@keithbostic and @bvpvamsikrishna, as discussed, here is a standalone test that exercises truncate together with timestamps and lookaside eviction.

In the process, I discovered another bug with setting commit timestamps for truncate operations: the `__wt_txn_update_needs_timestamp` call was treating the `op->u.ref` union pointer as a `WT_UPDATE *`.

Prior to the first set of changes in WT-3897 it probably succeeds (I didn't check) because truncate will always take the slow path.

Prior to the final change in WT-3897 (a8fde4b), it detects that updates and not restored from the lookaside table when reading a truncated page with history.

With all of the changes in WT-3897 (plus the above-mentioned fix to commit), this test now passes.

PR #4004, opened 2018-03-26
WT-3997 The cursor walk code can spin without sleeping on restart/split.

keithbostic:


src/include/misc.i
```diff
@@ -223,3 +223,24 @@ __wt_txn_context_check(WT_SESSION_IMPL *session, bool requires_txn)
...
+	 * We yield before retrying, and if we've yielded enough times, start
+	 * sleeping so we don't burn CPU to no purpose.
+	 */
+	if ((*yield_count) < WT_THOUSAND) {
+		(*yield_count)++;
+		__wt_yield();
+		return;
+	}
+
+	(*sleep_count) = WT_MIN((*sleep_count) + 100, WT_THOUSAND);
```
michaelcahill: I know this code has just been moved here, but what's the point of this line?  We're going to count up to 1000 yields, then after that we'll keep setting `sleep_count` to `min(1100, 1000)`, aka 1000.

Was this code originally intended to do something smarter?

keithbostic:
@michaelcahill 
> Is there some way to be confident that a walk is responsible for the stats reported in SERVER-34082?

I doubt a walk is responsible for the statistics in SERVER-34082. I guess it's possible, but I think it's unlikely.

I noticed this case when looking at `WT_RESTART` and cursor operations, it's just a place where I believe we can spin on a restart.

Background for anyone reading this: the interesting thing in that ticket was `extraordinarily high rate of restarted searches, averaging about 4 M/s over the 1 minute stall`. That statistic is associated with the `cursor_restart` counter, set at a high-level when a cursor operation (for example, `__wt_btcur_insert`), gets a `WT_RESTART`.

As I mentioned in SERVER-34082, I don't think spinning on restart is the problem in that ticket, I think it's something else and the restart counter spike was just a side-effect.

If I had to guess, a split didn't get resolved quickly, but I don't have any ideas on why that might be other than a page-lock wasn't available for some reason?

src/include/misc.i
keithbostic: We sleep for 100, 200, 300 ... 900 usecs, and from then on, 1000 usecs?

The sleep counter value is also fed back into the statistics counter of how long we slept, but I don't see any functionality there.

michaelcahill: Sorry, my mistake -- I conflated `sleep_count` and `yield_count`.

michaelcahill:

PR #4005, opened 2018-03-27
WT-3984 To use barrier instructions to sync prepare state.

bvpvamsikrishna:


bvpvamsikrishna:
@keithbostic 
Request your review.
I have added the barrier instructions to synching prepared state.
Request to check volatile keywork used for prepare_state in update & page_del structure as well.

src/btree/bt_delete.c
```diff
@@ -251,7 +251,7 @@ __wt_delete_page_rollback(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * Now mark the truncate aborted: this must come last because after
 	 * this point there is nothing preventing the page from being evicted.
 	 */
-	ref->page_del->prepare_state = WT_PREPARE_READY;
+	WT_PUBLISH(ref->page_del->prepare_state, WT_PREPARE_READY);
```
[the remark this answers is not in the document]
bvpvamsikrishna: I just discussed with michael and alex, and we are not going to change the prepare state when we abort (i.e. we leave the `prepare_state = WT_PREPARE_STATE`, as an aborted txnid we don't check the state anymore.

src/include/btree.i
```diff
@@ -1158,12 +1158,14 @@ static inline bool
...
 	WT_PAGE_DELETED *page_del;
+	uint8_t prepare_state;
 
 	if ((page_del = ref->page_del) == NULL)
 		return (false);
 	if (page_del->txnid == WT_TXN_ABORTED)
 		return (false);
-	if (page_del->prepare_state != WT_PREPARE_READY)
+	WT_ORDERED_READ(prepare_state, page_del->prepare_state);
+	if (prepare_state != WT_PREPARE_READY)
```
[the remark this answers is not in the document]
bvpvamsikrishna: 1. I will merge the two functions.

2. We re-read the prepare state in `__wt_txn_upd_visible_type` after checking `__wt_txn_visible`, so a torn read of a timestamp is possible, but the result will be ignored. We will also add a new prepare state to indicate the "DONE" state instead of transitioning back to "READY".  It makes it easier to reason about `__wt_txn_upd_visible_type`.

src/include/txn.i
```diff
@@ -581,29 +583,35 @@ __wt_txn_visible_page_deleted(
...
 
 		/*
 		 * The visibility check is only valid if the update does not
-		 * change state.  If the state does change, recheck visibility.
+		 * change its prepared state.  If it does change, recheck
+		 * visibility.
 		 */
-		if (upd->prepare_state == prepare_state)
+		WT_ORDERED_READ(after_prepare_state, upd->prepare_state);
+		if (after_prepare_state == before_prepare_state)
```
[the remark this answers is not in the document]
bvpvamsikrishna: Yes, you are right about read seeing the corrupt timestamp, hence we re-check the prepare state.

src/reconcile/rec_write.c
```diff
@@ -1344,8 +1345,9 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 		 * examining its updates. As prepared transaction id's are
 		 * globally visible, need to check the update state as well.
 		 */
+		WT_ORDERED_READ(prepare_state, upd->prepare_state);
 		if (F_ISSET(r, WT_REC_EVICT) &&
-		    (upd->prepare_state != WT_PREPARE_READY ||
+		    (prepare_state != WT_PREPARE_READY ||
```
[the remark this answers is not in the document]
bvpvamsikrishna: Yes, you are right. Will update.

src/txn/txn.c
```diff
@@ -842,10 +842,12 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
 				 * operation, we will manage using state.
 				 */
-				upd->prepare_state = WT_PREPARE_LOCKED;
+				WT_PUBLISH(upd->prepare_state,
+				    WT_PREPARE_LOCKED);
 				__wt_timestamp_set(
 				    &upd->timestamp, &txn->commit_timestamp);
-				upd->prepare_state = WT_PREPARE_READY;
+				WT_PUBLISH(upd->prepare_state,
+				    WT_PREPARE_READY);
```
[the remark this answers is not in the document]
bvpvamsikrishna: Agreed. We will have a `WT_WRITE_BARRIER` instead of `WT_FULL_BARRIER`.

src/include/btree.i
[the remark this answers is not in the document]
keithbostic: > We re-read the prepare state in `__wt_txn_upd_visible_type` after checking

This bothers me a little, I'm not sure if re-reading the state is our solution to timestamps where torn reads are possible, or just a side-effect of code we're writing anyway. If it's the former, maybe we need a comment?

I thought (but may be wrong), that our plan for torn timestamp reads was the `WT_WITH_TIMESTAMP_READLOCK` macro, so I'm not sure what's happening here.

src/include/txn.i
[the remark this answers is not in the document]
keithbostic: > hence we re-check the prepare state.

Same comment as above: why is this not done with `WT_WITH_TIMESTAMP_READLOCK`?

src/include/btree.i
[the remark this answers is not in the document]
michaelcahill: Yes, FTR the `prepare_state` design was to avoid the readlock in the (extremely common) case of reading an update.

src/btree/bt_delete.c
```diff
@@ -331,9 +328,9 @@ __tombstone_update_alloc(WT_SESSION_IMPL *session,
 	 * timestamp, do nothing.
 	 */
 	if (page_del != NULL) {
-		upd->txnid = page_del->txnid;
-		upd->prepare_state = page_del->prepare_state;
 		__wt_timestamp_set(&upd->timestamp, &page_del->timestamp);
+		upd->prepare_state = page_del->prepare_state;
+		upd->txnid = page_del->txnid;
```
keithbostic: I'm not sure about this change -- I'm guessing this was just a cosmetic change that fell out of some commit, in which case, it's fine with me. If it's more than cosmetic, then I don't understand it.

In general, I like initializing structures in the order the fields are declared simply because it makes it easier to figure out if any fields were forgotten.

src/btree/row_modify.c
```diff
@@ -289,6 +289,7 @@ __wt_update_alloc(WT_SESSION_IMPL *session, const WT_ITEM *value,
 		}
 	}
 	upd->type = (uint8_t)modify_type;
+	upd->prepare_state = WT_PREPARE_INIT;
```
keithbostic: This isn't necessary, `WT_PREPARE_INIT` is deliberately declared 0.

src/include/btmem.h
```diff
@@ -707,31 +707,41 @@ struct __wt_page {
...
+ *	This state has no impact on the visibility of the update's data.
+ *
+ * WT_PREPARE_INPROGRESS:
+ *	Update is in prepared phase.
  *
  * WT_PREPARE_LOCKED:
- *	State is locked as prepare state transition is in progress. Any reader
- *	of the state need to wait for state transition to complete.
+ *	State is locked as state transition is in progress from INPROGRESS to
+ *	RESOLVED. Any reader of the state need to wait for state transition to
```
keithbostic: "need" -> "needs"

src/include/btmem.h
```diff
@@ -707,31 +707,41 @@ struct __wt_page {
...
- * WT_PREPARE_STATE:
- *	State is prepared.
+ * WT_PREPARE_RESOLVED:
+ *	Represents the commit state of the prepared update.
  *
  * State Transition:
- * 	READY --> STATE --> LOCKED --> READY
+ * 	From prepare to commit:
+ * 	INIT --> INPROGRESS --> LOCKED --> RESOLVED
+ * 	LOCKED will be a momentarily phase during timestamp update.
```
keithbostic: "momentarily" -> "momentary", "temporary"

src/include/btmem.h
```diff
@@ -707,31 +707,41 @@ struct __wt_page {
...
+ *
+ * 	From prepare to rollback:
+ * 	INIT --> INPROGRESS
```
keithbostic: Is this still true? I thought we were no longer updating the update's prepare-state during rollback?

src/txn/txn.c
```diff
@@ -904,7 +908,7 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 					    &(*updp)->timestamp,
 					    &txn->commit_timestamp);
 					(*updp)->prepare_state =
-					    WT_PREPARE_READY;
+					    WT_PREPARE_RESOLVED;
```
keithbostic: This code doesn't have an associated `WT_WRITE_BARRIER()/WT_PUBLISH()` pair when setting `WT_PREPARE_LOCKED/WT_PREPARE_RESOLVED`. Is that correct?

bvpvamsikrishna: Yes, this code doesn't have an associated ` WT_WRITE_BARRIER()/WT_PUBLISH()`. As per my understanding `WT_WRITE_BARRIER()/WT_PUBLISH()` is not required as ref is in  `WT_REF_LOCKED` state.  Is my understanding correct?

keithbostic: If that's the case, then there's no need to set the `prepare_state`, basically, it's one or the other, I think.

src/include/btree.i
```diff
@@ -1155,18 +1155,25 @@ __wt_ref_block_free(WT_SESSION_IMPL *session, WT_REF *ref)
...
 		return (false);
-	if (page_del->prepare_state != WT_PREPARE_READY)
+	WT_ORDERED_READ(prepare_state, page_del->prepare_state);
+	if (prepare_state == WT_PREPARE_INPROGRESS ||
+	    prepare_state == WT_PREPARE_LOCKED)
 		return (true);
-	return (!__wt_txn_visible_all(session,
-	    page_del->txnid, WT_TIMESTAMP_NULL(&page_del->timestamp)));
+	return (!(visible_all ?
+	    __wt_txn_visible_all(session,
```
agorrod: Please move the `!` inside the ternary

src/include/btree.i
```diff
@@ -1155,18 +1155,25 @@ __wt_ref_block_free(WT_SESSION_IMPL *session, WT_REF *ref)
  *	Return if a truncate operation is active.
  */
 static inline bool
-__wt_btree_truncate_active(WT_SESSION_IMPL *session, WT_REF *ref)
+__wt_btree_truncate_active(
```
agorrod: Would prefer `__wt_page_del_active`

PR #4008, opened 2018-03-27
WT-3981 Update transaction snapshot when a read_timestamp is set.

michaelcahill:
Without this change, reads could miss committed updates with commit_timestamp lower than the read_timestamp if the commit raced with starting the read transaction.

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -691,18 +691,20 @@ __wt_txn_parse_read_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
 		txn_global = &S2C(session)->txn_global;
 		WT_RET(__wt_txn_parse_timestamp(session, "read", &ts, &cval));
 
-		/* Read timestamps imply / require snapshot isolation. */
+		/*
+		 * Read timestamps imply / require snapshot isolation.
+		 *
+		 * If we already have a snapshot, it may be too early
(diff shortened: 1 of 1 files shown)


src/txn/txn_timestamp.c
```diff
@@ -691,13 +691,14 @@ __wt_txn_parse_read_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
-			WT_RET_MSG(session, EINVAL, "setting a read_timestamp"
-			    " requires a transaction running at snapshot"
-			    " isolation");
+		/*
+		 * Read timestamps imply / require snapshot isolation.
+		 *
+		 * If we already have a snapshot, it may be too early to match
+		 * the timestamp.  Get a new one.
+		 */
+		txn->isolation = WT_ISO_SNAPSHOT;
```
agorrod: Changing the isolation level doesn't feel safe to me. Could this be changed to an error if it isn't already `WT_ISO_SNAPSHOT`? From what I can tell the isolation level will always have been setup by the time we get here.

michaelcahill: OK, I've reverted that part of the change.  I think we may want to reconsider in WT-4000, but let's keep this change simple.

agorrod:
Thanks @michaelcahill lgtm - I'll merge once testing finishes.

PR #4010, opened 2018-03-29
WT-4005 Check that ref->page_del is set before reading from it.

michaelcahill:


michaelcahill:
@keithbostic, I think this one is pretty straightforward: there are paths where `ref != NULL` but `ref->page_del == NULL` that we need to deal with.

I tried to clarify the code rather than creating an even more convoluted nested ternary expression...

PR #4011, opened 2018-03-29
WT-3931 cursor.prev split race

keithbostic:
The bug fix is in 268d1c4.


src/include/btree.i
```diff
@@ -1693,3 +1618,155 @@ __wt_split_descent_race(
...
+#endif
+	    );
+
+	/*
+	 * Expected failures: page not found or restart. Our callers list the
+	 * errors they're expecting to handle.
+	 */
+	if (LF_ISSET(WT_READ_NOTFOUND_OK) && ret == WT_NOTFOUND)
+		return (WT_NOTFOUND);
+	if (LF_ISSET(WT_READ_RESTART_OK)) {
```
agorrod: Is this a proxy for any caller doing a search rather than a next/prev? i.e: callers from search rather than bt_walk? I'd rather the condition where the prev race is checked to be made explicit.

michaelcahill: I think what you mean is that we should always check `prev_race`, and if set (and we're looking at an internal page), call `__wt_split_prev_race`.  If we detect a race but `WT_READ_RESTART_OK` is not passed in, the error case at line 1767 should apply: https://github.com/wiredtiger/wiredtiger/pull/4011/commits/268d1c46d2df6dce5250a57fcab08143259b89bc#diff-b2442811cb1095a2949962a9b1fa9c24R1767

Is that the change you'd like?

michaelcahill: Actually, isn't the logic backwards in that error mapping code?

```
+	/*
+	 * If we're returning an error, don't let it be one our caller expects
+	 * to handle as returned by page-in: the expectation includes the held
+	 * page not having been released, and that's not the case.
+	 */
+	if (LF_ISSET(WT_READ_NOTFOUND_OK) && ret == WT_NOTFOUND)
+		WT_RET_MSG(session,
+		    EINVAL, "page-release WT_NOTFOUND error mapped to EINVAL");
+	if (LF_ISSET(WT_READ_RESTART_OK) && ret == WT_RESTART)
+		WT_RET_MSG(session,
+		    EINVAL, "page-release WT_RESTART error mapped to EINVAL");
```

They should be `if(!LF_ISSET(WT_READ_NOTFOUND_OK) && ret == WT_NOTFOUND) ...` etc. unless I'm missing something?

agorrod: Thanks for stating my concern more clearly Michael.

keithbostic: @michaelcahill, @agorrod, fix pushed in 8ab5a47.

I also tightened down the conditions under which we check for a race, the race is only interesting when walking backwards through the tree.

src/include/btree.i
```diff
@@ -1693,3 +1618,154 @@ __wt_split_descent_race(
...
+}
+
+/*
+ * __wt_page_swap_func --
+ *	Swap one page's hazard pointer for another one when hazard pointer
+ * coupling up/down the tree.
+ */
+static inline int
+__wt_page_swap_func(WT_SESSION_IMPL *session,
+    WT_REF *held, WT_REF *want, bool prev_race, uint32_t flags
```
agorrod: Sorry to give the feedback piecemeal, but could we use `LF_ISSET(WT_READ_PREV)` instead of adding a new parameter?

agorrod:
@keithbostic I'd like for you to go ahead and merge this branch today, whether you decide the suggestion above is OK or not. We are planning to update the version of WiredTiger in MongoDB tomorrow (Sydney time), so it'd be nice to get this change into stress testing as early as we can (I know it's been dawdling primarily because of my slow review - sorry about that).

src/include/btree.i
keithbostic: @agorrod 
> could we use `LF_ISSET(WT_READ_PREV)` instead of adding a new parameter?

It's not just a cosmetic change.

The "additional parameter version" of the change doesn't check for the race when returning internal pages in a tree ascent. (If we are returning internal pages during a walk (that is, when doing eviction), we hazard-pointer couple to the internal parent of the current page in order to return it, and the changes means we'll detect and restart when the internal parent is splitting during that ascent, where we didn't before. I didn't check before, mostly because I was pretty sure the specific race we were chasing couldn't occur in that case. That said, making the change should avoid eviction skipping big blocks of records in some hot trees, which doesn't seem like a bad thing.)

All that said, I like the idea, it's cleaner than what I did. If it passes my testing, I'll go ahead and push it.

Thanks!

agorrod:
Thanks Keith - I like this version.

PR #4012, opened 2018-03-29
WT-4006 format should support stress timing configurations.

keithbostic:


src/conn/conn_api.c
```diff
@@ -1994,9 +1994,13 @@ __wt_timing_stress_config(WT_SESSION_IMPL *session, const char *cfg[])
 {
 	static const WT_NAME_FLAG stress_types[] = {
 		{ "checkpoint_slow",	WT_TIMING_STRESS_CHECKPOINT_SLOW },
```
agorrod: I think splitting the `split_race` conditions up is a great idea - it should make this mode more effective at uncovering problems. I'd like for that to be documented somewhere though - maybe in a comment here?

keithbostic: @agorrod 
> I'd like for that to be documented somewhere though - maybe in a comment here?

I tried to add a comment in 54dddf7, but not entirely sure what you wanted the comment to say. Feel free to just reword to match what you're looking for.

agorrod:
I've switched the timing stress Jenkins job over to this branch, with some script updates. Once that's looking OK I'll merge this.

agorrod:
Thanks to @lukech the Jenkins job that targets these timers is now updated to use the test/format configuration, rather than an environment variable. I'm going to go ahead and merge.

PR #4013, opened 2018-03-29
WT-4007 eviction instantiates pages from dead trees.

keithbostic:
I don't believe a handle can be marked dead while eviction is traversing the tree, but this fix isn't sufficient if I'm wrong (and that's possible).

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -1784,24 +1784,26 @@ __evict_walk_tree(WT_SESSION_IMPL *session,
 	 * eviction fairly visits all pages in trees with a lot of in-cache
 	 * content.
 	 */
-	switch (btree->evict_start_type) {
-	case WT_EVICT_WALK_NEXT:
-		break;
-	case WT_EVICT_WALK_PREV:
-		FLD_SET(walk_flags, WT_READ_PREV);
-		break;
-	case WT_EVICT_WALK_RAND_PREV:
-		FLD_SET(walk_flags, WT_READ_PREV);
-		/* FALLTHROUGH */
-	case WT_EVICT_WALK_RAND_NEXT:
-		if (btree->evict_ref == NULL) {
-			/* Ensure internal pages indexes remain valid */
-			WT_WITH_PAGE_INDEX(session, ret = __wt_random_descent(
-			    session, &btree->evict_ref, true));
-			WT_RET_NOTFOUND_OK(ret);
+	if (!F_ISSET(session->dhandle, WT_DHANDLE_DEAD))
+		switch (btree->evict_start_type) {
+		case WT_EVICT_WALK_NEXT:
+			break;
+		case WT_EVICT_WALK_PREV:
+			FLD_SET(walk_flags, WT_READ_PREV);
+			break;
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, this code should not be instantiating pages in any tree, regardless of whether it is marked dead.  It seems like the fix should be checking `LF_ISSET(WT_READ_CACHE))` for the `WT_REF_DELETED` case in `__wt_page_in_func`. 

Also, while it's not a common case, I think walking a dead tree for eviction should respect the "_PREV" modes of `evict_start_type`.

keithbostic:
@michaelcahill, I went through the same exercise with `WT_READ_CACHE` as you, and I agree.

I didn't think of `WT_READ_NO_WAIT`, but I don't think it's safe.

The code I'm wondering about is in `__rec_child_modify()`, the `WT_REF_MEM` case:
```
/*
 * If called during checkpoint, acquire a hazard pointer
 * so the child isn't evicted, it's an in-memory case.
 *
 * This call cannot return split/restart, we have a lock
 * on the parent which prevents a child page split.
 */
ret = __wt_page_in(session, ref,
    WT_READ_CACHE | WT_READ_NO_EVICT |
    WT_READ_NO_GEN | WT_READ_NO_WAIT);
if (ret == WT_NOTFOUND) {
        ret = 0;
        break;
}
```
If the page transitions into WT_REF_DELETED while we're doing this read, we'd skip it, and we could get into trouble?

That said, I don't understand how this code is safe in its current state.

`WT_READ_NO_WAIT` means we'll skip `WT_REF_LOCKED` pages.

Imagine reconciliation/checkpoint finds a page in the `WT_REF_MEM` state, the page briefly transitions into `WT_REF_LOCKED` (for example, in `__wt_delete_page()`), and then back to `WT_REF_MEM`. I think we could skip that page here, is that safe?

keithbostic:
@michaelcahill 
> Also, while it's not a common case, I think walking a dead tree for eviction should respect the "_PREV" modes of evict_start_type.

Can't we throw away any page we find? Why do we care about fairness?

michaelcahill:
> Can't we throw away any page we find? Why do we care about fairness?

Respecting the eviction walk choice isn't about fairness or throwing pages away: it's the mechanism we have for making sure the eviction server doesn't waste time / get stuck in a bad tree.

michaelcahill:
> I didn't think of WT_READ_NO_WAIT, but I don't think it's safe.

I agree that ` __rec_child_modify` looks unsafe (regardless of the change I'm suggesting).  I've pushed #4020 with the change I think we're talking about: can you please take a look and close this one down if you agree?

Note that `bt_walk.c` has this code, which is the way I think about `WT_READ_NO_WAIT`:

```
470                         if (LF_ISSET(WT_READ_CACHE)) {
471                                 /*
472                                  * Only look at unlocked pages in memory:
473                                  * fast-path some common cases.
474                                  */
475                                 if (LF_ISSET(WT_READ_NO_WAIT) &&
476                                     current_state != WT_REF_MEM &&
477                                     current_state != WT_REF_LIMBO)
478                                         break;
```

PR #4014, opened 2018-03-30
WT-3870 lint

keithbostic:


michaelcahill:

PR #4015, opened 2018-03-30
WT-3971 Cursor duplicates use the cursor cache.

ddanderson:
This fixes a cursor cache leak of sorts when dups of the same URI are repeatedly made.

### DIFF
--- a/src/cursor/cur_std.c
+++ b/src/cursor/cur_std.c
@@ -751,6 +751,7 @@ __wt_cursor_cache_get(WT_SESSION_IMPL *session, const char *uri,
 			return (WT_NOTFOUND);
 	}
 
+	WT_ASSERT(session, uri == NULL || to_dup == NULL);
 	if (to_dup != NULL) {
 		uri = to_dup->uri;
 		hash_value = to_dup->uri_hash;
--- a/src/cursor/cur_std.c
+++ b/src/cursor/cur_std.c
@@ -704,7 +704,7 @@ err:		WT_TRET(cursor->reopen(cursor, false));
  */
 int
 __wt_cursor_cache_get(WT_SESSION_IMPL *session, const char *uri,
-    const char *cfg[], WT_CURSOR **cursorp)
+    WT_CURSOR *to_dup, const char *cfg[], WT_CURSOR **cursorp)
 {
 	WT_CONFIG_ITEM cval;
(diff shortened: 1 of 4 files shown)


ddanderson:
@sueloverso, if you're available, can you please review?

test/suite/test_cursor13.py
```diff
@@ -538,6 +538,7 @@ def test_dup(self):
         cursor.close()
 
         c1 = self.session.open_cursor(uri, None)
+        c1.next()
```
sueloverso: Why do we need this line? If it is necessary please add a comment.

src/cursor/cur_std.c
```diff
@@ -751,11 +751,16 @@ __wt_cursor_cache_get(WT_SESSION_IMPL *session, const char *uri,
 			return (WT_NOTFOUND);
 	}
 
+	if (to_dup != NULL) {
+		uri = to_dup->uri;
+		hash_value = to_dup->uri_hash;
```
sueloverso: I'd add a comment that we expect the caller to check that one of `uri` or `to_dup` is NULL.

test/suite/test_cursor13.py
```diff
@@ -527,3 +527,22 @@ def test_cursor_sweep(self):
...
+    def test_dup(self):
+        self.cursor_stats_init()
+        uri = 'table:test_cursor13_dup'
+        self.session.create(uri, 'key_format=S,value_format=S')
+        cursor = self.session.open_cursor(uri)
+        cursor['A'] = 'B'
+        cursor.close()
+
+        c1 = self.session.open_cursor(uri, None)
+        c1.next()
```
sueloverso: It took a moment to understand the `next` call. Add a short comment that we need to position the cursor on the data before dup'ing it.

test/suite/test_cursor13.py
```diff
@@ -527,3 +527,22 @@ def test_cursor_sweep(self):
...
+        for notused in range(0, 100):
+            self.session.breakpoint()
+            c2 = self.session.open_cursor(None, c1, None)
+            c2.close()
+        stats = self.caching_stats()
+        self.assertGreaterEqual(stats[0], 100)
```
sueloverso: A comment about what stats we're checking here would be helpful.

agorrod:
lgtm

PR #4016, opened 2018-03-30
WT-4008 Add ARM NEON support for row search operations

stewart-daniel:
This is a contribution for adding AArch64 FTS fast byte vector support & NEON support for row search operations.


wtbuild:

PR #4019, opened 2018-04-01
WT-3724 Log an error if flushing with F_FULLSYNC fails

keithbostic:


michaelcahill:
If this is the whole change being proposed here, can we please make the commit title something like "Log an error if flushing with F_FULLSYNC fails", rather than the current title which makes it seem like this fixes a bug?

agorrod:
Thanks @keithbostic lgtm - I agree that the ticket/merge message title should be changed. Please go ahead and merge when you are ready (and JIRA is up)

PR #4021, opened 2018-04-03
WT-4009 Create fast path for cursor caching with "overwrite=false"

ddanderson:


ddanderson:
@agorrod, can you review? This change is a little tightly coupled with MongoDB code, but it does show some modest improvements for the capped collections (it was included in my recent perf runs).

agorrod:
Thanks @ddanderson lgtm

PR #4023, opened 2018-04-04
WT-4011 Checkpoint should not read truncated pages.

michaelcahill:


agorrod:
Thanks @michaelcahill this change makes sense to me, and I think the flag naming is clearer now. I'll go ahead and merge.

PR #4024, opened 2018-04-04
WT-4012 Fix lookaside entry count.

michaelcahill:
We now "pre-allocate" the lookaside entry count by incrementing it by 100 as records are inserted.  At the end, we need to fix up the final count to match the number that were actually inserted.  That is always a decrement.


agorrod:
I don't love ternary statements nested inside statements, but this change looks correct to me. I'll go ahead and merge once testing finishes.

PR #4027, opened 2018-04-04
WT-4012 Fix lookaside entry count.

agorrod:
We now "pre-allocate" the lookaside entry count by incrementing it by
100 as records are inserted.  At the end, we need to fix up the final
count to match the number that were actually inserted.  That is always a
decrement.

I merged this once and it trigger a stress [test failure](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress/65747/), so I reverted and re-created the PR based on the original branch.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -113,7 +113,7 @@ __las_page_instantiate_verbose(WT_SESSION_IMPL *session, uint64_t las_pageid)
  *	Instantiate lookaside update records in a recently read page.
  */
 static int
-__las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t btree_id)
+__las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 {
 	WT_CACHE *cache;
 	WT_CURSOR *cursor;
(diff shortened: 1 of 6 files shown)


michaelcahill:
@keithbostic, this change seems fine, but did you reproduce the failure and confirm that this change fixes it?

I'm confused if it does because `decrement_cnt` should always be zero.  The call to `__las_remove_block` should always be a noop now that lookaside page IDs are globally unique.

keithbostic:
@michaelcahill 
I did not reproduce the failure: I agreed with your change to switch the code to always decrement, but assumed the problem was keeping `decrement_cnt` in the calculation even if an error had occurred.

I'm a little confused: why are we calling `__las_remove_block` and looking at `decrement_cnt` at all, if it's a no-op?

agorrod:
> why are we calling __las_remove_block and looking at decrement_cnt at all, if it's a no-op?

That's an artefact of previous incarnations of lookaside, that I'm not brave enough to remove right now.

michaelcahill:
FTR, I pushed some changes in aeef510 that are cleanups I found while chasing the failure.  I don't have any reason to believe they fix anything, and I'm happy to split them out to a separate cleanup ticket once the bug is fixed.

agorrod:
sgtm - and lgtm, I'll go ahead and merge.

PR #4032, opened 2018-04-05
WT-4022 Avoid WT_RESTART error return during eviction walk

keithbostic:


keithbostic:
@agorrod, I split WT-4022 out of WT-4017, this is the fix for WT-4022.

agorrod:
Thanks @keithbostic this lgtm - I'd like to revisit the way tree walk flags work in the future, I'll open a new ticket for that.

PR #4033, opened 2018-04-05
WT-4023 naked error returns

keithbostic:
@sueloverso, I was chasing a `WT_ERROR` return today that didn't have an associated message, and so I went looking. I'm asking you for the review because `log.c` is involved.

Thanks!

### DIFF
--- a/src/bloom/bloom.c
+++ b/src/bloom/bloom.c
@@ -268,6 +268,7 @@ __wt_bloom_hash_get(WT_BLOOM *bloom, WT_BLOOM_HASH *bhash)
 	uint64_t h1, h2;
 	uint32_t i;
 	uint8_t bit;
+	int result;
 
 	/* Get operations are only supported by finalized bloom filters. */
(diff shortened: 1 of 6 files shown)


keithbostic:
autobuild: retest this please

src/bloom/bloom.c
```diff
@@ -292,20 +290,20 @@ __wt_bloom_hash_get(WT_BLOOM *bloom, WT_BLOOM_HASH *bhash)
 		WT_ERR(c->get_value(c, &bit));
 
 		if (bit == 0) {
-			result = WT_NOTFOUND;
+			ret = WT_NOTFOUND;
 			break;
 		}
 	}
 	WT_ERR(c->reset(c));
```
agorrod: This call to `WT_ERR` is going to overwrite the saved `ret` value now, isn't it?

keithbostic: @agorrod 
> This call to `WT_ERR` is going to overwrite the saved `ret` value now, isn't it?

Yes, you are correct -- I apologize for the noisy change.

PR #4035, opened 2018-04-06
WT-4017 When evicting during a checkpoint, avoid splits.

michaelcahill:
If eviction has to pull pages into cache (generally because they have lookaside history that is relevant to the checkpoint), we try to evict those pages immediately, writing the correct version to disk in the process.

However, if the eviction causes the page to split (which can happen because a different set of data is visible to the checkpoint than was visible when the page was previously evicted and the lookaside entries written), then splitting up the tree could invalidate the internal page image already calculated by the checkpoint.

The solution taken here is to give up on eviction if it would cause a page to split during a checkpoint.  The pages will stay in cache a little longer (though this should be rare in real workloads), but a valid snapshot of data will be written to disk.


agorrod:
lgtm - @keithbostic please review as well.

PR #4036, opened 2018-04-06
WT-4014 If eviction walk is interrupted, clean up the queue.

michaelcahill:
Later walks assume there is a clear range of the queue and interrupts could violate that assumption.


agorrod:
Thanks @michaelcahill this change looks OK to me. I was a bit worried that something holding the evict lock might wait for an interrupt to be processed (thus would block), but I couldn't see any cases where that would be true, and I think it would fail quickly if it was a problem - I'm going to go ahead and merge.

PR #4037, opened 2018-04-06
WT-3910 Fix evergreen build on macOS.

michaelcahill:
In particular, force the use of Python from mongodbtoolchain/v2 and fix up the libwiredtiger.so path in binaries.


agorrod:
Lgtm

lukech:
lgtm.

PR #4038, opened 2018-04-06
WT-3959 Manage recovery timestamp via a new schema entry.

sueloverso:
Not ready for review.

### DIFF
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1055,9 +1055,10 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
 	if (cval.val != 0)
 		F_SET(conn, WT_CONN_LEAK_MEMORY);
 	WT_TRET(__wt_config_gets(session, cfg, "use_timestamp", &cval));
-	if (cval.val != 0 && conn->txn_global.has_stable_timestamp) {
+	if (cval.val != 0) {
 		ckpt_cfg = "use_timestamp=true";
-		F_SET(conn, WT_CONN_CLOSING_TIMESTAMP);
+		if (conn->txn_global.has_stable_timestamp)
+			F_SET(conn, WT_CONN_CLOSING_TIMESTAMP);
 	}
 
 err:	/*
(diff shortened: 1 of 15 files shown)


src/conn/conn_api.c
```diff
@@ -1055,9 +1055,10 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
 	if (cval.val != 0)
 		F_SET(conn, WT_CONN_LEAK_MEMORY);
 	WT_TRET(__wt_config_gets(session, cfg, "use_timestamp", &cval));
-	if (cval.val != 0 && conn->txn_global.has_stable_timestamp) {
+	if (cval.val != 0) {
 		ckpt_cfg = "use_timestamp=true";
-		F_SET(conn, WT_CONN_CLOSING_TIMESTAMP);
+		if (conn->txn_global.has_stable_timestamp)
+			F_SET(conn, WT_CONN_CLOSING_TIMESTAMP);
 	}
```
michaelcahill: I'm slightly confused by this change -- what does it mean to set `use_timestamp=true` if there is no `stable_timestamp`?

src/include/meta.h
```diff
@@ -27,6 +27,7 @@
 #define	WT_METAFILE_URI		"file:WiredTiger.wt"	/* Metadata table URI */
 
 #define	WT_LAS_URI		"file:WiredTigerLAS.wt"	/* Lookaside table URI*/
+#define	WT_SYSTEM_URI		"system:info"		/* System info URI*/
```
michaelcahill: How about "system:checkpoint"?  If we want more "system" information in the metadata, I'd be inclined to keep it in separate records so we don't have to synchronize writes from different operation.

sueloverso:
@michaelcahill You might take another quick look. I added a new test to make sure that the timestamp information survives a backup (i.e. in the `WiredTiger.backup` file) but is not part of the filelist that backup generates.

src/conn/conn_api.c
michaelcahill: Okay, I understand the intent -- as we've discussed previously, I feel like we're twisting the checkpoint interface here in a way I find counterintuitive (it would make more sense to me to preserve the last timestamp at which a checkpoint completed and have MongoDB explicitly clear that via some version of `WT_CONNECTION::set_timestamp`).

That's not the subject of this change, though, and something we could revisit in future.

PR #4039, opened 2018-04-08
WT-4028 page-is-modified check requires the page be locked

keithbostic:
Revert part of 32793d9: it's not safe to call page-is-modified without first locking the WT_REF, the page can be evicted out from underneath.


agorrod:
Thanks @keithbostic lgtm

PR #4040, opened 2018-04-10
WT-4032 parent pages can be evicted while being split.

keithbostic:


agorrod:
@keithbostic I see the problem this change is solving - do you think it matters if failed splits mark pages as dirty now?

keithbostic:
> do you think it matters if failed splits mark pages as dirty now?

I thought about that (note, it's worth considering these are internal, not leaf, pages, and we're marking the page dirty before incrementing/decrementing the memory footprint), but I couldn't think of any reason it could make a difference.

Splits should pretty much never fail, so I'm confident there's no performance cost.

There might be a simplification where we mark the page dirty when we acquire a lock on it (so there are fewer places we mark pages dirty), but there are a bunch of different paths through the split code, if there's a simpler approach, I didn't see it.

cc: @michaelcahill

michaelcahill:
Thanks @keithbostic, this lgtm -- I agree that marking internal pages should be fine: we'll have to write them as part of the next checkpoint regardless.  I'll merge.

PR #4041, opened 2018-04-11
WT-4034 re-entering eviction can result in checkpoint corruption

keithbostic:


michaelcahill:
Thanks @keithbostic, nice catch!  This could be responsible for some other eviction-racing-with-checkpoint issues we have seen recently.  I also like the more general restricting on accidentally entering a newer generation.

I'll merge.

PR #4042, opened 2018-04-12
Revert "WT-3959 Manage recovery timestamp via a new schema entry."

agorrod:
Reverts wiredtiger/wiredtiger#4038


agorrod:
The original commit introduced test failures in a recovery test, so I'm going to revert for now.

PR #4043, opened 2018-04-12
WT-3959 Manage recovery timestamp via a new schema entry

agorrod:
This branch was merged once, but introduced [test failures](http://build.wiredtiger.com:8080/job/wiredtiger-test-spinlock/4246/), so I reverted.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -351,8 +351,6 @@ file_meta = file_config + [
         the file checkpoint entries'''),
     Config('checkpoint_lsn', '', r'''
         LSN of the last checkpoint'''),
-    Config('checkpoint_timestamp', '', r'''
-        stable timestamp of the last checkpoint'''),
     Config('id', '', r'''
         the file's ID number'''),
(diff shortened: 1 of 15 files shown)


agorrod:
I've reproduced the failure once in about 30 minutes running 4 reproducers in parallel. The configuration I was using is:

```
../configure --enable-python --with-spinlock=pthread_adaptive --enable-strict && make -j12
export WIREDTIGER_CONFIG="checkpoint_sync=0,transaction_sync=(method=none)"
for i in `seq 100`; do ./test_timestamp_abort -h WT_TEST.2 -T 5 -t 20 || break ; done
```

i.e: a slightly longer runtime than the configuration that failed in Jenkins.

agorrod:
It reproduced much faster running 20 in parallel:

```
#!/bin/bash
for j in `seq 100`; do
       	for i in `seq 20`; do
	       	./test_timestamp_abort -h WT_TEST.$i -T 5 -t 10 || break > run.$i.out &
       	done
       	wait
done
```

PR #4044, opened 2018-04-12
WT-3959 Read the recovery timestamp after rolling forward the metadata.

michaelcahill:


agorrod:
I've moved the spinlock test that was failing on the base branch here. I haven't been able to reproduce the failure on this branch in my testing.

agorrod:
@sueloverso would you mind reviewing this change which is built on top of your original one? If you are happy please merge this into your branch, then your branch into develop.

PR #4045, opened 2018-04-12
WT-4027 cursors should yield the CPU on repeated restart

keithbostic:


agorrod:
Thanks @keithbostic lgtm

PR #4046, opened 2018-04-12
WT-4036 Coverity tickets: #1129045, #1129046, #1382093 out-of-bounds access

keithbostic:
Alex, I've been talking with the Coverity folks about getting rid of this false positive -- it's pretty common because we have so many code paths that go through this function. Hopefully this will do the trick!


michaelcahill:

PR #4047, opened 2018-04-13
WT-3959 Remove metadata entry for checkpoint timestamp if zero.

michaelcahill:
We added a "system:checkpoint" entry to the metadata to store the most recent timestamp at which a checkpoint completed.  Current versions of WiredTiger treat a missing entry as equivalent to a zero timestamp.

However, this metadata entry can confuse older versions of WiredTiger (specifically, opening a backup cursor fails when the new entry exists). Since MongoDB will always checkpoint without a timestamp as part of the downgrade process, remove the metadata entry in that case so that older versions of WiredTiger never see it.

Also, check for the "system:" previous generally rather than this specific entry.  Then if future versions of WiredTiger introduce new entries, they will be silently skipped when not expected.

### DIFF
--- a/src/meta/meta_ckpt.c
+++ b/src/meta/meta_ckpt.c
@@ -502,12 +502,8 @@ __wt_meta_sysinfo_set(WT_SESSION_IMPL *session)
 {
 	WT_DECL_ITEM(buf);
 	WT_DECL_RET;
-	const char *cfg[3];
-	char *config, *newcfg;
 	char hex_timestamp[2 * WT_TIMESTAMP_SIZE + 2];
-	bool update;
 
-	config = newcfg = NULL;
 	WT_ERR(__wt_scr_alloc(session, 0, &buf));
 	hex_timestamp[0] = '0';
 	hex_timestamp[1] = '\0';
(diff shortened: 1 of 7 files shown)


src/meta/meta_ckpt.c
```diff
@@ -520,32 +516,23 @@ __wt_meta_sysinfo_set(WT_SESSION_IMPL *session)
...
-		cfg[0] = config;
-		cfg[1] = buf->mem;
-	} else if (ret == WT_NOTFOUND) {
-		update = false;
-		cfg[0] = buf->mem;
-		cfg[1] = NULL;
-	} else
-		WT_ERR(ret);
-	/* Replace or insert the system info entry. */
-	WT_ERR(__wt_config_collapse(session, cfg, &newcfg));
```
michaelcahill: The reason I removed this is that it (a) has no effect today, since there's only ever a single field in the configuration ("checkpoint_timestamp"), and (b) wasn't future proof, because `__wt_config_collapse` expects the first entry in the stack to contain the full set of permitted values.

agorrod:
It'd be nice if `system:` was a #define, so it easier to find.

agorrod:
Thanks @michaelcahill this change makes sense to me, and is a definite improvement - I'm going to go ahead and merge.

PR #4049, opened 2018-04-14
WT-4031 on-page zero-length row-store values can be discarded from checkpoints

keithbostic:


keithbostic:
One review comment: I don't like `__wt_cell_unpack_empty_value()` much, but I don't see an easy way to tease apart the calls into `__rec_txn_read()` and subsequently `__rec_append_orig_value()` so there'd be a way to pass down information that indicates `vpack` is NULL but there's still an associated value. If anybody sees something better, that'd be great. I thought about adding information to `vpack` itself, but that didn't seem any better.

michaelcahill:
@keithbostic, thanks for the fix: nice catch!

I'd be inclined to go the other way and always "unpack" a missing value cell as if it was a zero-length value.  If `__wt_row_leaf_value_cell` returned an unpacked value, rather than a pointer to the raw cell, that would be straightforward (AFAICT, all callers currently call `__wt_cell_unpack` right after).

Alternatively, we could keep around the bits corresponding to a zero-length value cell in a static array, and have `__wt_row_leaf_value_cell` return a pointer to that, but that seems messier to me.

This change would clean up various callers like `__cursor_row_slot_return` and `__value_return` that currently have to check for NULL and hard-code knowledge that NULL means an empty value in multiple parts of the code.

I've pushed a small step in that direction in 40ec4a929, which makes `__rec_row_leaf` more consistent with other reconciliation functions by having `vpack = &vpack` throughout.  Feel free to discard it if you disagree or I've misunderstood something.

keithbostic:
@michaelcahill, I like your suggestion a lot, but I think it's a relatively risky change. I'm going to merge this simpler fix, and then  we can look at the bigger change in a different ticket.

PR #4050, opened 2018-04-14
WT-4019 reduce prepare and snapshot-isolation interactions in format

keithbostic:


agorrod:
Thanks @keithbostic lgtm.

PR #4052, opened 2018-04-16
WT-4039 Move row-store missing-value support into the cell unpack code.

keithbostic:
@michaelcahill, I like your idea a lot, but the actual mechanics of it make me nervous.

I split this work into a different ticket from WT-4031 (#4049) because we don't need to backport this work to fix the bug, and second, I think we want to pound on this change hard before we let it out into the wild.


agorrod:
This seems like a really nice simplification. I'm inclined to take it once WT-4037 has had a chance to stress test for a while.

PR #4055, opened 2018-04-18
WT-4037 WT_REF structures freed while still in use

keithbostic:


keithbostic:
I've pushed some discussion into the ticket, and I think this is ready for review. There are two separate bug fixes.

One alternative to 62b5311 might be to acquire the session's split generation after acquiring the page lock, but I don't have any evidence there's a big performance advantage, and it's somewhat more complicated.

The 0272f3a change is potentially a performance issue, as splitting an internal page now prevents the moved child pages from themselves splitting until the internal page split completes.

One alternative might be to leave things as before and then update split-stash generations after the internal page split completes. The internal page split is holding a low split generation, preventing any WT_REF structures stashed by the problematic child page splits from being discarded, we know any stashed WT_REF structures aren't moving. We could walk the list of sessions, find any stashed objects with problematic generation values, and update those generation values to something after the internal page split completed. We could cull that list pretty  hard: tag the stashed objects with their type (or use the size), use a lower bound of the split generation when the internal page split started, and so on. I actually started in that direction, but it means we'd need some kind of locking on top of the session stash, because now we'd have threads accessing other thread's session stash structures. And, tbh, it's unclear that wouldn't introduce its own set of performance problems.

agorrod:
Thanks @keithbostic well done finding these, and thanks for the clear explanations.

I'm not a huge fan of passing around a reference to an array of pointers (`***`), but I don't see an obviously better solution. I'll go ahead and merge.

@michaelcahill please come back to this and review post merge.

PR #4057, opened 2018-04-18
WT-4044 add API to return if a generation is active

keithbostic:
Alex, feel free to discard without further discussion if you don't like it.


agorrod:
Thanks @keithbostic - this change looks good to me, and I agree it should be more efficient. I can't think of any problem with it, but I'm going to hold off merging until after WT-4037 has had some testing.

PR #4058, opened 2018-04-18
WT-3937 LAS sweep tuning

sueloverso:
Not yet ready for review. First draft of separating LAS sweep timing from dhandle sweep timing. Dynamic adjusting of number of entries per pass is still in progress.


lukech:
@autobuild: rebuild this please

lukech:
autobuild: retest this please

src/cache/cache_las.c
```diff
@@ -1078,12 +1077,16 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+		 * If we have processed enough entries and there is a reader
+		 * wanting to get the lock or we are between blocks, give up.
 		 */
-		if (cnt > 0)
+		++visit_cnt;
+		if (cnt > 0) {
 			--cnt;
-		else if (saved_key->size == 0)
+			if (visit_cnt > WT_LAS_SWEEP_ENTRIES &&
+			    cache->las_reader)
```
agorrod: Even with this condition (which i like), it's still potentially going to do twice as much work as before between relinquishing the lock. Do you think it makes sense to add another layer that releases the lock every ~1000 ops if there are readers waiting, but doesn't go back to sleep until it's done `cnt` things?

sueloverso: First, I'm happy to set the entries back to 10000 if you want. I think it was fairly arbitrarily chosen and 20000 is more but doesn't feel like too much either. I'm hesitant to keep it in there until `cnt` things is complete if it is getting interrupted or slowed down because sweep does need to get back to the caller and do the other sweep things. And another layer is just more complex.

sueloverso: Also, I don't feel like this is quite right yet. The current algorithm only breaks on a keyspace boundary allowing the count to stay zero while we process up to the boundary. I think the new code should do the same. I think other parts of the loop expect stopping on the boundary.

sueloverso:
@agorrod This is ready. The only open question I think is what the value of `WT_LAS_SWEEP_ENTRIES` should be.

src/cache/cache_las.c
sueloverso: I think I fixed the boundary issue.

agorrod:
@sueloverso thanks - this change is looking good. I have a slight reservation about doubling the minimum number of entries reviewed. I've asked Bruce to take a look in the JIRA ticket - if he is happy so am I.

michaelcahill:
Thanks @sueloverso, I'm happy with these changes: we'll do some more performance testing between here and MongoDB 4.0, but this change makes it easy to try different settings, so let see how far this gets us.

PR #4060, opened 2018-04-19
WT-3851 Optimize wt_compare* routines with NEON instructions for ARM platform

keithbostic:


keithbostic:
autobuild: retest this please

PR #4062, opened 2018-04-20
WT-4039 Move row-store missing-value support into the cell unpack code.

agorrod:
Change __wt_row_leaf_value_cell() to return an unpacked cell, where the
cell can potentially representing a missing row-store value, instead of
returning a pointer to the cell (which might be NULL if there was no value
cell), then subsequently handling the NULL cell explicitly or unpacking
the non-NULL cell.

This requires that the standard cell unpacking code handling a NULL cell
pointer (__wt_cell_unpack() can now handle a NULL cell reference and
create an "empty" unpacked cell), plus the __cell_data_ref() function
has to handle the empty unpacked cell created by __wt_cell_unpack().
The latter was already the case, but there's a minor change there so we
don't attempt to Huffman decode empty values.

This means a slightly slower code path in general, but simplifies some
higher-level code.


keithbostic:
@michaelcahill, can you please review this change when you return?
@agorrod, I think this can be merged (again!), but there's some tricky stuff in here, it needs good testing before a master drop.

agorrod:
I've switched the stress sanitizer branch to run on this branch:

agorrod:
This lgtm, and testing seems stable now.

PR #4063, opened 2018-04-20
WT-4039 Revert change to encapsulate NULL into cell decoding.

agorrod:


agorrod:
Reverting due to test failure in salvage:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format/6383/console

PR #4064, opened 2018-04-20
WT-3914 Add general documentation for prepare transaction.

agorrod:
prepare_doc

Attaching a screenshot showing the generated documentation.

@sueloverso Would you please review? I could definitely be convinced to move some/all the content into either the timestamps documentation on the same page, or the API doc in wiredtiger.in.

### DIFF
--- a/src/docs/transactions.dox
+++ b/src/docs/transactions.dox
@@ -38,6 +38,21 @@ transaction and their effects committed by calling
 WT_SESSION::commit_transaction, or discarded by calling
 WT_SESSION::rollback_transaction.
 
+Applications that use @ref transaction_timestamps can utilize the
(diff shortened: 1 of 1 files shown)


sueloverso:
@agorrod I think the text itself is fine. I have been looking at the full page in question. I would suggest either moving the entire paragraph into the timestamp section, or if you really feel it needs to be mentioned in the API section, only have some form of the first sentence in there, guiding the reader to the timestamp section. I'd put that sentence at the end of the section, though, not the middle. Most readers will want to know about commit/rollback or the whole timestamp section.

michaelcahill:
Thanks @sueloverso and @agorrod -- lgtm.  I'll merge.

PR #4065, opened 2018-04-23
WT-4052 Transaction resources should be freed on session reset.

ddanderson:


ddanderson:
@agorrod, can you review this please?

agorrod:
Thanks @ddanderson this change lgtm and I can understand how it solves the slow growth in memory reported as the symptom.

PR #4066, opened 2018-04-23
WT-4029 Bump log file version.

sueloverso:
Not yet ready for review.


sueloverso:
@agorrod this is ready for review. @dgottlieb and @ddanderson, FYI.

Here are things of note:
* The actual code changes are quite small given that there are no actual file content changes, only getting the new version written in the log file header.
* I've updated the `test_compat01.py` to deal with multiple log versions containing the previous LSN record. That required a lot of changes actually.
* I've manually tested this with a current develop branch to make sure it detects the newer log file version. But we don't have any good way to test this automatically.
* I bumped the WiredTiger version to 3.1 and modified the compatibility release string used in the test accordingly. We haven't yet decided on an official release number but if it is not 3.1 then it needs to be changed in a couple places, including `conn_log.c` which assumes V2 and V3 have the same major release number but different minor release numbers.

agorrod:
@sueloverso should I go ahead and merge this change?

sueloverso:
This is ready for review and merge. I set up a meeting next week when in NYC with interested parties to discuss this for going forward. But this PR should not wait for that.

agorrod:
Thanks @sue lgtm. I'll go ahead and merge.

@lukech be aware that this change alters WiredTiger version numbers - we should probably cut a WiredTiger release soon(ish) to capture this.

PR #4069, opened 2018-04-25
WT-4056 Add log file version statistic

sueloverso:
@agorrod Here is one fairly easy way to get the information needed by MongoDB to determine a database is an illegal 3.4-based without corrupting the database, if opened read-only.  I did run a test on a 3.4-based database directory, opening it read-only, grabbing the version and checking it, and then running `wt list -v` on the 3.4 directory after to make sure it is still usable. Again, there are no good ways to test it other than manually.

Also note that this branch is based off the WT-4029 branch.

@dgottlieb FYI.

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -367,6 +367,7 @@ connection_stats = [
     LogStat('log_compress_small', 'log records too small to compress'),
     LogStat('log_compress_write_fails', 'log records not compressed'),
     LogStat('log_compress_writes', 'log records compressed'),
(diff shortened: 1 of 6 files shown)


michaelcahill:
@sueloverso, it would make me sad for MongoDB to have to know about WiredTiger internals like this.  MongoDB already has a mapping for our "compatibility=" setting in `wiredtiger_open` -- I'll follow up in the ticket with a suggestion on using that instead of querying the log file version.

agorrod:
@sueloverso should we close this pull request in preference for the new approach outlined in WT-4056?

PR #4070, opened 2018-04-26
WT-4057 Update snapshot after handling round_to_oldest.

michaelcahill:


michaelcahill:
Note: these changes need a test case -- since we were already bumping the snapshot during `WT_SESSION::timestamp_transaction` but racing with an update of the `oldest_timestamp`, it isn't straightforward to trigger this from Python.

sueloverso:
@michaelcahill is this something that `timestamp_abort` could trigger and/or detect? That test program is very careful about how it updates the `oldest_timestamp`. Also that test does not use `timestamp_transaction` but always sets `commit_timestamp` on the `commit_transaction` call. Is that helpful in triggering the race?

agorrod:
Thanks @michaelcahill - that looks like a small window, but I see how the fix works.

I'm going to go ahead and merge to get this into testing. It would be good if you could open a new ticket to track adding testing for race conditions in this code.

PR #4074, opened 2018-04-26
WT-4058 Release a slot outside of the lock.

sueloverso:
@agorrod Please review this change. After the user opened the ticket and questioned this, I believe this is safe. I have been running on this branch for a few hours in both the `wiredtiger-test-recovery-stress` and `wiredtiger-perf` jobs and have not seen any problems.

Other uses of `wt_log_release` do not hold any locks so that function does not require it.

### DIFF
--- a/src/log/log_slot.c
+++ b/src/log/log_slot.c
@@ -297,7 +297,7 @@ __log_slot_switch_internal(
 	WT_LOG *log;
 	WT_LOGSLOT *slot;
 	uint32_t joined;
-	bool free_slot, release;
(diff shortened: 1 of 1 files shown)


sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

agorrod:
Some performance testing shows a regression due to this change - we should hold off merging for now.

agorrod:
The performance regression was an anomaly. I'm going to go ahead and merge.

PR #4075, opened 2018-04-26
WT-4063 Explicitly state log archiving is disabled. Fix a typo.

sueloverso:
@agorrod Please review these minor changes to the docs to be clearer when log archiving is disabled. This arose from confusion and a slack chat today.


agorrod:
Thanks @sueloverso lgtm

PR #4076, opened 2018-04-27
WT-4016 Change lookaside skew heuristic.

michaelcahill:
This heuristic attempts to ensure that most pages evicted with lookaside history get the versions on-disk that will be chosen by the next checkpoint.  Even if a stable_timestamp is set, under normal circumstances it will be moving forward regularly, so the next checkpoint is likely to choose the most recent committed versions to write to the backing file.

Only skew towards choosing the oldest versions of records if a page is evicted twice without the pinned timestamp moving forward.


agorrod:
@michaelcahill this change lgtm - should we run perf tests before merging, or is it good to go?

PR #4077, opened 2018-04-27
WT-4064 Relax checking of lookaside entry count.

michaelcahill:
If there is a race between inserting and removing lookaside records (or some bug where the entry count is not updated correctly), treat underflow the same as a zero entry count (i.e., treat the lookaside table as empty).

At worst, this will leave some data in the lookaside table that could potentially be swept.  It should avoid assertion failures during tests while we search for the source of the underflow.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -42,17 +42,29 @@ __las_restore_isolation(
 }
 
 /*
- * __wt_las_nonempty --
+ * __wt_las_entry_count --
+ *	Return when there are entries in the lookaside table.
+ */
+uint64_t
+__wt_las_entry_count(WT_CACHE *cache)
+{
+	uint64_t insert_cnt, remove_cnt;
+
+	insert_cnt = cache->las_insert_count;
+	remove_cnt = cache->las_remove_count;
+	WT_READ_BARRIER();
+
+	return (insert_cnt > remove_cnt ? insert_cnt - remove_cnt : 0);
+}
+
+/*
+ * __wt_las_empty --
  *	Return when there are entries in the lookaside table.
  */
 bool
-__wt_las_nonempty(WT_SESSION_IMPL *session)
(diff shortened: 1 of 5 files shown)


src/cache/cache_las.c
```diff
@@ -726,24 +739,15 @@ __wt_las_insert_block(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
...
-		} else
+		if (ret == 0)
+			WT_TRET(__wt_txn_commit(las_session, NULL));
+		else
 			WT_TRET(__wt_txn_rollback(las_session, NULL));
+
+		/* Adjust the entry count. */
+		if (ret == 0)
+			__wt_atomic_add64(
+			    &conn->cache->las_insert_count, insert_cnt);
```
keithbostic: `__wt_atomic_add64()` returns a value so needs a void cast (there are a couple of other instances, too).

src/cache/cache_las.c
```diff
@@ -908,8 +910,8 @@ __las_sweep_count(WT_CACHE *cache)
 	 * with lookaside entries are blocked during sweep, make sure we do
 	 * some work but don't block reads for too long.
 	 */
-	return ((uint64_t)WT_MAX(100,
-	    cache->las_entry_count / (WT_MINUTE * 5 / WT_LAS_SWEEP_SEC)));
+	return ((uint64_t)WT_MAX(100, __wt_las_entry_count(cache) /
+	     (5 * WT_MINUTE / WT_LAS_SWEEP_SEC)));
```
keithbostic: `WT_MAX` evaluates its arguments twice, so `__wt_las_entry_count()` is potentially called twice. It's not obvious to me that's a problem, but I'd suggest fixing it to remove any question.

michaelcahill:
Thanks @keithbostic, all done in 6635e46.

michaelcahill:
I got offline approval from @agorrod to merge, given that we're seeing regular test failures.

PR #4078, opened 2018-04-27
WT-4061 never rollback transactions during recovery

keithbostic:


keithbostic:
@agorrod, this was a `wt` command failure, if we're seeing eviction pressure in `wt` during recovery,  maybe we should increase the `wt` command cache size, or even the default cache size?

agorrod:
@keithbostic Your question above is similar to what I thought: "How can recovery pin 100MB of content into cache?" I guess that's 20MB of dirty content with our default settings, and a single thread pinning 20MB of content seems high, but not crazy - did something change to make us hit the hard limit though?

keithbostic:
@agorrod 
> "How can recovery pin 100MB of content into cache?"

Since this is a truncate operation, it could theoretically pin an entire table in memory?

michaelcahill:
I'm inclined to take this change then wait to see if recovery gets stuck somewhere.  There may have been some blip in I/O that caused eviction to go slow for a while and mark the cache stuck temporarily...

keithbostic:
@michaelcahill 
> then wait to see if recovery gets stuck somewhere.

FTR, I ran the test a lot after putting in this fix, testing for that case, and never saw a failure -- I believe it was likely a blip.

PR #4079, opened 2018-04-27
WT-4051 format configures too-small LSM caches.

keithbostic:


test/format/config.c
```diff
@@ -135,14 +137,20 @@ config_setup(void)
 			continue;
 
 		/*
-		 * Boolean flags are 0 or 1, but only set N in 100 where the
-		 * variable's min value is N.  Set the flag if we rolled >=
-		 * the min, 0 otherwise.
+		 * Boolean flags are 0 or 1, where the variables "min" value
+		 * is the percent chance the flag is "on" (so on if we rolled
+		 * <= N, otherwise off).
```
sueloverso: "variable's."
At first it took a while to parse this. It needs consistent use of double-quotes around all the strings. Then it will read easier.

test/format/config.c
```diff
@@ -252,6 +246,56 @@ config_setup(void)
...
+		/* Check if a minimum cache size has been specified. */
+		if (g.c_cache_minimum != 0 && g.c_cache < g.c_cache_minimum)
+			g.c_cache = g.c_cache_minimum;
+	}
+
+	/*
+	 * Maximum internal/leaf page size sanity.
+	 * Ensure we can service at least one operation per-thread concurrently
+	 * without filling the cache with pinned pages. We choose a multiplier
+	 * of three because the max configurations control on disk size and in
```
sueloverso: I realize that this comment and code is moved from another location and this isn't freshly written, but I'd remove "of three" from the comment because it doesn't actually say why 3 itself is a good number and not, say, 4. Since it is used in several places, can you make it a macro? `#define CACHE_MULTIPLIER`. It also makes me sad that we have the dirty trigger value hard coded. Can that be a macro?

test/format/config.c
keithbostic: Pushed in 632b009.

test/format/config.c
keithbostic: Pushed in 632b009.

PR #4080, opened 2018-04-28
WT-4051 format configures too-small LSM caches.

keithbostic:
@sueloverso, I'm going to merge this once it passes PR testing to get general testing back on track, but if you'd take a look when you're back next week, I'd really appreciate it.

Thanks!

### DIFF
--- a/src/include/lsm.h
+++ b/src/include/lsm.h
@@ -193,6 +193,13 @@ struct __wt_lsm_manager {
 #define	WT_LSM_AGGRESSIVE_THRESHOLD	2
 
 /*
+ * The minimum size for opening a tree: three chunks, plus one page for each
+ * participant in up to three concurrent merges.
+ */
+#define	WT_LSM_TREE_MINIMUM_SIZE(chunk_size, merge_max, maxleafpage)	\
+	(3 * (chunk_size) + 3 * ((merge_max) * (maxleafpage)))
+
+/*
  * WT_LSM_TREE --
  *	An LSM tree.
  */
--- a/test/format/config.c
+++ b/test/format/config.c
@@ -255,8 +255,19 @@ config_cache(void)
 {
 	uint32_t max_dirty_bytes, required;
 
-	if (config_is_perm("cache"))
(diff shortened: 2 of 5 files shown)


src/include/lsm.h
```diff
@@ -192,6 +192,13 @@ struct __wt_lsm_manager {
  */
 #define	WT_LSM_AGGRESSIVE_THRESHOLD	2
 
+/*
+ * The minimum size for opening a tree: three chunks, plus one page for each
+ * participant in up to three concurrent merges.
```
sueloverso: It took me a bit to parse and understand the comment. It would be good to remove "three" from the comment and make it flexible. I also changed the name to make clear what the number unit is. What do you think of:
```
#define WT_LSM_CONCURRENT_MERGE_MAX 3
#define WT_LSM_CHUNKS_MIN 3
/*
 * The minimum size for opening a tree: The minimum number of chunks, plus one page
 * for each participant for the maximum number of concurrent merges.
 */
#define WT_LSM_TREE_MINIMUM_BYTES(...)
```

test/format/config.c
```diff
@@ -283,18 +294,24 @@ config_cache(void)
 	for (;;) {
 		max_dirty_bytes = ((g.c_cache * WT_MEGABYTE) / 10) * 4;
 		if (SIZE_ADJUSTMENT * g.c_threads *
-		    (g.c_intl_page_max + g.c_leaf_page_max) <= max_dirty_bytes)
+		    (g.intl_page_max + g.leaf_page_max) <= max_dirty_bytes)
 			break;
-		g.c_cache += g.c_threads * g.c_leaf_page_max;
+		++g.c_cache;
```
sueloverso: Does this change make this loop really slow incrementing by one byte?

src/include/lsm.h
keithbostic: Thanks -- I'll tuck that change into my next "lint" PR.

test/format/config.c
keithbostic: `g.c_cache` is in units of MB, so it's not that slow.

PR #4083, opened 2018-05-01
WT-4069 transaction commit can return leaving a page permanently locked

keithbostic:


michaelcahill:
Thanks, @keithbostic: nice catch!  I'll merge...

PR #4085, opened 2018-05-02
WT-3698 Threads doing eviction can stall indefinitely when timestamps fail to advance.

keithbostic:
@michaelcahill, the problem I'm currently chasing is when a truncate operation dirties a relatively large number of pages (usually a truncate), and the operation resolves, but the thread that would allow format to update the current timestamp values and unpin a bunch of pages gets trapped in `__wt_cache_eviction_worker()`, and we eventually hang because eviction isn't making any progress.

Now that timestamp progress is essential to eviction success, it seems to me we don't want to co-opt  threads committing or rolling back transactions into eviction work. (We'll catch them on their next operation, of course.)

That said... the change to co-opt those threads into eviction went in about 6 months ago, ticket WT-3745, PR #3788, commit fd0c611, with this comment:
```
WT-3745 Favor oplog reads when under cache pressure. (#3788)

There are several behavior changes here to avoid situations where oplog
reads can block when a primary is under severe cache update pressure.
In particular:
* don't block if a scan encounters a large page;
* don't stall reads when the dirty limit of cache is reached;
* mark pages read from lookaside clean if it is safe to do so.

* Revert to including lookaside pages in the dirty limit.

Always check for cache full before starting a transaction, check the
dirty limit at the end of transactions that do updates.

* Don't retry page rewrites until transaction state changes.

We used to have this check for transaction IDs, extend it to also check
when the pinned timestamp hasn't moved forward and don't retry eviction.
```

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -671,7 +671,7 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_TXN_OP *op;
 	WT_UPDATE *upd;
 	u_int i;
-	bool locked, readonly;
+	bool locked;
 #ifdef HAVE_TIMESTAMPS
 	WT_REF *ref;
 	WT_UPDATE **updp;
@@ -689,7 +689,6 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_ASSERT(session, !F_ISSET(txn, WT_TXN_ERROR) ||
 	    txn->mod_count == 0);
 
-	readonly = txn->mod_count == 0;
 	/*
 	 * Look for a commit timestamp.
 	 */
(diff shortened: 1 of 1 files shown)


michaelcahill:
@keithbostic, WT-3745 moved waiting for the dirty cache limit from `begin_transaction` to `commit_transaction` / `rollback_transaction`.  In particular, WT-3745 meant that read-only transactions don't wait when the cache is at the dirty limit.

The change here is untenable: it means that ordinary applications (including MongoDB) that do no operations outside transactions, will stop respecting the dirty cache limit.  That is probably unnoticeable in `test/format` (other than some operations not being throttled) but is not okay at scale.

There is a change to `test/format` that will allow it to move its notion of the oldest timestamp forward even when the cache is full (either clean or dirty).  That is, to use `WT_SESSION::timestamp_transaction` rather than passing `read_timestamp` to `begin_transaction` or `commit_timestamp` to `commit_transaction`.

I'll push a branch, but so far haven't reproduced the stalls on develop with the config in the ticket.

keithbostic:
Close in favor of #4088.

PR #4088, opened 2018-05-03
WT-3698 Don't have format pin timestamp across begin/commit.

michaelcahill:


michaelcahill:
@keithbostic, here's what I'd propose rather than #4085.  In particular, it means threads in `test/format` only keep timestamps pinned across calls to `WT_SESSION::timestamp_transaction`.  That is closer to the behavior of MongoDB, particularly since SERVER-32876.

keithbostic:
@michaelcahill, this branch failed because it called `WT_SESSION.timestamp_transaction` on prepared transactions instead of continuing to set the timestamp on `WT_SESSION.commit_transaction` as prepared transactions require.

I pushed a change in 23c4e64 to fix that behavior, keeping your changes except for prepared transactions, and making them behave as they did before. I don't know if that's correct, I just fixed the symptom. Feel free to simply revert that commit (and the commit in 001bfbc, which is just whitespace/indentation/formatting noise).

For the record, this branch does appear to fix the problem that I've been chasing, the cache timeouts go away with the failing CONFIG.

michaelcahill:
Thanks for the fixes @keithbostic, I've opened WT-4075 to allow a call to `timestamp_transaction` to set the `commit_timestamp` after a transaction is prepared.  I'd like to get this fix in, so I don't want it to wait on that change.  I'll merge...

PR #4089, opened 2018-05-03
WT-4070 Avoid flushing individual files when a stable_timestamp is set.

michaelcahill:


michaelcahill:
@sueloverso, here's what I was trying to describe in our call today -- it seems reasonable to error out of syncing data to a tree when a `stable_timestamp` has been set.

AFAICT, it doesn't introduce any regressions in WT or MongoDB testing (https://evergreen.mongodb.com/version/5aea74f3e3c331252d157392).

Can you please take a look and see whether this addresses the issue identified in powercycle testing?

PR #4091, opened 2018-05-03
WT-4067 format LSM cache requirements are wrong

keithbostic:
@agorrod, feel free to just modify and merge this if there's anything you disagree with.

### DIFF
--- a/src/include/lsm.h
+++ b/src/include/lsm.h
@@ -193,11 +193,11 @@ struct __wt_lsm_manager {
 #define	WT_LSM_AGGRESSIVE_THRESHOLD	2
 
 /*
- * The minimum size for opening a tree: three chunks, plus one page for each
- * participant in up to three concurrent merges.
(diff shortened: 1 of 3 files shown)


keithbostic:
@agorrod, this change means our default cache size of 100MB is no longer sufficient when opening a simple LSM object, the error is:
```
LSM cache size 104857600 (100MB) too small, must be at least 346030080 (330MB)
```
Obviously, we don't  know what objects are going to be opened when we first size the cache.

I guess we could increase the default cache size across the board, or increase the cache size for all of the tests in the test suite, but neither of those sound all that great. Ideas?

agorrod:
@keithbostic I sent you down the wrong path here - there is an issue with how/when LSM is flushing chunks and how/when LSM cursors are opening cursors on live trees vs checkpoints. I'll need to dig more before I really understand and can be helpful.

agorrod:
I'm going to close this out for now - it's not the right approach

PR #4092, opened 2018-05-04
WT-4056 Add API to set minimum required file version.

michaelcahill:


michaelcahill:
@sueloverso, I saw the branch and assumed it was worth opening a pull request.  Feel free to discard if I've misunderstood.

sueloverso:
@agorrod or @michaelcahill (if you're not in transit) this new API is ready for review. The sooner we get this in the sooner it can be used in the server. @dgottlieb FYI.

sueloverso:
FYI, although I did it manually, I created a database via `timestamp-abort` under WiredTiger 2.9.3, then ran a develop `./test_timestamp_abort` with `require_min=3.0`to confirm I got the expected error.  I  then ran the `verify-only` phase on the same database directory under 2.9.3 again to make sure it didn't find any new-version files and could open and use that database.

agorrod:
@sulabhM please review this change

src/conn/conn_reconfig.c
```diff
@@ -25,9 +26,57 @@ __wt_conn_compat_config(WT_SESSION_IMPL *session, const char **cfg)
 	if (cval.len == 0) {
```
agorrod: It's a pity this parsing logic is duplicated for compatability.release and compatibility.require_min - can you see a path to sharing code?

test/suite/test_compat01.py
```diff
@@ -42,10 +42,6 @@ class test_compat01(wttest.WiredTigerTestCase, suite_subprocess):
     logmax = "100K"
     tablename = 'test_compat01'
     uri = 'table:' + tablename
-    sync_list = [
```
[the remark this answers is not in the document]
sueloverso: I started with `test_compat01.py` as the basis for `test_compat02.py`. And `test_compat01.py` had parts copied from some of the `test_txn` tests that focus on recovery. I noticed that each scenario was taking 15-20 seconds (and at 300 scenarios that would be a long time). I figured out that it was the `sync_list` and with the nature of this test there was no benefit to ever testing with sync turned on. So I removed it from both compatibiltiy tests.

src/conn/conn_reconfig.c
sueloverso: Fixed.

PR #4101, opened 2018-05-16
WT-4079 confirm WiredTiger builds under gcc8

keithbostic:


nehakhatri5:
@keithbostic I would like to review this code change. It would be great if you could share some context about this change. I would also like to understand more about strict.m4, how it is used in compilation.

keithbostic:
@nehakhatri5, the reason I started looking at this change was the release of gcc8 (I've installed gcc8 on our test machines, and updated the Jenkins PR compiler check to include gcc8).

The WiredTiger `--enable-strict` configuration option turns on all of the compiler warning flags we can turn on. The idea is that  helps us find problems, but as importantly, because we expect our users to compile our software, we have to make sure we can compile cleanly under whatever compiler flags they happen to use.

The Jenkins PR compiler check turns on `--enable-strict` so we get good testing.

When I went to add gcc8's new `-Wmultistatement-macros` warning flag, I realized we didn't set any warning flags for gcc5, and we were separately listing all of the warning flags for each gcc version and it was getting messy. So, I added flags for gcc5, and I changed the code to add new sets of warning flags for each successive version of gcc, rather than listing all of the flags each version supports separately.

The `strict.m4` flag is used to define the autoconf `AM_GCC_WARNINGS` and `AM_CLANG_WARNINGS` macros that support `--enable-strict`, you can find that code in `build_posix/configure.ac.in`.

One useful part of the review would be to check that I didn't miss any new warnings gcc8 supports that we should be configuring as part of `--enable-strict`.

Please let me know if you have any further questions, or I'm happy to walk you through the change, too.

build_posix/aclocal/strict.m4
```diff
@@ -41,38 +42,53 @@ AC_DEFUN([AM_GCC_WARNINGS], [
...
+		w="$w -Wlogical-op"
+		w="$w -Wunused-const-variable=2"
+	fi
+	if test $gcc7 -eq 1; then
 		w="$w -Walloca"
 		w="$w -Walloc-zero"
 		w="$w -Wduplicated-branches"
-		w="$w -Wduplicated-cond"
 		w="$w -Wformat-overflow=2"
 		w="$w -Wformat-signedness"
```
nehakhatri5: -Wformat-signedness flag is included by gcc5, can it be removed from gcc7's block?

agorrod:
lgtm to me - @nehakhatri5 please go ahead and merge once testing passes.

agorrod:
I also switched http://build.wiredtiger.com:8080/job/wiredtiger-configure-combinations over to run on this branch - please change it back to develop before merging.

build_posix/aclocal/strict.m4
keithbostic: Thanks!

nehakhatri5:
@keithbostic we saw the compiler error here first http://build.wiredtiger.com:8080/job/wiredtiger-pull-request-compilers/7822/console. 
Then I tried compiling on ocelot with gcc81 and saw the same error. I am not sure why it did not appear in the first commit e7c2ee6.

keithbostic:
Thank you, @nehakhatri5 -- this branch turned on `-Wunsafe-loop-optimizations` for gcc7 and gcc8, and that was a mistake. I've reverted and merged that change in #4109.

PR #4103, opened 2018-05-16
WT-4075 Allow timestamp_transaction after prepare.

sueloverso:
I made these modifications. @keithbostic please review the reversion (discussed in the ticket) to `test/format` for this. I ran this with the CONFIG from WT-3698 and it has run some iterations.

@michaelcahill or @bvpvamsikrishna please review the code and unit test changes.

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -657,6 +657,7 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
 
 	/* Look for a commit timestamp. */
 	ret = __wt_config_gets_def(session, cfg, "commit_timestamp", 0, &cval);
+	WT_RET_NOTFOUND_OK(ret);
 	if (ret == 0 && cval.len != 0) {
 #ifdef HAVE_TIMESTAMPS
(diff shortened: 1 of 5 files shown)


sueloverso:
autobuild: retest this please

src/txn/txn_timestamp.c
```diff
@@ -671,7 +671,13 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
 		    "version of WiredTiger built with timestamp support");
 #endif
-	}
+	} else
+		/*
+		 * We allow setting the commit timestamp after a prepare
+		 * but no other timestamp.
+		 */
+		WT_RET(__wt_txn_context_prepare_check(session));
+
```
keithbostic: The error handling here is a little funky. If `__wt_config_gets_def()` fails with `panic`, then we're still going to call `__wt_txn_context_prepare_check()` which doesn't make sense.

Maybe move `WT_RET_NOTFOUND_OK` to immediately after the `__wt_config_gets_def()` call?

keithbostic:
autobuild: retest this please

PR #4106, opened 2018-05-20
WT-4093 Temporarily disable test/format LSM testing

keithbostic:


agorrod:
autobuild: retest this please

keithbostic:
@agorrod, the structure in `format:config` is we have to check if a feature was explicitly configured before setting the data-source and then check again if the feature was randomly configured after setting the data-source.

The configuration code in `format` is very much a moving target and wasn't correct before, but I think it's worth incrementally cleaning up for the future.

I've reworked the change to do that before/after testing, and change the LSM code to match the already existing "reset the configuration if we picked incorrectly" functions for Helium & in-memory testing.

While I was in there, I reduced `LSM` and `file:` testing to 20%, with `table:` being the remaining 60%, I think that's a better balance for our current situation where we want to focus on `table:row-store` testing.

PR #4108, opened 2018-05-21
WT-3963 Add a new schema abort test

agorrod:


agorrod:
@sueloverso I updated the branch to the tip of develop, and fixed one bug that could cause an EINVAL:

```
../../../test/csuite/schema_abort/main.c:862:41: error: passing 'wt_thread_t' to parameter of incompatible type 'wt_thread_t *'; take the address with &
                testutil_check(__wt_thread_join(NULL, thr[i]));
```

I've been running on tinderbox for about 15 minutes without a failure. Would you mind taking a look?

agorrod:
We talked this through in the standup, and believe it's good to go.

PR #4109, opened 2018-05-21
WT-4079 confirm WiredTiger builds under gcc8

keithbostic:
Don't turn on "unsafe-loop-optimizations" after gcc7, it's too noisy to tolerate. This reverts part of a change in e7c2ee6.


keithbostic:
@nehakhatri5, this commit gets us back to clean builds under gcc7 -- sorry about the churn.

I'm going to merge it once it passes testing, but please take a look to be sure.

nehakhatri5:
Thanks @keithbostic, looks good.

PR #4111, opened 2018-05-22
WT-4076 Discard obsolete lookaside history from cache.

michaelcahill:
This unblocks workloads where internal pages containing references to leaf pages with lookaside history remain pinned in cache, even after the lookaside history is obsolete.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -480,7 +480,8 @@ __wt_las_page_obsolete_check(WT_SESSION_IMPL *session, WT_REF *ref)
 	if (!__wt_page_las_active(session, ref)) {
 		__wt_free(session, ref->page_las);
 		WT_PUBLISH(ref->state, previous_state == WT_REF_LOOKASIDE ?
-		    WT_REF_DISK : WT_REF_MEM);
+		    (ref->addr == NULL ? WT_REF_DELETED : WT_REF_DISK) :
+		    WT_REF_MEM);
 	} else
 		WT_PUBLISH(ref->state, previous_state);
 }
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -462,33 +462,6 @@ __wt_las_page_skip(WT_SESSION_IMPL *session, WT_REF *ref)
 }
 
 /*
- * __wt_las_page_obsolete_check --
- *	 Discard references from cache to obsolete lookaside history.
- */
-void
-__wt_las_page_obsolete_check(WT_SESSION_IMPL *session, WT_REF *ref)
-{
-	uint32_t previous_state;
-
-	if ((previous_state = ref->state) != WT_REF_LIMBO &&
-	    previous_state != WT_REF_LOOKASIDE)
-		return;
-
-	if (!__wt_atomic_casv32(&ref->state, previous_state, WT_REF_LOCKED))
-		return;
-
-	if (!__wt_page_las_active(session, ref)) {
-		__wt_free(session, ref->page_las);
-		if (previous_state == WT_REF_LOOKASIDE)
-			WT_PUBLISH(ref->state,
-			    ref->addr == NULL ?  WT_REF_DELETED : WT_REF_DISK);
-		else
-			WT_PUBLISH(ref->state, WT_REF_MEM);
-	} else
-		WT_PUBLISH(ref->state, previous_state);
-}
-
-/*
  * __las_remove_block --
  *	Remove all records for a given page from the lookaside store.
  */
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -479,9 +479,11 @@ __wt_las_page_obsolete_check(WT_SESSION_IMPL *session, WT_REF *ref)
 
 	if (!__wt_page_las_active(session, ref)) {
 		__wt_free(session, ref->page_las);
-		WT_PUBLISH(ref->state, previous_state == WT_REF_LOOKASIDE ?
-		    (ref->addr == NULL ? WT_REF_DELETED : WT_REF_DISK) :
-		    WT_REF_MEM);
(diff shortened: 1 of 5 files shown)


michaelcahill:
I'm going to run some tests against this branch before requesting review.

keithbostic:
@michaelcahill, failure this morning on kodkod: [WT-4096](https://jira.mongodb.org/browse/WT-4096).

Sorry, I created a ticket before I noticed the run wasn't on develop.

michaelcahill:
I think this change is ready for review: it's been running a stress job for a week and the only cases where the cache was stuck full of clean data was for LSM.  It's not perfect, but I think it's a step forward.

src/cache/cache_las.c
```diff
@@ -461,6 +461,33 @@ __wt_las_page_skip(WT_SESSION_IMPL *session, WT_REF *ref)
...
+		return;
+
+	if (!__wt_atomic_casv32(&ref->state, previous_state, WT_REF_LOCKED))
+		return;
+
+	if (!__wt_page_las_active(session, ref)) {
+		__wt_free(session, ref->page_las);
+		if (previous_state == WT_REF_LOOKASIDE)
+			WT_PUBLISH(ref->state,
+			    ref->addr == NULL ?  WT_REF_DELETED : WT_REF_DISK);
```
keithbostic: `NULL ?  WT_REF_DELETED` has extra white-space after `?`

keithbostic: Structurally, I mildly prefer moving this code into `__evict_child_check()`: (1) that's the only caller, (2) I prefer eviction "own" updating the `WT_REF` state, (3) it matches the existing `WT_REF_DELETED` code, but I don't feel that strongly about it.

src/cache/cache_las.c
```diff
@@ -461,6 +461,33 @@ __wt_las_page_skip(WT_SESSION_IMPL *session, WT_REF *ref)
...
+		else
+			WT_PUBLISH(ref->state, WT_REF_MEM);
+	} else
+		WT_PUBLISH(ref->state, previous_state);
+}
```
keithbostic: The final `WT_PUBLISH(ref->state, previous_state);` doesn't need to be a publish because we haven't written anything?

src/evict/evict_page.c
```diff
@@ -453,6 +455,16 @@ __evict_child_check(WT_SESSION_IMPL *session, WT_REF *parent)
...
+		case WT_REF_LOOKASIDE:
+			/*
+			 * Check if the lookaside history is obsolete, which
+			 * will flip the state to WT_REF_DISK.
+			 */
+			__wt_las_page_obsolete_check(session, child);
+			if (child->state != WT_REF_DISK &&
+			    child->state != WT_REF_DELETED)
+				return (EBUSY);
+			break;
```
keithbostic: `will flip the state to WT_REF_DISK.` -> `will flip the state to WT_REF_DISK or WT_REF_DELETED.`

keithbostic: In the case of the page reverting to `WT_REF_DELETED`, isn't there some chance that calling `__wt_page_del_active()` will allow us to proceed and evict the page?

keithbostic: You had to lock the `WT_REF` to free the `page_las` structure which makes sense to me, but that means I don't understand the comment in the `WT_REF_DELETED` case:
```
 * We don't have to lock the page, as no thread of
 * control can be running below our locked internal
 * page.
```
Why doesn't the `WT_REF_DELETED` case have to lock before calling `__wt_page_del_active()`?

keithbostic: I thought the only path to a page having both lookaside information and being deleted was if it was first written and switched into a lookaside state, and then "fast" deleted.

If we're reverting a lookaside page into the deleted state, I think my understanding is wrong -- can you please clarify that for me?

michaelcahill: @keithbostic, the difference between the eviction check of a `WT_REF_DELETED` ref and a `WT_REF_LOOKASIDE` ref is that the deleted check is read-only.  Other readers encountering the same deleted ref will make the same decision about whether it is active but we don't actively free the `ref->page_del` structure.

My first attempt tried to free obsolete lookaside history without flipping to locked and encountered errors.  However, that was because of the case where a page has lookaside history but no page image (discussed below).  I'll try backing out the locking (matching the existing comment), to see whether it is really necessary.

The reason a page can go from `WT_REF_LOOKASIDE` to `WT_REF_DELETED` is if all of the entries on the page were deleted, and reconciliation saw the page as empty, but there was history that could not be discarded.  Once that history becomes obsolete, the page is marked deleted directly.  By definition all of the history is obsolete at this point, so there should be no `page_del` structure and hence no need to call `__wt_page_del_active`.  I'll add a check for that.

src/cache/cache_las.c
michaelcahill: @keithbostic, it would be great to have clear layering for `ref->state` changes (e.g., to reads, truncates and eviction).  I'll move this one (unless I can get rid of it entirely), but there's more work to do...

michaelcahill:
@agorrod, I've removed the code that @keithbostic called out and this has been running for a day (without a single failure!).  OK to merge?

src/evict/evict_page.c
michaelcahill: FTR, I think the existing comment is correct and locking is not required for `WT_REF_DELETED` or `WT_REF_LOOKASIDE`.

At some point I was trying to make this work for `WT_REF_LIMBO`, but that is much closer to `WT_REF_MEM` in that a reader can be inside a child in that state, so it isn't safe to evict the parent (we'd have to lock *and* do a hazard check to make that safe and that doesn't seem worth it as we can just evict the limbo page separately, then try the parent again).

agorrod:

PR #4112, opened 2018-05-23
WT-4049 Adjust restarted search backoff algorithm.

agorrod:
If we backoff too aggressively it makes page splits less effective in some workloads.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -474,6 +474,7 @@ autocommit
 autoconf
 automake
 bInheritHandle
+backoff
 bal
 basecfg
 basho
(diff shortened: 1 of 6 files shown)


agorrod:
I switched Jenkins perf testing onto this branch, and kicked off the perf-long jobs.

agorrod:
The perf jobs show that performance is back to previous levels with this change.

agorrod:
@keithbostic I talked to @michaelcahill about how many times to spin before starting to yield and he wasn't comfortable with something larger than 10, so I'm inclined to take this change as-is.

Could you please do a review and merge if you are happy?

PR #4115, opened 2018-05-24
WT-4087 Store compatibility version in turtle file and check on require_min.

sueloverso:
Not yet ready for review.


sueloverso:
@michaelcahill and @agorrod  this is ready for review. I only store the compatibility version if the user explicitly set one with `compatibility=(release...)`. I only check it if the user sends in a `require_min`.

sueloverso:
FYI, reviewing and merging sooner than later is appreciated. The code location of changes in this ticket overlap significantly with WT-4098. I have some start to that ticket but it really needs some things from this one.

src/meta/meta_table.c
```diff
@@ -16,14 +16,18 @@ static bool
...
 		break;
 	case 'W':
-		if (strcmp(key, "WiredTiger version") == 0)
+		if (strcmp(key, WT_METADATA_VERSION) == 0)
 			return (true);
-		if (strcmp(key, "WiredTiger version string") == 0)
+		if (strcmp(key, WT_METADATA_VERSION_STR) == 0)
 			return (true);
 		break;
 	}
```
michaelcahill: Interesting, so there's no error from existing code for unexpected keys...  That makes this change much less scary.

src/meta/meta_turtle.c
```diff
@@ -286,9 +286,13 @@ err:	WT_TRET(__wt_fclose(session, &fs));
...
 	 * A file error or a missing key/value pair in the turtle file means
-	 * something has gone horribly wrong -- we're done.
+	 * something has gone horribly wrong, except for the compatibility
+	 * setting which is optional.
 	 */
-	return (ret == 0 ? 0 : __wt_illegal_value(session, WT_METADATA_TURTLE));
+	return (ret == 0 ? 0 :
+	    (WT_STRING_MATCH(
+	    key, WT_METADATA_COMPAT, strlen(WT_METADATA_COMPAT)) ? ret :
+	    __wt_illegal_value(session, WT_METADATA_TURTLE)));
```
michaelcahill: This seems kind of indirect, how about:
```
return (ret == 0 || strcmp(key, WT_METADATA_COMPAT) == 0 ? ret :
    __wt_illegal_value(session, WT_METADATA_TURTLE));
```
?

src/conn/conn_reconfig.c
```diff
@@ -71,40 +76,89 @@ __wt_conn_compat_config(
...
+	if (reconfig) {
+		/*
+		 * We rewrite the turtle file with the metadata information
+		 * that is already there. This call is only to update the
+		 * compatibility information. But since the turtle file is
+		 * entirely rewritten every time we need to send in the
+		 * metadata file information.
+		 */
+		WT_RET(__wt_metadata_search(session, WT_METAFILE_URI, &value));
+		WT_ERR(__wt_metadata_update(session, WT_METAFILE_URI, value));
```
agorrod: I've got two questions about this:

* The ordering seems wrong the following code compares the "saved" version which is retrieved from the metadata - but we've updated the metadata here. Is it comparing the same things? Should we only rewrite the metadata once the versioning checks are done?
* Could we avoid rewriting the metadata if nothing is going to change? That feels nicer than always doing it.

sueloverso: The metadata search/update is how we save or remove the compatibility setting from the turtle file. At this point we cannot tell if anything changed because we may be reading this on a restart. Reading the turtle file, parsing and comparing the compatibility contents here, and then conditionally doing the above lines feels like more complicated code for not much gain since we rarely do reconfigures.

agorrod: I do feel uncomfortable re-writing the metadata before doing the checks. Do we have tests that restart to make sure the database isn't left in a weird state after an error is returned? If so I'm happy for this change to be merged.

agorrod: I get it - it's only being re-written here via reconfigure so nothing should happen unless the user has called reconfigure.

I still think it'd be nice not to rewrite the turtle file unless there is an intent to change something in there, but see that it's not as big of a problem.

sueloverso: I will see if some reformatting and restructuring helps.

sueloverso: @agorrod Thanks for pushing back on this because when I restructured it, I think it uncovered a bug where someone could open with a valid required_min setting and then later reconfigure with a compatibility version that violates that requirement. I'll fix that and make sure the unit test covers that.

src/meta/meta_table.c
sueloverso: I'm not following. The macros (which already existed) should be an exact literal replacement here.

agorrod:
Thanks @sueloverso this looks better to me now. The logic is more complex, but I think the flow is pretty good.

agorrod:
@michaelcahill please take a look and merge if you are happy

src/meta/meta_table.c
michaelcahill: Sorry, my point (noticed here but not specifically tied to a line of code) was that there's no code that will result in an error because of an unrecognized key in the turtle file: we just skip over them.

src/meta/meta_table.c
```diff
@@ -16,20 +16,44 @@ static bool
...
+int
+__wt_metadata_turtle_rewrite(WT_SESSION_IMPL *session)
+{
+	WT_DECL_RET;
+	char *value;
+
+	value = NULL;
+	WT_RET(__wt_metadata_search(session, WT_METAFILE_URI, &value));
+	if ((ret = __wt_metadata_update(session, WT_METAFILE_URI, value)) == 0)
+		__wt_free(session, value);
```
michaelcahill: I believe there's a leak of `value` here on error.  I'll push the trivial fix.

michaelcahill:

PR #4116, opened 2018-05-31
WT-4105 Reduce cache use by modify operations with history pinned.

michaelcahill:


src/btree/bt_cursor.c
```diff
@@ -1360,9 +1365,14 @@ __cursor_chain_exceeded(WT_CURSOR_BTREE *cbt)
 	    page->modify != NULL && page->modify->mod_row_update != NULL)
 		upd = page->modify->mod_row_update[cbt->slot];
 
-	for (i = 0; upd != NULL; ++i, upd = upd->next) {
+	for (i = 0, upd_size = 0; upd != NULL; ++i, upd = upd->next) {
 		if (WT_UPDATE_DATA_VALUE(upd))
 			return (false);
+		upd_size += WT_UPDATE_MEMSIZE(upd);
+		if (!__wt_txn_upd_visible_all(session, upd))
+			continue;
```
michaelcahill: We should explain that when we have to maintain history, creating extra copies of a large document multiplies the cache pressure, so we avoid it.

src/include/btmem.h
```diff
@@ -1071,9 +1071,15 @@ struct __wt_update {
 /*
  * WT_MAX_MODIFY_UPDATE --
  *	Limit update chains to a small value to avoid penalizing reads and
- * permit truncation.
+ *	permit truncation.
  */
-#define	WT_MAX_MODIFY_UPDATE	10
+#define	WT_MAX_MODIFY_UPDATE	10000
```
michaelcahill: This is no longer a "small value", so if we stick with this we'll need to update the comment.

bvpvamsikrishna:
@michaelcahill, I have pushed the reproducer to this branch. It was aborting with in less than 200 modify count with out this code patch and with code patch is still running for more than 120K modify count when I killed it.

src/btree/bt_cursor.c
```diff
@@ -1360,9 +1365,19 @@ __cursor_chain_exceeded(WT_CURSOR_BTREE *cbt)
...
+		/*
+		 * When history has to be maintained, creating extra copies
+		 * of large documents multiplies cache pressure, so avoid it,
+		 * by growing the modify chain.
+		 */
+		if (upd_size >= WT_MODIFY_MEM_FACTOR * cursor->value.size)
+			return (true);
```
michaelcahill: I'm currently testing the performance with this check moved above the `__wt_txn_upd_visible_all` check.  In other words, when updates are a large proportion of the underlying record size, don't keep too many around, even if not all of them are visible.  The change still allows @bvpvamsikrishna's test to complete, but seems to have less impact on performance of automated tests.

I'll report back on the ticket when I have more solid data.

src/btree/bt_cursor.c
```diff
@@ -1360,9 +1365,19 @@ __cursor_chain_exceeded(WT_CURSOR_BTREE *cbt)
...
 			return (false);
+		upd_size += WT_UPDATE_MEMSIZE(upd);
+		if (upd_size >= WT_MODIFY_MEM_FACTOR * cursor->value.size)
+			return (true);
+		/*
+		 * When history has to be maintained, creating extra copies
+		 * of large documents multiplies cache pressure, so avoid it,
+		 * by growing the modify chain.
+		 */
+		if (!__wt_txn_upd_visible_all(session, upd))
```
bvpvamsikrishna: I think this check better either 'before previous if check (line 1372)' or 'after next if check (line 1381)'. In between seems to be a bit odd, as both if checks are in way to constrain the length of the update chain. @michaelcahill, request for your opinion.

michaelcahill: @bvpvamsikrishna, I've put it this way deliberately, maybe a comment can help?

Basically, the size check means we don't create huge amounts of cache pressure (10x 8 byte updates to a 10MB record creates 125000x cache pressure today).  The size check keeps the cache pressure to 2x.

The length check is mostly about read performance: we don't want to do too many apply operations on every read.

The reason to put the visibility check in between is to keep caching small updates when they can't be cleaned up.  It's okay to create new complete records when we can free up the last one.

Does that make sense, could I add a comment to clarify?

bvpvamsikrishna: @michaelcahill, Thanks for clarification. sounds good. 
Only thing is comment is not moved with the if check. I can move it, if we plan to merge this soon.

agorrod:
Thanks @michaelcahill lgtm.

PR #4118, opened 2018-06-04
WT-4098 Add require_max compatibility support.

sueloverso:
Not yet ready for review. It still needs unit tests added.


sueloverso:
@agorrod this is ready for review. It feels like it is getting fairly complex with all the various error conditions that are possible between min release, max release, current compatibility release and database creation release. The unit test scenarios capture them but they are hard to reason about.

The original intent of the `require_min/max` was to make sure an existing database met those on restart, but the unit test made it seem that they should be checked against the current requested release as well.

For instance, create a database with `compatibility=(release="3.0")`. Then reopening it with `compatibility=(release=,require_max="3.0)` will fail because the (default) `release=` will make the current open run at the current release, not the original "3.0", and that is larger than the "3.0" max required. Or should it strictly apply to the old/original log files but not new files going forward, which in this case would be "3.1". Right now it errors, which is more conservative and makes more sense to me, but this is hard to reason about, like I said.

sueloverso:
@dgottlieb if you could try this branch against the tests you run with the various compatibility calls that would be helpful. I've moved and altered some of the code and conditionals. I would like to make sure I didn't change any existing error/success paths that you're currently using and expecting.

agorrod:
Thanks @sueloverso  these changes are looking good to me, and I think you've done a great job of making the error checking clear and centralized.

I've kicked off a compatibility run on this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-compatibility/230/

agorrod:
@sueloverso lgtm - please go ahead and merge if it satisfies @dgottlieb

src/conn/conn_log.c
```diff
@@ -125,15 +125,23 @@ __logmgr_version(WT_SESSION_IMPL *session, bool reconfig)
...
 	 */
-	if (conn->compat_req_major != WT_CONN_COMPAT_NONE) {
-		if (conn->compat_req_major < WT_LOG_V2_MAJOR)
-			conn->log_req_version = 1;
-		else if (conn->compat_req_minor == WT_LOG_V2_MINOR)
-			conn->log_req_version = 2;
+	if (conn->req_max_major != WT_CONN_COMPAT_NONE) {
+		if (conn->req_max_major < WT_LOG_V2_MAJOR)
+			conn->log_req_max = 1;
+		else if (conn->req_max_minor == WT_LOG_V2_MINOR)
```
nehakhatri5: Now that we use both major and minor versions the comment in log.h [https://github.com/wiredtiger/wiredtiger/blob/518da6fac9bee33247d6cf879e68540cb37e9adf/src/include/log.h#L371](url) is not true any more.

dgottlieb:
lgtm!

PR #4120, opened 2018-06-05
WT-4076 Discard obsolete lookaside history

agorrod:


agorrod:
@michaelcahill should I go ahead and pull this back in?

PR #4121, opened 2018-06-06
WT-4110 Fix a checkpoint timestamp race that skipped recent commits.

michaelcahill:
Checkpoints were first starting a snapshot transaction without a read timestamp, than checking if a stable timestamp was set by the application.  If the stable timestamp was updated after checkpoint
established its snapshot, the checkpoint's transaction was invalid and could skip over records that were committed after the snapshot with timestamps less than the stable timestamp.

### DIFF
--- a/src/txn/txn_ckpt.c
+++ b/src/txn/txn_ckpt.c
@@ -627,11 +627,13 @@ __checkpoint_prepare(
 	WT_CONFIG_ITEM cval;
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
+	WT_DECL_TIMESTAMP(read_timestamp)
 	WT_TXN *txn;
 	WT_TXN_GLOBAL *txn_global;
 	WT_TXN_STATE *txn_state;
 	const char *txn_cfg[] = { WT_CONFIG_BASE(session,
 	    WT_SESSION_begin_transaction), "isolation=snapshot", NULL, NULL };
+	char timestamp_buf[64 + 2 * WT_TIMESTAMP_SIZE];
 	bool use_timestamp;
 
 	conn = S2C(session);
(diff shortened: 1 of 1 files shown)


src/txn/txn_ckpt.c
```diff
@@ -705,46 +751,6 @@ __checkpoint_prepare(
...
-		 * If the user wants timestamps then set the metadata
-		 * checkpoint timestamp based on whether or not a stable
-		 * timestamp is actually in use.  Only set it when we're not
-		 * running recovery because recovery doesn't set the recovery
-		 * timestamp until its checkpoint is complete.
-		 */
-		if (txn_global->has_stable_timestamp) {
-			__wt_timestamp_set(&txn->read_timestamp,
-			    &txn_global->stable_timestamp);
-			F_SET(txn, WT_TXN_HAS_TS_READ);
```
michaelcahill: The bug is caused by this code that sets a read timestamp on a running transaction.

src/txn/txn_ckpt.c
```diff
@@ -642,6 +644,50 @@ __checkpoint_prepare(
...
+	/*
+	 * Set the checkpoint transaction's timestamp, if requested.
+	 *
+	 * We rely on having the global transaction data locked so the oldest
+	 * timestamp can't move past the stable timestamp.
+	 */
+	WT_ASSERT(session, !F_ISSET(txn,
+	    WT_TXN_HAS_TS_COMMIT | WT_TXN_HAS_TS_READ |
+	    WT_TXN_PUBLIC_TS_COMMIT | WT_TXN_PUBLIC_TS_READ));
+
```
michaelcahill: Hmm, on looking more closely at this, I see that we did it to avoid the race mentioned in the comment above.  This is going to need some more work...

michaelcahill:
Hold off, I see that my first attempt introduces a different race (where the oldest_timestamp is updated after it is read but before it is used).  I'll push a different version.

michaelcahill:
I'm happy with where this has ended up: it matches what we're doing for `WT_SESSION::timestamp_transaction`.

@agorrod can you please review?

agorrod:
This lgtm @michaelcahill I agree that it's good that the semantics match with `timestamp_transaction`

sueloverso:
Although this has been merged, LGTM.

PR #4122, opened 2018-06-07
WT-4105 restore max modify update value and tune to run as short test.

bvpvamsikrishna:
Restored the max modify update value back to 10.
Tune the test to run as short test instead of long test.
print test statements are masked with verbose flag.


agorrod:
Thanks @bvpvamsikrishna lgtm.

PR #4124, opened 2018-06-08
WT-4116 Coverity #1393311 Copy-paste error

keithbostic:


bvpvamsikrishna:
@keithbostic, Thank you for the fixes.

PR #4130, opened 2018-06-12
WT-4123 Jenkins compiler tester on old branches fails with GCC8

keithbostic:


agorrod:
We decided to make a Jenkins change rather than doing this backport.

PR #4131, opened 2018-06-12
WT-4122 Add test and fix for MongoDB's downgrade usage.

sueloverso:
Not yet ready for review. This change, while fixing the problem may not detect an actual error condition properly. Adding more test scenarios and debugging that.


sueloverso:
@agorrod This is ready for review. One thing to note is that test/format does not support compatibility configuration any longer. That was removed a long time ago so there are no changes to it as part of this change. I'll update the ticket with more details. @dgottlieb you can review and/or check this branch against your compatibility tests you've been running.

src/conn/conn_log.c
```diff
@@ -69,15 +69,23 @@ __logmgr_force_ckpt(WT_SESSION_IMPL *session, uint32_t lognum)
...
 		    &tmp_session->iface, "force=1"));
-		WT_RET(WT_SESSION_CHECK_PANIC(tmp_session));
 		/*
 		 * Only sleep in the rare case that we had to come through
 		 * this loop more than once.
 		 */
-		if (yield++) {
-			WT_STAT_CONN_INCR(session, log_force_ckpt_sleep);
-			__wt_sleep(0, WT_THOUSAND);
+		if (yield++ != 0) {
```
agorrod: We have a fairly new `__wt_spin_backoff` that implements a shared version of this. I'd also be totally fine with not tracking the yield stat here - and just tracking sleeps.

michaelcahill:
Thanks @sueloverso and @agorrod -- lgtm.  Merging...

PR #4132, opened 2018-06-13
WT-4125 Repeated checkpoints at the same timestamp shouldn't read pages.

michaelcahill:
1. Add a test demonstrating the problem.

### DIFF
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -255,19 +255,31 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 
 			if (walk == NULL)
 				break;
+			page = walk->page;
 
 			/* Skip clean pages. */
-			if (!__wt_page_is_modified(walk->page))
+			if (page->modify == NULL)
 				continue;
-
-			/*
-			 * Take a local reference to the page modify structure
-			 * now that we know the page is dirty. It needs to be
-			 * done in this order otherwise the page modify
-			 * structure could have been created between taking the
-			 * reference and checking modified.
-			 */
-			page = walk->page;
+			if (!__wt_page_is_modified(page)) {
+				/*
+				 * Track the most recent update in the tree.
+				 * This determines when the tree can safely be
+				 * discarded from cache.
+				 */
+				if (WT_TXNID_LT(btree->rec_max_txn,
+				    page->modify->rec_max_txn))
+					btree->rec_max_txn =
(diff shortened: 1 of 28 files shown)


michaelcahill:
Note: testing is expected to fail until the underlying issue is addressed.

agorrod:
Thanks @michaelcahill this change lgtm. I think @keithbostic should take a look too - especially at the change to how context is carried across update/restores now.

src/btree/bt_sync.c
```diff
@@ -256,17 +256,21 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+			 * safely be discarded from cache.
 			 */
-			if (!__wt_page_is_modified(walk->page)) {
-				if (((mod = walk->page->modify) != NULL) &&
-				    mod->rec_max_txn > btree->rec_max_txn)
+			if ((mod = page->modify) == NULL)
+				continue;
+			if (!__wt_page_is_modified(page)) {
+				if (WT_TXNID_LT(btree->rec_max_txn,
+				    mod->rec_max_txn))
```
keithbostic: Checking `(mod = page->modify) == NULL)` isn't necessary, it could just be:
```
/*
 * Skip clean pages, but track the most recent update
 * in the tree. This determines when the tree can
 * safely be discarded from cache.
 */
if (!__wt_page_is_modified(page)) {
        mod = page->modify;
        if (WT_TXNID_LT(btree->rec_max_txn,
            mod->rec_max_txn))
```

michaelcahill:
Thanks @keithbostic, I've made all the changes you suggested.

agorrod:
Thanks @michaelcahill lgtm.

src/cache/cache_las.c
```diff
@@ -724,8 +715,7 @@ __wt_las_insert_block(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
 			 * table. (We check the length because row-store doesn't
 			 * write zero-length data items.)
 			 */
-			if (multi->page_las.las_skew_newest &&
-			    upd == list->onpage_upd &&
+			if (upd == list->onpage_upd &&
```
michaelcahill: @agorrod, @keithbostic this change in this ticket was not required for correctness -- I just couldn't work out why we were only using birthmarks when skewing newest.  If we make any progress reproducing WT-4144, reverting this change would be the first thing I'd try.

PR #4134, opened 2018-06-13
WT-4128 Skip creating checkpoints if the stable timestamp allows.

agorrod:
If the stable timestamp doesn't move between checkpoints, their content would be identical.
Also move the similar existing check for clean databases from the checkpoint server into the checkpoint code itself.

### DIFF
--- a/src/conn/conn_ckpt.c
+++ b/src/conn/conn_ckpt.c
@@ -101,37 +101,23 @@ __ckpt_server(void *arg)
 		if (!__ckpt_server_run_chk(session))
 			break;
 
-		/*
-		 * Checkpoint the database if the connection is marked dirty.
-		 * A connection is marked dirty whenever a btree gets marked
-		 * dirty, which reflects upon a change in the database that
-		 * needs to be checkpointed. Said that, there can be short
-		 * instances when a btree gets marked dirty and the connection
-		 * is yet to be. We might skip a checkpoint in that short
-		 * instance, which is okay because by the next time we get to
-		 * checkpoint, the connection would have been marked dirty and
-		 * hence the checkpoint will not be skipped this time.
-		 */
-		if (conn->modified) {
-			WT_ERR(wt_session->checkpoint(wt_session, NULL));
+		WT_ERR(wt_session->checkpoint(wt_session, NULL));
 
-			/* Reset. */
-			if (conn->ckpt_logsize) {
-				__wt_log_written_reset(session);
-				conn->ckpt_signalled = false;
+		/* Reset. */
+		if (conn->ckpt_logsize) {
+			__wt_log_written_reset(session);
+			conn->ckpt_signalled = false;
 
-				/*
-				 * In case we crossed the log limit during the
-				 * checkpoint and the condition variable was
-				 * already signalled, do a tiny wait to clear
-				 * it so we don't do another checkpoint
-				 * immediately.
-				 */
-				__wt_cond_wait(
-				    session, conn->ckpt_cond, 1, NULL);
-			}
-		} else
-			WT_STAT_CONN_INCR(session, txn_checkpoint_skipped);
+			/*
+			 * In case we crossed the log limit during the
+			 * checkpoint and the condition variable was
+			 * already signalled, do a tiny wait to clear
+			 * it so we don't do another checkpoint
+			 * immediately.
+			 */
+			__wt_cond_wait(
+			    session, conn->ckpt_cond, 1, NULL);
+		}
 	}
 
 	if (0) {
--- a/src/conn/conn_ckpt.c
+++ b/src/conn/conn_ckpt.c
@@ -83,6 +83,7 @@ __ckpt_server(void *arg)
 	WT_DECL_RET;
 	WT_SESSION *wt_session;
 	WT_SESSION_IMPL *session;
+	uint64_t checkpoint_gen;
 
 	session = arg;
 	conn = S2C(session);
@@ -101,10 +102,15 @@ __ckpt_server(void *arg)
 		if (!__ckpt_server_run_chk(session))
 			break;
 
+		checkpoint_gen = __wt_gen(session, WT_GEN_CHECKPOINT);
 		WT_ERR(wt_session->checkpoint(wt_session, NULL));
 
-		/* Reset. */
-		if (conn->ckpt_logsize) {
+		/*
+		 * Reset the log file size counters if the checkpoint wasn't
+		 * skipped.
+		 */
+		if (checkpoint_gen != __wt_gen(session, WT_GEN_CHECKPOINT) &&
+		    conn->ckpt_logsize) {
 			__wt_log_written_reset(session);
 			conn->ckpt_signalled = false;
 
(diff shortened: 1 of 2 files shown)


src/txn/txn_ckpt.c
```diff
@@ -994,8 +1076,17 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
-		__wt_timestamp_set(
-		    &conn->txn_global.last_ckpt_timestamp, &ckpt_tmp_ts);
+		/*
+		 * If timestamps were used to define the content of the
+		 * checkpoint update the saved last checkpoint timestamp,
+		 * otherwise leave it alone. If a checkpoint is taken without
+		 * timestamps, it's likely a bug, but we don't want to clear
+		 * the saved last checkpoint timestamp regardless.
+		 */
+		if (use_timestamp)
```
agorrod: This part of the change is causing a test failure - and I only added it out of paranoia - I'm going to back it out for now.

agorrod: I take that back - I just hadn't initialized use_timestamp..

agorrod:
@sulabhM please take a look at this change - it builds on some work you did related to optimizing checkpoints.

@keithbostic I'd appreciate your review as well, but please hold off merging until I've had a chance to do some performance tests to see if it's worthwhile.

src/txn/txn_ckpt.c
```diff
@@ -767,10 +764,93 @@ __checkpoint_prepare(
...
+/*
+ * __txn_checkpoint_can_skip --
+ *	Determine whether it's safe to skip taking a checkpoint. This function
+ *	also parses out some configuration options and hands them back to the
+ *	caller - make sure it does that parsing regardless of the result.
+ */
+static int
+__txn_checkpoint_can_skip(WT_SESSION_IMPL *session,
+    const char *cfg[], bool *fullp, bool *use_timestampp, bool *can_skipp)
+{
```
sueloverso: Having 3 boolean pointers as args makes the interface here clunky, especially since there is only a single caller of this function. Most of that is because of the config parsing. The interesting thing is that I don't see the removal of the same parsing elsewhere, so it is additional/duplicate work instead of moving the work.

src/conn/conn_ckpt.c
```diff
@@ -101,37 +101,23 @@ __ckpt_server(void *arg)
...
-				__wt_cond_wait(
-				    session, conn->ckpt_cond, 1, NULL);
-			}
-		} else
-			WT_STAT_CONN_INCR(session, txn_checkpoint_skipped);
+		WT_ERR(wt_session->checkpoint(wt_session, NULL));
+
+		/* Reset. */
+		if (conn->ckpt_logsize) {
+			__wt_log_written_reset(session);
```
keithbostic: There's a change here in that we'll reset the log-written value even if we decided to skip the checkpoint vs. the previous behavior where we only reset it if `conn.modified` was true. I have no idea if that's interesting or not, but it's a change. @sueloverso?

src/conn/conn_ckpt.c
```diff
@@ -101,37 +101,23 @@ __ckpt_server(void *arg)
...
+
+			/*
+			 * In case we crossed the log limit during the
+			 * checkpoint and the condition variable was
+			 * already signalled, do a tiny wait to clear
+			 * it so we don't do another checkpoint
+			 * immediately.
+			 */
+			__wt_cond_wait(
+			    session, conn->ckpt_cond, 1, NULL);
```
keithbostic: Whitespace, these lines can be joined.

src/txn/txn_ckpt.c
```diff
@@ -767,10 +764,93 @@ __checkpoint_prepare(
...
+	WT_CONFIG targetconf;
+	WT_CONFIG_ITEM cval, k, v;
+	WT_CONNECTION_IMPL *conn;
+	WT_TXN_GLOBAL *txn_global;
+	bool full, use_timestamp;
+
+	conn = S2C(session);
+	txn_global = &conn->txn_global;
+	/* Default to not skipping. */
+	*can_skipp = false;
```
keithbostic: I think our style (at the moment) is that side-effects are done before anything else.

src/txn/txn_ckpt.c
```diff
@@ -767,10 +764,93 @@ __checkpoint_prepare(
...
+	    session, cfg, __wt_checkpoint_get_handles));
 	return (ret);
 }
 
+/*
+ * __txn_checkpoint_can_skip --
+ *	Determine whether it's safe to skip taking a checkpoint. This function
+ *	also parses out some configuration options and hands them back to the
+ *	caller - make sure it does that parsing regardless of the result.
+ */
```
keithbostic: If it was me, I'd bring this comment down into the function code where it matters, rather than up here? I certainly don't feel strongly though.

src/txn/txn_ckpt.c
```diff
@@ -767,10 +764,93 @@ __checkpoint_prepare(
...
+		return (0);
+
+	/*
+	 * Skip checkpointing the database if nothing has been dirtied since
+	 * the last checkpoint. That said there can be short instances when a
+	 * btree gets marked dirty and the connection is yet to be. We might
+	 * skip a checkpoint in that short instance, which is okay because by
+	 * the next time we get to checkpoint, the connection would have been
+	 * marked dirty and hence the checkpoint will not be skipped this time.
+	 */
```
keithbostic: "skipped this time" -> "skipped that time".
We could re-order the setting of the connection modified field I suppose, not sure why it's the way it is.

keithbostic:
@sueloverso, there's a comment in that file:
```
/* Configure logging only if doing a full checkpoint. */
logging = FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED);
```
That is set without regard for if we're doing a full checkpoint. Might be worth fixing.

src/conn/conn_ckpt.c
sueloverso: This is an interesting question. If we skip a checkpoint, we should skip reseting how much log was written for the checkpoint. This function is to manage and account for the checkpoint setting of triggering a checkpoint based on log volume rather than time (with the internal thread). If stable isn't moving but log volume grows, this function would continually reset it and checkpoints would not occur at the right time. I think bugs may occur if someone is using timestamps and an internal checkpoint thread or at least we should have a general discussion about that.

sueloverso:
@agorrod regarding @keithbostic 's comment about:
```
/* Configure logging only if doing a full checkpoint. */
logging = FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED);
```
While you're in there, can you remove that comment. We use that variable just as a shorthand for logging being enabled overall. Full and logging need to be separate so the comment is outdated.

michaelcahill:
I agree with @sueloverso's comment that some further refactoring is possible now that config parsing is being done in `__txn_checkpoint_can_skip` and some flags returned, but don't think it's worth holding up this fix for that improvement.

Thanks @agorrod, @sueloverso and @keithbostic -- merging.

src/txn/txn_ckpt.c
sulabhM: Any update on this part of the change?
I didn't understand why it is a bug to take a checkpoint without timestamps. I thought it is legal to do so even in a timestamp build. I understand ckpt_tmp_ts will be set to transaction's read timestamp. If a checkpoint is without a timestamp, would read timestamp of the transaction be set to 0, if yes only in that case we were clearing the last_ckpt_timestamp.

PR #4135, opened 2018-06-13
WT-4127 Add compatibility prefix.

sueloverso:
@agorrod @dgottlieb Here are the changes for adding a prefix, and I changed the unit tests to reflect it.


sueloverso:
In particular, please take a look at the files changed and make sure you agree with the places I did not modify and chose NOT to put the prefix.

agorrod:
Talked to @dgottlieb and he is on board, so I'm going to merge.

PR #4136, opened 2018-06-15
WT-4133 Coverity 1393445, 1393446 Dereference before null check.

keithbostic:
The variables fullp and use_timestampp are dereferenced before being checked for NULL.


agorrod:
Thanks @keithbostic

PR #4137, opened 2018-06-18
WT-4111 Tune checkpoint scrubbing.

michaelcahill:
1. drop the floor at which eviction of dirty pages cut out from 5% to 1% of the cache size (making checkpoint more efficient for workloads with low write rates); and
2. don't slow down writes to reduce the amount of dirty data in cache (reducing the effect of scrubbing on workloads with high write rates).


michaelcahill:
@agorrod, for discussion only, at this stage.

agorrod:
Thanks @michaelcahill lgtm.

PR #4138, opened 2018-06-20
WT-4048 Extend timing_stress_for_test split options

keithbostic:
Make the timing stress function generally available.
Insert a random delay, up to 1/10th of a second, always delaying that long significantly slows down some applications which could lead to less useful testing.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -31,24 +31,6 @@ typedef enum {
 } WT_SPLIT_ERROR_PHASE;
 
 /*
- * __page_split_timing_stress --
- *	Optionally add delay to simulate the race conditions in page split for
- * debug purposes. The purpose is to uncover the race conditions in page split.
- */
-static void
-__page_split_timing_stress(
-    WT_SESSION_IMPL *session, uint64_t flag, uint64_t micro_seconds)
-{
-	WT_CONNECTION_IMPL *conn;
-
-	conn = S2C(session);
-
-	/* We only want to sleep when page split race flag is set. */
-	if (FLD_ISSET(conn->timing_stress_flags, flag))
-		__wt_sleep(0, micro_seconds);
-}
-
-/*
  * __split_safe_free --
  *	Free a buffer if we can be sure no thread is accessing it, or schedule
  *	it to be freed otherwise.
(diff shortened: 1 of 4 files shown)


src/btree/bt_split.c
```diff
@@ -1259,8 +1235,7 @@ __split_internal_lock(
 		parent = ref->home;
 
 		/* Encourage races. */
-		__page_split_timing_stress(
-		    session, WT_TIMING_STRESS_SPLIT_7, WT_THOUSAND);
```
agorrod: Did you deliberately change the sleep time for this option? I presume the sleep was very short here because it encouraged more races. How would you feel about adding a flags to `__wt_timing_stress` with a `SHORT` option?

src/include/misc.i
```diff
@@ -248,3 +248,26 @@ __wt_spin_backoff(uint64_t *yield_count, uint64_t *sleep_usecs)
...
+	WT_CONNECTION_IMPL *conn;
+	uint64_t sleep_usecs;
+
+	conn = S2C(session);
+
+	/* Only sleep when the specified configuration flag is set. */
+	if (!FLD_ISSET(conn->timing_stress_flags, flag))
+		return;
+
+	sleep_usecs = __wt_random(&session->rnd) % WT_TIMING_STRESS_MAX_DELAY;
```
agorrod: I like this idea - I actually think it'd be very cool if it could skew low so the pause is generally very short. A concern I have with this mode is that it slows things down so much we get less real work done and thus split races aren't as common..

src/btree/bt_split.c
keithbostic: I did change it deliberately, I couldn't think of a reason the lock stress timer would be different than any other stress timer, and I didn't find anything in the history of the changes that made me think there was something interesting happening. If there's any evidence that a shorter stress timer here contributes to finding races, we should change it back, of course.

If we do change it back, I'd suggest a "fixed-number-of-usecs" argument to `__wt_timing_stress()`, because then the checkpoint timer stress code can use it too.

src/include/misc.i
keithbostic: I did a uniform distribution because it was easy. You might be correct, but it's hard to know for sure. Generally in my own testing I use big sleeps, but I have no idea if that's optimal or not...

agorrod:

PR #4140, opened 2018-06-21
WT-4138 Add option to timeout operations waiting for cache.

agorrod:
The option is to wiredtiger_open called cache_max_wait, and applies per operation on a session.

Also:
Fixup error reporting in bloom filter code.
Enhance test/format to work with new timeout code.

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -417,6 +417,11 @@ connection_runtime_config = [
         maximum heap memory to allocate for the cache. A database should
         configure either \c cache_size or \c shared_cache but not both''',
         min='1MB', max='10TB'),
+    Config('cache_max_wait', '0', r'''
+        the maximum number of milliseconds an application thread will wait
+        for space to be available in cache before giving up. Default will
+        wait forever''',
+        min=0),
     Config('cache_overhead', '8', r'''
         assume the heap allocator overhead is the specified percentage, and
         adjust the cache usage by that amount (for example, if there is 10GB
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -417,7 +417,7 @@ connection_runtime_config = [
         maximum heap memory to allocate for the cache. A database should
         configure either \c cache_size or \c shared_cache but not both''',
         min='1MB', max='10TB'),
-    Config('cache_max_wait', '0', r'''
+    Config('cache_max_wait_ms', '0', r'''
         the maximum number of milliseconds an application thread will wait
         for space to be available in cache before giving up. Default will
         wait forever''',
--- a/src/conn/conn_cache.c
+++ b/src/conn/conn_cache.c
@@ -144,7 +144,7 @@ __cache_config_local(WT_SESSION_IMPL *session, bool shared, const char *cfg[])
 	conn->evict_threads_min = evict_threads_min;
 
 	/* Retrieve the wait time and convert from milliseconds */
-	WT_RET(__wt_config_gets(session, cfg, "cache_max_wait", &cval));
+	WT_RET(__wt_config_gets(session, cfg, "cache_max_wait_ms", &cval));
 	cache->cache_max_wait_us = (uint64_t)(cval.val * WT_THOUSAND);
 
 	return (0);
--- a/src/conn/conn_cache.c
+++ b/src/conn/conn_cache.c
@@ -143,6 +143,11 @@ __cache_config_local(WT_SESSION_IMPL *session, bool shared, const char *cfg[])
 	conn->evict_threads_max = evict_threads_max;
 	conn->evict_threads_min = evict_threads_min;
 
+	/* Retrieve the wait time and convert from milliseconds */
+	WT_RET(__wt_config_gets(session,
+	    cfg, "cache_max_wait", &cval));
+	cache->cache_max_wait_us = (uint32_t)cval.val * WT_THOUSAND;
+
 	return (0);
 }
 
--- a/src/conn/conn_cache.c
+++ b/src/conn/conn_cache.c
@@ -144,9 +144,8 @@ __cache_config_local(WT_SESSION_IMPL *session, bool shared, const char *cfg[])
 	conn->evict_threads_min = evict_threads_min;
 
 	/* Retrieve the wait time and convert from milliseconds */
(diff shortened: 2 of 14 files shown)


agorrod:
@keithbostic I'd appreciate if you can think about this change. I ran test format with:

`WIREDTIGER_CONFIG="eviction=(max_cache_wait=10000)" ./t`

Which seems to work OK.

test/format/ops.c
```diff
@@ -523,7 +524,14 @@ begin_transaction(TINFO *tinfo, WT_SESSION *session, u_int *iso_configp)
...
 
-	testutil_check(session->begin_transaction(session, config));
+	/*
+	 * Keep trying to start a new transaction if it's timing out - we
+	 * know there aren't any resources pinned so it should succeed
+	 * eventually.
+	 */
+	while ((ret = session->begin_transaction(
+	    session, config)) == WT_CACHE_FULL) {}
+	testutil_check(ret);
```
agorrod: The alternative would be to not return an error from begin_transaction - but that would hobble the concept. It's nice to know that threads will *never* wait forever..

test/format/lrt.c
```diff
@@ -110,8 +110,14 @@ lrt(void *arg)
...
-			testutil_check(session->begin_transaction(
-			    session, "snapshot=test"));
+			/*
+			 * Keep trying to start a new transaction if it's
+			 * timing out - we know there aren't any resources
+			 * pinned so it should succeed eventually.
+			 */
+			while ((ret = session->begin_transaction(
+			    session, "snapshot=test")) == WT_CACHE_FULL) {}
+			testutil_check(ret);
```
keithbostic: KNF, the rest of the code does:
```
while ((ret = 
    session->begin_transaction(session, config)) == WT_CACHE_FULL)
        ;
```

test/format/lrt.c
```diff
@@ -123,8 +129,9 @@ lrt(void *arg)
 			 * positioned. As soon as the cursor loses its position
 			 * a new snapshot will be allocated.
 			 */
-			testutil_check(session->begin_transaction(
-			    session, "isolation=snapshot"));
+			while ((ret = session->begin_transaction(
+			    session, "snapshot=snapshot")) == WT_CACHE_FULL) {}
+			testutil_check(ret);
```
keithbostic: Same KNF.

test/format/ops.c
keithbostic: Same KNF.

keithbostic: I agree `session.begin_transaction` should return the error.

keithbostic:
@agorrod, I'm torn about this change.

I think it means we'll no longer see and/or chase problems where we've configured sufficient cache, but for whatever reason, application threads are stalling in eviction, in other words, we're going to miss problems.

That said, defining "sufficient cache" is not easy to do, and we invest a fair amount of wasted effort chasing failures that aren't real, we simply didn't get the cache size calculation correct for all of the possible permutations of the workload.

I don't feel strongly about the change, I'm happy to defer to you & @michaelcahill.

keithbostic:
@agorrod, two other comments:

First, I'm guessing we want to set the new configuration for LSM format runs?

Second, we might want to rename `WT_CACHE_FULL` to `WT_OPERATION_STALL`, that way we could generalize the error if there's ever a case where it's not the cache that's the problem, and applications wouldn't have to change their error handling. (I haven't thought about that much, feel free to just reject it out of hand, just something that occurred to me while I was looking at this change.)

src/conn/conn_cache.c
```diff
@@ -143,6 +143,11 @@ __cache_config_local(WT_SESSION_IMPL *session, bool shared, const char *cfg[])
 	conn->evict_threads_max = evict_threads_max;
 	conn->evict_threads_min = evict_threads_min;
 
+	/* Retrieve the wait time and convert from milliseconds */
+	WT_RET(__wt_config_gets(session,
+	    cfg, "cache_max_wait", &cval));
+	cache->cache_max_wait_us = (uint32_t)cval.val * WT_THOUSAND;
+
```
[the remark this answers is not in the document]
keithbostic: I think the assignment is going to give you a truncation warning on some compilers, I think:
`cache->cache_max_wait_us = (uint32_t)(cval.val * WT_THOUSAND);` is safer.

dist/api_data.py
```diff
@@ -417,6 +417,11 @@ def __cmp__(self, other):
         maximum heap memory to allocate for the cache. A database should
         configure either \c cache_size or \c shared_cache but not both''',
         min='1MB', max='10TB'),
+    Config('cache_max_wait', '0', r'''
+        the maximum number of milliseconds an application thread will wait
+        for space to be available in cache before giving up. Default will
+        wait forever''',
+        min=0),
```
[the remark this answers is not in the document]
agorrod: I'll add `_ms` - I tossed a coin in the first place.

Regards the default: I'd like to change it eventually, but I think we need internal experience and a plan for telling our users about the change before we could turn this on by default.

test/format/lrt.c
agorrod: That line wrapping doesn't work, but I've changed from `{}` to `;`

agorrod:
bq. I think it means we'll no longer see and/or chase problems where we've configured sufficient cache, but for whatever reason, application threads are stalling in eviction, in other words, we're going to miss problems.

I agree - as your subsequent comment mentioned, I reached a tipping point when chasing LSM failures. Primarily because we don't rollback read-only transactions, and read-only transactions can pin a lot of content for LSM trees. This approach was more appealing than returning roll back for read only transactions.

OTOH: the goal here is to avoid cases where genuinely pinned content is causing stalls - since IMO that's an application configuration issue. I think we'll still see and chase cases where other factors cause the cache to become stuck full.

agorrod:
> Second, we might want to rename WT_CACHE_FULL to WT_OPERATION_STALL, that way we could generalize the error if there's ever a case where it's not the cache that's the problem, and applications wouldn't have to change their error handling.

I feel like we should have a broader conversation about error codes.. Maybe we can do that in a standup meeting next week?

agorrod:
Thanks for fixing this up Keith.

michaelcahill:

PR #4142, opened 2018-06-21
WT-4024 Fix a race between split and next/prev

keithbostic:


keithbostic:
@agorrod, @michaelcahill, I think this one is ready to go.

I think the code is simpler and generally less fragile, but it's going to need a significant amount of burn-in before release.

I'm continuing to stress test this on the PPC, that's the only place the errors reproduce.

agorrod:
I switched some Jenkins stress testing to run on this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer-ppc/configure

keithbostic:
One performance change I should have called out: cursor movements in this branch that have to be restarted are restarted from the original starting point of the movement, rather than from an intermediate page to which we hazard-pointer coupled successfully. That's necessary for cursors walking backwards through the tree, but not for cursors moving forwards through the tree. I did it for both forward and backward movements for simplicity, but we should watch for performance regressions.

michaelcahill:
I've stared at this for a while and I think I'm good with it.  There is some nice simplification in moving the prev-race check out of the generic `__wt_page_swap`, and the overall approach of going back to the original page whenever there is a race makes sense to me.

@keithbostic, are there any performance tests we could run before merging to check for regressions?

Otherwise, I'll look to merge tomorrow morning.

keithbostic:
> are there any performance tests we could run before merging to check for regressions?

I think the interesting focus would be cursor-forward movements in the face of a heavy-split workload, and I'm not sure if we have any of those MongoDB workloads around.

I did a bit of of micro-benchmarking of the walk code without seeing any significant slowdown, and the timing of the failing CONFIGs before & after the change didn't seem to change (but that testing has timing stress configurations, and might not be applicable).

I shut off my correctness testing on Friday -- I haven't seen any failures since these changes went in, so as far as I'm concerned, this is ready for merge.

I'd like to see a lot of burn-in before we release it, though.

michaelcahill:
Thanks @keithbostic -- lgtm.  I'll merge.

PR #4143, opened 2018-06-21
WT-4139 rename the cursor restart statistic to match implementation

keithbostic:


agorrod:
Thanks @keithbostic lgtm - FWIW: I try to sort statistics by their string descriptions not the name, which is the cause for some of the order shuffling, but I don't feel strongly.

keithbostic:
@agorrod 
> FWIW: I try to sort statistics by their string descriptions not the name, which is the cause for some of the order shuffling, but I don't feel strongly.

I never look them up by descriptions, I always look them up by name, didn't occur to me people were doing the reverse.

PR #4144, opened 2018-06-21
WT-4140 Cursor walk marks the wrong internal page for quick eviction

keithbostic:


keithbostic:
This is wrong -- removing the branch.

PR #4146, opened 2018-06-25
WT-4104 Fix test/format failure during comparing data content with berkeley db. (Not ready for review)

bvpvamsikrishna:


michaelcahill:
On reviewing in a little more detail, since the original failure was not detected by comparing dumps, we would need similar fixes around `ops:1275`, `ops.c:1477`, `ops.c:2075` and `ops.c:2109` (at least).

In short, I think the `SINGLETHREADED` condition that is repeated through `test/format` is insufficient for deciding whether to use Berkeley DB.  Instead, we should use the condition:
```c
g.use_berkeleydb = SINGLETHREADED && (!g.c_salvage || (g.c_delete_pct == 0 && g.c_truncate == 0));
```

test/format/salvage.c
```diff
@@ -164,14 +164,7 @@ wts_salvage(void)
...
 
-	/*
-	 * If no records were deleted, dump and compare against Berkeley DB.
-	 * (The problem with deleting records is salvage restores deleted
-	 * records if a page splits leaving a deleted record on one side of
-	 * the split, so we cannot depend on correctness in that case.)
-	 */
-	if (g.c_delete_pct == 0)
-		wts_dump("salvage", SINGLETHREADED);
+	wts_dump("salvage", USE_BERKELEY_DB);
```
michaelcahill: I think we should remove the second parameter to `wts_dump` and instead have `if (!USE_BERKELEY_DB) return;` near the start of that function.

There is no point calling out to the `s_dumpcmp.sh` shell script if Berkeley DB is not enabled.

test/format/ops.c
```diff
@@ -1079,6 +1080,8 @@ ops(void *arg)
 		 * a random direction.
 		 */
 		if (positioned) {
+			testutil_assert(F_ISSET(cursor, WT_CURSTD_KEY_INT));
+
```
michaelcahill: Maybe try moving this assertion into `nextprev` to catch cases where we need `g.dbc_reset`?

`test/format` tries fairly hard to track when it thinks a cursor is positioned -- what I think is happening here is some case where that tracking diverges from reality.

bvpvamsikrishna:

PR #4148, opened 2018-06-28
WT-3917 Update reserve() documentation to have a note around commit visibility

sulabhM:


sulabhM:
The Note is going to look like as follows:

screen shot 2018-06-28 at 15 01 21

agorrod:
lgtm

PR #4150, opened 2018-06-28
WT-4136 Add a new timing stress flag that yields during tree search

keithbostic:


agorrod:
@bvpvamsikrishna please take a look at this PR

agorrod:
lgtm too

PR #4157, opened 2018-07-01
WT-4145 Only include the checkpoint timestamp during checkpoints.

agorrod:
Otherwise the visibility check will compare to zero - and lead to bad decisions.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -444,7 +444,8 @@ __wt_txn_pinned_timestamp(WT_SESSION_IMPL *session, wt_timestamp_t *pinned_tsp)
 	    __wt_timestamp_set(&checkpoint_ts,
 	    &txn_global->checkpoint_timestamp));
 
-	if (__wt_timestamp_cmp(&checkpoint_ts, &pinned_ts))
+	if (!__wt_timestamp_iszero(&checkpoint_ts) &&
+	    __wt_timestamp_cmp(&checkpoint_ts, &pinned_ts))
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic the ticket is assigned to you - but I took a look at the format failures, and suspect this is the root cause.

agorrod:
@bvpvamsikrishna  Could you review this please?

src/include/txn.i
```diff
@@ -444,7 +444,8 @@ __wt_txn_pinned_timestamp(WT_SESSION_IMPL *session, wt_timestamp_t *pinned_tsp)
 	    __wt_timestamp_set(&checkpoint_ts,
 	    &txn_global->checkpoint_timestamp));
 
-	if (__wt_timestamp_cmp(&checkpoint_ts, &pinned_ts))
+	if (!__wt_timestamp_iszero(&checkpoint_ts) &&
+	    __wt_timestamp_cmp(&checkpoint_ts, &pinned_ts))
```
agorrod: The intent is to make this match the transaction ID semantic here:
https://github.com/wiredtiger/wiredtiger/blob/develop/src/include/txn.i#L391

bvpvamsikrishna: I think it is better to use '<= 0' with __wt_timestamp_cmp function as it can return "-1, 0, 1".

agorrod:
Thanks @bvpvamsikrishna could you please go ahead and merge.

PR #4159, opened 2018-07-03
WT-3955 Add function + line number context to error logging

dgottlieb:


wtbuild:

keithbostic:
@agorrod, this change requires format strings in all WT_ERR_MSG/WT_RET_MSG calls (similar to the changes we made to __wt_verbose awhile back). I generally like it, I think it's reasonable to universally have location marks on every error message.

The next step, I think, is @dgottlieb will take a run at verbose messages on naked error returns.

dgottlieb:
@sueloverso @keithbostic sorry for assigning reviewers a bit prematurely. Currently nothing uses the verbose logging flag that the first commit adds. I'm working on another commit that wraps naked returns with some code that will check for verbose error returns.

sueloverso:
Ok. I was about to make that comment that using verbose isn't in there yet. I'll be patient!

sueloverso:
autobuild: test this please

sueloverso:
@dgottlieb it does look like you need to make equivalent changes to `os_win` like you did in `os_posix`.

agorrod:
I don't love how invasive and error prone this would be for the relatively rare case where we want to pass additional parameters to the message error handling macros. I'd rather having:

```
WT_ERR_MSG(session, v, msg) \
    __wt_err(session, error, "%s: %d: " msg)
WT_ERR_MSG_VARG(session, v, msg, ...) \
    __wt_err(session, error, "%s: %d: " msg, __VA_ARGS__);
```

Then existing callers that just want to send a string through don't need to change, and more sophisticated usage switches to using a different macro.

ps: I didn't try compiling that - it's reasonably likely that it won't work quite like it does in my imagination..

keithbostic:
> I don't love how invasive and error prone this would be for the relatively rare case where we want to pass additional parameters to the message error handling macros. I'd rather having:

@agorrod, I agree about the invasiveness (we had the same discussion when we made the same change to `__wt_verbose()` calls), but I don't see where it's error prone? Won't any mistake will result in an immediate compiler error?

keithbostic:
autobuild: test this please

keithbostic:
autobuild: test this please

keithbostic:
autobuild: test this please

keithbostic:
autobuild: test this please

agorrod:
> I don't see where it's error prone? Won't any mistake will result in an immediate compiler error?

You are right - I should have said not obvious, so I'll be getting compilation warnings when doing development.

> I agree about the invasiveness (we had the same discussion when we made the same change to __wt_verbose() calls) Turns out, it's not a relatively rare case. I did a fast pass over calls to WT_ERR_MSG, and I see almost 50% (66 out of 140 pass additional arguments).

That makes me sad. I did a similar exercise for WT_RET_MSG and it's pretty much exactly half there as well.

This seems a bit different to the verbose case to me - in that case there were ~15 out of 260 uses that use a single string as the format specific.

I have a mild preference for a pair of macros (an argv and non-argv version), even if the code changes are going to be similarly invasive. I find the syntax of `WT_ERR_MSG(session, ret, "%s", "The world is ending");` pretty unnatural.  I'll go along with the choice you (Dan, Keith, Sue) make from here.

keithbostic:
I futzed with this more this morning, I can't find a path to any cpp magic that avoids the problem.

> I have a mild preference for a pair of macros (an argv and non-argv version).

I have a medium preference for a single macro: I prefer to avoid having two ways to do anything, I think we'll end up with code that uses the `"%s"` solution as well as code that uses the new macro name, if we support two macros). I won't fight about it.

@sueloverso, @dgottlieb?

sueloverso:
I'd prefer to have a single macro. I also agree that the usage is fairly unnatural. I felt that when we did it for `wt_verbose` and it took a while for me to internalize that but I think I can learn and remember that easier 2nd time around than having to think about multiple macros all the time.

That being said, looking through the places that changed, quite a number could actually be changed to include additional information. Two quick examples, the changes to `block_addr.c` could print out the version expected and what we found, and `block_ckpt.c:718` could print the unexpected number of entries.

dgottlieb:
Agreed with Sue that many of the messages can include useful diagnostic information. I think a single macro would encourage that.

src/support/scratch.c
```diff
@@ -400,16 +400,16 @@ __wt_scr_discard(WT_SESSION_IMPL *session)
 		if (*bufp == NULL)
 			continue;
 		if (F_ISSET(*bufp, WT_ITEM_INUSE))
-			__wt_errx(session,
-			    "scratch buffer allocated and never discarded"
+			__wt_errx(session, "%s",
```
sueloverso: @dgottlieb This is the cause of the error I suspect. This is a new, 2nd call and the if statement needs {}.

keithbostic: This was me, sorry @dgottlieb, fix pushed in c739f36.

keithbostic:
autobuild: retest this please

agorrod:
I'm happy to be outvoted, and it's a good point that encouraging more specific information is good. I'd be happy for this to get merged once it passed testing

sueloverso:
It does not yet use the `WT_VERB_ERROR_RETURNS` and print verbose messages, AFAIK. So it isn't ready for merge until that part of it is committed, right? Otherwise we'd have to open another ticket for that last piece and that does not seem useful.

src/include/wt_internal.h
```diff
@@ -369,14 +368,16 @@ union __wt_rand_state;
...
+#include "extern.h"			/* required by error.h */
 #ifdef _WIN32
 #include "extern_win.h"
 #else
 #include "extern_posix.h"
 #endif
 #include "verify_build.h"
 
+#include "error.h"
+
```
keithbostic: @dgottlieb, is this change still necessary?

src/os_posix/os_fs.c
```diff
@@ -703,12 +703,16 @@ __posix_open_file(WT_FILE_SYSTEM *file_system, WT_SESSION *wt_session,
...
-		    "%s: handle-open: open", name);
+	if (ret != 0) {
+		if (pfh->direct_io)
+			WT_ERR_MSG(session, ret,
+			    "%s: handle-open: open: failed with direct I/O "
+			    " configured, some filesystem types do not support "
+			    "direct I/O", name);
+		else
+			WT_ERR_MSG(session, ret,
+			    "%s: handle-open: open", name);
```
keithbostic: Lines can be joined.

src/include/wt_internal.h
dgottlieb: `error.h` now holds the `__wt_err[x]` macros which call into `__wt_err[x]_func`. Those methods are declared in `extern.h`, so we pulled `error.h` below it.

keithbostic: Yes, but those are #defines, so it doesn't matter?

dgottlieb: You're right, I moved it back and it works.

Which leaves me wondering what the compile problem was in the first place!

dgottlieb:
autobuild: retest this please

keithbostic:
I said I'd try and take a look at this build failure, but I can't get the test suite to run on a spawned Windows host, it seems to just hang after a few hundred tests.

agorrod:
I find that Windows has stronger checking for buffer overuns and memory corruption - so I'd try running the branch using either valgrind or an ASAN build to see if it shows anything up.

I've also reported some findings from Windows on the JIRA ticket

src/include/error.h
```diff
@@ -19,6 +19,15 @@
 #define	WT_DIAGNOSTIC_YIELD
 #endif
 
+#define	__wt_err(session, error, fmt, ...)				\
+	__wt_err_func(session, error, "%s: %d: " fmt,			\
+	    __func__, __LINE__, __VA_ARGS__)
```
agorrod: I don't think an explicit `fmt` is needed. The `__eventv` API already has equivalent func and line parameters. This could just become:
```
#define __wt_err(session, error, ...)        \
__wt_err_func(session, error, __func__, __LINE__, __VA_ARGS__);
```
With a corresponding change to the parameters in `__wt_err_func`.

Then we wouldn't need to change all the callers.

keithbostic: I could have sworn I tried to go down this path, but apparently not! Thank you, @agorrod, I totally missed this. @dgottlieb, sorry to have led you down the wrong path... Here's diff following Alex's point:
```
diff --git a/src/include/error.h b/src/include/error.h
index f94c3d7b8..6cb7ae133 100644
--- a/src/include/error.h
+++ b/src/include/error.h
@@ -19,6 +19,11 @@
 #define        WT_DIAGNOSTIC_YIELD
 #endif

+#define        __wt_err(session, error, ...)                                   \
+       __wt_err_func(session, error, __func__, __LINE__, __VA_ARGS__)
+#define        __wt_errx(session, ...)                                         \
+       __wt_errx_func(session, __func__, __LINE__, __VA_ARGS__)
+
 /* Set "ret" and branch-to-err-label tests. */
 #define        WT_ERR(a) do {                                                  \
        if ((ret = (a)) != 0)                                           \
diff --git a/src/support/err.c b/src/support/err.c
index 0569d0545..3459b6cd1 100644
--- a/src/support/err.c
+++ b/src/support/err.c
@@ -309,13 +309,14 @@ err:              if (fprintf(stderr,
 }

 /*
- * __wt_err --
+ * __wt_err_func --
  *     Report an error.
  */
 void
-__wt_err(WT_SESSION_IMPL *session, int error, const char *fmt, ...)
+__wt_err_func(WT_SESSION_IMPL *session,
+    int error, const char *file_name, int line_number, const char *fmt, ...)
     WT_GCC_FUNC_ATTRIBUTE((cold))
-    WT_GCC_FUNC_ATTRIBUTE((format (printf, 3, 4)))
+    WT_GCC_FUNC_ATTRIBUTE((format (printf, 5, 6)))
     WT_GCC_FUNC_ATTRIBUTE((visibility("default")))
 {
        va_list ap;
@@ -325,18 +326,20 @@ __wt_err(WT_SESSION_IMPL *session, int error, const char *fmt, ...)
         * an error value to return.
         */
        va_start(ap, fmt);
-       WT_IGNORE_RET(__eventv(session, false, error, NULL, 0, fmt, ap));
+       WT_IGNORE_RET(__eventv(session,
+           false, error, file_name, line_number, fmt, ap));
        va_end(ap);
 }

 /*
- * __wt_errx --
+ * __wt_errx_func --
  *     Report an error with no error code.
  */
 void
-__wt_errx(WT_SESSION_IMPL *session, const char *fmt, ...)
+__wt_errx_func(WT_SESSION_IMPL *session,
+    const char *file_name, int line_number, const char *fmt, ...)
     WT_GCC_FUNC_ATTRIBUTE((cold))
-    WT_GCC_FUNC_ATTRIBUTE((format (printf, 2, 3)))
+    WT_GCC_FUNC_ATTRIBUTE((format (printf, 4, 5)))
 {
        va_list ap;

@@ -345,7 +348,8 @@ __wt_errx(WT_SESSION_IMPL *session, const char *fmt, ...)
         * an error value to return.
         */
        va_start(ap, fmt);
-       WT_IGNORE_RET(__eventv(session, false, 0, NULL, 0, fmt, ap));
+       WT_IGNORE_RET(__eventv(session,
+           false, 0, file_name, line_number, fmt, ap));
        va_end(ap);
 }

diff --git a/src/support/scratch.c b/src/support/scratch.c
index 2ead79a1c..6d1bbad9f 100644
--- a/src/support/scratch.c
+++ b/src/support/scratch.c
@@ -400,16 +400,19 @@ __wt_scr_discard(WT_SESSION_IMPL *session)
                if (*bufp == NULL)
                        continue;
                if (F_ISSET(*bufp, WT_ITEM_INUSE))
+#ifdef HAVE_DIAGNOSTIC
                        __wt_errx(session,
                            "scratch buffer allocated and never discarded"
-#ifdef HAVE_DIAGNOSTIC
                            ": %s: %d",
                            session->
                            scratch_track[bufp - session->scratch].file,
                            session->
                            scratch_track[bufp - session->scratch].line
-#endif
                            );
+#else
+                       __wt_errx(session,
+                           "scratch buffer allocated and never discarded");
+#endif
 
                __wt_buf_free(session, *bufp);
                __wt_free(session, *bufp);
```
Dan, maybe it's easiest to drop back to develop, add Alex's change, plus your last commit a256544.

dgottlieb:
Closing this PR and re-opening with the new changes.

PR #4162, opened 2018-07-06
WT-4169 wt verify dump-pages failure

keithbostic:


agorrod:
Thanks @keithbostic this is a nice cleanup. I hesitated because of the complex macro, but I think it's warranted.

PR #4163, opened 2018-07-09
WT-3955 Add function + line number context to error logging

dgottlieb:


sueloverso:
I like this a whole lot better. I cloned this branch and did not see any missing returns or error settings that were not captured. However, I do get compilation errors when building on my Mac in some ifdefed code. I'll send you the details.

src/os_posix/os_fallocate.c
```diff
@@ -32,10 +32,13 @@ __posix_std_fallocate(
...
 #else
+	WT_SESSION_IMPL *session;
+
 	WT_UNUSED(file_handle);
-	WT_UNUSED(wt_session);
 	WT_UNUSED(offset);
-	return (ENOTSUP);
+
+	session = (WT_SESSION_IMPL *)wt_session;
+	return (__wt_set_return(session, ENOTSUP));
```
sueloverso: Although this builds, I think my preference is to avoid the local variable in all of these simple cases:
`return (__wt_set_return((WT_SESSION_IMPL *)wt_session, ENOTSUP));`

dgottlieb: Will do!

agorrod:
lgtm.

PR #4164, opened 2018-07-10
WT-4010 Simplify test/format timestamp handling.

keithbostic:


test/format/util.c
```diff
@@ -608,25 +609,18 @@ timestamp(void *arg)
 	 */
 	while (!g.workers_finished) {
```
michaelcahill: Is this loop still necessary?  What stops us using a "get=all_committed" query (under a lock) to update the oldest timestamp during the run?

If we could do that, then there would be no need to track `tinfo->last_timestamp`.

test/format/ops.c
```diff
@@ -535,60 +533,32 @@ begin_transaction(TINFO *tinfo, WT_SESSION *session, u_int *iso_configp)
...
+		/*
+		 * Prepare will return an error if the prepare timestamp is less
+		 * than any active read timestamp. Lock across allocating read
+		 * timestamps if prepare is configured, skip the read timestamp
+		 * if a prepare is in progress,
+		 */
+		locked = false;
 		if (g.c_prepare) {
-			testutil_check(pthread_rwlock_rdlock(&g.prepare_lock));
+			ret = pthread_rwlock_tryrdlock(&g.prepare_lock);
```
michaelcahill: What's wrong with waiting here?  Is there some operation that can deadlock, or that holds a write lock indefinitely?

test/windows/windows_shim.c
```diff
@@ -104,11 +97,24 @@ pthread_rwlock_unlock(pthread_rwlock_t *rwlock)
...
+	AcquireSRWLockShared(&rwlock->rwlock);
+	return (0);
+}
+
 int
 pthread_rwlock_trywrlock(pthread_rwlock_t *rwlock)
 {
 	if (TryAcquireSRWLockExclusive(&rwlock->rwlock)) {
-		rwlock->exclusive_locked = GetCurrentThreadId();
+		rwlock->exclusive_locked = 1;
```
michaelcahill: Tracking the thread that has exclusive access seems like useful information: why did you take that out?

test/format/util.c
keithbostic: I agree we could acquire a lock and use `get=all_committed`, but I think that would have to serialize with transaction-begin and transaction-prepare. I don't have data to show it, but it seems that might result in less thread independence and worse testing.

Just let me know if you prefer the lock.

test/format/ops.c
keithbostic: The reason not to wait here is waiting serializes transaction-begin with transaction-prepare, and it seemed potentially interesting to mix-and-match reads with and without timestamps. I don't have any data to show it's worth doing.

Just let me know if you prefer to always wait.

test/windows/windows_shim.c
keithbostic: We were mixing-and-matching `DWORD` and `int` types and we never used the value. If it's potentially interesting, I'll revert that part of the change and fix the type declaration instead.

Pushed in e1472a1.

keithbostic:
> my observation is "stable_timestamp" not used in test/format. Any specific reason ?

I don't think the subject has ever come up -- absent other issues, could the stable timestamp just track the oldest timestamp in format?

test/format/util.c
michaelcahill: @keithbostic sorry for letting this one sit.  My preference is to try to get all of the timestamp tracking out of test code, so can you please try this with a lock and see whether getting rid of `tinfo->last_timestamp` is feasible?

test/format/ops.c
michaelcahill: I'd prefer to wait here, then schedule a ticket to allow `WT_SESSION::timestamp_transaction` to set the prepare timestamp, and just do that part under the lock.

keithbostic:
@michaelcahill, changes pushed in 27b5a08.

Just after I pushed that change, however, I realized this branch breaks the long-running-transaction code (that is, any CONFIG with `long_running_txn=1`), even when `WT_SESSION.begin_transaction` is locked. The error is:
```
WT_SESSION.begin_transaction: __wt_txn_named_snapshot_get, 336:
Named snapshot 'snapshot' not found: Invalid argument
```

I'll figure that out tomorrow if it's not obvious to you from the error message.

test/format/lrt.c
```diff
@@ -115,10 +115,12 @@ lrt(void *arg)
 			 * timing out - we know there aren't any resources
 			 * pinned so it should succeed eventually.
 			 */
+			testutil_check(pthread_rwlock_wrlock(&g.txn_lock));
```
michaelcahill: My understanding of the lock is that it prevents races between timestamps being allocated and the oldest timestamp catching up to current.  Since these calls don't use timestamps, why do they need locking?

test/windows/windows_shim.c
```diff
@@ -104,6 +97,19 @@ pthread_rwlock_unlock(pthread_rwlock_t *rwlock)
 	return (0);
 }
 
+int
+pthread_rwlock_tryrdlock(pthread_rwlock_t *rwlock)
+{
+	return (TryAcquireSRWLockShared(&rwlock->rwlock) ? 0 : EBUSY);
+}
```
michaelcahill: Can this be removed?

michaelcahill:
I believe `long_running_txn` mode is broken on develop as well -- I'm not sure what that code is supposed to do but it's dropping all named snapshots then trying to use one, so I'd expect that to fail.

test/format/ops.c
```diff
@@ -525,68 +522,33 @@ begin_transaction(TINFO *tinfo, WT_SESSION *session, u_int *iso_configp)
 	*iso_configp = v;
 
 	/*
-	 * Keep trying to start a new transaction if it's timing out - we
-	 * know there aren't any resources pinned so it should succeed
-	 * eventually.
+	 * Lock out the oldest timestamp update and single-thread transaction
+	 * begin and prepare.
+	 */
+	testutil_check(pthread_rwlock_wrlock(&g.txn_lock));
```
michaelcahill: I only expected locking around allocating the timestamp and calling `timestamp_transaction`, not around the `begin_transaction`.  Does that not work for some reason?

test/format/lrt.c
keithbostic: Agreed, pushed in 043fd79.

test/windows/windows_shim.c
keithbostic: It can, but it's reasonable to leave it in, some utility might want it someday?

test/format/ops.c
keithbostic: Agreed, pushed in 043fd79.

keithbostic:
@michaelcahill, just to close the loop, as far as I know, this one is ready to merge.

michaelcahill:
Thanks @keithbostic, I'm happy with how this turned out.  I'll merge...

PR #4147, opened 2018-06-27
WT-4141 Allow dirty eviction during checkpoint as of a timestamp

michaelcahill:

PR #4166, opened 2018-07-11
WT-4160 Restore performance when timestamps are built but not in use.

sueloverso:
@michaelcahill Please review this change that modifies some of the changes made in WT-4141 that resulted in a perf drop for the overflow test. These are things I saw eyeballing that diff in #4147 that can affect a timestamp build that is not actually using timestamps. Here are the things to know:
1. The change to `rec_write.c` does NOT impact performance, but it feels like a correct change and avoids calling the function when timestamps are built but not being used.
2. The real change is in `btree.i`. In particular, the pre-WT-4141 code would return `true` in the case that timestamps were not in use. Restoring that via the `else` restores performance in all my local tests.
3. The other change in `btree.i` is a bug in the original code, IMO. The code gets the `pinned_ts` in the line before, but then never uses it.

### DIFF
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -1357,13 +1357,11 @@ __wt_page_evict_retry(WT_SESSION_IMPL *session, WT_PAGE *page)
 		return (true);
 
 #ifdef HAVE_TIMESTAMPS
-	if (!__wt_timestamp_iszero(&mod->last_eviction_timestamp)) {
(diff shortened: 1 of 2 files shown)


michaelcahill:
Thanks @sueloverso, I made a cosmetic change and I'll merge once it's done testing.

Nice catch!  lgtm.

PR #4167, opened 2018-07-11
WT-4171 Enabling tree walk timing stress causes excessive slowdown

keithbostic:


keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
@agorrod, slowing down the calls to `WT_DIAGNOSTIC_YIELD` by changing them to sleep, even for very short periods of time, slows down the test suite by a lot, so I've reverted that change.

Since I'm keeping `WT_DIAGNOSTIC_YIELD`, I (hopefully) solved the original problem of tree descent taking too long in stress tests by changing the tree descent code to call `WT_DIAGNOSTIC_YIELD` instead of `__wt_timing_stress`, that way we'll yield, but not sleep.

I left in the change to make `__wt_timing_stress` more likely to sleep a short time than a long time, I think that's a reasonable change to make to minimize the degradation caused by the split-code diagnostic tests.

If you agree, I suggest we declare victory and open another ticket for any further changes we want.

agorrod:
Thanks @keithbostic this approach makes a lot of sense to me. I'll go ahead and merge once it passes testing.

PR #4168, opened 2018-07-11
WT-4179 Expose WiredTiger crc32c functions

keithbostic:
@agorrod, this approach is intended to make it possible to build the checksum functions in a separate library from the WiredTiger library: in short, we don't use any WiredTiger include files or structures.

Returning a pointer to a function is intended to minimize the wrapping required for C++.


agorrod:
@keithbostic the approach looks fine to me, if it satisfies Mathias and passes testing I'm happy for you to merge.

keithbostic:
autobuild: retest this please

PR #4169, opened 2018-07-12
WT-3943 Display full error message on python test failure

nehakhatri5:
When there is an unexpected content in stderr, python test treats it
as a test failure and displays the content of stderr on console.
This content on console used to be truncated. This change elongates
the stderr content displayed on console to ease failure triage.

### DIFF
--- a/test/suite/wttest.py
+++ b/test/suite/wttest.py
@@ -101,7 +101,7 @@ class CapturedFd(object):
                                      ' unexpected ' + self.desc +
                                      ', contains:\n"' + contents + '"')
             testcase.fail('unexpected ' + self.desc + ', contains: "' +
-                      shortenWithEllipsis(contents, 1000) + '"')
+                      contents + '"')
         self.expectpos = filesize
 
     def checkAdditional(self, testcase, expect):
--- a/test/suite/wttest.py
+++ b/test/suite/wttest.py
@@ -101,7 +101,7 @@ class CapturedFd(object):
                                      ' unexpected ' + self.desc +
                                      ', contains:\n"' + contents + '"')
             testcase.fail('unexpected ' + self.desc + ', contains: "' +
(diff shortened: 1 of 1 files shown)


agorrod:
Is there a reason to shorten the content at all? Why not just print the entire contents out?

nehakhatri5:
I think that error messages might not exceed 1000 characters, so kept it to that limit.

agorrod:
What is the benefit to keeping a limit?

nehakhatri5:
I do not see any benefit of keeping it limited.

agorrod:
lgtm

sueloverso:
Although this is merged, I very often use `wt_errx(...)` to add debugging to a failure. The test then fails when it detects content in `stderr.txt` and I go edit that file to follow the debugging paths. Having a potentially large amount of content spewed to the screen is not helpful in the debugging case. I agree with @nehakhatri5 that a real error/failure will not exceed 1000 characters or some similar number. But when using that mechanism to debug it can become a problem.

agorrod:
The counterpoint is that, when a test runs via Jenkins or Evergreen the content of std*.txt is often not available, but the content output to the console is. If we truncate output that goes to the console there are situations where we can't get access to that later.

nehakhatri5:
I have seen that std*.txt is available after both Jenkins and Evergreen failures, so I believe that the error message could be retrieved from the failure artefacts.

agorrod:
For Jenkins in particular those files are removed as soon as another job is started - so it is only possible to retrieve their contents if you look at the failure immediately.

sueloverso:
Ah I had not thought of the Evergreen console aspect. This is merged anyway, so let's see how it goes. If it seems like too much for general development then we can reconsider then. Maybe after a time or two getting used to it, just having it on the screen will be easier than having to dig out the std*.txt files.

PR #4170, opened 2018-07-13
WT-4147 Detect and salvage corruption of log record size in log file.

ddanderson:


ddanderson:
This is not yet ready for review/merge!  The simple change has exposed a problem with the approach I tried, I'll comment in JIRA about it.

ddanderson:
This is closed in preference to https://github.com/wiredtiger/wiredtiger/pull/4177

PR #4172, opened 2018-07-17
WT-4135 Test github2jira

lukech:
Here is the first comment.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -1640,3 +1640,5 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
 
 	return (0);
 }
+
+// test


nehakhatri5:
I have read the ticket and certainly this change is very simple. However is there any information about impact in the areas Alex mentioned such as compact? We already have the file_extend configuration knob on wiredtiger_open. Perhaps documentation could be used for this instead of the code change. Does file_extend=(data=0) work or show the same performance as removing it entirely?

nehakhatri5:
@sulabhM, option 2 makes sense to me.                                        
                                                                                 
To make sure I understand, column store pages currently either have a fixed key range when the page is created, or are the last page in the tree with an append list.  Now that the last page can transition to not being at the end by still having an append list, we need to know the key range it finishes at.
                                                                                 
One question: can we infer the last record number during reconciliation by looking in the parent page at `pindex[slot + 1]`?

sulabhM:
@nehakhatri5 :                                                                  
                                                                                 
> To make sure I understand, column store pages currently either have a fixed key range when the page is created, or are the last page in the tree with an append list. Now that the last page can transition to not being at the end by still having an append list, we need to know the key range it finishes at.
                                                                                 
Yes.                                                                             
                                                                                 
> One question: can we infer the last record number during reconciliation by looking in the parent page at pindex[slot + 1]?
                                                                                 
I wondered about that, too, but splits/eviction scared me, I don't think we're guaranteed the page we're reconciling isn't the last entry in its parent's page-index, in which case we'd have to climb the tree and find the "next" page in the tree. We could do that, but it felt messy. Let me know if I'm over-complicating this, I'd much rather take the value from the "next" page than maintain a separate value in the page-modify structure.

nehakhatri5:
@sulabhM Do you have any thoughts about how we might be able to track down the remaining failure? Can we construct a test case that is going to be more likely to trigger a failure? Would it help to put in more aggressive error checking code in diagnostic mode?

sulabhM:
@nehakhatri5 , no ideas yet -- the symptom is an internal page with the wrong starting record number (which is the bug I thought the last set of changes fixed). Anyway, I'll get back to this one and chase it down next chance I get.

sulabhM:
@nehakhatri5 , found/fixed the problem in column-store splits, commit is 70ef762.     
                                                                                 
It's a problem in the develop branch, it just wasn't ever triggering without column-store splits.

sulabhM:
@lukech , @nehakhatri5 : this branch has now passed a few stress runs, and I think it's appropriate to consider it for merging again.
                                                                                 
I also believe this branch needs relatively careful review: there are significant changes in the split code, the cursor-prev code, and the reconciliation code. There aren't many changes by line count, but they're in sensitive places.

nehakhatri5:
@sulabhM , I've made some comments but basically this lgtm.  It would be good to get it merged tomorrow if possible.

src/txn/txn.c
```diff
@@ -1640,3 +1640,5 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
 
 	return (0);
 }
+
+// test
```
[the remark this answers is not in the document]
sulabhM: > I really dislike nested ternery statements. Can this be untangled?             
                                                                                 
Sure, done.

src/txn/txn.c
```diff
@@ -1640,3 +1640,5 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
 
 	return (0);
```
sulabhM: @nehakhatri5 Do you think we can enhance this portion?

nehakhatri5: How could this be set?  We start with zeroed memory and only set the fields we want.                                                                                 
I understand wanting the comment, maybe it can go near line 1716?

PR #4175, opened 2018-07-18
WT-4179 Expose WiredTiger crc32c functions

keithbostic:
@agorrod, this change is probably worth consideration.

I missed it in the first round of changes, and what it means is we can no longer use autoconf configuration to turn off hardware checksum acceleration in WiredTiger. I went back through the original checksum tickets, and I didn't find any reason why we wanted that ability, but I wanted to call it out, just in case.


keithbostic:
autobuild: retest this please

PR #4176, opened 2018-07-19
WT-4182 Use conservative approach for log checksum errors.

ddanderson:
A checksum error in a log does not necessarily imply corruption. Any log entry that was only partially written will result in a checksum mismatch.  This change reverts a selected portion of the change for WT-3270 that basically marked the end of the log to be recovered unconditionally as soon as a checksum error happened.

Certain test scenarios unfortunately need to be watered down as well.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -2316,7 +2316,9 @@ advance:
 			WT_STAT_CONN_INCR(session, log_scan_rereads);
 		}
 		/*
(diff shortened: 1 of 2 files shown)


ddanderson:
@agorrod, can you please review?  This is the narrow fix we were talking about. I'll have a somewhat better approach for log file corruption in changes for WT-4147.

agorrod:

PR #4178, opened 2018-07-20
WT-4183 Add verbose option to log more messages on error returns

keithbostic:
Review WiredTiger error returns for `WT_ERROR`, `WT_PANIC` and `WT_RUN_RECOVERY`, ensure the appropriate places include verbose error messages.

### DIFF
--- a/src/btree/bt_slvg.c
+++ b/src/btree/bt_slvg.c
@@ -1770,7 +1770,8 @@ __slvg_row_trk_update_start(
 	 * would have discarded it, we wouldn't be here.  Therefore, this test
 	 * is safe.  (But, it never hurts to check.)
 	 */
-	WT_ERR_TEST(!found, WT_ERROR);
+	if (!found)
+		WT_ERR_MSG(session, WT_ERROR, "expected on-page key not found");
 	WT_ERR(__slvg_key_copy(session, &trk->row_start, key));
 
 	/*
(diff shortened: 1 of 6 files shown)


agorrod:
@bvpvamsikrishna

bvpvamsikrishna:
@keithbostic 
Code changes in this PR lgtm for me.
I was checking for messages WT_PANIC and I encountered this instance.  In  [__wt_optrack_record_funcid](https://github.com/wiredtiger/wiredtiger/blob/03ad4b47e006a35085fcf8729ba7cd54d54d576d/src/optrack/optrack.c#L16), we have statement `err:		WT_PANIC_MSG(session, ret, "%s", __func__);`.
May be this can have a more descriptive message. 

My main concern in this function is statements

 `WT_ERR(__wt_scr_alloc(session, strlen(func) + 32, &tmp));`
`__wt_spin_lock(session, &conn->optrack_map_spinlock);`

and the error path is 

`err:		WT_PANIC_MSG(session, ret, "%s", __func__);`
          `}`
          `__wt_spin_unlock(session, &conn->optrack_map_spinlock);`
          `__wt_scr_free(session, &tmp);`

when `__wt_scr_alloc` fails, we will call `__wt_spin_unlock` without `__wt_spin_lock`.

Since this is panic error which results in WiredTiger shutdown, this will not cause any harm, but I think this can be cleaned up. 
Request for your opinion.

keithbostic:
@bvpvamsikrishna 
> May be this can have a more descriptive message.
> when __wt_scr_alloc fails, we will call __wt_spin_unlock without __wt_spin_lock.

Agreed, and nice catch!

I've pushed fixes for both of those in db1fb6b.

It also occurred to me I never checked for uses of `__func__` that we no longer need (because `__wt_err` and `__wt_errx` always include a function name. I've cleaned up two cases in fee2ebb.

bvpvamsikrishna:
@keithbostic Thank you.

agorrod:
Thank @keithbostic lgtm

PR #4179, opened 2018-07-20
WT-4187 Coverity: unused value complaints

keithbostic:


agorrod:
@sulabhM please review this change

test/format/t.c
```diff
@@ -104,7 +104,6 @@ main(int argc, char *argv[])
 		default:
 			usage();
 		}
-	argc -= __wt_optind;
```
sulabhM: Was this supposed to be the removal of the next line here.

keithbostic: > Was this supposed to be the removal of the next line here.

Please check me to be sure, but I think this is correct, that is, `argc` is not subsequently used and so this assignment is unnecessary, `argv` is subsequently used and its assignment should remain.

PR #4183, opened 2018-07-25
WT-4196 Make log corruption checking work regardless of the machine byte order

ddanderson:


ddanderson:
@agorrod, can you please review?  I'm presuming that this will fix WT-4196 on zseries.  Do we have a way to try the zseries jobs against a pull request? It's possible that WT-4197 is related.

lukech:
I just switched the branch to `wt-4196-log-byte-order` for job wiredtiger-test-format-stress-zseries and started a run http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-zseries/43941/

agorrod:
Thanks @ddanderson this change looks good to me - assuming it passes testing.

@sueloverso is more familiar with the code though - so it'd be good if she can review as well.

@lukech thanks for switching the testing over - it's looking better with this change.

src/log/log.c
```diff
@@ -2452,24 +2462,12 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
...
 		buf->size = reclen;
 		logrec = (WT_LOG_RECORD *)buf->mem;
-		WT_ERR(__log_record_verify(session, log_fh, rd_lsn.l.offset,
-		    logrec, &corrupt));
-		if (corrupt) {
-			need_salvage = true;
-			WT_ERR(__log_salvageable_error(session, log_fh->name,
-			    "", rd_lsn.l.offset));
-		}
-
```
sueloverso: While this all looks fine to me, to be clear, the original code was verifying the header on every record read in. Now this fix will only verify the header when there is a checksum failure.

ddanderson: That's correct.  I needed to reorder the code anyway (can't byte swap before computing the checksum), and I realized that the reason for the header checks is a guard against corruption.  If there is no checksum failure, there isn't corruption. Given that the header check now needed to be more involved (copy the header, and byte swap it), it seemed silly to make the check when we didn't really need it.

PR #4184, opened 2018-07-25
WT-4195 when encountering an illegal value, log the value that failed

keithbostic:
@agorrod, here are a set of changes to clean up and enhance our use of `__wt_illegal_value()`.

The first set are all cleaning up places where we called `__wt_illegal_value()` and it didn't make a lot of sense, basically, places that passed an additional message to `__wt_illegal_value()` so an error message could be logged. For most of those, I replaced the call with some variant of the `WT_PANIC` macro with an error message. I did those as separate commits so it would be easier to review the changes, that is, each change that needed a different fix got its own commit.

Once that was done, I removed the additional message being passed to `__wt_illegal_value()`, and added a "failing value" argument that's now displayed by the function.

### DIFF
--- a/dist/log.py
+++ b/dist/log.py
@@ -331,7 +331,7 @@ for optype in log_data.optypes:
 })
 
 tfile.write('''
-\tWT_ILLEGAL_VALUE(session);
+\tWT_ILLEGAL_VALUE(session, optype);
 \t}
 
 \treturn (0);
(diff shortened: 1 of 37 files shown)


keithbostic:
autobuild: retest this please

agorrod:
Thanks @keithbostic lgtm

PR #4186, opened 2018-07-25
WT-4198 Some supported MongoDB architectures don't support crc32 hardware

keithbostic:


agorrod:
Thanks @keithbostic lgtm.

PR #4188, opened 2018-07-26
WT-4199 Fix log corruption errantly reported due to buffer overrun

ddanderson:


ddanderson:
@sueloverso, can you please review this fix?

src/log/log.c
```diff
@@ -1904,9 +1904,9 @@ __log_has_hole(WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t log_size,
 			 * hole.  We can do some simple data verification to
 			 * ensure that this part of the file is not corrupted.
 			 */
-			for (p = buf; remainder > 0;
-			     remainder -= (wt_off_t)rdlen, p += rdlen) {
-				rdlen = WT_MIN(WT_LOG_ALIGN, (size_t)remainder);
+			for (p = buf, buf_left = rdlen; buf_left > 0;
+			     buf_left -= rdlen, p += rdlen) {
+				rdlen = WT_MIN(WT_LOG_ALIGN, buf_left);
```
sueloverso: I've been reading through this and thinking about this and understand it better. Although it isn't changed in this diff, I'm thinking `WT_LOG_ALIGN` should be `log->allocsize`. That is the unit that log records are aligned on. So if someone is using a non-standard log alignment, it should use that. (Might be good to add a non-standard log allocation size in the corruption tests too.)

src/log/log.c
```diff
@@ -1904,9 +1904,9 @@ __log_has_hole(WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t log_size,
 			 * hole.  We can do some simple data verification to
 			 * ensure that this part of the file is not corrupted.
```
sueloverso: I suggest getting rid of the "We can do..." sentence here. This code right here is only walking within the buffer to find the non-zero part. We're not verifying here yet.

src/log/log.c
```diff
@@ -1916,7 +1916,7 @@ __log_has_hole(WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t log_size,
 			 * present.  This record may also be partially written.
```
sueloverso: I think this comment here needs more based on my new understanding of this code today. What we're NOT doing is determining whether the rest of the file is corrupted. What we ARE doing is looking at the first non-zero bit that made us decide we have a hole and determining if that first non-zero part is a valid log record header (i.e. it is a hole) or if the first non-zero part is corruption/flipped bits, etc. I suggest:
/*
 * A presumed log record begins here where the buffer becomes non-zero. If we have
 * enough of a log record present in the buffer, we either have a valid header or corruption.
 * Verify the header of this record to determine whether it is just a hole or corruption.
 */

sueloverso:
LGTM merge whenever you're ready

ddanderson:
Thanks, @sueloverso, I'll merge.

PR #4189, opened 2018-07-27
WT-4206 Fix error handling in cursor close routines

keithbostic:


keithbostic:
@ddanderson, there's a specific change I'd like you to look at, `src/cursor/cur_file.c`.

There's a problem where `CURSOR_API_CALL_PREPARE_ALLOWED` has failed, but `__wt_cursor_cache_release` succeeds, setting `released` to true. I'm not sure what we should be doing there.

I think I made it better, but I'm not sure if there's a better way to handle the problem.

src/cursor/cur_file.c
```diff
@@ -469,15 +469,20 @@ __curfile_close(WT_CURSOR *cursor)
...
-	released = false;
+err:
 
 	/*
-	 * If releasing the cursor fails in any way, it will be left
-	 * in a state that allows it to be normally closed.
+	 * If releasing the cursor fails in any way, it will be left in a state
+	 * that allows it to be normally closed.
 	 */
 	WT_TRET(__wt_cursor_cache_release(session, cursor, &released));
```
ddanderson: @keithbostic, I don't think we should be caching a cursor in an error path.  I pushed an alternative branch for your consideration.

src/cursor/cur_file.c
```diff
@@ -469,15 +469,20 @@ __curfile_close(WT_CURSOR *cursor)
 
 	cbt = (WT_CURSOR_BTREE *)cursor;
 	CURSOR_API_CALL_PREPARE_ALLOWED(cursor, session, close, cbt->btree);
-	released = false;
```
ddanderson: Does remove this explicit assignment open us up to complaints from code analyzers?  If we want it after all, it should be after the `err:` label.

keithbostic: > Does remove this explicit assignment open us up to complaints from code analyzers?

There are two reasons to remove that line: first, it's not necessary, because by definition every called function that has a returned side-effect should immediately set the side-effect to its default at the top of the function. Admittedly, that's coding practice, but I think it's important.

Second, if you remove that line, and the called function doesn't set the side-effect correctly, you'll get a complaint from the static analyzer because it's not necessarily initialized.

So, I'd argue it's cleaner to remove it and leads to better static analysis results.

ddanderson: Sounds reasonable.  With the merged changes, the changes to `cur_file.c` LGTM

michaelcahill:
I find the `err:` labels early in close routines (with blank lines afterwards) a little odd looking, but maybe that's the intent?

Regardless, I understand the need for close to free resources after errors, so these changes lgtm.  I'll merge.

PR #4190, opened 2018-07-30
WT-4206 cursor close routines can fail, leaking resources

ddanderson:


src/cursor/cur_file.c
```diff
@@ -471,17 +471,15 @@ __curfile_close(WT_CURSOR *cursor)
...
 		/*
-		 * XXX
-		 * We're ignoring there might have already been an error.
+		 * If releasing the cursor fails in any way, it will be left in
+		 * a state that allows it to be normally closed.
 		 */
-		goto done;
+		WT_TRET(__wt_cursor_cache_release(session, cursor, &released));
+		if (released)
+			goto done;
```
keithbostic: Static analyzers are likely to complain about the use of `WT_TRET` there, because `ret` is known to be 0. I'd go with `ret = __wt_cursor_cache_release()` just to make sure we never get a complaint.

keithbostic:
autobuild: retest this please

PR #4191, opened 2018-07-30
WT-4208 tree walks can be interrupted by locked internal pages

keithbostic:


src/btree/bt_walk.c
```diff
@@ -414,24 +414,41 @@ __tree_walk_internal(WT_SESSION_IMPL *session,
...
+				ret = __wt_page_swap(session,
+				    couple, ref, WT_READ_NOTFOUND_OK | flags);
+				if (ret == 0) {
+					/* Success, "couple" released. */
+					couple = NULL;
+					*refp = ref;
+					goto done;
+				}
+
+				 WT_ASSERT(session, ret == WT_NOTFOUND);
```
michaelcahill: Is this assertion supposed to stay long term?  Can't we get `WT_RESTART` or `EBUSY` here (at least)?

michaelcahill: Sorry, I see the comment about restart above: still curious if WT_NOTFOUND is really the only possible case.

keithbostic: I don't think `WT_RESTART` should happen here (and I don't see the path where `EBUSY` matters, can you explain further what you're thinking?)

So, I was planning to leave the assert in, just so we know if it ever happens. I think the `WT_ERR_NOTFOUND_OK` macro should do the right thing in non-diagnostic builds, so we won't fail in the field.

michaelcahill: I'm also concerned about spurious failures in testing, but I'm happy to merge and see what happens.

keithbostic: > I'm also concerned about spurious failures in testing

Agreed. I kind of want to see what happens, though, this problem could explain some weirdness in eviction because it could prevent us from evicting big parts of a busy tree.

PR #4192, opened 2018-07-31
WT-3856 Add schema test to truncate and recover from every LSN.

sueloverso:
I created a new Python test that does some schema operations, walks the log, gathering LSNs and then truncates the log to each of those offsets and then runs recovery in each of those.

I'm slightly unsatisfied with it. @keithbostic was correct in the ticket that detailed verification is difficult as we don't know what the system should look like at any given LSN. I copy the database and then truncate the log, but I don't know whether, at that LSN, the actual file should or should not exist during that point in the processing. So the only thing I do is run recovery and get `list -v`. So that makes sure the metadata can be read.

Another path I started on was a `test/csuite` program that would try to make sure we created and then ran recovery on every type of log record and log operation type that exists. That is more complicated, so writing this test was much faster turnaround.

Anyway, this test is certainly not harmful. We still might want to have a different-focused test to verify that every log record and operation is tested.

### DIFF
--- a/test/suite/test_schema08.py
+++ b/test/suite/test_schema08.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python
+#
+# Public Domain 2014-2018 MongoDB, Inc.
+# Public Domain 2008-2014 WiredTiger, Inc.
+#
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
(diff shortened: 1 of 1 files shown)


agorrod:
@sueloverso I like how this fell out - a Python test is preferable to a C test, I think.

Should I go ahead and merge, or do you think we should pursue something more involved as well?

PR #4193, opened 2018-08-01
WT-4134 Rework assertion that we don't discard required history.

michaelcahill:


michaelcahill:
@agorrod, I'm not sure how much clearer this is, but I think (a) the existing comment doesn't match the assertion, and (b) there are some cases the existing assertion doesn't cover, which I think could explain why it fired in testing.  In particular, after a forced drop when we mark a tree dead, the usual rules about maintaining history are broken.

agorrod:
Thanks @michaelcahill this change looks better to me. I'm not sure if it's possible for there to be a NULL pointer dereference in there somewhere, but I'm going to go ahead and merge. We'll find out soon enough.

PR #4194, opened 2018-08-01
WT-4213 Remove redundant and incorrect stat description text.

sueloverso:
@bdlucas1 and @agorrod here's the proposed text changes for the incorrect locking statistic descriptions. The change is simply to remove the redundant phrase from the end `for the dhandle lock` from all applicable stats. I did not otherwise change the text.


bdlucas1:
Thanks, I think this looks good.

agorrod:
Thanks @sueloverso

PR #4199, opened 2018-08-06
WT-4207 Guard against null log pointer dereference

ddanderson:


ddanderson:

PR #4200, opened 2018-08-06
WT-4210 Start timer after checkpoint completes.

sueloverso:
@agorrod This fixes the failures on tinderbox that result from very slow IO on that machine and a false positive timeout. Please review.


agorrod:
Thanks @sueloverso lgtm.

PR #4201, opened 2018-08-07
WT-4235 Fix workgen tracking of table runtime across workloads.

michaelcahill:
Previously, there was code approximating realloc in C++, but it assumed the wrong size of TableRuntime objects in an array.

### DIFF
--- a/bench/workgen/workgen.cxx
+++ b/bench/workgen/workgen.cxx
@@ -254,12 +254,11 @@ ContextInternal::~ContextInternal() {
 }
 
 int ContextInternal::create_all() {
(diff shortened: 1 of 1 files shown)


michaelcahill:
@ddanderson, this took me a while to track down, and maybe I'm using `Workload` in unexpected ways (see https://jira.mongodb.org/secure/attachment/193452/wt4194.py for details).  Regardless, avoiding `memcpy` and `memset` here seems clearer to me.  Can you please take a look and let me know if I've misunderstood the intent?

ddanderson:
@michaelcahill, I agree.  That's clearer and safer.

PR #4205, opened 2018-08-08
WT-4215 Allow recovery of log files from a backup without salvage.

ddanderson:
Because they are being read concurrently with being written, log files copied as part of a backup may have various portions that are zeroed. In particular, some log records may have their first part zeroed, and subsequent parts filled in. This should never happen for log records used after an application crash, so we must apply different rules for these cases. For backup log files, we allow checksums to fail, or holes in log files to end without the start of a valid record. Either one of these 'corruptions' is unreported, but causes the log to truncate at that point.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1916,14 +1916,21 @@ __log_has_hole(WT_SESSION_IMPL *session, WT_FH *fh, wt_off_t log_size,
 			 * present in the buffer, we either have a valid header
 			 * or corruption.  Verify the header of this record to
 			 * determine whether it is just a hole or corruption.
+			 *
+			 * We don't bother making this check for backup copies,
+			 * as records may have their beginning zeroed, hence
+			 * the part after a hole may in fact be the middle of
+			 * the record.
 			 */
-			logrec = (WT_LOG_RECORD *)p;
-			if (buf_left >= sizeof(WT_LOG_RECORD)) {
-				off += p - buf;
-				WT_ERR(__log_record_verify(session, fh,
-				    (uint32_t)off, logrec, &corrupt));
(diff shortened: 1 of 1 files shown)


ddanderson:

src/log/log.c
```diff
@@ -2481,29 +2482,49 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
...
+					 */
+					WT_ERR(__log_record_verify(session,
+					    log_fh, rd_lsn.l.offset, logrec,
+					    &corrupt));
+					if (corrupt) {
+						need_salvage = true;
+						WT_ERR(__log_salvageable_error(
+						    session, log_fh->name, "",
+						    rd_lsn.l.offset));
+					}
```
sueloverso: My minor complaint is that it is a bit of a shame to have another layer of conditional in that code. It is pretty deep already. Maybe something like this:
```
if (!__log_checksum_mismatch(...)) {
    if (log != NULL)
        log->trunc_lsn = rd_lsn;
    if (LF_ISSET(WT_LOGSCAN_ONE))
        ret = WT_NOTFOUND;
    if (F_ISSET(conn, WT_CONN_WAS_BACKUP))
        break;
    if (!log_check_partial_write(...)) {
    } else { ...
    }
    break;
}
```
That helps one layer (if it works, I haven't thought too deeply about the line reordering correctness).

ddanderson: @sueloverso, I like it. I'll make that change and test it out.  BTW, it has previously been failing at the first or second iteration of test/format on a specific config.  It ran cleanly for > 10 on the one you reviewed.

PR #4206, opened 2018-08-08
WT-4193 test/format snapshot-isolation search mismatch

keithbostic:


michaelcahill:
Thanks @keithbostic, I'm happy with this change, and I think it's strong enough as-is.  That is, for read-committed, as long as the cursor stays positioned, which it will if the application does the obvious thing, then the results will be correct.  If the cursor is not positioned, `modify` will do its own search and we'll keep a snapshot active afterwards.

michaelcahill:
@agorrod, this is an API change -- (a) are you comfortable with it, and (b) any special handling / sequencing required?  At minimum, I guess it needs upgrading notes...

keithbostic:
@michaelcahill, @agorrod: I got almost 3000 runs of the failing CONFIG last night, but there were 5 failures.

They look similar to this failure, that is, a snapshot-isolation mismatch, but I don't see any way we could still be mixing-and-matching implicit transactions and modify operations, so I think it's something different.

I'm going to start looking for the failure, but I'm inclined to merge this ticket and open another ticket when we find the next problem. I'm only going to get a test run or so a day, absent finding some way to make it reproduce more reliably, so this next one could take awhile.

I'll put something in the upgrading notes, and I'll add some stronger wording to the `WT_CURSOR.modify` docs.

agorrod:
Thanks @keithbostic lgtm

PR #4207, opened 2018-08-08
WT-4176 Expose a WT_SESSION.query_timestamp method.

dgottlieb:
Accepts configuration `get=read_timestamp` which returns a timestamp representing which updates are visible to the running transaction.


dist/api_data.py
```diff
@@ -1162,6 +1162,14 @@ def __cmp__(self, other):
         type='list'),
 ]),
 
+'WT_SESSION.query_timestamp' : Method([
+    Config('get', 'read_timestamp', r'''
+        specify which timestamp to query: \c read_timestamp returns the
+        timestamp the transaction is reading at.  See @ref
+        transaction_timestamps''',
+        choices=['read_timestamp']),
```
sueloverso: I have two comments on the API:
1. I have a preference for it to just be `read` as the `timestamp` part is implied in the method name, and it is consistent with the usage of the connection's `query_timestamp` of `oldest, pinned` etc.
2. Now with all the infrastructure in place, I'd like to see support for all the timestamps available in the txn. May as well add them now and be complete. `commit|first_commit|prepare|read`

src/txn/txn_timestamp.c
```diff
@@ -302,6 +302,32 @@ __txn_global_query_timestamp(
 done:	__wt_timestamp_set(tsp, &ts);
 	return (0);
 }
+
+/*
+ * __txn_query_timestamp --
+ *	Query a timestamp.
```
sueloverso: Here (and the next function) I'd like the comment to clearly state "Query a timestamp within this session's txn." so that it isn't confused with the connection/global function.

src/txn/txn_timestamp.c
```diff
@@ -302,6 +302,32 @@ __txn_global_query_timestamp(
...
+	WT_CONFIG_ITEM cval;
+	WT_TXN *txn;
+	wt_timestamp_t ts;
+
+	txn = &session->txn;
+
+	WT_STAT_CONN_INCR(session, session_query_ts);
+	WT_RET(__wt_config_gets(session, cfg, "get", &cval));
+	if (WT_STRING_MATCH("read_timestamp", cval.str, cval.len)) {
+		if (!F_ISSET(txn, WT_TXN_RUNNING))
```
sueloverso: I think this conditional should be moved after the STAT increment. Particularly if/when the other timestamps are added here, it is a general conditional that would apply to all of them.

dist/api_data.py
```diff
@@ -1162,6 +1162,16 @@ def __cmp__(self, other):
...
 ]),
 
+'WT_SESSION.query_timestamp' : Method([
+    Config('get', 'read', r'''
+        specify which timestamp to query: \c read returns the timestamp the
+        transaction is reading at.  \c commit returns the most recently set
+        commit_timestamp. \c first_commit returns the first set
+        commit_timestamp. \c prepare returns the timestamp used in preparing a
+        transaction. See @ref transaction_timestamps''',
+        choices=['read', 'commit', 'first_commit', 'prepare']),
```
sueloverso: Please alphabetize the choices and their order in the description above to also be alphabetized.

dist/api_data.py
```diff
@@ -1162,6 +1162,16 @@ def __cmp__(self, other):
         type='list'),
 ]),
 
+'WT_SESSION.query_timestamp' : Method([
+    Config('get', 'read', r'''
+        specify which timestamp to query: \c read returns the timestamp the
+        transaction is reading at.  \c commit returns the most recently set
```
sueloverso: Please rephrase this to be `\c read returns the timestamp at which the transaction is reading.`

src/txn/txn_timestamp.c
```diff
@@ -302,11 +302,44 @@ __txn_global_query_timestamp(
...
+	WT_TXN *txn;
+
+	txn = &session->txn;
+
+	WT_STAT_CONN_INCR(session, session_query_ts);
+	if (!F_ISSET(txn, WT_TXN_RUNNING))
+		return (WT_NOTFOUND);
+
+	WT_RET(__wt_config_gets(session, cfg, "get", &cval));
+	if (WT_STRING_MATCH("read", cval.str, cval.len))
```
sueloverso: Please alphabetize the `WT_STRING_MATCH` usage.

src/txn/txn_timestamp.c
```diff
@@ -705,6 +738,28 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
 	return (0);
 }
 
+/*
+ * __wt_txn_query_timestamp --
+ *	Query a timestamp.
```
sueloverso: Add same addition as internal function: `Query a timestamp within this session's txn.`

test/suite/test_timestamp13.py
```diff
@@ -0,0 +1,146 @@
...
+    conn_config = 'log=(enabled)'
+
+    def test_degenerate_timestamps(self):
+        if not wiredtiger.timestamp_build():
+            self.skipTest('requires a timestamp build')
+
+        self.session.create(self.uri,
+            'key_format=i,value_format=i' + self.extra_config)
+
+        times_to_query = ['read', 'commit', 'first_commit', 'prepare']
```
sueloverso: Two comments:
1. Alphabetize the list.
2. I suggest calling it something like `txn_queries` or `query_choices`. `times` has unrelated connotations.

test/suite/test_timestamp13.py
```diff
@@ -0,0 +1,146 @@
...
+        # Querying a session's timestamps will error when not in a transaction.
+        for time_to_query in times_to_query:
```
sueloverso: For all these loops, I suggest something like `ts_to_query` or just `query`. Again `time` has other implications.

test/suite/test_timestamp13.py
```diff
@@ -0,0 +1,146 @@
...
+            self.assertTimestampsEqual(
+                self.session.query_timestamp('get=' + time_to_query), '0')
+
+        self.session.commit_transaction()
+        # Querying a session's timestamps will error when not in a transaction.
+        for time_to_query in times_to_query:
+            self.assertRaises(
+                wiredtiger.WiredTigerError,
+                lambda: self.session.query_timestamp('get=' + time_to_query))
```
sueloverso: I love all these tests! I suggest one more in this function to verify the error on a call with `get=unknown` to test that else clause too.

src/txn/txn_timestamp.c
```diff
@@ -302,11 +302,44 @@ __txn_global_query_timestamp(
 done:	__wt_timestamp_set(tsp, &ts);
 	return (0);
 }
+
+/*
+ * __txn_query_timestamp --
+ *	Query a timestamp within this session's txn.
+ */
```
agorrod: We try to avoid abbreviations in comments could you "txn" => "transaction"

src/txn/txn_timestamp.c
```diff
@@ -705,6 +738,28 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
 	return (0);
 }
 
+/*
+ * __wt_txn_query_timestamp --
+ *	Query a timestamp.
+ */
+int
+__wt_txn_query_timestamp(
```
agorrod: It's a bit sad that so much of this is duplicated with the existing `__wt_txn_global_query_timestamp` - did you think about instead renaming that to `__wt_txn_query_timestamp` and adding a bool parameter to differentiate between the session and global versions?

PR #4208, opened 2018-08-09
WT-4241 GNU-stack section should never be conditionally compiled out

keithbostic:


keithbostic:
autobuild: retest this please

agorrod:
lgtm - please merge if platforms team confirms this fixes their issue.

src/checksum/power8/crc32.sx
```diff
@@ -775,11 +775,13 @@ FUNC_START(__crc32_vpmsum)
...
+#endif
 
 /*
  * Make sure the stack isn't executable with GCC (regardless of platform).
+ * Note: this section must always be output and can never be #ifdef'd out as
+ * part of conditional compilation.
  */
 #ifdef __ELF__
 .section	.note.GNU-stack,"",@progbits
 #endif
```
acmorrow: Do you want to add a scary
/* DO NOT SAY #ifndef BELOW THIS LINE, per above comment */

The comment you have is good, but will anyone read it before slapping an #endif on the last line of the file?

src/checksum/zseries/crc32le-vx.sx
```diff
@@ -279,11 +279,13 @@ crc32_le_vgfm_generic:
...
 	lmg     %r14,%r15,240(%r15)
 	br	%r14
+#endif
 
 /*
  * Make sure the stack isn't executable with GCC (regardless of platform).
+ * Note: this section must always be output and can never be #ifdef'd out as
+ * part of conditional compilation.
  */
 #ifndef __clang__
```
acmorrow: Not to be addressed in this review, but I wonder a bit. This file makes the GNU-stack note conditional on not using clang, the other makes it conditional on using ELF. I wonder which is right. Something for another day.

acmorrow:
Note that this should be backported to all branches where WT-4198 was fixed (https://github.com/wiredtiger/wiredtiger/commit/96f004261d6336f81fbd0bc245e8aa07e7cc24ac)

src/checksum/power8/crc32.sx
keithbostic: Will do.

src/checksum/zseries/crc32le-vx.sx
keithbostic: Good catch, thanks -- I'll use `__ELF__`, afaict, that's accepted practice.

PR #4210, opened 2018-08-12
WT-4104 keys reappear after fast truncate.

bvpvamsikrishna:
As per my understanding, checkpoint is marking the btree as clean, which is resulting in not constructing the update chain when fast truncate page is read back from disk.

### DIFF
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -357,10 +357,9 @@ __wt_delete_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	/*
 	 * Give the page a modify structure.
 	 *
-	 * If the tree is already dirty and so will be written, mark the page
-	 * dirty.  (We'd like to free the deleted pages, but if the handle is
-	 * read-only or if the application never modifies the tree, we're not
-	 * able to do so.)
+	 * Mark tree dirty, unless the handle is read-only.
+	 * (We'd like to free the deleted pages, but if the handle is read-only,
+	 * we're not able to do so.)
 	 */
 	WT_RET(__wt_page_modify_init(session, page));
 	if (!F_ISSET(btree, WT_BTREE_READONLY))
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -337,6 +337,7 @@ __tombstone_update_alloc(WT_SESSION_IMPL *session,
 int
 __wt_delete_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 {
+	WT_BTREE *btree;
 	WT_DECL_RET;
 	WT_INSERT *ins;
 	WT_INSERT_HEAD *insert;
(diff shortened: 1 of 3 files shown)


bvpvamsikrishna:
autobuild: retest this please

src/btree/bt_delete.c
```diff
@@ -363,7 +363,7 @@ __wt_delete_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * able to do so.)
 	 */
 	WT_RET(__wt_page_modify_init(session, page));
-	if (btree->modified)
+	if (session->dhandle->checkpoint == NULL)
```
michaelcahill: @bvpvamsikrishna, in https://github.com/wiredtiger/wiredtiger/blob/develop/src/btree/bt_handle.c#L89-L91 there is this code:

```
 89         /* Checkpoint files are readonly. */
 90         readonly = dhandle->checkpoint != NULL ||
 91             F_ISSET(S2C(session), WT_CONN_READONLY);
```

That is, there is another condition that leads to read-only trees.  That is more unusual, but MongoDB does exercise it (e.g., for queryable backups).  I think this situation of a reference to a deleted child page could occur there as well.

Given that, I think we should add a `WT_BTREE_READONLY` flag to btree handles and set it in `bt_handle.c` based on this condition, then turn the change in `bt_delete.c` into `if (!F_ISSET(btree, WT_BTREE_READONLY)) ...`

bvpvamsikrishna:
I made some changes to test/format to reproduce issue reported in WT-4104, but I don't think there is any value in checking them in to develop.

bvpvamsikrishna:
@agorrod Request your review for this PR.

src/btree/bt_delete.c
```diff
@@ -363,7 +363,7 @@ __wt_delete_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * able to do so.)
```
agorrod: I think this comment should be updated - the code will now always mark the tree dirty unless it's read only.

src/btree/bt_handle.c
```diff
@@ -120,7 +121,8 @@ __wt_btree_open(WT_SESSION_IMPL *session, const char *op_cfg[])
 		WT_ERR_MSG(session, EINVAL, "expected a 'file:' URI");
 
 	WT_ERR(__wt_block_manager_open(session, filename, dhandle->cfg,
-	    forced_salvage, readonly, btree->allocsize, &btree->bm));
+	    forced_salvage, F_ISSET(btree, WT_BTREE_READONLY),
+	    btree->allocsize, &btree->bm));
```
agorrod: Do the `__wt_block_manager_open` and `bt->checkpoint_load` calls still need the readonly parameter? Could they instead check the btree structure?

bvpvamsikrishna: Currently btree structure is not referenced in block manager. so I think better to have readonly parameter.

michaelcahill:
Thanks @bvpvamsikrishna, this change lgtm and I agree with you that there is a layering between the block manager code and the btree handle that we should maintain.  I'll merge...

PR #4212, opened 2018-08-13
WT-4144 Fix rollback_to_stable with lookaside history.

michaelcahill:
There were several bugs, resulting in incorrect data after a rollback_to_stable call in the presence of lookaside history.

The main issue was with pages images containing updates newer than the stable timestamp (which is a common case because WiredTiger evicts optimistically, assuming the stable timestamp will move forward by the time the next checkpoint starts).  Since rollback_to_stable doesn't modify page images, it instead was marking such pages "invalid" so the history would always be loaded before the pages could be read correctly.  However, the required history to make sense of such pages could be removed by a lookaside sweep, which makes decisions solely on the contents of the lookaside table without access to the "invalid" flag.

The solution here is to read the history into cache for such pages during rollback_to_stable.  The page is marked dirty, with updates newer than the stable timestamp rolled back, then it can be evicted or rewritten in memory.  This eliminates the special "invalid" flag.

In the process of finding that bug, a number of other issues were found and fixed, including:
* don't read all lookaside content into cache during rollback_to_stable;
* avoid races between eviction threads and rollback_to_stable (fix nesting of locks and transactions, use an explicit flag check during eviction rather than relying on the eviction server thread to update cache-wide flags); and
* simplify functions accessing the lookaside table so each function has a single session.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -394,9 +394,6 @@ __wt_las_page_skip_locked(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * We also need to instantiate a lookaside page if this is an update
 	 * operation in progress.
 	 */
-	if (ref->page_las->invalid)
-		return (false);
-
 	if (F_ISSET(txn, WT_TXN_UPDATE))
 		return (false);
 
@@ -470,18 +467,23 @@ __wt_las_page_skip(WT_SESSION_IMPL *session, WT_REF *ref)
  *	Remove all records for a given page from the lookaside store.
  */
 static int
-__las_remove_block(WT_SESSION_IMPL *session,
+__las_remove_block(
     WT_CURSOR *cursor, uint64_t pageid, bool lock_wait, uint64_t *remove_cntp)
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
 	WT_ITEM las_key;
+	WT_SESSION_IMPL *session;
+	WT_TXN_ISOLATION saved_isolation;
 	uint64_t las_counter, las_pageid;
 	uint32_t las_id;
+	bool local_txn;
 
 	*remove_cntp = 0;
 
+	session = (WT_SESSION_IMPL *)cursor->session;
 	conn = S2C(session);
+	local_txn = false;
 
 	/* Prevent the sweep thread from removing the block. */
 	if (lock_wait)
@@ -489,6 +491,10 @@ __las_remove_block(WT_SESSION_IMPL *session,
 	else
 		WT_RET(__wt_try_writelock(
 		    session, &conn->cache->las_sweepwalk_lock));
+
+	__las_set_isolation(session, &saved_isolation);
+	WT_ERR(__wt_txn_begin(session, NULL));
+	local_txn = true;
 
 	/*
 	 * Search for the block's unique btree ID and page ID prefix and step
@@ -508,7 +514,15 @@ __las_remove_block(WT_SESSION_IMPL *session,
 	}
 	WT_ERR_NOTFOUND_OK(ret);
 
-err:	__wt_writeunlock(session, &conn->cache->las_sweepwalk_lock);
+err:	if (local_txn) {
+		if (ret == 0)
+			ret = __wt_txn_commit(session, NULL);
+		else
+			WT_TRET(__wt_txn_rollback(session, NULL));
+	}
+
+	__las_restore_isolation(session, saved_isolation);
+	__wt_writeunlock(session, &conn->cache->las_sweepwalk_lock);
 	return (ret);
 }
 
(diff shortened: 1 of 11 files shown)


src/include/btree.i
```diff
@@ -1189,7 +1189,7 @@ __wt_page_las_active(WT_SESSION_IMPL *session, WT_REF *ref)
 
 	if ((page_las = ref->page_las) == NULL)
 		return (false);
-	if (page_las->invalid || !ref->page_las->skew_newest)
+	if (!ref->page_las->skew_newest)
```
keithbostic: We can strip the `ref->` from this -- I'm sure it makes no difference, but it read oddly.

src/txn/txn_rollback_to_stable.c
```diff
@@ -230,9 +229,47 @@ static int
...
+		WT_RET(__wt_page_in(session, ref, read_flags));
+		WT_ASSERT(session, ref->state != WT_REF_LIMBO &&
+		    ref->page != NULL && __wt_page_is_modified(ref->page));
+		local_read = true;
+	}
+
+	/* Review deleted page saved to the ref */
+	if (ref->page_del != NULL && __wt_timestamp_cmp(
+	    rollback_timestamp, &ref->page_del->timestamp) < 0)
+		WT_ERR(__wt_delete_page_rollback(session, ref));
```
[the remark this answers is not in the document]
michaelcahill: The timestamp comparison is going the other way: the truncate is more recent than the rollback timestamp, so it has to be rolled back.

src/evict/evict_lru.c
```diff
@@ -628,11 +628,10 @@ __evict_update_work(WT_SESSION_IMPL *session)
 	 * (3) the cache is more than half way from the dirty target to the
 	 *     dirty trigger.
 	 */
-	if (!F_ISSET(conn, WT_CONN_EVICTION_NO_LOOKASIDE) &&
-	    (__wt_cache_stuck(session) ||
```
agorrod: I don't understand how this change is related - and it makes me a bit apprehensive that we'll queue pages we wouldn't have queued before (maybe that won't happen?)

src/txn/txn_rollback_to_stable.c
```diff
@@ -230,9 +229,47 @@ static int
...
     WT_SESSION_IMPL *session, WT_REF *ref, wt_timestamp_t *rollback_timestamp)
 {
+	WT_DECL_RET;
 	WT_PAGE *page;
+	uint32_t read_flags;
+	bool local_read;
+
+	local_read = false;
+	read_flags = WT_READ_WONT_NEED;
+	if (ref->page_las != NULL && ref->page_las->skew_newest &&
```
agorrod: A comment describing why this is necessary would be good. i.e: sweep could cleanup the required history after rollback has completed if the previous reconciliation wrote the most recent version of data.

src/evict/evict_lru.c
michaelcahill: This was fixing a bug: `rollback_to_stable` sets `WT_CONN_EVICTION_NO_LOOKASIDE` as its way to make sure no lookaside eviction happens, but only had an effect after the eviction ran and clearing `WT_CACHE_EVICT_LOOKASIDE` from the cache flags.

What I did was move the `WT_CONN_EVICTION_NO_LOOKASIDE` check into `__evict_review` so it will be checked immediately (or at least reliably after `__txn_rollback_eviction_drain` finishes).  I hope that makes sense: let me know if I can add comments to clarify.

src/txn/txn_rollback_to_stable.c
michaelcahill: Done.

src/cache/cache_las.c
```diff
@@ -742,21 +755,21 @@ __wt_las_insert_block(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
 err:	/* Resolve the transaction. */
 	if (local_txn) {
 		if (ret == 0)
-			WT_TRET(__wt_txn_commit(las_session, NULL));
+			WT_TRET(__wt_txn_commit(session, NULL));
```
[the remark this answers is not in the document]
michaelcahill: Done.

agorrod:
Thanks @michaelcahill lgtm

src/txn/txn_rollback_to_stable.c
```diff
@@ -252,10 +289,12 @@ __txn_abort_newer_updates(
 	case WT_PAGE_ROW_LEAF:
 		__txn_abort_newer_row_leaf(session, page, rollback_timestamp);
 		break;
-	WT_ILLEGAL_VALUE(session, page->type);
+	WT_ILLEGAL_VALUE_ERR(session, page->type);
 	}
 
-	return (0);
+err:	if (local_read)
+		WT_TRET(__wt_page_release(session, ref, read_flags));
```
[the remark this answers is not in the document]
michaelcahill: That's true today, but I'd prefer if we have a set of flags that are used consistently for all paired calls to `__wt_page_in` and `__wt_page_release` -- it's easy to imagine wanting to know as part of releasing a page whether it should stay in cache or not.

PR #4214, opened 2018-08-13
WT-3894 Timestamp queue statistics and usage

sueloverso:
@michaelcahill and @agorrod this is ready for merge. There are a number of interesting things to note:
1. I pulled in Michael's change to walk the queues backward. Some performance was better and none seemed worse and it still only takes the lock once.
2. I created changes to `timestamp_abort` to use the timestamps in a way similar to MongoDB when not using prepared transactions. This is using Michael's change but supporting prepared as well.
3. I added in two `walked` statistics to count how many entries in the timestamp queues we walk per insert.
4. I changed the use of the `read_timestamp_queue` to work the same way as the timestamp queue. This change was committed separately so we can undo that one commit if you want. This means we take the write lock only once rather than both on set/clear. I modeled it after the new commit timestamp queue so that it walks backward also.


src/session/session_api.c
```diff
@@ -184,7 +184,8 @@ __wt_session_release_resources(WT_SESSION_IMPL *session)
 /*
  * __session_clear_commit_queue --
  *	We're about to clear the session and overwrite the txn structure.
- *	Remove ourselves from the commit timestamp queue if we're on it.
+ *	Remove ourselves from the commit timestamp queue and the read
+ *	timestamp queue if we're on either of them.
  */
 static void
 __session_clear_commit_queue(WT_SESSION_IMPL *session)
```
michaelcahill: I moved this code to `txn_timestamp.c` and renamed the function `__wt_txn_clear_timestamp_queues` to match the new functionality.

sueloverso:
Thanks @michaelcahill I agree with that refactoring.

PR #4215, opened 2018-08-14
WT-4212 Modify lookaside schema to carry prepared update state

sulabhM:


src/btree/bt_delete.c
```diff
@@ -314,8 +314,8 @@ __tombstone_update_alloc(WT_SESSION_IMPL *session,
 {
 	WT_UPDATE *upd;
 
-	WT_RET(
-	    __wt_update_alloc(session, NULL, &upd, sizep, WT_UPDATE_TOMBSTONE));
+	WT_RET(__wt_update_alloc(session,
+	    NULL, &upd, sizep, WT_UPDATE_TOMBSTONE, WT_PREPARE_INIT));
```
sulabhM: Instead of setting prepare_state to WT_PREPARE_INIT, should I be checking for ```ref->page_del->prepare_state``` in the caller ```__wt_delete_page_instantiate()``` and use that instead ? This concern is coming from line 369, considering that it is possible to have a deleted prepared page ?
I suspect not because I do not see ```upd->prepare_state``` getting set around this code.

src/cache/cache_las.c
```diff
@@ -1113,8 +1113,8 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 		 * Remove entries from the lookaside that have aged out and are
```
sulabhM: How do we clear the prepared entries once they are committed/rollbacked?
Does aging out takes care of these entries or we need to put extra logic to take care of sweeping prepared updates once not needed.

src/txn/txn_rollback_to_stable.c
```diff
@@ -63,8 +63,8 @@ __txn_rollback_to_stable_lookaside_fixup(WT_SESSION_IMPL *session)
 		if (__bit_test(conn->stable_rollback_bitstring, las_id))
 			continue;
```
sulabhM: Do we have to do anything special here to remove prepared updates from the las file ?

src/btree/bt_delete.c
bvpvamsikrishna: The value of WT_PREPARE_INIT is choosed as '0' to avoid it setting specifically.  __wt_update_alloc will initialize to zero. My preference is not to modify __wt_update_alloc signature. A page with prepared update cannot be freed and hence page will not be instantiated.

src/cache/cache_las.c
bvpvamsikrishna: For a transaction to commit / rollback, it will mark the update structures as such, hence the pages modified by the transaction will be in memory. When a transaction commit / rollback is in progress all the pages modified by it will be loaded to memory, so the corresponding las entries will also be added back to page and will be removed from las. Hence no need for extra logic to take care of sweeping prepared updates from las.

src/txn/txn_rollback_to_stable.c
bvpvamsikrishna: we don't do anything special for prepare updates, when the entries of a page are removed from las, it includes committed updates as well as prepared updates.

src/btree/bt_delete.c
sulabhM: When reading an update from lookaside we need to pass the corresponding prepare_state. Hence we need to change the signature of __wt_update_alloc(). eg:
https://github.com/wiredtiger/wiredtiger/pull/4215/files#diff-e80f0967af0f71afc192ac61c363333cR187

What I have a minor doubt is when we are re-initiating a deleted page with __wt_delete_page_instantiate(), from the code at bt_delete.c:369 it looks like the deleted page can be in prepared state. Question I have over there, that in such a case, would the updates for the records on that page be in the same prepare_state as the page itself or be default initialised to WT_PREPARE_INIT. I think they will be default initialised to WT_PREPARE_INIT, because that is what happens today and I don't see how long running prepared will change that.

src/cache/cache_las.c
sulabhM: I was concerned as to at what stage do we remove lookaside entries upon reading them back. I confirmed with @agorrod that here in this case reading prepared updates back for a commit/rollback will cause page to be eventually reconciled in the memory and deleted from lookaside. So, we won't need to sweep prepared updates out of lookaside.

src/txn/txn_rollback_to_stable.c
sulabhM: I think you are right.

src/btree/bt_delete.c
bvpvamsikrishna: I mean instead of modify __wt_update_alloc signature, where ever required we can update the prepared state of update structure after it is allocated.

bvpvamsikrishna: If a page is deleted, it will have page_del structure, and prepare state will be set to page_del structure. It will not have updates.

sulabhM: > I mean instead of modify __wt_update_alloc signature, where ever required we can update the prepared state of update structure after it is allocated

I will like to keep it as I have in the change at the moment, ie to pass the prepare_state through the __wt_update_alloc() call. I think this is more consistent with the notion that we pass all the needed information to this function and expect update structure to be allocated and initialised to the given information, eg: modify_type is passed similarly as an argument to this function. Also this prevents from forgetting to set prepare_type when using __wt_update_alloc() and forces you to put a thought about what the prepare_state should be, even when default.

bvpvamsikrishna: prepare is the state of the transaction, which is inherited by update. May be later we can get away with prepare_state inside update structure and might only maintain at txn level. In my opinion, prepare state we can consider in conjunction with txnid. We will discuss.

src/include/btmem.h
```diff
@@ -223,7 +223,7 @@ struct __wt_ovfl_reuse {
 #endif
 #define	WT_LAS_CONFIG							\
     "key_format=" WT_UNCHECKED_STRING(QIQu)				\
-    ",value_format=" WT_UNCHECKED_STRING(QuBu)				\
+    ",value_format=" WT_UNCHECKED_STRING(QuBuB)				\
```
bvpvamsikrishna: Please update the comment with the change.

src/cache/cache_las.c
```diff
@@ -1113,8 +1113,8 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
 		 * now no longer needed.
 		 */
-		WT_ERR(cursor->get_value(cursor,
-		    &las_txnid, &las_timestamp, &upd_type, &las_value));
+		WT_ERR(cursor->get_value(cursor, &las_txnid,
+		    &las_timestamp, &upd_type, &las_value, &prepare_state));
```
bvpvamsikrishna: we can skip the prepared updates as part of sweep.  Below call to __wt_txn_visible_all will return "True" for prepared updates.  add this condition also to below if.

src/include/btmem.h
sulabhM: To discuss: Is there any value in having the schema as "QuBBu" instead of the one I have in the change now.

michaelcahill: The encoding is a little more efficient if the last field is a 'u' -- we can save storing the length in that case and just treat the rest of the bytes as the data for a `WT_ITEM`.

src/cache/cache_las.c
```diff
@@ -1124,11 +1124,13 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
-		 * If it is visible then perform additional checks to see
-		 * whether it has aged out of a live file.
+		 * If this entry isn't globally visible or if it is a part of a
+		 * prepared transaction, then we cannot remove it. Otherwise,
+		 * perform additional checks to see whether it has aged out of a
+		 * live file.
 		 */
-		if (!__wt_txn_visible_all(session, las_txnid, val_ts)) {
+		if (!__wt_txn_visible_all(session, las_txnid, val_ts) ||
+		    prepare_state == WT_PREPARE_INPROGRESS) {
```
agorrod: Why don't you need to check other prepared states here?

sulabhM: @bvpvamsikrishna pointed out that updates that are in the prepared state, ie state is set to ```WT_PREPARE_INPROGRESS``` will be globally visible, but we still want the sweep to skip it. Hence I have added this check here. The state here can't be ```WT_PREPARE_LOCKED``` because that is a transient state sweep server wont see in lookaside records. The state can't be ```WT_PREPARE_RESOLVED``` either because a commit/rollback will cause the state to transition to ```WT_PREPARE_RESOLVED``` and once that happens if this record goes into the lookaside, it will be treated like a normal commited update.

PR #4216, opened 2018-08-14
WT-4249 Attempt to discard dirty page during verify operation.

bvpvamsikrishna:
This is fall out from WT-4104 fix, where handles opened by verify are
missed to mark as read-only.


michaelcahill:
Thanks @bvpvamsikrishna, lgtm.  I'll merge once testing finishes.

PR #4217, opened 2018-08-14
WT-4156 Add salvage API option to wiredtiger_open.

sueloverso:
Not ready for merge, but first modifications to get the API and new error code in and make it work for the existing log recovery salvage code paths.

@ddanderson and @agorrod if you could eyeball this before I go too far down the path. Next up is the metadata salvage which is new code.


src/conn/conn_log.c
```diff
@@ -308,13 +308,6 @@ __logmgr_config(
 		    session, cfg, "log.recover", 0, &cval));
 		if (WT_STRING_MATCH("error", cval.str, cval.len))
 			FLD_SET(conn->log_flags, WT_CONN_LOG_RECOVER_ERR);
-		else if (WT_STRING_MATCH("salvage", cval.str, cval.len)) {
```
ddanderson: Have we lost the check that salvage is incompatible with starting readonly? Or is that enforced some other way?

test/suite/test_txn19.py
```diff
@@ -303,7 +303,7 @@ def test_corrupt_log(self):
 
         if expect_fail:
             self.check_file_contains_one_of(errfile,
-            ['/log file.*corrupted/', 'WT_ERROR: non-specific WiredTiger error'])
```
ddanderson: It's quite possible now we can change this to `self.check_files_contains('WT_DATA_CORRUPTION: database corruption detected')` since we now have a single error code that we should always get for these errors.

sueloverso:
@agorrod and @ddanderson this is ready for review. I have a `csuite` test that test corruption. I create a number of different tables so that the metadata corruption I add will be on a different page than some table. If you have better ideas for the details of that testing, please let me know.

Michael asked for a test that tested a turtle file and metadata file out of sync with each other. This branch does not include that. This gets the API into the library and a new ticket can be added for that new test.

lukech:
@sueloverso it looks the `salvage` choice is removed from the `recover` API, and a new `salvage` API is introduced instead. Do you think it warrants some writing in the `upgrading.dox`?

ext/test/fail_fs/fail_fs.c
```diff
@@ -269,7 +269,7 @@ fail_file_read(WT_FILE_HANDLE *file_handle,
 			    fail_fh->iface.name,
 			    (uintmax_t)len, (uintmax_t)offset,
 			    wtext->strerror(wtext, NULL, errno));
-			ret = (nr == 0 ? WT_ERROR : errno);
+			ret = (nr == 0 ? WT_DATA_CORRUPTION : errno);
```
agorrod: I was hoping that we could restrict `WT_DATA_CORRUPTION` to only be returned by `wiredtiger_open` for now - could we achieve that by converting `WT_DATA_CORRUPTION` into WT_ERROR in the `API_END` macro?

That way applications won't need to change all of their WiredTiger API callers to be able to handle a new error return code.

dist/api_data.py
```diff
@@ -886,6 +886,11 @@ def __cmp__(self, other):
         methods that may modify a database are disabled.  See @ref readonly
         for more information''',
         type='boolean'),
+    Config('salvage', 'false', r'''
+        open connection and salvage any WiredTiger database files that it
+        detects as corrupted. Should only be used after getting an error
+        return of WT_DATA_CORRUPTION. This may be a destructive setting.''',
```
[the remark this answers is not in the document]
sueloverso: Done.

src/conn/conn_api.c
```diff
@@ -2625,6 +2601,24 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
...
 
+	WT_ERR(__wt_config_gets(session, cfg, "cache_cursors", &cval));
+	if (cval.val)
+		F_SET(conn, WT_CONN_CACHE_CURSORS);
+
+	WT_ERR(__wt_config_gets(session, cfg, "checkpoint_sync", &cval));
+	if (cval.val)
+		F_SET(conn, WT_CONN_CKPT_SYNC);
+
+	WT_ERR(__wt_config_gets(session, cfg, "direct_io", &cval));
```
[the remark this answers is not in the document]
sueloverso: You're correct. I moved them back, but below the comment and add more to the comment so that they stay together.

ext/test/fail_fs/fail_fs.c
sueloverso: It cannot be added to `API_END` because not all callers send in a `ret` variable that can be assigned. There are `API_END_RET` and friends that take a `ret` variable (although it doesn't appear it must - we just use it that way now).

test/csuite/wt4156_metadata_salvage/main.c
```diff
@@ -0,0 +1,462 @@
...
+static void
+cursor_insert(WT_SESSION *session, const char *uri)
+{
+	WT_CURSOR *cursor;
+	WT_ITEM vu;
+	char keybuf[100], valuebuf[100];
+	bool recno;
+
+	/* Reserve requires a running transaction. */
+	testutil_check(session->begin_transaction(session, NULL));
```
ddanderson: I don't see a matching commit_transaction, and I'm confused about the comment "Reserve requires a running transaction."

test/csuite/wt4156_metadata_salvage/main.c
```diff
@@ -0,0 +1,462 @@
...
+	/* Open a cursor. */
+	testutil_check(session->open_cursor(session, uri, NULL, NULL, &cursor));
+
+	/* Operations change based on the key/value formats. */
+	recno = strcmp(cursor->key_format, "r") == 0;
+	if (recno)
+		cursor->set_key(cursor, (uint64_t)1);
+	else {
+		strcpy(keybuf, KEY);
+		cursor->set_key(cursor, keybuf);
```
ddanderson: I don't think you need `keybuf`, just use the constant `KEY`.

test/csuite/wt4156_metadata_salvage/main.c
```diff
@@ -0,0 +1,462 @@
...
+	 * Open the file, read its contents. Find the string "corrupt" and
+	 * modify one byte at that offset. That will cause a checksum error
+	 * when WiredTiger next reads it.
+	 */
+	sprintf(path, "%s/%s", home, WT_METAFILE);
+	testutil_check(stat(path, &sb));
+	meta_size = (size_t)sb.st_size;
+	buf = dmalloc(meta_size);
+	memset(buf, 0, meta_size);
+	if ((fp = fopen(path, "r+")) == NULL)
```
ddanderson: I think we should have the `'b'` option for `fopen`.  Not sure if this will ever run on Windows, but just in case.

src/conn/conn_api.c
```diff
@@ -2734,6 +2743,12 @@ wiredtiger_open(const char *home, WT_EVENT_HANDLER *event_handler,
 	 */
 	WT_ERR(__wt_turtle_init(session));
 
+	/*
+	 * If the user wants to salvage, do so before opening the
+	 * metadata cursor.
+	 */
+	if (F_ISSET(conn, WT_CONN_SALVAGE))
```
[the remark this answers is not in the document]
sueloverso: This feels like a chicken and egg problem. Although moving the call works with the test case I have, this makes me very uneasy. The `wt_turtle_init` call does a lot of work and may create a metadata file from backup files by moving it into place (overwriting anything we did if we salvaged before this call).
I think we want to salvage the metadata that we're going to use and leave the call where it is. If there are other cases added later for dealing with the turtle file, calls to another new function can be added later.

sueloverso:
@lukech That is a good question and I am not sure. The `recover=salvage` API change is not on the `master` branch, therefore it has not yet been released. So I think we don't need to change the upgrading documentation to note its removal. It is just something that came and went on the `develop` branch. @agorrod is that correct?

sueloverso:
Thanks @ddanderson and @agorrod  I have pushed changes to address all the comments.

lukech:
@sueloverso You are right - the `recover=salvage` option introduced by WT-3276 which is not in WiredTiger master branch. Thanks for pointing that out. I agree there's no need to change the upgrading doc.

agorrod:
Thanks @sueloverso I'm happy with this now. I don't expect fallout, but I've switched [stress](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer) testing over here to give us more confidence before merging.

agorrod:
Looks like this causes issues with test/format:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/21320/console
I've switched testing back to develop

sueloverso:
I can reproduce (without sanitizer) and I'm chasing.

sueloverso:
@agorrod I found the issue. It is that the decompress function returned WT_ERROR and the code in `bt_slvg.c` originally converted that to 0. My change had the code looking only for `WT_DATA_CORRUPTION`. This change:
https://github.com/wiredtiger/wiredtiger/blob/1e869355ceb2699b4329cd7d4ddb917149fc5cda/src/btree/bt_slvg.c#L399
fixes that issue but brings up a general discussion.

There are a number of places (search for "return (WT_ERROR)" in the codebase on this branch) where it does not feel appropriate to return `WT_DATA_CORRUPTION` but retain the `WT_ERROR`. However, it seems possible that those paths could get returned in some corruption situations (similarly for decryption errors). It makes me question the feasibility of having a new return value `WT_DATA_CORRUPTION` or at least only having it at the end of `wiredtiger_open` and translating any `WT_ERROR` from that to `WT_DATA_CORRUPTION`.

lukech:
There's a cache stuck failure in below job with WT-4156 feature branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/21333/
My understanding is changes in this ticket so far have negligible impact to test/format cache handling, and the failure could be unrelated.

agorrod:
You are right @lukech - that is failure is WT-4119

sueloverso:
@agorrod This is again ready for merge. I changed all the compressor locations to return WT_DATA_CORRUPTION per our discussion. And the stress-sanitizer job has been running this branch for a day. We'll let it run with these latest changes until the beginning of your week.

sueloverso:
@agorrod This is ready. It has run the Jenkins stress sanitizer job successfully for 4 days modulo known failures for WT-4119. (I'm going to move that back to develop now.)

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

agorrod:
Thanks @sueloverso lgtm

PR #4220, opened 2018-08-15
WT-4233 Disable random_abort on OS/X, it is currently unreliable

ddanderson:


ddanderson:
@lukech, can you please review?

agorrod:
@ddanderson I was thinking about this overnight, and realized that we've seen failures in other automated tests as well (e.g: via MongoDB crash recovery tests). Do you think it would make sense to instead change the log salvage failure to a warning message on OS X - so we don't need to keep patching tests and test environments?

ddanderson:
@agorrod, I'm not yet convinced of that approach. We'd have some system-conditional code in logging, which seems less than optimal.  I _think_ there are a small number of tests (like this one that involve signaling) that tickle this bug. So I think we can deal with it by patching tests. Though I guess we don't really know until we go down that path.

Maybe we should ask instead what we want the user experience to be?  If the bug is triggered and less data is written than we think should be (some earlier writes are lost), should that be cause for a 1) fatal error during normal recovery that requires salvage 2) warning message 3) no message.

ddanderson:
I'm closing this pull request.  As discussed in the ticket https://jira.mongodb.org/browse/WT-4233, we need to take a different approach.  A new pull request is in #4227.

PR #4222, opened 2018-08-16
WT-4090 low priority reads

dgottlieb:
This is not ready for merging.

@michaelcahill @sueloverso, I've found all the callers of the btree walk methods and `__wt_page_in` and either:
* Used an immediately available `WT_CURSOR_BTREE` to inform the local flags variable.
* Or left a comment asking "can this be called from contexts with a cursor that we care about honoring a `read_once` setting".

Please review those assessments and let me know if it's important for potential cursor information to be plumbed through deeper in any of those locations. I'm also happy to dig up another layer if you need more investigation to make a judgment (the btree splitting methods were suspicious to me, but perhaps they're only called in the context of a writer and less important to honor any `read_once` settings).

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -691,7 +691,7 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 			__wt_page_evict_soon(session, cbt->ref);
 		cbt->page_deleted_count = 0;
 
-		if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
+		if (F_ISSET(cbt, WT_CBT_READ_ONCE))
 			LF_SET(WT_READ_WONT_NEED);
 		WT_ERR(__wt_tree_walk(session, &cbt->ref, flags));
 		WT_ERR_TEST(cbt->ref == NULL, WT_NOTFOUND);
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -691,6 +691,8 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 			__wt_page_evict_soon(session, cbt->ref);
 		cbt->page_deleted_count = 0;
 
+		if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
+			LF_SET(WT_READ_WONT_NEED);
 		WT_ERR(__wt_tree_walk(session, &cbt->ref, flags));
 		WT_ERR_TEST(cbt->ref == NULL, WT_NOTFOUND);
 	}
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -261,6 +261,7 @@ restart:	/*
 		 * On other error, simply return, the swap call ensures we're
 		 * holding nothing on failure.
 		 */
+		// WT-4090: Any `read_once` cursor flags to consider?
 descend:	if ((ret = __wt_page_swap(
 		    session, current, descent, flags)) == 0) {
 			current = descent;
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -169,7 +169,8 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
  *	Find a random page in a tree for either sampling or eviction.
  */
 int
-__wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, bool eviction)
+__wt_random_descent(WT_SESSION_IMPL *session,
+    WT_REF **refp, bool eviction, uint32_t additional_flags)
 {
 	WT_BTREE *btree;
 	WT_DECL_RET;
@@ -186,10 +187,12 @@ __wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, bool eviction)
 
 	/* Eviction should not be tapped to do eviction. */
 	if (eviction)
-		flags = WT_READ_CACHE | WT_READ_NO_EVICT | WT_READ_NO_GEN |
-		    WT_READ_NO_WAIT | WT_READ_NOTFOUND_OK | WT_READ_RESTART_OK;
+		flags = additional_flags |
+			WT_READ_CACHE | WT_READ_NO_EVICT |
+			WT_READ_NO_GEN | WT_READ_NO_WAIT |
+			WT_READ_NOTFOUND_OK | WT_READ_RESTART_OK;
 	else
-		flags = WT_READ_RESTART_OK;
+		flags = additional_flags | WT_READ_RESTART_OK;
 
 	if (0) {
 restart:	/*
@@ -302,11 +305,15 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	WT_UPDATE *upd;
 	wt_off_t size;
 	uint64_t n, skip;
+	uint32_t descent_flags;
 	bool valid;
 
 	btree = cbt->btree;
 	cursor = &cbt->iface;
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
+	descent_flags = 0;
+	if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
+		FLD_SET(descent_flags, WT_CBT_READ_WONT_NEED);
 
 	/*
 	 * Only supports row-store: applications can trivially select a random
@@ -337,7 +344,8 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	if (cbt->ref == NULL || cbt->next_random_sample_size == 0) {
 		WT_ERR(__cursor_func_init(cbt, true));
 		WT_WITH_PAGE_INDEX(session,
-		    ret = __wt_random_descent(session, &cbt->ref, false));
+		    ret = __wt_random_descent(
+			session, &cbt->ref, false, descent_flags));
 		if (ret == 0)
 			goto random_page_entry;
 
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -184,6 +184,11 @@ __wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, uint32_t flags)
 	btree = S2BT(session);
 	current = NULL;
 	retry = 100;
+	/*
+	 * This function is called by eviction to find a random page in the
+	 * cache. That case is indicated by the WT_READ_CACHE flag. Ordinary
+	 * lookups in a tree will read pages into cache as needed.
+	 */
 	eviction = LF_ISSET(WT_READ_CACHE);
 
 	if (0) {
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -261,7 +261,6 @@ restart:	/*
 		 * On other error, simply return, the swap call ensures we're
 		 * holding nothing on failure.
 		 */
-		// WT-4090: Any `read_once` cursor flags to consider?
 descend:	if ((ret = __wt_page_swap(
 		    session, current, descent, flags)) == 0) {
 			current = descent;
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -303,9 +303,9 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	btree = cbt->btree;
 	cursor = &cbt->iface;
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
-	read_flags = 0;
-	if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
-		FLD_SET(read_flags, WT_CBT_READ_WONT_NEED);
+	read_flags = WT_READ_RESTART_OK;
+	if (F_ISSET(cbt, WT_CBT_READ_ONCE))
+		FLD_SET(read_flags, WT_READ_WONT_NEED);
 
 	/*
 	 * Only supports row-store: applications can trivially select a random
@@ -336,8 +336,7 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	if (cbt->ref == NULL || cbt->next_random_sample_size == 0) {
 		WT_ERR(__cursor_func_init(cbt, true));
 		WT_WITH_PAGE_INDEX(session,
-		    ret = __wt_random_descent(
-			session, &cbt->ref, read_flags));
+		    ret = __wt_random_descent(session, &cbt->ref, read_flags));
 		if (ret == 0)
 			goto random_page_entry;
 
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -169,30 +169,22 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
  *	Find a random page in a tree for either sampling or eviction.
  */
 int
-__wt_random_descent(WT_SESSION_IMPL *session,
-    WT_REF **refp, bool eviction, uint32_t additional_flags)
+__wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, uint32_t flags)
 {
 	WT_BTREE *btree;
 	WT_DECL_RET;
 	WT_PAGE *page;
 	WT_PAGE_INDEX *pindex;
 	WT_REF *current, *descent;
-	uint32_t flags, i, entries, retry;
+	uint32_t i, entries, retry;
+	bool eviction;
 
 	*refp = NULL;
 
 	btree = S2BT(session);
 	current = NULL;
 	retry = 100;
-
-	/* Eviction should not be tapped to do eviction. */
-	if (eviction)
-		flags = additional_flags |
-			WT_READ_CACHE | WT_READ_NO_EVICT |
-			WT_READ_NO_GEN | WT_READ_NO_WAIT |
-			WT_READ_NOTFOUND_OK | WT_READ_RESTART_OK;
-	else
-		flags = additional_flags | WT_READ_RESTART_OK;
+	eviction = LF_ISSET(WT_READ_CACHE);
 
 	if (0) {
 restart:	/*
@@ -305,15 +297,15 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	WT_UPDATE *upd;
 	wt_off_t size;
 	uint64_t n, skip;
-	uint32_t descent_flags;
+	uint32_t read_flags;
 	bool valid;
 
 	btree = cbt->btree;
 	cursor = &cbt->iface;
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
-	descent_flags = 0;
+	read_flags = 0;
 	if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
-		FLD_SET(descent_flags, WT_CBT_READ_WONT_NEED);
+		FLD_SET(read_flags, WT_CBT_READ_WONT_NEED);
 
 	/*
 	 * Only supports row-store: applications can trivially select a random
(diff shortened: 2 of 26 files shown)


src/btree/bt_random.c
```diff
@@ -169,7 +169,8 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
  *	Find a random page in a tree for either sampling or eviction.
  */
 int
-__wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, bool eviction)
+__wt_random_descent(WT_SESSION_IMPL *session,
+    WT_REF **refp, bool eviction, uint32_t additional_flags)
```
dgottlieb: This method calls `__wt_page_swap`, but without enough context to pass in any cursor flags. I wasn't sure how to funnel the context in here. There are three callers into this method. One is from eviction and the other two are the next/prev implementations for random btree cursors. For this method I've considered:
* Passing in a `cbt` pointer. That seemed inappropriate as `__evict_walk_tree` didn't seem to manage those, so that caller would be passing `null`.
* Passing in a second boolean, `read_once`. From my limited exposure to WT code, this approach did not seem widely adopted when a slightly more flexible API might be useful. I didn't think the small addition to clarity necessarily offset flexibility. Though methods managing two sets of "flag" variables seems uncommon as well.
* Removing the `eviction` boolean and having the callers pass in `flags` altogether. However `eviction` is also used in a couple other spots in the method such that it couldn't entirely be removed.

src/btree/bt_random.c
```diff
@@ -186,10 +187,12 @@ __wt_random_descent(WT_SESSION_IMPL *session, WT_REF **refp, bool eviction)
 
 	/* Eviction should not be tapped to do eviction. */
 	if (eviction)
-		flags = WT_READ_CACHE | WT_READ_NO_EVICT | WT_READ_NO_GEN |
-		    WT_READ_NO_WAIT | WT_READ_NOTFOUND_OK | WT_READ_RESTART_OK;
+		flags = additional_flags |
```
dgottlieb: I think I may have taken the style rule of "successive wrapped lines should be longer than the above line" too far.

src/btree/bt_random.c
```diff
@@ -302,11 +305,15 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	WT_UPDATE *upd;
 	wt_off_t size;
 	uint64_t n, skip;
+	uint32_t descent_flags;
```
dgottlieb: Was unsure if I should just keep this as simple `flags` or because this was only being forwarded into the descent method, a little more descriptive.

src/btree/col_srch.c
```diff
@@ -73,7 +73,7 @@ __wt_col_search(WT_SESSION_IMPL *session,
 	WT_PAGE_INDEX *pindex, *parent_pindex;
 	WT_REF *current, *descent;
 	uint64_t recno;
-	uint32_t base, indx, limit;
+	uint32_t base, indx, limit, page_swap_flags;
```
dgottlieb: Same naming question as `descent_flags`

src/cursor/cur_file.c
```diff
@@ -693,6 +693,10 @@ __curfile_create(WT_SESSION_IMPL *session,
 		cacheable = false;
 	}
 
+	WT_ERR(__wt_config_gets_def(session, cfg, "read_once", 0, &cval));
+	if (cval.val != 0)
+		F_SET(cbt, WT_CBT_READ_WONT_NEED);
```
dgottlieb: I'm still investigating exactly what it takes to address this, but the patch as shown, to my knowledge only "works" if `read_once` is passed into `open_cursor` specifically for a `file:` URI or a simple `table:` URI. I suspect when higher level cursors create a lower level btree cursor, the configuration is not forwarded down? Still investigating.

src/include/cursor.h
```diff
@@ -228,10 +228,11 @@ struct __wt_cursor_btree {
 #define	WT_CBT_ITERATE_NEXT	0x004u	/* Next iteration configuration */
 #define	WT_CBT_ITERATE_PREV	0x008u	/* Prev iteration configuration */
 #define	WT_CBT_NO_TXN   	0x010u	/* Non-txn cursor (e.g. a checkpoint) */
-#define	WT_CBT_RETRY_NEXT	0x020u	/* Next, resulted in prepare conflict */
-#define	WT_CBT_RETRY_PREV	0x040u	/* Prev, resulted in prepare conflict */
-#define	WT_CBT_SEARCH_SMALLEST	0x080u	/* Row-store: small-key insert list */
-#define	WT_CBT_VAR_ONPAGE_MATCH	0x100u	/* Var-store: on-page recno match */
+#define	WT_CBT_READ_WONT_NEED	0x020u	/* Page in with WT_READ_WONT_NEED */
```
dgottlieb: I'm not happy with the comment here, but I couldn't make it fit in under 80 characters. I wasn't sure how to style the definition if it exceeded 80 characters.

src/btree/bt_random.c
sueloverso: You're right that once we start needing multiple boolean args, what we tend to do is have function flags that then gives flexibility to add new flags in the future too. You're on the right path in your thoughts. My suggestion would be to:
* Remove the `eviction` boolean and only have a new arg `flags`.
* Rename the local `flags` to `page_flags`.
* Callers pass in (via new `flags`) new flags `WT_DESCENT_EVICTION | WT_DESCENT_READ_ONCE`.
* Code then can be:
```
bool eviction;
eviction = LF_ISSET(WT_DESCENT_EVICTION);
if (eviction)
    page_flags = ...;
else
   page_flags = ...;
if (LF_ISSET(WT_DESCENT_READ_ONCE))
    FLD_SET(page_flags, ...);
```

src/btree/bt_random.c
sueloverso: The more descriptive name is helpful since it is only sent to descent.

src/btree/bt_random.c
```diff
@@ -302,11 +305,15 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
...
 	bool valid;
 
 	btree = cbt->btree;
 	cursor = &cbt->iface;
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
+	descent_flags = 0;
+	if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
+		FLD_SET(descent_flags, WT_CBT_READ_WONT_NEED);
```
sueloverso: If you take my suggestion this would become `WT_DESCENT_READ_ONCE` (or maybe call it `WT_DESCENT_READ_WONT_NEED`).

src/evict/evict_lru.c
```diff
@@ -1793,7 +1793,7 @@ __evict_walk_tree(WT_SESSION_IMPL *session,
 		if (btree->evict_ref == NULL) {
 			/* Ensure internal pages indexes remain valid */
 			WT_WITH_PAGE_INDEX(session, ret = __wt_random_descent(
-			    session, &btree->evict_ref, true));
+			    session, &btree->evict_ref, true, 0));
```
sueloverso: This would become `WT_DESCENT_EVICTION`.

src/btree/bt_random.c
michaelcahill: Rather than creating a new set of "descent" flags, I'm inclined to use the existing read flags here -- i.e., have callers pass in exactly the flags that will be passed to `__wt_page_swap`.

Also as a general comment, I'd go with `read_flags` as the variable name generally rather than `descent_flags`, `page_swap_flags`, etc.

src/cursor/cur_file.c
dgottlieb: Concern withdrawn. I can see the configuration being passed down.

src/btree/bt_random.c
```diff
@@ -169,27 +169,22 @@ __wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
...
 	current = NULL;
 	retry = 100;
-
-	/* Eviction should not be tapped to do eviction. */
-	if (eviction)
-		flags = WT_READ_CACHE | WT_READ_NO_EVICT | WT_READ_NO_GEN |
-		    WT_READ_NO_WAIT | WT_READ_NOTFOUND_OK | WT_READ_RESTART_OK;
-	else
-		flags = WT_READ_RESTART_OK;
+	eviction = LF_ISSET(WT_READ_CACHE);
```
sueloverso: A comment here about why `WT_READ_CACHE` means eviction would be helpful.

src/btree/bt_random.c
```diff
@@ -302,11 +297,15 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
...
 	uint64_t n, skip;
+	uint32_t read_flags;
 	bool valid;
 
 	btree = cbt->btree;
 	cursor = &cbt->iface;
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
+	read_flags = 0;
+	if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
+		FLD_SET(read_flags, WT_CBT_READ_WONT_NEED);
```
sueloverso: Two things:
1. I think `read_flags` needs to be initialized to `WT_READ_RESTART_OK`.
2. I think you need to set `WT_READ_WONT_NEED` into `read_flags` (not the CBT value).

src/btree/bt_random.c
```diff
@@ -337,7 +336,8 @@ __wt_btcur_next_random(WT_CURSOR_BTREE *cbt)
 	if (cbt->ref == NULL || cbt->next_random_sample_size == 0) {
 		WT_ERR(__cursor_func_init(cbt, true));
 		WT_WITH_PAGE_INDEX(session,
-		    ret = __wt_random_descent(session, &cbt->ref, false));
+		    ret = __wt_random_descent(
+			session, &cbt->ref, read_flags));
```
sueloverso: Nit: I think these lines can be joined again.

sueloverso:
One more post-review thought: Do we want to add a statistic to count and know how many cursors are opened this way? We don't count other things like `readonly` cursors, but it might be interesting. I could go either way on it.

src/btree/bt_curnext.c
```diff
@@ -691,6 +691,8 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 			__wt_page_evict_soon(session, cbt->ref);
 		cbt->page_deleted_count = 0;
 
+		if (F_ISSET(cbt, WT_CBT_READ_WONT_NEED))
```
sueloverso: @dgottlieb I had another review thought on this. I think it would be good to name the CBT flag `WT_CBT_READ_ONCE` both to keep it similar to the API naming and to avoid the similarity with the cache flag so it is visually more obvious.

src/btree/bt_random.c
dgottlieb: @michaelcahill can you provide some assistance with this?

src/btree/bt_random.c
dgottlieb: Correct on both accounts, thanks for the catch!

dgottlieb:
I'll let @michaelcahill make the call on whether a `read_once` statistics makes sense and what exactly it should be counting (only application cursors using `read_once`? every time `WT_READ_WONT_NEED` is exercised? something in between?)

Otherwise, I've now added a test which tries to observe eviction side-effects. I had to explore a bit how to trigger eviction (while preferably keeping the document count low, for debug output sanity). I'm not sure if I've stumbled upon a happy solution. I'm also unsure if the statistics I'm asserting on are maybe more flaky than some alternatives. I chose them because toggling `read_once` give resulted in 0 <-> ~20, which seemed like pretty definitive evidence of the flag being properly passed down.

src/btree/bt_random.c
michaelcahill: I'd add a comment saying something like "This function is called by eviction to find a random page in cache.  That case is indicated by the WT_READ_CACHE flag.  Ordinary lookups in a tree will read pages into cache as needed."

src/cursor/cur_file.c
```diff
@@ -693,6 +693,10 @@ __curfile_create(WT_SESSION_IMPL *session,
 		cacheable = false;
 	}
 
+	WT_ERR(__wt_config_gets_def(session, cfg, "read_once", 0, &cval));
+	if (cval.val != 0)
+		F_SET(cbt, WT_CBT_READ_ONCE);
```
michaelcahill: I think we either have to set `cacheable = false` here and add a `read_once` case to https://github.com/wiredtiger/wiredtiger/blob/develop/src/cursor/cur_std.c#L731-L761, or deal with setting / clearing `WT_CBT_READ_ONCE` in this code: https://github.com/wiredtiger/wiredtiger/blob/develop/src/cursor/cur_std.c#L801-L822

I'd prefer the latter, so that we can use the cursor cache with `read_once` set.

test/suite/test_read_once.py
```diff
@@ -0,0 +1,81 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# test_timestamp13.py
+#   Timestamps: session query_timestamp
```
sueloverso: Update the test name and brief description here!
I also suggest naming it `test_cursor15.py` (and likewise in the class and test definitions below) so that it gets run as part of `run.py cursor`.

test/suite/test_read_once.py
```diff
@@ -0,0 +1,81 @@
...
+        conn_config = 'cache_size=1M,statistics=(all)'
+        self.conn = wiredtiger.wiredtiger_open(self.home, conn_config)
+        self.session = self.conn.open_session()
+
+    def test_read_once(self):
+        # This test is configured to use 1MB of cache. It will insert 20
+        # documents, each 100KB. Manipulate the table to result in one page per
+        # document.
+        self.session.create(self.uri,
+                            'key_format=i,value_format=S,leaf_page_max=108K,leaf_value_max=108K')
```
sueloverso: The continuation line should be indented 4 spaces.

test/suite/test_read_once.py
```diff
@@ -0,0 +1,81 @@
...
+    def ConnectionOpen(self):
+        '''
+        Specify a custom ConnectionOpen method to flush WT's cache during a test
+        after inserting data.
+        '''
+        self.home = '.'
+
+        conn_config = 'cache_size=1M,statistics=(all)'
+        self.conn = wiredtiger.wiredtiger_open(self.home, conn_config)
+        self.session = self.conn.open_session()
```
sueloverso: I think you do not need the `ConnectionOpen` function.
You can just put `conn_config` up near the `uri` and it will do what you want (see below too).

test/suite/test_read_once.py
```diff
@@ -0,0 +1,81 @@
...
+                            'key_format=i,value_format=S,leaf_page_max=108K,leaf_value_max=108K')
+
+        cursor = self.session.open_cursor(self.uri, None, None)
+        for key in range(0, 20):
+            cursor[key] = '1' * (100 * 1024)
+        cursor.close()
+
+        # Restart the database to clear the cache.
+        self.conn.close()
+        self.ConnectionOpen()
```
sueloverso: (Related to above) you can just do `self.reopen_conn()` here instead of close/open and it will do what you want.

test/suite/test_read_once.py
```diff
@@ -0,0 +1,81 @@
...
+        for key, value in cursor:
+            pass
+        cursor.close()
+
+        # Although 2MB of data was touched, paging in from disk never had to
+        # perform eviction. Additionally, the eviction server was not enlisted
+        # in performing eviction.
+        stat_cursor = self.session.open_cursor('statistics:', None, None)
+        self.assertEqual(stat_cursor[stat.conn.cache_eviction_get_ref][2], 0)
+        self.assertEqual(stat_cursor[stat.conn.cache_eviction_server_evicting][2], 0)
```
sueloverso: I suggest putting the cursor walk and statistics check into a function, and then calling it twice. Once to walk with `read_once` and once to walk with a normal cursor and confirm the stats are non-zero.

test/suite/test_read_once.py
```diff
@@ -0,0 +1,81 @@
...
+#
+# test_timestamp13.py
+#   Timestamps: session query_timestamp
+#
+
+import random
+import wiredtiger, wttest
+from suite_subprocess import suite_subprocess
+from wiredtiger import stat
+from wtscenario import make_scenarios
```
sueloverso: I think you don't need most of these imports.

test/suite/test_read_once.py
sueloverso: Yes. You can run an individual test or a `prefix` set of tests.

test/suite/test_read_once.py
dgottlieb: Good idea. This is causing the test to fail which I think might be related to @michaelcahill's catch on the cursor caching.

test/suite/test_cursor15.py
```diff
@@ -0,0 +1,79 @@
...
+            # Without the flag, the same statistics can be observed to be
+            # non-zero.
+            stat_cursor = self.session.open_cursor('statistics:', None, None)
+            if stats_are_zero:
+                self.assertEqual(stat_cursor[stat.conn.cache_eviction_get_ref][2], 0)
+                self.assertEqual(stat_cursor[stat.conn.cache_eviction_server_evicting][2], 0)
+            else:
+                self.assertGreater(stat_cursor[stat.conn.cache_eviction_get_ref][2], 0)
+                self.assertGreater(stat_cursor[stat.conn.cache_eviction_server_evicting][2], 0)
+            stat_cursor.close()
```
sueloverso: Sorry I missed this the first time. You need to keep the main test stuff at the end of the test. If you look, every test file has this at the bottom:
```
if __name__ == '__main__':
    wttest.run()
```

src/cursor/cur_file.c
dgottlieb: I got #2 to work. That, in combination with @sueloverso's testing suggestion, I could observe the caching logic from being broken (not clearing the flag) to working. Two remarks:
1) Should we spend time coming up with a more preferable way to detect cursor caching bugs regarding `read_once`? A failing test result is certainly subtle to diagnose.
2) I looked at the first character of the uri + internal_uri to determine if the `WT_CURSOR` object was really a `WT_CURSOR_BTREE` object, but I'm not sure if I can just assume `internal_uri` is always available when pulling from the cache and the right thing to check?

test/suite/test_read_once.py
dgottlieb: I figured there was a better way. I had grepped for `wiredtiger_open` to find tests that finely controlled database lifetimes. I don't feel too bad for not discovering `reopen_conn`!

michaelcahill:
Thanks @dgottlieb, lgtm -- merging.

dgottlieb:
@sueloverso @michaelcahill 

I'm not surprised the eviction statistic was not stable enough. I'm not sure what options we have for testing this, I need help moving this forward.

I know using a performance test was suggested, but if the eviction statistic isn't reliable enough, I'd be surprised if the performance test could be trusted to not give false positives. Should we reconsider adding a `WT_READ_WONT_NEED` statistic? Leave the feature untested?

sueloverso:
@dgottlieb here are a few thoughts on possibilities:
* Which side of the statistics was not reliable? Was it only the non-`read_once`? If so, then one possibility is to only check stats are zero for `read_once` and have a comment in the test for the other case.
* Or at worst, remove the statistics checking. That tests the API but does not verify its internal effects but is probably sufficient for merging this ticket (barring adding the new statistic you suggested that Michael prefers not to add).

dgottlieb:
The `read_once` cursor had a `1` for one of the eviction statistics (the expected result was 0).

sueloverso:
Is it reliable to say read_once-stat < normal-stat?

PR #4223, opened 2018-08-17
WT-3735 Split stress test workgen

nehakhatri5:
A workload to stress page splits.

### DIFF
--- a/bench/workgen/runner/split_stress.py
+++ b/bench/workgen/runner/split_stress.py
@@ -41,71 +41,49 @@ from runner import *
 from wiredtiger import *
 from workgen import *
 
-# Helper functions.
-def op_append(ops, op):
-    if ops == None:
-        ops = op
-    else:
-        ops += op
-    return ops
-
-def make_op(optype, table, key, value = None):
-    if value == None:
-        return Operation(optype, table, key)
-    else:
-        return Operation(optype, table, key, value)
-
-def operations(optype, tables, key, value = None, ops_per_txn = 0, logtable = None):
-    txn_list = []
-    ops = None
-    nops = 0
-    for table in tables:
-        ops = op_append(ops, make_op(optype, table, key, value))
-        if logtable != None:
-            ops = op_append(ops, make_op(optype, logtable, logkey, value))
-        nops += 1
-        if ops_per_txn > 0 and nops % ops_per_txn == 0:
-            txn_list.append(txn(ops))
-            ops = None
-    if ops_per_txn > 0:
-        if ops != None:
-            txn_list.append(txn(ops))
-            ops = None
-        for t in txn_list:
-            ops = op_append(ops, t)
-    return ops
-
-# Connection and table configuration.
 context = Context()
-conn_config="create,cache_size=100MB,log=(enabled=false),statistics=[fast],statistics_log=(wait=5,json=false)"
-conn = wiredtiger_open("WT_TEST", conn_config)
-s = conn.open_session()
+# Connection configuration.
+conn_config = "cache_size=100MB,log=(enabled=false),statistics=[fast],statistics_log=(wait=5,json=false)"
+conn = wiredtiger_open("WT_TEST", "create," + conn_config)
+s = conn.open_session("")
 
-table_config="leaf_page_max=8k,internal_page_max=8k,leaf_item_max=1433,internal_item_max=3100,type=file,memory_page_max=1MB,split_deepen_min_child=100"
-
+# Table configuration.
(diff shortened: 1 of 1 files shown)


nehakhatri5:
@agorrod I have created a workgen test that stresses the page splits. The associated wtperf config also specified `reopen_connection=false`. It is not clear that how to set this option in the workgen test. It will good to set this option in the new workgen test too. ptal.
Meanwhile the `wiredtiger-pull-request-compilers` has failed due to unforeseen reason.

agorrod:
autobuild: retest this please

agorrod:
@ddanderson it would be great if you can take a look at this - we are trying to upskill on workgen - so feedback on good usage patterns would be appreciated.

ddanderson:
@nehakhatri5, this looks basically good, I'm glad you're getting into workgen.  To answer the question about `reopen_connection=false`, that is the default in workgen. You already have `conn = wiredtiger_open("WT_TEST", conn_config)` in your script.  If you want to reopen your connection, you use the WT Python interface to do the reopen.  That is: `conn.close()` and `conn = wiredtiger_open(....)`, and pass the new connection into your second `workload.run(conn)`.  Hope that makes sense.

All that said, I _think_ there might be a more straightforward way to port or directly run an existing wtperf script.  And that is via the `bench/workgen/wtperf.py` script, which takes the wtperf configuration file as a command line arg and generates the Python workgen script, and runs it internally.  The `wtperf.py` script does not know all the wtperf options yet.  If it stumbles, you can comment out the offending line, run `python wtperf.py --python perf_script.wtperf > myperf.py` and then edit `myperf.py` to your liking.

bench/workgen/runner/split_stress.py
```diff
@@ -0,0 +1,89 @@
...
+conn_config = "cache_size=100MB,log=(enabled=false),statistics=[fast],statistics_log=(wait=5,json=false)"
+conn = wiredtiger_open("WT_TEST", "create," + conn_config)
+s = conn.open_session("")
+
+# Table configuration.
+table_config = "leaf_page_max=8k,internal_page_max=8k,leaf_item_max=1433,internal_item_max=3100,type=file,memory_page_max=1MB,split_deepen_min_child=100"
+tables = []
+table_count = 3
+for i in range(0, table_count):
+    tname = "file:test" + str(i) + ".wt"
```
agorrod: Could you switch this to `tname = "file:test" + str(i)` instead? We discourage direct use of `file:` uris.

bench/workgen/runner/split_stress.py
```diff
@@ -0,0 +1,89 @@
...
+# and multiple threads inserting keys in random order. It stresses the page
+# splits in order to catch split races.
+#
+from runner import *
+from wiredtiger import *
+from workgen import *
+
+context = Context()
+# Connection configuration.
+conn_config = "cache_size=100MB,log=(enabled=false),statistics=[fast],statistics_log=(wait=5,json=false)"
```
agorrod: Statistics logging once per second is more consistent with other uses.

bench/workgen/runner/split_stress.py
```diff
@@ -0,0 +1,89 @@
...
+
+# Run phase.
+ops = Operation(Operation.OP_INSERT, tables[0])
+ops = op_multi_table(ops, tables, False)
+thread0 = Thread(ops)
+
+workload = Workload(context, 20 * thread0)
+workload.options.report_interval=5
+workload.options.run_time=10
+print('Split stress workload running...')
```
agorrod: Does workgen have an option to automatically print out information like this?

bench/workgen/runner/split_stress.py
```diff
@@ -0,0 +1,89 @@
...
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+
+# Drive a constant high workload through, even if WiredTiger isn't keeping
+# up by dividing the workload across a lot of threads. This needs to be
+# tuned to the particular machine so the workload is close to capacity in the
+# steady state, but not overwhelming.
```
agorrod: Is the final sentence here true? It doesn't seem to match the implementation.

bench/workgen/runner/split_stress.py
nehakhatri5: I could not find an option to automatically print.

bench/workgen/runner/split_stress.py
nehakhatri5: Err, copy-paste error. Fixed.

agorrod:
lgtm - I'll merge.

PR #4229, opened 2018-08-20
WT-3637 Fix a heap use after free from evicting of a page that just split (v3.4 backport)

lukech:
(cherry picked from commit a86638be657f647ce1656d24d8f653e72a9e6eee)

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -391,6 +391,18 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
 	 */
 	WT_PAGE_LOCK(session, page);
 
+	/*
+	 * Now that the page is locked, if attempting to evict it, check again
+	 * whether eviction is permitted. The page's state could have changed
+	 * while we were waiting to acquire the lock (e.g., the page could have
+	 * split).
+	 */
+	if (LF_ISSET(WT_REC_EVICT) &&
+	    !__wt_page_can_evict(session, ref, NULL)) {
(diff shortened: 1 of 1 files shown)


src/reconcile/rec_write.c
```diff
@@ -391,6 +391,18 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
 	 */
 	WT_PAGE_LOCK(session, page);
 
+	/*
+	 * Now that the page is locked, if attempting to evict it, check again
+	 * whether eviction is permitted. The page's state could have changed
+	 * while we were waiting to acquire the lock (e.g., the page could have
+	 * split).
+	 */
+	if (LF_ISSET(WT_REC_EVICT) &&
```
agorrod: @michaelcahill This flag changed from `WT_EVICTING` to `WT_REC_EVICT` in WT-3435 - could you confirm that the right thing to do for this backport is to use `WT_EVICTING`?

agorrod: Link to the commit in WT-3435 if it helps
https://github.com/wiredtiger/wiredtiger/commit/c227e057df39d2ef4c3eab9bf47e08cc84402bdb

michaelcahill: Yes, the `WT_EVICTING` flag was renamed `WT_REC_EVICT` to create a group of reconciliation flags with a common prefix, so it is correct to use `WT_EVICTING` for the backport.

lukech: Thanks for the confirmation @michaelcahill

PR #4230, opened 2018-08-21
WT-4256 Loosen check during rollback_to_stable.

michaelcahill:
In particular, cope with updates that are already rolled back, and don't fire for updates that were never assigned a timestamp unless "assert=(commit_timestamp=X)" is configured for the table.

### DIFF
--- a/src/txn/txn_rollback_to_stable.c
+++ b/src/txn/txn_rollback_to_stable.c
@@ -110,8 +110,8 @@ __txn_abort_newer_update(WT_SESSION_IMPL *session,
 		/*
 		 * Updates with no timestamp will have a timestamp of zero and
 		 * will never be rolled back.  If the table is configured for
-		 * strict timestamp checking, assert that we never have such
-		 * updates in the list ahead of updates we do roll back.
+		 * strict timestamp checking, assert that all more recent
+		 * updates were also rolled back.
 		 */
 		if (upd->txnid == WT_TXN_ABORTED && upd == first_upd)
 			first_upd = upd->next;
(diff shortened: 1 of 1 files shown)


agorrod:
@michaelcahill this change looks pretty good to me. I'm still trying to figure out what happens if an update chain has a mixture of updates with timestamps and those without - I think this will mark as aborted any with older timestamps from the middle of the list  - which feels odd.

It also got me wondering about how this would play with modify. Are we sufficiently happy that the issue here is only around truncate in capped collections, so the code won't do weird things with complex update chains?

@sueloverso I've added you as a reviewer - it would be great if you can take a look at the restructure.

src/txn/txn_rollback_to_stable.c
```diff
@@ -98,31 +98,38 @@ err:	if (ret == 0) {
...
-	for (next_upd = upd; next_upd != NULL; next_upd = next_upd->next) {
+	for (upd = first_upd; upd != NULL; upd = upd->next) {
 		/*
-		 * Updates with no timestamp will have a timestamp of zero
-		 * which will fail the following check and cause them to never
-		 * be aborted.
+		 * Updates with no timestamp will have a timestamp of zero and
+		 * will never be rolled back.  If the table is configured for
+		 * strict timestamp checking, assert that we never have such
+		 * updates in the list ahead of updates we do roll back.
```
sueloverso: Minor nit: I find `ahead` really confusing because I can interpret it two ways and I don't know which you mean. Do you mean ahead of `upd` that is walking the list and we haven't gotten to it yet (i.e. older) or ahead as in earlier in the list (i.e. younger)? Can you use `earlier/later in the list` or `younger/older`?

agorrod:
Thanks @michaelcahill lgtm - I'll merge once you've fixed the comment per the request from Sue

PR #4231, opened 2018-08-21
WT-4257 Don't assume timestamps from lookaside are aligned in memory.

michaelcahill:


agorrod:
Thanks @michaelcahill lgtm.

PR #4234, opened 2018-08-21
WT-4251 Prepared updates cannot be discarded.

michaelcahill:
We have a check in ordinary eviction that prepared updates are not discarded, but not in other places (e.g., the code that garbage collects obsolete updates).  Rather than find all of the places, this change checks for prepared updates as part of the "visible all" check for updates.


michaelcahill:
Yes, thanks @keithbostic -- we deal with prepared updates explicitly just above there, at https://github.com/wiredtiger/wiredtiger/blob/develop/src/reconcile/rec_write.c#L1337-L1353 so I don't think the change to `__wt_txn_upd_visible_all` will cause any behavior changes here.

PR #4235, opened 2018-08-22
WT-4239 Checkpoint doing insert-splits can corrupt the tree.

keithbostic:
@agorrod, @michaelcahill and I figured this out today, the problem is checkpoint can do insert-splits and that can corrupt the tree.

I'm running overnight to confirm, but I think this is ready for review.

This should probably be backported to any release where checkpoint attempts to evict pages.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -67,8 +67,8 @@ __split_verify_intl_key_order(WT_SESSION_IMPL *session, WT_PAGE *page)
 	WT_ITEM *next, _next, *last, _last, *tmp;
 	WT_REF *ref;
 	uint64_t recno;
+	uint32_t slot;
 	int cmp;
-	bool first;
 
 	btree = S2BT(session);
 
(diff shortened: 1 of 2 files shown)


keithbostic:
autobuild: retest this please

agorrod:
autobuild: retest this please

It's failed twice on the same machine due to a timeout - I've taken that machine off rotation. Let's try again to see if it's something in this branch or load on the machine causing the timeout.

agorrod:
Looks like this branch is hanging while running fops in make check.

keithbostic:
> It's failed twice on the same machine due to a timeout - I've taken that machine off rotation. Let's try again to see if it's something in this branch or load on the machine causing the timeout.

This was real, and I've pushed a fix for it in 65253a2.
I don't see where you pulled kodkod out of rotation, so I didn't reverse that change.

keithbostic:
@agorrod, this ran overnight without failure (in either the stress-sanitizer tests or my local format runs). I'm going to merge it so I can watch it run today.

agorrod:
Thanks @keithbostic - the additional fix looked right to me.

> I don't see where you pulled kodkod out of rotation, so I didn't reverse that change.

Sorry - I added it back in after confirming that the timeout wasn't machine related.

PR #4236, opened 2018-08-22
WT-4119 Avoid restarts updating / removing during a column store scan

michaelcahill:


michaelcahill:
@keithbostic, let's talk this one through -- I tried a few things here and this was the smallest change but the layering is ... questionable.  I'd be happy to find a cleaner solution but I hope this illustrates the problem.

keithbostic:
@michaelcahill, the changes in this ticket are ready for review, but the pull request won't pass testing until #4245 is merged.

agorrod:
lgtm - will update once WT-4267 is merged.

agorrod:
Thanks @keithbostic and @michaelcahill this change lgtm now. I'll go ahead merge. @michaelcahill I'm confident enough about this change to merge, but think it's still worth you reviewing after the fact.

PR #4237, opened 2018-08-22
WT-4262 In eviction of internal pages, lock deleted children.

michaelcahill:
Elsewhere we assume that flipping `ref->state` from `DELETED` to `LOCKED` makes it safe to free `ref->page_del`.  If we're attempting to evict an internal page, we have it locked, so no thread can descend into it, but a walk could be attempting to couple into a child page, so we can't assume exclusive access to all of the children.

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -450,6 +450,7 @@ static int
 __evict_child_check(WT_SESSION_IMPL *session, WT_REF *parent)
 {
 	WT_REF *child;
+	bool active;
 
 	WT_INTL_FOREACH_BEGIN(session, parent->page, child) {
 		switch (child->state) {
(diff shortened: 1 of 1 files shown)


src/evict/evict_page.c
```diff
@@ -460,11 +461,13 @@ __evict_child_check(WT_SESSION_IMPL *session, WT_REF *parent)
...
-			 * control can be running below our locked internal
-			 * page.
 			 */
-			if (__wt_page_del_active(session, child, true))
+			if (!__wt_atomic_casv32(
+			    &child->state, WT_REF_DELETED, WT_REF_LOCKED))
+				return (__wt_set_return(session, EBUSY));
+			active = __wt_page_del_active(session, child, true);
+			WT_PUBLISH(child->state, WT_REF_DELETED);
+			if (active)
```
keithbostic: I don't think the `WT_PUBLISH` call is useful, `__wt_page_del_active` hasn't written anything.

We could add a barrier to flush the child state update, but I don't feel strongly.

michaelcahill:
@keithbostic, yes, I tried to explain it in the commit comment but I'll put something inline: you're right, we shouldn't be able to descend (i.e., search) into an internal page at this point, but we can walk into it, which is what happened in this case.  From the ticket, thread T41 freed `page_del` as part of a walk in service of a `__curfile_prev`.

agorrod:
Thanks @michaelcahill

PR #4238, opened 2018-08-22
WT-4263 When copying a key for a lookaside write, use the right tree.

michaelcahill:
Previously this code juggled two session handles, one of which referenced the original tree and the other was specific to the lookaside table.  Now that the code has been simplified to just use one session, we need to take care when we need to operate on the original tree.


agorrod:
Thanks @michaelcahill lgtm

PR #4239, opened 2018-08-22
WT-4246 Transaction changes to support indirect references for prepared transactions.

bvpvamsikrishna:
This PR is to make changes to transaction structure such that during commit/rollback, it can access the update structure through indirect references ( indirect references are needed as prepared updates can be stored to lookaside), hence commit/rollback might be referring to an update structure stored in lookaside.

Each TXN_OP structure will store the key of the operation as indirect reference, which can be used to retrieve the update structure.
Operation key is not available in TXN_OP or WT_UPDATE structure to use during prepare time, hence coping the key to TXN_OP as part of the operation itself.
Key is needed only in case of single record level operation (not for truncate).
Changes made to TXN_OP so that sizeof it will not increase.

New types are added to facilitate the release of key in __wt_txn_op_free. ( In case of column store, the key will not be allocated)


src/include/txn.h
```diff
@@ -177,13 +177,26 @@ struct __wt_txn_op {
 		WT_TXN_OP_NONE,
 		WT_TXN_OP_BASIC,
 		WT_TXN_OP_INMEM,
+		WT_TXN_OP_BASIC_ROW,
+		WT_TXN_OP_INMEM_ROW,
```
agorrod: These new types are necessary to identify the difference between a row store update and a column store update (since they are stored in a union) when deciding whether to free the key.

src/include/txn.h
```diff
@@ -177,13 +177,26 @@ struct __wt_txn_op {
...
+		 * WT_TXN_OP_INMEM, WT_TXN_OP_INMEM_ROW
+		 */
+		struct {
+			WT_UPDATE *upd;
+			union {
+				/* WT_TXN_OP_BASIC_ROW, WT_TXN_OP_INMEM_ROW */
+				WT_ITEM row_key;
+				/* WT_TXN_OP_BASIC, WT_TXN_OP_INMEM */
+				uint64_t recno;
+			} key; /* Key to search prepared update. */
```
agorrod: Each time a new TXN_OP is created the key needs to be copied in, so the code will save both the key and update structure. That's done because there is no obvious way to get the from the TXN_OP or WT_UPDATE structure at prepare time.

bvpvamsikrishna:
autobuild: test this please.

bvpvamsikrishna:
@agorrod I have updated the initial description with your comments, so that commit will not mask them.

bvpvamsikrishna:
@sulabhM request your review of this.

src/include/txn.h
```diff
@@ -176,14 +176,27 @@ struct __wt_txn_op {
 	enum {
```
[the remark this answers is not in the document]
bvpvamsikrishna: @agorrod, I will check this, but file id using also seems to be feasible, as recovery is using the file id only.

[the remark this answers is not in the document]
michaelcahill: @bvpvamsikrishna you're correct that recovery works this way (because btree IDs are what we write into the log), but that depends on recovery maintaining a mapping from btree IDs to cursors, which doesn't exist at runtime.

bvpvamsikrishna:
autobuild:retest this please

bvpvamsikrishna:
I have changed fileid of txn op structure to data handle. currently I am incrementing sessions_inuse variable of data handle for reference count. 
Not sure, whether need to add a seperate variable in data handle for transaction op reference count  or using session_inuse is fine.

src/include/txn.i
```diff
@@ -297,10 +302,40 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
+		 * be prepared.
+		 */
+		if (!WT_SESSION_IS_CHECKPOINT(session) &&
+		    ((WT_BTREE *)op->dhandle->handle)->id !=
+		    S2C(session)->cache->las_fileid) {
+			/*
+			 * Store the key, to search the update incase of
+			 * prepared transaction.
+			 */
+			if (btree->type == BTREE_ROW) {
```
agorrod: I'm sad that we need code that knows the difference between row and column stores here, but I don't see an obvious alternative.
@keithbostic Could you take a look and let us know if you could see an alternative?

src/include/txn.i
```diff
@@ -297,10 +302,40 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
+				WT_CLEAR(op->u.single_op.key.row_key);
+				item = &op->u.single_op.key.row_key;
+				WT_RET(__wt_cursor_get_raw_key(
+				    &cbt->iface, item));
+				WT_RET(__wt_buf_set(session,
+				    item, item->data, item->size));
```
agorrod: I can see that other places where __wt_cursor_get_raw_key is used are also calling __wt_buf_set, but I don't see why - do you know if it might be doing something to the copied memory?

src/include/dhandle.i
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * See the file LICENSE for redistribution information.
+ */
+
+/*
+ * __wt_dhandle_incr_use --
+ *	Increment the in-use counter in the data source.
+ */
+static inline void
+__wt_dhandle_incr_use(WT_DATA_HANDLE *dhandle)
```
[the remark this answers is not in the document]
agorrod: @bvpvamsikrishna you are right - we need to increment `session_inuse` not `session_ref` - since `session_ref` doesn't stop sweep from closing a handle. Could you remove `__wt_cursor_dhandle_incr_use` and call this function instead? I don't like having that functionality duplicated.

bvpvamsikrishna:
@agorrod  I tried to use session_inuse variable instead of session_ref variable of dhandle, to increase the reference count, just based on the variable name, as session_inuse seems to be more relevant to usage in this PR. Happy to change to session_ref if you feel it is more appropriate.

src/include/txn.h
```diff
@@ -172,18 +172,31 @@ typedef enum __wt_txn_isolation {
  *	records during commit or undo the operations during rollback.
  */
 struct __wt_txn_op {
-	uint32_t fileid;
+	WT_DATA_HANDLE *dhandle;
 	enum {
 		WT_TXN_OP_NONE,
 		WT_TXN_OP_BASIC,
```
sulabhM: Is WT_TXN_OP_BASIC for column store that we have a separate one for row store. Probably we can rename it to WT_TXN_OP_BASIC_COL. Similarly for WT_TXN_OP_INMEM to WT_TXN_OP_INMEM_COL.

src/include/txn.i
```diff
@@ -297,10 +302,39 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
+		 * be prepared.
+		 */
+		if (!WT_SESSION_IS_CHECKPOINT(session) &&
+		    !F_ISSET(btree, WT_BTREE_LOOKASIDE)) {
+			/*
+			 * Store the key, to search the update incase of
+			 * prepared transaction.
+			 */
+			if (btree->type == BTREE_ROW) {
+				WT_CLEAR(op->u.single_op.key.row_key);
```
sulabhM: Can this be changed to:
```
item = &op->u.single_op.key.row_key;
WT_CLEAR(*item);
```

src/include/txn.i
sulabhM: I think __wt_buf_set() is needed because when getting raw key, __wt_cursor_get_keyv() sets WT_ITEM data and size to refer to the cursors. __wt_buf_set() takes care of copying the external memory into the WT_ITEM's memory. If we do not do so, the cursor might move in the meanwhile and we end up referring to some other unintended key.

src/include/txn.h
bvpvamsikrishna: WT_TXN_OP_BASIC represents both column store and row store in case of non-timestamp builds. Introduced WT_TXN_OP_BASIC_ROW and WT_TXN_OP_INMEM_ROW just to call __wt_buf_free in __wt_txn_op_free()

src/include/txn.i
keithbostic: We could also skip metadata updates here as well, if that matters, since they aren't prepared either.

bvpvamsikrishna:

src/txn/txn.c
```diff
@@ -1079,21 +1085,23 @@ __wt_txn_prepare(WT_SESSION_IMPL *session, const char *cfg[])
 	/* Prepare updates. */
 	for (i = 0, op = txn->mod; i < txn->mod_count; i++, op++) {
 		/* Assert it's not an update to the lookaside file. */
-		WT_ASSERT(session,
-		    S2C(session)->cache->las_fileid == 0 ||
-		    op->fileid != S2C(session)->cache->las_fileid);
+		WT_ASSERT(session, S2C(session)->cache->las_fileid == 0 ||
+		    ((WT_BTREE *)op->dhandle->handle)->id !=
+		    S2C(session)->cache->las_fileid);
```
agorrod: I think this reads better as:
```
!F_ISSET((WT_BTREE *)op->dhandle->handle, WT_BTREE_LOOKASIDE);
```

Similarly below.

If you saved a `WT_BTREE` in the `WT_TXN_OP` structure instead of a `WT_DHANDLE` you could avoid these casts - the btree has a pointer back to the `WT_DHANDLE`.

src/include/txn.h
agorrod: I think this would be clearer if the types were always distinguished - does that make this change a lot more invasive?

bvpvamsikrishna:
@agorrod, Thank you for the changes.

PR #4240, opened 2018-08-22
WT-4248 Fix checkpoints in schema_abort for slow machines.

sueloverso:
@agorrod Please review this PR. It contains changes to `schema_abort` only for the failures on slow machines where the stable timestamp was never set. The former code was taking a checkpoint immediately and therefore wrote out (all) data before the stable timestamp was ever set, that could result in the failures we saw. 

Now this code does not take a checkpoint until it sees the stable timestamp set the first time.


agorrod:
Thanks @sueloverso lgtm.

PR #4241, opened 2018-08-23
WT-4264 Compact can race with page modifications

keithbostic:


agorrod:
Thanks @keithbostic this looks good to me, and I believe you about locking the page, but your commit comment doesn't sound very absolute - should we put this branch through stress testing for a day before merging to gain more confidence?

keithbostic:
> should we put this branch through stress testing for a day before merging to gain more confidence?

I've been running it all day, so I'm pretty confident it's OK, if you agree on the eviction logic.

agorrod:
Sounds good. I've just switched [a stress job](http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress/configure) over, I'll let it run for a few hours then merge if @michaelcahill doesn't see any problems.

michaelcahill:
Thanks @keithbostic -- I think it's a little surprising to remove locking in a change that is addressing a race, but I don't see why locking should be necessary there.  I'll merge...

keithbostic:
> I think it's a little surprising to remove locking in a change that is addressing a race

Yes, sorry, it should have been a separate ticket.

PR #4242, opened 2018-08-23
WT-4216 Use separate counters for page_swap yield and sleep.

sueloverso:
@keithbostic Please review this. It slightly modifies the change you made in WT-4208. Basically I separate out the two different callers to `wt_spin_backoff` so they do not interfere with each other. This change recovers a bit more than half the performance loss.


sueloverso:
Interestingly enough, if I reset the `swap` counters just outside the `for (;;) {` loop near line 424 I get slightly worse and less consistent performance. So I went with initializing them at the top.

PR #4245, opened 2018-08-26
WT-4267 fixed-length column store operations can corrupt data

keithbostic:


agorrod:
Thanks @keithbostic this change looks correct to me, but I think it'll be hard to remember why it's needed in the future.

Do you think the check could be moved inside `__cursor_page_pinned`? Maybe passing in a parameter to identify callers who care about "on a page", vs "at a specific record" (I think that's what matters) to figure out how fixed-length column stores should return.

keithbostic:
> Do you think the check could be moved inside `__cursor_page_pinned`?

Search & search-near use `__cursor_page_pinned`, and `BTREE_COL_FIX` isn't excluded from those uses.

Search-near also excludes `BTREE_ROW` from the page-pinned check and does that check inline instead of inside `__cursor_page_pinned`.

In short, I think that's messier, but open to suggestions?

We could certainly add comments.

agorrod:
Please add comments - I'd hoped to pull the logic into a single place, but agree that even though `__cursor_page_pinned` is tempting it's not quite right.

keithbostic:
@agorrod 
> Please add comments

Pushed in 62043f8.

lukech:
test this please

PR #4247, opened 2018-08-27
WT-4154 Surface the oldest read timestamp.

nehakhatri5:
Added a new option `oldest_reader` to the `query_timestamp` API. When `query_timestamp` is used with this option, WT will return the minimum of the read timestamps of all the active transactions which includes the checkpoint transaction. If there are no active readers at the time of issuing this API, WT will return WT_NOTFOUND.
@bvpvamsikrishna please review.

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -202,7 +202,7 @@ __txn_get_pinned_timestamp(
    bool include_oldest)
 {
 	WT_CONNECTION_IMPL *conn;
-	WT_DECL_TIMESTAMP(tmp_ts);
+	WT_DECL_TIMESTAMP(tmp_ts)
 	WT_TXN *txn;
 	WT_TXN_GLOBAL *txn_global;
 
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -194,11 +194,12 @@ __wt_txn_parse_timestamp(WT_SESSION_IMPL *session, const char *name,
 
 /*
  * __txn_get_pinned_timestamp --
- *	Calculate the current pinned timestamp.
+ * 	Calculate the current pinned timestamp.
  */
 static int
 __txn_get_pinned_timestamp(
-   WT_SESSION_IMPL *session, wt_timestamp_t *tsp, bool include_checkpoint)
+   WT_SESSION_IMPL *session, wt_timestamp_t *tsp, bool include_checkpoint,
+   bool include_oldest)
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_TXN *txn;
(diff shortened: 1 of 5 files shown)


src/txn/txn_timestamp.c
```diff
@@ -198,24 +198,29 @@ __wt_txn_parse_timestamp(WT_SESSION_IMPL *session, const char *name,
...
+   WT_SESSION_IMPL *session, wt_timestamp_t *tsp, bool include_checkpoint,
+   bool include_oldest)
 {
 	WT_CONNECTION_IMPL *conn;
 	WT_TXN *txn;
 	WT_TXN_GLOBAL *txn_global;
 
 	conn = S2C(session);
 	txn_global = &conn->txn_global;
+	__wt_timestamp_set_zero(tsp);
```
sueloverso: This has the side effect that if the caller gets back `WT_NOTFOUND` that the caller's timestamp has been zeroed. That is a new side effect that it didn't do before. Perhaps move it as an `else` clause at about line 216 below instead.

nehakhatri5: Sure, done.

src/txn/txn_timestamp.c
```diff
@@ -198,24 +198,29 @@ __wt_txn_parse_timestamp(WT_SESSION_IMPL *session, const char *name,
...
 	__wt_readlock(session, &txn_global->rwlock);
-	__wt_timestamp_set(tsp, &txn_global->oldest_timestamp);
+	if (include_oldest)
+		__wt_timestamp_set(tsp, &txn_global->oldest_timestamp);
 
 	/* Check for a running checkpoint */
 	if (include_checkpoint &&
 	    !__wt_timestamp_iszero(&txn_global->checkpoint_timestamp) &&
-	    __wt_timestamp_cmp(&txn_global->checkpoint_timestamp, tsp) < 0)
+	    (!include_oldest ||
```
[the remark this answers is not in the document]
bvpvamsikrishna: I think better not to use include_oldest in the checkpoint related block, it will make coupling between whether to consider oldest timestamp or checkpoint timestamp. (i.e. instead of include_oldest, might be __wt_timestamp_iszero(tsp) will suffice.

src/txn/txn_timestamp.c
```diff
@@ -207,15 +208,20 @@ __txn_get_pinned_timestamp(
...
-	if (!txn_global->has_oldest_timestamp)
+	if (include_oldest && !txn_global->has_oldest_timestamp)
 		return (WT_NOTFOUND);
+
 	__wt_readlock(session, &txn_global->rwlock);
-	__wt_timestamp_set(tsp, &txn_global->oldest_timestamp);
+	if (include_oldest)
+		__wt_timestamp_set(tsp, &txn_global->oldest_timestamp);
+	else
+		__wt_timestamp_set_zero(tsp);
```
bvpvamsikrishna: I think better to use local variable for timestamp so that we will not modify the user passed parameter incase of returning WT_NOTFOUND.

PR #4248, opened 2018-08-27
WT-4185 Don't remove all lookaside entries when reading a page.

nehakhatri5:
The lookaside sweep code removes entries in the background so stop removing lookaside entries when reading pages from lookaside tree.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -519,17 +519,6 @@ skip_read:
 		break;
 	}
 
-	/*
-	 * We no longer need lookaside entries once the page is instantiated.
-	 * There's no reason for the lookaside remove to fail, but ignore it
-	 * if for some reason it fails, we've got a valid page.
-	 *
-	 * Don't free WT_REF.page_las, there may be concurrent readers.
-	 */
-	if (final_state == WT_REF_MEM && ref->page_las != NULL)
-		WT_IGNORE_RET(__wt_las_remove_block(
-		    session, ref->page_las->las_pageid, false));
-
 	WT_PUBLISH(ref->state, final_state);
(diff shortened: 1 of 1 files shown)


agorrod:
@nehakhatri5 please run some Evergreen correctness and perf tests on this branch

agorrod:
This change is potentially destabilising, am going to hold off moving forward with it for now.

michaelcahill:
@sulabhM, how does WT-4218 currently prevent prepared updates in lookaside from being removed by sweep?

sulabhM:
@michaelcahill , there was no change as part of WT-4218 to prevent prepared updates in lookaside from being removed, that change already exists here in __wt_las_sweep():

```
                /*                                                               
                 * If this entry isn't globally visible or if it is a part of a  
                 * prepared transaction, then we cannot remove it. Otherwise,    
                 * perform additional checks to see whether it has aged out of a 
                 * live file.                                                    
                 */                                                              
                if (!__wt_txn_visible_all(session, las_txnid, val_ts) ||         
                    prepare_state == WT_PREPARE_INPROGRESS) {                    
                        saved_key->size = 0;                                     
                        continue;                                                
                }   
```  

While working on WT-4218 we had a discussion about whether we need to do something special for prepared transactions so that sweep will clear them out. The conclusion from there was that it is okay for sweep to ignore these entries because commit/rollback will read them back in, and the read will remove them from lookaside.

michaelcahill:
Thanks @sulabhM.

@agorrod, is there some testing we could do to unblock this change?

agorrod:
I'm good to go with this now that we've made a drop to master, I'll go ahead and merge.

PR #4249, opened 2018-08-27
WT-4268 Random abort should wait until record files exist before starting timer.

sueloverso:
Waiting for the table to exist was insufficient as that was created before the threads were started.
Instead wait for the record files that each thread creates for itself.


sueloverso:
autobuild: retest this please

agorrod:
lgtm

PR #4250, opened 2018-08-27
WT-4270 Add an operation field to know where threads hang.

sueloverso:
@agorrod This is a branch to add debugging for schema_abort when threads hang. This will give us some more information if it fails on a sanitizer build. We still need a core dump to know exactly where they're all hanging. But this is a step forward.


agorrod:
lgtm

PR #4251, opened 2018-08-28
WT-4218 Evict prepared updates to lookaside, gated behind ifdef

sulabhM:


agorrod:
autobuild: retest this please

agorrod:
autobuild: retest this please

sulabhM:
autobuild: retest this please

sulabhM:
autobuild: retest this please

sulabhM:
Pull request passes testing, but isn't properly getting reflected on github yet. The review need not wait.

agorrod:
Thanks @sulabhM lgtm

src/reconcile/rec_write.c
```diff
@@ -1342,14 +1342,35 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
-		if (F_ISSET(r, WT_REC_EVICT) &&
-		    (upd->prepare_state == WT_PREPARE_LOCKED ||
-		    upd->prepare_state == WT_PREPARE_INPROGRESS ||
-		    (F_ISSET(r, WT_REC_VISIBLE_ALL) ?
-		    WT_TXNID_LE(r->last_running, txnid) :
-		    !__txn_visible_id(session, txnid)))) {
-			uncommitted = r->update_uncommitted = true;
-			continue;
+		if (F_ISSET(r, WT_REC_EVICT)) {
+		       if (upd->prepare_state == WT_PREPARE_LOCKED ||
```
bvpvamsikrishna: prepare check is not required in the final code. This can be part of temporary macro below.

src/reconcile/rec_write.c
```diff
@@ -1342,14 +1342,35 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+#define	DISABLE_LOOKASIDE_FOR_PREPARED
+#ifdef DISABLE_LOOKASIDE_FOR_PREPARED
+		       if (prepared) {
+			       prepared = false;
+			       uncommitted = r->update_uncommitted = true;
+		       }
+#endif
+#undef DISABLE_LOOKASIDE_FOR_PREPARED
+
+		       if (prepared || uncommitted)
```
bvpvamsikrishna: This prepared condition also can be inside the above macro.

test/suite/test_prepare_lookaside01.py
```diff
@@ -40,33 +40,66 @@ class test_prepare_lookaside01(wttest.WiredTigerTestCase):
...
-        # If we increase nsessions below, then hang will occur.
-        nsessions = 1
+        # Do another set of updates in that many transactions. This forces the
+        # pages that have been evicted to lookaside to be re-read and brought in
+        # memory. Hence testing if we can read prepared updates from lookaside.
+
+        # Start with setting a stable timestamp to pin history in cache
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(1))
+
+        # Commit some updates to get eviction and lookaside fired up
```
bvpvamsikrishna: I am not clear as to why we need committed updates for evicting prepared updates.
my understanding is prepared updates should be treated same as committed updates from eviction point of view. committed updates are evicted with out any prerequisites, similarly prepared updates should be evicted with out any prerequisites, committed updates should not be prerequisite for prepared updates eviction.

lukech:
test this please

src/reconcile/rec_write.c
sulabhM: As discussed offline, with the understanding that we are not necessarily avoiding a stall if cache is full of prepared updates, please review if this comment still holds good.

src/reconcile/rec_write.c
sulabhM: As discussed offline, with the understanding that we are not necessarily avoiding a stall if cache is full of prepared updates, please review if this comment still holds good

test/suite/test_prepare_lookaside01.py
sulabhM: As discussed offline, with the understanding that we are not necessarily avoiding a stall if cache is full of prepared updates, please review if this comment still holds good

src/reconcile/rec_write.c
```diff
@@ -1342,14 +1342,35 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
...
+			       continue;
```
bvpvamsikrishna: Is it not required to reset the prepared to "false" before continue ?

src/reconcile/rec_write.c
```diff
@@ -1395,7 +1416,7 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 			 * discard an uncommitted update.
 			 */
 			if (F_ISSET(r, WT_REC_UPDATE_RESTORE) &&
-			    *updp != NULL && uncommitted) {
+			    *updp != NULL && (uncommitted || prepared)) {
```
bvpvamsikrishna: Will the evaluation of (uncommitted || prepared) can ever be true ? as we continue in the loop at line 1373, if it is "true".

test/suite/test_prepare_lookaside01.py
bvpvamsikrishna: If we understand how it is working is better, instead of ignoring it.

src/reconcile/rec_write.c
sulabhM: No, if we find an uncommitted or prepared update, we continue to not process the rest of the for loop because other than exceptions we do not see a stable update after an uncommitted and hence we wont find a suitable candidate for eviction.

src/reconcile/rec_write.c
sulabhM: It can be true, we discussed this offline

test/suite/test_prepare_lookaside01.py
sulabhM: We discussed this one earlier.
I think what happens is eviction notices the committed updates which it tries to evict. It is able to evict one of them, but unable to populate the eviction queue any further because the rest of them are pinned in memory (these updates being newer than the stable timestamp). This inability to populate the eviction queue repeatedly raises the lookaside score beyond a certain value (I believe 80) which sets the cache in a mode that says, it-is-okay-to-try-lookaside-eviction, fkag "WT_CACHE_EVICT_LOOKASIDE". We need the cache in this state otherwise lookaside is not attempted even though the cache might be full with prepared updates.
To conclude, we need at-least some amount of regular eviction while holding updates pinned to raise lookaside score to be able to lookaside evict the prepared updates.

PR #4254, opened 2018-08-28
WT-4274 Free buffer.

sueloverso:


sueloverso:
self-reviewing this one line fix.

PR #4255, opened 2018-08-28
WT-4225 Add test to simulate random system crashes using direct IO

ddanderson:


ddanderson:
This is not ready for review, but I may be filing tickets based on some failures I've seen so far, so it seemed like a good time to push it.

agorrod:
autobuild: retest this please

ddanderson:
retest this please

ddanderson:
@agorrod, could you please review this?

agorrod:
Thanks @ddanderson functionally the test looks really good - thanks for putting it together.

I'm going to ask Neha to do a code review as well - I only really skimmed the code for style oddities.

test/csuite/random_directio/smoke.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/sh
+
+set -e
+
+# Smoke-test random-abort as part of running "make check".
+
+echo TEST_WRAPPER = $TEST_WRAPPER
+$TEST_WRAPPER ./test_random_directio  -t 5 -T 5
```
sueloverso: Given all the differently options the program supports, I would like to see `smoke.sh` expanded greatly to run as many of them as is practical. Even if they're turning up bugs, having them in there, but commented out (like `timestamp_abort/smoke.sh` comments out the LSM variants because LSM does not work with timestamps for now).

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+ * function).  When forming the key, the large string is truncated so
+ * that the key is effectively padded to the right length.
+ *
+ * The key space for the main table is designed to be interleaved tightly
+ * among all the threads.  The matching values in the main table are the
+ * same, except with the xxxx string reversed.  So the keys and values
+ * are the same size.
+ *
+ * There is also a reverse table where the keys/values are swapped.
+
```
sueloverso: Extra blank line.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+/* 64 spaces */
+#define	SPACES								\
+	"                                                                "
+
+/*
+ * Set higher to be less stressful for schema operations.
+ */
+#define	SCHEMA_OP_FREQUENCY	100
+#define	SCHEMA_OP(id, offset)						\
```
sueloverso: A small comment on this calculation would be helpful.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+	    "  -m method      \t"
+	    "sync method: fsync, dsync, none [none]\n"
+	    "  -n num_cycles  \t"
+	    "number of copy/recover cycles [5]\n"
+	    "  -p             \t"
+	    "populate only [false]\n"
+	    "  -S             \t"
+	    "schema operations on [false]\n"
+	    "  -T num_threads \t"
+	    "number of threads in writer[ 4]\n"
```
sueloverso: `[ 4]` should be ` [4]` (swap space and open bracket).

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+	kvsize = td->datasize / 2;
+	buf1 = td->data;
+	buf2 = &td->data[kvsize];
+
+	/*
+	 * Continuing writing until we're killed.
+	 */
+	printf("Thread %" PRIu32 "\n", td->id);
+	for (i = 0; ; ++i) {
+	again:
```
sueloverso: Label at left margin.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+		rev->set_value(rev, buf1);
+		testutil_check(rev->insert(rev));
+
+		/*
+		 * If we are doing a schema test, generate operations
+		 * for additional tables.  Each table has a 'lifetime'
+		 * of 4 values of the id.
+		 */
+		if (td->schema_test) {
+			if (!retryable_error && SCHEMA_OP(i, 0)) {
```
sueloverso: Every conditional within this is `!retryable_error` so that can be put on line 354 as `if (td->schema_test && !retryable_error) {`. Can the `SCHEMA_OP` be modified so that it can something like `switch ((op = SCHEMA_OP(i))` with cases 0-3 instead of sequential conditionals? Or if that is really what we want, please add a comment describing the testing strategy here.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+				/*
+				 * XXX
+				 * We notice occasional EBUSY errors from
+				 * rename, even though neither URI should be
+				 * used by any other thread.
+				 */
+				/*
+				printf(" rename(\"%s\", \"%s\")\n", buf1, buf2);
+				*/
+				if ((ret = session->rename(session, buf1, buf2,
```
sueloverso: Just FYI, until WT-2329 is fixed there is a definitely possibility of files/metadata out of sync with `rename`. A crash at the wrong time of rename can leave things orphaned.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+			}
+			if (!retryable_error && SCHEMA_OP(i, 3)) {
+				/* Update the single value in the table. */
+				gen_table_name(buf1, kvsize, i - 3, td->id);
+				gen_table2_name(buf2, kvsize, i - 3, td->id);
+				testutil_check(session->open_cursor(session,
+				    buf2, NULL, NULL, &sch));
+				sch->set_key(sch, buf1);
+				sch->set_value(sch, buf2);
+				testutil_check(sch->insert(sch));
```
sueloverso: Should this be `sch->update` instead of `sch->insert`? Or at least that's what the comment at line 401 implied to me.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+					    " retrying transaction\n", buf1);
+					retryable_error = true;
+				} else {
+					if (ret != 0)
+						printf("FAIL: "
+						    "drop(\"%s\") "
+						    "returns %d: %s\n",
+						    buf1, ret,
+						    wiredtiger_strerror(ret));
+					testutil_check(ret);
```
sueloverso: This could save a level of indentation with:
```
    ...
    retryable_error = true;
    ret = 0;
} else if (ret != 0)
    printf("FAIL: " ...
testutil_check(ret);
```

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+			printf("FAIL: expected rev file to have: %s\n", key);
+			testutil_assert(!exists);
+		}
+	} else {
+		testutil_check(ret);
+		if (!exists) {
+			printf("FAIL: unexpected key in rev file: %s\n", key);
+			testutil_assert(exists);
+		}
+
```
sueloverso: Remove the blank line.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+	/*
+	 * We're most interested in the final records on disk.
+	 * Rather than walk all records, we do a quick scan
+	 * to find the last complete set of written ids.
+	 * Each thread writes each id, along with the thread id,
+	 * so they are interleaved.  Once we have the neighborhood
+	 * where some keys may be missing, we'll back up to do a scan
+	 * from that point.
+	 */
+
```
sueloverso: Remove blank line.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+		/*
+		 * See if the expected thread has finished at this point.
+		 * If so, remove it from the thread map.
+		 */
+		while (gotth != th) {
+			if ((threadmap & (0x1U << th)) != 0) {
+				threadmap &= ~(0x1U << th);
+				lastid[th] = id - 1;
+				/*
+				 * Any newly removed should not be present
```
sueloverso: Missing a word, `Any new removed <something> should not...`

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+				    "Child process %" PRIu64 " terminated "
+				    " with signal %d", (uint64_t)pid,
+				    WTERMSIG(status));
+		} else if (got == -1)
+			testutil_die(errno, "waitpid");
+
+		--seconds;
+		sleep(1);
+	}
+}
```
sueloverso: This seems more robust than what was done in the other `*_abort` programs. Does it make sense to do this there too? Or maybe make it a testutil function?

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+			break;
+		case 'v':
+			verify_only = true;
+			break;
+		default:
+			usage();
+		}
+	argc -= __wt_optind;
+	if (argc != 0)
+		usage();
```
sueloverso: I recently added a line to `timestamp_abort` that prints out `CONFIG: <args>` for reproducing. That has been very helpful there. (I'd have sworn I also added it to `random_abort` too but it is not in there and should be too. If you add it here, please add it there too.)

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+	 * We make a copy of the directory (possibly using direct IO)
+	 * for recovery and checking, and an identical copy that
+	 * keeps the state of all files before recovery starts.
+	 */
+	testutil_check(__wt_snprintf(buf, sizeof(buf),
+	    "H='%s'; C=$H.CHECK; S=$H.SAVE; rm -rf $C $S;"
+	    " mkdir $C $S; for f in `ls $H/`; do "
+	    " dd if=$H/$f of=$C/$f bs=4096 %s >/dev/null 2>&1 || exit 1; done;"
+	    " cp -pr $C $S",
+	    home, directio ? "iflag=direct" : ""));
```
sueloverso: It would be more helpful if `WT_TEST.random-directio.SAVE` ended up with the files in the top level directory rather than `WT_TEST.random-directio.SAVE/WT_TEST.random-directio.CHECK/<files>`.

test/csuite/random_directio/main.c
nehakhatri5: The execution of the test with default settings, creates random number of writer threads. So why are we saying that default number of writer threads is 4 in the usage?

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+	if (!has_direct_io()) {
+		fprintf(stderr, "**** test_random_directio: this system does "
+		    "not support direct I/O.\n**** Skipping test.\n");
+		return (EXIT_SUCCESS);
+	}
+	while ((ch = __wt_getopt(progname, argc, argv,
+	    "dh:i:m:n:pST:t:v")) != EOF)
+		switch (ch) {
+		case 'd':
+			datasize = (uint32_t)atoi(__wt_optarg);
```
nehakhatri5: Providing an "-d 50" input argument causes a segmentation fault. I saw that __wt_optarg was NULL.

test/csuite/random_directio/main.c
michaelcahill: So is the plan to change this test to avoid exercising rename until WT-2329 is fixed?

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1127 @@
...
+	 * for recovery and checking, and an identical copy that
+	 * keeps the state of all files before recovery starts.
+	 */
+	testutil_check(__wt_snprintf(buf, sizeof(buf),
+	    "H='%s'; C=$H.CHECK; S=$H.SAVE; rm -rf $C $S;"
+	    " mkdir $C $S; for f in `ls $H/`; do "
+	    " dd if=$H/$f of=$C/$f bs=4096 %s >/dev/null 2>&1 || exit 1; done;"
+	    " cp -pr $C $S",
+	    home, directio ? "iflag=direct" : ""));
+	printf("Shell command: %s\n", buf);
```
nehakhatri5: It useful to print a verb about the work being done here.

test/csuite/random_directio/main.c
ddanderson: I've modified the test to optionally do rename, and actually to control more what gets done and what gets checked. I find that we can do both create and drop and reliably finish recovery in all situations (methods set to {dsync,fsync,none}) but we can't expect to see the created or dropped tables to appear or disappear at the right moment.  So those options with be in smoke.sh, with the various things we can't yet do commented out.

test/csuite/random_directio/main.c
ddanderson: Right!  Set defaults for -T and -t to [random] in the usage text, to match the code.

test/csuite/random_directio/main.c
ddanderson: @sueloverso, I've reworked the code and commented, hopefully the next round will be clearer.

test/csuite/random_directio/main.c
ddanderson: I'm resolving this here, and opening a new ticket for some followup work on other tests: https://jira.mongodb.org/browse/WT-4319

test/csuite/random_directio/main.c
ddanderson: I'm resolving this here, and opening a new ticket for some followup work on other tests: https://jira.mongodb.org/browse/WT-4319

ddanderson:
> @ddanderson since I've been reading the test for WT-4284, I added myself as a reviewer. A general comment: this test seems more comprehensive for the schema validation than `schema_abort` which does not really keep track of what should be there. It only makes sure that the crashed database can recover without error. It never has turned up a problem, nor did it reproduce the original problem we were trying to repro. It does use timestamps too though. Should this test replace that one?

@sueloverso, I've opened a ticket https://jira.mongodb.org/browse/WT-4320 to look at this possibility.

ddanderson:
@sueloverso and @nehakhatri5, thanks for your reviews and your patience!  I've resolved all your comments, I think. Please take another look.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1264 @@
...
+		 * Check metadata to see if there are any tables
+		 * present that shouldn't be there.
+		 */
+		testutil_check(session->open_cursor(session, "metadata:", NULL,
+		    NULL, &meta));
+		while ((ret = meta->next(meta)) != WT_NOTFOUND) {
+			testutil_check(ret);
+			meta->get_key(meta, &gotkey);
+			/*
+			 * Names involved in schema testing are off the form:
```
sueloverso: off -> of

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1264 @@
...
+			 * xxx corresponds to the id inserted into the main
+			 * table when the table was created, and t corresponds
+			 * to the thread id that did this.
+			 */
+			if (WT_PREFIX_SKIP(gotkey, "table:") &&
+			    (*gotkey == 'A' || *gotkey == 'B')) {
+				gotid = (uint64_t)strtol(gotkey + 1, &p, 10);
+				testutil_assert(*p == '-');
+				th = (uint32_t)strtol(p + 1, &p, 10);
+				testutil_assert(*p == '\0');
```
sueloverso: Is there any checking we can or want to do on `gotid` or `th`? If not, maybe name them `dummyid, dummyth` or `unusedid, unusedth`

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1264 @@
...
+					LF_SET(SCHEMA_DATA_CHECK);
+				else if (WT_STREQ(arg, "drop"))
+					LF_SET(SCHEMA_DROP);
+				else if (WT_STREQ(arg, "drop_check"))
+					LF_SET(SCHEMA_DROP_CHECK);
+				else if (WT_STREQ(arg, "rename"))
+					LF_SET(SCHEMA_RENAME);
+				else if (WT_STREQ(arg, "verbose"))
+					LF_SET(SCHEMA_VERBOSE);
+				else if (WT_STREQ(arg, "none"))
```
sueloverso: I think we want the `none` case alphabetized to its normal place.

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1264 @@
...
+			usage();
+		}
+	argc -= __wt_optind;
+	if (argc != 0)
+		usage();
+
+	testutil_work_dir_from_path(home, sizeof(home), working_dir);
+	/*
+	 * If the user wants to verify they need to tell us how many threads
+	 * there were so we can find the old record files.
```
sueloverso: This comment is similar to, or from `random_abort` is "old record files" the right phrasing here?

test/csuite/random_directio/main.c
```diff
@@ -0,0 +1,1264 @@
...
+#endif
+}
+
+/*
+ * main --
+ *	Top level test.
+ */
+int
+main(int argc, char *argv[])
+{
```
sueloverso: This is a large, comprehensive and complex test program. Do you want to consider breaking the code up into a few files for easier understanding? child.c, verify.c, main.c or others that seem suitable.

test/csuite/random_directio/main.c
ddanderson: Fixed, added check.

test/csuite/random_directio/main.c
ddanderson: @sueloverso, I think I'd do this as a followup.

PR #4256, opened 2018-08-28
WT-4261 Add out of sync metadata/turtle files to salvage test.

sueloverso:
@agorrod The changes for this test to test out of sync are in and working. Please review. It is a lot of directory and file copying, so it isn't all that pretty. But we are testing corruption and out of sync now.


sueloverso:
@agorrod hold off on review because it isn't detecting corruption like it should.

sueloverso:
autobuild: retest this please

agorrod:
autobuild: retest this please

agorrod:
Thanks @sueloverso this test case looks good to me. I'm going to let you own merging this in case you want to get a better understanding of the OS X behavior.

sueloverso:
autobuild: retest this please

sueloverso:
@agorrod and @michaelcahill this is ready for a new review. @michaelcahill please make sure the out-of-sync cases capture your intended cases. It works in both directions of metadata/turtle out of sync and salvageable when they're out of sync by 1 checkpoint. If I increase the distance to 2 then many of the cases do not pass. There's also a bunch of refactoring and cleanup in this change.

sueloverso:
autobuild: retest this please

michaelcahill:
Thanks for adding this testing @sueloverso.  I tried it out with `CKPT_DISTANCE == 2` and understand why that case is failing (opening a cursor on the ordinary table fails because the metadata references a checkpoint that doesn't exist yet).  This shouldn't be a problem for MongoDB: when running `mongod --repair`, it will attempt to verify each table and salvage any for which verification fails.

I'll merge -- thanks again.

PR #4257, opened 2018-08-28
WT-4277 Make truncate in column stores more efficient

keithbostic:


agorrod:
Thanks @keithbostic this fix makes sense to me - I think it's simpler to follow if it's setup prior to looking at the existing insert list anyway.

I'm going to go ahead and merge.

PR #4259, opened 2018-08-29
WT-4282 Restructure __page_read and avoid moving from limbo to lookaside

agorrod:
Sometimes it may be possible to avoid the transition.

It's likely the check in __wt_page_in avoids this transition - but it's not clear to me that the code always goes through that path - so I think we might as well be explicit in __page_read as well.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -369,6 +369,43 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_REF *ref)
 }
 
 /*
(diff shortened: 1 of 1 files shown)


agorrod:
@michaelcahill please take a look at this and let me know what you think - it means we'll call `__wt_las_page_skip_locked` a bit more often - but I find the code easier to understand.

michaelcahill:
Thanks @agorrod, this is an important bug fix and I like the refactoring.  I'll merge.

PR #4265, opened 2018-08-30
WT-4177 Force a log switch when opening a backup cursor.

dgottlieb:


src/cursor/cur_backup.c
```diff
@@ -181,6 +181,8 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
 	if (conn->log) {
 		WT_ERR(__wt_log_get_all_files(
 		    session, &logfiles, &logcount, &cb->maxid, active));
+		F_SET(conn->log, WT_LOG_FORCE_NEWFILE);
```
dgottlieb: I'm thinking I should clear this flag on error, but wasn't sure if there were considerations. I imagine this thread is already holding a lock that prevents concurrent access to the `conn->log` object such that it's not possible to accidentally clear a different thread's setting of the flag.

src/cursor/cur_backup.c
```diff
@@ -181,6 +181,8 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
...
+		WT_ERR(__wt_log_force_write(session, true, NULL));
```
dgottlieb: This method takes a retry boolean, returns an integer (error?) and also has a `did_work` out parameter.

The out-paramter is immediately be set to true, but I do believe this method call can return `0`, and also have `did_work` set to false. I'm not sure if I'm using this method in the appropriate way for what needs to be guaranteed.

dgottlieb:
autobuild: retest this please

test/suite/test_backup09.py
```diff
@@ -0,0 +1,119 @@
...
+        if self.checkpoint:
+            self.session.checkpoint()
+
+        for i in range(10):
+            doc_id += 1
+            cursor[doc_id] = doc_id
+
+        last_doc_in_backup = doc_id
+        self.assertEqual(1, len(filter(lambda x: x.startswith('WiredTigerLog.'),
+                                        os.listdir('.'))))
```
sueloverso: The continuation line should be indented four spaces. Perhaps break the line before `lambda`.

test/suite/test_backup09.py
```diff
@@ -0,0 +1,119 @@
...
+        backup_cursor = self.session.open_cursor('backup:')
+        # Opening a backup cursor will immediately force a new log file.
```
sueloverso: I thought the flag is a lazy file creation when the next log record gets written and we had discussed possibly adding a log printf to the test or just checking after adding more data. What log record is getting written?

test/suite/test_backup09.py
```diff
@@ -0,0 +1,119 @@
...
+class test_backup09(wttest.WiredTigerTestCase):
+    conn_config = 'config_base=false,create,log=(enabled)'
+    uri = 'table:coll1'
+    dir = 'backup.dir'
+
+    types = [
+        ('checkpoint', {'checkpoint': True}),
+        ('no_checkpoint', {'checkpoint': False}),
+    ]
+    scenarios = make_scenarios(types)
```
sueloverso: You might add a set to the scenarios that toggles the type of copying done. The choices would be 'walking the cursor' and 'copying existing files'. In the former, the later data does not exist, but in the latter, it would exist because log 2 would get copied.

sueloverso:
@dgottlieb this looks in good shape once we figure out why `txn02` is failing. The cause is not clear to me.

test/suite/test_backup09.py
dgottlieb: We talked about this and it resulted in add a forced a log write after setting the flag: https://github.com/wiredtiger/wiredtiger/commit/8b423cbcf1b317bf617ff54f787e1d7f3b544bd6#diff-f948a1575eef3a5b686ffed3802c2b5aR185

Before that change, I was seeing:
1. Perform write A
2. Assert 1 log file
3. Open a backup cursor
4. Perform write B
5. Assert 2 log files

Which was failing the test. I had expected writes concurrent with opening a backup cursor may or may not appear in the log files to copy, but not writes causally after the backup cursor.

I now realized though that this feature/test in particular wants to assert that taking a checkpoint between steps 3 and 4 results in a new log file. When I remove the forced log write and add the checkpoint call, the two log files become visible via the OS right after the checkpoint without any additional "flushing" calls.

The only part I don't completely understand is that taking a checkpoint will respect the `WT_LOG_FORCE_NEWFILE` flag, but a "typical write" does not. Does it have something to do with log slots and which operations can dispatch an active, not full slot?

My next patch will remove the forced log write and add a checkpoint call in the test.

dgottlieb: FWIW, removing the line that forces the log write fixed some scenarios in the txn test (but not all).

sueloverso: Ah! Right! Forcing out earlier writes and forcing a log slot switch causes the flag to be noticed when setting up the next, new slot. You cannot remove the call to {{wt_log_force_write}}. Its purpose is to force out earlier records in the slot's log buffer memory and write(2) them out to the log file. It is NOT to force some new record to be written into the log.

You must force out those records before returning the backup cursor so that all existing records in the log are written to the OS before the backup cursor returns that filename to be copied.

dgottlieb: Are you saying that, today, opening up a backup cursor does not force out earlier records?

sueloverso: AFAICT, that is correct. That is a bug, just realized now, and fixed with this.

test/suite/test_txn02.py
```diff
@@ -201,7 +208,7 @@ def check_log(self, committed):
     def test_ops(self):
         self.backup_dir = os.path.join(self.home, "WT_BACKUP")
         self.session2 = self.conn.open_session()
-        # print "Creating %s with config '%s'" % (self.uri, self.create_params)
```
dgottlieb: I found these print statements useful, but turning them on, and off, and on again to keep commits clean annoying. I hope this is an agreeable, localized, alternative.

test/suite/test_backup09.py
```diff
@@ -0,0 +1,151 @@
...
+
+        last_doc_in_backup = doc_id
+        self.assertEqual(1, len(filter(lambda x:
+            x.startswith('WiredTigerLog.'), os.listdir('.'))))
+        backup_cursor = self.session.open_cursor('backup:')
+        # This test asserts that writes after a checkpoint will not be part of
+        # the data/log files returned by a backup cursor that was opened prior
+        # to the checkpoint. Taking a checkpoint will also force a log write to
+        # a new file, due to the lazy-evaluated flag set by the backup cursor.
+        self.session.checkpoint()
```
sueloverso: I think this checkpoint call may not be necessary. By setting the flag before `log_force_write` the logging subsystem will not lazy-evaluate it and will create the new file when it switches log slots.

test/suite/test_backup09.py
```diff
@@ -0,0 +1,151 @@
...
+
+        cursor.close()
+        return backup_cursor, last_doc_in_backup, doc_id
+
+    def copy_and_restore(self, backup_cursor, last_doc_in_backup, last_doc_in_data):
+        log_files_to_copy = 0
+        os.mkdir(self.backup_dir)
+        if self.all_log_files:
+            log_files_copied = 0
+            for file_name in os.listdir('.'):
```
sueloverso: I think you should use `copy_wiredtiger_home` here instead of doing it yourself. (It does the mkdir of dest directory too.) Then you can do an `os.listdir`, filter on the filename and grab its length.

test/suite/test_txn02.py
```diff
@@ -93,6 +93,13 @@ class test_txn02(wttest.WiredTigerTestCase, suite_subprocess):
     checklog_calls = 100 if wttest.islongtest() else 2
     checklog_mod = (len(scenarios) / checklog_calls + 1)
 
+    _debug = False
+    def debug(self, msg):
+        if not self._debug:
+            return
+
```
sueloverso: I'm okay with this function. I'd suggest removing the blank line here so the `print(msg)` isn't just floating on its own.

src/log/log.c
```diff
@@ -469,16 +470,35 @@ __wt_log_get_all_files(WT_SESSION_IMPL *session,
...
 
 	/*
-	 * These may be files needed by backup.  Force the current slot
-	 * to get written to the file.
+	 * Capture the LSN written, before forcing a new log file. This LSN
+	 * represents the latest journal file that needs to be copied. Note the
+	 * checkpoint selected for backup may exceed this value. In that case,
+	 * the journal files are wasted.
 	 */
+	max_lsn = log->write_lsn;
```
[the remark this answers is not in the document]
dgottlieb: All sound good to me, done!

src/log/log.c
```diff
@@ -469,16 +470,35 @@ __wt_log_get_all_files(WT_SESSION_IMPL *session,
...
+	 */
+	min_lsn = log->ckpt_lsn;
+
+	/*
+	 * These may be files needed by backup.  Force the current slot to get
+	 * written to the file. Also switch to using a new log file.  That log
+	 * file will be removed from the list of files returned. New writes will
+	 * not be included in the backup.
+	 */
+	F_SET(log, WT_LOG_FORCE_NEWFILE);
```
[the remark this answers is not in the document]
dgottlieb: Added. I didn't perform the rename as the documentation for the method refers to the active files and I don't think I'm familiar enough with "cold backups" to know how to rephrase it.

test/suite/test_backup09.py
```diff
@@ -0,0 +1,151 @@
...
+    types = [
+        # checkpoint: whether to explicitly checkpoint some data before opening
+        #   the backup cursor.
+        #
+        # all_log_files: whether to copy all files in the source directory, or
+        #   only the files returned from the backup cursor. Copying the
+        #   additional log files will result in more operations being recovered.
+        ('checkpoint', {'checkpoint': True, 'all_log_files': False}),
+        ('no_checkpoint', {'checkpoint': False, 'all_log_files': False}),
+        ('all_log_files', {'checkpoint': True, 'all_log_files': True}),
```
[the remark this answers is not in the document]
dgottlieb: Done. Curious if there's an advantage to doing it this way that I'm not seeing?

test/suite/test_backup09.py
dgottlieb: I had left the checkpoint in because the contract we need to provide is weaker (providing a relationship between backup cursors and writes only in the face of checkpoints). Though that contract is subtle and probably confusing to others compared to simply "backup cursors change the log file".

So I took this out, and one of the tests started failing. Back to debugging :)

test/suite/test_backup09.py
dgottlieb: Done

dgottlieb:
Force pushed a patch addressing @sueloverso's comments and removed the modification to `test/csuite/test_wt4156_metadata_salvage`. This patch also adds `test/csuite/test_wt4156_metadata_salvage` to the `.gitignore` file.

sueloverso:
LGTM. Thanks @dgottlieb for pushing this through!

PR #4268, opened 2018-08-31
 WT-4243 Fix lookaside sweep to not remove required entries

agorrod:


agorrod:
This isn't yet ready for a real review - up here for a discussion at the moment.

src/include/txn.i
```diff
@@ -300,42 +348,49 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd)
...
+#ifdef HAVE_TIMESTAMPS
+	__wt_txn_update_set_timestamp(session, op, false, NULL);
+        /*
+         * Copy the key into the transaction op structure, so the update
+         * can be evicted to lookaside, and we have a chance of finding it
+         * again. This is only possible for transactions that are in the
+         * prepared state, but we don't know at this stage if a transaction
+         * will be prepared or not.
+         */
+        if (!WT_SESSION_IS_CHECKPOINT(session) &&
```
agorrod: @bvpvamsikrishna this is quite a significant change - I moved some functionality inside `__wt_txn_update_needs_timestamp` (and renamed it). When I was resolving merge conflicts, I got to thinking that this code wasn't working as expected before. I believe it always needs to save a copy of the key here - not only when the transaction already has a timestamp set.
That could explain why we didn't see any performance change with the original merge of WT-4246.

bvpvamsikrishna: You observation is correct, there is some problem here. But, it is other way around for optimisation reason. 
This is required for prepared transactions and an update done by a transaction, which will go to prepared state will not have a timestamp. So, this code to be enabled only when the update doesn't have a timestamp.

agorrod:
@michaelcahill there are still some testing/diagnostic changes in here I need to revert - will let you know when it's ready for review.

agorrod:
@michaelcahill - please review when you are available.

michaelcahill:
Thanks @agorrod, lgtm -- I like the simplification of the sweep code.  I'll merge.

PR #4271, opened 2018-09-04
WT-4217 Enhance commit and rollback to re-read prepared updates

bvpvamsikrishna:
This PR is for review only, not to be merged.
This handles re-reading of prepared updates for commit and rollback.


bvpvamsikrishna:

src/include/txn.i
```diff
@@ -193,6 +193,172 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	case WT_TXN_OP_NONE:
+		break;
+	case WT_TXN_OP_BASIC_COL:
+	case WT_TXN_OP_BASIC_ROW:
+	case WT_TXN_OP_INMEM_COL:
+	case WT_TXN_OP_INMEM_ROW:
+		cursor = NULL;
+		WT_ERR(__wt_open_cursor(session, op->btree->dhandle->name,
+		    NULL, open_cursor_cfg, &cursor));
+		if (op->type == WT_TXN_OP_BASIC_ROW ||
```
bvpvamsikrishna: Suggestion for restructuring are most welcome. I don't like the if condition on "op->type" again with in a switch which is based on "op->type".

src/txn/txn.c
```diff
@@ -1092,10 +1105,16 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
 		case WT_TXN_OP_BASIC_ROW:
 		case WT_TXN_OP_INMEM_COL:
 		case WT_TXN_OP_INMEM_ROW:
-			WT_ASSERT(session,
-			    upd->txnid == txn->id ||
-			    upd->txnid == WT_TXN_ABORTED);
-			upd->txnid = WT_TXN_ABORTED;
+			if (prepare) {
+				F_CLR(txn, WT_TXN_PREPARE);
```
bvpvamsikrishna: As transaction's in prepared state are not allowed any other operations apart from commit/ rollback, but txn_op_resolve need to perform a cursor search, the prepare state is cleared temporarily.
Suggestions to restructure are most welcome.

src/btree/col_modify.c
```diff
@@ -241,6 +241,13 @@ __wt_col_modify(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt,
 			WT_ERR(__wt_insert_serial(
 			    session, page, cbt->ins_head, cbt->ins_stack,
 			    &ins, ins_size, skipdepth, exclusive));
+
+		/*
+		 * We need to update the recno in txn op, as recno is retrieved
+		 * now only.
+		 */
+		if (logged)
+			__txn_op_modify_recno(session, cbt->recno);
```
sulabhM: It doesn't look right in the layering. If I consider this function alone, I can't tell why is a cursor search operation trying to find the latest txn_op in the running transaction and updating its recno if it happens to be column store.

bvpvamsikrishna:
autobuild: retest this please

lukech:
retest this please

bvpvamsikrishna:
@sulabhM, I have updated as per our discussion, please check.

src/btree/bt_cursor.c
```diff
@@ -470,6 +470,46 @@ __wt_btcur_reset(WT_CURSOR_BTREE *cbt)
...
+{
+	WT_BTREE *btree;
+	WT_CURSOR *cursor;
+	WT_DECL_RET;
+	WT_SESSION_IMPL *session;
+	WT_UPDATE *upd;
+
+	btree = cbt->btree;
+	cursor = &cbt->iface;
+	session = (WT_SESSION_IMPL *)cursor->session;
```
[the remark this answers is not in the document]
sulabhM: Resolving as Vamsi pointed out that other functions in this file are similar.

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	case WT_TXN_OP_INMEM_ROW:
+		WT_RET(__wt_open_cursor(session,
+		    op->btree->dhandle->name, NULL, open_cursor_cfg, &cursor));
+
+		/*
+		 * Transaction prepare is cleared temporarily as cursor
+		 * functions are not allowed for prepared transactions.
+		 */
+		F_CLR(txn, WT_TXN_PREPARE);
+		if (op->type == WT_TXN_OP_BASIC_ROW ||
```
sulabhM: My preference is that we be consistent with the conditional we used when we populated the key, ie we check for "btree->type == BTREE_ROW".

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+			 *
+			 * As updating timestamp might not be an atomic
+			 * operation, we will manage using state.
+			 */
+			upd->prepare_state = WT_PREPARE_LOCKED;
+			WT_WRITE_BARRIER();
+			__wt_timestamp_set(
+			    &upd->timestamp, &txn->commit_timestamp);
+			WT_PUBLISH(upd->prepare_state, WT_PREPARE_RESOLVED);
+
```
sulabhM: Not sure if we need this new line here.

src/txn/txn.c
```diff
@@ -1073,6 +1085,7 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
 		WT_TRET(txn->notify->notify(txn->notify, (WT_SESSION *)session,
 		    txn->id, 0));
 
+	prepare = F_ISSET(txn, WT_TXN_PREPARE);
```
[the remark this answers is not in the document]
sulabhM: Did you consider prepare not being a local variable ?

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+    ]
+
+    # Transaction end types
+    txn_end = [
+        ('txn_commit', dict(txn_commit=True)),
+        ('txn_rollback', dict(txn_commit=False)),
+    ]
+
+    scenarios = make_scenarios(types, txn_end)
+    conn_config = 'log=(enabled)'
```
sulabhM: This 'conn_config' doesn't get used, can be removed.

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+        self.session.begin_transaction(self.txn_config)
+        c[1] = 1
+        # update the value with in this txn
+        self.session.prepare_transaction('prepare_timestamp=' + timestamp_str(100))
+        if self.txn_commit == True:
+            self.session.commit_transaction('commit_timestamp=' + timestamp_str(101))
+        else:
+            self.session.rollback_transaction()
+
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(150))
```
sulabhM: Do we need setting stable/oldest timestamp and checkpointing in this test. If it isn't functionally required, my preference would be to remove it to keep the test simple.

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+        # update a uncommitted value, insert and update a key.
+        c[2] = 1
+        c[2] = 2
+        self.session.prepare_transaction('prepare_timestamp=' + timestamp_str(200))
+        if self.txn_commit == True:
+            self.session.commit_transaction('commit_timestamp=' + timestamp_str(201))
+        else:
+            self.session.rollback_transaction()
+
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(250))
```
sulabhM: same as before, if we don't need to, lets avoid setting timestamp and checkpoints.

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(250))
+        self.session.checkpoint()
+
+        # Scenario: 3
+        # Check remove operation
+        #   remove an existing key.
+        #   remove an previously updated key.
```
sulabhM: "a previously"

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+            self.session.rollback_transaction()
+
+        # commit some keys, to generate the update chain subsequently.
+        self.session.begin_transaction(self.txn_config)
+        c[1] = 1
+        c[2] = 1
+        c[3] = 1
+        self.session.commit_transaction('commit_timestamp=' + timestamp_str(301))
+
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(350))
```
sulabhM: Similar to my comment before, do we need to advance the timestamp ?
We probably do not need a checkpoint here, connection reopen should take one.

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+
+    scenarios = make_scenarios(types, txn_end)
+    conn_config = 'log=(enabled)'
+
+    def test_prepare_conflict(self):
+        if not wiredtiger.timestamp_build():
+            self.skipTest('requires a timestamp build')
+
+        self.session.create(self.uri,
+            'key_format=i,value_format=i' + self.extra_config)
```
sulabhM: 'key_format=i,value_format=i' string gets used twice in this test, it might be cleaner to put it in a config as a local variable.

test/suite/test_prepare_lookaside02.py
```diff
@@ -0,0 +1,174 @@
...
+        # Remove a updated key
+        cur.set_key(3)
+        cur.remove()
+        self.session.prepare_transaction('prepare_timestamp=' + timestamp_str(400))
+        if self.txn_commit == True:
+            self.session.commit_transaction('commit_timestamp=' + timestamp_str(401))
+        else:
+            self.session.rollback_transaction()
+
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(450))
```
sulabhM: Similar to before.

src/btree/bt_cursor.c
```diff
@@ -470,6 +470,46 @@ __wt_btcur_reset(WT_CURSOR_BTREE *cbt)
...
+	    __cursor_col_search(session, cbt, NULL));
+
+	WT_ASSERT(session, cbt->compare == 0);
+
+	/*
+	 * Get the uncommitted update from the cursor.
+	 */
+	if (cbt->ins != NULL)
+		upd = cbt->ins->upd;
+	else {
```
agorrod: Why is this condition simpler than the similar check in `__cursor_chain_exceeded`?

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
 
 #endif /* HAVE_TIMESTAMPS */
 
+/*
+ * __txn_op_modify_recno --
+ *      Modify the latest txn op with the given recno.
+ */
+static inline void
+__txn_op_modify_recno(WT_SESSION_IMPL *session, uint64_t recno)
```
agorrod: It's not required with inline headers, but we generally keep the convention of using {{__wt}} prefix if the function is called outside of this file.

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
 
 #endif /* HAVE_TIMESTAMPS */
 
+/*
+ * __txn_op_modify_recno --
```
agorrod: Could you call this `set_recno` rather than `modify_recno` - we have WT_MODIFY structures, so the modify word can be confusing.

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	op = txn->mod + txn->mod_count - 1;
+
+	WT_ASSERT(session, op->type == WT_TXN_OP_BASIC_COL ||
+	    op->type == WT_TXN_OP_INMEM_COL);
+
+	op->u.op_col.recno = recno;
+}
+
+/*
+ * __txn_prepared_op_resolve --
```
agorrod: Does this need to be inline? Is it in a performance path?

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	WT_CURSOR *cursor;
+	WT_DECL_RET;
+	WT_TXN *txn;
+	WT_UPDATE *upd;
+	const char *open_cursor_cfg[] = {
+	    WT_CONFIG_BASE(session, WT_SESSION_open_cursor), NULL };
+
+	txn = &session->txn;
+
+	switch (op->type) {
```
agorrod: This switch doesn't look worthwhile to me. Could you do:
```
if (op->type == WT_TXN_OP_NONE || op->type == WT_TXN_OP_REF_DELETE || op->type == WT_TXN_OP_TRUNCATE_COL || op->type == WT_TXN_OP_TRUNCATE_ROW)
    return (0);
```
Then remove the switch? That allows you to remove a level of indenting as well.

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+		WT_WITH_BTREE(session,
+		    op->btree, ret = __wt_btcur_search_uncommitted(
+		    (WT_CURSOR_BTREE *)cursor, &upd));
+		WT_TRET(cursor->close(cursor));
+		WT_RET(ret);
+		WT_ASSERT(session, upd != NULL);
+		op->u.op_upd = upd;
+
+		for (; upd != NULL; upd = upd->next) {
+			/* Process all the uncommitted updates of this txn. */
```
agorrod: Can this transaction have committed updates? Use full words - i.e: not `txn`.

In fact, in light of the more extensive comment below, I think this comment can be removed.

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+			if (!commit) {
+				/* Rollback is just to update txn id. */
+				upd->txnid = WT_TXN_ABORTED;
+				continue;
+			}
+
+			/*
+			 * Newer updates are inserted at head of update chain,
+			 * and txn operations are added at the tail of the txn
+			 * modify chain.
```
agorrod: This should use full words and correct sentence structure, so:
```
Newer updates are inserted at the head of the update chain and transaction operations are added at the tail of the transaction modify chain.
```

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+			 *
+			 * For example, a transaction has modified [k,v] as
+			 * [k,v] -> [k, u1]   (txn_op : txn_op1)
+			 * [k, u1] -> [k, u2] (txn_op : txn_op2)
+			 * update chain : u2->u1
+			 * txn_mod      : txn_op1->txn_op2.
+			 *
+			 * we save only key in the txn op, hence during
+			 * commit/rollback we cannot identify "txn_op1"
+			 * corresponds to either "u2" or "u1".
```
agorrod: Grammar:

Only the key is saved in the transaction op structure, hence we cannot identify whether "txn_op1" corresponds to "u2" or "u1" during commit/rollback.

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+			 * multiple updates of a key will be resolved as part of
+			 * the first txn op resolution of that key, and
+			 * subsequent txn op resolution of the same key will be
+			 * effectively a no-op.
+			 *
+			 * In the above example, we will resolve "u2" and "u1"
+			 * as part of resolving "txn_op1" and will not do any
+			 * significant thing as part of "txn_op2".
+			 */
+			if ( upd->prepare_state == WT_PREPARE_RESOLVED)
```
agorrod: Spacing is wrong - no space after the opening brace

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+			upd->prepare_state = WT_PREPARE_LOCKED;
+			WT_WRITE_BARRIER();
+			__wt_timestamp_set(
+			    &upd->timestamp, &txn->commit_timestamp);
+			WT_PUBLISH(upd->prepare_state, WT_PREPARE_RESOLVED);
+
+		}
+#ifdef HAVE_DIAGNOSTIC
+		upd = op->u.op_upd;
+		/* Ensure that we have not missed update of this txn. */
```
agorrod: Is better as: `Ensure that we have not missed any of this transactions updates.`

src/include/txn.i
```diff
@@ -193,6 +193,166 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+			 * In case of a prepared transaction, the order of
+			 * modification of the prepare timestamp to commit
+			 * timestamp in the update chain will not affect the
+			 * data visibility, a reader will encounter a prepared
+			 * update resulting in prepare conflict.
+			 *
+			 * As updating timestamp might not be an atomic
+			 * operation, we will manage using state.
+			 */
+			upd->prepare_state = WT_PREPARE_LOCKED;
```
agorrod: Did you consider setting the update in the WT_TXN_OP structure and using a call to `__wt_txn_op_set_timestamp` here? It'd save duplicating this subtle code.

bvpvamsikrishna: we cannot call `__wt_txn_op_set_timestamp` as this txn will have a commit timestamp, to avoid the duplication of this subtle code, I have added an inline function `__txn_resolve_prepared_update`.

src/btree/bt_cursor.c
```diff
@@ -470,6 +470,45 @@ __wt_btcur_reset(WT_CURSOR_BTREE *cbt)
...
+	WT_DECL_RET;
+	WT_SESSION_IMPL *session;
+	WT_UPDATE *upd;
+
+	btree = cbt->btree;
+	cursor = &cbt->iface;
+	session = (WT_SESSION_IMPL *)cursor->session;
+	upd = NULL;					/* -Wuninitialized */
+
+	WT_ERR(btree->type == BTREE_ROW ?
```
agorrod: This should be WT_RET, since there is no special cleanup needed.

src/btree/bt_cursor.c
```diff
@@ -470,6 +470,45 @@ __wt_btcur_reset(WT_CURSOR_BTREE *cbt)
...
+	    __cursor_row_search(session, cbt, NULL, false) :
+	    __cursor_col_search(session, cbt, NULL));
+
+	WT_ASSERT(session, cbt->compare == 0);
```
agorrod: This assert deserves a comment - it's not normal for a search function to expect a match.

src/btree/bt_cursor.c
```diff
@@ -470,6 +470,45 @@ __wt_btcur_reset(WT_CURSOR_BTREE *cbt)
...
+	else if (cbt->btree->type == BTREE_ROW) {
+		WT_ASSERT(session, cbt->ref->page->modify != NULL &&
+		    cbt->ref->page->modify->mod_row_update != NULL);
+		upd = cbt->ref->page->modify->mod_row_update[cbt->slot];
+	}
+
+	WT_ASSERT(session, upd != NULL);
+
+	*updp = upd;
+err:	return (ret);
```
agorrod: This probably becomes `return (0);`

src/btree/col_modify.c
```diff
@@ -241,6 +241,14 @@ __wt_col_modify(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt,
 			WT_ERR(__wt_insert_serial(
 			    session, page, cbt->ins_head, cbt->ins_stack,
 			    &ins, ins_size, skipdepth, exclusive));
+
+		/*
+		 * In case of append, the recno(key) for the value is assigned
+		 * now, hence the recno in the latest txn op representing this
+		 * operation need to be updated.
```
agorrod: Space before `(key)`, `txn` => `transaction`,`need` => `needs`.

src/include/txn.i
```diff
@@ -193,6 +193,176 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	WT_UNUSED(upd);
+#endif
+}
+
+/*
+ * __txn_prepared_op_resolve --
+ *      Resolve a transaction operation indirect references.
+ *
+ *      In case of prepared transactions, the prepared updates could be evicted
+ *      using cache overflow mechanism. Transaction operations referring these
```
agorrod: referring these => referring to these

src/include/txn.i
```diff
@@ -193,6 +193,176 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	else
+		((WT_CURSOR_BTREE *)cursor)->iface.recno =
+		    op->u.op_col.recno;
+	F_SET(txn, WT_TXN_PREPARE);
+
+	WT_WITH_BTREE(session,
+	    op->btree, ret = __wt_btcur_search_uncommitted(
+	    (WT_CURSOR_BTREE *)cursor, &upd));
+	WT_ERR(ret);
+	WT_ASSERT(session, upd != NULL);
```
agorrod: I think this is going away, but it's also duplicated inside __wt_btcur_search_uncommitted - choose one or the other.

src/include/txn.i
```diff
@@ -193,6 +193,176 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+	op->u.op_upd = upd;
+
+	for (; upd != NULL; upd = upd->next) {
+		 if (upd->txnid != txn->id)
+			continue;
+
+		if (!commit) {
+			/* Abort the updates of this transaction. */
```
agorrod: No need for this comment.

src/include/txn.i
```diff
@@ -193,6 +193,172 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+{
+	WT_TXN *txn;
+	WT_TXN_OP *op;
+
+	txn = &session->txn;
+
+	WT_ASSERT(session, txn->mod_count > 0 && recno != WT_RECNO_OOB);
+	op = txn->mod + txn->mod_count - 1;
+
+	WT_ASSERT(session, op->type == WT_TXN_OP_BASIC_COL ||
```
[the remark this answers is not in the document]
agorrod: We talked about this in person - resolving.

agorrod:
Thanks @bvpvamsikrishna I'm good with this change now.

@michaelcahill could you please take a look, and merge if you are happy?

src/include/txn.i
```diff
@@ -193,6 +193,205 @@ __wt_timestamp_subone(wt_timestamp_t *ts)
...
+#endif
+	}
+
+#ifdef HAVE_DIAGNOSTIC
+	/*
+	 * Reduce by one, to exclude the current processed one as to keep track
+	 * of only pre-processed ones.
+	 */
+	if (pre_processed_count < txn->pre_processed_count)
+		txn->pre_processed_count--;
```
michaelcahill: I find this logic really hard to follow (decrementing, then incrementing, now decrementing again).

Could something like this work instead?
```
if (upd != op->u.op_upd) {
    if (upd->prepare_state == WT_PREPARE_RESOLVED) {
        WT_ASSERT(session, txn->multi_update_count > 0);
        --txn->multi_update_count;
    } else
        ++txn->multi_update_count;
}
```

bvpvamsikrishna:

agorrod:
Thanks @bvpvamsikrishna I'm happy with this change now - @michaelcahill please do another round of review too.

michaelcahill:
Thanks @bvpvamsikrishna -- lgtm!  I'll merge.

PR #4272, opened 2018-09-04
WT-4272 Increase startup timeout to 30 seconds for slow I/O systems.

sueloverso:
@agorrod Here's the change to increase the timeout to 30 seconds, as discussed in the ticket.


agorrod:
lgtm

PR #4275, opened 2018-09-07
WT-3879 checkpoint race when WT_SESSION_LOCKED_METADATA isn't held

keithbostic:


michaelcahill:
As discussed [on the ticket](https://jira.mongodb.org/browse/WT-3879?focusedCommentId=1995954&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1995954), while I think the code here was confusing, I believe it did always acquire the metadata lock.

Closing in favor of #4276, which includes a cleanup to clarify this code.  I don't believe LSM should hold the metadata lock while flushing chunks -- it should only be held when flushing the metadata itself.

PR #4276, opened 2018-09-07
WT-3879 Fix a race if checkpoint evicts a metadata page.

michaelcahill:
The existing WT_SESSION_IS_CHECKPOINT macro is too broad: it allows a session-wide checkpoint to bypass the usual rules that make eviction of dirty pages safe.  Since we allow operations like WT_SESSION::create to run concurrent with a checkpoint and flush the metadata, we can't have the checkpoint being tapped for forced eviction in the metadata while it is being flushed.

Track the tree a checkpoint is currently flushing, and only allow checkpoint to bypass eviction rules in that tree.

### DIFF
--- a/src/block/block_ext.c
+++ b/src/block/block_ext.c
@@ -505,6 +505,10 @@ __wt_block_alloc(
 	WT_EXT *ext, **estack[WT_SKIP_MAXDEPTH];
 	WT_SIZE *szp, **sstack[WT_SKIP_MAXDEPTH];
 
+	/* If a sync is running, no other sessions can allocate blocks. */
+	WT_ASSERT(session, !WT_BTREE_SYNCING(S2BT(session)) ||
+	    WT_SESSION_TREE_SYNC(session));
+
 	/* Assert we're maintaining the by-size skiplist. */
 	WT_ASSERT(session, block->live.avail.track_size != 0);
 
@@ -622,8 +626,9 @@ __wt_block_off_free(
 {
 	WT_DECL_RET;
 
-	WT_ASSERT(session, S2BT(session)->checkpointing != WT_CKPT_RUNNING ||
-	   WT_SESSION_TREE_SYNC(session));
+	/* If a sync is running, no other sessions can free blocks. */
+	WT_ASSERT(session, !WT_BTREE_SYNCING(S2BT(session)) ||
+	    WT_SESSION_TREE_SYNC(session));
 
 	/*
 	 * Callers of this function are expected to have already acquired any
--- a/src/block/block_ext.c
+++ b/src/block/block_ext.c
@@ -506,8 +506,7 @@ __wt_block_alloc(
 	WT_SIZE *szp, **sstack[WT_SKIP_MAXDEPTH];
 
 	/* If a sync is running, no other sessions can allocate blocks. */
-	WT_ASSERT(session, !WT_BTREE_SYNCING(S2BT(session)) ||
-	    WT_SESSION_TREE_SYNC(session));
+	WT_ASSERT(session, WT_SESSION_BTREE_SYNC_SAFE(session, S2BT(session)));
 
 	/* Assert we're maintaining the by-size skiplist. */
 	WT_ASSERT(session, block->live.avail.track_size != 0);
@@ -627,8 +626,7 @@ __wt_block_off_free(
 	WT_DECL_RET;
 
 	/* If a sync is running, no other sessions can free blocks. */
-	WT_ASSERT(session, !WT_BTREE_SYNCING(S2BT(session)) ||
-	    WT_SESSION_TREE_SYNC(session));
+	WT_ASSERT(session, WT_SESSION_BTREE_SYNC_SAFE(session, S2BT(session)));
 
 	/*
 	 * Callers of this function are expected to have already acquired any
--- a/src/block/block_ext.c
+++ b/src/block/block_ext.c
@@ -622,6 +622,9 @@ __wt_block_off_free(
 {
 	WT_DECL_RET;
 
+	WT_ASSERT(session, S2BT(session)->checkpointing != WT_CKPT_RUNNING ||
+	   WT_SESSION_TREE_SYNC(session));
+
 	/*
 	 * Callers of this function are expected to have already acquired any
 	 * locks required to manipulate the extent lists.
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -240,14 +240,13 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 		 * Set the checkpointing flag to block such actions and wait for
 		 * any problematic eviction or page splits to complete.
 		 */
-		WT_ASSERT(session, btree->checkpointing == WT_CKPT_OFF);
+		WT_ASSERT(session, btree->syncing == WT_BTREE_SYNC_OFF &&
+		    btree->sync_session == NULL);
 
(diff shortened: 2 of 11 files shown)


src/block/block_ext.c
```diff
@@ -622,6 +622,9 @@ __wt_block_off_free(
 {
 	WT_DECL_RET;
 
+	WT_ASSERT(session, S2BT(session)->checkpointing != WT_CKPT_RUNNING ||
+	   WT_SESSION_TREE_SYNC(session));
+
```
michaelcahill: This is the check I used to track down when things were going wrong: ideally I'd like to catch the general case without a layering violation.

src/include/session.h
```diff
@@ -66,6 +66,10 @@ struct __wt_session_impl {
 
 	WT_DATA_HANDLE *dhandle;	/* Current data handle */
 
+	WT_DATA_HANDLE *sync_dhandle;	/* Data handle being synced */
+#define	WT_SESSION_TREE_SYNC(session)					\
+	(session->dhandle == session->sync_dhandle)
+
```
michaelcahill: This macro could use a better name but I haven't thought of one yet.  Suggestions welcome.

src/meta/meta_track.c
```diff
@@ -307,12 +307,11 @@ __wt_meta_track_off(WT_SESSION_IMPL *session, bool need_sync, bool unroll)
...
-		WT_WITH_METADATA_LOCK(session,
-		    WT_WITH_DHANDLE(ckpt_session,
-		    WT_SESSION_META_DHANDLE(session),
-		    ret = __wt_checkpoint(ckpt_session, NULL)));
-		F_CLR(ckpt_session, WT_SESSION_LOCKED_METADATA);
+		WT_ASSERT(session,
+		    !F_ISSET(session, WT_SESSION_LOCKED_METADATA));
+		WT_WITH_DHANDLE(ckpt_session, WT_SESSION_META_DHANDLE(session),
+		    WT_WITH_METADATA_LOCK(ckpt_session,
+			ret = __wt_checkpoint(ckpt_session, NULL)));
```
michaelcahill: I believe this code was structured oddly in case the session passed in already held the metadata lock.  Asserting that isn't the case instead: I can't see the path, but the assertion should fire in testing if I'm wrong.

agorrod:
@michaelcahill this change makes sense to me - I can't think of a better name for the variable or the macro - but I do feel like it'd be nice if it were linked together with WT_SESSION_IS_CHECKPOINT a bit more strongly. I'd be happy for this to be merged without a name change..

src/block/block_ext.c
keithbostic: I don't feel strongly about this kind of assert in the default block manager, it's good with me. Would it be worth adding a more general assertion if a block is allocated or freed in a file that's being checkpointed?

keithbostic: Trivial white-space error, the follow-on line is incorrectly indented.

src/btree/bt_sync.c
```diff
@@ -240,10 +240,15 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
 		 */
+		WT_ASSERT(session, btree->checkpointing == WT_CKPT_OFF);
+
 		btree->checkpointing = WT_CKPT_PREPARE;
 		(void)__wt_gen_next_drain(session, WT_GEN_EVICT);
 		btree->checkpointing = WT_CKPT_RUNNING;
 
+		/* Note in the session that we are syncing this tree. */
+		session->sync_dhandle = session->dhandle;
+
```
keithbostic: This won't get flushed and might get reordered. I don't think it causes problems but it might affect behavior.

Can we move the set above the `__wt_gen_next_drain` call? Pairing it with the initial set of `btree.checkpoint` makes sense to me, it's already paired with the clear.

src/include/btree.i
```diff
@@ -1236,6 +1236,14 @@ __wt_leaf_page_can_split(WT_SESSION_IMPL *session, WT_PAGE *page)
...
 	btree = S2BT(session);
 
+	/*
+	 * Checkpoints can not do in-memory splits in the tree they are
+	 * walking: that can lead to corruption when the parent internal page
+	 * is updated.
+	 */
+	if (WT_SESSION_TREE_SYNC(session))
+		return (false);
+
```
keithbostic: Typo, "can not" -> "cannot".

src/include/session.h
keithbostic: I'm with @agorrod, I don't have a better name.

src/btree/bt_sync.c
keithbostic: Why isn't `WT_SESSION.sync_dhandle` a flag or boolean?

I'm not positive of this, but isn't it the same as holding `WT_SESSION_LOCKED_CHECKPOINT`? (I'm not excited about actually using `WT_SESSION_LOCKED_CHECKPOINT`, but hopefully that helps explain the point.)

Alternatively, it might be more intuitive to have `WT_SESSION.sync_btree`, that makes it more obvious we're referencing the same tree as `WT_BTREE.checkpointing != WT_CKPT_OFF`.

keithbostic: Edit: never mind, of course it can't be a boolean.

I still think `WT_SESSION.sync_btree` makes a little more sense than `WT_SESSION.sync_dhandle`.

michaelcahill: FTR, there is subtlety: if a checkpoint is currently working in ordinary tree T, and as part of that it updates T's metadata, the ordinary rules should be followed for the metadata write.  In particular, there can be a concurrent flush of the metadata (e.g., as part of a `WT_SESSION::create`), and the checkpoint should not be able to evict or split a metadata page in that case.

michaelcahill: I don't think there's any flushing issue in this version because session handles aren't shared between threads.  That said, I'm about to push a different version that reorders as you suggest.

src/btree/bt_sync.c
```diff
@@ -383,7 +387,8 @@ __sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 		__wt_txn_release_snapshot(session);
 
 	/* Clear the checkpoint flag. */
-	btree->checkpointing = WT_CKPT_OFF;
+	btree->syncing = WT_BTREE_SYNC_OFF;
+	btree->sync_session = NULL;
```
agorrod: Is the order important here? Do we need to make sure that syncing state change is visible to other threads before clearing `sync_session`?

michaelcahill:
Thanks @agorrod -- I'm fairly happy with this branch now, so please go ahead and merge.

src/btree/bt_sync.c
michaelcahill: I don't think it matters -- the value of `sync_session` is only relevant if it is equal to `session`: in other words, it's only the current thread that will do anything different depending on the value, so the order of operations here isn't important.

It's also worth noting that this code already assumes that there is some higher level concurrency control preventing multiple threads from syncing the same tree at the same time.  So if you're worried that some other thread will flip `syncing` to `WT_BTREE_SYNC_RUNNING` and have its `sync_session` cleared, I don't think that's possible.

All of that said, if the code reads better with manipulation of `sync_session` nested inside `syncing` (or vice versa), I'm fine with that.

agorrod: Thanks for the explain - makes sense, and I don't think it requires further changes for clarity. I'll go ahead and merge.

PR #4277, opened 2018-09-07
WT-3736 Add a statistics to measure contention on lookaside cursor.

nehakhatri5:


dist/stat_data.py
```diff
@@ -260,6 +260,7 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_hazard_walks', 'hazard pointer check entries walked'),
     CacheStat('cache_inmem_split', 'in-memory page splits'),
     CacheStat('cache_inmem_splittable', 'in-memory page passed criteria to be split'),
+    CacheStat('cache_lookaside_cursor', 'sleep for cache overflow table cursor'),
```
sulabhM: I agree with Michael, we should count the total time we wait here. It will be awesome if we can rename "cache_lookaside_cursor" to something accordingly, "cache_lookaside_cursor_acquire_time" maybe.

dist/stat_data.py
```diff
@@ -260,6 +260,7 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_hazard_walks', 'hazard pointer check entries walked'),
     CacheStat('cache_inmem_split', 'in-memory page splits'),
     CacheStat('cache_inmem_splittable', 'in-memory page passed criteria to be split'),
+    CacheStat('cache_lookaside_cursor_wait', 'cache overflow cursor wait time (usecs)'),
```
agorrod: I think it would be worthwhile to split this into two statistics, similar to many of our other wait based statistics. Similar to [lock_table_wait_application and lock_table_wait_internal](https://github.com/wiredtiger/wiredtiger/blob/develop/dist/stat_data.py#L353)

You can differentiate between the different cases using the [WT_SESSION_INTERNAL flag](https://github.com/wiredtiger/wiredtiger/blob/develop/src/evict/evict_lru.c#L2108)

agorrod:
@sueloverso please take a look and merge if you are happy.

PR #4279, opened 2018-09-09
WT-4300 setting the update timestamp can overwrite the WT_REF.addr field

keithbostic:
@bvpvamsikrishna, @agorrod, I'm not sure this fix is correct, but hopefully 2551e61 points out where the problem is.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -279,7 +279,7 @@ __wt_txn_update_set_timestamp(WT_SESSION_IMPL *session,
 	 */
 	timestamp = op->type == WT_TXN_OP_REF_DELETE ?
 	    &op->u.ref->page_del->timestamp : &op->u.op_upd->timestamp;
-	needs_timestamp = 
+	needs_timestamp =
 	    __wt_timestamp_iszero(timestamp) || F_ISSET(txn, WT_TXN_PREPARE);
 	if (!needs_timestamp)
 		return;
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -235,14 +235,11 @@ static inline void
 __wt_txn_unmodify(WT_SESSION_IMPL *session)
 {
 	WT_TXN *txn;
-	WT_TXN_OP *op;
 
 	txn = &session->txn;
 	if (F_ISSET(txn, WT_TXN_HAS_ID)) {
 		WT_ASSERT(session, txn->mod_count > 0);
-		op = txn->mod + txn->mod_count - 1;
-		__wt_txn_op_free(session, op);
-		txn->mod_count--;
+		__wt_txn_op_free(session, txn->mod + --txn->mod_count);
 	}
 }
 
@@ -262,9 +259,10 @@ __wt_txn_update_set_timestamp(WT_SESSION_IMPL *session,
 	wt_timestamp_t *timestamp;
 	bool needs_timestamp;
 
-	txn = &session->txn;
 	if (was_set != NULL)
 		*was_set = false;
+
+	txn = &session->txn;
 
 	/*
 	 * The timestamp is in the page deleted structure for truncates, or
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -259,54 +259,37 @@ __wt_txn_op_commit_page_del(WT_SESSION_IMPL *session, WT_REF *ref)
 
 	txn = &session->txn;
 
-	/*
-	 * The page-deleted list can be discarded by eviction,
-	 * lock the WT_REF to ensure we don't race.
-	 */
+	/* Avoid locking the page if a previous eviction already cleaned up. */
 	if (ref->page_del->update_list == NULL)
 		return;
 
+	/*
+	 * Lock the ref to ensure we don't race with eviction freeing the
+	 * page deleted update list.
+	 */
 	for (;; __wt_yield()) {
 		previous_state = ref->state;
 		if (previous_state != WT_REF_LOCKED &&
 		    __wt_atomic_casv32(
 		    &ref->state, previous_state, WT_REF_LOCKED))
-			return;
+			break;
 	}
 
-	if ((updp = ref->page_del->update_list) == NULL) {
-		/*
-		 * Publish to ensure we don't let the page be
-		 * evicted and the updates discarded before
-		 * being written.
-		 */
-		WT_PUBLISH(ref->state, previous_state);
-		return;
-	}
-
-	for (; *updp != NULL; ++updp) {
-		if (F_ISSET(txn, WT_TXN_PREPARE)) {
+	for (updp = ref->page_del->update_list;
+	    updp != NULL && *updp != NULL; ++updp) {
+		__wt_timestamp_set(&(*updp)->timestamp, &txn->commit_timestamp);
+		if (F_ISSET(txn, WT_TXN_PREPARE))
 			/*
-			 * As ref state is LOCKED, timestamp
-			 * and prepare state are updated in
-			 * exclusive access, hence no need for
-			 * temporary state WT_PREPARE_LOCKED
-			 * and BARRIER.
+			 * Holding the ref locked means we have exclusive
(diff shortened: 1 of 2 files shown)


keithbostic:
As a separate comment, I think there's a problem in `__wt_txn_update_set_timestamp()`.

In this clause:
```
if (op->type == WT_TXN_OP_REF_DELETE)
        timestamp = op->u.ref == NULL || op->u.ref->page_del == NULL ?
            NULL : &op->u.ref->page_del->timestamp;
else    
        timestamp = op->u.op_upd == NULL ?
            NULL : &op->u.op_upd->timestamp;

needs_timestamp = timestamp == NULL || 
    __wt_timestamp_iszero(timestamp) || F_ISSET(txn, WT_TXN_PREPARE);
if (!needs_timestamp)
        return; 
```
the variable `timestamp` is (reasonably) anticipated to be NULL, but if it is actually NULL, then I think we're going to drop core when we call `__wt_timestamp_set()` a little further on in this function.

I'm pretty sure that in cases other than `WT_TXN_OP_REF_DELETE` timestamp cannot be NULL, (but I'm less sure about that case). Anyway, my guess is this should be:
```
timestamp = op->type == WT_TXN_OP_REF_DELETE :
    &op->u.ref->page_del->timestamp : &op->u.op_upd->timestamp;
```

src/include/txn.i
```diff
@@ -235,14 +235,11 @@ static inline void
...
 	WT_TXN *txn;
-	WT_TXN_OP *op;
 
 	txn = &session->txn;
 	if (F_ISSET(txn, WT_TXN_HAS_ID)) {
 		WT_ASSERT(session, txn->mod_count > 0);
-		op = txn->mod + txn->mod_count - 1;
-		__wt_txn_op_free(session, op);
-		txn->mod_count--;
+		__wt_txn_op_free(session, txn->mod + --txn->mod_count);
```
agorrod: I'm sure this is correct, but I find it hard to reason about. I don't really like getting array offsets via addition - which is why I preferred to make it explicit. I'd presumed the compiler was smart enough to get rid of the local variable for me.

src/include/txn.i
```diff
@@ -262,36 +259,32 @@ __wt_txn_update_set_timestamp(WT_SESSION_IMPL *session,
...
 	 * Updates in the metadata never get timestamps (either now or at
 	 * commit): metadata cannot be read at a point in time, only the most
 	 * recently committed data matches files on disk.
 	 */
-	needs_timestamp = !WT_IS_METADATA(op->btree->dhandle) &&
-	    F_ISSET(txn, WT_TXN_HAS_TS_COMMIT) &&
-	    (timestamp == NULL || __wt_timestamp_iszero(timestamp) ||
-	    F_ISSET(txn, WT_TXN_PREPARE));
+	if (WT_IS_METADATA(op->btree->dhandle) ||
+	    !F_ISSET(txn, WT_TXN_HAS_TS_COMMIT))
```
agorrod: I don't believe it's OK to exit early if the transaction is in WT_TXN_PREPARE.

src/include/txn.i
```diff
@@ -262,36 +259,32 @@ __wt_txn_update_set_timestamp(WT_SESSION_IMPL *session,
...
+		return;
 
+	/*
+	 * The timestamp is in the page deleted structure for truncates, or
+	 * in the update for other operations.
+	 */
+	timestamp = op->type == WT_TXN_OP_REF_DELETE ?
+	    &op->u.ref->page_del->timestamp : &op->u.op_upd->timestamp;
```
agorrod: This simplification seems OK to me right now, but we won't always have a valid `op->u.op_upd` - that's part of the ongoing prepared transaction work. I don't like nested ternary statements, which is why I exploded it out.
ps: I know I can't expect you to know what I'm thinking will happen soon.

src/include/txn.i
```diff
@@ -262,36 +259,32 @@ __wt_txn_update_set_timestamp(WT_SESSION_IMPL *session,
...
+	needs_timestamp =
+	    __wt_timestamp_iszero(timestamp) || F_ISSET(txn, WT_TXN_PREPARE);
 	if (!needs_timestamp)
 		return;
 
-	if (for_prepare) {
+	if (for_prepare && op->type != WT_TXN_OP_REF_DELETE) {
```
agorrod: Nice catch - this makes me wonder if we should pull the code from `__wt_txn_commit` dealing with prepared `WT_TXN_OP_REF_DELETE` into here as well. I didn't because it also locks the ref, and I didn't think this was an obvious place to go looking for that.
@bvpvamsikrishna do you have an opinion about whether we should write a comment here describing why REF_DELETE isn't handled in this function, or moving the functionality in here?

bvpvamsikrishna: @agorrod, my opinion is to move functionality in here and rename the function as "__wt_txn_op_set_timestamp" (changed the update --> op).

src/include/txn.i
```diff
@@ -262,36 +259,32 @@ __wt_txn_update_set_timestamp(WT_SESSION_IMPL *session,
...
+		return;
 
+	/*
+	 * The timestamp is in the page deleted structure for truncates, or
+	 * in the update for other operations.
+	 */
+	timestamp = op->type == WT_TXN_OP_REF_DELETE ?
+	    &op->u.ref->page_del->timestamp : &op->u.op_upd->timestamp;
+	needs_timestamp =
+	    __wt_timestamp_iszero(timestamp) || F_ISSET(txn, WT_TXN_PREPARE);
```
bvpvamsikrishna: cosmetic : the value of "for_prepare" will be same as "F_ISSET(txn, WT_TXN_PREPARE)", could you just one for consistency.

src/include/txn.i
agorrod: @bvpvamsikrishna pointed out that a transaction in WT_TXN_PREPARE will always have a commit timestamp - so it's OK not to test it explicitly.

agorrod:
autobuild: retest this please

agorrod:
@bvpvamsikrishna and @michaelcahill please take a look at how this landed - it ended up being more a more significant change than I expected, but I think the new code is simpler..

src/include/txn.i
```diff
@@ -235,83 +235,137 @@ static inline void
...
+	 */
+	if (ref->page_del->update_list == NULL)
+		return;
+
+	for (;; __wt_yield()) {
+		previous_state = ref->state;
+		if (previous_state != WT_REF_LOCKED &&
+		    __wt_atomic_casv32(
+		    &ref->state, previous_state, WT_REF_LOCKED))
+			return;
```
bvpvamsikrishna: this should be break; not return;
as we will lock and proceed down.

bvpvamsikrishna:
lgtm apart from changing return --> break in txn.i:274

agorrod:
@keithbostic please take a look at the changes I pushed in here today.

I've updated a stress test branch to run here as well: http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/

src/include/txn.i
keithbostic: @agorrod, I'm sure you're correct, the compiler will kill the local store.

FTR, my reasoning is I don't like to see things done more than once, so the `x - 1` followed by `--x` triggered for me. I'm good either way if anyone else has a preference.

src/include/txn.i
keithbostic: > @bvpvamsikrishna pointed out that a transaction in WT_TXN_PREPARE will always have a commit timestamp

I think that's worth a comment.

src/include/txn.i
keithbostic: > This simplification seems OK to me right now, but we won't always have a valid `op->u.op_upd`

Then maybe the clearest choice is to explicitly handle a NULL timestamp reference now.

agorrod:
Thanks Keith, sounds good

PR #4282, opened 2018-09-10
WT-4306 Fix mode if metadata pages need eviction.

michaelcahill:
Specifically, as a checkpoint is running, it writes updates to the metadata table, which can cause pages to exceed maximum sizes.  It's fine to attempt eviction in that situation, but it should be ordinary eviction that respects the usual visibility rules.


agorrod:
Thanks @michaelcahill this change makes sense to me - I'm going to go ahead and merge.

PR #4283, opened 2018-09-11
WT-4308 Insert split during sync should not free blocks.

michaelcahill:
It is not currently safe to allocate or free blocks in a tree where checkpoint is running. That's because the checkpoint needs to calculate a map of free space that corresponds to the pages referenced by the checkpoint (otherwise either verify reports holes in the file).

What we saw was an in-memory split is causing an on-disk block to be freed. That was because the split freed any deleted references it found in the parent, which includes freeing their blocks.


agorrod:
Thanks @michaelcahill lgtm

PR #4284, opened 2018-09-14
WT-4314 Don't access unmapped pages when evicting a tree.

michaelcahill:


michaelcahill:
This feels like whack-a-mole -- @bvpvamsikrishna wrote a test that found a case where we were accessing mapped memory after the tree was marked dead (when we unmap everything).  This branch fixes that specific case, but I don't have a lot of confidence that this is the only place we can get into trouble.

michaelcahill:
Closing in favor of #4287.

PR #4285, opened 2018-09-17
WT-4321 In direct IO test, added checkpoint after table creation so they will be available for checking

ddanderson:


ddanderson:
@agorrod, can you please review?  I'm hoping this will resolve the problem we're seeing.

agorrod:
Thanks @ddanderson this change makes sense to me - I'll go ahead and merge once it passes testing. ps: I'm jealous of the ticket number..

@lukech please hold off reverting WT-4225

agorrod:
@ddanderson I've gone a step further and disabled the test from `make check` for now - I'll open up a follow on ticket.

PR #4287, opened 2018-09-17
WT-4314 Don't access unmapped pages when evicting a tree.

keithbostic:


michaelcahill:
@keithbostic, we'll need to talk this through.  Forced drops aren't just for "performance reasons": by the time a forced drop succeeds, applications expect that the file has been removed so the namespace can be reused.  Keeping a memory mapping open after the forced drop completes will break that contract (at least on Windows).  It still might be the best solution, but it needs more discussion.

michaelcahill:
Oh, maybe I'm misreading (there is at least one too many `bool` in that function for me to keep in my head).  If I'm understanding, this change discards the tree from cache if it is mapped, but does that up front, so that there should be no need to mark the tree dead -- you're suggesting we completely discard mapped trees?

lukech:
retest this please

michaelcahill:
@keithbostic I've pushed a change that makes `__wt_conn_dhandle_close` easier for me to understand, and should mean that mapped handles are immediately closed (they don't hang around in the system with `WT_DHANDLE_DEAD` set, which was the case in the original failure).  Otherwise, it should be functionally equivalent to what you pushed.

Can you please take a look, and merge if you like it?

keithbostic:
@michaelcahill, this change looks correct to me (although I still find the function terribly confusing).

The only problem I see is that we'll call `__wt_cache_op(..., WT_SYNC_DISCARD)` twice on mapped files, which should work as far as I can tell, so if that was deliberate on your part, I'm good with the change, let's merge.

michaelcahill:
Yes, @keithbostic -- calling `__wt_cache_op(..., WT_SYNC_DISCARD)` multiple times is safe (the first thing it does is bail out if there is no root page), and it got rid of one of the conditionals in the previous version of the code (`discard`).

I'll go ahead and merge...

PR #4288, opened 2018-09-17
WT-4149 If salvaging, recovery should checkpoint and complete.

sueloverso:


sueloverso:
Not ready for merge.

sueloverso:
I think this is ready for review. There are some things to note. There was a bit of whack-a-mole with this. I ended up adding a flag to know when recovery failed versus when wiredtiger_open failed so we could do the right thing on the error path out.

One change that is not obvious is the removal of the check for `ENOENT` in two locations in `txn_recover.c`. That check was added in WT-1936 several years ago. A new test was added as part of that work as well. The test still passes. I believe later changes made that `ENOENT` code no longer needed. In particular, the changes that added a `checkpoint_start` record to the log and not writing out the metadata on an idle checkpoint.

Right now any failure from `wt_log_scan` during salvage is ignored. As part of WT-4260, I intend to detect and set `WT_CONN_DATA_CORRUPTION` and enhance that check as well as the considerations of the warning versus error. So that will be part of that work.

test/suite/test_txn19.py
```diff
@@ -111,9 +111,9 @@ def includeFunc(name, dictarg):
         # NOTE:
         # The removal or truncation of a middle log file (not first or last)
         # that would be used in recovery is not currently handled gracefully.
-        if (kind == 'removal' or kind == 'truncate') and \
-           corruptpos != 6 and corruptpos > chkpt:
-            return False
```
ddanderson: It would be good to remove the NOTE comment and the commented out code here.

src/txn/txn_recover.c
```diff
@@ -321,6 +320,15 @@ __txn_log_recover(WT_SESSION_IMPL *session,
 	/* First, peek at the log record type. */
 	WT_RET(__wt_logrec_read(session, &p, end, &rectype));
 
+	/*
+	 * Record the highest LSN we process during the metadata phase.
+	 * If not the metadata phase, then stop at that LSN.
+	 */
+	if (r->metadata_only)
+		r->max_rec_lsn = *next_lsnp;
+	else if (__wt_log_cmp(lsnp, &r->max_rec_lsn) > 0)
```
ddanderson: Should this comparison be `if (__wt_log_cmp(lsnp, &r->max_rec_lsn) >= 0)` instead of `... > 0)` ?

src/txn/txn_recover.c
```diff
@@ -592,6 +602,12 @@ __wt_txn_recover(WT_SESSION_IMPL *session)
 	 * First, do a pass through the log to recover the metadata, and
 	 * establish the last checkpoint LSN.  Skip this when opening a hot
 	 * backup: we already have the correct metadata in that case.
+	 *
+	 * If we're running with salvage and we hit an error, we ignore it
+	 * and continue. In salvage we want to recover whatever part of the
+	 * data we can from the last checkpoint up until whatever problem we
+	 * detect in the log file. In salvage we only care about errors from
+	 * scanning the log. Other errors remain errors.
```
ddanderson: I think the second to last sentence of this comment doesn't make sense with my read of the code.  Looks like when salvaging we are ignoring errors from scanning the log.

sueloverso:

PR #4289, opened 2018-09-18
WT-4322 Reenable direct IO 'crash' test, with directory copying in C

ddanderson:


ddanderson:
@agorrod, can you please review?  I have run this on aws-build-test2 10 times with success. I haven't seen the recovery failure that you saw either.  That could well be a case of _bad things happen when you have an incomplete directory copy_. I'm going to keep testing with different options just to see if I can catch that failure mode again.

agorrod:
Thanks @ddanderson this change makes sense to me. On the other hand I believe WT-4328 is a real issue that this test will also uncover - I'm going to hold off merging until we have a better handle on what to do about that ticket.

Alternatively I'd be very happy to take this change without the part that reverts smoke.sh right now, and enable testing later. @sueloverso I believe that would be better if the immediate solution to WT-4328 is to make changes to the test case.

ddanderson:
I'm thinking about the wisdom of @sueloverso's suggestion, I'll look at it tomorrow.  I'm pondering another possibility - have an option to put the schema operations outside the transactions. That makes the checking side more complex, since at many stop points, either of two possibilities can be considered "correct": the current schema operation completed or it didn't. It does have the benefit that it is closer to what MongoDB does, and probably more in line with what typical WT users do.  (Are there WT users that want to put one or more schema operations in the same transaction with a data operation)? It has the downside of adding a little more complexity to a test that is already complex.

ddanderson:
@sueloverso and @agorrod, I have modified the test so the default case will be closer to what I think MongoDB does.  That is, there are still schema operations, and by default they done outside any explicit transactions. We'll still expect recovery to succeed, of course, and we'll check the sanity of the main table and index table, but in the default case, we don't check the status of the schema files.  It seems like we _should_ be able to pass this.  We can get back the more difficult test by turning on -S "integrated" (which integrates schema operations into table transactions).

Let me know what you think.

test/csuite/random_directio/smoke.sh
```diff
@@ -25,10 +22,10 @@ for threads in $TEST_THREADS; do
 
         # Here are successively tougher schema tests that do not yet
         # reliably pass.  'verbose' can be added to any.
-        #$RUN_TEST -T $threads -S create,create_check       || exit 1
-        #$RUN_TEST -T $threads -S create,drop,drop_check    || exit 1
-        #$RUN_TEST -T $threads -S create,rename             || exit 1
-        #$RUN_TEST -T $threads -S create,rename,drop_check  || exit 1
-        #$RUN_TEST -T $threads -S all,verbose               || exit 1
```
sueloverso: If we take the integrated changes, then these can be uncommented, right?

ddanderson: If we take the integrated changes, all the new lines with "integrated" are legal, but they won't consistently pass. The old (deleted) lines without "integrated" might pass, except that create_check/drop_check/data_check don't make any sense without "integrated" being set.  I just added a check to disallow that combination.  So the only deleted one that might pass is "create,rename". We don't use rename in MongoDB to my knowledge, so that test seems less important for now.

PR #4292, opened 2018-09-23
WT-4333 WiredTiger cursor cache doesn't handle all possible locked handle states

keithbostic:


keithbostic:
@ddanderson, I think this one is likely yours.

What's happening is the `__wt_session_lock_dhandle()` call in `__curfile_reopen()` can return a locked handle that is no longer open, if an exclusive operation opened/closed the handle in-between operations.

It seems to me that the cursor cache only wants handles that are open, and so I added a test for `WT_DHANDLE_OPEN`. I don't understand cursor caching in any detail, so don't hesitate to discard this work if there's a better solution.

keithbostic:
@ddanderson, one other thing. This code in the original looks wrong to me:

```
if (!WT_DHANDLE_CAN_REOPEN(dhandle))
        ret = WT_NOTFOUND;
if (!check_only) {
        session->dhandle = dhandle;
        WT_TRET(__wt_session_lock_dhandle(session, 0, &is_dead));
```

Imagine if `ret` is set to `WT_NOTFOUND`, then `check_only` is true, and the call to `__wt_session_lock_dhandle` returns `WT_PANIC`. Because `ret` is already set, `WT_TRET` will ignore the `WT_PANIC` which isn't right.

I don't know what the right fix is here, but I think there's a potential problem, and clarification would be good.

michaelcahill:
FWIW, the specific case of `WT_TRET` handling `WT_PANIC` when `ret == WT_NOTFOUND` will set `ret = WT_PANIC`: https://github.com/wiredtiger/wiredtiger/blob/develop/src/include/error.h#L80-L87

I think the general point about clarity still holds, though.

keithbostic:
> FWIW, the specific case of WT_TRET handling WT_PANIC

I'm even more wrong than that -- `WT_TRET` checks for `ret == WT_NOTFOUND`, so pretty much any other error will overwrite the `WT_NOTFOUND`. That said, I'd still appreciate clarity. :)

ddanderson:
@keithbostic, your changes look correct to me.  It led to some simplifications, I just pushed here - I think it makes things clearer.  I can add some commentary if not. Thanks!

ddanderson:
retest this please

PR #4293, opened 2018-09-24
WT-3995 Enhance timestamp abort to accept more number of threads.

nehakhatri5:


test/csuite/timestamp_abort/main.c
```diff
@@ -451,6 +453,12 @@ run_workload(uint32_t nth)
 	if (compat)
 		strcat(envconf, ENV_CONFIG_COMPAT);
 
+	if (nth > SESSION_MAX) {
```
agorrod: I don't think this needs to be conditional - just configure enough sessions so that the maximum thread count won't run out. It's also not going to be OK to configure 99 threads with this logic.

test/csuite/timestamp_abort/main.c
```diff
@@ -56,9 +56,10 @@ static char home[1024];			/* Program working dir */
  * Each worker thread creates its own records file that records the data it
  * inserted and it records the timestamp that was used for that insertion.
  */
+#define	INT_SESSION_BUF	36
```
agorrod: I don't find this name intuitive. BUF makes me think of memory allocation.

sulabhM:
LGTM.

No strong opinion about it, but ENV_CONFIG_TXNSYNC being incomplete feels weird to me. Have you considered an alternative implementation similar to yours (haven't tried coding this one, just throwing an idea to consider):

```
#define	ENV_CONFIG_TXNSYNC					\
    "create,log=(archive=false,file_max=10M,enabled),"			\
    "transaction_sync=(enabled,method=none)"

. . . .

    (void)__wt_snprintf(envconf, sizeof(envconf),
        "%s,session_max=%" PRIu32 ",", ENV_CONFIG_TXNSYNC, MAX_TH + INT_SESSION_CNT);
```

nehakhatri5:
> No strong opinion about it, but ENV_CONFIG_TXNSYNC being incomplete feels weird to me. Have you considered an alternative implementation similar to yours (haven't tried coding this one, just throwing an idea to consider):
> 
I saw an [example](https://github.com/wiredtiger/wiredtiger/blob/81305b4ade4b92bde247931334c99c32fb148c44/test/csuite/timestamp_abort/main.c#L68) macro in this test program and followed that. I thought it would be good to keep the `session_max` within `ENV_CONFIG_TXNSYNC`.

test/csuite/timestamp_abort/main.c
```diff
@@ -56,9 +56,10 @@ static char home[1024];			/* Program working dir */
  * Each worker thread creates its own records file that records the data it
  * inserted and it records the timestamp that was used for that insertion.
  */
+#define	INT_SESSION_CNT 36
 #define	INVALID_KEY	UINT64_MAX
 #define	MAX_CKPT_INVL	5	/* Maximum interval between checkpoints */
-#define	MAX_TH		12
+#define	MAX_TH		200
```
sueloverso: Why do we need 36 extra sessions? Where did that number come from? The code inside WiredTiger already adds in an extra 20 for internal WT threads. The program only uses 3 internal threads to do other work.
Since it is a constant, I'd prefer it to be defined here as a constant rather than down in the code. Something like:
```
#define SESSION_CNT MAX_TH + 3 /* Maximum threads plus program worker threads */
```

michaelcahill:
@nehakhatri5, are you waiting for further review of this branch?  If so, does it have to be @sueloverso or would you like me to take a look?

PR #4294, opened 2018-09-24
Code simplifications.

ddanderson:
@keithbostic, can you review please?  Your changes look right, and it leads to some nice simplification.  The diff is a little confusing, better to look at the raw code.  I just added the early return which changed indenting for the function.  Thanks for finding/fixing this!


ddanderson:
@keithbostic, I'm pushing this on to your branch instead.

PR #4297, opened 2018-09-26
WT-4337 Coverity #1395811 Time of check time of use

keithbostic:
Don't call stat(), save the file size and then open the file, there's a potential race. Instead, open the file and use fstat().


ddanderson:
LGTM, merging.

PR #4298, opened 2018-09-26
WT-4338 New WT_TXN_TS_XXX flags break #undef HAVE_TIMESTAMPS build

keithbostic:


nehakhatri5:
Thanks @keithbostic. lgtm I'll merge.

PR #4299, opened 2018-09-26
WT-4340 the cursor caching layer can incorrectly release too many handle locks.

keithbostic:
@ddanderson, I introduced a bug in #4292, __wt_session_lock_dhandle() returns 0 with is_dead set, and we're not holding a lock in those cases.


ddanderson:
LGTM!

PR #4301, opened 2018-09-27
WT-4214 Simplify timestamp handling for timestamp abort test

sueloverso:
@nehakhatri5 I created a pull request of your branch in order to see the code differences more easily.


test/csuite/timestamp_abort/main.c
```diff
@@ -313,26 +296,33 @@ thread_run(void *arg)
...
-		 * transactions, it must be set after the prepare.
-		 */
-		if (use_ts && !use_prep) {
-			testutil_check(__wt_snprintf(tscfg, sizeof(tscfg),
-			    "commit_timestamp=%" PRIx64, stable_ts));
-			testutil_check(
-			    session->timestamp_transaction(session, tscfg));
+
+		if (use_ts) {
+			testutil_check(pthread_rwlock_wrlock(&ts_lock));
```
agorrod: Could this be a read lock? Doesn't it only need to stop the co-ordinating thread from attempting an update?

test/csuite/timestamp_abort/main.c
```diff
@@ -313,26 +296,33 @@ thread_run(void *arg)
...
+			stable_ts = __wt_atomic_addv64(&global_ts, 1);
+			/*
+			 * If not using prepared transactions set the timestamp
+			 * now before performing the operation. If we are using
+			 * prepared transactions, it must be set after the
+			 * prepare.
+			 */
+			if (!use_prep) {
```
agorrod: I'm concerned with how long the lock is held when doing prepared transactions. I can see that the test program writes the timestamp into the value, so wants to allocate one before doing work.

Because of two sessions/transactions are used when prepared is in use, we know that there will be a transaction using the timestamp at this point (in `session`), so the global value can't be updated. I think it's OK to release the lock here, as long as the prepare call on `oplog_session` happens prior to the `session->commit` call.

test/csuite/timestamp_abort/main.c
```diff
@@ -357,6 +347,7 @@ thread_run(void *arg)
 		data.data = obuf;
 		cur_oplog->set_value(cur_oplog, &data);
 		testutil_check(cur_oplog->insert(cur_oplog));
+
```
agorrod: Spurious whitespace change

test/csuite/timestamp_abort/main.c
```diff
@@ -313,26 +296,33 @@ thread_run(void *arg)
...
-		 */
-		if (use_ts && !use_prep) {
-			testutil_check(__wt_snprintf(tscfg, sizeof(tscfg),
-			    "commit_timestamp=%" PRIx64, stable_ts));
-			testutil_check(
-			    session->timestamp_transaction(session, tscfg));
+
+		if (use_ts) {
+			testutil_check(pthread_rwlock_wrlock(&ts_lock));
+			stable_ts = __wt_atomic_addv64(&global_ts, 1);
```
agorrod: Please rename this variable to something other than `stable_ts` - `active_ts` would be better.

test/csuite/timestamp_abort/main.c
nehakhatri5: Releasing the lock here causes a test failure : 
`[1538524024:257641][16257:0x7f945b7f6700], WT_SESSION.prepare_transaction: __wt_txn_parse_prepare_timestamp, 863: prepare timestamp 7dc0 is older than the oldest timestamp 7f87 : Invalid argument
test_timestamp_abort: FAILED: thread_run/361: session->prepare_transaction( session, tscfg): Invalid argument`
I had tried this during initial implemetation, hence moved the lock release after `oplog_session`'s commit for prepared txns.

agorrod: Even if you re-order the prepare and commit further down?

sueloverso: The reason is that if this thread allocates the timestamp earlier (with the lock even) and gets N, and some other thread commits N+1 and then the timestamp-coordinator thread calls `query_timestamp` it doesn't know anything about N and N+1 is the last-committed. So oldest/stable get updated to N+1 and when this thread starts using N it gets that error.

test/csuite/timestamp_abort/main.c
```diff
@@ -311,28 +298,39 @@ thread_run(void *arg)
...
+			    sizeof(tscfg), "commit_timestamp=%" PRIx64,
+			    active_ts));
+			/*
+			 * Set the transaction's timestamp now before performing
+			 * the operation. If we are using prepared transactions,
+			 * set the timestamp for the oplog session. The
+			 * collection session in that case would also use this
+			 * timestamp.
+			 */
+			if (use_prep)
```
agorrod: I think this code would be clearer if the prepare path used the same session for the non-prepared transaction as the non prepare path. Then there wouldn't need to be an `if` condition here, and I feel as though other parts of this function would simplify as well. @nehakhatri5  could you try making that change?

PR #4302, opened 2018-09-27
WT-4164 Evict clean internal pages where possible when there is cache pressure

keithbostic:
@michaelcahill, this change takes the in-memory page maximum size into account when sizing the cache, guaranteeing every worker thread can pin two maximum-sized pages into memory without blowing out the cache. This change significantly increases the memory that format will allocate to a cache in the presence of larger numbers of threads and so has an outsized impact on our small-cache testing.

On the other hand, this should resolve the category of small-cache problems where we're simply filling the cache.

I don't like it, but I like playing whack-a-mole even less, and I think that's where we've ended up.

### DIFF
--- a/test/format/config.c
+++ b/test/format/config.c
@@ -296,8 +296,7 @@ config_cache(void)
 	 * are lots of threads, it grows the cache to N megabytes per thread.
 	 */
 	v = 2 * g.c_threads * MEGABYTE(g.c_memory_page_max);
-	g.c_cache = WT_MAX(g.c_cache, (v + (WT_MEGABYTE - 1)) / WT_MEGABYTE);
-	v = ((g.c_cache * WT_MEGABYTE) / 10) * 4;
+	v = (v / 4) * 10;
 	g.c_cache = WT_MAX(g.c_cache, (v + (WT_MEGABYTE - 1)) / WT_MEGABYTE);
 
 	/*
--- a/test/format/config.c
+++ b/test/format/config.c
@@ -260,12 +260,13 @@ config_setup(void)
 static void
 config_cache(void)
 {
-	uint32_t max_dirty_bytes, required;
+	uint32_t required, v;
 
 	/* Page sizes are powers-of-two for bad historic reasons. */
 	g.intl_page_max = 1U << g.c_intl_page_max;
 	g.leaf_page_max = 1U << g.c_leaf_page_max;
 
+	/* Check if a minimum cache size has been specified. */
 	if (config_is_perm("cache")) {
 		if (config_is_perm("cache_minimum") &&
 		    g.c_cache_minimum != 0 && g.c_cache < g.c_cache_minimum)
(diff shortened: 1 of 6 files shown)


keithbostic:
autobuild: retest this please

test/format/config.c
```diff
@@ -275,36 +276,29 @@ config_cache(void)
...
-		max_dirty_bytes = ((g.c_cache * WT_MEGABYTE) / 10) * 4;
-		if (SIZE_ADJUSTMENT * g.c_threads *
-		    (g.intl_page_max + g.leaf_page_max) <= max_dirty_bytes)
-			break;
-		++g.c_cache;
-	}
+	v = 2 * g.c_threads * MEGABYTE(g.c_memory_page_max);
+	g.c_cache = WT_MAX(g.c_cache, (v + (WT_MEGABYTE - 1)) / WT_MEGABYTE);
+	v = ((g.c_cache * WT_MEGABYTE) / 10) * 4;
+	g.c_cache = WT_MAX(g.c_cache, (v + (WT_MEGABYTE - 1)) / WT_MEGABYTE);
```
michaelcahill: I don't understand the last two lines: don't they boil down to `cache = WT_MAX(cache, 0.4 * cache)` -- i.e., it's always going to be smaller than the first calculation, so won't have any effect?

test/format/config.h
```diff
@@ -250,6 +250,10 @@ static CONFIG c[] = {
 	  "the number of LSM worker threads",
 	  0x0, 3, 4, 20, &g.c_lsm_worker_threads, NULL },
 
+	{ "memory_page_max",
+	  "maximum size of in-memory pages",
+	  0x0, 1, 10, 128, &g.c_memory_page_max, NULL },
```
michaelcahill: I expected to see a `CONFIG_APPEND` line in `wts.c` that uses this value: did I miss it?

test/format/config.c
keithbostic: Yes, I did that calculation incorrectly, I've pushed a change in 30cb00f that I think gets it right.

We talked about this briefly: I recall @agorrod made that change, configuring the total number of dirty pages by all the threads to never exceed 40% of the cache. Maybe he remembers his thinking on this.

Regardless, this change as written, is going to significantly increase the size of pretty much every cache we run with. I think I'm OK with that, given the relative maturity of WiredTiger and the kinds of debugging we're doing (vs. the kind of debugging we were doing a few years ago).

test/format/config.h
keithbostic: You're correct, I missed that, pushed in b077abf.

test/format/config.c
agorrod: I still don't really understand the adjustment to `v` - it's doing a calculation based on threads and maximum page sizes, then multiplying that by 2.5. Isn't that the same as basing the calculation off 5 pages? Is what we want here to actually use 3 - 2 leaf pages and an internal page? That assumes a tree depth of 1 (plus root level), but I think it's reasonable.

Oh - I see. It's adjusting so that those pages can be dirty. Could we instead use 3 here, and set a higher `eviction_dirty_trigger` like we already do for LSM trees? Do we expect test/format to meaningfully want or test different clean/dirty targets?

Or use a higher constant multiplier? The dirty trigger is a hard limit, so using 40% seems a bit weird (I doubt it was a hard limit when I made the change originally, or maybe I just didn't think hard about it).

agorrod:
Thanks @keithbostic and @michaelcahill lgtm

PR #4303, opened 2018-09-27
WT-4328 Use an internal session handle for schema operations in a txn.

sueloverso:
Not yet ready for review.

### DIFF
--- a/src/schema/schema_alter.c
+++ b/src/schema/schema_alter.c
@@ -254,10 +254,14 @@ __wt_schema_alter(WT_SESSION_IMPL *session,
 	WT_SESSION_IMPL *int_session;
 
 	WT_RET(__wt_schema_internal_session(session, &int_session));
-	wt_session = &int_session->iface;
+	if (int_session != session)
+		wt_session = &int_session->iface;
+	else
+		wt_session = NULL;
 	WT_ERR(__wt_meta_track_on(int_session));
 	ret = __schema_alter(int_session, uri, newcfg);
 	WT_TRET(__wt_meta_track_off(int_session, true, ret != 0));
-err:	WT_TRET(wt_session->close(wt_session, NULL));
+err:	if (wt_session != NULL)
+		WT_TRET(wt_session->close(wt_session, NULL));
 	return (ret);
 }
--- a/src/schema/schema_alter.c
+++ b/src/schema/schema_alter.c
@@ -250,18 +250,12 @@ __wt_schema_alter(WT_SESSION_IMPL *session,
     const char *uri, const char *newcfg[])
 {
 	WT_DECL_RET;
-	WT_SESSION *wt_session;
 	WT_SESSION_IMPL *int_session;
 
 	WT_RET(__wt_schema_internal_session(session, &int_session));
-	if (int_session != session)
-		wt_session = &int_session->iface;
-	else
-		wt_session = NULL;
 	WT_ERR(__wt_meta_track_on(int_session));
 	ret = __schema_alter(int_session, uri, newcfg);
 	WT_TRET(__wt_meta_track_off(int_session, true, ret != 0));
-err:	if (wt_session != NULL)
-		WT_TRET(wt_session->close(wt_session, NULL));
+err:	WT_TRET(__wt_schema_session_release(session, int_session));
 	return (ret);
 }
--- a/src/schema/schema_alter.c
+++ b/src/schema/schema_alter.c
@@ -250,9 +250,18 @@ __wt_schema_alter(WT_SESSION_IMPL *session,
     const char *uri, const char *newcfg[])
 {
 	WT_DECL_RET;
+	WT_SESSION *wt_session;
+	WT_SESSION_IMPL *int_session;
 
-	WT_RET(__wt_meta_track_on(session));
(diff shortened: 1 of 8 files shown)


src/schema/schema_alter.c
```diff
@@ -250,9 +250,18 @@ __wt_schema_alter(WT_SESSION_IMPL *session,
...
+	WT_RET(__wt_schema_internal_session(session, &int_session));
+	if (int_session != session)
+		wt_session = &int_session->iface;
+	else
+		wt_session = NULL;
+	WT_ERR(__wt_meta_track_on(int_session));
+	ret = __schema_alter(int_session, uri, newcfg);
+	WT_TRET(__wt_meta_track_off(int_session, true, ret != 0));
+err:	if (wt_session != NULL)
+		WT_TRET(wt_session->close(wt_session, NULL));
```
michaelcahill: Consider a `__wt_schema_session_release(session, int_session)` function that does the close if the two session pointers differ.  I think that would simplify all of the calling code and avoid the repeated `wt_session` dance.

sueloverso: That is a good cleanup. I'll make that change.

ddanderson:
Hi @sueloverso, the changes make sense to me. I just ran the directio test 10 times under valgrind, and all success. Not sure if you wanted @michaelcahill to take a second look?

src/schema/schema_util.c
```diff
@@ -62,6 +62,48 @@ __wt_schema_get_source(WT_SESSION_IMPL *session, const char *name)
...
+	 * schema operations are not logged and buffered with any log records
+	 * in the transaction.  The new session inherits its flags from the
+	 * original.
+	 */
+	*int_sessionp = session;
+	if (F_ISSET(&session->txn, WT_TXN_RUNNING)) {
+		/* We should not have a schema txn running now. */
+		WT_ASSERT(session, !F_ISSET(session, WT_SESSION_SCHEMA_TXN));
+		WT_RET(__wt_open_internal_session(S2C(session), "schema",
+		    true, session->flags, int_sessionp));
```
michaelcahill: I didn't understand how `WT_TXN_RUNNING` could be set on a transaction in a new session, so I removed the code that cleared it.  (I first ran the test suite asserting that it wasn't set).  Is there some case I'm missing?

src/schema/schema_util.c
```diff
@@ -62,6 +62,48 @@ __wt_schema_get_source(WT_SESSION_IMPL *session, const char *name)
...
+int
+__wt_schema_session_release(
+    WT_SESSION_IMPL *session, WT_SESSION_IMPL *int_session)
+{
+	WT_SESSION *wt_session;
+
+	if (session != int_session) {
+		wt_session = &int_session->iface;
+		WT_RET(wt_session->close(wt_session, NULL));
+	}
```
michaelcahill: Variables don't have to be used in every block, so the `WT_UNUSED` that was here shouldn't be necessary.

src/txn/txn.c
```diff
@@ -1003,7 +1003,8 @@ __wt_txn_prepare(WT_SESSION_IMPL *session, const char *cfg[])
 		WT_ASSERT(session, S2C(session)->cache->las_fileid == 0 ||
 		    !F_ISSET(op->btree, WT_BTREE_LOOKASIDE));
 
-		/* Metadata updates are never prepared. */
+		/* Metadata updates should never be prepared. */
+		WT_ASSERT(session, !WT_IS_METADATA(op->btree->dhandle));
 		if (WT_IS_METADATA(op->btree->dhandle))
 			continue;
```
michaelcahill: Ideally we'd remove the `if ... continue;` here as well but I chickened out.

src/schema/schema_alter.c
```diff
@@ -250,9 +250,12 @@ __wt_schema_alter(WT_SESSION_IMPL *session,
     const char *uri, const char *newcfg[])
 {
 	WT_DECL_RET;
+	WT_SESSION_IMPL *int_session;
 
-	WT_RET(__wt_meta_track_on(session));
-	ret = __schema_alter(session, uri, newcfg);
-	WT_TRET(__wt_meta_track_off(session, true, ret != 0));
+	WT_RET(__wt_schema_internal_session(session, &int_session));
+	WT_ERR(__wt_meta_track_on(int_session));
```
agorrod: Why does alter need to use meta tracking and other operations don't? A comment would be useful since this is different to other usage patterns.

src/schema/schema_util.c
sueloverso: No, I was losing a level of indirection! The session flags are set unconditionally from the original session and I was applying that in my head to the txn flags in the session too. Thanks for catching that.

src/schema/schema_alter.c
sueloverso: The new session code didn't change/add the meta_tracking usage. It only changed a `WT_RET` or `WT_TRET` into a `WT_ERR` so we could release the session.  All the other operations do tracking, just in other functions.

sueloverso:
I thought about but didn't change any docs because we're not changing the way any users interact with the system and this should be invisible to the user. Everything felt more confusing to me. I'm going to merge this. If you feel strongly about documenting something here, we can open a follow on ticket.

PR #4304, opened 2018-09-28
WT-4224 statistics for long running prepared transactions.

bvpvamsikrishna:
Added initial set of statistics.
  1.  Number of active prepared updates ( represents the number of updates unlined from txn op)
   2. Number of prepared updates resolved ( represents the number of updates resolved through indirect references).
   3. Number of prepared updates added to lookaside.


dist/stat_data.py
```diff
@@ -264,12 +264,15 @@ def __init__(self, name, desc, flags=''):
...
     CacheStat('cache_lookaside_insert', 'cache overflow table insert calls'),
+    CacheStat('cache_lookaside_prepare_insert', 'prepared updates added to cache overflow'),
     CacheStat('cache_lookaside_remove', 'cache overflow table remove calls'),
     CacheStat('cache_lookaside_score', 'cache overflow score', 'no_clear,no_scale'),
     CacheStat('cache_overhead', 'percentage overhead', 'no_clear,no_scale'),
     CacheStat('cache_pages_dirty', 'tracked dirty pages in the cache', 'no_clear,no_scale'),
     CacheStat('cache_pages_inuse', 'pages currently held in the cache', 'no_clear,no_scale'),
     CacheStat('cache_pages_requested', 'pages requested from the cache'),
+    CacheStat('cache_prepared_updates_active', 'prepared updates active in the cache'),
+    CacheStat('cache_prepared_updates_resolved', 'prepared updates resolved in the cache'),
```
michaelcahill: A few comments:

* I'd probably expect these two, and maybe the lookaside stat, to be in the `TxnStat` section.  Since prepared updates don't pin pages in cache, that makes more sense to me than grouping them with the cache stats.
* The "updates active" only counts updates added to cache: it is never decremented.  Is that expected?
* As far as I can tell, "resolved" doesn't include rolled back.  Either add rollbacks in here or add a new stat.

What I think triage engineers would want to know is "how many updates are currently in the prepared state" -- if we can have a single stat for that, which goes up and down, that would probably be better than having to combine multiple stats to figure it out.

src/cache/cache_las.c
```diff
@@ -755,6 +755,9 @@ __wt_las_insert_block(WT_CURSOR *cursor,
 				    &las_timestamp, upd->prepare_state,
 				    upd->type, &las_value);
 
+			if (upd->prepare_state == WT_PREPARE_INPROGRESS)
```
sulabhM: I think we should increment "txn_prepared_updates_in_lookaside" after a successful cursor->update call. If update() fails for some reason, we shouldn't increment this value.

dist/stat_data.py
```diff
@@ -526,6 +526,9 @@ def __init__(self, name, desc, flags=''):
     TxnStat('txn_prepare_active', 'prepared transactions currently active'),
     TxnStat('txn_prepare_commit', 'prepared transactions committed'),
     TxnStat('txn_prepare_rollback', 'prepared transactions rolled back'),
+    TxnStat('txn_prepared_updates_active', 'prepared active updates'),
```
sulabhM: "active updates" suggest that this many updates are prepared at the moment, whereas this counter measures how many updates we have marked as prepared until now (a commit / rollback doesnt decrease this count so these are not active at the moment). Can we call it something on the lines of :
```TxnStat('txn_prepared_updates', 'prepared updates count')

File: dist/stat_data.py [hunk 4]
```diff
@@ -526,6 +526,9 @@ def __init__(self, name, desc, flags=''):
...
+    TxnStat('txn_prepared_updates_in_lookaside', 'prepared updates in cache overflow'),
+    TxnStat('txn_prepared_updates_resolved', 'prepared resolved updates'),
```
sulabhM: suggest renaming description to "prepared updates resolved"

sulabhM: Do you think we need a separate commited/rollbacked rather than single resolved ?

src/cache/cache_las.c
sulabhM: Is the idea that WT-4346 will take care of decrementing this counter when prepared updates get removed ?
At the moment this counter doesn't get decremented, which means it is rather a count of updates that have been put in the lookaside till now, and not a count of updates that exist in the lookaside at the moment.

dist/stat_data.py
bvpvamsikrishna: We need whether prepared updates are resolved or not, to check whether we missed to resolve any prepared update, not interested in mechanism (commit/rollback) resolution.

src/cache/cache_las.c
bvpvamsikrishna: Currently there is no statistics for counting the removals from lookaside, as we need to get the value of the record to examine the prepared flag, which will have some performance impact. my opinion is prepared_updates_lookaside_inserts will tell whether prepared updates are added to cache overflow mechanism or not which some what represents that life span of prepared updates. measuring removals will be useful to test our changes correctness but will not represent much user scenario diagnostics.

agorrod:
Thanks @bvpvamsikrishna lgtm.

PR #4305, opened 2018-09-28
WT-4315 Skip timestamp order check in rollback to stable

nehakhatri5:
Unless a table is configured with assert=(commit_timestamp=XX)
skip validating the timestamp ordering during rollback to stable.

### DIFF
--- a/src/txn/txn_rollback_to_stable.c
+++ b/src/txn/txn_rollback_to_stable.c
@@ -104,10 +104,6 @@ __txn_abort_newer_update(WT_SESSION_IMPL *session,
     WT_UPDATE *first_upd, wt_timestamp_t *rollback_timestamp)
 {
 	WT_UPDATE *upd;
-	bool skip_timestamp_checks;
-
-	skip_timestamp_checks = !FLD_ISSET(S2BT(session)->assert_flags,
-	    WT_ASSERT_COMMIT_TS_ALWAYS | WT_ASSERT_COMMIT_TS_KEYS);
 
 	for (upd = first_upd; upd != NULL; upd = upd->next) {
 		/*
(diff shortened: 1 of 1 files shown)


src/txn/txn_rollback_to_stable.c
```diff
@@ -116,23 +112,28 @@ __txn_abort_newer_update(WT_SESSION_IMPL *session,
...
 			 * If any updates are aborted, all newer updates
 			 * better be aborted as well.
+			 *
+			 * Validate timestamp ordering only if configured.
 			 */
-			WT_ASSERT(session, upd == first_upd);
+			WT_ASSERT(session,
+			    !FLD_ISSET(S2BT(session)->assert_flags,
+			    WT_ASSERT_COMMIT_TS_ALWAYS |
+			    WT_ASSERT_COMMIT_TS_KEYS) ||
```
michaelcahill: After looking more closely at `__txn_commit_timestamp_validate`, specifically https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn.c#L611-L615, I think we should check for `WT_ASSERT_COMMIT_TS_KEYS` here.

src/txn/txn_rollback_to_stable.c
```diff
@@ -116,23 +112,28 @@ __txn_abort_newer_update(WT_SESSION_IMPL *session,
...
 		    rollback_timestamp, &upd->timestamp) < 0) {
-			upd->txnid = WT_TXN_ABORTED;
-			WT_STAT_CONN_INCR(session, txn_rollback_upd_aborted);
-			__wt_timestamp_set_zero(&upd->timestamp);
-
 			/*
 			 * If any updates are aborted, all newer updates
 			 * better be aborted as well.
+			 *
+			 * Validate timestamp ordering only if configured.
```
michaelcahill: Please explain the connection between this check and timestamp validation on commit.  In other words, we can only rely on timestamps being ordered here if it is enforced during commit.

src/txn/txn_rollback_to_stable.c
nehakhatri5: Agreed. I had this in my head but was not sure.

michaelcahill:
Thanks @nehakhatri5 -- lgtm.  I'll merge.

PR #4306, opened 2018-09-28
WT-4341 Support million collection test in Evergreen

lukech:
This part of the change is about Evergreen yaml configuration in order to support million collection testing in Evergreen CI system. 

The solution is to put the new `million-collect-test` task into a new `large-scale-test' build variant in the existing `WiredTiger` project.

The related script changes for `largescale/run-million-collection-test.sh` within `mongo-tests` repo will be covered by BUILD-6194.

### DIFF
--- a/test/mciproject.yml
+++ b/test/mciproject.yml
@@ -16,6 +16,11 @@ functions:
         remote_file: wiredtiger/${build_variant}/${revision}/binaries/${build_id}.tgz
         bucket: build_external
         extract_to: wiredtiger
+  "fetch mongo-tests repo" :
+    command: shell.exec
+    params:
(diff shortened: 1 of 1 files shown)


lukech:
@agorrod That's a good question. We should add `batchtime` setting for the new task, but won't be able to tell the effect until the changes are merged. It's not obvious on the wiredtiger/wiredtiger repo side (webhook configuration), need further investigation/check on the evergreen side how they handle the REST request. Will follow up.

lukech:
@agorrod I pushed a commit to set `batchtime` to 1440 (1 day) for the new build variant - similar as what we did before for those WiredTiger develop variants.

PR #4307, opened 2018-09-28
WT-3633 workgen support for checkpoints and better tracking of latencies

ddanderson:
In workgen, added an internal implementation to Operations, created a
sleep operation, a checkpoint operation and a no-op operation.
Added tracking of checkpoint operations.
Track the number of operations in progress to make it easier to tell if
we are in the middle of a checkpoint.
Removed latency smoothing, it doesn't help.
Added support to the wtperf simulation script for checkpoint_threads,
checkpoint_interval and added a new (non-wtperf) option sample_file.
Added a wtperf file to track latency of operations during checkpoints.


ddanderson:
WT-4297 is the more relevant ticket number now, so closing.

PR #4308, opened 2018-09-28
WT-4297 workgen support for checkpoints and better tracking of latencies

ddanderson:
- In workgen, added an internal implementation to Operations, created a sleep operation, a checkpoint operation and a no-op operation.
- Added tracking of checkpoint operations.
- Track the number of operations in progress to make it easier to tell if we are in the middle of a checkpoint.
- Removed latency smoothing, it doesn't help.
- Added support for 50/95/99 percentile latencies
- Added support to the wtperf simulation script for checkpoint_threads, checkpoint_interval and added a new (non-wtperf) option sample_file.
- Added a wtperf file to track latency of operations during checkpoints.

Workgen previously had capability of generating the latency information as time series that could be read by t2. These changes enhance the information and expose it as an option to .wtperf files.


ddanderson:
Running the newly committed workload like this `python wtperf.py runner/checkpoint-latency.wtperf`, produces a `WT_TEST/mon.json` file that can be imported into `t2`.  
t2-2018-09-28-12-50-06
I'll continue to modify the workload as part of WT-4298/WT-3633, but I think the base work to support the analysis is ready for review.

ddanderson:
@agorrod, can you please review when you have a chance?

bench/workgen/wtperf.py
```diff
@@ -467,9 +481,10 @@ def translate_inner(self):
                 if len(line) == 0:
                     continue
                 (key, val) = self.split_assign(line)
+                # Note: 'sample_file' is not an original wtperf option.
                 if key in [ 'max_latency', 'report_file', 'report_interval',
-                            'run_time', 'sample_interval', 'sample_rate',
-                            'warmup' ]:
+                            'run_time', 'sample_file', 'sample_interval',
```
nehakhatri5: sample_stat_file or stat_file might be a more meaningful name for this parameter.

bench/workgen/workgen.cxx
```diff
@@ -1029,33 +1098,44 @@ void Operation::describe(std::ostream &os) const {
...
+        case OP_REMOVE:
+            stats.remove.ops += multiplier;
+            break;
+        case OP_SEARCH:
+            stats.read.ops += multiplier;
+            break;
+        case OP_UPDATE:
+            stats.update.ops += multiplier;
+            break;
+        case OP_CHECKPOINT:
```
nehakhatri5: `OP_CHECKPOINT` is not a table op as per definition of `is_table_op()`. Would this case ever be excercised?

bench/workgen/workgen.cxx
```diff
@@ -553,7 +568,7 @@ void ThreadRunner::op_create_all(Operation *op, size_t &keysize,
     tint_t tint;
 
     op->create_all();
-    if (op->_optype != Operation::OP_NONE) {
+    if (op->is_table_op()) {
```
nehakhatri5: Line #696 should have the same check. Throttle applies only to insert, read and update.

bench/workgen/workgen.h
```diff
@@ -276,8 +282,10 @@ struct Value {
 
 struct Operation {
     enum OpType {
-	OP_NONE, OP_INSERT, OP_REMOVE, OP_SEARCH, OP_UPDATE };
+	OP_NONE, OP_INSERT, OP_REMOVE, OP_SEARCH, OP_UPDATE,
+	OP_CHECKPOINT, OP_SLEEP, OP_NOOP };
```
sueloverso: Do these need to remain the same order? Otherwise please alphabetize.

bench/workgen/wtperf.py
```diff
@@ -76,7 +76,8 @@ def fatal_error(self, msg, errtype = 'configuration error'):
         self.error(msg)
         raise TranslateException(errtype)
 
-    supported_opt_list = [ 'close_conn', 'compression', 'compact',
+    supported_opt_list = [ 'checkpoint_interval', 'checkpoint_threads',
+                           'close_conn', 'compression', 'compact',
```
sueloverso: Alphabetize `compact` before `compression`.

bench/workgen/workgen.cxx
```diff
@@ -713,11 +728,18 @@ int ThreadRunner::op_run(Operation *op) {
         recno = op_get_key_recno(op, range, tint);
         break;
     case Operation::OP_NONE:
+    case Operation::OP_NOOP:
+        recno = 0;
+        break;
+    case Operation::OP_CHECKPOINT:
+        recno = 0;
+        track = &_stats.checkpoint;
+        break;
```
sueloverso: Alphabetize this please.

bench/workgen/workgen.cxx
```diff
@@ -981,28 +1016,62 @@ Operation& Operation::operator=(const Operation &other) {
...
+    case OP_SLEEP:
+        if (other == NULL)
+            _internal = new SleepOperationInternal();
         else
-            THROW("operation has illegal config: \"" << _config << "\"");
+            _internal = new SleepOperationInternal(
+              *(SleepOperationInternal *)other);
+        break;
+    case OP_CHECKPOINT:
```
sueloverso: Alphabetize these too.

bench/workgen/runner/checkpoint-latency.wtperf
```diff
@@ -0,0 +1,26 @@
+# A stress configuration to create checkpoints while doing a mix of inserts
```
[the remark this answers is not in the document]
ddanderson: I had put this .wtperf file in the workgen directory because it used an option (`sample_file`) that was not available with the standard wtperf.  I just pushed a better resolution that creates a `"monitor.json"` file whenever a `"monitor"` file is created (that is, when `sample_interval` is used).  Then the incompatible option can be removed, and this file can be moved into the `wtperf/runners` directory as suggested.

bench/workgen/wtperf.py
ddanderson: Removed this new option, see previous comment.

PR #4311, opened 2018-10-02
WT-4346 Remove prepared updates from lookaside on reading the page back

sulabhM:


michaelcahill:
Thanks @sulabhM -- this change lgtm.  Can you post details of the test (and stat differences after rolling back) on the ticket?

sulabhM:
The stat difference is as follows:

```
Oct 02 16:54:51 6939 . transaction: prepared transactions LAS in
Oct 02 16:54:51 6645 . transaction: prepared transactions LAS out
Oct 02 16:54:51 6660 . transaction: prepared transactions LAS out seen
```

I am counting successful updates in __wt_las_insert_block() and removes in __las_remove_block() when prepare_state is WT_PREPARE_INPROGRESS:

```
diff --git a/src/cache/cache_las.c b/src/cache/cache_las.c
index cf28027d8..419406dc0 100644
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -482,10 +482,13 @@ __las_remove_block(
        WT_CONNECTION_IMPL *conn;
        WT_DECL_RET;
        WT_ITEM las_key;
+       WT_ITEM las_timestamp, las_value;
        WT_SESSION_IMPL *session;
        WT_TXN_ISOLATION saved_isolation;
        uint64_t las_counter, las_pageid;
+       uint64_t las_txnid;
        uint32_t las_id;
+       uint8_t upd_type, prepare_state;
        bool local_txn;
 
        *remove_cntp = 0;
@@ -513,12 +516,20 @@ __las_remove_block(
            ret == 0; ret = cursor->next(cursor)) {
                WT_ERR(cursor->get_key(cursor,
                    &las_pageid, &las_id, &las_counter, &las_key));
-
+               WT_ERR(cursor->get_value(
+                   cursor, &las_txnid, &las_timestamp,
+                   &prepare_state, &upd_type, &las_value));
                /* Confirm that we have a matching record. */
-               if (las_pageid != pageid)
+               if (prepare_state == WT_PREPARE_INPROGRESS)
+                       WT_STAT_CONN_INCR(session, txn_prepare_las_out_seen);
+               if (las_pageid != pageid) {
+                       printf("BLAH\n");
                        break;
+               }
 
+               if (prepare_state == WT_PREPARE_INPROGRESS)
+                       WT_STAT_CONN_INCR(session, txn_prepare_las_out);
                ++*remove_cntp;
        }
        WT_ERR_NOTFOUND_OK(ret);
@@ -762,6 +773,8 @@ __wt_las_insert_block(WT_CURSOR *cursor,
                         * between inserts.
                         */
                        WT_ERR(cursor->update(cursor));
+                       if (upd->prepare_state == WT_PREPARE_INPROGRESS)
+                               WT_STAT_CONN_INCR(session, txn_prepare_las_in);
                        ++insert_cnt;
                } while ((upd = upd->next) != NULL);
        }
``` 

The test is test_prepare_lookaside01.py, with some modifications, it does the following:
1. For small cache size (50MB) creates a table with a small amount of initial data
2. inserts 10000 rows with large data (500 characters long string)
3. checkpoints
4. sets timestamp of 1
5. key range 1 to 12000 commits updates with new data of same size with timestamp equal to the key, ie ranging from 1 to 12000
6. in 3 sessions non-overlapping key range of 4000 each puts prepare updates for the existing data with prepare timestamp of 2
7. key range 1 to 12000 commits updates with new data of same size with timestamp equal to the key + 3, ie ranging from 4 to 12003
8. Closes 3 sessions prepared was called from, hence initiating a rollback.

In addition, I have set the test to checkpoint after setting a stable time stamp greater than any timestamp. The test then sleeps so that i can look at the statsitics.

```
--- a/test/suite/test_prepare_lookaside01.py
+++ b/test/suite/test_prepare_lookaside01.py
@@ -29,6 +29,7 @@
 from helper import copy_wiredtiger_home
 import wiredtiger, wttest
 from wtdataset import SimpleDataSet
+import time
 
 def timestamp_str(t):
     return '%x' % t
@@ -127,11 +128,15 @@ class test_prepare_lookaside01(wttest.WiredTigerTestCase):
         # Check if lookaside is working properly with prepare transactions.
         # We put prepared updates in multiple sessions so that we do not hang
         # because of cache being full with uncommitted updates.
-        # TODO: Increase the nsessions below to start testing lookaside eviction
-        # of prepared updates.
-        nsessions = 1
+        nsessions = 3
         nkeys = 4000
         self.prepare_updates(uri, ds, nrows, nsessions, nkeys)
 
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str((nsessions * nkeys) + 10))
+
+        print 'Test done: wait for sweep'
+        time.sleep(1200)
+
 if __name__ == '__main__':
     wttest.run()
```

src/btree/bt_read.c
```diff
@@ -536,24 +541,24 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
...
 	 *
 	 * There is no reason for the lookaside remove should fail, but ignore
 	 * it if for some reason it fails, we've got a valid page.
 	 *
 	 * Don't free WT_REF.page_las, there may be concurrent readers.
 	 */
 	if (final_state == WT_REF_MEM &&
-	    ref->page_las != NULL && !ref->page_las->skew_newest)
+	    ref->page_las != NULL && (prepared || !ref->page_las->skew_newest))
 		WT_IGNORE_RET(__wt_las_remove_block(
```
sulabhM: I figured out why we have entries that aren't being removed from lookaside, because we are ignoring failure of __wt_las_remove_block(). When lookaside sweep or eviction to lookaside has taken a lock, removal from lookaside here fails to obtain the lock and failure gets ignored and we never remove these entries.

To confirm the above theory, I make this function wait on lock (pass true for the last argument), my test passes and there are no prepared entries in lookaside at shutdown. Statistics for removal and insertion also seem to match.

sulabhM: @agorrod I am not sure if WT_IGNORE_RET(..) here is okay to do. I am assuming it was being done here because it isn't safe to bail out of this function anymore ?

agorrod: I don't think it's OK for that to fail any longer - we are relying on those removes succeeding now.

sulabhM:
@agorrod @bvpvamsikrishna this is ready for review.

src/btree/bt_read.c
```diff
@@ -536,26 +541,23 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
...
+	 * to be read back into cache or they will be lost).
 	 *
 	 * Don't free WT_REF.page_las, there may be concurrent readers.
 	 */
 	if (final_state == WT_REF_MEM &&
-	    ref->page_las != NULL && !ref->page_las->skew_newest)
-		WT_IGNORE_RET(__wt_las_remove_block(
-		    session, ref->page_las->las_pageid, false));
+	    ref->page_las != NULL && (prepared || !ref->page_las->skew_newest))
+		WT_ERR(__wt_las_remove_block(
```
bvpvamsikrishna: lock_wait is always passed as true, we can remove that arguement from declaration.

src/btree/bt_read.c
```diff
@@ -536,26 +541,23 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
...
 	/*
 	 * Once the page is instantiated, we no longer need the history in
 	 * lookaside.  We leave the lookaside sweep thread to do most cleanup,
-	 * but it can only remove keys that skew newest (if there are entries
-	 * in the lookaside newer than the page, they need to be read back into
-	 * cache or they will be lost).
-	 *
-	 * There is no reason for the lookaside remove should fail, but ignore
-	 * it if for some reason it fails, we've got a valid page.
+	 * but it can only remove committed updates and keys that skew newest
```
bvpvamsikrishna: Enhance the comment to indicate that we will remove if the page has prepared updates.

src/btree/bt_read.c
```diff
@@ -188,6 +188,10 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 		total_incr += incr;
 		upd->txnid = las_txnid;
 		upd->prepare_state = prepare_state;
+		// If it isn't already done, set that we found prepared updates.
```
bvpvamsikrishna: use c-style comments.

src/btree/bt_read.c
```diff
@@ -188,6 +188,10 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
+		if (preparedp &&
```
bvpvamsikrishna: is there a case where preparedp can be NULL ?

sulabhM: Yes, if a caller doesn't need and pass it as NULL. I have tried to code this function as a utility, if the caller doesn't care whether there are prepared updates or not, it can chose to pass as NULL. That said, there are not such callers at the moment.

sulabhM:
@bvpvamsikrishna I have addressed your comments, have another look at the review.

src/btree/bt_read.c
bvpvamsikrishna: I think we can remove this check. Any way it is not a big structure where caller can pass NULL and save some memory, it is just a bool, caller can always pass the pointer and ignore the return value if not bothered.

sulabhM: Is there any reason you feel like we should not check that passed variable is null ?

Doing so offers the benefit of not relying on the caller passing a variable to us.
It is also beneficial for the caller, if it doesn't need to know about the prepared state, it can just pass null, rather than going through the exercise of allocating a variable and then calling WT_UNUSED(..) on it.

I do not see any cons of doing a NULL check, while it offers the above said benefits.
__wt_log_release() for instance does something similar.

If you do not have strong reasons to remove this check I will like to keep it.

agorrod:
Thanks @bvpvamsikrishna lgtm - please merge once testing passes and @bvpvamsikrishna approves.

PR #4314, opened 2018-10-04
WT-4351 Ensure resolving prepared transactions use updates from itself

bvpvamsikrishna:
As part of indirect reference resolution in prepared transactions, the transaction operation structure currently always refer the first update in the update chain. This assumption is wrong, first update might be from a different transaction in case of lower isolation levels.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -359,16 +359,18 @@ __wt_txn_resolve_prepared_op(
 	 * case.
 	 */
 	WT_ASSERT(session, upd != NULL || txn->multi_update_count != 0);
-	if (upd == NULL)
+	if (upd == NULL && commit)
 		--txn->multi_update_count;
 #endif
 
-	op->u.op_upd = upd;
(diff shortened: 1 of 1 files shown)


bvpvamsikrishna:
@sulabhM @agorrod request your review for this pull request.
This is a bug fix in long running prepared transactions.

bvpvamsikrishna:
we are tracking the update count only commit, not for rollback, as our tracking is based on txn id, and incase of rollback, we set it to WT_TXN_ABORTED.

src/include/txn.i
```diff
@@ -359,17 +359,19 @@ __wt_txn_resolve_prepared_op(
 	 * case.
 	 */
 	WT_ASSERT(session, upd != NULL || txn->multi_update_count != 0);
-	if (upd == NULL)
```
agorrod: Why not do this check for rollback any more?

agorrod:
Thanks @bvpvamsikrishna lgtm - please merge once testing has passed.

PR #4315, opened 2018-10-04
WT-4293 WT_CURSOR.remove can lose a cursor position

keithbostic:
@michaelcahill, the simplest solution I could find was to maintain the "positioned" status of the cursor in the upper-level cursor code. That's arguably a layering violation, let me know if you see something better.

I also reworked error handling in the Btree cursor remove code, there were paths were we could lose errors.


michaelcahill:
Thanks @keithbostic, I don't see a better solution and this change lgtm.  I'll merge.

PR #4317, opened 2018-10-08
WT-4343 Switch slot after releasing to avoid deadlock.

sueloverso:
Release slot lock if waiting for a new one.

Not yet ready for review. Want to run log perf tests on branch to see if there's any perf regression.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -2899,17 +2899,8 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
 	 * detects a panic.
 	 */
 	__wt_log_slot_join(session, rdup_len, flags, &myslot);
-	/*
-	 * If the addition of this record crosses the buffer boundary,
-	 * switch in a new slot.
-	 */
 	force = LF_ISSET(WT_LOG_FLUSH | WT_LOG_FSYNC);
-	ret = 0;
-	if (myslot.end_offset >= WT_LOG_SLOT_BUF_MAX ||
-	    F_ISSET(&myslot, WT_MYSLOT_UNBUFFERED) || force)
-		ret = __wt_log_slot_switch(session, &myslot, true, false, NULL);
-	if (ret == 0)
-		ret = __wt_log_fill(session, &myslot, false, record, &lsn);
+	ret = __wt_log_fill(session, &myslot, false, record, &lsn);
 	release_size = __wt_log_slot_release(&myslot, (int64_t)rdup_len);
 	/*
 	 * If we get an error we still need to do proper accounting in
@@ -2918,7 +2909,15 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
 	 */
 	if (ret != 0)
 		myslot.slot->slot_error = ret;
-	WT_ASSERT(session, ret == 0);
+	/*
+	 * If the addition of this record crosses the buffer boundary,
+	 * switch in a new slot. We have to do it after we release the
+	 * slot so that we cannot deadlock ourselves if all slots are in use.
+	 */
+	if (myslot.end_offset >= WT_LOG_SLOT_BUF_MAX ||
+	    F_ISSET(&myslot, WT_MYSLOT_UNBUFFERED) || force)
+		WT_TRET(__wt_log_slot_switch(
+		session, &myslot, true, false, NULL));
 	if (WT_LOG_SLOT_DONE(release_size)) {
 		WT_ERR(__wt_log_release(session, myslot.slot, &free_slot));
 		if (free_slot)
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -1230,8 +1230,15 @@ __log_newfile(WT_SESSION_IMPL *session, bool conn_open, bool *created)
 	WT_ASSERT(session, F_ISSET(session, WT_SESSION_LOCKED_SLOT));
 	for (yield_cnt = 0; log->log_close_fh != NULL;) {
 		WT_STAT_CONN_INCR(session, log_close_yields);
+		/*
+		 * Processing slots will conditionally signal the file close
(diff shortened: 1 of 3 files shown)


sueloverso:
@agorrod this is ready for review. I have it running a full suite of the random_abort and timestamp_abort tests over the weekend since there are places we now release and reacquire a lock. My reasoning thinks it is okay but I want to shake that out with testing. The normal logging perf tests look good with the latest changes, on par with a few better perf runs. I have kicked off a perf-long run to see what the multi-hour tests with logging show.

agorrod:
@sueloverso sorry I've let this one sit, I'll do a proper review tomorrow and get it moving ahead again.

agorrod:
@sueloverso sorry I've let this one sit, I'll do a proper review tomorrow and get it moving ahead again.

sueloverso:
@agorrod although we have seen no failures nor performance degradation, the long tests show more times of latency, but not longer latency. http://build.wiredtiger.com:8080/job/wiredtiger-perf-long/plot/. That seems to be the only side effect of this change. 

The normal logging perf tests show no performance impact. There is a minor change that the gap between min and max throughput widened some but overall throughput stayed the same. http://build.wiredtiger.com:8080/job/wiredtiger-perf-log-consolidated/plot/

src/log/log.c
```diff
@@ -1230,8 +1230,18 @@ __log_newfile(WT_SESSION_IMPL *session, bool conn_open, bool *created)
...
+		 * Processing slots will conditionally signal the file close
+		 * server thread. But if we've tried a while, signal the
+		 * thread directly here.
+		 */
 		__wt_log_wrlsn(session, NULL);
-		if (++yield_cnt > 10000)
+		if (++yield_cnt % WT_THOUSAND == 0) {
+			__wt_spin_unlock(session, &log->log_slot_lock);
+			__wt_cond_signal(session, conn->log_file_cond);
+			__wt_spin_lock(session, &log->log_slot_lock);
```
agorrod: I don't understand exactly what the `log_slot_lock` is protecting - so I'm not confident that releasing and re-acquiring is OK. I spent some time trying to figure it out without success - could you explain?

src/log/log.c
```diff
@@ -2918,12 +2918,20 @@ __log_write_internal(WT_SESSION_IMPL *session, WT_ITEM *record, WT_LSN *lsnp,
...
-	WT_ASSERT(session, ret == 0);
+	/*
+	 * If the addition of this record crosses the buffer boundary,
+	 * switch in a new slot. We have to do it after we release the
+	 * slot so that we cannot deadlock ourselves if all slots are in use.
+	 */
+	if (myslot.end_offset >= WT_LOG_SLOT_BUF_MAX ||
+	    F_ISSET(&myslot, WT_MYSLOT_UNBUFFERED) || force)
+		WT_TRET(__wt_log_slot_switch(
+		session, &myslot, true, false, NULL));
```
agorrod: Is this going in the opposite direction to https://jira.mongodb.org/browse/WT-4095 and the previous attempt in https://jira.mongodb.org/browse/WT-4058? I know that correctness definitely trumps performance, but I'd like to understand how these fit together. Will the ordering change here mean that we don't make a new slot available until after the I/O is performed?
Ah, I think the answer is "no" it's `__wt_log_release` that does a write so the switch still happens before that, right?

src/log/log_slot.c
```diff
@@ -264,21 +267,30 @@ __log_slot_new(WT_SESSION_IMPL *session)
...
 		 */
 		WT_STAT_CONN_INCR(session, log_slot_no_free_slots);
 		__wt_cond_signal(session, conn->log_wrlsn_cond);
-		__wt_yield();
+		++wait_count;
+		if (wait_count < WT_THOUSAND)
+			__wt_yield();
+		else {
+			__wt_spin_unlock(session, &log->log_slot_lock);
+			__wt_sleep(0, WT_THOUSAND);
```
agorrod: It seems aggressive to sleep after a thousand spins.

src/log/log_slot.c
```diff
@@ -264,21 +267,30 @@ __log_slot_new(WT_SESSION_IMPL *session)
...
+			__wt_spin_lock(session, &log->log_slot_lock);
+			goto retry;
```
agorrod: Using the goto means the wait_count will be reset after each 1000 spins - I don't see how it could ever reach a million now - which is the condition to timeout. Could this be refactored to avoid a goto?

src/log/log.c
sueloverso: Mostly the code movement here will cause joining threads to wait for the memory copy if a slot is full. So the thread that fills the slot buffer will do its copy into the memory buffer before setting up the new slot. The answer is mostly "no". The `log_fill` function in the common case will do the memory copy. The uncommon case involves I/O (write(2) to OS only) for unbuffered log records.

src/log/log_slot.c
sueloverso: I will refactor and I will yield instead of sleep to try to mitigate some of this.

sueloverso:
Thanks for modifying Jenkins. I'll make the review changes and then point the perf tests to the branch to measure the impact.

src/log/log.c
sueloverso: Effectively the lock protects the `log->active_slot` field. The owner of the lock is the one who is switching slots.

sueloverso:
@agorrod I have addressed the review comments. The code changes in `log_slot_new` are now very tiny. I realized we could just unlock if we cannot find a slot at all and many changes and code paths go away. I have changed the Jenkins perf jobs to point to this branch for the weekend. We'll see what it shows in the new many-threads plot.

sueloverso:
@agorrod Let's not merge this yet. I'm going to run some experiments with the failing condition and only the changes to the locking to see if those changes alone fix the problem, and probably without introducing the performance variances.

agorrod:
We took another version of this branch - so I'm going to close this one.

PR #4318, opened 2018-10-09
WT-4355 transaction rollback crash for prepared updates.

bvpvamsikrishna:
multi_update_count is tracked only for commit, but not for rollback. Hence moved the assertion to check only in case of commit.


sulabhM:
@bvpvamsikrishna , I am going to merge this in, the change is simple and in debug code, doesn't need a second reviewer.

PR #4319, opened 2018-10-11
WT-4358 operating on closed/evicted tree

keithbostic:
@agorrod, this is the test program I've been using to try and reproduce the handle-locking issues.

It pushes the handle code pretty hard, but it's not finding any problems.

Anyway, I figured I'd go ahead and merge the test program, absent any other ideas on ways to make it push the system harder.


agorrod:
Thanks for this test @keithbostic it makes sense to me - I'm going to go ahead and merge.

agorrod:
Thanks for this test @keithbostic it makes sense to me - I'm going to go ahead and merge.

PR #4320, opened 2018-10-15
WT-4339 Revert part of a previous commit that simplified handle locking

agorrod:
There was a bug in that change that we don't fully understand yet. Also add a test enhancement that lets us see the bug quickly.

### DIFF
--- a/src/conn/conn_dhandle.c
+++ b/src/conn/conn_dhandle.c
@@ -244,12 +244,10 @@ __wt_conn_dhandle_close(
 	WT_CONNECTION_IMPL *conn;
 	WT_DATA_HANDLE *dhandle;
 	WT_DECL_RET;
-	bool close_handle, is_btree, is_mapped, no_schema_lock;
+	bool discard, is_btree, is_mapped, marked_dead, no_schema_lock;
 
 	conn = S2C(session);
 	dhandle = session->dhandle;
-	close_handle = true;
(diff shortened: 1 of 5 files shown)


agorrod:
@sueloverso I've used the ticket you own to push a change here - it's a simple revert of the second half of the change in WT-4314, since my local testing showed it was the root cause:

```
export "TESTUTIL_ENABLE_TIMING_TESTS=1"
for i in `seq 10`; do ./test/csuite/test_wt3363_checkpoint_op_races -h WT_TEST.$i & done
```

Reproduces the failure in WT-4339 in a single iteration for me, and with this revert I haven't seen a failure.

michaelcahill:
@sueloverso -- I discussed this change with @agorrod: it reverts changes in WT-4314 that were supposed to simplify / clarify `__wt_conn_dhandle_close` but obviously broke something.  I'm good with it if you are.

sueloverso:
I am good with the revert changes. I don't understand the testutil changes. Are those extra that just happened to be in your tree? Should they be part of this change? They were not part of the original WT-4314.

agorrod:
@bvpvamsikrishna it would be great if you can review this change. There would be value in applying the test changes here and ensuring you can confirm that the change here fixes the problem. I outlined the steps I used above - which were reproducing a failure within 5 minutes for me.

PR #4321, opened 2018-10-20
WT-3898 Keep prepared updates in lookaside until they are read.

michaelcahill:
We need to consider a lookaside block with prepared updates "active" until they are read back into cache (as part of commit or rollback).  Otherwise we can decide that the history is no longer relevant and transition a page to the WT_REF_MEM state, ignoring its history.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -554,11 +554,9 @@ skip_read:
 	 * Don't free WT_REF.page_las, there may be concurrent readers.
 	 */
 	if (final_state == WT_REF_MEM && ref->page_las != NULL &&
-	    (!ref->page_las->skew_newest || ref->page_las->has_prepares)) {
-		ret = __wt_las_remove_block(
-		    session, ref->page_las->las_pageid);
-		WT_ERR(ret);
-	}
+	    (!ref->page_las->skew_newest || ref->page_las->has_prepares))
+		WT_ERR(__wt_las_remove_block(
+		    session, ref->page_las->las_pageid));
 
 	WT_REF_SETSTATE(ref, final_state);
 	return (ret);
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -113,7 +113,7 @@ __las_page_instantiate_verbose(WT_SESSION_IMPL *session, uint64_t las_pageid)
  *	Instantiate lookaside update records in a recently read page.
  */
 static int
-__las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref, bool *preparedp)
+__las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 {
 	WT_CACHE *cache;
 	WT_CURSOR *cursor;
@@ -166,7 +166,6 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref, bool *preparedp)
 	__wt_readlock(session, &cache->las_sweepwalk_lock);
 	WT_PUBLISH(cache->las_reader, false);
 	locked = true;
-	*preparedp = false;
 	for (ret = __wt_las_cursor_position(cursor, las_pageid);
 	    ret == 0;
 	    ret = cursor->next(cursor)) {
@@ -189,8 +188,6 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref, bool *preparedp)
 		total_incr += incr;
 		upd->txnid = las_txnid;
 		upd->prepare_state = prepare_state;
-		if (prepare_state == WT_PREPARE_INPROGRESS)
-			*preparedp = true;
 #ifdef HAVE_TIMESTAMPS
 		WT_ASSERT(session, las_timestamp.size == WT_TIMESTAMP_SIZE);
 		memcpy(&upd->timestamp, las_timestamp.data, las_timestamp.size);
@@ -284,6 +281,7 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref, bool *preparedp)
 		FLD_SET(page->modify->restore_state, WT_PAGE_RS_LOOKASIDE);
 
 		if (ref->page_las->skew_newest &&
+		    !ref->page_las->has_prepares &&
 		    !S2C(session)->txn_global.has_stable_timestamp &&
 		    __wt_txn_visible_all(session, ref->page_las->unstable_txn,
 		    WT_TIMESTAMP_NULL(&ref->page_las->unstable_timestamp))) {
@@ -378,7 +376,7 @@ __evict_force_check(WT_SESSION_IMPL *session, WT_REF *ref)
  */
 static inline int
 __page_read_lookaside(WT_SESSION_IMPL *session, WT_REF *ref,
-    uint32_t previous_state, uint32_t *final_statep, bool *preparedp)
+    uint32_t previous_state, uint32_t *final_statep)
 {
 	/*
 	 * Reading a lookaside ref for the first time, and not requiring the
@@ -390,8 +388,8 @@ __page_read_lookaside(WT_SESSION_IMPL *session, WT_REF *ref,
 			WT_STAT_CONN_INCR(
 			    session, cache_read_lookaside_skipped);
 			ref->page_las->eviction_to_lookaside = true;
-			*final_statep = WT_REF_LIMBO;
 		}
+		*final_statep = WT_REF_LIMBO;
 		return (0);
 	}
 
@@ -403,7 +401,7 @@ __page_read_lookaside(WT_SESSION_IMPL *session, WT_REF *ref,
 			    cache_read_lookaside_delay_checkpoint);
 	}
 
-	WT_RET(__las_page_instantiate(session, ref, preparedp));
+	WT_RET(__las_page_instantiate(session, ref));
 	ref->page_las->eviction_to_lookaside = false;
 	return (0);
 }
@@ -422,7 +420,7 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
 	uint64_t time_start, time_stop;
 	uint32_t page_flags, final_state, new_state, previous_state;
 	const uint8_t *addr;
-	bool prepared, timer;
+	bool timer;
 
 	time_start = time_stop = 0;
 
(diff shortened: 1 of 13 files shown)


michaelcahill:
I have pushed a branch showing my working -- I expect the final change to be considerably smaller, but hope this version helps explain how I got here.

First, I found some places where we could get a `WT_ROLLBACK` error due to cache pressure while restoring a page with lookaside history.  That fixed some of the test cases that were getting rollback errors during commit/rollback, but not others.

Then I added a log of state transitions to the `WT_REF` structure so I could see where things were going wrong: a page with history evicted in lookaside was ending up in the `WT_REF_MEM` state without the history loaded into cache.  This was happening as part of `__evict_page_clean_update` (first to `WT_REF_DISK` so on the next read `ref->page_las` was ignored).

src/btree/bt_read.c
```diff
@@ -390,8 +388,8 @@ __page_read_lookaside(WT_SESSION_IMPL *session, WT_REF *ref,
 			WT_STAT_CONN_INCR(
 			    session, cache_read_lookaside_skipped);
 			ref->page_las->eviction_to_lookaside = true;
-			*final_statep = WT_REF_LIMBO;
 		}
+		*final_statep = WT_REF_LIMBO;
```
michaelcahill: @agorrod this seems like a straight bug: if we got here with a limbo page and decided not to read the history, it would transition to `WT_REF_MEM`.  Maybe this is unlikely because of the higher level checks on the state in `__wt_page_in_func` but I don't think it's impossible to be handling a page that is already `WT_REF_LIMBO` here, given the right interleaving.

src/cache/cache_las.c
```diff
@@ -829,26 +830,23 @@ __wt_las_cursor_position(WT_CURSOR *cursor, uint64_t pageid)
...
+		 * we have a key lower than we expect.
+		 *
+		 * There may be no block of lookaside entries if they
+		 * have been removed by
+		 * WT_CONNECTION::rollback_to_stable.
+		 */
+		WT_RET(cursor->get_key(cursor,
+		    &las_pageid, &las_id, &las_counter, &las_key));
+		if (las_pageid >= pageid)
+			return (0);
```
michaelcahill: This is just paranoia: no need to keep it.

src/include/btree.i
```diff
@@ -1189,7 +1189,7 @@ __wt_page_las_active(WT_SESSION_IMPL *session, WT_REF *ref)
 
 	if ((page_las = ref->page_las) == NULL)
 		return (false);
-	if (!page_las->skew_newest)
+	if (!page_las->skew_newest || page_las->has_prepares)
```
michaelcahill: This is one of the core fixes.

src/include/txn.i
```diff
@@ -1339,7 +1339,7 @@ __wt_txn_am_oldest(WT_SESSION_IMPL *session)
 	txn = &session->txn;
 	txn_global = &conn->txn_global;
 
-	if (txn->id == WT_TXN_NONE)
+	if (txn->id == WT_TXN_NONE || F_ISSET(txn, WT_TXN_PREPARE))
```
michaelcahill: We shouldn't roll back a prepared transaction here. We set `WT_SESSION_IGNORE_CACHE_SIZE` in a lot of places that deal with restoring from lookaside, but I saw cases where we got here with `WT_TXN_PREPARE` set.

src/include/btmem.h
```diff
@@ -856,6 +857,24 @@ struct __wt_ref {
...
+#define	WT_REF_SETSTATE(ref, s) do {					\
+	ref->hist[ref->histoff].session = session;			\
+	ref->hist[ref->histoff].name = session->name;			\
+	ref->hist[ref->histoff].file = __FILE__;			\
+	ref->hist[ref->histoff].line = __LINE__;			\
+	ref->hist[ref->histoff].state = s;				\
+	ref->histoff = (ref->histoff + 1) % (int)WT_ELEMENTS(ref->hist);\
+	WT_PUBLISH(ref->state, s);					\
+} while (0)
+
```
keithbostic: I've done something like this a few times as well -- maybe we could #ifdef this on something and leave it in place? We could probably crank down the size by a bit, pass session as an argument (if session & session.name are sufficiently useful to keep permanently).

src/cache/cache_las.c
```diff
@@ -403,7 +403,7 @@ __wt_las_page_skip_locked(WT_SESSION_IMPL *session, WT_REF *ref)
 	 * We also need to instantiate a lookaside page if this is an update
 	 * operation in progress.
 	 */
-	if (F_ISSET(txn, WT_TXN_UPDATE))
+	if (F_ISSET(txn, WT_TXN_PREPARE | WT_TXN_UPDATE))
```
keithbostic: Comment probably needs tweaking.

keithbostic:
> First, I found some places where we could get a WT_ROLLBACK error due to cache pressure while restoring a page with lookaside history. That fixed some of the test cases that were getting rollback errors during commit/rollback, but not others.

This fixes the core failure in WT-4362
```
file:wt.wt, WT_SESSION.commit_transaction: __wt_txn_resolve_prepared_op,
360: txn->multi_update_count > 0
```
But the run still fails with:
```
t: FAILED: ops/1058: ret == 0 || ret == WT_ROLLBACK
```

keithbostic:
> Then I added a log of state transitions to the WT_REF structure so I could see where things were going wrong: a page with history evicted in lookaside was ending up in the WT_REF_MEM state without the history loaded into cache. This was happening as part of __evict_page_clean_update (first to WT_REF_DISK so on the next read ref->page_las was ignored).

Couldn't we assert that never happens?

src/btree/bt_read.c
agorrod: I agree - this looks like a bug fix. Nice catch.

src/btree/bt_read.c
```diff
@@ -556,12 +553,14 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
...
 	 */
-	if (final_state == WT_REF_MEM &&
-	    ref->page_las != NULL && (prepared || !ref->page_las->skew_newest))
-		WT_ERR(__wt_las_remove_block(
-		    session, ref->page_las->las_pageid));
+	if (final_state == WT_REF_MEM && ref->page_las != NULL &&
+	    (!ref->page_las->skew_newest || ref->page_las->has_prepares)) {
+		ret = __wt_las_remove_block(
+		    session, ref->page_las->las_pageid);
+		WT_ERR(ret);
```
sueloverso: Was there other code you temporarily had here that the call was split into `ret =` and `WT_ERR(ret)`? It can be put back together.

src/include/btmem.h
sueloverso: I'll add my hat to the ring that this seems really useful, probably under diagnostic.

src/include/btmem.h
```diff
@@ -856,6 +857,24 @@ struct __wt_ref {
...
 	volatile uint32_t state;	/* Page state */
 
+	struct {
+		WT_SESSION_IMPL *session;
+		const char *name;
+		const char *file;
+		int line;
+		uint32_t state;
+	} hist[3];
+	int histoff;
```
sueloverso: My only minor comment about the `WT_REF_SETSTATE` macro is trying to find a way to enforce its usage. How difficult would it be to look for `ref->state = ` in one of the dist scripts to flag it perhaps? That should not hold up the merge, just an idea.

michaelcahill:
Thanks, @agorrod – I probably won't get much dedicated time this week, so if you and @bvpvamsikrishna  can get the change into shape for merging, that would be great.

src/btree/bt_read.c
michaelcahill: Yes, I was asserting that this call didn't fail (and didn't see it fail with the test cases I tried).  I agree this part should be reverted.

src/include/btmem.h
michaelcahill: Yes, we have checks like this in `s_style`: adding something like this (and the `WT_PUBLISH` equivalent) makes sense to me.

sueloverso:
This branch fixed the failures in WT-4364. I do see a significant percentage of those processes (running 20 in parallel) fail with stuck cache. Maybe 6-8 out of 20. Since so many of the 20 were failing in WT-4364 (12 of 20), I cannot tell if this branch unblocks us to hit cache-stuck or if it makes it more likely.

bvpvamsikrishna:
@agorrod, could you take a look at the recent updates to address review comments.

src/cache/cache_las.c
keithbostic: @agorrod, should this be reverted before the merge? I don't think it's a necessary change, and it potentially changes the performance of `__wt_las_cursor_position()`.

(I don't know if that matters or not.)

src/include/btmem.h
keithbostic: @sueloverso, @michaelcahill 

I think that's going to be difficult to do, it's easy to split the `=` operator from the `WT_REF_XXX` name, and we have places we set `WT_REF.state` in interesting ways. For example the `new_state` and `final_statep` uses:

```
% egrep ' = WT_REF_' ../*/*.[chyis]
../btree/bt_handle.c:	root_ref->state = WT_REF_MEM;
../btree/bt_read.c:		*final_statep = WT_REF_LIMBO;
../btree/bt_read.c:		new_state = WT_REF_READING;
../btree/bt_read.c:		new_state = WT_REF_LOCKED;
../btree/bt_read.c:	final_state = WT_REF_MEM;
../btree/bt_rebalance.c:	copy->state = WT_REF_DISK;
../btree/bt_split.c:	child->state = WT_REF_MEM;
../btree/bt_split.c:	child->state = WT_REF_MEM;
../evict/evict_lru.c:	*previous_statep = WT_REF_MEM;
../evict/evict_page.c:			child->state = WT_REF_DELETED;
```

An alternative would be to obscure the field name, and add `WT_REF_STATE_READ` and `WT_REF_STATE_WRITE` macros.

src/cache/cache_las.c
agorrod: I figured that since @michaelcahill was suspicious of the previous behavior, and it's a reasonable change we might as well keep it. We are in the cache overflow path - so I'm not really concerned that doing a get_key will noticeably change performance.

agorrod:
Thanks @bvpvamsikrishna and @michaelcahill I'm going to go ahead and merge this. Please open a ticket if you can think of follow up work that would make this change better.

PR #4323, opened 2018-10-22
WT-4366 format doesn't handle WT_PREPARE_CONFLICT returns on reads

keithbostic:


keithbostic:
@bvpvamsikrishna, with this change, the WT-4366 CONFIG still fails, with a variety of errors (for example, cursor next/prev keys returned out-of-order, keys return not-found during snapshot isolation).

I'm intending to chase those problems, but before I do that I'd like to verify that the behavior here, especially for cursor search-near and next/prev operations,  is as you expected it to be.

bvpvamsikrishna:
@keithbostic, I think behavior is as I expect, one not significant suggestion is instead of a tight loop, may be better have a __wt_sleep, before retrying.

keithbostic:
@bvpvamsikrishna,
>  one not significant suggestion is instead of a tight loop, may be better have a __wt_sleep, before retrying.

Is our expectation a prepare-conflict will take a significant amount of time to resolve, that is, it's application dependent?

keithbostic:
Sorry, folks -- I merged the wrong branch, I had to revert some stuff. I think it's fixed.

bvpvamsikrishna:
> @bvpvamsikrishna,
> 
> > one not significant suggestion is instead of a tight loop, may be better have a __wt_sleep, before retrying.
> 
> Is our expectation a prepare-conflict will take a significant amount of time to resolve, that is, it's application dependent?

No, we don't expect prepare-conflict to take significant time to resolve.

agorrod:
I opened a new pull request with the same code change, based off develop.

PR #4324, opened 2018-10-22
WT-4374 Fix a bug where a page could transition from WT_REF_LIMBO incorrectly

agorrod:
It's possible to transition from WT_REF_LIMBO to WT_REF_MEM incorrectly.


agorrod:
@bvpvamsikrishna please review this change once testing has passed and merge if you are happy with it.

@michaelcahill I pulled this change out of the WT-3898 branch, and attempted to fix up JIRA to match.

PR #4325, opened 2018-10-24
WT-4366 Fix how test/format handles prepare conflict errors

agorrod:
It should retry until a different result is seen.


agorrod:
The previous branch was based off the changes in WT-3898, which would make it hard to uncouple in the future, so I've pushed a new branch against develop.

agorrod:
@bvpvamsikrishna lgtm'ed the previous version of this change.

PR #4326, opened 2018-10-24
WT-4371 Change ordering of referenced shared libraries

ddanderson:


ddanderson:
Oops, wrong ticket number, closing.

PR #4327, opened 2018-10-24
WT-4387 Fix ordering of referenced shared libraries

ddanderson:


ddanderson:
@sueloverso, can you please review?  The rationale is in the ticket.

ddanderson:
Sure, good idea!

PR #4328, opened 2018-10-24
WT-4298 Workgen fixes. Retry on WT_ROLLBACK. Clear ops after warmup.

sueloverso:
Not yet ready for review. There is still one more area that needs fixing.

### DIFF
--- a/bench/workgen/workgen.cxx
+++ b/bench/workgen/workgen.cxx
@@ -778,8 +778,8 @@ int ThreadRunner::op_run(Operation *op) {
         if (op->_transaction != NULL) {
             if (_in_transaction)
                 THROW("nested transactions not supported");
-            _session->begin_transaction(_session,
-              op->_transaction->_begin_config.c_str());
+            WT_ERR(_session->begin_transaction(_session,
+              op->_transaction->_begin_config.c_str()));
             _in_transaction = true;
         }
         if (op->is_table_op()) {
@@ -797,7 +797,7 @@ int ThreadRunner::op_run(Operation *op) {
                 if (ret == WT_NOTFOUND) {
                     ret = 0;
                     track = &_stats.not_found;
-		}
+                }
                 break;
             case Operation::OP_UPDATE:
                 ret = cursor->update(cursor);
@@ -816,7 +816,7 @@ int ThreadRunner::op_run(Operation *op) {
             else {
                 retry_op = true;
                 track->rollbacks++;
-                WT_TRET(_session->rollback_transaction(_session, NULL));
+                WT_ERR(_session->rollback_transaction(_session, NULL));
                 _in_transaction = false;
                 ret = 0;
             }
--- a/bench/workgen/workgen.cxx
+++ b/bench/workgen/workgen.cxx
@@ -1301,14 +1301,15 @@ void TableOperationInternal::parse_config(const std::string &config)
     }
 }
 
-Track::Track(bool latency_tracking) : ops_in_progress(0), ops(0),
+Track::Track(bool latency_tracking) : ops_in_progress(0), ops(0), rollbacks(0),
     latency_ops(0), latency(0), bucket_ops(0), min_latency(0), max_latency(0),
     us(NULL), ms(NULL), sec(NULL) {
     track_latency(latency_tracking);
 }
 
 Track::Track(const Track &other) : ops_in_progress(other.ops_in_progress),
-    ops(other.ops), latency_ops(other.latency_ops), latency(other.latency),
+    ops(other.ops), rollbacks(other.rollbacks),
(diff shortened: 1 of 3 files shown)


sueloverso:
@ddanderson please review this change. It is not yet ready to merge for two superficial reasons:
1. I'm not sure of the coding style for the workgen source. It appears some lines I changed use tabs and some use spaces. What is the indentation depth and type?
2. When adding support for rollback it felt like a good idea to note that we got a rollback. So I added a counter in the tracking code. But other than increment it, I don't do anything with that field and it is never reported. It should either be removed or something real done. Nothing felt quite right to me, so I'm leaning toward removing it.

ddanderson:
@sueloverso, at least for the *.cxx file, there should not be any tabs. It's possible this is not true of the .h files, they may have tabs enforced by the dist scripts!  We should fix that.  Superficially (with regards to spacing, indentation), I tried to follow MongoDB standards.  Indent 4, continuation +2. There are some things we don't really follow (I think MongoDB requires {} around any nested statement). And beyond the superficial level, our C++ style is much looser. I think that's appropriate, but there's no formalized approach.

ddanderson:
On the issue of tracking rollback, I'm glad you added that.  I think it should probably get out into the monitor.json file so it is visible when we're doing analysis.

ddanderson:
I'll review now, and ignore spacing issues.

bench/workgen/workgen.cxx
```diff
@@ -775,30 +771,57 @@ int ThreadRunner::op_run(Operation *op) {
...
+                break;
+            case Operation::OP_UPDATE:
+		notfound_ok = true;
+                ret = cursor->update(cursor);
+                break;
+            default:
+                ASSERT(false);
+            }
+	    // Assume success and no retry unless ROLLBACK.
+            retry_op = false;
```
ddanderson: I don't think this handles the `WT_NOTFOUND` case when `notfound_okay` is `false`.  For example, if there is a `cursor->search` that returns `WT_NOTFOUND`, an error is not reported, and it should be.

ddanderson:
As far as reporting rollbacks, I think it should be one of the fields emitted by the `TRACK_JSON` macro.

bench/workgen/workgen.cxx
sueloverso: @ddanderson can you look at the current version in the PR? I redid this and I believe it now matches what the original code does.

bench/workgen/workgen.cxx
```diff
@@ -774,53 +775,56 @@ int ThreadRunner::op_run(Operation *op) {
...
-		notfound_ok = true;
                 ret = cursor->remove(cursor);
+                if (ret == WT_NOTFOUND)
+                    ret = 0;
                 break;
             case Operation::OP_SEARCH:
                 ret = cursor->search(cursor);
+                if (ret == WT_NOTFOUND) {
+                    ret = 0;
+                    track = &_stats.not_found;
```
ddanderson: I think the closing brace has a spacing error.

bench/workgen/workgen.cxx
```diff
@@ -774,53 +775,56 @@ int ThreadRunner::op_run(Operation *op) {
...
+            if (ret != 0 && ret != WT_ROLLBACK)
+                WT_ERR(ret);
+            if (ret == 0)
                 cursor->reset(cursor);
-	    else {
+            else {
                 retry_op = true;
-		track->rollbacks++;
+                track->rollbacks++;
                 WT_TRET(_session->rollback_transaction(_session, NULL));
```
ddanderson: I think the use of `WT_TRET` here should be `WT_ERR`?

bench/workgen/workgen.cxx
```diff
@@ -774,53 +775,56 @@ int ThreadRunner::op_run(Operation *op) {
     }
     // Retry on rollback until success.
     while (retry_op) {
-	if (op->_transaction != NULL) {
+        if (op->_transaction != NULL) {
             if (_in_transaction)
                 THROW("nested transactions not supported");
             _session->begin_transaction(_session,
```
ddanderson: This isn't part of your changes, but the call to `begin_transaction` should be inside a `WT_ERR` macro.

bench/workgen/workgen.cxx
sueloverso: Fixed.

bench/workgen/workgen.cxx
sueloverso: Good catch. Fixed.

bench/workgen/workgen.cxx
sueloverso: Fixed.

sueloverso:
Thanks @ddanderson. I have fixed all the comments. I will merge once the PR testers complete.

PR #4330, opened 2018-10-24
WT-4366 Fix how test/format handles prepare conflict errors

lukech:
There was fallout from the initial merge of #4325 regarding cursor out-of-order:
http://build.wiredtiger.com:8080/job/wiredtiger-test-race-condition-stress-sanitizer/27811/
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress/68652/

@keithbostic @agorrod I recreated the PR based on the old `wt-4366-format-prepare-conflict2`  branch for further work.

PR #4331, opened 2018-10-25
WT-4385 Prepare-conflict during a cursor scan can return the wrong key

keithbostic:
@bvpvamsikrishna, the core bug fix is 96a3059.

I think 7ae8a03 probably needs close review. If I'm wrong in my assertion, we'll still need some kind of change there because it must be wrong to fall into the iteration code.

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -604,24 +604,22 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 	F_CLR(cbt, WT_CBT_RETRY_PREV);
 	if (F_ISSET(cbt, WT_CBT_RETRY_NEXT)) {
 		WT_ERR(__wt_cursor_valid(cbt, &upd, &valid));
+		WT_ASSERT(session, valid == true);
+
+		/* The update that returned prepared conflict is now visible. */
 		F_CLR(cbt, WT_CBT_RETRY_NEXT);
-		if (valid) {
-			/*
-			 * If the update, which returned prepared conflict is
-			 * visible, return the value.
-			 *
-			 * The underlying key-return function uses a comparison
-			 * of 0 to indicate the search function has pre-built
-			 * the key we want to return. That's not the case here,
-			 * make sure we don't take that path.
-			 */
-			cbt->compare = 1;
-			WT_ERR(__cursor_kv_return(session, cbt, upd));
+
+		/*
+		 * The underlying key-return function uses a comparison value
+		 * of 0 to indicate the search function has pre-built the key
+		 * we want to return. That's not the case, don't take that path.
+		 */
+		cbt->compare = 1;
+		WT_ERR(__cursor_kv_return(session, cbt, upd));
 #ifdef HAVE_DIAGNOSTIC
-			WT_ERR(__wt_cursor_key_order_check(session, cbt, true));
+		WT_ERR(__wt_cursor_key_order_check(session, cbt, true));
 #endif
-			return (0);
-		}
+		return (0);
 	}
 
 	WT_ERR(__cursor_func_init(cbt, false));
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -603,7 +603,7 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 	 */
 	F_CLR(cbt, WT_CBT_RETRY_PREV);
 	if (F_ISSET(cbt, WT_CBT_RETRY_NEXT)) {
-		WT_RET(__wt_cursor_valid(cbt, &upd, &valid));
+		WT_ERR(__wt_cursor_valid(cbt, &upd, &valid));
 		F_CLR(cbt, WT_CBT_RETRY_NEXT);
 		if (valid) {
 			/*
@@ -616,11 +616,15 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 			 * make sure we don't take that path.
 			 */
 			cbt->compare = 1;
-			return (__cursor_kv_return(session, cbt, upd));
+			WT_ERR(__cursor_kv_return(session, cbt, upd));
+#ifdef HAVE_DIAGNOSTIC
+			WT_ERR(__wt_cursor_key_order_check(session, cbt, true));
+#endif
+			return (0);
 		}
 	}
 
-	WT_RET(__cursor_func_init(cbt, false));
+	WT_ERR(__cursor_func_init(cbt, false));
 
 	/*
 	 * If we aren't already iterating in the right direction, there's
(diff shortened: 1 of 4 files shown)


keithbostic:
Once this PR is merged, I think we can merge #4330 as well.

keithbostic:
@agorrod, @bvpvamsikrishna --

This change results in a compiler error if `--disable=shared` is configured:
```
make src/btree/bt_curnext.o
  CC       src/btree/bt_curnext.o
In file included from ./src/include/wt_internal.h:396:0,
                 from src/btree/bt_curnext.c:9:
./src/include/cursor.i: In function '__cursor_row_slot_return':
./src/include/cursor.i:379:25: error: '_kpack.prefix' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  WT_CELL_UNPACK *kpack, _kpack, *vpack, _vpack;
                         ^
./src/include/cursor.i:379:25: error: '_kpack.data' may be used uninitialized in this function [-Werror=maybe-uninitialized]
./src/include/cursor.i:379:25: error: '_kpack.size' may be used uninitialized in this function [-Werror=maybe-uninitialized]
./src/include/cursor.i:379:25: error: '_kpack.__len' may be used uninitialized in this function [-Werror=maybe-uninitialized]
cc1: all warnings being treated as errors
make: *** [src/btree/bt_curnext.o] Error 1
```
Which makes no sense at all to me. It's easy enough to trigger in the develop branch, just use this diff:
```
index c9cccc6..834c360 100644
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -594,6 +594,9 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 
        F_CLR(cursor, WT_CURSTD_KEY_SET | WT_CURSTD_VALUE_SET);
 
+       if (cursor == NULL)
+               return (EINVAL);
+
        /*
         * In case of retrying a next operation due to a prepare conflict,
         * cursor would have been already positioned at an update structure
```
In other words, if we return "early" out of `__wt_btcur_next()`, the compiler complains that data structures in `__cursor_row_slot_return()` are used uninitialized, and as far as I can tell, an early return just means that function won't be called.

I'm going to push a change to shut the compiler up, but that makes me pretty unhappy. Can either of you spot why the compiler is unhappy with an early return?

bvpvamsikrishna:
@keithbostic, Thank you for chasing this down.

src/btree/bt_curnext.c
```diff
@@ -603,18 +603,26 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
-			return (__cursor_kv_return(session, cbt, upd));
-		}
+
+		/*
+		 * The underlying key-return function uses a comparison value
+		 * of 0 to indicate the search function has pre-built the key
+		 * we want to return. That's not the case, don't take that path.
+		 */
+		cbt->compare = 1;
+		WT_ERR(__cursor_kv_return(session, cbt, upd));
```
bvpvamsikrishna: I think this can handle even when upd == NULL ?
If the prepared_updated we were pointing to is committed and if it is a delete operation, then we should continue and get the next value. My understanding is this is already handled as part of __cursor_kv_return.

src/btree/bt_curnext.c
```diff
@@ -603,18 +603,26 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
-			 */
-			return (__cursor_kv_return(session, cbt, upd));
-		}
+
+		/*
+		 * The underlying key-return function uses a comparison value
+		 * of 0 to indicate the search function has pre-built the key
+		 * we want to return. That's not the case, don't take that path.
+		 */
+		cbt->compare = 1;
```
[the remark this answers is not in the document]
keithbostic: That was deliberate.

Basically, the **correct** value is 0, because we're pointing to a key which is an exact match for what we're "searching for". The problem is that "searching for" doesn't have a lot of meaning for a scan.

We never called `__cursor_kv_return()` from the cursor scan code before it was added to support prepare-conflict.  The `__cursor_kv_return()` function wasn't ever intended for scans, it was there for updates and other cursor operations that involve searches, and adding it got us into trouble.

The code I'm trying to avoid is here:
```
__key_return(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
{               
        WT_CURSOR *cursor;
        WT_ITEM *tmp;
        WT_PAGE *page;
        WT_ROW *rip;
                        
        page = cbt->ref->page;
        cursor = &cbt->iface;   
                                
        if (page->type == WT_PAGE_ROW_LEAF) {
                rip = &page->pg_row[cbt->slot];
                                
                /*      
                 * If the cursor references a WT_INSERT item, take its key.
                 * Else, if we have an exact match, we copied the key in the
                 * search function, take it from there.
                 * If we don't have an exact match, take the key from the
                 * original page.
                 */             
                if (cbt->ins != NULL) {
                        cursor->key.data = WT_INSERT_KEY(cbt->ins);
                        cursor->key.size = WT_INSERT_KEY_SIZE(cbt->ins);
                        return (0); 
                }               
                        
                if (cbt->compare == 0) {         <<<< THIS TEST
```
And, setting `compare = 1/-1` doesn't make any difference for avoiding that test, it just can't be 0.

To be clear, I agree with the change to call `__cursor_kv_return()` in the scan code. The alternative to calling `__cursor_kv_return()` is a lot worse (at least I wasn't able to find anything l liked better when I was messing around with it today). You either end up duplicating a bunch of code, or trying to set the key BEFORE returning a prepare-conflict error on the value, and neither of those led me anywhere I wanted to go.

So, in summary, it doesn't matter to what we set `compare` in the scan code, and setting it to `-1` explicitly seemed misleading, at best.

If you can think of a way to refactor this that would make it more explicit, I'd be a fan. I don't like how we're handing it at the moment, it's seems fragile and dangerous.

But, as I said, when I messed with it today, I didn't find anything I liked better. (The hard part isn't avoiding the test, for example, we could add a cursor field that indicates search has built a key for us. The problem is figuring out when/where to clear that new field, and that gets hard.)

The only approach I didn't hate was changing `__cursor_kv_return()` and all its children (down to `__key_return()`), to take an additional argument to configure if the `cbt->compare == 0` test should be performed, that is, an argument that says "the current cursor position is the result of a search". That wasn't too bad -- I don't honestly remember exactly why I decided against it at the end.

If you like that better, let's talk!

src/btree/bt_curnext.c
keithbostic: That's one of the reasons I flagged this change for you. I wasn't sure if a prepare-conflict followed by a committed delete would return `WT_NOTFOUND` or `0 with upd=NULL, valid=false`. If it's the latter, then I think I agree, we should fall out of that clause into a normal iteration.

I don't understand what you mean by "handled as part of __cursor_kv_return". What specific code are you thinking of that handles this?

bvpvamsikrishna:
@keithbostic, I am writing a python test, to test the cursor next/move with prepared transactions and will merge this along with that test. 
Code changes are lgtm for me.
This bug fix is for prepared transactions and I feel is not very urgent to merge in develop. If you feel otherwise, I am happy to merge this first.

keithbostic:
@bvpvamsikrishna 
> This bug fix is for prepared transactions and I feel is not very urgent to merge in develop.

fbm.

PR #4333, opened 2018-10-28
WT-4376 Single-thread index open with WT_WITH_TABLE_WRITE_LOCK.

keithbostic:


src/cursor/cur_backup.c
```diff
@@ -144,11 +144,15 @@ __wt_curbackup_open(WT_SESSION_IMPL *session,
...
-	 * lock, we need a consistent view when creating a copy.
+	 * lock, we need a consistent view when creating a copy.  Further, if
+	 * backing up individual tables, we have to include indexes, which may
+	 * involve opening those indexes.  Acquire the table lock in write mode
+	 * for that case.
 	 */
 	WT_WITH_CHECKPOINT_LOCK(session,
 	    WT_WITH_SCHEMA_LOCK(session,
-		ret = __backup_start(session, cb, cfg)));
+		WT_WITH_TABLE_WRITE_LOCK(session,
```
agorrod: I'm a little surprised this is required - I thought backup went through {{__wt_meta_apply_all}} - avoiding the schema level operations. Does it do a schema apply as well?

michaelcahill: Yes, this is a "targeted backup" of named tables.

agorrod: Got it - now I see. It's unfortunate that the layering means we need to take that lock even for the more common full and incremental backup cases. I'll open a different ticket to address that.

agorrod:
One more question.. Do we not need this for compact because that code is going through the normal handle-open routines?

src/cursor/cur_backup.c
michaelcahill: I've moved it in the latest commit so we only acquire that lock when doing targeted checkpoints.

michaelcahill:
Thanks @agorrod, nice catch: I've updated compact to also acquire the table lock when figuring out which handles to operate on.

agorrod:
Thanks @michaelcahill lgtm now - I'll merge as long as it passes testing.

PR #4335, opened 2018-10-29
WT-4371 workgen improvements to create a MongoDB-like workload

ddanderson:
Includes the ability specify a fractional sample_interval,
an operator for log_flush, and more realistic log_like files
(enables logging on that file, and adds a flush/sleep thread).
Resolved some warnings for variables and parameters that were unused.


ddanderson:
Not quite ready for review.

ddanderson:
@sueloverso, can you take a look at this please?  I think this moves us forward, but an unfortunate effect is that the strong correlation between poor latencies and checkpoints has largely disappeared.  I have some suspicions about this, and will investigate.

sueloverso:
@ddanderson my reaction to the `sample_interval` changes is that I'd suggest instead of messing with floats and conversions, why not just have the value be milliseconds instead of seconds. Using a float makes it incompatible with `wtperf` anyway. I think milliseconds offers ample granularity.

ddanderson:
@sueloverso, if `sample_interval` was changed to be ms, then wtperf.py could no longer run old `wtperf` scripts.  I'd rather introduce a `sample_interval_ms` or even an alternate notation `sample_interval=500ms` so we could keep back compatibility.  Changing `sample_interval` to a `double` value was an attempt to keep backward compatibility.

sueloverso:
@ddanderson fair enough. As I was typing that I didn't think I had a preference, I discovered I did! I like your suggestion of adding a unit to `sample_interval`. If no unit is specified it can stay the default of seconds. And if a unit is specified then we use it.

PR #4336, opened 2018-10-29
WT-4343 Unlock when sleeping to allow other log threads to make progress.

sueloverso:
@agorrod here's a lighter version of the same fix from the other branch. I found that simply releasing and re-acquiring the locks allows threads to make progress and I am unable to reproduce the many-thread failure with these smaller changes (well over 500 iterations). On develop I can reproduce usually in under 30 iterations.

I also moved the Jenkins log perf test to this branch and it avoids the issues the original branch showed. The min/max throughput is back where it was and the performance of the logging variants is within normal variations.

I believe this branch will have less impact on MongoDB. If you agree, this should be reviewed and merged and the other one closed.


agorrod:
Thanks @sueloverso this version looks good to me - I'm much more comfortable with this version.

PR #4337, opened 2018-10-30
WT-4393 Fix read committed transaction visibility problem of next / prev

bvpvamsikrishna:
For read committed isolation, each operation needs to get a transaction snapshot.


agorrod:
We decided to update the documentation rather than make this behavior change.

PR #4338, opened 2018-10-30
WT-4394 Reduce Evergreen Ubuntu build variant runtime

lukech:
This is NOT READY FOR REVIEW yet!

Changes include:
- Add SMP command option (-j 4) to the 'make check' command
- Remove (disable) 'TESTUTIL_ENABLE_LONG_TESTS=1' setting before calling 'make check'
- Split csuite tests into a separate task


test/mciproject.yml
```diff
@@ -80,7 +81,13 @@ tasks:
                 install_name_tool -change /usr/local/lib/libwiredtiger-$WT_VERSION.dylib $(pwd)/.libs/libwiredtiger-$WT_VERSION.dylib .libs/wt
               fi
 
-              ${test_env_vars|} TESTUTIL_ENABLE_LONG_TESTS=1 ${make_command|make} VERBOSE=1 check 2>&1
+              # Exclude csuite test which is covered by a separate task
+              dirs=$(cat build_posix/Make.subdirs | grep -v ^# | grep -v ^$ | sed '/\./d' | cut -d' ' -f1 | grep -v test/csuite)
+              echo ${dirs}
+              for dir in ${dirs}
+              do
+                ${test_env_vars|} ${make_command|make} VERBOSE=1 check -C ${dir} ${smp_command|}
```
lukech: Do we really want to disable the long test option (i.e. `TESTUTIL_ENABLE_LONG_TESTS=1`) for the Ubuntu build variant? There's tradeoff here we need to decide between testing scope and duration. There were over 20mins saving if disabling the long test option (see comparison table [here](https://jira.mongodb.org/browse/BUILD-4395?focusedCommentId=2042941&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2042941)).

agorrod: I'd like to run the long tests individually as separate jobs. That is going to require work as more tests are added, but I think it's worthwhile for the benefit of parallelizing our testing.

test/mciproject.yml
```diff
@@ -99,6 +106,38 @@ tasks:
...
+          working_dir: "wiredtiger"
+          script: |
+            set -o errexit
+            set -o verbose
+            if [ "Windows_NT" = "$OS" ]; then
+              scons.bat --enable-python=c:\\swigwin-3.0.2\\swig.exe --enable-diagnostic ${smp_command|}
+            else
+              ./build_posix/reconf
+              ${configure_env_vars|} ./configure --enable-diagnostic --enable-python --enable-zlib --enable-strict
+              ${make_command|make} ${smp_command|} 2>&1
```
agorrod: I was expecting to have a compile job, that just compiles, then other jobs that copies the artifacts to run tests. I understand that compile is pretty fast for standalone WiredTiger, but it's more idiomatic method for Evergreen.

test/mciproject.yml
lukech: When talking about `separate jobs`, do you mean `separate build variants` in evergreen?

test/mciproject.yml
lukech: That's actually the next thing I planned to do :) 
The only concern I have is the overhead of s3 put (for the compile task) and s3 get (for all other dependent tasks). But will test and see.

test/mciproject.yml
agorrod: I meant separate `tasks`

lukech:
With above 2 commits, all 3 build variants (Ubuntu, OS X, Windows) passed both the `compile` and `unit-test` tasks. https://evergreen.mongodb.com/version/5be2951b2a60ed2cb0be0608We can start the work of splitting tasks based on this

lukech:
@agorrod I pushed a few commits today. In current configuration there's one example task (`examples-c-test`) for 'make check' specific directory, and another example task (`csuite-col-append-test`) for breaking down csuite long running tests. Latest patch build is [here](https://evergreen.mongodb.com/version/5bea5f132a60ed32bca741be). Both tasks were setup to depend on the new `compile` task (pure compiling, no make check), fetch artifacts, re-configure and re-make, then run the test. Arguably letting each task fetch wiredtiger repo and do a clean compile could save overall runtime for the build variant, but my preference is to keep the `compile` task dependency and consistent workspace when possible. Please let me know if that's what you expected.

test/mciproject.yml
```diff
@@ -58,69 +83,86 @@ tasks:
...
-            # Protection enabled, it ignores DYLD_LIBRARY_PATH and hence
-            # doesn't find the WiredTiger library in the local tree.
-            ${test_env_vars|} python ./test/suite/run.py -v 2 ${smp_command|} 2>&1
+            if [ "Windows_NT" = "$OS" ]; then
+              ${test_env_vars|} python ./test/suite/run.py -v 2 ${smp_command|} 2>&1
+            elif [ "$(uname -s)" == "Darwin" ]; then
+              # Avoid /usr/bin/python, at least on macOS: with System Integrity
+              # Protection enabled, it ignores DYLD_LIBRARY_PATH and hence
+              # doesn't find the WiredTiger library in the local tree.
+              ${test_env_vars|} python ./test/suite/run.py -v 2 ${smp_command|} 2>&1
```
agorrod: Is this exactly the same as the Windows command?

test/mciproject.yml
```diff
@@ -58,69 +83,86 @@ tasks:
...
+            else # Ubuntu
+              # Change directory to where the local installed 'wt' binary is located, 
+              # to avoid libtool generated 'wt' script from being selected by run.py,
+              # which invokes relink_command that tries to changing to a non-existed 
+              # /data/mci/<uniq> directory, as 'make' is done by a separate 'compile' task.
+              cd LOCAL_INSTALL/bin
```
agorrod: Seems like this `cd` is the primary difference - could you do this the other way, and `if #Ubuntu` do the `cd`, and then all other places run the same python command?

I see that the following line uses `test_env_vars_py` rather than `test_env_vars` - that distinction confuses me - I feel like it should either be made the same as other cases or renamed to make the difference more obvious

test/mciproject.yml
```diff
@@ -58,69 +83,86 @@ tasks:
...
-             - "./**"
-      - command: s3.put
+          working_dir: "wiredtiger"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            ${test_env_vars|} ${make_command|make} VERBOSE=1 check -C examples/c  ${smp_command|} 2>&1
+
+  - name: csuite-col-append-test
```
agorrod: I'd prefer it if this name were auto-generateable from either the directory structure or the test name - so we can easily script the creation of this file at some stage.

test/mciproject.yml
lukech: `cd` here is critical here to make sure `run.py` locates the `wt` binary (instead of the `wt` script) to avoid non-existed directory error. I can change the checking sequence as suggestion. 

The `test_env_vars_py` addition is a bit ugly. It is needed for python test suite as the relative directory (to `$(pwd)`) becomes different after the `cd` command. Does `python_test_env_vars` make more sense?

test/mciproject.yml
lukech: For now I manipulate the test name a bit to get the task name in a manual approach - e.g. for test name `test_wt2246_col_append`, remove the `test_` prefix and `wt2246_` ticket info, then  add `csuite-` prefix and `-test` postfix to get the task name `csuite-col-append-test`. Should be okay to scriptize it later.

lukech: How to figure out if a test supports `TESTUTIL_ENABLE_LONG_TESTS` is an open question. Searching the keyword in source code feels less than ideal.

lukech:
Separate tasks had been added for each csuite test (both long-running and non-long-running). 
Path build: https://evergreen.mongodb.com/version/5bed1cc62fbabe6de4a3906f

lukech:
Now we should have a full coverage of the `make check` test scope, with below 2 set of new tasks added into the `Ubuntu 14.04` build variant:
- one task for each make check **applicable** directory (except for `test/csuite` which is covered by the 2nd set)
- one task for each csuite test (both long-running and non-long running tests)
Note: The make check applicable directories in the 1st set were figured out by using the list of directories in `build_posix/Make.subdirs` as base, and removing directories that had `Nothing to be done` for make check according to the task log of a current Evergreen [run](https://evergreen.mongodb.com/task_log_raw/wiredtiger_ubuntu1404_compile_b9701cbebaac842fa6ea3d5a264bf3ddeb239f15_18_11_13_16_52_17/0?type=T&text=true).  

The latest patch build is here: https://evergreen.mongodb.com/version/5bee18062a60ed092b9afa8a
The `Makespan` for Ubuntu 14.04 build variant was **44m 0s**, mainly dragged by the longest-running task `unit-test` (**17m 46s**, not being tuned yet), and the delay of task scheduling - `unit-test` task got scheduled **21m** after `compile` task being finished. There could be ways to get the dependent tasks start sooner that we can check with dev-prod team. 

@agorrod, before we talk with David B. about the `30 mins patch build` tooling and do further improvements, I think we've completed the base work of splitting `make check` test into multiple parallel tasks - the initial goal of this ticket.

test/mciproject.yml
```diff
@@ -184,11 +834,56 @@ buildvariants:
...
-    configure_env_vars: CC=/opt/mongodbtoolchain/bin/gcc CXX=/opt/mongodbtoolchain/bin/g++
+    configure_env_vars: CC=/opt/mongodbtoolchain/bin/gcc CXX=/opt/mongodbtoolchain/bin/g++ PATH=/opt/mongodbtoolchain/v2/bin:$PATH
   tasks:
     - name: compile
+    - name: lang-python-test
+    - name: examples-c-test
+    - name: bloom-test
+    - name: checkpoint-test
+    - name: cursor-order-test
+    - name: fops-test
```
lukech: Does the scope of this new `fops-test` task cover that of the old `fops` task? The new task ran much longer time, and the old task has more verbose log output.

test/mciproject.yml
lukech: I tried to adjust the OS checking sequence as suggested, but got a failed [patch](https://evergreen.mongodb.com/version/5bf4ac0ee3c33109cf6bc2a1) - the `if [ "$(uname -s)" == "Linux" ]` checking did not seem working for some reason. I tried before on a ubuntu14.04-test spawn host before and `uname -s` returned `Linux`. 
Considering we are going to split unit-test task in WT-4410, in which the whole OS checking logic here will be simplified/removed, I'd prefer to keep the current change as is.

test/evergreen.yml
```diff
@@ -0,0 +1,925 @@
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/test/csuite"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            ${test_env_vars|} $(pwd)/test_truncated_log 2>&1
+
+  - name: csuite-col-efficiency-test
```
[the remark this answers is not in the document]
lukech: Initially I intentionally hid the ticket numbers to avoid distracting readers over the test scenarios. Since you asked for it implies the ticket numbers do help with troubleshooting, I'm happy to add them back.

PR #4339, opened 2018-10-30
WT-4395 Use WT_TRET so error return is not overwritten.

sueloverso:
@ddanderson Please review this small change. The code path changed slightly with the log corruption/salvage work a couple months ago and in the specific case of a log cursor getting an error, that error return would get overwritten and lost.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -2501,6 +2501,8 @@ advance:
 			 * the first pass through recovery.  In the second pass
 			 * where we truncate the log, this is where it should
 			 * end.
+			 * Continue processing where possible, so remember any
+			 * error returns, but don't skip to the error handler.
 			 */
 			if (log != NULL)
 				log->trunc_lsn = rd_lsn;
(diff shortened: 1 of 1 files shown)


src/log/log.c
```diff
@@ -2537,7 +2537,7 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
 				 * must be salvaged.
 				 */
 				need_salvage = true;
-				WT_ERR(__log_salvage_message(session,
+				WT_TRET(__log_salvage_message(session,
```
agorrod: @sueloverso it's not obvious but `WT_TRET` doesn't return or goto error, it just sets up `ret` - is that what you want here? Or should there be a subsequent `WT_ERR` call?

sueloverso: This code is in the "checksums don't match" area. After the various salvage calls, there is an unconditional `break`.

ddanderson: @sueloverso, I think the change is correct.  But I agree with Alex that it's hard to reason about (an unusual usage of `WT_TRET`).  Perhaps clearer would be keep `WT_TRET`, but add `goto truncate`, with the new label just after the loop?  And there's another occurrence of the same a few lines away.

agorrod: Got it - do we need to add a `WT_ERR(ret);` immediately [after the](https://github.com/wiredtiger/wiredtiger/pull/4339/files#diff-764f5432bdeaa6ffb0296d56d50ca6ceR2599) loop then?

agorrod:
@sueloverso and @ddanderson I've made a minor change in here - if it makes sense to you and passes testing please go ahead and merge.

sueloverso:
@agorrod I'm happy with the change and comments you added.

PR #4340, opened 2018-10-30
WT-4396 when retrying after prepare-conflict, cursor unable to find a valid update

keithbostic:
@bvpvamsikrishna, in this failure, I'm seeing update chains where all of the updates are in the `WT_PREPARE_INPROGRESS` state. I'm guessing the right approach is to retry until the update we're waiting for moves out of that state?

Here's the relevant state at one of the failures:

```
(gdb) p session.txn
$3 = {
  id = 940387,
  isolation = WT_ISO_READ_COMMITTED,
  forced_iso = 0,
  snap_min = 940344,
  snap_max = 940400,
  snapshot = 0x7f197cf70a70,
  snapshot_count = 6,
  txn_logsync = 0,
  commit_timestamp = {
    val = 605637
  },
  first_commit_timestamp = {
    val = 605637
  },
  prepare_timestamp = {
    val = 605602
  },
  read_timestamp = {
    val = 605651
  },
  commit_timestampq = {
    tqe_next = 0x0,
    tqe_prev = 0x242fe88
  },
  read_timestampq = {
    tqe_next = 0x7f19c03db460,
    tqe_prev = 0x242fec0
  },
  clear_commit_q = false,
  clear_read_q = false,
  mod = 0x7f197c1cb080,
  mod_alloc = 1040,
  mod_count = 3,
  multi_update_count = 0,
  logrec = 0x0,
  notify = 0x0,
  ckpt_lsn = {
    l = {
      offset = 0,
      file = 0
    },
    file_offset = 0
  },
  ckpt_nsnapshot = 0,
  ckpt_snapshot = 0x0,
  full_ckpt = false,
  rollback_reason = 0x0,
  flags = 4108
}

(gdb) p cbt.ins
$4 = (WT_INSERT *) 0x7f19900272b0

(gdb) yyy cbt.ins.upd
$5 = {
  txnid = 18446744073709551615,
  timestamp = {
    val = 605670
  },
  next = 0x7f19983f03a0,
  size = 11,
  type = 4 '\004',
  prepare_state = 1 '\001',
  data = 0x7f198c0b9a0e "0000172032/"
}
$6 = {
  txnid = 18446744073709551615,
  timestamp = {
    val = 605553
  },
  next = 0x7f1990027300,
  size = 11,
  type = 4 '\004',
  prepare_state = 1 '\001',
  data = 0x7f19983f03be "0000172032/"
}                                     
$7 = {
  txnid = 18446744073709551615,
  timestamp = {                       
    val = 605510
  },
  next = 0x0,
  size = 11,
  type = 4 '\004',
  prepare_state = 1 '\001',
  data = 0x7f199002731e "0000172032/"
}
```


bvpvamsikrishna:
> @bvpvamsikrishna, in this failure, I'm seeing update chains where all of the updates are in the `WT_PREPARE_INPROGRESS` state. I'm guessing the right approach is to retry until the update we're waiting for moves out of that state?

@keithbostic , yes we should retry until the update we're waiting for moves out of that state. 

I think it will be better if we move the changes to test/format/ops in this PR to WT-4366 PR.

FYI, As part of my cursor operation testing with prepared transactions, I am observing some 'not expected' behaviours. eg:- WT-4393. So, it will be a while before I merge WT-4366.

agorrod:
@bvpvamsikrishna I agree about moving the format change - could you go ahead and do that?

PR #4342, opened 2018-10-30
WT-4043 Fix assertion failure while dumping the cache

nehakhatri5:
Acquire a read lock on the dhandle just before dumping cache information. This would prevent concurrent updates to dhandle state. Also prevent dumping the handle if it is already marked discarded.

### DIFF
--- a/src/evict/evict_lru.c
+++ b/src/evict/evict_lru.c
@@ -2639,6 +2639,41 @@ __verbose_dump_cache_single(WT_SESSION_IMPL *session,
 }
 
 /*
+ * __verbose_dump_cache_apply --
+ *	Apply dumping cache for all the dhandles.
+ */
+static int
+__verbose_dump_cache_apply(WT_SESSION_IMPL *session,
+    uint64_t *total_bytesp, uint64_t *total_dirty_bytesp)
+{
+	WT_CONNECTION_IMPL *conn;
+	WT_DATA_HANDLE *dhandle;
+	WT_DECL_RET;
+
+	conn = S2C(session);
+	for (dhandle = NULL;;) {
+		WT_DHANDLE_NEXT(session, dhandle, &conn->dhqh, q);
+		if (dhandle == NULL)
+			break;
+
+		/* Skip if the tree is marked discarded by another thread. */
+		if (dhandle->type != WT_DHANDLE_TYPE_BTREE ||
+		    !F_ISSET(dhandle, WT_DHANDLE_OPEN) ||
+		    F_ISSET(dhandle, WT_DHANDLE_DISCARD))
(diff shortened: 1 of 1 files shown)


src/evict/evict_lru.c
```diff
@@ -2673,13 +2673,27 @@ __wt_verbose_dump_cache(WT_SESSION_IMPL *session)
...
 		if (dhandle == NULL)
 			break;
+		/*
+		 * Lock the handle to prevent changes in the handle's state
+		 * when it is being used to dump the cache information.
+		 */
+		if ((ret = __wt_try_readlock(session, &dhandle->rwlock)) ==
+		    EBUSY)
+			continue;
+		WT_RET(ret);
```
michaelcahill: Since this code only exists for diagnostics, is there any problem with just wrapping the whole loop in `WT_WITH_HANDLE_LIST_READ_LOCK`?

nehakhatri5: I think that this code can be invoked by the `debug_info` API too. This API is not documented but it can be used by the test programs. In that case, wrapping the loop with the `WT_WITH_HANDLE_LIST_READ_LOCK`, could that interfere with other operations?

agorrod: My concern with that is if there are cases where we want to dump the cache but can't get that lock. They are orthogonal, so this approach seems reasonable if I think about it.

agorrod: @nehakhatri5 can you answer your question?

nehakhatri5: We actually do not want other threads to get a lock on the dhandle and change its state when dumping the cache. So it is the purpose rather than interference.

nehakhatri5:
@bvpvamsikrishna please review the change.

src/evict/evict_lru.c
```diff
@@ -2638,6 +2638,39 @@ __verbose_dump_cache_single(WT_SESSION_IMPL *session,
...
+		if (dhandle->type != WT_DHANDLE_TYPE_BTREE ||
+		    !F_ISSET(dhandle, WT_DHANDLE_OPEN) ||
+		    F_ISSET(dhandle, WT_DHANDLE_DISCARD))
+			continue;
+
+		WT_WITH_DHANDLE(session, dhandle,
+		    ret = __verbose_dump_cache_single(
+		    session, total_bytesp, total_dirty_bytesp));
+		if (ret != 0)
+			break;
```
agorrod: As @bvpvamsikrishna this would be better as a WT_RET(ret), then the return at the end of the function can return 0.

agorrod:
Thanks @nehakhatri5

PR #4345, opened 2018-10-31
WT-4401 workgen: wtperf emulation: fix sample_interval with integer values

ddanderson:


ddanderson:
@sueloverso, can you take a look at this please?

PR #4346, opened 2018-11-01
WT-4402 Add rollback support and monitor JSON output.

sueloverso:
@ddanderson please review this. While debugging throttling in workgen, I'm running the configuration on wtperf. I'm seeing stalls during checkpoints and wanted the monitoring JSON that is in workgen, and it needed rollback support for the log like table.


ddanderson:
LGTM

PR #4347, opened 2018-11-02
WT-4405 Fix cursor next & prev for prepared transactions.

bvpvamsikrishna:
I have added a python test to test the cursor next / prev behaviour with respect to prepared transactions.
I have made some code changes as well.

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -603,7 +603,8 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 		WT_ERR(__cursor_check_prepared_update(cbt, &visible));
 		if (visible) {
 #ifdef HAVE_DIAGNOSTIC
-			ret = __wt_cursor_key_order_check(session, cbt, true);
+			WT_ERR(
+			    __wt_cursor_key_order_check(session, cbt, true));
 #endif
 			return (0);
 		}
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -603,8 +603,7 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 		WT_ERR(__cursor_check_prepared_update(cbt, &visible));
 		if (visible) {
 #ifdef HAVE_DIAGNOSTIC
-			WT_ERR(
-			    __wt_cursor_key_order_check(session, cbt, true));
+			WT_ERR(__wt_cursor_key_order_check(session, cbt, true));
 #endif
 			return (0);
 		}
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -582,9 +582,8 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
 	WT_DECL_RET;
 	WT_PAGE *page;
 	WT_SESSION_IMPL *session;
-	WT_UPDATE *upd;
 	uint32_t flags;
-	bool newpage, valid;
+	bool newpage, visible;
 
 	cursor = &cbt->iface;
 	session = (WT_SESSION_IMPL *)cbt->iface.session;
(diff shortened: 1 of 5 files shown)


keithbostic:
@bvpvamsikrishna, my understanding was we could have update chains that would result in `__wt_cursor_valid()` returning 0 with `valid == false`, and at the same time we'd be in a snapshot configuration that implied the cursor could not skip the key, but instead must wait until the situation resolves.

That's why I was looping on a prepare-conflict error, because I believed falling through that clause would iterate to the next key, and so potentially skip a key where all the values happen to be unavailable.

Can you please explain what I'm misunderstanding?

keithbostic:
@bvpvamsikrishna, the branch that I have been seeing fail is the CONFIG in ticket WT-4397, with the changes from WT-4366 applied to a branch.

If I take this branch, apply the changes from WT-4366 and run 15 parallel format runs using the WT-4397 CONFIG, it fails almost immediately.

To be clear, there may be something else going on, I've not tried to debug the failure, so we can't assume there's not something else going on here (or even a bug in the WT-4366 change, for that matter).

bvpvamsikrishna:
> @bvpvamsikrishna, my understanding was we could have update chains that would result in `__wt_cursor_valid()` returning 0 with `valid == false`, and at the same time we'd be in a snapshot configuration that implied the cursor could not skip the key, but instead must wait until the situation resolves.
> 

If we are reading after the commit_timestamp and the operation is 'delete' , I think we can skip the key.

I think the difference in understanding is whether we are using 'read_timestamp' or not.
The current test program I have added as part of this PR doesn't use timestamps for read operations. I am currently adding the cases with read_timestamps as well. That should reduce our confusion.

bvpvamsikrishna:
@keithbostic , I have updated the python test in this PR to use the timestamped reads. This version of test is working fine.

But when I add the column store configuration to this python test, then I can see some failures.
I am trying to debug the issues.
{noformat}
--- a/test/suite/test_prepare_cursor01.py
+++ b/test/suite/test_prepare_cursor01.py
@@ -40,6 +40,7 @@ class test_prepare_cursor01(wttest.WiredTigerTestCase):
      keyfmt = [
         ('integer', dict(keyfmt='i')),
+        ('recno', dict(keyfmt='r')),
     ]
{noformat}

bvpvamsikrishna:
@keithbostic 
I have pushed a couple of fixes commit id : "f87b148" & "73414b4". Could you take a look at the fix and let me know your opinion.

The commit "f01e647" is just code reorg, no logical change.

src/btree/bt_curnext.c
```diff
@@ -710,7 +688,12 @@ err:	switch (ret) {
 	case 0:
 		F_SET(cursor, WT_CURSTD_KEY_INT | WT_CURSTD_VALUE_INT);
 #ifdef HAVE_DIAGNOSTIC
-		ret = __wt_cursor_key_order_check(session, cbt, true);
+		/*
+		 * Skip key order check, if next is called after a prev returned
+		 * a prepare conflict error.
+		 */
+		if (!F_ISSET(cbt, WT_CBT_RETRY_PREV))
+			ret = __wt_cursor_key_order_check(session, cbt, true);
```
keithbostic: I think we should clear `WT_CBT_RETRY_NEXT` and `WT_CBT_RETRY_PREV` in `__wt_btcur_iterate_setup()`, along with `WT_CBT_ITERATE_NEXT` and `WT_CBT_ITERATE_PREV`, in other words, if we're switching directions, we clear all of the iteration flags. (For the same reason, we might want to consider renaming `CBT_RETRY_NEXT/PREV` to be `CBT_ITERATE_RETRY_NEXT/PREV`, but I don't feel terribly strongly.)

Further, I think that means `__cursor_check_prepared_update()` should not be called as early in `__wt_btcur_next/prev`, that is, we'd optionally call `__wt_btcur_iterate_setup()`, then `__cursor_func_init()`, and then `__cursor_check_prepared_update`.

I'm hoping that change means this code is no longer necessary, but I'm not 100% sure of that. I really don't like this code appearing here, I'd like to see it go away, but I don't completely understand the need for it yet.

src/btree/bt_curprev.c
```diff
@@ -667,7 +643,11 @@ err:	switch (ret) {
 	case 0:
 		F_SET(cursor, WT_CURSTD_KEY_INT | WT_CURSTD_VALUE_INT);
 #ifdef HAVE_DIAGNOSTIC
-		ret = __wt_cursor_key_order_check(session, cbt, false);
+		/* Skip key order check, if prev is called after a next returned
+		 * a prepare conflict error.
+		 */
+		if (!F_ISSET(cbt, WT_CBT_RETRY_NEXT))
+			ret = __wt_cursor_key_order_check(session, cbt, false);
```
keithbostic: Hopefully this is going away (see above), but the first line should be split, that is text only appears on the second line of a block comment.

src/btree/bt_curnext.c
```diff
@@ -595,32 +594,19 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
-		 * we want to return. That's not the case, don't take that path.
-		 */
-		cbt->compare = 1;
-		WT_ERR(__cursor_kv_return(session, cbt, upd));
+	if (F_ISSET(cbt, WT_CBT_ITERATE_RETRY_NEXT)) {
+		WT_ERR(__cursor_check_prepared_update(cbt, &visible));
+		if (visible) {
 #ifdef HAVE_DIAGNOSTIC
-		WT_ERR(__wt_cursor_key_order_check(session, cbt, true));
+			ret = __wt_cursor_key_order_check(session, cbt, true);
```
agorrod: It's unusual for us to ignore an error return code - even from a diagnostic function. If we are going to ignore it we have a macro: `WT_IGNORE_RET`. Could we handle this somehow?

src/btree/bt_curnext.c
```diff
@@ -595,32 +594,20 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
-		 */
-		cbt->compare = 1;
-		WT_ERR(__cursor_kv_return(session, cbt, upd));
+	if (F_ISSET(cbt, WT_CBT_ITERATE_RETRY_NEXT)) {
+		WT_ERR(__cursor_check_prepared_update(cbt, &visible));
+		if (visible) {
 #ifdef HAVE_DIAGNOSTIC
-		WT_ERR(__wt_cursor_key_order_check(session, cbt, true));
+			WT_ERR(
+			    __wt_cursor_key_order_check(session, cbt, true));
```
keithbostic: This line got split and didn't need to?

bvpvamsikrishna: Fixed. Thank you.

agorrod:
Thanks @bvpvamsikrishna lgtm - I'll go ahead and merge (woo!)

PR #4348, opened 2018-11-02
WT-4372 Add a script to measure and compare latency runs in a standard way

ddanderson:


ddanderson:
@sueloverso, can you please review?  I put some sample output in the ticket.

bench/workgen/latency_metric.py
```diff
@@ -0,0 +1,244 @@
...
+
+    def set_value(self, value):
+        self.value = value
+
+# A set of latency metrics collected for a single run.
+class FileMetrics:
+    def __init__(self, filename):
+        self.filename = filename
+        all = []
+        self.latency_avg = m = Metric('Average latency reads us',
```
sueloverso: I think this is a great starting place. Looking at the sample output in the ticket, some of the descriptions are fairly brief. That's fine. But it would be good if the comments in this function compensated for that to make sure readers understand what they're looking at.

agorrod:
@ddanderson if you make the small comment change suggested by @sueloverso I'm happy to get this merged.

ddanderson:
@agorrod, thanks for the reminder, done!

PR #4351, opened 2018-11-08
WT-4411 Added connection statistic for current total of cached cursors.

ddanderson:
Renamed data source statistic for current total.  Tweeked descriptions
of other cache cursor statistics.  Added test to verify new statistic.

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -291,7 +291,8 @@ connection_stats = [
     ##########################################
     # Cursor operations
     ##########################################
-    CursorStat('cursor_cache', 'cursors cached on close'),
+    CursorStat('cursors_cached', 'cursors currently cached', 'no_clear,no_scale'),
+    CursorStat('cursor_cache', 'cursor close calls that result in cache'),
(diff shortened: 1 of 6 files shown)


ddanderson:
@bdlucas1, the diff for `dist/stat_data.py` is hard to see in isolation.  The difference between `CursorStat` and `SessionStat` is merely the `"cursor: "` or `"session: "` prefix that appears in the description.  What is more relevant is that the first `cursors_cached` metric (newly added) is in a group of `connection_stats`.  That is indeed a current count of cached cursors global to all sessions.  That's the only new statistic.

The other `SessionStat('session_cursors_cached', ...)` is in a group called `dsrc_stats`. This one is not new, it is only renamed to more closely match the new one (and avoid confusion**).  This particular one you asked about in HELP-8101. It is the current count of cursors cached that match a particular URI (for example, the oplog).

** Since there are `cursor_cache` metrics that track number of calls to internally cache a cursor, I named the running totals to be `cursors_cached` and `session_cursors_cached` to help distinguish them.

agorrod:

PR #4353, opened 2018-11-12
WT-4421 Add a way to calculate modify operations.

michaelcahill:


src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+		if (start || i == gap) {
+			if (start) {
+				start = false;
+				gap = WT_CM_STARTGAP;
+				hstart = __cm_hash(p1);
+			} else if (i == gap) {
+				p1 += gap;
+				gap *= 2;
+				gap = WT_MIN(gap, (size_t)(
+				    cms.e1 - p1 - WT_CM_BLOCKSIZE));
```
keithbostic: gcc/clang is going to ding you for `x - y - z`

src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+	    p1++, p2++)
+		;
+
+	for (; m1 >= cms->s1 && m2 >= cms->s2 && *m1 == *m2;
+	    m1--, m2--)
+		;
+
+	match->m1 = m1 + 1;
+	match->m2 = m2 + 1;
+	match->len = p1 > m1 ? (size_t)(p1 - m1 - 1) : 0;
```
keithbostic: gcc/clang is going to ding you for `x - y - z`

src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+ *	exceeded.
+ */
+static int
+__cm_add_modify(WT_CM_STATE *cms, const uint8_t *p1, size_t len1,
+    const uint8_t *p2, size_t len2, WT_MODIFY *entries, int *nentriesp)
+{
+	WT_MODIFY *mod;
+
+	if (*nentriesp >= cms->maxentries || len2 > cms->maxdiff)
+		return (WT_NOTFOUND);
```
keithbostic: I'd go  with `EINVAL`, not "not-found" (our general "map not-found to something else" code scares me, but I'm not sure what we didn't find?

src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+	return (0);
+}
+
+/*
+ * wiredtiger_calc_modify --
+ *	Calculate a set of WT_MODIFY operations to represent an update.
+ */
+int
+wiredtiger_calc_modify(const WT_ITEM *prev, const WT_ITEM *newv,
+    size_t maxdiff, WT_MODIFY *entries, int *nentriesp)
```
keithbostic: In general, I read "prev" as a single word, not "pre version". If I'm not the only one, I'd suggest going to old/new or current/replace, here and in the documentation. Don't hesitate to ignore if it's just me.

keithbostic:
Is this work based on any published algorithms?

src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+/*
+ * __cm_hash --
+ *	Calculate a "hash" of a block of bytes.
+ */
+static uint64_t
+__cm_hash(const uint8_t *p)
+{
+	uint64_t h;
+
+	memcpy(&h, p, sizeof h);
```
keithbostic: `sizeof(h)`

src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+int
+wiredtiger_calc_modify(const WT_ITEM *prev, const WT_ITEM *newv,
+    size_t maxdiff, WT_MODIFY *entries, int *nentriesp)
+{
+	WT_CM_MATCH match;
+	WT_CM_STATE cms;
+	size_t gap, i;
+	uint64_t h, hend, hstart;
+	const uint8_t *p1, *p2, *used1, *used2;
+	int size_delta;
```
keithbostic: Shouldn't `size_delta` be a `size_t`, it's accumulating memory sizes?

src/conn/api_calc_modify.c
```diff
@@ -0,0 +1,193 @@
...
+	 * separated by a gap in the pre-image.  If the current point in the
+	 * post-image matches either marker, try to extend the match to find a
+	 * (large) range of matching bytes.  If the end of the range is reached
+	 * in the post-image without finding a good match, double the size of
+	 * the gap, update the markers and keep trying.
+	 */
+	p1 = cms.s1;
+	h = hstart = hend = 0;
+	i = gap = 0;
+	for (p2 = cms.s2, start = true;
```
keithbostic: Initialize both `p1` and `p2` in the for loop?

michaelcahill:
Thanks for the review @keithbostic -- I expect to do some more cleanup before it's really ready for review, so I'll incorporate your feedback.  As far as published algorithms, no -- I did a lot of reading but didn't find anything that quite does what I was looking for.
Useful general background: https://neil.fraser.name/writing/diff/
Closest thing I found to what I came up with: http://pop.xmailserver.net/xdfs.pdf

src/conn/api_calc_modify.c
michaelcahill: We didn't find a set of `WT_MODIFY` operations within the given constraints (maximum number of deltas, maximum bytes of difference).

src/conn/api_calc_modify.c
michaelcahill: Done

src/conn/api_calc_modify.c
michaelcahill: I've never seen this warning with any version of gcc or clang, maybe it's from lint or coverity?  Regardless, I've changed the code.

src/conn/api_calc_modify.c
michaelcahill: Yes, except it's signed (bytes can be added or removed), so `ssize_t`...

src/conn/api_calc_modify.c
michaelcahill: Done.

agorrod:
@keithbostic would you mind doing a final review of this code - and merging if you are happy with it?

PR #4356, opened 2018-11-12
WT-3756 Adjust pre-allocated file amount downward if we're not using them quickly enough.

sueloverso:
@agorrod, here's one possible algorithm to adjust the number of pre-allocated log files downward. It may be just right or too aggressive.

I have also modified a test that generates a lot of log files in a burst so that I can detect it being adjusted once it is idle.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -447,7 +447,14 @@ __log_prealloc_once(WT_SESSION_IMPL *session)
 		__wt_verbose(session, WT_VERB_LOG,
 		    "Missed %" PRIu32 ". Now pre-allocating up to %" PRIu32,
 		    log->prep_missed, conn->log_prealloc);
-	}
+	} else if (reccount > conn->log_prealloc / 2)
+		/*
+		 * If we used less than half, then adjust how many we allocate
+		 * downward. Cut in half, plus 1 so that we always allocate
+		 * at least one.
+		 */
+		conn->log_prealloc = conn->log_prealloc / 2 + 1;
+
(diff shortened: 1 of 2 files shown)


sueloverso:
autobuild: retest this please

src/conn/conn_log.c
```diff
@@ -447,7 +447,14 @@ __log_prealloc_once(WT_SESSION_IMPL *session)
...
 		    "Missed %" PRIu32 ". Now pre-allocating up to %" PRIu32,
 		    log->prep_missed, conn->log_prealloc);
-	}
+	} else if (reccount > conn->log_prealloc / 2)
+		/*
+		 * If we used less than half, then adjust how many we allocate
+		 * downward. Cut in half, plus 1 so that we always allocate
+		 * at least one.
+		 */
+		conn->log_prealloc = conn->log_prealloc / 2 + 1;
```
nehakhatri5: I noticed that at the time of checkpoint not as many PrepLog files would be renamed to log files as usual. The log server thread here would still think that so many prep logs are not required and adjust downward. Would that be right?

Also should the downward adjustment be gradual e.g. reducing one file at a time. Sometimes `reccount` might show just a temporary reduction in use of Prep logs. In that situation the `log_prealloc` might need to be re-adjusted upwards.

src/conn/conn_log.c
```diff
@@ -447,7 +447,14 @@ __log_prealloc_once(WT_SESSION_IMPL *session)
...
+
```
nehakhatri5: A log message like the one at line#447 would be useful.

PR #4360, opened 2018-11-13
WT-4159 Start timer in log_server inside loop.

sueloverso:
@ddanderson Please review this small change to make the timer in the log server thread more accurate. It is needed for PR #4356.


ddanderson:
LGTM.

PR #4361, opened 2018-11-13
WT-4409 Fix workgen throttling

ddanderson:
Fix wtperf.py to not modify the throttling number of a thread to account for log-like operations, it was doing the wrong calculation before.

Fix workgen so that small throttle numbers, or generally numbers that aren't divisible by 20 (internal THROTTLE_PER_SEC constant) are still handled.

### DIFF
--- a/bench/workgen/workgen.cxx
+++ b/bench/workgen/workgen.cxx
@@ -715,10 +715,15 @@ int ThreadRunner::op_run(Operation *op) {
     retry_op = true;
     range = op->_table.options.range;
     if (_throttle != NULL) {
-        if (_throttle_ops >= _throttle_limit && !_in_transaction) {
-            WT_ERR(_throttle->throttle(_throttle_ops,
-              &_throttle_limit));
+        while (_throttle_ops >= _throttle_limit && !_in_transaction && !_stop) {
+            // Calling throttle causes a sleep until the next time division,
+            // and we are given a new batch of operations to do before calling
+            // throttle again.  If the number of operations in the batch is
+            // zero, we'll need to go around and throttle again.
+            WT_ERR(_throttle->throttle(_throttle_ops, &_throttle_limit));
             _throttle_ops = 0;
+            if (_throttle_limit != 0)
+                break;
         }
         if (op->is_table_op())
             ++_throttle_ops;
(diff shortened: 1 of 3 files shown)


ddanderson:
@sueloverso, can you please review this?

Here's an excerpt from a run with `threads=((count=100,updates=1,throttle=8),(count=100,reads=1,throttle=32))`:
```
3216 reads, 795 inserts, 795 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (894 total secs)
3208 reads, 800 inserts, 800 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (895 total secs)
3164 reads, 800 inserts, 800 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (896 total secs)
3218 reads, 802 inserts, 802 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (897 total secs)
3206 reads, 795 inserts, 795 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (898 total secs)
3190 reads, 801 inserts, 801 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (899 total secs)
Executed 2885830 read operations (66%) 3205 ops/sec
Executed 0 not found operations (0%) 0 ops/sec
Executed 720012 insert operations (16%) 799 ops/sec
Executed 720012 update operations (16%) 799 ops/sec
Executed 0 truncate operations (0%) 0 ops/sec
Executed 0 remove operations (0%) 0 ops/sec
Executed 0 checkpoint operations (0%) 0 ops/sec
Run completed: 900.138 seconds
```

sueloverso:
@ddanderson I noticed that although the output now looks like it is correctly throttling, there is an error in the initial output. See below, the reads are throttled to 3200 and updates to 800:
```
Starting workload: 201 threads, 10000 reads, 10000 inserts, 10000 updates, 0 tru
ncates, 0 removes, 0 checkpoints
3180 reads, 799 inserts, 794 updates, 0 truncates, 0 removes, 0 checkpoints in 1
 secs (1 total secs)
```

sueloverso:
Also, interestingly enough, I'm trying out the no-checkpoint version, although `wtperf` can sustain the throttled workload for the entire 15 minutes, running `workgen` with the same is showing a significant variation in throughput.

ddanderson:
@sueloverso, as for the small skew from the throttle on the first second report (and I think it's the same for all seconds), I think this may be in the two approaches to throttling.  wtperf says: here you go, thread 0, you should do 32 operations and then sleep until the next second marker.  workgen divides the second into 20 parts and divvies out the operations, something like 2 operations the first 0.05 second, 2 operations the next...  And to make sure that threads aren't operating in lock-step, it varies the sleep amount so that it could be 0.10 seconds or it could be 0.0 seconds (or anything inbetween). It's keeping a running score of operations and where it "should" be, so it will tend to play catch-up if it's behind.  So, it should even out pretty much over time, but it may never be totally exact.  If this feels wrong or unintuitive, I could change it so it's more likely to be synchronized on the second boundaries.

PR #4362, opened 2018-11-14
WT-4419 big-endian machines incorrectly configure little-endian crc32c support

keithbostic:


src/checksum/software/checksum.c
```diff
@@ -1128,41 +1124,85 @@ __wt_checksum_sw(const void *chunk, size_t len)
...
+ * __wt_checksum_sw --
+ *	Determine if on big- or little-endian system and return a checksum for
+ * a chunk of memory, computed in software.
+ */
+uint32_t (*__wt_checksum_sw(void))(const void *, size_t)
+{
+	uint64_t v;
+
+	v = 1;
+	return (*((uint8_t *)&v) == 0 ? __checksum_sw_be : __checksum_sw_le);
```
michaelcahill: Does the fact that we're checking at runtime mean we can't rely on `WORDS_BIGENDIAN`?  Can we not figure this out during configure because of cross-compilation?

agorrod: I had the same thought, Keith answered in his commit message:

```
MongoDB had a bug, they depended on the autoconfig #define WORDS_BIGENDIAN,
which is no longer included by the crc32c code because MongoDB doesn't
necessarily have them available.
```

agorrod: On the other hand - I am a bit nervous about this check - it feels like something the compiler could break. Is it a known idiom?

keithbostic: > Does the fact that we're checking at runtime mean we can't rely on `WORDS_BIGENDIAN`?  Can we not figure this out during configure because of cross-compilation?

This code can't depend on any of the normal autoconfigure work because it's separately compiled for MongoDB services when we may not be building a WiredTiger library.

keithbostic: > On the other hand - I am a bit nervous about this check - it feels like something the compiler could break. Is it a known idiom?

I stared at the standard and searched around some, and this seems as safe/portable as anything I could find. It seems to me that as long as a `uint64_t` is larger than a `uint8_t`, it has to work (famous last words).

We use that same check in `support/global.c:__wt_endian_check()` to detect if the build being run matches the machine on which we're running, and that at least tells us that every compiler we've seen so far has no complaints.

keithbostic: And then, there's this: [Finally, one-line endianness detection in the C preprocessor](http://esr.ibiblio.org/?p=5095)

keithbostic:
Closing in favor of #4363.

PR #4363, opened 2018-11-15
WT-4419 big-endian machines incorrectly configure little-endian crc32c support

keithbostic:
Replacement for #4362.


keithbostic:
@agorrod, Gabriel confirms this branch passes s390x testing, merging now.

PR #4367, opened 2018-11-19
WT-4403 Add a statistic that tracks accumulated dirty cache content.

agorrod:


agorrod:
@sulabhM and @michaelcahill would you mind reviewing this?

dist/stat_data.py
```diff
@@ -191,6 +191,7 @@ def __init__(self, name, desc, flags=''):
     # Cache and eviction statistics
     ##########################################
     CacheStat('cache_bytes_dirty', 'tracked dirty bytes in the cache', 'no_clear,no_scale,size'),
+    CacheStat('cache_bytes_dirty_total', 'bytes that have been dirtied in the cache', 'no_clear,no_scale,size'),
```
michaelcahill: "that have been dirtied in the cache" seems clunky, how about "bytes dirty in the cache cumulative" or just "bytes dirty cumulative" (since it's already a cache statistic)?

michaelcahill: FTR, I wish "tracked dirty bytes ..." had been "bytes dirty current" or similar to keep it sorted with related stats.

agorrod:
Thanks @michaelcahill I've made the wording change you suggested and opened https://jira.mongodb.org/browse/WT-4448 to explore renaming the dirty tracked stat.

dist/stat_data.py
```diff
@@ -618,6 +619,7 @@ def __init__(self, name, desc, flags=''):
...
```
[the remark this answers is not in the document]
agorrod: Yep - thanks. Fixed.

PR #4370, opened 2018-11-20
WT-4417 Make os_cache_max and os_cache_dirty_max reconfigurable.

sueloverso:
@agorrod This is a simple change that moves `os_cache[_dirty]_max` to reconfigurable via alter(). The ticket discusses another approach which can be something done at a later date. If we do that additional work, these options still need to be reconfigurable. Please review.

### DIFF
--- a/test/suite/test_alter04.py
+++ b/test/suite/test_alter04.py
@@ -126,7 +126,7 @@ class test_alter04(wttest.WiredTigerTestCase):
         self.verify_metadata(new_param)
 
         # Run through all combinations of the alter commands
-        # for all allowed settings.  
+        # for all allowed settings.
         for a in self.cache_alter:
             alter_param = '%s=%s' % (self.setting, a)
             self.session.alter(uri, alter_param)


test/suite/test_alter04.py
```diff
@@ -0,0 +1,209 @@
...
+        c.close()
+
+        # Verify the string in the metadata
+        self.verify_metadata(new_param)
+
+        # Run through all combinations of the alter commands
+        # for all allowed settings.  This tests having only one or
+        # the other set as well as having both set.  It will also
+        # cover trying to change the setting to its current value.
+        for a in self.cache_alter:
```
sulabhM: Probably you meant dirty_alter here instead of cache_alter

test/suite/test_alter04.py
```diff
@@ -0,0 +1,209 @@
...
+            alter_param = ''
+            if a != '':
+                new_str = 'os_cache_max=%s' % a
+                alter_param += '%s' % new_str
+                if alter_param != '':
```
sulabhM: alter_param will never be ' ' here, since it will be at least 'os_cache_max=<something>. Am is missing something ?

test/suite/test_alter04.py
```diff
@@ -0,0 +1,209 @@
...
+                        self.reopen_conn()
+                    special = self.use_cg or self.use_index
+                    if not special:
+                        self.verify_metadata(new_str)
+                    else:
+                        self.session.alter(suburi, alter_param)
+                        self.verify_metadata(new_str)
+
+    # Alter: Change the os_cache_dirty_max setting after creation
+    def test_alter04_dirty(self):
```
sulabhM: test_alter04_dirty and test_alter04_cache are exactly the same other than new_param using 'os_cache_dirty_max' vs 'os_cache_max' and testing over cache_alter vs dirty_alter. If I have followed the test correctly I think we can make a single routine for the test that passes these parameters separately hence not duplicating the code.

But it might be harder than it looks with create_params and complex_params depending on the new_param and maybe not worth it for python testing.

I don't have a strong opinion about reducing code duplication here.

sueloverso:
@sulabhM Thanks for the good review. I have made the changes and consolidations. It is much better for it.

sulabhM:
@agorrod , ready for your review

PR #4371, opened 2018-11-21
WT-4381 Reset session statistics on WT_SESSION::reset()

nehakhatri5:


nehakhatri5:
@bvpvamsikrishna Please review this change.

PR #4373, opened 2018-11-22
WT-4552 big-endian checksums being written on zSeries platforms.

keithbostic:
@agorrod, the WiredTiger library assumed that checksums were written in big-endian format on big-endian hardware, and that's not the case: both the software and hardware implementations return checksums in little-endian format, even on big-endian hardware.

There's some cleanup and bug fix changes in the `salvage` test program, and clarifying the library verbose and error messages, and then the bug fix commit is f280c8a.


keithbostic:
FTR, this isn't yet passing all of the standard tests, at least `csuite/test_wt4156_metadata_salvage` is failing.

agorrod:
@keithbostic should we hold off even merging this PR until the automated testing @lukech is working on is ready to be enabled?

keithbostic:
Yes, that sounds right to me. This code has only been very lightly tested at this point, and I don’t see any gain in moving quickly?

keithbostic:
autobuild: retest this please

PR #4374, opened 2018-11-23
WT-4410 Split 'unit-test' task to reduce Evergreen Ubuntu build variant runtime

lukech:
Split 'unit-test' Evergreen task (Python test/suite) into 8 buckets/tasks to reduce Makespan for the WiredTiger Ubuntu build variant. The split was based on test name (alphabetic order for the 1st letter after 'test_' prefix) and test runtime, a few alphabetics were grouped together to get them 'relatively' evenly-distributed runtime for each bucket/task. 

There's a plan to implement a mechanism to auto-group and auto-generate buckets/tasks based on historic runtime for each tests, which would be covered later in WT-4441. Added a REAME under test/suite to describe the situation and plan. 

A patch build of the change is [here](https://evergreen.mongodb.com/version/5bf745092a60ed72f37df437).

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -808,7 +808,7 @@ tasks:
             set -o verbose
 
             cd .libs
-            ${python_test_env_vars|} python ../test/suite/run.py [de] -v 2 ${smp_command|} 2>&1
+            ${python_test_env_vars|} python ../test/suite/run.py [defg] -v 2 ${smp_command|} 2>&1
 
   - name: unit-test-bucket03
     depends_on:
@@ -823,7 +823,7 @@ tasks:
             set -o verbose
 
             cd .libs
-            ${python_test_env_vars|} python ../test/suite/run.py [hij] -v 2 ${smp_command|} 2>&1
+            ${python_test_env_vars|} python ../test/suite/run.py [hijk] -v 2 ${smp_command|} 2>&1
 
   - name: unit-test-bucket04
     depends_on:
@@ -838,7 +838,7 @@ tasks:
             set -o verbose
 
             cd .libs
-            ${python_test_env_vars|} python ../test/suite/run.py [lmnop] -v 2 ${smp_command|} 2>&1
+            ${python_test_env_vars|} python ../test/suite/run.py [lmnopq] -v 2 ${smp_command|} 2>&1
 
   - name: unit-test-bucket05
     depends_on:
@@ -883,7 +883,7 @@ tasks:
             set -o verbose
 
             cd .libs
-            ${python_test_env_vars|} python ../test/suite/run.py [uv] -v 2 ${smp_command|} 2>&1
+            ${python_test_env_vars|} python ../test/suite/run.py [uvwxyz] -v 2 ${smp_command|} 2>&1
 
   # End of Python unit test tasks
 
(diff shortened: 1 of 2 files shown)


test/evergreen.yml
```diff
@@ -749,21 +751,141 @@ tasks:
...
+      - func: "fetch artifacts"
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            cd .libs
+            ${python_test_env_vars|} python ../test/suite/run.py [ab] -v 2 ${smp_command|} 2>&1
```
agorrod: It bothers me that there are missing letters in the buckets - if we add a test with one of those prefixes it wouldn't be run, and we wouldn't notice.

test/suite/README
```diff
@@ -0,0 +1,13 @@
+The test/suite directory includes a collection of Python unit tests
```
agorrod: Thanks for adding this - as described above, I'd like to change the bucketing so it won't miss new tests.

I'd meant to ask for a README in the csuite directory - describing what we'd done there.

lukech: The README for csuite would be covered by WT-4454.

test/evergreen.yml
sulabhM: I agree with Alex, we should maybe club all the missing ones together for any test we add in the future

test/evergreen.yml
```diff
@@ -749,21 +751,141 @@ tasks:
...
-              ${test_env_vars|} python ./test/suite/run.py -v 2 ${smp_command|} 2>&1
-            else # Ubuntu
-              # Change directory to where the local installed 'wt' binary is located, 
-              # to avoid libtool generated 'wt' script from being selected by run.py,
-              # which invokes relink_command that tries to changing to a non-existed 
-              # /data/mci/<uniq> directory, as 'make' is done by a separate 'compile' task.
-              cd .libs
-              ${python_test_env_vars|} python ../test/suite/run.py -v 2 ${smp_command|} 2>&1
-            fi
+            # Only Windows and OS X variants are expected to run this task
```
sulabhM: I do not understand, how are we limiting this task to run on windows and osx here.

test/evergreen.yml
lukech: @agorrod I checked into the `test/suite/run.py` program and found this wild-casting [line](https://github.com/wiredtiger/wiredtiger/blob/develop/test/suite/run.py#L216) could help us achieve the wanted result by simply appending those missing letters to those existing letters in the existing 8 buckets. E.g. by adding missing letters 'f' and 'g' into bucket02 to let it run `${python_test_env_vars|} python ../test/suite/run.py [defg] -v 2 ${smp_command|} 2>&1`, the program will attempt to match `test_[defg]*.py`, and won't error out because 'd' and 'e' guarantee the existence of matched files. I pushed a commit just now which passed the patch build [here](https://evergreen.mongodb.com/version/5bfb30f1e3c331687f38bb11). Please let me know if you like this change.

test/evergreen.yml
lukech: The scope of this ticket (and recent works) is for Ubuntu build variant, and we plan to cover Windows & OS X variants later in a separate effort. Before that happens, I'd keep both Windows and OS X running the old `unit-test` task (the un-split Python suite tests).

agorrod:
lgtm

PR #4376, opened 2018-11-26
WT-4383 Update session statistics to reflect the operation statistics.

nehakhatri5:
This change set increments the new session statistics at the desired locations. It does not add any decrement macro yet because s_define complains about the unused macros.

### DIFF
--- a/src/include/mutex.h
+++ b/src/include/mutex.h
@@ -53,6 +53,7 @@ struct __wt_rwlock {			/* Read/write lock */
 	int16_t stat_write_count_off;	/* write acquisitions offset */
 	int16_t stat_app_usecs_off;	/* waiting application threads offset */
 	int16_t stat_int_usecs_off;	/* waiting server threads offset */
+	int16_t stat_session_usecs_off;	/* waiting session offset */
 
 	WT_CONDVAR *cond_readers;	/* Blocking readers */
 	WT_CONDVAR *cond_writers;	/* Blocking writers */
@@ -65,17 +66,24 @@ struct __wt_rwlock {			/* Read/write lock */
  * Implemented as a macro so we can pass in a statistics field and convert
  * it into a statistics structure array offset.
  */
-#define	WT_RWLOCK_INIT_TRACKED(session, l, name) do {                   \
-	WT_RET(__wt_rwlock_init(session, l));                           \
-	(l)->stat_read_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(   \
-	    S2C(session)->stats, lock_##name##_read_count);             \
-	(l)->stat_write_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(  \
-	    S2C(session)->stats, lock_##name##_write_count);            \
-	(l)->stat_app_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(    \
-	    S2C(session)->stats, lock_##name##_wait_application);       \
-	(l)->stat_int_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(    \
-	    S2C(session)->stats, lock_##name##_wait_internal);          \
+#define	WT_RWLOCK_INIT_TRACKED(session, l, name) do {			\
+	WT_RET(__wt_rwlock_init(session, l));				\
+	(l)->stat_read_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_read_count);		\
+	(l)->stat_write_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_write_count);		\
+	(l)->stat_app_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_wait_application);	\
+	(l)->stat_int_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_wait_internal);		\
 } while (0)
+
+#define	WT_RWLOCK_INIT_SESSION_TRACKED(session, l, name) {		\
+	WT_RWLOCK_INIT_TRACKED(session, l, name);			\
+	(l)->stat_session_usecs_off =					\
+	    (int16_t)WT_SESSION_STATS_FIELD_TO_OFFSET(			\
+	    &(session->stats), lock_##name##_wait);			\
+}
 
 /*
  * Spin locks:
@@ -112,6 +120,7 @@ struct __wt_spinlock {
 	int16_t stat_count_off;		/* acquisitions offset */
 	int16_t stat_app_usecs_off;	/* waiting application threads offset */
 	int16_t stat_int_usecs_off;	/* waiting server threads offset */
+	int16_t stat_session_usecs_off;	/* waiting session offset */
 
 	int8_t initialized;		/* Lock initialized, for cleanup */
 
--- a/src/include/mutex.h
+++ b/src/include/mutex.h
@@ -78,12 +78,12 @@ struct __wt_rwlock {			/* Read/write lock */
 	    S2C(session)->stats, lock_##name##_wait_internal);		\
 } while (0)
 
-#define	WT_RWLOCK_INIT_SESSION_TRACKED(session, l, name) {		\
+#define	WT_RWLOCK_INIT_SESSION_TRACKED(session, l, name) do {		\
 	WT_RWLOCK_INIT_TRACKED(session, l, name);			\
 	(l)->stat_session_usecs_off =					\
 	    (int16_t)WT_SESSION_STATS_FIELD_TO_OFFSET(			\
 	    &(session->stats), lock_##name##_wait);			\
-}
+} while (0)
 
 /*
  * Spin locks:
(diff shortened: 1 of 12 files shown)


src/include/mutex.h
```diff
@@ -65,18 +66,25 @@ struct __wt_rwlock {			/* Read/write lock */
...
+	    S2C(session)->stats, lock_##name##_read_count);		\
+	(l)->stat_write_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_write_count);		\
+	(l)->stat_app_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_wait_application);	\
+	(l)->stat_int_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_wait_internal);		\
 } while (0)
 
+#define	WT_RWLOCK_INIT_SESSION_TRACKED(session, l, name) {		\
```
sulabhM: The macro needs a "do .. while(0)" to make it act like a single statement, say when used in a single statement conditional.

src/include/mutex.i
```diff
@@ -286,6 +286,13 @@ __wt_spin_unlock(WT_SESSION_IMPL *session, WT_SPINLOCK *t)
 	    S2C(session)->stats, lock_##name##_wait_internal);		\
 } while (0)
 
+#define	WT_SPIN_INIT_SESSION_TRACKED(session, t, name) {                \
```
sulabhM: same as above .. needs a do while(0)

src/include/mutex.i
```diff
@@ -286,6 +286,13 @@ __wt_spin_unlock(WT_SESSION_IMPL *session, WT_SPINLOCK *t)
...
+	WT_SPIN_INIT_TRACKED(session, t, name);                         \
+	(t)->stat_session_usecs_off = \
```
sulabhM: the '\' at the end of this line isn't aligned properly.

src/include/mutex.i
```diff
@@ -294,20 +301,24 @@ static inline void
...
+		session_stats = (int64_t *)&(session->stats);
 		stats[session->stat_bucket][t->stat_count_off]++;
 		if (F_ISSET(session, WT_SESSION_INTERNAL))
 			stats[session->stat_bucket][t->stat_int_usecs_off] +=
 			    (int64_t)WT_CLOCKDIFF_US(time_stop, time_start);
-		else
+		else {
 			stats[session->stat_bucket][t->stat_app_usecs_off] +=
 			    (int64_t)WT_CLOCKDIFF_US(time_stop, time_start);
+			session_stats[t->stat_session_usecs_off] +=
```
sulabhM: @bvpvamsikrishna , have we decided to not maintain these statistics for internal sessions ?

bvpvamsikrishna: I think it is better to track these statistics for internal sessions as well as other statistics we are not differentiating. This will also help us in diagnosis of some scenario's.

src/support/mtx_rw.c
```diff
@@ -260,12 +262,16 @@ stall:			__wt_cond_wait(session,
...
+		time_diff = WT_CLOCKDIFF_US(time_stop, time_start);
 		if (F_ISSET(session, WT_SESSION_INTERNAL))
 			stats[session->stat_bucket][l->stat_int_usecs_off] +=
-			    (int64_t)WT_CLOCKDIFF_US(time_stop, time_start);
-		else
+			    (int64_t)time_diff;
+		else {
 			stats[session->stat_bucket][l->stat_app_usecs_off] +=
-			    (int64_t)WT_CLOCKDIFF_US(time_stop, time_start);
+			    (int64_t)time_diff;
```
bvpvamsikrishna: Better to track this statistics to internal sessions as well.

src/support/mtx_rw.c
```diff
@@ -260,12 +262,16 @@ stall:			__wt_cond_wait(session,
...
 	if (set_stats) {
 		time_stop = __wt_clock(session);
+		time_diff = WT_CLOCKDIFF_US(time_stop, time_start);
 		if (F_ISSET(session, WT_SESSION_INTERNAL))
 			stats[session->stat_bucket][l->stat_int_usecs_off] +=
-			    (int64_t)WT_CLOCKDIFF_US(time_stop, time_start);
-		else
+			    (int64_t)time_diff;
+		else {
 			stats[session->stat_bucket][l->stat_app_usecs_off] +=
```
sulabhM: same as above, have we decided not to track this stat for the internal sessions?

src/support/mtx_rw.c
```diff
@@ -433,12 +441,16 @@ __wt_writelock(WT_SESSION_IMPL *session, WT_RWLOCK *l)
...
-			    (int64_t)WT_CLOCKDIFF_US(time_stop, time_start);
+			    (int64_t)time_diff;
```
sulabhM: same as above, have we decided not to track this stat for the internal sessions?

test/suite/test_schema06.py
```diff
@@ -42,6 +42,10 @@ class test_schema06(wttest.WiredTigerTestCase):
         ('lsm', { 'idx_config' : ',type=lsm' }),
     ])
 
+    # Enable statistics collection for session level schema lock statistics test.
```
bvpvamsikrishna: comment can be updated, currently we don't have enable/disable for session level statistics.

test/suite/test_schema06.py
```diff
@@ -87,6 +91,14 @@ def test_index_stress(self):
         self.drop_index(0)
         self.drop_index(1)
 
+        # Test the session level schema lock wait statistic.
```
sulabhM: can you add a comment why we test only lsm here

test/suite/test_schema06.py
```diff
@@ -87,6 +91,14 @@ def test_index_stress(self):
...
+        if self.idx_config == ',type=lsm':
+            stat_cur = self.session.open_cursor('statistics:session', None, None)
+            while stat_cur.next() == 0:
+                [desc, pvalue, stat_value] = stat_cur.get_values()
+                if desc == 'session: schema lock wait time (usecs)':
```
sulabhM: What happens if we do not find 'session: schema lock wait time (usecs)' in the stats, eg if the description of this stat gets changed tomorrow. Maybe we should also check if we found this stat and assert if we dont.

test/suite/test_stat08.py
```diff
@@ -0,0 +1,89 @@
...
+                            "key_format=i,value_format=S")
+        cursor =  self.session.open_cursor('table:test_stat08', None, None)
+        # Write the entries.
+        for i in range(0, self.nentries):
+            cursor[i] = self.entry_value
+        cursor.reset()
+
+        # Read the entries.
+        for key, value in cursor:
+            self.pr('read %d -> %s' % (key, value))
```
sulabhM: This will generate over 350000 lines in results.txt and I don't think we really need it.

test/suite/test_stat08.py
```diff
@@ -0,0 +1,89 @@
...
+        cursor.reset()
+
+        stat_cur = self.session.open_cursor('statistics:session', None, None)
+        while stat_cur.next() == 0:
+            [desc, pvalue, value] = stat_cur.get_values()
+            self.printVerbose(2, '  stat: \'' + desc + '\', \'' +
+                              pvalue + '\', ' + str(value))
+            if desc == 'session: time waiting for cache (usecs)' or \
```
sulabhM: If we change any of the descriptions for the stats, this test will miss checking on that statistics.

test/suite/test_stat08.py
```diff
@@ -0,0 +1,89 @@
...
+        stat_cur.reset()
+        for i in range(0, 10000):
+            cursor[i + self.nentries] = self.entry_value
+        self.session.checkpoint()
+
+        while stat_cur.next() == 0:
+            [desc, pvalue, value] = stat_cur.get_values()
+            self.printVerbose(2, '  stat: \'' + desc + '\', \'' +
+                              pvalue + '\', ' + str(value))
+            if desc == 'session: bytes written from cache':
```
sulabhM: Same as above, we should explore ways of not relying on these stats descriptions to stay the same OR build a way for these tests to fail if the descriptions change tomorrow.

src/include/mutex.i
bvpvamsikrishna: No need for us differentiate between internal or external sessions. We collect stats for all sessions.

src/include/mutex.i
nehakhatri5: I tried aligning the \ , but s_all always brings it back to where it is.

src/include/mutex.h
```diff
@@ -65,16 +66,23 @@ struct __wt_rwlock {			/* Read/write lock */
...
+	WT_RET(__wt_rwlock_init(session, l));				\
+	(l)->stat_read_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_read_count);		\
+	(l)->stat_write_count_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_write_count);		\
+	(l)->stat_app_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_wait_application);	\
+	(l)->stat_int_usecs_off = (int16_t)WT_STATS_FIELD_TO_OFFSET(	\
+	    S2C(session)->stats, lock_##name##_wait_internal);		\
+} while (0)
```
agorrod: I don't understand what changed in these lines - was there a different spacing?

nehakhatri5: It replaced spaces with tabs.

nehakhatri5:
@agorrod I have updated the python test to only check the bytes read. It passes on zseries. This change set is good to go now.

agorrod:
Thanks @nehakhatri5

PR #4377, opened 2018-11-27
WT-4434-update-zstd-compression-level

bvpvamsikrishna:
Updated the zstd compression level from 3 to 6.
I believe this change will not have any backward compatibility issue.


bvpvamsikrishna:
need to rename commit to match WT

PR #4381, opened 2018-11-27
WT-4421 wiredtiger calc modify test

keithbostic:
@michaelcahill, @sulabhM, when I was looking at WT-4421 today, I hooked the new API into the `csuite/wt3338_partial_update` test program (that's code that tests if the underlying WiredTiger modify API does the right thing).

This branch fixes a minor problem, if the old/new data sizes are 0, in commit b023cb0.

### DIFF
--- a/src/conn/api_calc_modify.c
+++ b/src/conn/api_calc_modify.c
@@ -116,7 +116,7 @@ wiredtiger_calc_modify(WT_SESSION *wt_session,
 	const uint8_t *p1, *p2;
 	bool start;
 
-	if (oldv->size == 0 || newv->size == 0)
+	if (oldv->size < WT_CM_MINMATCH || newv->size < WT_CM_MINMATCH)
 		return (WT_NOTFOUND);
 
 	cms.session = (WT_SESSION_IMPL *)wt_session;
--- a/src/conn/api_calc_modify.c
+++ b/src/conn/api_calc_modify.c
@@ -116,8 +116,6 @@ wiredtiger_calc_modify(WT_SESSION *wt_session,
 	const uint8_t *p1, *p2;
 	bool start;
 
-	*nentriesp = 0;
-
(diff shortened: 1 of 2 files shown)


keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

src/conn/api_calc_modify.c
```diff
@@ -116,6 +116,11 @@ wiredtiger_calc_modify(WT_SESSION *wt_session,
 	const uint8_t *p1, *p2;
 	bool start;
 
+	*nentriesp = 0;
+
+	if (oldv->size == 0 || newv->size == 0)
+		return (WT_NOTFOUND);
```
michaelcahill: This could be extended to `if (oldv->size <= WT_CM_MINMATCH || newv->size <= WT_CM_MINMATCH)`

src/conn/api_calc_modify.c
```diff
@@ -124,7 +129,6 @@ wiredtiger_calc_modify(WT_SESSION *wt_session,
 	cms.e2 = cms.s2 + newv->size;
 	cms.maxdiff = maxdiff;
 	cms.maxentries = *nentriesp;
-	*nentriesp = 0;
```
michaelcahill: Moving this introduces a bug.  `*nentriesp` is an in/out parameter.  The input value needs to be copied before it is overwritten.  We don't make any promises about the return value when the call fails, so I don't see any reason to move it.

keithbostic: Ouch, sorry -- reverted.

src/conn/api_calc_modify.c
keithbostic: Pushed in dcc612f.

agorrod:
Thanks @sulabhM and @keithbostic lgtm - I'll go ahead and merge.

PR #4383, opened 2018-11-28
WT-4442 Add duplicate backup cursor functionality.

sueloverso:
Not yet ready for merge.


sueloverso:
@agorrod and @ddanderson the duplicate backup cursor change is ready for review. I think the code in `session_api.c` needs a clean-up, refactoring. I'll open a tech-debt ticket for that. I'd prefer to unblock this current project sooner than it would take to do that.

sueloverso:
(Oh and I goofed on the ticket number - I had it in my head that it was WT-4452 instead of WT-4442. I edited the ticket title to the correct number, but the branch name is incorrect.)

src/cursor/cur_backup.c
```diff
@@ -216,48 +239,67 @@ __backup_start(
...
+		 * generating the list of targets in backup_uri.  This file will
+		 * later be renamed to the correct name depending on whether or
+		 * not we're doing an incremental backup.  We need a temp file
+		 * so that if we fail or crash while filling it, the existence
+		 * of a partial file doesn't confuse restarting in the source
+		 * database.
+		 */
+		WT_ERR(__wt_fopen(session, WT_BACKUP_TMP,
+		    WT_FS_OPEN_CREATE, WT_STREAM_WRITE, &cb->bfs));
+	} else
```
ddanderson: This is a no-op assert.  Maybe you want to assert that other has this flag?

src/cursor/cur_backup.c
```diff
@@ -89,7 +106,13 @@ __curbackup_close(WT_CURSOR *cursor)
 	 * discarded when the cursor is closed), because that cursor will never
 	 * not be responsible for cleanup.
 	 */
-	if (F_ISSET(cb, WT_CURBACKUP_LOCKER))
+	if (F_ISSET(cb, WT_CURBACKUP_DUP)) {
+		__backup_free(session, cb);
+		/* Make sure the original backup cursor is still open. */
+		WT_ASSERT(session, F_ISSET(session, WT_SESSION_BACKUP_CURSOR));
```
ddanderson: Can there be more than one dup cursor?  If so, then a single session flag for backup dup is not going to work.

src/cursor/cur_backup.c
```diff
@@ -216,48 +239,67 @@ __backup_start(
...
 	/*
 	 * If targets were specified, add them to the list. Otherwise it is a
 	 * full backup, add all database objects and log files to the list.
 	 */
 	target_list = false;
 	WT_ERR(__backup_uri(session, cfg, &target_list, &log_only));
+	/*
+	 * We're either a top-level cursor, or if a duplicate cursor the target
+	 * must be log only.
+	 */
```
ddanderson: What's to prevent user from making a dup cursor that is not configured log_only?  That is, shouldn't the assertion be a check instead?

test/suite/test_backup10.py
```diff
@@ -0,0 +1,138 @@
...
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+
+import wiredtiger, wttest
+import os, shutil
+from helper import compare_files
+from suite_subprocess import suite_subprocess
+from wtdataset import simple_key
+from wtscenario import make_scenarios
+
```
ddanderson: If multiple duplicates of a backup cursor are allowed, it should be tested.  Also making a duplicate of a duplicate.  If not supported, there should still be tests (and checks within the code to disallow).

test/suite/test_backup10.py
```diff
@@ -0,0 +1,138 @@
...
+# test_backup10.py
+# Test cursor backup with a duplicate backup cursor.
+class test_backup10(wttest.WiredTigerTestCase, suite_subprocess):
+    dir='backup.dir'                    # Backup directory name
+    logmax="100K"
+    newuri="table:newtable"
+
+    pfx = 'test_backup'
```
ddanderson: Scenarios are generally not needed if there is just one.

src/cursor/cur_backup.c
sueloverso: I don't think so. For now I'll forbid it and we can support it later if necessary.

src/cursor/cur_backup.c
sueloverso: I just removed it.

src/cursor/cur_backup.c
sueloverso: Fixed.

test/suite/test_backup10.py
sueloverso: Fixed.

test/suite/test_backup10.py
sueloverso: Added several new error case tests.

PR #4388, opened 2018-11-30
WT-4454 Auto-generate Evergreen configuration when adding new 'make check' or csuite tests

lukech:
The changes in this PR include: 
- A new Python CLI script `test/evergreen/evg_cfg.py` that supports `check` (to identify missing tests in Evergreen configuration file) and `generate` (to generate and apply Evergreen configuration for missing tests) action types, both support 3 different test types - `make_check`, `csuite`, and `all`. 
- Hook the `check` action call of the new script into developer workflow through `dist/s_evergreen`, which will be called by `dist/s_all`.
- A README for `test/csuite` 

Some sample output for the script runs can be found [here](https://jira.mongodb.org/browse/WT-4454?focusedCommentId=2076413&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2076413).


agorrod:
@keithbostic would you please take a look at this change?

test/evergreen/evg_cfg.py
```diff
@@ -0,0 +1,335 @@
...
+
+"""
+This program provides an CLI interface to check and generate Evergreen configuration.
+"""
+
+import os
+import sys
+import re
+import subprocess
+import docopt
```
keithbostic: My system didn't have `docopt` installed:
```
Traceback (most recent call last):
  File "/home/bostic/wiredtiger/test/evergreen/evg_cfg.py", line 11, in <module>
    import docopt
ModuleNotFoundError: No module named 'docopt'
```
We might want to add better error handling if this is a common/expected problem.

test/evergreen/evg_cfg.py
```diff
@@ -0,0 +1,335 @@
...
+            for t in TEST_TYPES:
+                # Aggregate the missing tests for each test type together here
+                missing_tests.update(check_missing_tests(t))
+        else:
+            sys.exit("Unsupported test type '%s'" % test_type)
+
+        # If any missing test is identified, prompt users to run the 'generate' action
+        # which will auto-generate the Evergreen configuration for those missing tests.
+        if missing_tests:
+            prompt = ("\n*** Some tests are missing in Evergreen configuraiton ***\nPlease\n" +
```
keithbostic: typo `configuraiton`

test/evergreen/evg_cfg.py
```diff
@@ -0,0 +1,335 @@
...
+    return csuite_dirs
+
+def check_missing_tests(test_type):
+    """
+    Check to see if any tests are missing from the Evergreen configuration.
+    Loop through the list of directories in 'Make.subdirs' file and skip a few known
+    directories that do not require any test.
+    """
+
+    # Retrive the directories that are applicable for testing based on test type
```
keithbostic: typo `Retrive`

test/evergreen/evg_cfg.py
```diff
@@ -0,0 +1,335 @@
...
+        elif test_type == 'csuite':
+            # The Evergreen task name for each 'csuite' test is worked out from sub directory name
+            # E.g. for 'test/csuite' sub directory 'wt3184_dup_index_collator', the corresponding
+            # Evergreen task name will be 'csuite-wt3184-dup-index-collator-test'.
+
+            evg_task_name = 'csuite-' + d.replace('_', '-') + '-test'
+            debug("Evergreen task name for csuite sub directory '%s' is: %s" % (d, evg_task_name))
+
+        else:
+            sys.exit("Unknow test_type '%s'" % test_type)
```
keithbostic: typo `Unknow`

test/evergreen/evg_cfg.py
```diff
@@ -0,0 +1,335 @@
...
+    "ext/extractors/csv",
+    "ext/test/kvs_bdb",
+    "ext/test/fail_fs",
+    "api/leveldb",
+    "lang/java",
+    "test/utility",
+    "examples/java",
+    "test/csuite",  # csuite has its own set of Evergreen tasks, skip the checking here
+    "test/syscall",
+    "bench/workgen",
```
keithbostic: This list is long enough I think it should be sorted.

keithbostic: How was this list generated? I'm confused, because I doubt `lang/python` has `make check` testing, either? My guess is there are more directories that don't involve `make check` testing, than do involve it.

That said, could we instead check the `Makefile` in a directory, and see if there's a `TESTS =` target?

Or, have an empty set of tests in directories that don't include testing so we don't have to skip directories?

dist/s_all
```diff
@@ -100,6 +100,7 @@ COMMANDS="
 2>&1 ./s_void > ${t_pfx}s_void
 2>&1 ./s_whitespace > ${t_pfx}s_whitespace
 2>&1 ./s_win > ${t_pfx}s_win
+2>&1 ./s_evergreen > ${t_pfx}s_evergreen
```
keithbostic: This list is sorted.

lukech: Ack.

test/evergreen/evg_cfg.py
lukech: Good suggestion. Will add some error handling here.

test/evergreen/evg_cfg.py
lukech: Ack.

lukech: Here is the set of directories I got while grep `TESTS =` from Makefile.am as below. `api/leveldb` showed up here, but was not in the `make check` testing scope for some reason. All other directories make sense (`test/csuite` is split up separately and will be an exception we need to skip). I'll go ahead implementing this solution, and likely need to keep the skip list (with limited number of entries though) to exclude `api/leveldb` and `test/csuite`.

> $ find . -name Makefile.am -exec grep -H "^TESTS =" {} \;
> ./api/leveldb/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./bench/wtperf/Makefile.am:TESTS = smoke.sh
> ./examples/c/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./lang/python/Makefile.am:TESTS = run-ex_access
> ./test/bloom/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./test/checkpoint/Makefile.am:TESTS = smoke.sh
> ./test/csuite/Makefile.am:TESTS = $(all_TESTS)
> ./test/cursor_order/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./test/fops/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./test/format/Makefile.am:TESTS = smoke.sh
> ./test/huge/Makefile.am:TESTS = smoke.sh
> ./test/manydbs/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./test/packing/Makefile.am:TESTS = smoke.sh
> ./test/readonly/Makefile.am:TESTS = smoke.sh
> ./test/salvage/Makefile.am:TESTS = $(noinst_PROGRAMS)
> ./test/thread/Makefile.am:TESTS = smoke.sh

lukech:
@keithbostic Please see my rework in the 3 latest commits. Changed the way of figuring out `make check` test directories by probing `Makefile.am` for `TESTS =` keyword as suggested. Also enhanced the error handling for Python module missing.

test/evergreen/evg_cfg.py
```diff
@@ -0,0 +1,327 @@
...
+CSUITE_TEST_TMPLT = "test/evergreen/csuite_test_evg_task.template"
+MAKE_CHECK_TEST_SEARCH_STR = "  # End of normal make check test tasks"
+CSUITE_TEST_SEARCH_STR = "  # End of csuite test tasks"
+WIREDTIGER_REPO = "git@github.com:wiredtiger/wiredtiger.git"
+
+# This list of sub directories will be skipped from checking.
+# They are not expected to trigger any 'make check' testing.
+make_check_subdir_skips = [
+    "api/leveldb",  # no need to test
+    "test/csuite",  # csuite has its own set of Evergreen tasks, skip the checking here
```
keithbostic: Should this be `test/`, since the evergreen file is at that level, not at `test/csuite`?

lukech: `test/csuite` will be queried out in the new approach as its Makefile.am contains `TESTS =`. The reason the `test/csuite` (make check) directory is put into skip list is the csuite tests under it had been extended to separate Evergreen tasks, and we don't want to duplicate the csuite tests again with a `csuite-test` Evergreen task. The change is aimed to advise updates needed for `test/evergreen.yml` - the main Evergreen configuration file for WiredTiger (Evergreen) project - when new make check test applicable directory is introduced.

lukech:
@keithbostic That's actually a good point. It looks after WT-4470 and WT-4463 all references to`TESTUTIL_ENABLE_LONG_TESTS` in test code had been removed. There's no point to leave the references in Evergreen configuration then. Will remove remaining references from `test/evergreen.yml`. Thanks!

lukech:
The patch [build](https://evergreen.mongodb.com/version/5c0861ce2a60ed69298c9604) with all references to `TESTUTIL_ENABLE_LONG_TESTS ` removed from `test/evergreen.yml` had a comparable turnaround time compared to without this change. Will go ahead to merge.

PR #4391, opened 2018-12-02
WT-4470 Minimize testing that requires TESTUTIL_ENABLE_LONG_TESTS

keithbostic:
@lukech, @agorrod, with these changes, I think we're close to getting rid of `TESTUTIL_ENABLE_LONG_TESTS`.

The only remaining uses of `TESTUTIL_ENABLE_LONG_TESTS` are:
```
test/evergreen.yml
test/csuite/time_shift_test.sh
```

@bvpvamsikrishna, I'm not sure what's going on with `time_shift_test.sh`. I can't find any place in the system that runs that shell script, and when I tried to run it, it hung on me.

Most of the commits simply remove checks for `TESTUTIL_ENABLE_LONG_TESTS`, because the tests involved were almost always reasonably fast, there wasn't any point in limiting the tests to "long" runs.

There were two tests I changed in other ways:

* 675b0c8: I increased the run-time of `wt2535_insert_race`, and changed it so the threads were queued behind a gate and didn't start until all of them were ready to run, to avoid them executing serially. This is more of a bug fix than anything else, the test wasn't really doing anything useful.

* 50519a3: I reworked `wt2909_checkpoint_integrity` in a couple of ways, first to reduce the amount of testing done to find a midpoint, and second, to only do a single test run per value near the midpoint. @ddanderson, this probably deserves real review.

### DIFF
--- a/test/checkpoint/smoke.sh
+++ b/test/checkpoint/smoke.sh
@@ -6,9 +6,6 @@ set -e
 echo "checkpoint: 3 mixed tables"
 $TEST_WRAPPER ./t -T 3 -t m
 
-# We are done unless long tests are enabled.
-test "$TESTUTIL_ENABLE_LONG_TESTS" = "1" || exit 0
-
 echo "checkpoint: 6 column-store tables"
 $TEST_WRAPPER ./t -T 6 -t c
 
--- a/test/csuite/rwlock/main.c
+++ b/test/csuite/rwlock/main.c
@@ -55,10 +55,6 @@ main(int argc, char *argv[])
 	pthread_t dump_id, id[MAX_THREADS];
 	int i;
 
-	/* Ignore unless requested */
-	if (!testutil_is_flag_set("TESTUTIL_ENABLE_LONG_TESTS"))
-		return (EXIT_SUCCESS);
-
 	opts = &_opts;
 	memset(opts, 0, sizeof(*opts));
 	opts->nthreads = 100;
@@ -110,7 +106,8 @@ thread_rwlock(void *arg)
 	    opts->conn->open_session(opts->conn, NULL, NULL, &wt_session));
 	session = (WT_SESSION_IMPL *)wt_session;
 
-	printf("Running rwlock thread\n");
+	if (opts->verbose)
+		printf("Running rwlock thread\n");
 	for (i = 1; i <= opts->nops; ++i) {
 		writelock = (i % READS_PER_WRITE == 0);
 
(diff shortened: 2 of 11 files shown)


ddanderson:
LGTM.  I mainly looked at changes for `wt2909_checkpoint_integrity`.  I like this change - if a bug in this area appears, we may not catch it quite as fast as running the 'long' version of the original test (doing 3 times at each point), but since we'll be running the test more often as the 'short' version, it should more than even out.

agorrod:
I've switched the PPC Jenkins job to this branch:
http://build.wiredtiger.com:8080/job/wiredtiger-test-unit-ppc/1652/

If the runtime there doesn't change significantly I'm good to go with merging this.

lukech:
@keithbostic Thanks for making the changes to help reducing turnaround time for csuite tests. I'd like to provide some background info for `test/evergreen.yml`:  This file was named `test/mciproject.yml`, and recently I made some updates (including rename the filename) to split csuite tests into separate Evergreen tasks and get `TESTUTIL_ENABLE_LONG_TESTS` set for each applicable test in Evergreen configuration. So, if a decision were made to remove `TESTUTIL_ENABLE_LONG_TESTS` setting for one csuite test, the corresponding line in `test/evergreen.yml` for the same test should be updated (removing the `TESTUTIL_ENABLE_LONG_TESTS` setting).

keithbostic:
@ddanderson, would you please re-review the changes to `wt2909_checkpoint_integrity`? I've made some changes so we shorten the run-time in most cases, but extend the run-time as necessary to ensure the test passes.

lukech:
@keithbostic @agorrod I just pushed a commit to remove the unnecessary TESTUTIL_ENABLE_LONG_TESTS settings for tests/tasks in `test/evergreen.yml` that this PR is touching. Let me know if any comment.

lukech:
I made a Evergreen patch [build](https://evergreen.mongodb.com/build/wiredtiger_ubuntu1404_patch_1a6c5d42ab08da3108ace11e3fcd05164e023206_5c05baa8c9ec440d1badee66_18_12_03_23_22_17) to verify the task turnaround time with changes in this PR, the result is very positive - the longest running `csuite-wt2909-checkpoint-integrity-test` took only 5m 1s to finish.

lukech:
I switched `wiredtiger-test-unit-zseries` job to this feature branch to give us more confidence. http://build.wiredtiger.com:8080/job/wiredtiger-test-unit-zseries/2707/
With the test passed just now, I'm going ahead to merge the changes.

PR #4395, opened 2018-12-03
WT-4433 Revert Enable dynamic compression for snappy

bvpvamsikrishna:
To check the perf regression reported for ycsb small docs, reverting the change made to enable dynamic compression for snappy WT-4433.


bvpvamsikrishna:
This PR is to dig the commit which caused perf regression. Not to merge before conclusion.

agorrod:
Going to back out this change for now - re-using the original ticket.

PR #4398, opened 2018-12-04
WT-4444 Rework cursor statistics

agorrod:
Make the set of statistics consistent between connection and dhandles.
Reword so they sort together in groups.
Extend to add more information about how much data is being touched by
each operation type.


dist/stat_data.py
```diff
@@ -702,22 +709,25 @@ def __init__(self, name, desc, flags=''):
     CursorStat('cursor_cache', 'close calls that result in cache'),
     CursorStat('cursor_create', 'create calls'),
     CursorStat('cursor_insert', 'insert calls'),
-    CursorStat('cursor_insert_bulk', 'bulk-loaded cursor-insert calls'),
-    CursorStat('cursor_insert_bytes', 'cursor-insert key and value bytes inserted', 'size'),
-    CursorStat('cursor_modify', 'modify calls'),
+    CursorStat('cursor_insert_bulk', 'bulk loaded cursor insert calls'),
+    CursorStat('cursor_insert_bytes', 'insert key and value bytes', 'size'),
```
agorrod: I believe it's safe to change the words in these statistics - they are dhandle stats that aren't widely used at the moment.

agorrod:
@ddanderson please review this change - it's mostly extending the set of statistics we track on aggregate for cursor operations.

src/btree/bt_cursor.c
```diff
@@ -1490,10 +1490,11 @@ __wt_btcur_modify(WT_CURSOR_BTREE *cbt, WT_MODIFY *entries, int nentries)
...
 	WT_ERR(__cursor_size_chk(session, &cursor->value));
-	if (new > orig)
-		WT_STAT_DATA_INCRV(session, cursor_update_bytes, new - orig);
-	else
-		WT_STAT_DATA_DECRV(session, cursor_update_bytes, orig - new);
+
+	WT_STAT_CONN_INCRV(session, cursor_update_bytes_changed,
+	    new > orig ? new - orig : orig - new);
+	WT_STAT_DATA_INCRV(session, cursor_update_bytes_changed,
+	    new > orig ? new - orig : orig - new);
```
ddanderson: This is a funny statistic.  It now measures sum of the absolute value of all size changes.  It used to measure the actual sum of size changes.  That is, if the original size was 10 then it was 12, then 7, the net change of this statistic used to be 2 - 5 => -3.  Now it is 2 + 5 = 7.  It seems like an improvement of meaning.  If this is the intention, maybe should be called out in the ticket, as the statistics will appear to change, if anyone notices.

agorrod: ack - it was intentional. Since this was a data handle only statistic I think it's safe to change the meaning. I've called it out on the ticket as well.

agorrod:
Thanks for the review @ddanderson - I'm going to go ahead and merge.

PR #4399, opened 2018-12-04
WT-4438 Use accurate statistics for cursor cache totals in connection.

ddanderson:


ddanderson:

agorrod:
@ddanderson did you consider using the `WT_STAT_*_ATOMIC` macros instead of a shared variable in the connection? Those macros have the benefit of using a set of values - so should lead to less contention.

ddanderson:
@agorrod, thanks, that's a great idea.  I've made that change.

agorrod:
Thanks @ddanderson lgtm

PR #4401, opened 2018-12-06
WT-4280 Add debugging to know which session has a hazard pointer.

sueloverso:
[NOT FOR MERGE] @keithbostic is debugging like this helpful in the case of the assertion failure here? My quick look at the tarball it seemed like this would help to know which session still has a hazard pointer unexpectedly.


agorrod:
ping @keithbostic

sueloverso:
Actually @keithbostic messaged me that just getting the session itself would be useful because that is the only thing one would do with the index anyway. I need to push that version instead.

lukech:
rebuild this please

lukech:
retest this please

src/support/hazard.c
```diff
@@ -408,18 +411,21 @@ bool
...
 		if (!waitfor || ++i > 100)
 			break;
 		__wt_sleep(0, 10000);
 	}
 	__wt_errx(session,
-	    "hazard pointer reference to discarded object: (%p: %s, line %d)",
-	    (void *)hp->ref, hp->func, hp->line);
+	    "hazard pointer reference to discarded object: "
+	    "(%p: session %p name %s: %s, line %d)",
+	    (void *)hp->ref, (void *)s, s->name, hp->func, hp->line);
```
michaelcahill: I think we need to be careful about `s->name == NULL` here -- that is a pointer to string constants, so if we get any non-NULL value, I think it's fine to use it.  Some `snprintf` implementations will segault on NULL, though.

PR #4403, opened 2018-12-07
WT-4483 Improve caching of small updates to large values.

michaelcahill:


michaelcahill:
autobuild: retest this please

src/btree/bt_cursor.c
```diff
@@ -1425,10 +1425,11 @@ __cursor_chain_exceeded(WT_CURSOR_BTREE *cbt)
 		upd_size += WT_UPDATE_MEMSIZE(upd);
 		if (upd_size >= WT_MODIFY_MEM_FACTOR * cursor->value.size)
 			return (true);
-		if (__wt_txn_upd_visible_all(session, upd) &&
-		    i >= WT_MAX_MODIFY_UPDATE)
-			return (true);
 	}
+	if (upd != NULL && upd->type == WT_UPDATE_STANDARD &&
```
agorrod: So that I remember - one of the key effects of this change is that we won't shorten modify based update chains unless they contain at least one full entry. i.e: if there is a full entry on the disk-image, the `WT_UPDATE_STANDARD` test won't ever succeed.

agorrod:
Automated performance testing looks positive for this change - I'm going to go ahead and merge.

PR #4405, opened 2018-12-09
WT-4426 Change WT data format to optionally include timestamps in cells

keithbostic:


keithbostic:
@agorrod, @michaelcahill, @sulabhM, I think I've made all of the changes we discussed, so this branch should be ready for review/merge.

agorrod:
autobuild: retest this please

lukech:
retest this please

keithbostic:
@michaelcahill, I just made a change in e65bc73 that, I believe skips cells that weren't stable during shutdown, when rolling back to a version that doesn't support timestamp page formats. The tricky bit is the `WT_CELL_FOREACH_BEGIN` macro at the end of `src/include/cell.i` and needs a close look.

src/include/cell.i
```diff
@@ -209,72 +226,86 @@ __wt_cell_pack_data(WT_CELL *cell, uint64_t rle, size_t size)
...
-		if (rle)
-			WT_RET(__wt_vunpack_uint(&b, 0, &bv));
-		WT_RET(__wt_vunpack_uint(&b, 0, &bv));	/* Length */
+		if (__wt_process.page_version_ts) {
+			WT_RET(__wt_vunpack_uint(&b, 0, &v));	/* Skip TS */
+			WT_RET(__wt_vunpack_uint(&b, 0, &v));
+		}
+		if (rle)					/* Skip RLE */
+			WT_RET(__wt_vunpack_uint(&b, 0, &v));
+		WT_RET(__wt_vunpack_uint(&b, 0, &blen));	/* Length */
```
agorrod: The repetition in this code feels a bit unwieldy, but it's performance sensitive so possibly not worth refactoring.

src/include/cell.i
```diff
@@ -143,6 +149,8 @@ struct __wt_cell_unpack {
 
 	uint64_t v;			/* RLE count or recno */
 
+	wt_timestamp_t start, stop;	/* start/stop timestamps */
```
agorrod: I'd prefer if the names had `timestamp` in them. When @michaelcahill first added the timestamp changes I pushed to use the full word rather than `ts`, but we've started moving towards `ts` and `TS` recently. Either would be fine with me - but I think we should move towards a consistent usage pattern.

src/include/txn.h
```diff
@@ -41,7 +41,10 @@
 #define	WT_SESSION_IS_CHECKPOINT(s)					\
 	((s)->id != 0 && (s)->id == S2C(s)->txn_global.checkpoint_id)
 
-#define	WT_TS_NONE	0		/* No timestamp */
+#define	WT_TS_NONE	0		/* No (or earliest) timestamp) */
+#define	WT_TS_FIXME	37		/* Fake timestamp */
```
agorrod: Would you be OK with merging with this in here? It seems OK to me, but is a little odd so I thought I'd seek confirmation.

src/btree/bt_vrfy_dsk.c
```diff
@@ -38,6 +38,12 @@ static int __verify_dsk_row(
 	return (WT_ERROR);						\
 } while (0)
 
+#define	WT_CELL_FOREACH_VRFY(btree, dsk, cell, unpack, i)		\
```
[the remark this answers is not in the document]
keithbostic: Happy to do that, but not sure what'd you'd like to have called out?

src/include/txn.h
keithbostic: I agree it's odd.

We don't yet use any of these timestamps, and this is literally a flag meaning "this timestamp isn't right and needs to be fixed before it's used for real", maybe there's a better name?

src/include/cell.i
keithbostic: Will do. I'll go with `ts` so they aren't so long (start and stop aren't too bad, but `oldest_start` is getting longer).

src/include/cell.i
```diff
@@ -209,72 +226,86 @@ __wt_cell_pack_data(WT_CELL *cell, uint64_t rle, size_t size)
...
+		if (rle < 2) {
+			size -= WT_CELL_SIZE_ADJUST;
+			cell->__chunk[0] = WT_CELL_VALUE;	/* Type */
+		} else {
+			cell->__chunk[0] = WT_CELL_VALUE | WT_CELL_64V;
+			(void)__wt_vpack_uint(&p, 0, rle);	/* RLE */
+		}
+		(void)__wt_vpack_uint(&p, 0, (uint64_t)size);	/* Length */
 	}
-	(void)__wt_vpack_uint(&p, 0, (uint64_t)size);	/* Length */
```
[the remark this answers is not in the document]
keithbostic: I think your reading is incorrect, but I'm not completely confident. :)

Please take another look and let me know if you still disagree.

src/include/cell.i
keithbostic: I'm open to refactoring this, can you share further the path you're seeing?

src/include/cell.i
keithbostic: Pushed in 48f1f3a.

keithbostic:
autobuild: retest this please

src/btree/bt_vrfy_dsk.c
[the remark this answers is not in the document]
agorrod: It's specifically the assignment to `(cell)` that I find challenging - that statement is complex enough I think a comment would help, perhaps: "Iterate through each cell on a page, starting at the last cell and working backward. The size of each cell is used to decode the block of memory into data structures."
(I don't love the comment, but I've been sitting on responding for nearly a day, so I'm just going to post..)

src/include/cell.i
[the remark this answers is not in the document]
agorrod: I see now. There used to be an early return from the special short case. Sorry.

src/include/cell.i
agorrod: The code from lines 268->286 and 288->306 are the same apart from operating on `a` or `b` - that code could be pulled into a shared function.

It'd also be possible to make an inline function called `__skip_encoded_timestamps` (or similar) that does:
```
 		if (__wt_process.page_version_ts) {
 			WT_RET(__wt_vunpack_uint(&b, 0, &v));	/* Skip TS */
 			WT_RET(__wt_vunpack_uint(&b, 0, &v));
 		}
```

src/btree/bt_vrfy_dsk.c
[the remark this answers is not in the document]
keithbostic: Pushed in 56f90b1.

src/include/cell.i
keithbostic: I've futzed with this a little, and I don't see anything that I like better than the current code. Unless you feel strongly, I'm planning to leave this alone.

src/btree/bt_debug.c
```diff
@@ -1313,6 +1310,15 @@ __debug_cell(WT_DBG *ds, const WT_PAGE_HEADER *dsk, WT_CELL_UNPACK *unpack)
...
 
+	if (unpack->start_ts != WT_TS_NONE) {
+		__wt_timestamp_to_hex_string(hex_timestamp, unpack->start_ts);
+		WT_RET(ds->f(ds, ", ts %s-", hex_timestamp));
+	}
+	if (unpack->stop_ts != WT_TS_NONE) {
+		__wt_timestamp_to_hex_string(hex_timestamp, unpack->stop_ts);
+		WT_RET(ds->f(ds, "%s", hex_timestamp));
+	}
+
```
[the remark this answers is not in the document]
keithbostic: Thanks, @sulabhM. Pushed in 096a87d.

PR #4407, opened 2018-12-10
WT-4482 lint

keithbostic:


src/os_common/os_fhandle.c
```diff
@@ -146,6 +146,7 @@ __open_verbose(WT_SESSION_IMPL *session,
 	 * It's useful to track file opens when debugging platforms, take some
 	 * effort to output good tracking information.
 	 */
+	file_type_tag = "unknown open type";
```
nehakhatri5: Is this needed: /* -Werror=maybe-uninitialized */?

PR #4408, opened 2018-12-11
WT-4447 Add subsystem I/O throttling ability.

sueloverso:
@ddanderson Here's a branch for us to share.


lukech:
retest this please

ddanderson:
retest this please

ddanderson:
@sueloverso, added an (untested) throttling function.

sueloverso:
autobuild: retest this please

agorrod:
autobuild: retest this please

src/os_common/os_fhandle.c
```diff
@@ -355,6 +355,101 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+			 */
+			/*
+			 * Even if we get an error, we still need to lock and
+			 * decrement our reference counts.
+			 */
+			if (ret == 0) {
+				F_CLR(fh, WT_FH_DIRTY);
+				WT_STAT_CONN_INCR(session, fsync_all_fh);
+				fh->last_sync = now;
+				WT_PUBLISH(fh->written, 0);
```
ddanderson: Shouldn't we use `__wt_atomic_store64` instead of `WT_PUBLISH`?  Not sure about this, but `WT_PUBLISH` does a barrier operation before setting the value, so seems useful for forcing the order of operations, but does it guarantee the new value gets pushed out at this moment?

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,504 @@
...
+ * fewer sleeps with the risk of more choppy behavior as this number
+ * is larger.
+ */
+#define	WT_CAPACITY_SLEEP_CUTOFF_US	100
+
+/*
+ * When given a total capacity, divide it up for each subsystem. We allow
+ * and expect the sum of the subsystems to exceed 100.
+ * We aim for:
+ *    checkpoint: 10% of total
```
ddanderson: Checkpoint comment needs update (10% -> 5%), or just refer to defines below.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,504 @@
...
+    uint64_t *result)
+{
+	uint64_t res_len, res_value;
+
+	if (capacity != 0) {
+		res_len = WT_RESERVATION_NS(bytes, capacity);
+		res_value = __wt_atomic_add64(reservation, res_len);
+		if (now_ns > res_value && now_ns - res_value > WT_BILLION)
+			/*
+			 * If the total reservation clock is out of date,
```
ddanderson: Remove `total` from this comment.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,504 @@
...
+	    F_ISSET(conn, WT_CONN_RECOVERING))
+		return;
+
+	/*
+	 * There may in fact be some reads done under the umbrella of log
+	 * I/O, but they are mostly done under recovery. And if we are
+	 * recovering, we don't reach this code.
+	 */
+	if (type != WT_THROTTLE_READ) {
+		/* Sizes larger than this may overflow */
```
ddanderson: This comment belongs next to the assert for `bytes < 16 * WT_GIGABYTE`.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,504 @@
...
+		(void)__wt_atomic_addv64(&conn->capacity_written, bytes);
+		WT_STAT_CONN_INCRV(session, capacity_bytes_written, bytes);
+		__wt_capacity_signal(session);
+	} else
+		WT_STAT_CONN_INCRV(session, capacity_bytes_read, bytes);
+	WT_ASSERT(session, bytes < 16 * (uint64_t)WT_GIGABYTE);
+	WT_ASSERT(session, capacity != 0);
+
+	/* Get he current time in nanoseconds since the epoch. */
```
ddanderson: Comment typo.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,504 @@
...
+			steal = &conn->reservation_read;
+			steal_capacity = conn->capacity_read;
+			best_res = this_res;
+		}
+
+		/*
+		 * We have a subsystem that has enough spare capacity to
+		 * steal.  We'll take a small slice and add it to our
+		 * own subsystem.
+		 */
```
ddanderson: This comment should move after the opening to the `if` statement to make sense.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,504 @@
...
+				(void)__wt_atomic_sub64(&
+				    conn->reservation_total,
+				    WT_RESERVATION_NS(bytes, total_capacity));
+				goto again;
+			}
+
+			/*
+			 * We've actually stolen capacity in terms of bytes,
+			 * not nanoseconds, so we need to convert it.
+			 */
```
ddanderson: Remove blank line.

src/include/connection.h
```diff
@@ -104,6 +104,16 @@ struct __wt_named_extractor {
...
 };
 
+typedef enum {
+	WT_THROTTLE_CKPT,		/* Checkpoint throttle */
+	WT_THROTTLE_EVICT,		/* Eviction throttle */
+	WT_THROTTLE_LOG,		/* Logging throttle */
+	WT_THROTTLE_READ		/* Read throttle */
+} WT_THROTTLE_TYPE;
+
+#define	WT_THROTTLE_MIN		WT_MEGABYTE
```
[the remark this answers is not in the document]
sueloverso: It is used in `WT_CAPACITY_CHK` to check that if non-zero, the capacity is at least 1M.

src/include/connection.h
```diff
@@ -293,6 +303,34 @@ struct __wt_connection_impl {
...
+	WT_CONDVAR	*capacity_cond;	/* Capacity wait mutex */
+	uint64_t capacity_threshold;	/* Capacity size period */
+	volatile uint64_t capacity_written;	/* Written this period */
+	volatile bool	 capacity_signalled;	/* Capacity signalled */
+
+	uint64_t  capacity_usecs;	/* Capacity timer */
+
+	/*
+	 * A reservation is a point in the time (usually in the future) when
+	 * a write for a subsystem can be scheduled, so as not to overrun the
```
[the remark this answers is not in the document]
sueloverso: I like that.

src/os_common/os_fhandle.c
```diff
@@ -355,6 +355,101 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+	TAILQ_FOREACH_SAFE(fh, &conn->fhqh, q, fhtmp) {
+		handle = fh->handle;
+		WT_STAT_CONN_INCR(session, fsync_all_fh_total);
+		if (!F_ISSET(fh, WT_FH_DIRTY) ||
+		    handle->fh_sync_nowait == NULL ||
+		    fh->written < WT_CAPACITY_FILE_THRESHOLD) {
+			WT_STAT_CONN_INCR(session, fsync_too_small);
+			continue;
+		}
+		/* Skip over WiredTiger owned files. */
```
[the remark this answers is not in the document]
sueloverso: Oh, I like that. I think that string operations in the loop are likely a performance drag. This is an internal thread, so not a performance path, but I still like that idea.

src/os_common/os_fhandle.c
sueloverso: I think the PUBLISH does imply pushing it out. And near this line of code but unrelated, I think we can get rid of the `WT_FH_DIRTY` flag. The `fh->written` value being non-zero implies the same thing.

src/conn/conn_capacity.c
sueloverso: Since I just changed it (incorrectly) today, I agree with just referring to the values below.

src/include/os.h
```diff
@@ -109,11 +109,17 @@ struct __wt_fh {
 	const char *name;			/* File name */
 
 	uint64_t name_hash;			/* hash of name */
+	uint64_t last_sync;			/* time of background fsync */
+	volatile uint64_t written;		/* written since fsync */
```
sueloverso: This field is now, essentially, a duplicate of the `block->os_cache_dirty` field that is used with `os_cache_dirty_max`. It would be good to remove that redundancy. This is a thread that is coming in through the file handles, so it isn't using the block manager.

sueloverso:
@agorrod and @keithbostic this is ready for review. @keithbostic I thought you might be interested in looking at the file handle layer changes.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+__capacity_signal(WT_SESSION_IMPL *session)
+{
+	WT_CAPACITY *cap;
+	WT_CONNECTION_IMPL *conn;
+
+	conn = S2C(session);
+	cap = conn->capacity;
+	if (cap->written >= cap->threshold && !cap->signalled) {
+		__wt_cond_signal(session, conn->capacity_cond);
+		WT_PUBLISH(cap->signalled, true);
```
keithbostic: Not sure what writes this `WT_PUBLISH` is intended to flush?

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+ *	Make a reservation for the given number of bytes against
+ * the capacity of the subsystem.
+ */
+static void
+__capacity_reserve(WT_SESSION_IMPL *session, uint64_t *reservation,
+    uint64_t bytes, uint64_t capacity, uint64_t now_ns, uint64_t *result)
+{
+	uint64_t res_len, res_value;
+
+	WT_UNUSED(session);
```
keithbostic: Since `__capacity_reserve` is private to this file, why pass in a `WT_SESSION_IMPL` at all?

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+				(void)__wt_atomic_sub64(&cap->reservation_total,
+				    WT_RESERVATION_NS(bytes, total_capacity));
+				goto again;
+			}
+
+			/*
+			 * We've actually stolen capacity in terms of bytes,
+			 * not nanoseconds, so we need to convert it.
+			 */
+			stolen_bytes = steal_capacity / 16;
```
keithbostic: I'm a little confused why dividing by 16 converts bytes to nanoseconds?

Maybe a `#define` plus comment?

src/docs/tune-capacity.dox
```diff
@@ -0,0 +1,38 @@
...
+
+@snippet ex_all.c Configure capacity
+
+When a total capacity is set, the database ensures that the volume of
+system reads and writes totalled will not exceed the given I/O capacity.
+If a read or write is scheduled and would overflow the capacity, the issuing
+thread will sleep to guarantee the capacity ceiling.  The policy used is
+fair to all threads, and gives some weight to both readers and writers to
+try to ensure that each session can make progress when bandwidth
+resources are limited.
```
keithbostic: Not sure what `the database` means in this context. I think you can just delete `the database ensures that` and not lose meaning. I'd suggest `When a total capacity is set, the total volume of system reads and writes will not exceed the given I/O capacity.`

src/include/capacity.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+typedef enum {
+	WT_THROTTLE_CKPT,		/* Checkpoint throttle */
+	WT_THROTTLE_EVICT,		/* Eviction throttle */
+	WT_THROTTLE_LOG,		/* Logging throttle */
+	WT_THROTTLE_READ		/* Read throttle */
+} WT_THROTTLE_TYPE;
+
+#define	WT_THROTTLE_MIN			WT_MEGABYTE /* Config minimum size */
```
keithbostic: Unexpected indentation, I'd suggest one <tab> before `WT_MEGABYTE` and one <tab> before the comment.

src/include/capacity.h
```diff
@@ -0,0 +1,74 @@
...
+	volatile bool	 signalled;	/* Capacity signalled */
+
+	/*
+	 * A reservation is a point in the time when a write for a subsystem
+	 * can be scheduled, so as not to overrun the given capacity.  These
+	 * values hold the next available reservation, in nanoseconds since
+	 * the epoch. Getting a reservation with a future time implies sleeping
+	 * until that time; getting a reservation with a past time implies that
+	 * the operation can be done immediately.
+	 */
```
keithbostic: `point in the time` -> `point in time`?

src/os_common/os_fhandle.c
```diff
@@ -240,6 +241,20 @@ __wt_open(WT_SESSION_IMPL *session,
...
+	sep = __wt_path_separator();
+	WT_ASSERT(session, strlen(sep) == 1);
+	fp = strrchr(name, sep[0]);
+	if (fp == NULL)
+		fp = name;
+	else
+		++fp;
+	if (WT_PREFIX_MATCH(fp, "WiredTiger"))
+		F_SET(fh, WT_FH_WIREDTIGER_OWNED);
+
```
keithbostic: This code surprised me a bit, I was expecting a check for `WT_FS_OPEN_FILE_TYPE_DATA | WT_FS_OPEN_FILE_TYPE_LOG`. How is this different from that test?

keithbostic: Why not use `strstr(3)` instead, and skip the magic?

src/os_common/os_fhandle.c
```diff
@@ -355,6 +370,117 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
 	return (__handle_close(session, fh, true));
 }
 
+/*
+ * __wt_fsync_background_chk --
+ *	Return if background fsync is supported.
+ */
+int
+__wt_fsync_background_chk(WT_SESSION_IMPL *session)
```
keithbostic: `__wt_fsync_background_chk()` could more simply return a boolean.

src/os_common/os_fhandle.c
```diff
@@ -355,6 +370,117 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+{
+	WT_CONNECTION_IMPL *conn;
+	WT_DECL_RET;
+	WT_FH *fh, *fhnext;
+	WT_FILE_HANDLE *handle;
+
+	conn = S2C(session);
+	WT_ASSERT(session, !F_ISSET(conn, WT_CONN_READONLY));
```
keithbostic: With respect to an earlier comment, this might need `WT_CONN_IN_MEMORY`, too. (Or, this is safe, you could just remove the assert as far as I'm concerned.)

src/os_common/os_fhandle.c
```diff
@@ -355,6 +370,117 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+			    F_ISSET(fhnext, WT_FH_WIREDTIGER_OWNED))
+				fhnext = TAILQ_NEXT(fhnext, q);
+			/*
+			 * Increment our ref count on the current and next
+			 * handle. That way both are guaranteed valid when we
+			 * lock again after the fsync.
+			 */
+			++fh->ref;
+			if (fhnext != NULL)
+				++fhnext->ref;
```
keithbostic: What if there's one other reference to the handle, we increment the reference count, then the other reference is closed (and our reference prevents the handle from being really closed), and then we decrement our reference count. Won't that leave a handle with a reference count of 0?

src/os_common/os_fhandle.c
```diff
@@ -355,6 +370,117 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+__wt_fsync_background(WT_SESSION_IMPL *session)
+{
+	WT_CONNECTION_IMPL *conn;
+	WT_DECL_RET;
+	WT_FH *fh, *fhnext;
+	WT_FILE_HANDLE *handle;
+	uint64_t now;
+
+	conn = S2C(session);
+	WT_ASSERT(session, !F_ISSET(conn, WT_CONN_READONLY));
```
keithbostic: With respect to an earlier comment, this might need `WT_CONN_IN_MEMORY`, too. (Or, this is safe, you could just remove the assert as far as I'm concerned.)

src/os_common/os_fhandle.c
```diff
@@ -355,6 +370,117 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+			 * time would be done here.
+			 */
+			/*
+			 * Even if we get an error, we still need to lock and
+			 * decrement our reference counts.
+			 */
+			if (ret == 0) {
+				WT_STAT_CONN_INCR(session, fsync_all_fh);
+				fh->last_sync = now;
+				WT_PUBLISH(fh->written, 0);
```
keithbostic: Did you consider having two values in the `WT_FH` structure, one for the written bytes and one for the point at which fsync wrote the bytes? That should mean less collisions, given that the background sync thread is the only thread writing one of those values.

dist/api_data.py
```diff
@@ -519,6 +519,17 @@ def __cmp__(self, other):
...
+    Config('io_capacity', '', r'''
+        control how many bytes per second are written and read. Exceeding
+        the capacity results in throttling.''',
+        type='category', subconfig=[
+        Config('total', '0', r'''
+            number of bytes per second available to all subsystems in total.
+            When set, decisions about what subsystems are throttled, and in
+            what proportion, are made internally. The minimum non-zero setting
+            is 1MB.''',
+            min='0', max='1TB'),
```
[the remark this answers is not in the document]
sueloverso: @agorrod is correct. For this particular one, just numbers felt more natural than adding an `enabled` setting.

dist/stat_data.py
```diff
@@ -148,10 +148,24 @@ def __init__(self, name, desc, flags=''):
     ##########################################
     # System statistics
     ##########################################
+    ConnStat('capacity_bytes_ckpt', 'throttled capacity bytes checkpoint'),
+    ConnStat('capacity_bytes_evict', 'throttled capacity bytes eviction'),
+    ConnStat('capacity_bytes_log', 'throttled capacity bytes log'),
+    ConnStat('capacity_bytes_read', 'throttled capacity bytes read'),
+    ConnStat('capacity_bytes_written', 'throttled capacity bytes written'),
```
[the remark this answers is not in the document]
sueloverso: Done

dist/stat_data.py
```diff
@@ -148,10 +148,24 @@ def __init__(self, name, desc, flags=''):
...
+    ConnStat('capacity_threshold', 'throttled capacity threshold to fsync'),
```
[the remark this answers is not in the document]
sueloverso: It is the threshold where the capacity thread calls fsync. I modified the text.

dist/stat_data.py
```diff
@@ -148,10 +148,24 @@ def __init__(self, name, desc, flags=''):
...
+    ConnStat('capacity_time_ckpt', 'throttled capacity time in checkpoints (usecs)'),
+    ConnStat('capacity_time_evict', 'throttled capacity time in eviction (usecs)'),
+    ConnStat('capacity_time_log', 'throttled capacity time in logging (usecs)'),
+    ConnStat('capacity_time_read', 'throttled capacity time in reads (usecs)'),
+    ConnStat('capacity_time_total', 'throttled time in total capacity (usecs)'),
```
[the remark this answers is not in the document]
sueloverso: Fixed.

src/block/block_read.c
```diff
@@ -200,6 +200,7 @@ __wt_block_read_off_blind(WT_SESSION_IMPL *session,
 	 * the first allocation-size block.
 	 */
 	WT_RET(__wt_scr_alloc(session, block->allocsize, &tmp));
+	__wt_capacity_throttle(session, block->allocsize, WT_THROTTLE_READ);
```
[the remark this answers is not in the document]
sueloverso: Fixed.

src/block/block_read.c
```diff
@@ -248,6 +249,7 @@ __wt_block_read_off(WT_SESSION_IMPL *session, WT_BLOCK *block,
 		bufsize = WT_MAX(size, buf->memsize + 10);
 	}
 	WT_RET(__wt_buf_init(session, buf, bufsize));
+	__wt_capacity_throttle(session, size, WT_THROTTLE_READ);
```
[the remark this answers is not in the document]
keithbostic: > I'd be inclined to push the call to __wt_capacity_throttle into `__wt_read` and `__wt_write`- otherwise we aren't going to remember to track calls to read/write in the future.

This concerns me, because it implies that any locking in higher layers will potentially give us problems.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
+/*
+ * Copyright (c) 2014-2018 MongoDB, Inc.
+ * Copyright (c) 2008-2014 WiredTiger, Inc.
```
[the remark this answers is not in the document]
sueloverso: Fixed.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+		cap->evict = WT_CAPACITY(total, WT_CAP_EVICT);
+		cap->log = WT_CAPACITY(total, WT_CAP_LOG);
+		cap->read = WT_CAPACITY(total, WT_CAP_READ);
+
+		/*
+		 * Set the threshold to the percent of our capacity to
+		 * periodically asynchronously flush what we've written.
+		 */
+		cap->threshold = (cap->ckpt + cap->evict + cap->log) /
+		    100 * WT_CAPACITY_PCT;
```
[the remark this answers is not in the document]
sueloverso: Fixed.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+ *	Start the capacity server thread.
+ */
+static int
+__capacity_server_start(WT_CONNECTION_IMPL *conn)
+{
+	WT_SESSION_IMPL *session;
+
+	/* Nothing to do if the server is already running. */
+	if (conn->capacity_session != NULL)
+		return (0);
```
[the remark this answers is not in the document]
sueloverso: You're correct. The destroy function takes care of that on a reconfigure. Fixed.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+
+	conn = S2C(session);
+
+	/*
+	 * If it is a read only connection or if background fsync is not
+	 * supported, then there is nothing to do.
+	 */
+	if (F_ISSET(conn, WT_CONN_READONLY) ||
+	    (__wt_fsync_background_chk(session) == WT_NOTFOUND))
+		return (0);
```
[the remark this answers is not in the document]
sueloverso: Fixed.

src/conn/conn_capacity.c
sueloverso: I think the `volatile` is sufficient for the purposes we want, which is to not cache the values.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+		 * the signal was missed.
+		 */
+		__wt_cond_wait(session,
+		    conn->capacity_cond, WT_MILLION, __capacity_server_run_chk);
+
+		/* Check if we're quitting or being reconfigured. */
+		if (!__capacity_server_run_chk(session))
+			break;
+
+		WT_PUBLISH(cap->signalled, false);
```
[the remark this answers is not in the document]
sueloverso: Fixed

src/conn/conn_capacity.c
sueloverso: I will let @ddanderson make the call on this one. I agree with this sentiment. But it is there because we originally had debugging/verbose messages in there, and needed the session. Might we need that again sometime?

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+	if (capacity != 0) {
+		res_len = WT_RESERVATION_NS(bytes, capacity);
+		res_value = __wt_atomic_add64(reservation, res_len);
+		if (now_ns > res_value && now_ns - res_value > WT_BILLION)
+			/*
+			 * If the reservation clock is out of date, bring it
+			 * to within a second of a current time.
+			 */
+			__wt_atomic_store64(reservation,
+			    now_ns - WT_BILLION + res_len);
```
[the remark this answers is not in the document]
sueloverso: Fixed.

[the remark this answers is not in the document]
sueloverso: Fixed.

src/conn/conn_capacity.c
sueloverso: @ddanderson please add this comment. I forget why this works, so I think a comment will be helpful to remember in the future.

src/include/capacity.h
sueloverso: Fixed.

src/docs/tune-capacity.dox
sueloverso: Fixed.

src/include/capacity.h
```diff
@@ -0,0 +1,74 @@
...
+ * exceed 100, as often they are not at their maximum at the same time. In any
+ * event, we track the total capacity separately, so it is never exceeded.
+ */
+#define	WT_CAPACITY(total, pct)	((total) * (pct) / 100)
+#define	WT_CAP_CKPT		5
+#define	WT_CAP_EVICT		50
+#define	WT_CAP_LOG		30
+#define	WT_CAP_READ		55
+
+struct __wt_capacity {
```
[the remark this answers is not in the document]
sueloverso: Good catch! This was added very early, and the new structure was done very late and I didn't notice the overlap. Thanks! (and fixed.)

src/include/capacity.h
sueloverso: Fixed.

src/include/os.h
sueloverso: This issue is now in WT-4547.

src/os_common/os_fhandle.c
sueloverso: @ddanderson Please investigate these comments.

src/os_common/os_fhandle.c
sueloverso: Fixed.

src/os_common/os_fhandle.c
sueloverso: Removed. It is checked at a higher level.

src/os_common/os_fhandle.c
```diff
@@ -355,6 +370,117 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+	WT_FILE_HANDLE *handle;
+
+	conn = S2C(session);
+	WT_ASSERT(session, !F_ISSET(conn, WT_CONN_READONLY));
+	__wt_spin_lock(session, &conn->fh_lock);
+	/*
+	 * Look for the first non-WiredTiger file handle and see if
+	 * the fsync nowait function is supported.
+	 */
+	TAILQ_FOREACH_SAFE(fh, &conn->fhqh, q, fhnext) {
```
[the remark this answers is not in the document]
sueloverso: Good catch. That will clean up some of that code.

src/os_common/os_fhandle.c
sueloverso: Removed.

src/os_common/os_fhandle.c
sueloverso: Having it `volatile` should make sure other threads see the value.

src/include/capacity.h
```diff
@@ -0,0 +1,74 @@
...
+ * capacity limitation, but one that would even out. That is, possibly
+ * fewer sleeps with the risk of more choppy behavior as this number
+ * is larger.
+ */
+#define	WT_CAPACITY_SLEEP_CUTOFF_US	100
+
+/*
+ * When given a total capacity, divide it up for each subsystem. These defines
+ * represent the percentage of the total capacity that we allow for each
+ * subsystem's capacity. We allow and expect the sum of the subsystems to
```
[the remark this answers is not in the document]
sueloverso: How bizarre. It is okay with `subsystem` and `subsystems` but not the possessive? Fixed.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,461 @@
...
+	} else
+		res_value = now_ns;
+
+	*result = res_value;
+}
+
+/*
```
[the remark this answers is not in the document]
sueloverso: Although I understand where you're coming from we kept it one file because it really is fairly small. The entire file is just 450 lines. Breaking it up to the config/create/destroy/thread and the throttle pieces will result in very small files that you then have to think about where each piece is located. But the two pieces have no shared code either. How strongly do you feel?

src/block/block_read.c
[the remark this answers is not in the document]
sueloverso: I agree with Keith that throttling at the lowest level and holding a spinlock while we sleep may give us problems. Moving it to a higher level avoids that. Block manager and logging have been the only two paths out to the disk for many, many years. There exists the potential for another subsystem to be created that also does I/O sometime. But adding it down in `wt_read` or `wt_write` as insurance but impacting current, common code paths that might hold a lock seems like the worse tradeoff.

We could open a backlog ticket that did something like record the number of bytes coming through the throttle function (we already do that, but there are "nothing to do so return" paths that may skip it) and add in a comparison in `wt_read` or `wt_write` for total bytes written and triggers if those values are off by some amount as insurance.

src/block/block_write.c
```diff
@@ -335,6 +335,8 @@ __block_write_off(WT_SESSION_IMPL *session, WT_BLOCK *block,
 	WT_RET(ret);
 
 	/* Write the block. */
+	__wt_capacity_throttle(session, align_size,
+	     checkpoint_io ? WT_THROTTLE_CKPT : WT_THROTTLE_EVICT);
```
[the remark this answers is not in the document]
sueloverso: Hmm, as I looked into this, there is a tension here. There are a number of callers that circumvent `bm_write()` and call `wt_block_write_off` (a wrapper for `block_write_off` where the throttle call currently is) directly. They don't use `bm_write`. Moving it up there would mean those writes are not accounted for at all.

src/os_common/os_fhandle.c
ddanderson: We are looking for the final separator, and there is no `strrstr` (reverse string search).

keithbostic: I was thinking `strstr` for the separator until you run out of matches, and then a `strcmp` for the `WiredTiger` string.

I just really hate that `WT_ASSERT`.

How about what you have, but after the `strrchr`, do `WT_PREFIX_SKIP` of the separator string, followed by `WT_PREFIX_MATCH` of `WiredTiger`?

src/conn/conn_capacity.c
ddanderson: I've added clarifying comments and a macro.

src/conn/conn_capacity.c
ddanderson: Fixed per @keithbostic's suggestion.

src/os_common/os_fhandle.c
ddanderson: Thanks, @keithbostic.  Looks like checking `file_type != WT_FS_OPEN_FILE_TYPE_DATA` could be used to determine if a file is owned by WT.  The one counter example I've seen is the `WiredTigerLAS` file, that is marked as `WT_FS_OPEN_FILE_TYPE_DATA`.  So it would participate in file syncing, where it wouldn't with the current code.  That's probably not a bad thing?  I'm running a performance run to see if there's any noticeable affect for my test case.

keithbostic: `WiredTigerLAS` has the right prefix, so why doesn't i tmatch? I'm sure I'm missing something terribly obvious...

Regardless, I agree with you that the LAS file should participate in file syncing, there's a lot of traffic to that file in some workloads, and I suspect there will be more in the future.

sueloverso: I thought the point of your comment about the file types was to look at the type instead of doing the string comparisons. So doing that would remove the code about matching the prefix.

keithbostic: Yes: if we can use the file type flags instead, all of this goes away.

If we can't make that change, then this suggestion might get rid of the path separator assert.

ddanderson: @keithbostic, I've pushed a change to use the file type. `WiredTigerLAS` is opened as `WT_FS_OPEN_FILE_TYPE_DATA`, so it will participate in file sync, it didn't before.  I didn't see any change in performance - I guess that's not really surprising, as I don't think we we're doing anything that requires retaining history.

ddanderson: I think `WT_FH_WIREDTIGER_OWNED` is confusing. Should we change it to something like `WT_FH_BACKGROUND_SYNC` (and reverse when it's set)?

ddanderson: Done.

ddanderson:
@sueloverso, please check the last change (https://github.com/wiredtiger/wiredtiger/pull/4408/commits/661df287ba4fff0744354df49b315843f133263a).  Some code reordering was needed to resolve a test error on windows, where reconfigure didn't get an error when using a capacity that was too small.

src/os_common/os_fhandle.c
ddanderson: This discussion is now moot, we no longer look at file names.

ddanderson:
@keithbostic, can you take another pass please?  I think we've resolved all your comments. Please don't merge yet, I made a final modification that I'd like Sue to review.

src/conn/conn_capacity.c
```diff
@@ -0,0 +1,464 @@
...
+	conn = S2C(session);
+
+	WT_RET(__wt_config_gets(session, cfg, "io_capacity.total", &cval));
+	if (cval.val != 0 && cval.val < WT_THROTTLE_MIN)
+		WT_RET_MSG(session, EINVAL,
+		    "total I/O capacity value %" PRId64 " below minimum %d",
+		    cval.val, WT_THROTTLE_MIN);
+
+	if (cval.val != 0) {
+		WT_RET(__wt_calloc_one(session, &conn->capacity));
```
keithbostic: It occurs to me, this allocation is 104 bytes, it's not worth the error handling to deal with the possible allocation failure, just add it to the `WT_CONNECTION` structure.

src/conn/conn_capacity.c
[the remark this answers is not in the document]
keithbostic: FWIW, I'm mildly on the side of leaving it in a single file.

src/include/os_fhandle.i
```diff
@@ -196,6 +196,7 @@ __wt_write(WT_SESSION_IMPL *session,
 	time_stop = __wt_clock(session);
 	__wt_stat_msecs_hist_incr_fswrite(session,
 	    WT_CLOCKDIFF_MS(time_stop, time_start));
+	__wt_atomic_addv64(&fh->written, len);
```
keithbostic: `__wt_atomic_addv64()` returns a value.

src/os_common/os_fhandle.c
```diff
@@ -355,6 +357,101 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+			continue;
+		now = __wt_clock(session);
+		if (fh->last_sync == 0 ||
+		    WT_CLOCKDIFF_SEC(now, fh->last_sync) > 0) {
+			/*
+			 * Increment our ref count on the current and next
+			 * handle. That way both are guaranteed valid when we
+			 * lock again after the fsync.
+			 */
+			++fh->ref;
```
keithbostic: I think there's still an open question on this code:

> What if there's one other reference to the handle, we increment the reference count, then the other reference is closed (and our reference prevents the handle from being really closed), and then we decrement our reference count. Won't that leave a handle with a reference count of 0?

ddanderson: Thanks, you're right.  I just pushed a fix.

src/os_common/os_fhandle.c
ddanderson: A fix has been pushed for this.

src/conn/conn_capacity.c
keithbostic: Oh, I think `WT_CAPACITY` is a good idea, I'm just saying don't bother allocating the structure, just declare it.

(On second thought, that means all of your `p->` code would become `p.`, which is maybe just annoying pain at this point. Oh, forget it. :)

sueloverso: I committed a change for this. It was actually very small.

sueloverso:
@keithbostic can you let us know if there's anything else in this PR you want addressed?

src/os_common/os_fhandle.c
```diff
@@ -355,6 +357,108 @@ __wt_close(WT_SESSION_IMPL *session, WT_FH **fhp)
...
+			__wt_spin_lock(session, &conn->fh_lock);
+			--fh->ref;
+			WT_ERR(ret);
+			/*
+			 * The last open reference to the file handle may
+			 * have been released while we dropped the lock.
+			 * If so, we're responsible for closing.
+			 */
+			if (fh->ref == 0)
+				return (__handle_close(session, fh, true));
```
keithbostic: If `__wt_fsync` fails, we don't panic, but we potentially leak a file handle. Any, why are we returning out of the loop here if we're closing a  handle, aren't there other handles to sync (plus, we're returning while holding the spin lock).

I'd suggest:
```
*** os_fhandle.c        Wed Feb  6 15:08:45 2019
--- -   Wed Feb  6 15:23:46 2019
***************
*** 434,458 ****
                         * is blocking, adding a clock call and checking the
                         * time would be done here.
                         */
-                       /*
-                        * Even if we get an error, we still need to lock and
-                        * decrement our reference counts.
-                        */
                        if (ret == 0) {
                                WT_STAT_CONN_INCR(session, fsync_all_fh);
                                fh->last_sync = now;
                                fh->written = 0;
                        }
!                       __wt_spin_lock(session, &conn->fh_lock);
!                       --fh->ref;
!                       WT_ERR(ret);
                        /*
!                        * The last open reference to the file handle may
!                        * have been released while we dropped the lock.
!                        * If so, we're responsible for closing.
                         */
                        if (fh->ref == 0)
!                               return (__handle_close(session, fh, true));
                }
        }
  err:  __wt_spin_unlock(session, &conn->fh_lock);
--- 434,460 ----
                         * is blocking, adding a clock call and checking the
                         * time would be done here.
                         */
                        if (ret == 0) {
                                WT_STAT_CONN_INCR(session, fsync_all_fh);
                                fh->last_sync = now;
                                fh->written = 0;
                        }
!
                        /*
!                        *
!                        * Even if we get an error, we still need to lock and
!                        * decrement our reference counts.
!                        *
!                        * The last open reference to the file handle may have
!                        * been released while we dropped the lock. If so, we're

!                        * responsible for closing.
                         */
+                       __wt_spin_lock(session, &conn->fh_lock);
+                       --fh->ref;
                        if (fh->ref == 0)
!                               WT_TRET(__handle_close(session, fh, true));
!
!                       WT_ERR(ret);
                }
        }
  err:  __wt_spin_unlock(session, &conn->fh_lock);
```

ddanderson: @keithbostic, I've just pushed some changes that are a little different than you suggest.  `__handle_close` when called with `true`, releases the `fh_lock`.  I assume we want it to do that, as the handle close may take some time.  So we need to reacquire the lock.  Also, I followed through on your thought that we shouldn't exit the loop on account of any single failure.

keithbostic: @ddanderson, I think this is close, but I think you'll now need `TAILQ_FOREACH_SAFE`, because we're using `fh` after it's been freed, aren't we?

ddanderson: @keithbostic, yes you're right.  If we have a temp variable, we'll need to increase its ref count also so it doesn't get deleted out from under us.  I think we did that in a past version of this code.

ddanderson: @keithbostic, I've pushed a fix for your consideration.

keithbostic: I think we can still get into trouble:

First, if `fhtmp` goes to 0, but that handle isn't a file handle that we will flush on the next iteration of the outer loop, we `continue` the outer loop instead of falling into the check-for-handle-close code, so it won't get closed.

Second, once we decrement the `fhtmp` reference count, can't the close of the `fh` handle (discarding the spinlock), allow `fhtmp` to be closed out from under us?

Also, I don't think it matters, but checking for `fh->ref > 0` looks wrong to me. We're holding the spinlock, is it possible for a handle to be on that list with a zero ref count?

I can't see any way to make the existing loop work without a bunch of goto's and special cases.

I've pushed a branch in 92e175a for your consideration.

I don't think it's any slower, and it seems cleaner and less fiddly to me -- we've had enough problems getting this correct I'd like to avoid anything complicated. Feel free to simply revert the change if you don't like it, or I'm misunderstanding the code, or you just have something you like better!

ddanderson:
@keithbostic, your latest change LGTM, thanks!  Is there anything else you'd like to see fixed?

src/os_common/os_fhandle.c
sueloverso: Thanks to both of you for working through this. LGTM.

sueloverso:
Is there anything else outstanding in this PR?

PR #4409, opened 2018-12-12
WT-4430 Fix race between prepare and page instantiate with fast truncate

sulabhM:


sulabhM:
@bvpvamsikrishna, I don't think comments go beyond column 80. s_all and hence pull req testing captures long lines. Our pull req testing shouldn't pass if there are long lines.
I checked again and do not see any long lines.

src/include/txn.i
```diff
@@ -343,51 +343,58 @@ __wt_txn_unmodify(WT_SESSION_IMPL *session)
...
 	 * Lock the ref to ensure we don't race with eviction freeing the
-	 * page deleted update list.
+	 * page deleted update list or with a page instantiate.
 	 */
 	for (;; __wt_yield()) {
 		previous_state = ref->state;
+		WT_ASSERT(session, previous_state != WT_REF_READING);
 		if (previous_state != WT_REF_LOCKED &&
+		    (previous_state == WT_REF_MEM ||
+		    previous_state == WT_REF_DELETED) &&
```
agorrod: I don't think this is OK. I think the ref could be `WT_REF_LOOKASIDE` or `WT_REF_LIMBO`. Even if it couldn't the consequence of a ref being in that state unexpectedly is that this code would spin forever.

src/include/txn.i
```diff
@@ -343,51 +343,58 @@ __wt_txn_unmodify(WT_SESSION_IMPL *session)
...
+		ts = txn->commit_timestamp;
+		prepare_state = WT_PREPARE_RESOLVED;
+	} else {
+		ts = txn->prepare_timestamp;
+		prepare_state = WT_PREPARE_INPROGRESS;
+	}
 	for (updp = ref->page_del->update_list;
 	    updp != NULL && *updp != NULL; ++updp) {
-		(*updp)->timestamp = txn->commit_timestamp;
-		if (F_ISSET(txn, WT_TXN_PREPARE))
```
agorrod: Was this just redundant before? Should we be asserting it now?

sulabhM: I think this was redundant before, this function was called in an if condition that already checked for WT_TXN_PREPARE state. Now this function can be called as part of prepare() call, before WT_TXN_PREPARE is set on txn, so we shouldn't be asserting. Now, this function will be called with and without WT_TXN_PREPARE set on txn.

src/include/txn.i
sulabhM: I guess this needs a discussion. This check is from the notes I took when we had a meeting with @michaelcahill

michaelcahill: We talked about various options here but I thought we ended up agreeing to add the assertion above (`previous_state != WT_REF_READING`) but not make these changes.

src/include/txn.i
```diff
@@ -343,51 +343,58 @@ __wt_txn_unmodify(WT_SESSION_IMPL *session)
...
-			/*
-			 * Holding the ref locked means we have exclusive
-			 * access, so don't need to use the prepare locked
-			 * transition state.
-			 */
-			(*updp)->prepare_state = WT_PREPARE_RESOLVED;
+		(*updp)->timestamp = ts;
+		/*
+		 * Holding the ref locked means we have exclusive access, so
+		 * if we are committing we don't need to use the prepare locked
```
bvpvamsikrishna: @sulabhM , my comment was not about above 80, it is not upto 80, like "if" can be part of the previous line itself.

sulabhM: Ahh ! got you, my mistake. I will fix that.

src/include/txn.i
sulabhM: Fixed.

src/include/txn.i
sulabhM: Fixed.

sulabhM:
@agorrod I have addressed the comments, can you please review again

PR #4410, opened 2018-12-12
WT-4482 lint

keithbostic:
Revert a commit in the #4407 branch, it's incorrect and can potentially lead to a column-store core dump.


keithbostic:
@lukech, I attached this to the WT-4482 ticket since I don't think that change has made it out into the wild. Would you please merge this change if that's correct? If that's incorrect, just let me know and I'll create a new ticket. -- Thanks!

agorrod:
autobuild: retest this please

lukech:
@keithbostic You are right the initial commit of WT-4482 has not been merged to other branches (or sync-ed to any mongo branch). I will go ahead to merge this partial revert change.

PR #4411, opened 2018-12-16
WT-4499 Fix prepare transactions key order check failure.

bvpvamsikrishna:
This PR has two changes.
1.A prepared delete operation after resolution will be visible, but should not use the key and value.

2.cbt->compare to be set to non-zero to indicate that key is not pre-built.


bvpvamsikrishna:

PR #4414, opened 2018-12-19
WT-4498 Add Evergreen test to verify data files portability

lukech:
In order to create a test infrastructure we can use to periodically verify wt data files portability across different platforms (big_endian & little_endian), below changes were made in this pull request:

- Created a new shell script, try listing and dumping wt data files in the WT_TEST directory
- Created two new Evergreen build variants ('big-endian' and 'little-endian'), and 3 tasks ('compile', generate data files, and dump data files from a different platform) for each build variant. Inter-variant dependency was setup across tasks of the 2 build variants so that data files generated by one platform can be verified by the other platform.

Currently the patch [build](https://evergreen.mongodb.com/version/5c196e952fbabe28277f85f8) is still failing with the `s3.get` artifacts problem, which needs to be addressed.


lukech:
Lots of wt data files were generated when running python suite with preserve (-p) option. It caused timeout while extracting the artifacts tarball on a different platform. Patch build: https://evergreen.mongodb.com/version/5c25b5ba2a60ed51248cc8c9
Will check with Evergreen team how to tune the timer (for s3.get).

lukech:
The 5 mins timer for `s3.get` (which currently includes tarball extraction) is not adjustable in Evergreen. Applied a workaround of separating `archive.targz_extract` from `s3.get` then it progressed further. 
https://evergreen.mongodb.com/version/5c2d43e42fbabe957fc70999

- The big-endian dump of little-endian data files failed on a table (WT-4515 filed).
- The artifacts archiving timed out (after 15 mins) and failed because of the large number of (data) files to be archived. Need to check with Evergreen team and apply a workaround.
- The little-endian dump of big-endian data files failed at the compile step (on little-endian platform) with mismatched libtool version issue. Considering a solution of only including `WT_TEST` directory in the archive in the dump tasks (currently the whole wiredtiger workspace is archived), hopefully also mitigate the long duration of artifacts archiving.

lukech:
In order to address the 3rd issue (bullet) mentioned above, I pushed a commit to archive only `WT_TEST` directory (instead of the whole wiredtiger workspace) which includes the list of wt data files to be fetched and verified on a different platform. Also added a dependency to corresponding 'compile' task in the 'dump' task Evergreen configuration, so a 'dump' task can fetch binaries/libraries compiled on the same platform, and use them to verify wt data files (through `WT_TEST` fetching) generated on a different platform.

lukech:
Here is the summary of the changes made in this PR by creating a test infrastructure to periodically verify wt data files portability across different platforms (big_endian & little_endian):

- Created a new shell script, try listing and dumping wt data files in the WT_TEST directory
- Created two new daily-run Evergreen build variants (`big-endian` and `little-endian`), and 3 tasks for each build variant (`compile`, `generate` data files, and `dump` data files from a different platform). Inter-variant dependency was setup across tasks of the 2 build variants so that data files generated by one platform can be dumped/verified by the other platform. Specifically, `generate` task depends on `compile` task of the same build variant, and `dump` task depends on both `compile` task of the same build variant and `generate` task of the other build variant. `generate` task runs Python test suite and preserves the wt files (using '-p' command option of the test/suite/run.py script). 

There were a couple of Evergreen configuration workarounds applied along the way e.g. separate tarball extraction out to avoid hitting the 5 mins 's3.get' timer in `dump` tasks; archive only '.libs/WT_TEST' directory to minimise the artifacts size and avoid file permission issues in `generate` tasks; compile wiredtiger locally instead of using the artifacts archived by a different platform to avoid libtool version mismatch issue.

The latest patch build with the 2 new build variants can be used as a reference to get a feel what have been achieved so far: https://evergreen.mongodb.com/version/5c31945ce3c3313ca39bbba6 

There are still a few known issues we'd need to address, either in this PR or separate tickets:
1. The big-endian dump of little-endian data files failed on a table in the big-endian `dump` task. WT-4515 was filed.
2. The artifacts archiving took over 15 mins and timed out for `dump` tasks. Need to find out a workaround (e.g. try shelling out to `tar` instead of `archive.targz_pack` buildin command).
3. The little-endian `dump` task failed to find the new shell script, caused by an overwritten artifacts - turned out both `compile` and `generate` of the same build variant (little-endian) attempted to write to the same location of s3 bucket, since `build_id` Evergreen expansion were parsed to the same value for the 2 tasks which feels abnormal. Need to check with Evergreen team to find a solution. 

@agorrod @keithbostic Though this might not be the full set of changes to be included in this PR, it might be worthwhile for you to start reviewing the changes I got so far. I will continue chasing up the 3 known issues listed above. 

Another aspect I'd hope to get your advice is about the wt data file testing scope - right now the full Python test suite is executed, which generates a large number of wt (and corresponding txt, lock, basecfg, turtle) files. It takes extra long time to run the testing and archive the artifacts. I'm wondering if it's safe to only verify a subset of those wt data files (and suite tests) to make the Evergreen tasks agile and hitting the point (instead of exhausting the whole test suite and wt data files they generate). Also I acknowledge picking specific tests needs extra human input and might introduce potential future work/overhead to evaluate if new Python tests need to be included in the data file portability tasks. Please let me know your opinions. Thanks.

lukech:
For issue 3 (artifacts overwritten), I tried to use expansion `${task_id}` to replace `${build_id}` as the artifacts name, and it worked well - making the artifacts name unique within the project. I think we should make a base change by using `${task_id}` for the project level `Post` commands so that each task has a unique artifacts file generated in s3, which can be used for debugging purpose.  For `compile` task specifically, using `${build_id}` as the artifacts name is more reasonable as it's used as a measure to share build artifacts between `compile` and testing tasks. In order to avoid `compile` task archiving 2 identical artifacts (with different names `${build_id}` and `${task_id}`), I'm planning to make a trick by removing `wiredtiger` directory as part of archiving `${task_id}` artifacts - so that the s3.put step would fail and save ~1 min time for `compile` task.

Will push the commit once it passes the patch build (which takes time as it involves big-endian tasks).

lukech:
With @nehakhatri5's recent commits, the little-endian dump wt datafiles Evergreen task progressed further, and failed on a later case `test_baseconfig`. WT-4559 was opened to track this new issue.

lukech:
retest this please

lukech:
retest this please

lukech:
Will close this PR as we decided to pursue a different approach to verify data file portability.

PR #4415, opened 2018-12-19
WT-4505 Add ability to manage when prepared transactions appear in checkpoints

bvpvamsikrishna:
Update the WT_UPDATE and WT_TXN structure to add a new member to store durable_timestamp.
As part of commit transaction, copy the given commit_timestamp to durable_timestamp if it is a prepared transaction and durable_timestamp is not specified.


agorrod:
Is this ready for review @bvpvamsikrishna ? If so please remove the locked label.

src/btree/bt_read.c
```diff
@@ -190,6 +190,9 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 		upd->txnid = las_txnid;
 		upd->timestamp = las_timestamp;
 		upd->prepare_state = prepare_state;
+		/* Copy the commit timestamp as durable timestamp. */
```
agorrod: I guess this is OK for now, since we aren't writing prepared updates to lookaside, it'll always be correct. I think it's worth expanding the comment to reflect that:

Use the commit timestamp as the durable timestamp, since prepared updates don't currently get written to lookaside, so the two timestamps should always be identical.

src/include/misc.h
```diff
@@ -366,3 +366,11 @@ union __wt_rand_state {
 		    session, buf, (buf)->size + __len + 1));		\
 	}								\
 } while (0)
+
+/*
+ * HAVE_DURABLE_TIMESTAMPS
+ * 	To enable functionality of oplog durability feature i.e. using
+ * durable timestamps.
+ */
+#define	HAVE_DURABLE_TIMESTAMPS
```
agorrod: Can this be removed?

src/include/txn.i
```diff
@@ -438,11 +443,19 @@ __wt_txn_op_set_timestamp(WT_SESSION_IMPL *session, WT_TXN_OP *op)
 		/*
 		 * The timestamp is in the page deleted structure for
 		 * truncates, or in the update for other operations.
+		 * Both commit timestamp and durable timestamp needs to be
+		 * updated.
```
agorrod: Both commit and durable timestamps need to be updated.

Is better phrasing.

src/include/txn.i
```diff
@@ -438,11 +443,19 @@ __wt_txn_op_set_timestamp(WT_SESSION_IMPL *session, WT_TXN_OP *op)
...
 		timestamp = op->type == WT_TXN_OP_REF_DELETE ?
 		    &op->u.ref->page_del->timestamp : &op->u.op_upd->timestamp;
-		if (*timestamp == WT_TS_NONE)
+		if (*timestamp == WT_TS_NONE) {
 			*timestamp = txn->commit_timestamp;
+
+			timestamp = op->type == WT_TXN_OP_REF_DELETE ?
+			    &op->u.ref->page_del->durable_timestamp :
+			    &op->u.op_upd->durable_timestamp;
+			*timestamp = txn->durable_timestamp;
```
agorrod: Can the original assignment to `*timestamp` be replaced with a comment here? The comment would explain that "The durable timestamp is usually the commit timestamp, except when dealing with prepared transactions."

src/include/txn.i
```diff
@@ -795,6 +808,21 @@ __wt_txn_upd_visible_type(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
+/*
+ * __wt_txn_upd_durable --
+ *	Can the current transaction make the given update durable.
+ */
+static inline bool
+__wt_txn_upd_durable(WT_SESSION_IMPL *session, WT_UPDATE *upd)
+{
+	/* If update is visible then check if it is durable. */
+	if (__wt_txn_upd_visible_type(session, upd) == WT_VISIBLE_TRUE)
+	{
```
agorrod: Extra braces around a 1-line if. Could it be:
```
if (__wt_txn_upd_visible_type(session, upd) != WT_VISIBLE_TRUE)
    return (false);
return (__wt_txn_visible(session, upd->txnid, upd->durable_timestamp));
```
?

src/reconcile/rec_write.c
```diff
@@ -1268,9 +1276,11 @@ __rec_txn_read(WT_SESSION_IMPL *session, WT_RECONCILE *r,
 		if (*updp == NULL && r->las_skew_newest)
 			*updp = upd;
 
-		if (F_ISSET(r, WT_REC_VISIBLE_ALL) ?
+		/* Consider non durable updates as unstable. */
```
agorrod: Could you replace `unstable` with `uncommitted` for consistency? i.e: switch to using the same comment as on line `1245`

src/txn/txn.c
```diff
@@ -699,14 +699,20 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	    txn->mod_count == 0);
 
 	readonly = txn->mod_count == 0;
+
```
agorrod: whitespace

src/txn/txn.c
```diff
@@ -716,8 +722,46 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		WT_ERR_MSG(session, EINVAL,
 		    "commit_timestamp is required for a prepared transaction");
 
+#ifdef HAVE_DURABLE_TIMESTAMPS
+	/*
+	 * Durable timestamp is required for a prepared transaction.
+	 */
+	if (prepare) {
+		/*
+		 * Look for a durable timestamp.
```
agorrod: This comment should be on a single line. Actually, I don't think it's necessary because of the one above the `if (prepare)`

src/txn/txn.c
```diff
@@ -716,8 +722,46 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		WT_ERR_MSG(session, EINVAL,
 		    "commit_timestamp is required for a prepared transaction");
 
+#ifdef HAVE_DURABLE_TIMESTAMPS
+	/*
+	 * Durable timestamp is required for a prepared transaction.
```
agorrod: Does this fit on a single line?

src/txn/txn.c
```diff
@@ -716,8 +722,46 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+#endif
+
+	if (cval.len != 0)
+		WT_ERR(__wt_timestamp_validate(
+		    session, "durable", txn->durable_timestamp, &cval, true));
+
 	WT_ERR(__txn_commit_timestamp_validate(session));
 
+#ifdef HAVE_DURABLE_TIMESTAMPS
+		// TODO : assert durable_timestamp
```
agorrod: Should switch to C-style comment

src/include/txn.i
agorrod: Ignore this one

PR #4419, opened 2018-12-24
WT-4510 Failure when verifying internal pages of older version data files

keithbostic:


keithbostic:
The review here is tricky: essentially, review means checking code that unpacks cells (callers of anything that ends up at `__wt_cell_unpack_safe()`), and confirm none of them subsequently use `unpack.cell` to reference the underlying copy of the cell.

In other words, if a cell can be of type `WT_CELL_VALUE_COPY` (which means it references a different "copy" cell that appears earlier in the block), then any callers using `unpack.cell` after unpacking a cell must be intending that reference to be to the original cell, not the copy cell.

PR #4426, opened 2019-01-10
WT-4520 prepare transactions fallout during cursor navigation.

bvpvamsikrishna:
When scanning a new page and prepared update is encountered, the cbt->slot is not being updated, it still represents the slot from the earlier page.

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -305,6 +305,7 @@ __cursor_row_next(WT_CURSOR_BTREE *cbt, bool newpage)
 		cbt->ins_head = WT_ROW_INSERT_SMALLEST(page);
 		cbt->ins = WT_SKIP_FIRST(cbt->ins_head);
 		cbt->row_iteration_slot = 1;
+		cbt->slot = 0;
 		cbt->rip_saved = NULL;
 		goto new_insert;
 	}
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -178,6 +178,7 @@ __cursor_var_next(WT_CURSOR_BTREE *cbt, bool newpage)
 
 	/* Initialize for each new page. */
 	if (newpage) {
+		cbt->slot = UINT32_MAX;
 		cbt->last_standard_recno = __col_var_last_recno(cbt->ref);
 		if (cbt->last_standard_recno == 0)
 			return (WT_NOTFOUND);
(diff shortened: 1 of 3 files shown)


bvpvamsikrishna:
@keithbostic 
With the change in this PR, I was not able to reproduce wt-4520.
I raised this PR to discuss with you.

My observations during debugging:
1. If I modify the test/format to call the prepare_transaction all the times, instead of just 10%, even with out truncate also I can reproduce this issue. So, this issue is not just with truncate.
2. During debugging I observed that when prepare_conflict is reported, the cbt->row_iteration_slot is updated but not the cbt->slot.  My understanding of the scenario leading to the issue is : 
      a) cursor->next has returned a valid value.
      b) cursor->next is called again.
      c) cursor scan moved to a new page.
      d) page insert list has a prepared update, so __cursor_row_next returns from bt_curnext.c:324 with prepare_conflict error.
      e) prepared transaction is rolledback. som prepared update is not visible. 
      e) test/format calls the cursor->next() again.
      f) In {{__cursor_check_prepared_update}} we call {{__wt_cursor_valid}}.
      g) Since prepared update is not visible, {{__wt_cursor_valid}} will enter the {{case BTREE_ROW:}} and hits the assert. 

I think this is what is happening. 
I don't understand this code completely, so I might got it completely wrong.

Tomorrow morning I will discuss with you.

bvpvamsikrishna:
@keithbostic, Thank you for the quick review.
This change is not complete. I want to get your opinion on whether I am in right direction or not before making changes in other code flow paths.

As of now, I don't have a static test.  I will try to write one.

bvpvamsikrishna:
@keithbostic, Could you review the changes in this PR.

I think cbt->slot to be initialied in {{__cursor_var_prev}} as well. If yes, could you help me with the change. I am not able to understand to what value `cbt->slot` to be initialized, it seems the page can have the repeats as well (i.e. I am not able to understand the usage of {{page->pg_var_nrepeats}}.

bvpvamsikrishna:
@keithbostic, I have remove `cbt.slot=0` from `__cursor_var_next` as `cbt->slot` is assigned before call to `__wt_txn_read`. Same is the case with `__cursor_var_prev` as well.

michaelcahill:
@bvpvamsikrishna, I've taken a look at these changes.  Do you have plans to add any testing to this branch, or are the changes ready for final review?

bvpvamsikrishna:
@michaelcahill I have just pushed some changes based on discussions to this PR. No testing will be added to this branch, I have verified with the failed test case only.

src/btree/bt_curnext.c
```diff
@@ -178,6 +178,7 @@ __cursor_var_next(WT_CURSOR_BTREE *cbt, bool newpage)
 
 	/* Initialize for each new page. */
 	if (newpage) {
+		cbt->slot = UINT32_MAX;
```
agorrod: I'd like a comment around each of these, along the lines of `/* Be paranoid and set the slot out of bounds when moving to a new page. */`

bvpvamsikrishna:
@keithbostic , could you take a look at this PR now. I believe all changes are complete.

keithbostic:
lgtm, I'll merge so it runs over the weekend.

PR #4427, opened 2019-01-10
WT-4524 Ignore slowness failures when running valgrind.

sueloverso:
@keithbostic please review. If valgrind is running then ignore any slow operations.


sueloverso:
autobuild: retest this please

PR #4431, opened 2019-01-14
WT-4521 Change WiredTiger to format printed timestamps as 4B integer pairs

keithbostic:
There are two interesting changes here:

* change af46378, switching most of our timestamp displays from hexadecimal to integers, making sure that I switched all of the ones I should have, and none of the ones I shouldn't,
* change 852f06c, shuffling some code in `__wt_txn_parse_read_timestamp()` so we don't format timestamps unless we need to, and we never hold a lock while formatting a timestamp.

### DIFF
--- a/src/btree/bt_debug.c
+++ b/src/btree/bt_debug.c
@@ -1169,7 +1169,7 @@ __debug_modify(WT_DBG *ds, WT_UPDATE *upd)
 static int
 __debug_update(WT_DBG *ds, WT_UPDATE *upd, bool hexbyte)
 {
-	char hex_timestamp[WT_TS_HEX_SIZE];
+	char ts_string[WT_TS_INT_STRING_SIZE];
 
 	for (; upd != NULL; upd = upd->next) {
 		switch (upd->type) {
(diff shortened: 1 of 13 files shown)


michaelcahill:
This generally lgtm.

There's something specific I'd like to check in the Python API that I'll need to do tomorrow, then I'll do a final pass.

I haven't done a careful check, but is there really much overlap between this change and WT-4491?  i.e., could this change be made directly against develop without a ton of conflicts?

keithbostic:
> is there really much overlap between this change and WT-4491? i.e., could this change be made directly against develop without a ton of conflicts?

I did a quick test and there were 4 places it failed to patch cleanly. I don't know any way to do that other than applying the diff and cleaning it up, let me know if you know of a better approach.

src/btree/bt_debug.c
```diff
@@ -1267,8 +1268,7 @@ __debug_cell(WT_DBG *ds, const WT_PAGE_HEADER *dsk, WT_CELL_UNPACK *unpack)
 	WT_DECL_ITEM(buf);
 	WT_DECL_RET;
 	WT_SESSION_IMPL *session;
-	char hex_ts[3][WT_TS_HEX_SIZE];
-	const char *type;
```
sueloverso: My only comment on this is if the `type` is really not useful in debugging. It seems like removing the type is both a gratuitous change and probably should not be folded into the timestamp work.

keithbostic: I believe the `type` is duplicated information, compare the output of the initial output lines in `__debug_cell`:
```
   1275         WT_RET(ds->f(ds, "\t%s: len %" PRIu32,
   1276             __wt_cell_type_string(unpack->raw), unpack->size));
```
and the `type` assignments that were removed.

Let me know if you want this change pulled out into a separate ticket.

sueloverso: Thanks for the explanation. I only viewed the diff. I agree it is duplicated.

PR #4433, opened 2019-01-15
WT-4489 Store correct timestamp in cells on disk

keithbostic:
The branch to clean up on-disk timestamps: drive timestamps down into reconciliation's update-select code, and change the variable-length and row-store leaf-page reconciliation to pass along those timestamps when writing updates to the page.


keithbostic:
@agorrod, @michaelcahill, @bvpvamsikrishna:

A few weeks ago I had a conversation with Alex about using `_ts` or `_timestamp` as the timestamp variable suffix (for example, `durable_ts` vs. `durable_timestamp`). IIRC, Alex wanted something, but didn't feel strongly about which I used.

(I've been using `_ts` myself, because "damn 80-character window", but that's a different discussion.)

My unscientific count is there are 155 instances of `_timestamp`, and 210 instances of `_ts` in this branch, but that's unfair because I've been adding `start_ts` and `stop_ts`. The develop branch is more evenly divided at 159 vs. 149.

Should we pick one and do a naming pass?

I'd vote for `_ts`, but I won't fight about it.

keithbostic:
@agorrod, @michaelcahill: once you've had a chance to look at this branch a little, can we have a conversation about where we are?

I'm starting to run into the end of the changes I think I can make without breaking the existing implementation, which we obviously don't want to do until after the next release.

sueloverso:
If we're going to do a naming pass, I too vote of `_ts` because I'm more of a minimalist if we don't lose context and I think `ts` gives context.

agorrod:
I'm OK to move to `_ts` everywhere except in externally visible uses. I'd originally pushed back because I didn't really want yet another shortening to remember, and searching for `ts` is pretty much useless.
I'd definitely prefer to make that change as part of a different ticket.

keithbostic:
> There are two places where you flag TIMESTAMP-FIXME. Are they the places we'll need to update when switching to the new format? i.e: we can't get them right until we make a backward breaking change?

1. There's a TIMESTAMP-FIXME in `__rec_upd_select()`, and that's the bottom place where I drove timestamps, into the code that decides what item we're going to write onto this particular page. I'm anticipating that code will have to return the stop timestamp (either from the `WT_UPDATE` itself or from another `WT_UPDATE` in the chain or the underlying page, and I haven't done any work to make that stop timestamp "correct". We could probably drive the work down a bit further, but I don't think we can go much further and it doesn't seem worth doing before the next release.

1. The other 3 TIMESTAMP-FIXME notations are all in `__rec_col_var()`, in places where we are running through a list of records that may or may not be there. My guess is a missing record is going to require a tombstone on the page, but I'm honestly not sure about that, and I don't really have enough information to figure it out, yet. So, those are waiting for the tombstone discussion to happen. 

> I see that you were fairly conservative about where WT_UPDATE_SELECT was used - do you imagine that it will become a bit more pervasive over time? I ask because there seems to be a bit of a pattern of passing the WT_UPDATE_SELECT to a function, then unpacking the members into variables.

I think it may become more pervasive over time. I wanted a structure because `__rec_upd_select()` (the old `__rec_txn_read()`), returns 5 items, and I wanted to bundle them somehow. I immediately converted to the local variables we were using before because it made the change a lot simpler, I didn't want to rewrite any more of the principle reconciliation loops than I had to.

sulabhM:
@agorrod , I am looking at it at the moment.

src/include/btmem.h
```diff
@@ -827,8 +827,9 @@ struct __wt_page {
  */
 struct __wt_page_deleted {
 	volatile uint64_t txnid;		/* Transaction ID */
-	wt_timestamp_t timestamp;
-	wt_timestamp_t durable_timestamp;	/* aligned uint64_t timestamp */
+
+	wt_timestamp_t timestamp;		/* Timestamps */
```
sulabhM: Are we not calling these start_ts, durable_ts like at other places.

keithbostic: We're pretty evenly split between `XXX_timestamp` and `XXX_ts`. I've been just leaving it alone, once this ticket gets merged, I plan to do a pass and bring everything into alignment using `XXX_ts`.

sulabhM:
@keithbostic I have merged this change but was unable to delete the branch. Probably you are using this branch for a change on top of it?

PR #4434, opened 2019-01-18
WT-4529 Update lookaside schema to store durable timestamp

nehakhatri5:
As of this change the lookaside schema has been updated to store the durable timestamp of a WT_UPDATE structure.  A LAS read or write also reads/writes the durable timestamp.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -123,7 +123,7 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	WT_ITEM las_key, las_value;
 	WT_PAGE *page;
 	WT_UPDATE *first_upd, *last_upd, *upd;
-	wt_timestamp_t las_timestamp;
+	wt_timestamp_t durable_timestamp, las_timestamp;
 	size_t incr, total_incr;
 	uint64_t current_recno, las_counter, las_pageid, las_txnid, recno;
 	uint32_t las_id, session_flags;
@@ -183,20 +183,14 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 		/* Allocate the WT_UPDATE structure. */
 		WT_ERR(cursor->get_value(
 		    cursor, &las_txnid, &las_timestamp,
-		    &prepare_state, &upd_type, &las_value));
+		    &durable_timestamp, &prepare_state, &upd_type, &las_value));
 		WT_ERR(__wt_update_alloc(
 		    session, &las_value, &upd, &incr, upd_type));
 		total_incr += incr;
 		upd->txnid = las_txnid;
 		upd->timestamp = las_timestamp;
 		upd->prepare_state = prepare_state;
-		/*
-		 * Use the commit timestamp as the durable timestamp, since
-		 * non durable committed updates don't currently get written to
-		 * lookaside, so the two timestamps should always be identical.
-		 */
-		if (prepare_state != WT_PREPARE_INPROGRESS)
-			upd->durable_timestamp = upd->timestamp;
+		upd->durable_timestamp = durable_timestamp;
 
 		switch (page->type) {
 		case WT_PAGE_COL_FIX:
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -205,8 +205,8 @@ struct __wt_ovfl_reuse {
  * this way so that overall the lookaside table is append-mostly), a counter
  * (used to ensure the update records remain in the original order), and the
  * record's key (byte-string for row-store, record number for column-store).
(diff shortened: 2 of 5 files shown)


src/btree/bt_read.c
```diff
@@ -183,20 +183,14 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
 		upd->timestamp = las_timestamp;
 		upd->prepare_state = prepare_state;
-		/*
-		 * Use the commit timestamp as the durable timestamp, since
-		 * non durable committed updates don't currently get written to
-		 * lookaside, so the two timestamps should always be identical.
-		 */
-		if (prepare_state != WT_PREPARE_INPROGRESS)
-			upd->durable_timestamp = upd->timestamp;
+		upd->durable_timestamp = durable_timestamp;
```
agorrod: Did you remove the conditional setting of the durable timestamp for a reason?

src/include/btmem.h
```diff
@@ -205,8 +205,8 @@ struct __wt_ovfl_reuse {
  * this way so that overall the lookaside table is append-mostly), a counter
  * (used to ensure the update records remain in the original order), and the
  * record's key (byte-string for row-store, record number for column-store).
- * The value is the WT_UPDATE structure's transaction ID, timestamp, update's
- * prepare state, update type and value.
+ * The value is the WT_UPDATE structure's transaction ID, timestamp, durable
+ * timestamp, update's prepare state, update type and value.
```
agorrod: This list is getting long - I think a line per item would make it more understandable.

test/suite/test_durable_ts03.py
```diff
@@ -0,0 +1,76 @@
...
+def timestamp_str(t):
+    return '%x' %t
+
+# test_durable_ts03.py
+#    Check that the updates with durable timestamp newer than the stable
+#    timestamp fill up the cache and leave it stuck.
+class test_durable_ts03(wttest.WiredTigerTestCase):
+    # Reduce the cache size to 10MB to get the stuck cache.
+    def conn_config(self):
+        return 'cache_size=50MB'
```
agorrod: The comment doesn't match the number used. It's better to make the comment general to avoid that divergence. Something like "Use a small cache size so the cache can be pinned full with less data"

test/suite/test_durable_ts03.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_durable_ts03(self):
+        # Create a table.
+        uri = 'table:test_durable_ts03'
+        nrows = 600000
+        self.session.create(uri, 'key_format=i,value_format=u')
+        value1 = "aaaaa" * 100
+        value2 = "bbbbb" * 100
```
agorrod: This looks like about 285MB of data (600k entries, each at 500 bytes). Why do we need this much data to get a 50MB (10MB dirty) cache stuck full?

test/suite/test_durable_ts03.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Load the data into the table.
+        session = self.conn.open_session()
+        cursor = session.open_cursor(uri, None)
+        for i in range(0, nrows):
+            cursor[i] = value1
+        cursor.close()
+
+        # Set the stable timestamp to checkpoint initial data set.
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(100))
```
agorrod: It's quite unusual to switch from not using timestamps, to using timestamps - and also more difficult to reason about expected behavior. You should switch to setting an oldest/stable timestamp at the start. Then using a timestamp of 50 for all of the intermediate writes, then updating the oldest timestamp?

test/suite/test_durable_ts03.py
```diff
@@ -0,0 +1,76 @@
...
+        self.session.checkpoint()
+
+        # Update all the values within transaction. Commit the transaction with
+        # a durable timestamp newer than the stable timestamp.
+        cursor = session.open_cursor(uri, None)
+        self.assertEquals(cursor.next(), 0)
+        for i in range(1, nrows):
+            session.begin_transaction()
+            cursor.set_value(value2)
```
agorrod: Why do you use a different method of setting the content here than you do above? i.e above you use `cursor[i] = value`, here you use `cursor->next() && cursor.update(value)`

src/btree/bt_read.c
nehakhatri5: Yes - per discussion with Vamsi this check is no longer required. With this change, for every update that we write to the LAS table, we also write the durable timestamp. Therefore at the time of LAS read, it can be read unconditionally.

test/suite/test_durable_ts03.py
nehakhatri5: Updated to use the same method for setting the content.

PR #4437, opened 2019-01-21
WT-4537 WiredTiger cursor prev/next traversal failure on prepare retry

keithbostic:
Discussion in the ticket: [WT-4537](https://jira.mongodb.org/browse/WT-4537).


keithbostic:
autobuild: retest this please

bvpvamsikrishna:
@keithbostic, my understanding is we have to make similar changes for `prev` flow as well. Am I missing something?

keithbostic:
> my understanding is we have to make similar changes for `prev` flow as well. Am I missing something?

No, I think you're correct. I was planning to take a run at this one as soon as I get the last timestamp change working, so should be in a day or so.

bvpvamsikrishna:
@keithbostic, I have committed the changes to prev flow, based on next flow. Request you to review.

src/btree/bt_curprev.c
```diff
@@ -334,14 +351,14 @@ __cursor_var_prev(WT_CURSOR_BTREE *cbt, bool newpage)
...
-		goto new_page;
+		goto retry_insert;
 	}
 
 	/* Move to the previous entry and return the item. */
 	for (;;) {
 		__cursor_set_recno(cbt, cbt->recno - 1);
 
-new_page:	if (cbt->recno < cbt->ref->ref_recno)
+retry_insert:	if (cbt->recno < cbt->ref->ref_recno)
```
bvpvamsikrishna: @keithbostic, Could you review this change, I wanted your attention as I have doubt.  My doubt is "retry_insert", should it be labelled on line 364 instead of line 361 ? similar to next.

src/btree/bt_curnext.c
```diff
@@ -38,6 +42,7 @@ __cursor_fix_append_next(WT_CURSOR_BTREE *cbt, bool newpage)
 	 */
 	__cursor_set_recno(cbt, cbt->recno + 1);
 
+retry_insert:
```
keithbostic: I think that works (and maybe that's how I originally did it), but thinking about it now, I'd go with:
```
--- ../btree/bt_curnext.c       2019-02-15 17:00:05.955313586 -0500
+++ -   2019-02-15 17:00:17.121389922 -0500
@@ -63,7 +63,7 @@
                cbt->v = 0;
                cbt->iface.value.data = &cbt->v;
        } else {
-               WT_RET(__wt_txn_read(session, cbt->ins->upd, &upd));
+retry_insert:  WT_RET(__wt_txn_read(session, cbt->ins->upd, &upd));
                if (upd == NULL) {
                        cbt->v = 0;
                        cbt->iface.value.data = &cbt->v;
```
That makes more sense to me: `__wt_txn_read()` is the function that caused the problem, the state should be as we left it before that call, so we'll repeat the call rather than re-evaluate anything else about our current state. I think it's the pragmatic choice too -- yes, if we repeat the code we should end up at the same spot, but why chance that?

This comment applies to some of the other edits as well, of course.

src/btree/bt_curnext.c
```diff
@@ -86,21 +91,25 @@ __cursor_fix_next(WT_CURSOR_BTREE *cbt, bool newpage)
...
+		goto retry_insert;
 	}
 
 	/* Move to the next entry and return the item. */
 	if (cbt->recno >= cbt->last_standard_recno)
 		return (WT_NOTFOUND);
 	__cursor_set_recno(cbt, cbt->recno + 1);
 
-new_page:
+retry_insert:
```
keithbostic: It's going to be ugly, but I'd double up on the labels, so:
```
new_page:
retry_insert:    /* Code for new pages and restarts. */
```

bvpvamsikrishna:
@keithbostic, I have made the changes as per your suggestions, could you take another look.

keithbostic:
@bvpvamsikrishna, my apologies, I wasn't clear enough on my comments. I've pushed a change for your consideration, 5c9f5f8.

I tried to make the label names consistent (I like "restart_read", because I don't want confusion around "retry", and we're restarting a "read", not an "insert").

That's just semantic sugar, though.

More importantly, I think we want to make a statement that if we're restarting a read after a prepare conflict, the BTREE cursor should be correct, and absent local variable configuration we should
start at the same `__wt_txn_read()` that failed. The change there is that we jump as close to the `__wt_txn_read()` that failed as we can, setting local variables as necessary.

It seems to me that statement **should** be correct, and if it's not, then we run some risk of getting this wrong. Do you agree?

If I'm just missing the point here, feel free to revert that commit, of course.

keithbostic:
@bvpvamsikrishna, and, as you can see, I messed up and missed the setting of a local variable. We might be able to rework the code to avoid that additional/repeated search and tighten down the code we execute on the repeated attempt to read, but that's kind of what we're trying to avoid.

Anyway, let me know what you think!

bvpvamsikrishna:
>It seems to me that statement **should** be correct, and if it's not, then we run some risk of getting this wrong. Do you agree?

@keithbostic, I agree. I don't understand, but I have a doubt in a particular `prepared_delete` case. It is  
1. Insert (k,1)       : Txn 1, Session 1
2. commit_transaction : Txn 1, Session 1
3. delete (k)     : Txn 2, Session 1
4. prepare_transaction    : Txn 2, Session 1
5. At this point another Txn 3, Session 2  search(k-1)
6. cursor->next() : Txn 3, Session 2   --> return PREPARE_CONFLICT.
7. commit_transactin : Txn 2, Session 1.
If reconcilation runs, page will not have the insert list which was holding key k. So, if insert list doesn't have an entry for k, will cursor search be able to fetch the correct key ?
8. cursor->next() : Txn 3, Session 2. --> here the key is deleted, so it should fetch (K+1)

keithbostic:
@bvpvamsikrishna, if I understand correctly, reconciliation cannot run in step 7 because the cursor is positioned on the page.

bvpvamsikrishna:
> @bvpvamsikrishna, if I understand correctly, reconciliation cannot run in step 7 because the cursor is positioned on the page.

Yes. Changes look good to me.

michaelcahill:
I've spent some time staring at this (and in particular, trying to understand the state transitions we're tracking with various flags).  I think I understand it, and I can't see any problems.

Thanks @keithbostic and @bvpvamsikrishna: this has been tough, but I'm glad it is over the line.

I'll merge.

PR #4441, opened 2019-01-23
WT-4529 Update lookaside schema to store durable timestamp

nehakhatri5:


nehakhatri5:
I will create a new branch with the fix from WT-4539 included and open a pull request using that branch.

PR #4443, opened 2019-01-24
WT-4508 Fix a leak of index structure memory

sulabhM:


src/schema/schema_open.c
```diff
@@ -297,8 +297,19 @@ __schema_open_index(WT_SESSION_IMPL *session,
...
 		WT_ERR(cursor->get_key(cursor, &uri));
 		name = uri;
-		if (!WT_PREFIX_SKIP(name, tmp->data))
+
+		if (!WT_PREFIX_SKIP(name, tmp->data)) {
+			/*
+			 * We reached the end of index list, remove the rest of
+			 * in memory indices, they no longer exist.
+			 */
+			while (i < table->nindices) {
```
sulabhM: I am not sure if I should only do this if "idxname == NULL" which would imply that we are scanning all the indices.

agorrod:
@ddanderson would you please do a review of this fix?

src/schema/schema_open.c
ddanderson: It makes sense to me to do the removal in all cases, not just when `idxname == NULL`.

sulabhM:
@keithbostic, can you please review this one.

src/schema/schema_open.c
```diff
@@ -297,8 +297,19 @@ __schema_open_index(WT_SESSION_IMPL *session,
...
+				WT_ERR(__wt_schema_destroy_index(session,
```
keithbostic: Don't we want to proceed even if `__wt_schema_destroy_index()` returns an error, that is, continue freeing the memory? Otherwise I think we'll leak it. (That's what we do in `__wt_schema_close_table()`.)

I don't feel strongly.

PR #4444, opened 2019-01-25
WT-4532 Fix shared access to the default session.

michaelcahill:
In particular, `WT_CONNECTION::rollback_to_stable` can run for long enough that other periodic tasks (e.g., calls to `WT_CONNECTION::query_timestamp`) are called concurrently.  Disallow any access to data handles via the default session (outside open / close connection), and use a temporary internal session for rollback_to_stable.


michaelcahill:
autobuild: retest this please

(test timed out on tinderbox)

PR #4445, opened 2019-01-25
WT-4536 Compile wiredtiger under build_posix directory for Evergreen tests

lukech:
This changes in this PR is to address one technical debt to make wiredtiger compilation happen under `build_posix` directory (previously it was under top-level directory), to keep consistent with Jenkins tests and developer habit. 

Thanks @ddanderson for contributing 2 critical fixes to make Python test suite working, and now we have the flexibility of executing run.py under either `build_posix` or `build_posix/.libs` directory. The former option was chosen to be consistent with other test suites, and getting rid of the ugly `python_test_env_vars`.

In the mean time, another base change was made to use shell `tar` to replace evergreen build-in `archive.targz_pack` command in the Post section, to avoid the unwanted result of missing `build_posix/.lib` directory in the artifacts - `archive.targz_pack` seems having issue including hidden directories under the non-top-level directories. With this change, in order to avoid duplicated artifacts archiving for compile task, I made a 'trick' to add explicitly an archiving step followed by s3.put with `${build_id}`, and remove the wiredtiger directory (and the wiredtiger.tgz file) so that the common Post archiving commands fail to trigger. There could be errors at the end of compile task log, but won't be treated as task failure so I think it's fine.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -123,7 +123,7 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	WT_ITEM las_key, las_value;
 	WT_PAGE *page;
 	WT_UPDATE *first_upd, *last_upd, *upd;
-	wt_timestamp_t las_timestamp;
+	wt_timestamp_t durable_timestamp, las_timestamp;
 	size_t incr, total_incr;
 	uint64_t current_recno, las_counter, las_pageid, las_txnid, recno;
 	uint32_t las_id, session_flags;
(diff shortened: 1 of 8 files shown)


ddanderson:
@lukech, I needed to explicitly set LD_LIBRARY_PATH in run.py to guarantee that the 'wt' subprocess can find subordinate libraries.  Can you please review that change and redo the local regressions tests you ran yesterday?

lukech:
@ddanderson I pushed a change to use `os.environ.get` method to retrieve environment variable. Accessing the `os.environ` list directly error out (failing to finding `LD_LIBRARAY_PATH` environment variable) though I explicitly set it in the prefix of the python command. Regression testing was good on both Linux and MacOS now. 

I also triggered an Evergreen patch build: https://evergreen.mongodb.com/version/5c511fb161837d5bad0ee4d8

ddanderson:
@lukech, LGTM!  I'll let you merge, since I don't know how many tests you need to run on this.

lukech:
@agorrod I created another patch build, and this time I did not see the same failure pattern of s3.get timeout on the 2 macos tasks which failed in the previous patch build. 
https://evergreen.mongodb.com/version/5c52829661837d5bed0102cd
I think the code is ready for review.

PR #4448, opened 2019-01-29
WT-4296 Add aggressive mode to sweep server

guddnjs91:
Wake up sweep server to run every 10 msec in aggressive sweep mode.


src/conn/conn_sweep.c
```diff
@@ -292,8 +292,13 @@ __sweep_server(void *arg)
 	__wt_seconds(session, &last);
 	for (;;) {
 		/* Wait until the next event. */
-		__wt_cond_wait(session, conn->sweep_cond,
-		    min_sleep * WT_MILLION, __sweep_server_run_chk);
+		if (FLD_ISSET(conn->timing_stress_flags,
+			WT_TIMING_STRESS_AGGRESSIVE_SWEEP))
```
sulabhM: If a line wrap overs onto the next line, we indent it with 4 white spaces in addition to the indentation of the previous line. Please fix the indentation here.

src/conn/conn_sweep.c
```diff
@@ -292,8 +292,13 @@ __sweep_server(void *arg)
...
+			__wt_cond_wait(session, conn->sweep_cond,
+				min_sleep * 10000, __sweep_server_run_chk);
```
sulabhM: Same comment about indentation as above. Also, it is easier to read `10 * WT_THOUSAND` as compared to `10000`.

src/conn/conn_sweep.c
```diff
@@ -292,8 +292,13 @@ __sweep_server(void *arg)
...
+		else
+			__wt_cond_wait(session, conn->sweep_cond,
+				min_sleep * WT_MILLION, __sweep_server_run_chk);
```
sulabhM: Same comment about indentation as above.

src/include/connection.h
```diff
@@ -456,16 +456,17 @@ struct __wt_connection_impl {
...
-#define	WT_TIMING_STRESS_LOOKASIDE_SWEEP	0x002u
-#define	WT_TIMING_STRESS_SPLIT_1		0x004u
-#define	WT_TIMING_STRESS_SPLIT_2		0x008u
-#define	WT_TIMING_STRESS_SPLIT_3		0x010u
-#define	WT_TIMING_STRESS_SPLIT_4		0x020u
-#define	WT_TIMING_STRESS_SPLIT_5		0x040u
-#define	WT_TIMING_STRESS_SPLIT_6		0x080u
-#define	WT_TIMING_STRESS_SPLIT_7		0x100u
-#define	WT_TIMING_STRESS_SPLIT_8		0x200u
+#define	WT_TIMING_STRESS_AGGRESSIVE_SWEEP   0x001u
```
sulabhM: Indentation doesn't look correct. I suspect that you edited these values by hand. An easier way is to let them autogenerate. Look at `dist/flags.py` to understand how to get them autogenerating:

`To add a new flag declare it at the top of the flags list as:                  
#define WT_NEW_FLAG_NAME      0x0u`

test/format/config.h
```diff
@@ -334,6 +334,10 @@ static CONFIG c[] = {
 	  "maximum time to run in minutes",
 	  C_IGNORE, 0, 0, UINT_MAX, &g.c_timer, NULL },
 
+    { "timing_stress_aggressive_sweep",
```
sulabhM: The indentation and white spaces do not follow the rest of the file. Take another look and see if you can fix these.

src/conn/conn_sweep.c
sulabhM: We will check more often with reduced sleep, but we will still run lookaside sweep or file handle sweep at the same duration as before. Let us talk offline to figure out how we can resolve that.

src/conn/conn_sweep.c
```diff
@@ -350,6 +355,9 @@ __sweep_server(void *arg)
 
 		if (dead_handles > 0)
 			WT_ERR(__sweep_remove_handles(session));
+
+		/* Remember last sweep time. */
+		last = now;
```
sulabhM: @agorrod , I think we had a bug that went unnoticed. We were never updating the `last` time sweep ran. I had noticed in T2 data from Atlas that we were running sweep every 2 seconds, most likely because of this.

sulabhM: screen shot 2019-01-31 at 13 53 02

For instance, those connection sweeps happen every 2 seconds.

src/conn/conn_sweep.c
nehakhatri5: One way is to reduce the `conn->sweep_interval` config value [here](https://github.com/wiredtiger/wiredtiger/blob/92719d6bc9a83ce45c337db6a67adcc1354cca32/src/conn/conn_sweep.c#L387) and also reduce the `WT_LAS_SWEEP_SEC` in the aggressive mode.

guddnjs91:
autobuild: Please test this.

guddnjs91:
autobuild: test this please.

sulabhM:
@nehakhatri5 , please have a look.

src/conn/conn_sweep.c
```diff
@@ -279,21 +279,33 @@ __sweep_server(void *arg)
 	WT_SESSION_IMPL *session;
 	time_t last, now;
 	uint64_t last_las_sweep_id, min_sleep, oldest_id;
+	uint64_t sweep_interval;
```
nehakhatri5: nit: This can be declared in line # 281 along with other uint64_t variables,

PR #4449, opened 2019-01-29
WT-4516 Change WT data format to include timestamps in checkpoints

keithbostic:
@agorrod, here are the changes to aggregate the btree timestamps into the metadata during checkpoint.

As I hoped, it's a relatively simple change, assuming I haven't missed anything!

Note, this is a PR against the WT-4489 PR #4433.


src/reconcile/rec_write.c
```diff
@@ -5731,12 +5743,24 @@ __rec_write_wrapup(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
...
 			    true, F_ISSET(r, WT_REC_CHECKPOINT),
 			    r->wrapup_checkpoint_compressed));
 
+			/*
+			 * Checkpoint timestamps are saved in the Btree handle,
+			 * copied into the checkpoint when it completes.
+			 */
+			btree->oldest_start_ts = r->multi->addr.oldest_start_ts;
+			btree->newest_start_ts = r->multi->addr.newest_start_ts;
+			btree->newest_stop_ts = r->multi->addr.newest_stop_ts;
```
agorrod: I thought this would be an aggregated value, something like 
```
if (btree->oldest_start_ts > r->multi->addr.oldest_start_ts)
    btree->oldest_start_ts = r->multi->addr.oldest_start_ts;
```

Could you help me understand?

src/meta/meta_ckpt.c
```diff
@@ -427,23 +439,44 @@ __wt_meta_ckptlist_set(WT_SESSION_IMPL *session,
...
+		    ckpt->newest_start_ts, ckpt->newest_stop_ts);
 		if (strcmp(ckpt->name, WT_CHECKPOINT) == 0)
 			WT_ERR(__wt_buf_catfmt(session, buf,
 			    "%s%s.%" PRId64 "=(addr=\"%.*s\",order=%" PRId64
-			    ",time=%" PRIuMAX ",size=%" PRIu64
+			    ",time=%" PRIuMAX
+			    ",size=%" PRIu64
+			    ",oldest_start=%" PRIu64
+			    ",newest_start=%" PRIu64
+			    ",newest_stop=%" PRIu64
```
sueloverso: I think these should all have a `_ts` suffix to match the strings you're looking for in `__ckpt_load`.

src/include/meta.h
```diff
@@ -76,6 +76,9 @@ struct __wt_ckpt {
 
 	uint64_t write_gen;			/* Write generation */
 
+						/* Timestamps */
+	wt_timestamp_t oldest_start_ts, newest_start_ts, newest_stop_ts;
+
```
sueloverso: Why have these all on the same line?

src/meta/meta_ckpt.c
```diff
@@ -427,23 +439,44 @@ __wt_meta_ckptlist_set(WT_SESSION_IMPL *session,
...
 			    ckpt->write_gen));
 		else
 			WT_ERR(__wt_buf_catfmt(session, buf,
 			    "%s%s=(addr=\"%.*s\",order=%" PRId64
-			    ",time=%" PRIuMAX ",size=%" PRIu64
+			    ",time=%" PRIuMAX
+			    ",size=%" PRIu64
+			    ",oldest_start=%" PRIu64
+			    ",newest_start=%" PRIu64
+			    ",newest_stop=%" PRIu64
```
sueloverso: Same here.

src/meta/meta_ckpt.c
keithbostic: Thank you, @sueloverso!

And the fact I didn't catch this means I forgot to hook the checkpoint entries into the verify code.

src/include/meta.h
keithbostic: The addresses are a triplet, that is, there shouldn't ever be one or two of them alone (in previous incarnations of the code it was actually a structure). That's why putting them all on the same line made sense to me.

Let me know if you feel strongly.

src/meta/meta_ckpt.c
keithbostic: Fixed in the next push.

src/include/meta.h
sueloverso: I totally did not understand why they were on the same line, so that visual did not make sense to me. I don't think we do that visual anywhere else, do we? They're already grouped by whitespace. I think having them on separate lines and adding a comment before the group saying `/* These are a triplet, all or none should always appear. */ is more clear (to me anyway).

keithbostic:
@sueloverso, I've connected the checkpoint timestamps to the verify code, and I think this branch is ready to merge once you're happy with it.

The commit f77e7e8 is non-trivial. We've potentially been writing metadata files we can't read and they're out in the field. Hopefully, you'll find my explanation of why this doesn't matter convincing. Michael, I'd appreciate your review of that commit as well.

cc: @michaelcahill

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

michaelcahill:
I've done a review and made a few cosmetic changes in d9769c3, including addressing one of @sueloverso's comments about grouping timestamp fields onto a single line in `WT_CKPT`.

I agree with the change in f77e7e8 and I believe the argument that we should not have any broken  metadata in the wild (both because of the ranges of those values and because I think we would have heard about it).  There is some subtlety in passing timestamps where `int64_t`s are expected and coping with overflow, but that isn't unique to this change.

lgtm, I'll merge once my change passes testing.

PR #4450, opened 2019-01-29
WT-4548 Move timing locations to check if an individual operation took too long.

sueloverso:


ddanderson:
@agorrod, I was not aware of the specifics of the original problem.  Given that description, I'd have to agree with your assessment. @sueloverso, I don't have any better idea than to disable this test on machines that we know to be slow.

sueloverso:
Close this PR. We're not taking it.

PR #4451, opened 2019-01-29
WT-4053 Add statistic for data handle size

ZombieKim:
A stat variable "dh_conn_handle_size" takes the size of dhandle structure.

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -333,6 +333,7 @@ connection_stats = [
     # Dhandle statistics
     ##########################################
     DhandleStat('dh_conn_handle_count', 'connection data handles currently active', 'no_clear,no_scale'),
+    DhandleStat('dh_conn_handle_size', 'connection data handle size'),
     DhandleStat('dh_session_handles', 'session dhandles swept'),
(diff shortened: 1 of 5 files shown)


sulabhM:
This looks good to me.
@agorrod , can you have a look and check if this change is in the right direction.

dist/stat_data.py
```diff
@@ -333,6 +333,7 @@ def __init__(self, name, desc, flags=''):
     # Dhandle statistics
     ##########################################
     DhandleStat('dh_conn_handle_count', 'connection data handles currently active', 'no_clear,no_scale'),
+    DhandleStat('dh_conn_handle_size', 'connection data handle size'),
```
agorrod: I'd have expected this to have the `no_clear` and `no_scale` options set, otherwise the value will be scaled by t2, and will be reset to 0 when statistics are periodically reset, and never retr

nehakhatri5: It should also have the `size` flag set, because the value is a byte count.

sulabhM:
autobuild: test this please

nehakhatri5:
LGTM.

PR #4453, opened 2019-01-31
WT-4501 Increment cache_pool ref count under lock.

sueloverso:
@agorrod you've done the most work in the shared cache pool area. Please review this change. I've kicked off a run on tinderbox to see if it fixes the failure seen (after 10 hours, sadly). I'll let it run at least 24 hours.

I discovered the global reference count is not incremented under the same lock that it is decremented, and all the insert/remove calls are done on the connection list. I fixed that by dropping the process lock, acquiring the cache pool lock, and then re-acquiring the process lock (it is necessary to acquire in that order to avoid deadlock with other code taking them in that order).

While I believe I've maintained the original intent, I'm wondering why we hold the `wt_process` lock for the entire function. Looking later in the function, AFAICT we don't modify or access the `wt_process` structure again. So I think we might be able to drop the process lock and just acquire the `cp->cache_pool_lock` without re-acquiring the process lock.

### DIFF
--- a/src/conn/conn_cache_pool.c
+++ b/src/conn/conn_cache_pool.c
@@ -52,10 +52,10 @@ __wt_cache_pool_config(WT_SESSION_IMPL *session, const char **cfg)
 	WT_DECL_RET;
 	uint64_t chunk, quota, reserve, size, used_cache;
 	char *pool_name;
-	bool created, updating;
+	bool cp_locked, created, updating;
 
 	conn = S2C(session);
-	created = updating = false;
+	cp_locked = created = updating = false;
 	pool_name = NULL;
 	cp = NULL;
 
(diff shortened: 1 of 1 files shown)


sueloverso:
@agorrod Can you please review this change? I checked tinderbox today and the test has been running successfully for 5 days and well over 700 iterations.

src/conn/conn_cache_pool.c
```diff
@@ -117,7 +117,16 @@ __wt_cache_pool_config(WT_SESSION_IMPL *session, const char **cfg)
 		    "Attempting to join a cache pool that does not exist: %s",
 		    pool_name);
 
+	/*
+	 * At this point we have a cache pool to use. We need to take its
+	 * lock. We need to drop the process lock first to avoid deadlock
+	 * and acquire in the proper order.
+	 */
+	__wt_spin_unlock(session, &__wt_process.spinlock);
```
agorrod: I feel like releasing this lock here opens the possibility of the cache pool being destroyed by another thread. I think we'd need to validate the `cache pool` we'd saved is still OK after getting the process lock again.

Did you think about changing the other usage to acquire the process lock before the cache pool lock?

sueloverso: The destroy cannot happen because the creation code does not set the `WT_CONN_CACHE_POOL` flag until the very end, after we've relocked in the proper order. If a destroy tries to happen in the interim when the create code is at the point of redoing the locks, the destroy is a no-op and just returns due to that flag.

I think that the locking and create/destroy could use a general review but I think this change solves this bug.

agorrod:
Sounds good to me - thanks for the explanation.

PR #4455, opened 2019-02-04
WT-4514 Implement assert=durable_timestamp

bvpvamsikrishna:
I made changes to existing assert=commit_timestamp behaviour as well. The change is that when configured with assert=(commit_timestamp=always) it will perform the key consistent check as well.
@sueloverso, request you to comment on this change.


src/txn/txn.c
```diff
@@ -605,10 +605,13 @@ __txn_commit_timestamp_validate(WT_SESSION_IMPL *session)
...
 		    "timestamp set on this transaction");
 
+	commit_ts_keys = F_ISSET(txn,
+			     WT_TXN_TS_COMMIT_ALWAYS | WT_TXN_TS_COMMIT_KEYS);
+	durable_ts_keys = F_ISSET(txn, WT_TXN_TS_DURABLE_ALWAYS);
 	/*
 	 * If we're not doing any key consistency checking, we're done.
 	 */
-	if (!F_ISSET(txn, WT_TXN_TS_COMMIT_KEYS))
+	if (!commit_ts_keys && !durable_ts_keys)
```
bvpvamsikrishna: @sueloverso, I have changed this logic to perform key consistent check even if WT_TXN_TS_COMMIT_ALWAYS only is configured. I would like to know your opinion on this change. Do you think it is ok, or should we do key consistency check only if WT_TXN_TS_COMMIT_KEYS is configured?

src/btree/bt_handle.c
```diff
@@ -437,6 +437,14 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
...
 	else if (WT_STRING_MATCH("never", cval.str, cval.len))
 		FLD_SET(btree->assert_flags, WT_ASSERT_COMMIT_TS_NEVER);
+	WT_RET(__wt_config_gets(session,
+	    cfg, "assert.durable_timestamp", &cval));
+	/* Durable timestamp make sense only with a commit timestamp. */
+	if (WT_STRING_MATCH("always", cval.str, cval.len))
+		FLD_SET(btree->assert_flags,
+		    WT_ASSERT_DURABLE_TS_ALWAYS | WT_ASSERT_COMMIT_TS_ALWAYS);
+	else if (WT_STRING_MATCH("never", cval.str, cval.len))
+		FLD_SET(btree->assert_flags, WT_ASSERT_DURABLE_TS_NEVER);
```
sueloverso: Should this contain `WT_ASSERT_COMMIT_TS_NEVER`? That is, are durable and commit timestamps always together?

src/docs/transactions.dox
```diff
@@ -202,6 +202,13 @@ WT_PREPARE_CONFLICT error will be returned indicating that it is not possible
 to choose a version of data to return until a prepared transaction is resolved,
 it is reasonable to retry such operations.
 
+Durability of the data updates performed by a prepared transaction, on tables
+configured with log=(enabled=false), can be controlled by specifying a durable
+timestamp during WT_SESSION::commit_transaction. Checkpoint will consider the
+durable timestamp, instead of commit timestamp for persisting the data updates.
+If durable timestamp is not specified, commit timestamp will be considered as
```
sueloverso: Minor wording: If the durable timestamp is not specified, then the commit timestamp will be considered as the durable timestamp.

src/txn/txn.c
```diff
@@ -709,45 +718,42 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+	 * error if durable timestamp is specified for non-prepared
+	 * transactions.
+	 */
+	WT_ERR(__wt_config_gets_def(
+	    session, cfg, "durable_timestamp", 0, &cval));
+	if (cval.len != 0) {
+		if (!prepare)
+			WT_ERR_MSG(session, EINVAL,
+			    "durable_timestamp should not be given for "
+			    "non-prepared transaction");
```
sueloverso: This part of the change seems unrelated to implementing the `assert` functionality. Is it just a small bug fix or should it be part of a differrent ticket?

src/txn/txn.c
sueloverso: The reason that `key_consistent` is an explicit choice for commit is that it adds a lot of overhead. The `always` flag simply checks a bit. I think the `key_consistent` check is a level lower/deeper than just "this table should always/never use commit timestamps". I'd prefer to keep it separate. You may want to consider adding in a `key_consistent` setting for the durable timestamp too.

src/txn/txn.c
```diff
@@ -605,10 +605,13 @@ __txn_commit_timestamp_validate(WT_SESSION_IMPL *session)
 		WT_RET_MSG(session, EINVAL, "no commit_timestamp required and "
 		    "timestamp set on this transaction");
```
sueloverso: I am surprised there is no additional code checking the status of always/never and some transaction flags for the durable timestamp at this point in the code similar to what is there for commit and read.

src/btree/bt_handle.c
```diff
@@ -437,6 +437,14 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
 		FLD_SET(btree->assert_flags, WT_ASSERT_COMMIT_TS_KEYS);
 	else if (WT_STRING_MATCH("never", cval.str, cval.len))
 		FLD_SET(btree->assert_flags, WT_ASSERT_COMMIT_TS_NEVER);
+	WT_RET(__wt_config_gets(session,
+	    cfg, "assert.durable_timestamp", &cval));
+	/* Durable timestamp make sense only with a commit timestamp. */
+	if (WT_STRING_MATCH("always", cval.str, cval.len))
+		FLD_SET(btree->assert_flags,
+		    WT_ASSERT_DURABLE_TS_ALWAYS | WT_ASSERT_COMMIT_TS_ALWAYS);
```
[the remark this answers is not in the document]
sueloverso: What does `always` mean in this context? Does it mean that the `durable_timestamp` is always explicitly set, as opposed to the transaction defaulting to the `commit_timestamp`?

src/txn/txn.c
```diff
@@ -709,45 +718,42 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
-	WT_ERR(__txn_commit_timestamp_validate(session));
+		    session, "durable", ts, &cval, true));
+		txn->durable_timestamp = ts;
+	}
 
-	/* TODO : assert durable_timestamp. */
+	WT_ERR(__txn_commit_timestamps_validate(session));
 
 	/*
 	 * The default sync setting is inherited from the connection, but can
```
sueloverso: Not related to this file, but at the end of all the changes - I'm surprised not to find test changes or a new test. There are at least 4 assertion tests `test_assert0[1234].py` in the test suite. This should be added to existing tests or a new one created.

test/suite/test_assert_durable_ts01.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python
```
sueloverso: @bvpvamsikrishna thanks for writing this test. Please rename it to `test_assert05.py` so that it is clearly related to the other assert tests. All those tests perform tests with timestamps and the `assert` setting. This is exactly in the same area.

src/btree/bt_handle.c
bvpvamsikrishna: There can be a commit timestamp without a durable timestamp, but vice versa should never happen. Durable timestamp is only applicable for prepared transactions for which commit timestamp is mandatory. I will modify the `WT_ASSERT_DURABLE_TS_ALWAYS` to include `WT_ASSERT_COMMIT_TS_ALWAYS` .

src/txn/txn.c
bvpvamsikrishna: Yes, this part is not part of assert functionality, it is more of error checking if the user has given durable timestamp for a non-prepared transaction.

src/txn/txn.c
bvpvamsikrishna: I have added new test test_assert_durable_ts01.py

src/txn/txn_timestamp.c
```diff
@@ -646,11 +646,23 @@ __wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
...
+	if (F_ISSET(txn, WT_TXN_HAS_TS_DURABLE) &&
+	    txn->durable_timestamp < txn->commit_timestamp) {
+		__wt_timestamp_to_string(
+		    txn->durable_timestamp, ts_string[0], sizeof(ts_string[0]));
+		__wt_timestamp_to_string(
+		    txn->commit_timestamp, ts_string[1], sizeof(ts_string[1]));
+		WT_RET_MSG(session, EINVAL,
+		    "%s timestamp %s older than the commit timestamp %s "
+		    "for this transaction",
+		    name, ts_string[0], ts_string[1]);
```
bvpvamsikrishna: This is also not part of assert functionality.

test/suite/test_assert_durable_ts01.py
bvpvamsikrishna: sure. Thank you.

bvpvamsikrishna:
@sueloverso, I need you opinion about error handling of durable timestamp being part of this PR. Do you feel it is better I create a new ticket for error handling of durable timestamp and move the code changes to a separate ticket ? A part from that I believe other comments are addressed.

src/txn/txn.c
```diff
@@ -604,11 +608,21 @@ __txn_commit_timestamp_validate(WT_SESSION_IMPL *session)
...
+	    F_ISSET(txn, WT_TXN_HAS_TS_DURABLE) &&
+	    txn->mod_count != 0)
+		WT_RET_MSG(session, EINVAL, "no durable_timestamp required and "
+		    "durable timestamp set on this transaction");
 
 	/*
 	 * If we're not doing any key consistency checking, we're done.
 	 */
-	if (!F_ISSET(txn, WT_TXN_TS_COMMIT_KEYS))
+	if (!F_ISSET(txn, WT_TXN_TS_COMMIT_KEYS | WT_TXN_TS_DURABLE_ALWAYS))
```
sueloverso: This comment is a follow-on to an earlier discussion. I think you should add a `key_consistent` option for `durable_timestamp` and have this code triggered only on that. The key consistency checks are a lot more processing. The `always` and `never` settings are meant to be fast high-level sanity checks.

When you do that, you'll have to consider whether durable's key consistency checking does, or does not, imply the commit key consistency checks. Right now the code below durable does imply commit consistency checking. But key consistency checking, particularly for durable timestamps is now a very large performance hit that we might not want to pay with a simple `always` check.

src/txn/txn.c
sueloverso: That is fine that it is a small fix you noticed and added while you were working in this area of code. I just wanted to make sure I wasn't misunderstanding what I was reading.

test/suite/test_assert05.py
```diff
@@ -0,0 +1,120 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# test_assert05.py
+#   Timestamps: assert durable timestamp settings
```
sueloverso: This test looks good. Thanks. If you add `key_consistency` checking, the already-existing test for it for commit timestamps is in `test_assert04.py`.

bvpvamsikrishna:
@sueloverso , I believe all your comments are incorporated, could you take another look at it.

src/include/btree.h
```diff
@@ -103,9 +103,10 @@ struct __wt_btree {
 #define	WT_ASSERT_COMMIT_TS_KEYS	0x02u
 #define	WT_ASSERT_COMMIT_TS_NEVER	0x04u
 #define	WT_ASSERT_DURABLE_TS_ALWAYS	0x08u
-#define	WT_ASSERT_DURABLE_TS_NEVER	0x10u
-#define	WT_ASSERT_READ_TS_ALWAYS	0x20u
-#define	WT_ASSERT_READ_TS_NEVER		0x40u
+#define	WT_ASSERT_DURABLE_TS_KEYS   0x10u
```
sueloverso: Did you add this yourself? The spacing between the name and value isn't correct. It should be a tab not spaces. The way to add a new flag is to add it at the top of the list (right below the `/* AUTOMATIC FLAG VALUE GENERATION START */`) give it the name you want and add its value as `0x00u`. Then the scripts will automatically alphabetize it and recompute the bit fields. But you need to have the tabs aligned to start.

test/suite/test_assert06.py
```diff
@@ -0,0 +1,403 @@
...
+    def apply_timestamps(self, timestamp):
+        self.session.prepare_transaction(
+            'prepare_timestamp=' + timestamp_str(timestamp))
+        self.session.timestamp_transaction(
+            'commit_timestamp=' + timestamp_str(timestamp))
+        self.session.timestamp_transaction(
+            'durable_timestamp=' + timestamp_str(timestamp))
+
+    def test_timestamp_alter(self):
+        base = 'assert04'
```
sueloverso: This should be `assert06` or generically named `assert` (only that it doesn't have the wrong test number in it - feel free to make `test_assert04.py` generic too if you go that route).

test/suite/test_assert06.py
```diff
@@ -0,0 +1,403 @@
...
+
+        c = self.session.open_cursor(uri)
+        self.session.begin_transaction()
+        c['key_ts1'] = 'value4'
+        self.session.prepare_transaction(
+            'prepare_timestamp=' + timestamp_str(4))
+        self.session.timestamp_transaction(
+            'commit_timestamp=' + timestamp_str(4))
+        self.session.timestamp_transaction(
+            'durable_timestamp=' + timestamp_str(4))
```
sueloverso: Either the above should call `self.apply_timestamps(4)` or it should have a comment about why that isn't appropriate here. I see no differences so I think it should just call the function.

test/suite/test_assert06.py
```diff
@@ -0,0 +1,403 @@
...
+        c = self.session.open_cursor(uri)
+        self.session.begin_transaction()
+
+        self.assertEquals(c['key_ts1'], 'value5')
+        
+        self.assertEquals(c['key_nots'], 'value_nots3')
+        self.session.commit_transaction()
+
+        c.close()
+
```
sueloverso: From here down it looks like the test is not quite finished. Key consistency should be able to be turned off, AFAIK.

src/txn/txn.c
```diff
@@ -664,7 +703,11 @@ __txn_commit_timestamp_validate(WT_SESSION_IMPL *session)
 				op_timestamp = txn->commit_timestamp;
 			if (op_timestamp < upd->start_ts)
```
sueloverso: This code is now assuming that `WT_TXN_TS_DURABLE_KEYS` implies `WT_TXN_TS_COMMIT_KEYS`. Either the commit-related flag should be set when we parse the durable-related one in `bt_handle.c` (and a comment/documentation added stating that) or if they're separate, this `if` needs to also be conditional on `WT_TXN_TS_COMMIT_KEYS` being set.

test/suite/test_assert06.py
```diff
@@ -0,0 +1,384 @@
...
+        self.assertRaisesWithMessage(wiredtiger.WiredTigerError,
+            lambda: self.session.commit_transaction(), msg_usage)
+        c.close()
+
+        # Detect using a timestamp on the non-timestamp key.
+        c = self.session.open_cursor(uri)
+        self.session.begin_transaction()
+        c['key_nots'] = 'value_nots3'
+        self.session.commit_transaction()
+        c.close()
```
sueloverso: Do we need this operation here? At first I was confused thinking that you forgot to "use the timestamp" in the lines above. I eventually saw the lines below. If the operation above is not needed, remove it. If it is needed, please add a comment like: "Detect using a timestamp on the non-timestamp key. We first must perform a non-timestamped operation so that..."

test/suite/test_assert06.py
```diff
@@ -0,0 +1,384 @@
...
+            lambda: self.session.commit_transaction(), msg_usage)
+        c.close()
+        self.session.checkpoint()
+
+        c = self.session.open_cursor(uri)
+        self.assertEquals(c['key_ts1'], 'value5')
+        self.assertEquals(c['key_nots'], 'value_nots3')
+        c.close()
+        # Test to make sure that key consistency cannot be turned off
+        # after turning it on
```
sueloverso: Is this true? Why can't the assert be turned off? (In any case, please end the sentence with a period.)

test/suite/test_assert06.py
```diff
@@ -0,0 +1,384 @@
...
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(5))
+        self.session.alter(uri, cfg_off)
```
sueloverso: Hmm, if it cannot be turned off, I'd expect this to be a lambda error checking. I think the comment needs to be corrected above.

src/btree/bt_handle.c
```diff
@@ -437,6 +437,22 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
...
+	 */
+	WT_RET(__wt_config_gets(session,
+	    cfg, "assert.durable_timestamp", &cval));
+	if (WT_STRING_MATCH("always", cval.str, cval.len))
+		FLD_SET(btree->assert_flags,
+		    WT_ASSERT_COMMIT_TS_ALWAYS | WT_ASSERT_DURABLE_TS_ALWAYS);
+	else if (WT_STRING_MATCH("key_consistent", cval.str, cval.len))
+		FLD_SET(btree->assert_flags, WT_ASSERT_DURABLE_TS_KEYS);
+	else if (WT_STRING_MATCH("never", cval.str, cval.len))
+		FLD_SET(btree->assert_flags, WT_ASSERT_DURABLE_TS_NEVER);
```
sueloverso: While reviewing the test, it occurs to me that we don't check for non-sensical combinations with the `commit_timestamp` setting. Something like `durable_timestamp=always,commit_timestamp=never` should get an error (I think that might be the only one actually).

bvpvamsikrishna: The error check for the combination `durable_timestamp=always, commit_timestamp=never` is actually part of the `commit_transaction` API.  `commit_timestamp` is optional for non-prepared transactions only, and if `durable_timestamp` is specified for non-prepared transaction, error will be returned. So, we thought no need to check this condition as part of assert.  If you feel strongly that it is better we add this check as part of assert, will do it.

sueloverso: Thanks for that explanation. That's fine.

PR #4456, opened 2019-02-05
WT-4369 random_directio shouldn't assume children threads wrote any data

sulabhM:


sueloverso:
@sulabhM Good work finding this. I've been thinking about whether we should have code that instead forces the parent to wait until at least one record is written. But I think that ends up adding complexity.

PR #4463, opened 2019-02-07
WT-4492 Add Evergreen compatibility tests for versioned data format

lukech:
Add a new Evergreen test (task) to verify backward and forward compatibility for a set of mongodb release pairs, to make sure the newer release wt binary is able to read objects generated by older release, and vice versa. The recent data format changes (PM-1309, targeting for MongoDB 4.2) will also be covered by this new test.

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -965,6 +965,18 @@ tasks:
             ulimit -c unlimited 
             largescale/run-million-collection-test.sh .
 
+  - name: compatibility-test-for-mongodb-releases
+    commands: 
+      - func: "fetch source"
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger"
+          script: | 
+            set -o errexit
+            set -o verbose
+            test/evergreen/compatibility_test_for_mongodb_releases.sh
+
+
 buildvariants:
 - name: ubuntu1404
   display_name: Ubuntu 14.04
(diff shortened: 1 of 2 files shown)


test/evergreen/compatibility_test_for_mongodb_releases.sh
```diff
@@ -0,0 +1,136 @@
...
+##############################################################################################
+# Check releases to ensure forward and backward compatibility.
+##############################################################################################
+
+###########################################################################
+# Return the most recent version of the tagged release.
+###########################################################################
+get_release()
+{ 
+	echo "$(git tag | grep "^mongodb-$1.[0-9]" | sort -n | sed -e '$p' -e d)"
```
lukech: `sort -n` seems ok to handle 1 decimal digit version numbers in this case. If there are 2 decimal digits available e.g. 3.4.17, it won't query out the latest version. `sort -V` seems working better.

test/evergreen/compatibility_test_for_mongodb_releases.sh
```diff
@@ -0,0 +1,136 @@
...
+	git clone --quiet https://github.com/wiredtiger/wiredtiger.git "wt.$1" > /dev/null || return 1
+	cd "wt.$1" || return 1
+
+	config=""
+	config+="--enable-snappy "
+
+	case "$1" in
+	"develop")
+		branch="develop";;
+	"develop-timestamps")
```
lukech: Would be useful to add comments here to mention develop here means the future 4.2 - the only mongodb release that supports both enabling and disabling timestamps in data format. Once 4.2 is released we'd need to update this script.

lukech:
@keithbostic I made a few minor changes to your shell script. Let me know if they do not make sense. 
@agorrod Could you help to review the changes? 

The patch build for the new Evergreen variant/task: https://evergreen.mongodb.com/version/5c5bb10f61837d36dfcb4ab1

test/evergreen.yml
```diff
@@ -1041,6 +1053,14 @@ buildvariants:
   tasks:
     - name: million-collection-test
 
+- name: weekly-tests
+  display_name: Weekly run tests
```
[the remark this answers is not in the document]
agorrod: I prefer `compatibility tests`, we might want to change the frequency at which we run the compatibility tests, and they seem to define a unique category well to me.

lukech:
I pushed a change to update the variant name to `Compatibility tests`. PR testing passed, will go ahead to merge.

PR #4464, opened 2019-02-07
WT-4483 Improve caching of small updates to large values. (#4403)

lukech:
Be less aggressive about creating full copies of a value, especially when the values are large.

(cherry picked from commit 42865162b40203367904627e5fc39d3274665cc4)


lukech:
@agorrod This is a clean cherry-pick. Can you help to take a quick look?

PR #4469, opened 2019-02-08
WT-4484 workgen wtperf: copy generated python source to test directory, remove stray temp file

ddanderson:


ddanderson:
@sueloverso, yes there are no diffs to add a remove call.  The remove call was already there, but due to a some quirk in how execfile() was working, the script started over at the top, so a second temp file was made (and removed) without the first one being removed.  Very odd - I found no documentation for this, but changing the call from execfile (which is obsolete in Python 3) fixed it.

PR #4471, opened 2019-02-08
WT-4559 Create a marker file when a test leaves a database corrupted.

ddanderson:


ddanderson:
@lukech please review this and see if this helps with the ticket.  Tests that leave databases corrupted will have a special marker file.  I marked the two tests that I know about.

lukech:
@ddanderson Thanks for making the changes. LGTM. 
@keithbostic do you mind to take another look. 
I have no more idea if there are more tests that are designed to corrupt the database. But the new dump script in WT-4498 should be able to tell if there are more.

keithbostic:
@lukech, this looks good to me. Obviously, you need to hook the new file into your script, but I think that should resolve WT-4559.

I'm happy to review any future failures that you find, to confirm if they should be removed from your script's consideration or if they are real issues we should investigate, just send them my way.

lukech:
@keithbostic I'm going to add the checking logic for the new marker files in the dump script as part WT-4498. Will let you know if there are more failures observed with that change. Will merge this PR.

PR #4473, opened 2019-02-11
WT-4526 Allow durable updates to be evicted to the lookaside file.

nehakhatri5:
This changeset evicts the durable updates to LAS, ensures that the checkpoint honors the durable timestamp and skips reading from LAS respecting the durable timestamp of the update.


nehakhatri5:
Test this please.

src/cache/cache_las.c
```diff
@@ -1137,7 +1147,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 			 */
 			if (upd_type == WT_UPDATE_BIRTHMARK &&
 			    __wt_txn_visible_all(
-			    session, las_txnid, las_timestamp) &&
+			    session, las_txnid, durable_timestamp) &&
```
agorrod: This change isn't obvious to me - can you add an assertion that `durable_timestamp != WT_TS_MAX && durable_timestamp >= las_timestamp`, that will make me more comfortable.

src/cache/cache_las.c
```diff
@@ -426,16 +426,26 @@ __wt_las_page_skip_locked(WT_SESSION_IMPL *session, WT_REF *ref)
...
+		 * Skip lookaside pages during checkpoint if all the unstable
+		 * durable updates are in the future.
+		 */
+		if (WT_SESSION_IS_CHECKPOINT(session) &&
+		    txn->read_timestamp <
+		    ref->page_las->unstable_durable_timestamp)
+			return (true);
+
+		if (txn->read_timestamp < ref->page_las->unstable_timestamp)
+			return (true);
```
agorrod: This is the only code that uses `unstable_timestamp` anymore, why is it important for the two cases to be dealt with differently?

i.e: what is the advantage of tracking both `unstable_timestamp` and `unstable_durable_timestamp`?

nehakhatri5: `unstable_timestamp` is required for ordinary cursor reads. An ordinary read should not skip reading LAS if `read_timestamp(210)` < `durable_timestamp(220)` but `read_timestamp(210)` > `commit_timestamp(200)`.  Ordinary read should still honor the `commit_timestamp` stored in the `unstable_timestamp`. That's why we still track the `unstable_timestamp`.

agorrod: Does that mean that keeping just the unstable check here would be correct, but might mean that checkpoints read more content from lookaside than necessary?

nehakhatri5: That's right.

agorrod: It would be valuable to capture that in a comment somewhere (probably here). Could you expand the comments?

src/reconcile/rec_write.c
```diff
@@ -51,6 +51,7 @@ typedef struct {
 
 	/* Lookaside boundary tracking. */
 	uint64_t unstable_txn;
+	wt_timestamp_t unstable_durable_timestamp;
 	wt_timestamp_t unstable_timestamp;
```
agorrod: The `unstable_durable_timestamp` doesn't seem to be being initialized in the same place as the `unstable_timestamp` - was that deliberate?

nehakhatri5: That was not intentional.

nehakhatri5:
Test this please.

src/reconcile/rec_write.c
```diff
@@ -2968,6 +2977,8 @@ done:	if (F_ISSET(r, WT_REC_LOOKASIDE)) {
 		WT_ASSERT(session, r->unstable_txn != WT_TXN_NONE);
 		multi->page_las.max_timestamp = r->max_timestamp;
 		multi->page_las.unstable_timestamp = r->unstable_timestamp;
+		multi->page_las.unstable_durable_timestamp =
+		    r->unstable_durable_timestamp;
```
agorrod: Since the more general assertion didn't turn out to be OK - could you add an assertion in here that:

`r->unstable_durable_timestamps > r->unstable_timestamp`

nehakhatri5:
@agorrod take a look at the updated changes and acknowledge if it addresses your concerns.

src/cache/cache_las.c
```diff
@@ -1127,6 +1143,32 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+			 * locked state.
+			 */
+			WT_ASSERT(session,
+			    prepare_state != WT_PREPARE_LOCKED);
+			/*
+			 * Never expect an update with durable timestamp
+			 * as max timestamp.
+			 */
+			WT_ASSERT(session,
+			    durable_timestamp != WT_TS_MAX);
```
agorrod: Please fold the three enabled assertions into a single statement.

src/reconcile/rec_write.c
```diff
@@ -1394,8 +1395,8 @@ __rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 		r->max_txn = max_txn;
 
 	/* Update the maximum timestamp. */
-	if (first_ts_upd != NULL && r->max_timestamp < first_ts_upd->start_ts)
-		r->max_timestamp = first_ts_upd->start_ts;
+	if (first_ts_upd != NULL && r->max_timestamp < first_ts_upd->durable_ts)
+		r->max_timestamp = first_ts_upd->durable_ts;
```
agorrod: Why did you choose to switch to the durable timestamp here?

src/reconcile/rec_write.c
```diff
@@ -1476,9 +1477,11 @@ __rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 	if (F_ISSET(r, WT_REC_LOOKASIDE) && r->las_skew_newest) {
 		if (WT_TXNID_LT(r->unstable_txn, first_upd->txnid))
 			r->unstable_txn = first_upd->txnid;
-		if (first_ts_upd != NULL &&
-		    r->unstable_timestamp < first_ts_upd->start_ts)
-			r->unstable_timestamp = first_ts_upd->start_ts;
+		if (first_ts_upd != NULL)
+			if (r->unstable_durable_timestamp <
```
agorrod: It's not compliant with coding standards to have nested `if` statements without need - these should be folded together.
On the other hand the change looks wrong to me. I'd have expected the original assignment to happen as well. i.e this would become:

```
         if (first_ts_upd != NULL) {                                     
                        WT_ASSERT(session,                                      
                            first_ts_upd->start_ts <= first_ts_upd->durable_ts);
                        if (r->unstable_durable_timestamp <                     
                            first_ts_upd->durable_ts)                           
                                r->unstable_durable_timestamp =                 
                                    first_ts_upd->durable_ts;                   
                        if (r->unstable_timestamp < first_ts_upd->start_ts)     
                                r->unstable_timestamp = first_ts_upd->start_ts; 
                }
```

src/reconcile/rec_write.c
```diff
@@ -1488,7 +1491,11 @@ __rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 			    WT_TXNID_LT(upd->txnid, r->unstable_txn))
 				r->unstable_txn = upd->txnid;
 			if (upd->start_ts < r->unstable_timestamp)
-				r->unstable_timestamp = upd->start_ts;
+				r->unstable_timestamp =
+				r->unstable_durable_timestamp = upd->start_ts;
+			if (upd->durable_ts < r->unstable_durable_timestamp &&
+				upd->durable_ts > r->unstable_timestamp)
```
agorrod: I don't understand the second part of the clause here, I tried the following instead and it passed the tests I tried with it:

```
                        /*                                                      
                         * The durable timestamp is always set, and usually     
                         * the same as the start timestamp, which makes it OK   
                         * to use the two independently and be confident both   
                         * will be set.                                         
                         */                                                     
                        WT_ASSERT(session, upd->durable_ts >= upd->start_ts);   
                        if (upd->start_ts < r->unstable_timestamp)              
                                r->unstable_timestamp = upd->start_ts;          
                        if (upd->durable_ts < r->unstable_durable_timestamp)    
                                r->unstable_durable_timestamp = upd->durable_ts;
```

src/reconcile/rec_write.c
```diff
@@ -2967,7 +2974,11 @@ done:	if (F_ISSET(r, WT_REC_LOOKASIDE)) {
 		multi->page_las.unstable_txn = r->unstable_txn;
 		WT_ASSERT(session, r->unstable_txn != WT_TXN_NONE);
 		multi->page_las.max_timestamp = r->max_timestamp;
+		WT_ASSERT(session, r->unstable_durable_timestamp >=
+		    r->unstable_timestamp);
```
agorrod: You should split after `session,` - it's always preferred to keep logical expressions on a single line.

src/reconcile/rec_write.c
nehakhatri5: As discussed, the suggested assertion would fail for updates with prepare in-progress state. So I have further augmented this assertion to check for the prepare state the update.

src/reconcile/rec_write.c
nehakhatri5: As discussed, this assertion would fail too if all the updates were with prepare in-progress state. So introduced a boolean to track the prepare in-progress state for all the visited updates.

nehakhatri5:
Test this please.

src/reconcile/rec_write.c
nehakhatri5: As discussed offline, it is OK to switch to durable timestamp here. Taken a look at all the places where `max_timestmap` can have an effect. This is a behavior change and can have some knock-on effect in eviction

nehakhatri5:
Test this please.

nehakhatri5:
Test this please.

nehakhatri5:
autobuild: test this please.

lukech:
retest this please

lukech:
retest this please

PR #4474, opened 2019-02-12
WT-4558 WiredTiger connection statistics cursor incorrectly provides doubled-up values.

ZombieKim:


src/cursor/cur_stat.c
```diff
@@ -357,6 +357,7 @@ __curstat_conn_init(WT_SESSION_IMPL *session, WT_CURSOR_STAT *cst)
 	 * Optionally clear the connection statistics.
 	 */
 	__wt_conn_stat_init(session);
+	memset(&cst->u.conn_stats, 0, sizeof(WT_CONNECTION_STATS));
```
nehakhatri5: Would this  clear the value for the `no_clear` statistics too?

ZombieKim: Yes, it is. But only clear the aggregated conn stats in the cursor.
Global conn stat is cleared at `__wt_stat_coninection_clear_all()` considering the `no_clear` flag, which is at right after this `memset`,

ZombieKim: Hello @keithbostic, Can you please let me know if there is some reason for using "+=" operator rather than "=" operator inside the `__wt_stat_connection_aggregate()` function? Or can it be simply changed to "=" operator?

keithbostic: > Can you please let me know if there is some reason for using "+=" operator rather than "=" operator inside the `__wt_stat_connection_aggregate()` function? Or can it be simply changed to "=" operator?

I'm sorry, but I just don't remember!

My guess is it could be "=", but "+=" makes the auto-generation of the code  simpler because it's all the same, and there was some chance there would be a need for aggregation from multiple connections someday. The only downside I can see is it's marginally slower, but I can't imagine that matters here.

keithbostic: Looking at this, I was thinking a slightly different change might be better:
```
diff --git a/src/cursor/cur_stat.c b/src/cursor/cur_stat.c
index 0118f3429..f4aee9a8c 100644
--- a/src/cursor/cur_stat.c
+++ b/src/cursor/cur_stat.c
@@ -357,6 +357,7 @@ __curstat_conn_init(WT_SESSION_IMPL *session, WT_CURSOR_STAT *cst)
         * Optionally clear the connection statistics.
         */
        __wt_conn_stat_init(session);
+       __wt_stat_connection_init_single(&cst->u.conn_stats);
        __wt_stat_connection_aggregate(conn->stats, &cst->u.conn_stats);
        if (F_ISSET(cst, WT_STAT_CLEAR))
                __wt_stat_connection_clear_all(conn->stats);
```
That is entirely untested and maybe I'm missing the point, so be careful.

Anyway, the idea is to make `__curstat_conn_init()` and `__curstat_file_init()` look the same.

keithbostic:
Are we going to add a static test for this bug? If we do, it might be interesting to include a similar test for a complex table with multiple column groups and indexes, and confirm we're correctly clearing ad aggregating statistics in that case as well. There's some special handling of that in `__wt_curstat_table_init()`, from 1673308c, and I doubt it's explicitly tested anywhere.

src/cursor/cur_stat.c
ZombieKim: Thank you for your answer. Now I can understand it!

ZombieKim:
autobuild: test this please.

ZombieKim:
autobuild: test this please.

ZombieKim:
> Are we going to add a static test for this bug? If we do, it might be interesting to include a similar test for a complex table with multiple column groups and indexes, and confirm we're correctly clearing ad aggregating statistics in that case as well. There's some special handling of that in `__wt_curstat_table_init()`, from [1673308](https://github.com/wiredtiger/wiredtiger/commit/1673308ce976e05cd4b7f6d0179a99cead44c395), and I doubt it's explicitly tested anywhere.



> Are we going to add a static test for this bug? If we do, it might be interesting to include a similar test for a complex table with multiple column groups and indexes, and confirm we're correctly clearing ad aggregating statistics in that case as well. There's some special handling of that in `__wt_curstat_table_init()`, from [1673308](https://github.com/wiredtiger/wiredtiger/commit/1673308ce976e05cd4b7f6d0179a99cead44c395), and I doubt it's explicitly tested anywhere.

ZombieKim:
autobuild: test this please.

sulabhM:
LGTM. I will merge

PR #4477, opened 2019-02-12
WT-4576 Reworked code collecting capacity stats to address coverity error

ddanderson:


ddanderson:
@sueloverso can you review?  I think this will pass coverity since it uses the same style switch statement as in earlier in the code.

PR #4478, opened 2019-02-13
WT-4577 test/format LSM deadlock

keithbostic:
@sueloverso, @agorrod, this change reverts the functional part of the changes for WT-4561 (commit 6aafde0c).

The race that commit was supposed to fix doesn't exist:
```
__lsm_free_chunks() ->
__lsm_drop_file() ->
__wt_schema_drop() ->
__schema_drop() ->
__drop_file() ->
__wt_schema_backup_check() acquires the hot-backup lock.
```
The original code was:
```
/*
 * Don't remove files if a hot backup is in progress.
 *
 * The schema lock protects the set of live files, this check
 * prevents us from removing a file that hot backup already
 * knows about.
 */
if (S2C(session)->hot_backup)
        break;
```
which is wrong because the check can race with the backup cursor, the check doesn't do anything meaningful (other than maybe skipping a schema drop attempt that's likely to fail). I don't know enough about hot backup to know if there is a real issue here for LSM, or if `__wt_schema_backup_check()` is a sufficient check for LSM discarding chunks.

This change resolves the deadlock that WT-4561 introduced, and doesn't make anything worse. If `__wt_schema_backup_check()` is insufficient for LSM chunk discard, then we should open a new ticket to deal with that issue.

### DIFF
--- a/src/lsm/lsm_work_unit.c
+++ b/src/lsm/lsm_work_unit.c
@@ -818,10 +818,7 @@ err:	/* Flush the metadata unless the system is in panic */
 int
 __wt_lsm_free_chunks(WT_SESSION_IMPL *session, WT_LSM_TREE *lsm_tree)
 {
-	WT_CONNECTION_IMPL *conn;
 	WT_DECL_RET;
-
-	conn = S2C(session);
 
 	if (lsm_tree->nold_chunks == 0)
 		return (0);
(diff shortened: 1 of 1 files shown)


sueloverso:
Do we want to consider, instead, just reverting the original change? I think this diff leaves it slightly worse off than the original code because now we don't check conn->hot_backup at all, so if a hot backup is actually in progress, this code keeps going. My vote is to just revert the original change.

What was the origin of the original ticket? It doesn't say. Did you find this by inspection or did some test fail?

keithbostic:
> Do we want to consider, instead, just reverting the original change? I think this diff leaves it slightly worse off than the original code because now we don't check conn->hot_backup at all, so if a hot backup is actually in progress, this code keeps going. My vote is to just revert the original change.

I don't want to revert to the original change (I think the code isn't useful and the comment is wrong). Hot backup will eventually be checked in the schema drop code.

> Did you find this by inspection or did some test fail?

By inspection.

sueloverso:
Ok. I double checked and you're correct that schema drop will check that. My other thought regarding the reverting suggestion was that the subfunction and wrapper are really not necessary now. I don't feel strongly about pulling `lsm_free_chunks` back up into `wt_lsm_free_chunks`. I have a slight preference but not a lot.

PR #4480, opened 2019-02-13
WT-4517 Change WT data format to avoid writing stable timestamps

keithbostic:


src/include/cell.i
```diff
@@ -58,7 +58,10 @@
  * Bit 3 marks an 8B packed, uint64_t value following the cell description byte.
  * (A run-length counter or a record number for variable-length column store.)
  *
- * Bit 4 is unused.
+ * Bit 4 marks a value with durable timestamps. For a start, stop value pair,
+ * it's the same as WT_TS_NONE and WT_TS_MAX. For an oldest_start, newest_start,
+ * newest_stop address triplet, it's the same as WT_TS_NONE, WT_TS_NONE and
+ * WT_TS_MAX.
```
michaelcahill: Did you consider flipping this, so bit 4 being on means that timestamps are present?  I don't feel strongly, that just seems to better match the existing format without timestamps.

src/include/cell.i
```diff
@@ -193,14 +190,42 @@ __wt_timestamp_value_check(
...
+	 *
+	 * TIMESTAMP-FIXME
+	 * Values (presumably) have associated transaction IDs, but we haven't
+	 * yet decided how to handle them. For now, ignore them in determining
+	 * value durability.
+	 */
+	if (*stop_tsp != WT_TS_MAX &&
+	    __wt_txn_visible_all(session, WT_TXN_NONE, *stop_tsp)) {
+		*start_tsp = WT_TS_NONE;
+		*stop_tsp = WT_TS_MAX;
```
michaelcahill: If this test were true, it would mean that the value was removed, and the remove is now stable.  In that case, unless the oldest timestamp has been updated while reconciliation is running, we shouldn't be storing the record at all (i.e., it should have been filtered out by `__rec_upd_select`).

For now, I think this check should be flipped to consider start timestamps for the common case where an insert becomes stable and isn't removed.

src/include/cell.i
```diff
@@ -256,7 +281,9 @@ __wt_cell_pack_addr(WT_SESSION_IMPL *session,
 {
 	uint8_t *p;
 
-	p = cell->__chunk + 1;
+	cell->__chunk[0] = '\0';
+	p = cell->__chunk;
```
michaelcahill: Maybe worth a comment: it isn't immediately clear to me why we were skipping a byte here before, and now we have to initialize it here.

src/include/cell.i
keithbostic: > Did you consider flipping this, so bit 4 being on means that timestamps are present? I don't feel strongly, that just seems to better match the existing format without timestamps.

I did not think of that (I added timestamps before thinking about durable timestamps being special). That approach makes things better because "short" values use 6 bits of the cell for the data length, which means the bit 0x08 isn't available to flag a special timestamp case. The way I did it, we can use short values if there are timestamps included, but we can't use short values if the timestamps are globally durable because we'll need bit 0x08 to flag the missing timestamps. In your approach, we get to use short values if there aren't timestamps included, and that should be the default case over time, that is, I expect more values without timestamps than with timestamps.

tl;dr
I didn't think of it, but I really like it.

keithbostic: This change worked out really nicely, it's much better this way. Pushed in 3c3b841.

src/include/cell.i
keithbostic: I don't completely follow this comment -- please feel free to update the code to match what you expect.

The test I'm trying to make here is if the specified timestamp pair is globally visible. There are paths through here that aren't removes, or even through `__rec_upd_select()`, and as I think you're pointing out, the transaction state may have changed since timestamps for this value were chosen. In other words, I'm thinking of this as "we chose a value to write and it has some timestamps, but before we write it, check to see if those timestamps are necessary based on the current state of the system".

src/include/cell.i
keithbostic: This is messy.

Previous versions of these functions set `p` to point past the cell descriptor byte, then used `p` to store data and other values that are associated with the cell descriptor byte, and somewhere in the top-level function did an explicit assignment of some value to `cell->__chunk[0]`. (That meant we didn't need to clear `cell->__chunk[0]`).

Current versions call underlying timestamp functions to optionally store timestamp information, in some cases setting a bit in the cell descriptor byte. That means we need to set `p` to the cell descriptor byte so those functions can set that bit, and we need to clear the cell descriptor byte because the function might not set the bit. When the underlying timestamp function returns, we set additional bits in the cell descriptor byte (so, instead of `cell->__chunk[0] = XXX`, it's `cell->__chunk[0] |= XXX`).

I tried a couple of different approaches, for example, clearing the byte in the underlying timestamp function. Anyway, everything I tried made the various cell functions diverge and look different -- value cells may or may not store timestamps, address cells always store timestamps, and key cells don't have timestamps at all.

So, I ended up with:
```
cell->__chunk[0] = '\0';
p = cell->__chunk[0];
```
as a function prefix everywhere, that is, always clear the cell descriptor byte and set `p` to point to it, even in cases where it's not necessary (for example, because address cells don't set a timestamp bit, we could pass the function that writes address timestamps a pointer past the cell descriptor byte). I thought optimizing those cases was unlikely to be useful (and the compiler is likely to catch those cases anyway), so I just gave every function the same prefix, including clearing the cell descriptor byte.

There's also an obvious question if there's a better overall structure: could we layout the cell without jumping around so much?

tl;dr If you have a suggestion of a better way to structure this code, or a suggestion about what that comment might look like, feel free to add it, or I can.

michaelcahill: In a055e22, I turned this into:
```
	/* Start building a cell: the descriptor byte starts zero. */
	p = cell->__chunk;
	*p = '\0';
```
That said, this was mostly for my benefit understanding the change: if you prefer it the other way, I won't fight.

src/include/cell.i
michaelcahill: I pushed the checks I expected in a055e22: please take a look and let me know if that doesn't make sense.

keithbostic:
@michaelcahill, your changes in a055e22 look good to me. I pushed a matching change in d224e2b94, if you agree with it, then please merge.

michaelcahill:

PR #4481, opened 2019-02-15
WT-4578 test/format should drop core even when deadlocked

keithbostic:


test/format/ops.c
```diff
@@ -64,6 +64,39 @@ modify_repl_init(void)
...
+{
+#ifndef _WIN32
+	struct itimerspec timer_val;
+	timer_t timer_id;
+
+	(void)signal(SIGALRM, on_alarm);
+
+	testutil_check(timer_create(CLOCK_REALTIME, NULL, &timer_id));
+	memset(&timer_val, 0, sizeof(timer_val));
+	timer_val.it_value.tv_sec = 60 * 2;
```
sueloverso: I think one minute or even 30 seconds is sufficient. Two minutes seems like overkill.

keithbostic: I agree, but I want to avoid any chance of a false positive, that is, I don't want any risk of debugging cases where the timer went off erroneously (and I didn't see any advantage in tightening down the timer). If you can think of a reason to tighten down the timer, let me know,

PR #4482, opened 2019-02-15
WT-4588 test/format should drop core even when Jenkins or connection terminates

keithbostic:


keithbostic:
autobuild: retest this please

test/format/t.c
```diff
@@ -38,6 +38,19 @@ static void usage(void)
 extern int __wt_optind;
 extern char *__wt_optarg;
 
+/*
+ * signal)handler --
```
agorrod: ) should be _

test/format/t.c
```diff
@@ -51,6 +64,16 @@ main(int argc, char *argv[])
 
 	(void)testutil_set_progname(argv);
 
+#ifdef SIGALRM
```
agorrod: These ifdefs are surprising enough that I think they warrant a comment. Something like "Windows and Linux support a different set of signals, so be conservative about installing handlers."

test/format/t.c
keithbostic: Thanks, fixed in e9f44ce1b.

test/format/t.c
keithbostic: Added in e9f44ce1b.

PR #4483, opened 2019-02-15
WT-4568 Fix operation tracking visualization.

fedorova:
Used to display callstacks incorrectly.

### DIFF
--- a/tools/optrack/find-latency-spikes.py
+++ b/tools/optrack/find-latency-spikes.py
@@ -786,7 +786,47 @@ def generateTSSlicesForBuckets():
 
     return bucketFilenames;
 
-def processFile(fname):
+#
+# After we have cleaned up the data by getting rid of incomplete function
+# call records (e.g., a function begin record is presend but a function end
+# is not or vice versa), we optionally dump this clean data into a file, so
+# it can be re-processed by other visualization tools. The output format is
+#
+# <0/1> <funcname> <timestamp>
+#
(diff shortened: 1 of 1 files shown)


tools/optrack/find-latency-spikes.py
```diff
@@ -280,13 +282,13 @@ def assignStackDepths(dataframe):
 
     for i in range(len(df.index)):
 
-        myStartTime = df.at[i, 'start'];
+        myEndTime = df.at[i, 'end'];
 
         # Pop all items off stack whose end time is earlier than my
```
luke-pearson: Comment here needs to be updated for end time

luke-pearson: Other than that this looks good to me

fedorova:
@luke-pearson, thank you for reviewing the branch. I pushed the comment update that you suggested.  Autobuild fails for mysterious reasons. The python build shows as failed, but if you zoom in, there are no problems identified. How do you suggest to proceed? May I merge the branch? Should I wait? Thanks!

fedorova:
@luke-pearson, forgot to mention, the evergreen build says that the patch must be manually authorized. Not sure what the protocol here is. Can I do this? What are the rules?

keithbostic:
autobuild: retest this please

lukech:
@fedorova you might have some evergreen credential issue there (if failing to trigger an evergreen build). Can you send me an email with screenshots of your problem?

PR #4484, opened 2019-02-17
WT-4413 Add optional compact progress messages.

nehakhatri5:


src/btree/bt_compact.c
```diff
@@ -103,21 +103,44 @@ __compact_rewrite_lock(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
...
+}
+
 /*
  * __wt_compact --
  *	Compact a file.
  */
 int
 __wt_compact(WT_SESSION_IMPL *session)
 {
+	struct timespec now, last;
```
sueloverso: These should be alphabetized.

src/btree/bt_compact.c
```diff
@@ -103,21 +103,44 @@ __compact_rewrite_lock(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
...
 	WT_BM *bm;
 	WT_DECL_RET;
 	WT_REF *ref;
+	uint64_t time_diff;
 	u_int i;
 	bool skip;
 
 	bm = S2BT(session)->bm;
 	ref = NULL;
+	__wt_epoch(session, &last);
```
sueloverso: Do we want to consider `wt_clock` here? Although we only print the progress messages once per second, we may be coming through here more frequently.

nehakhatri5: Yes, it's worth using `__wt_clock`.

src/btree/bt_compact.c
```diff
@@ -187,6 +209,14 @@ __wt_compact(WT_SESSION_IMPL *session)
 
 		session->compact_state = WT_COMPACT_SUCCESS;
 		WT_STAT_DATA_INCR(session, btree_compact_rewrite);
+
+		if (WT_VERBOSE_ISSET(session, WT_VERB_COMPACT_PROGRESS)) {
+			now = __wt_clock(session);
+			time_diff = WT_CLOCKDIFF_SEC(now, last);
+			if (time_diff >= 1)
+				__compact_progress(session, time_diff);
+			last = now;
```
agorrod: Shouldn't this only be updated when a message is actually output?

I think the timer variable should be moved to the connection - otherwise there might be 1000 calls to this function that cumulatively take a long time, but none of them are individually long enough to report a message.

I'd also prefer for all of this logic to be within the `__compact_progress` function - similar to the checkpoint progress message.

nehakhatri5: @agorrod I have added the timer into the `compact` state in the `session` structure and moved the logic within the `__compact_progress` function. Take a look and LGTM if looks OK.

src/btree/bt_compact.c
```diff
@@ -103,6 +103,34 @@ __compact_rewrite_lock(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
...
+__compact_progress(WT_SESSION_IMPL *session)
+{
+	WT_BM *bm;
+	uint64_t cur_time, time_diff;
+
+	bm = S2BT(session)->bm;
+	cur_time = __wt_clock(session);
+
+	time_diff = WT_CLOCKDIFF_SEC(cur_time, session->compact->last);
+	if (time_diff > 1) {
```
agorrod: I'd prefer this structure to be more similar to that used for checkpoint progress messages:
https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn_ckpt.c#L470

Including matching how often messages are generated

src/btree/bt_compact.c
```diff
@@ -187,6 +215,9 @@ __wt_compact(WT_SESSION_IMPL *session)
 
 		session->compact_state = WT_COMPACT_SUCCESS;
 		WT_STAT_DATA_INCR(session, btree_compact_rewrite);
+
+		if (WT_VERBOSE_ISSET(session, WT_VERB_COMPACT_PROGRESS))
```
agorrod: I'm inclined to move this check into the progress function

src/btree/bt_compact.c
```diff
@@ -187,6 +215,9 @@ __wt_compact(WT_SESSION_IMPL *session)
...
+			__compact_progress(session);
```
agorrod: I think this should move to the code block that contains `__wt_session_compact_check_timeout` - it's similar code and calling this progress function less often is appealing.

nehakhatri5:
@sueloverso I have updated the frequency of the progress message such that it would be called less often. So it is OK to switch back to using `__wt_epoch`. The logic is more consistent with the logic of checkpoint progress message now.

src/btree/bt_compact.c
```diff
@@ -103,6 +103,37 @@ __compact_rewrite_lock(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
...
+	uint64_t time_diff;
+
+	if (!WT_VERBOSE_ISSET(session, WT_VERB_COMPACT_PROGRESS))
+		return;
+
+	bm = S2BT(session)->bm;
+	__wt_epoch(session, &cur_time);
+
+	time_diff = WT_TIMEDIFF_SEC(cur_time, session->compact->begin);
+	if (time_diff / 20 > session->compact->prog_msg_count) {
```
sueloverso: I'd like a comment here because it isn't obvious what the connection between a time difference and message count is. Something like: "Wait 20 seconds before printing out any progress messages and then print out one message every 20 seconds."
It seems like the `20` should be a macro because that time might want to be adjusted.

src/include/misc.h
```diff
@@ -30,6 +30,8 @@
 
 #define	WT_MINUTE	(60)
 
+#define	WT_MSG_PERIOD	(20)
```
agorrod: This is too imprecise `WT_PROGRESS_MSG_PERIOD` would be better. You should also switch the checkpoint progress code to use the define.

src/session/session_compact.c
```diff
@@ -404,6 +404,7 @@ __wt_session_compact(
 	WT_ERR(__wt_config_gets(session, cfg, "timeout", &cval));
 	session->compact->max_time = (uint64_t)cval.val;
 	__wt_epoch(session, &session->compact->begin);
+	session->compact->prog_msg_count = 0;
```
agorrod: This isn't necessary - the structure is 0 initialized - so explicit initialization is redundant.

PR #4485, opened 2019-02-17
WT-4571 Improve transaction ID allocation

guddnjs91:
Acquiring a spinlock to allocate transacion ID can be a latent
bottleneck in WiredTiger. This attempts to remove using spinlock while
allocating transaction ID and retry to get valid ID from transaction
getting its own snapshot.


src/include/txn.i
```diff
@@ -965,15 +965,16 @@ __wt_txn_id_alloc(WT_SESSION_IMPL *session, bool publish)
 
 	/*
 	 * Allocating transaction IDs involves several steps.
+	 * (Note that this requires proof-of-concept now)
 	 *
-	 * Firstly, we do an atomic increment to allocate a unique ID.  The
-	 * field we increment is not used anywhere else.
+	 * Firstly, publish the current transaction ID to the global transaction
```
sulabhM: the comment needs updating, looks like you have moved from an invalid ID to a flag.

src/include/txn.i
```diff
@@ -1187,8 +1183,17 @@ __wt_txn_am_oldest(WT_SESSION_IMPL *session)
...
+			/*
+			 * If the transaction is still allocating its ID,
+			 * then we spin here until it gets its valid ID.
+			 */
+			if (s->is_allocating != false) {
+				WT_PAUSE();
+				goto retry;
+			}
 			return (false);
+		}
```
sulabhM: Can we restructure the code as follows:

```
while ((id = s->id) != WT_TXN_NONE && WT_TXNID_LT(id, txn->id)) {
    if (!s->is_allocating)
        return (false);
    WT_PAUSE();
}
```

src/include/txn.h
```diff
@@ -98,6 +98,7 @@ struct __wt_txn_state {
 	volatile uint64_t id;
 	volatile uint64_t pinned_id;
 	volatile uint64_t metadata_pinned;
+	volatile bool is_allocating;
```
guddnjs91: @michaelcahill, I added this variable in __wt_txn_state which is padded in cache line size for optimization.
I think it would trigger not a bit performance issue here because it would not extend the cache line size,
but I found that once the transaction id allocation is done, session->txn sets the flag WT_TXN_HAS_ID.
Do you think it is better to use that flag for indicating transaction is still allocating txn id instead of this additional boolean flag?

src/txn/txn.c
```diff
@@ -210,9 +210,23 @@ __wt_txn_get_snapshot(WT_SESSION_IMPL *session)
...
+			 * higher if the transaction is quickly done its job.
+			 * In this case, we ignore this transaction since
+			 * it would not be visible anyway in current snapshot.
+			 */
+			if (WT_TXNID_LT(current_id, id))
+				continue;
+			if (s->is_allocating != false) {
+				WT_PAUSE();
+				goto retry;
+			}
```
ZombieKim: This code can make an incorrect state if the execution order is
1. Thread A (txn allocating its txn id) set its `is_allocating` value as true and set its id with the `current` value. (up to __wt_txn_id_alloc:992)
2. Thread B (txn creating its snapshot) check if the Thread A's txn id is smaller than its own id. (up to __wt_txn_get_snapshot:225)
3. Thread A set its real txn id and turn its `is_allocating` flag as false. (up to __wt_txn_id_alloc:995)
4. Thread B sees the `is_allocating` flag as false (line 226) and then store the id as Thread A's txn id, which is different from the real B's ID.

I think the root cause of this problem is that the `is_allocating` flag and the id value are decoupled into different variables so any thread can see the middle state of it. IMO, it can be simpler to just steal 1 bit of the id variable and use it for the is_allocating flag so we can easily & atomically read the transaction's state. One thing I'm worried about is that this bit stealing can reduce the flexibility of the code so I think it's better to get confirm from another expert.

guddnjs91: To the best of my knowledge, I resolved this bug by checking `id == s->id` again after seeing `is_allocating` as false. If once fetched `id` is not the same with current `s->id`, then this `id` might not be a valid one and by iterating loop it will catch the valid one (or ignore if this transaction already completed its job or become higher). I'm not sure where to put the correct read barrier in this changed code yet and it requires discussion.

src/include/txn.i
michaelcahill: Actually, most of the change to this function can probably be reverted.  The `__wt_txn_am_oldest` function exists to abort the oldest transaction when activity is blocked (because everything is pinned in cache).  It should be able to ignore any transaction with `is_allocating = true`, though we may want a comment explaining why.  Something like:

```
/*
 * We are checking if this transaction is the oldest one in the system.
 * It is safe to ignore any sessions that are allocating transaction IDs:
 * since we already have an ID, they are guaranteed to be newer.
 */
if (!s->is_allocating && (id = s->id) != WT_TXN_NONE &&
    WT_TXNID_LT(id, txn->id)) {
        ...
```

src/include/txn.h
michaelcahill: I think it's better to have something in `WT_TXN_STATE` -- that is intended to be in a single cache line rather than having threads pull parts of `WT_TXN` into each CPU's cache.

I could go either way on whether to use a bit of `id` or a separate field.  The approach here (using a separate field) has the definite advantage that `id` always contains an ordinary transaction ID and no masking / special handling is required.

src/txn/txn.c
michaelcahill: I agree that a read barrier is required in between checking `is_allocating` and re-checking `id`.  Avoiding this complexity is definitely a point in favour of stealing a bit from the transaction ID.

guddnjs91: @michaelcahill, I also added another read barrier at line 225 to prevent reordering between reading `id` (at line 218) and reading `is_allocating` (at line 226) because this order should not be changed too.
@ZombieKim and I discussed and we're not sure if this kind of reordering could occur. Because we think `id` and `is_allocating` have no dependency logically so this might happen, though this two variables would be in the same cache line. Do you think this is necessary?

src/include/txn.h
guddnjs91: Thank you @michaelcahill . Now I understand that using a boolean flag in `WT_TXN_STATE` would be better than using the flag `WT_TXN_HAS_ID`

Related to this one, it is still not clear for me to decide between using `is_allocating` or stealing a bit from `id`. The advantage if we steal one bit of `id` is that the code (below) in `__wt_get_snapshot` and `__txn_oldest_scan` would be simpler. But if you think the comment explanation is enough, I personally think it is good to separate the flag instead of stealing one bit of `id` (since this might be used later)

src/include/txn.i
ZombieKim: I think we need to consider about simply ignoring the transaction id with `is_allocating = true` because it can miss the real oldest transaction id at that time. A possible history is
1. Thread A (real oldest) set the `is_allocating` flag as `true` and get its transaction id by calling `fetch_and_add(&current, 1)`. And then sleep.
2. Thread B (caller of `__wt_txn_am_oldest`) get its transaction id higher than Thread A, and then call `__wt_txn_am_oldest`. It sees the transaction id of Thread A with `is_allocating = true`, ignore it and decide itself as the oldest one.
3. Thread A begin to execute its transaction with the real oldest id.

To strictly check if this transaction is oldest, I think we need to wait for the transaction id with `is_allocating = true` unless the id value is greater than the current transaction id.
Can you please confirm this @michaelcahill ?

src/txn/txn.c
```diff
@@ -253,18 +279,47 @@ __txn_oldest_scan(WT_SESSION_IMPL *session,
...
+		 *    can happen if the transaction is quickly done its job.
+		 *    In this case, we ignore this transaction since it would
+		 *    not be visible anyway in current snapshot.
+		 */
+		while ((id = s->id) != WT_TXN_NONE &&
 		    WT_TXNID_LE(prev_oldest_id, id) &&
-		    WT_TXNID_LT(id, last_running))
-			last_running = id;
+		    WT_TXNID_LT(id, last_running) &&
+		    WT_TXNID_LT(id, current_id)) {
```
ZombieKim: I think this line (comparing `id` with `current_id`) doesn't necessary. `last_running` value starts from the `current_id`(line 282), and `last_running` value is monotonically decreased in this loop. So if the `id` value is less than `last_running` value (line 299), it must be less than `current_id` value.

src/include/txn.h
michaelcahill: Let's stick with a separate `bool`.

src/txn/txn.c
michaelcahill: I wouldn't make assumptions based on cache lines: WT runs on multiple CPU architectures.

At this point, assuming there is no performance regression on any of our workloads, I am happy with the clarity of having two read barriers.  I suspect we can rewrite this code to read `is_allocating` first, so it has a single read barrier, but I don't think that should hold up getting this change merged.

ZombieKim: > I suspect we can rewrite this code to read `is_allocating` first, so it has a single read barrier

I think this way can make another problem. If `is_allocating` value has read before setting to `true` and then `id` value has read as a temporal id value which could be different from real txn id value, snapshot can miss the real txn id.

src/include/txn.i
michaelcahill: @ZombieKim, I think this situation is okay: even if `__wt_txn_am_oldest` returns false in this case, thread B will call it again if eviction is really stuck, and the second time it won't race with any concurrent ID allocation, so it will return `true`.  I think that is preferable to slowing down the common case by waiting for allocating threads.

PR #4487, opened 2019-02-19
WT-4580 Make transactions with ignore_prepare=true readonly.

michaelcahill:


agorrod:
Closing for now, since this is blocked on some MongoDB server changes. Will resurrect when those are finalized.

luke-pearson:
Re-opening this PR as the related server work is now complete.

luke-pearson:
retest this please

src/include/txn.i
```diff
@@ -511,9 +511,14 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 
 	txn = &session->txn;
 
-	if (F_ISSET(txn, WT_TXN_READONLY))
+	if (F_ISSET(txn, WT_TXN_READONLY)) {
+		if (F_ISSET(txn, WT_TXN_IGNORE_PREPARE))
```
louiswilliams: @luke-pearson It looks like this returns  ENOTSUP even when `ignore_prepare=force`. Is there more work needed to suppress the error in that case? I applied this locally with the server changes, but I still get failures.

src/txn/txn.c
```diff
@@ -529,9 +529,14 @@ __wt_txn_config(WT_SESSION_IMPL *session, const char *cfg[])
...
-	    (cval.len > 0 && WT_STRING_MATCH("force", cval.str, cval.len)))
-		F_SET(txn, WT_TXN_IGNORE_PREPARE);
+	if (cval.val)
+	{
+		if (cval.len > 0 &&
+		    WT_STRING_MATCH("force", cval.str, cval.len))
+			F_SET(txn, WT_TXN_IGNORE_PREPARE);
+		else
+			F_SET(txn, WT_TXN_IGNORE_PREPARE | WT_TXN_READONLY);
+	}
```
michaelcahill: Suggest losing the top `if (cval.val)` and replacing with `if (cval.len > 0 && WT_STRING_MATCH("force", cval.str, cval.len) ... else if (cval.val) ...`

src/include/txn.i
luke-pearson: I've put in a fix for the issue, could you please retest?

louiswilliams: This does work now, thanks.

PR #4491, opened 2019-02-21
WT-4324 Re-instate lookaside and checkpoint collision function/test.

sueloverso:
I have a bunch of you listed as reviewers. This change reverts part of WT-4094 that removed the check for lookaside and checkpoint collisions. It retains the other portion of the change for the `las_skew` and the transaction id. @nehakhatri5 can you check if this change still satisfies the YCSB issue in the WT-4094 ticket?

### DIFF
--- a/src/reconcile/rec_write.c
+++ b/src/reconcile/rec_write.c
@@ -571,6 +571,42 @@ __wt_reconcile(WT_SESSION_IMPL *session, WT_REF *ref,
 }
 
 /*
+ * __rec_las_checkpoint_test --
+ *	Return if the lookaside table is going to collide with a checkpoint.
+ */
+static inline bool
+__rec_las_checkpoint_test(WT_SESSION_IMPL *session, WT_RECONCILE *r)
+{
+	WT_BTREE *btree;
+
+	btree = S2BT(session);
+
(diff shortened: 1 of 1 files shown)


michaelcahill:
@sueloverso, thanks: nice catch!

I'm not sure if anyone has had a chance to look at this yet, but it would be good if someone can write up what is actually going wrong.  I'm concerned about reinstating this check because it will cause all eviction to give up in some cases where we are already constrained (i.e., under pressure, when all eviction is lookaside eviction, everything will stop during checkpoints).  Of course we need checkpoints to be correct when there is lookaside eviction, but I'd like to make sure we're only interrupting eviction when it is absolutely necessary.

The comment says "the alternative is to have checkpoint drain lookaside table reconciliations" -- I thought this check was safe to remove because we now do exactly that with the call to `__wt_gen_next_drain(session, WT_GEN_EVICT)` in `__wt_sync_file`.

Anyway, I hope I'll get a chance to dig into this early next week, but if anyone understands what is actually going wrong and why this check fixes it, it would be great to see that written up.

agorrod:
I've closed this - since we took another solution see WT-4324 for details.

PR #4494, opened 2019-02-26
WT-4573 Reducing calls to __wt_epoch from session reset.

guddnjs91:
__wt_epoch is called everytime in __wt_session_cursor_cache_sweep
when session reset is called. This change is to set countdown to
call less often cursor cache sweep during session reset.

### DIFF
--- a/src/include/session.h
+++ b/src/include/session.h
@@ -40,7 +40,7 @@ struct __wt_hazard {
 typedef TAILQ_HEAD(__wt_cursor_list, __wt_cursor)	WT_CURSOR_LIST;
 
 /* Number of cursors cached to trigger cursor sweep. */
-#define	WT_SESSION_CURSOR_SWEEP_COUNTDOWN	20
+#define	WT_SESSION_CURSOR_SWEEP_COUNTDOWN	40
 
 /* Minimum number of buckets to visit during cursor sweep. */
 #define	WT_SESSION_CURSOR_SWEEP_MIN		5
--- a/src/session/session_api.c
+++ b/src/session/session_api.c
@@ -1016,7 +1016,11 @@ __session_reset(WT_SESSION *wt_session)
 
 	WT_TRET(__wt_session_reset_cursors(session, true));
 
-	WT_TRET(__wt_session_cursor_cache_sweep(session));
+	if (--session->cursor_sweep_countdown == 0) {
+		session->cursor_sweep_countdown =
+		    WT_SESSION_CURSOR_SWEEP_COUNTDOWN;
+		WT_TRET(__wt_session_cursor_cache_sweep(session));
+	}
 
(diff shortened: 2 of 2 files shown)


sulabhM:
@ddanderson, can you think of any negative impact of not calling cursor_cache_sweep each time we reset a session? Is there any MongoDB test that you can suggest that would help us here ?

src/include/session.h
```diff
@@ -40,7 +40,7 @@ struct __wt_hazard {
 typedef TAILQ_HEAD(__wt_cursor_list, __wt_cursor)	WT_CURSOR_LIST;
 
 /* Number of cursors cached to trigger cursor sweep. */
-#define	WT_SESSION_CURSOR_SWEEP_COUNTDOWN	20
+#define	WT_SESSION_CURSOR_SWEEP_COUNTDOWN	40
```
nehakhatri5: Why is it increased to 40?

guddnjs91: Currently, there are 2 places to trigger the cursor cache sweep(`__wt_session_cursor_cache_sweep`): `__curfile_close` and `__session_reset`. Before the change, the countdown variable (`session->cursor_sweep_countdown`) is only used from `__curfile_close` and not from `__session_reset`. That's why `__wt_epoch` was called every time we reset the session.  
Now in this change, because we use the same countdown variable to trigger less cursor cache sweep from `__session_reset` just like `__curfile_close`, I doubled the initial variable 20 to 40.

ddanderson:
@sulabhM, I'm not aware of any standard tests.  When I was doing cursor testing, I adapted a mongod test based on one I got from Michael, with a bunch of collections and random drops. I'd be happy to send it.  You'd probably need to fiddle with parameters to make it more interesting for this use.

Some general comments. I've convinced myself that an idle session cannot stay idle for too long (it is eventually closed), and probably has some kind of activity: a steady diet of either session reset or cursors released to the cache. And at the end of every session, all its cursors are closed (for the default way of using the cursor cache in MongoDB).  So with maybe 5 cursors in a typical session (it would include indices), we might hit the countdown after 8 sessions are used (40==5*8), instead of every 1 before this change.  So the ratio might be 8:1, of course could be higher if fewer cursors used on average.

In the worst case, when there are lots of cached cursors that should be closed (that would happen in a system with lots of collection drops), there would be a buildup of closable cursors in the session's cache.  With a build up of more closable cursors, more sweeping is done.  At maximum up to 32 buckets of 512 are visited and cleaned up for each call to sweep.  So 16 calls to sweep cleans up everything. If there is not such a buildup of closable cursors, fewer buckets are visited each time, it would take more calls to sweep everything, but then, we're not so worried about wasted memory in that case (since there are fewer closable cursors).

So that's the theory, but it would still be good to run some tests.

PR #4498, opened 2019-02-28
WT-4606 workgen: remove lsm from the default table type in wtperf emulation

ddanderson:


ddanderson:
FWIW I reran some of the tests I did for WT-4447, and saw big differences.  I think adjustments to the parameters of the test need to be made: I don't see any stalls for the 0 case, so really our load is too low. Both the 30M and 35M bring ops/sec way down, so those numbers may be too low as well. 1G does make improvements over the 0 case in latency, as expected.  I'm not sure we need to explore that any more, it's well covered by the actual wtperf runs.

ddanderson:
@sueloverso, can you please take a look at this small change?  Thanks so much for catching the error!

PR #4499, opened 2019-03-01
WT-4607 Allow prepare timestamp to commit behind oldest timestamp.

bvpvamsikrishna:
Function name changes for less ambiguity.
Mask the check of commit timestamp < oldest timestamp.

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -579,14 +579,15 @@ set:	__wt_writelock(session, &txn_global->rwlock);
 }
 
 /*
- * __wt_timestamp_validate --
- *	Validate a timestamp to be not older than the global oldest and global
- *	stable and running transaction commit timestamp and running transaction
- *	prepare timestamp.
+ * __wt_commit_timestamp_validate --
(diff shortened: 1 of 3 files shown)


src/txn/txn_timestamp.c
```diff
@@ -579,14 +579,15 @@ set:	__wt_writelock(session, &txn_global->rwlock);
...
+ * __wt_commit_timestamp_validate --
+ *	Validate a timestamp to be not older than running transaction commit
+ *	timestamp and running transaction prepare timestamp. Validate a durable
+ *	timestamp to be not older than the global oldest and global stable
+ *	timestamp.
  */
 int
-__wt_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
-    wt_timestamp_t ts, WT_CONFIG_ITEM *cval, bool compare_stable)
+__wt_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
```
agorrod: While you are changing the name, a prefix of `__wt_txn_XXX` is more consistent with other functions in the file.

PR #4500, opened 2019-03-01
WT-4630 Add Clang Format script

tetsuo-cpp:
This PR contains an initial attempt at a `.clang-format` file for the WiredTiger formatting style as well as a script in `dist/` to format everything under `src/`.

The `s_clang-format.list` is for specifying files that we don't want to be formatted (mostly files that get auto generated by other scripts within `dist/`).

I created the `.clang-format` by dumping the full set of LLVM options and reviewing the values one by one.
`clang-format -style=llvm -dump-config > .clang-format`


tetsuo-cpp:
retest this please

keithbostic:
This is great!!

tetsuo-cpp:
@bvpvamsikrishna @keithbostic 
What files do you think belong in the `s_clang-format.list` file?
I've currently added files that contain generated code (by scripts under `dist`) however I may have missed some. Also I'm thinking maybe that CityHash file belongs on this list too.

keithbostic:
> What files do you think belong in the s_clang-format.list file?
I've currently added files that contain generated code (by scripts under dist/) however I may have missed some. Also I'm thinking maybe that CityHash file belongs on this list too.

I agree with adding CityHash (we don't maintain it, so I'm hesitant to mess around with it, the closer it is to the originally imported version, the better).

I disagree with adding auto-generated files to this list. I'd go the other way and simplify the auto-generation process (just dump out the code instead trying to make it look pretty, checking line lengths and so on), and let clang do the formatting for us. We do still have to read those files sometimes, they should look familiar.

tetsuo-cpp:
> I disagree with adding auto-generated files to this list. I'd go the other way and simplify the auto-generation process (just dump out the code instead trying to make it look pretty, checking line lengths and so on), and let clang do the formatting for us. We do still have to read those files sometimes, they should look familiar.

18a185c
Understood! I've removed those files from the list and added the two imported files containing hashing code.

lukech:
@tetsuo-cpp Please remember to create a new WT ticket (and update PR title) as WT-4602 had been resolved. There's a 'one ticket one commit' policy in the team.

tetsuo-cpp:
> @tetsuo-cpp Please remember to create a new WT ticket (and update PR title) as WT-4602 had been resolved. There's a 'one ticket one commit' policy in the team.

Thanks @lukech. I've made a new ticket and changed the PR title.

bvpvamsikrishna:
@keithbostic, would you mind taking a look at this. If you are good with this PR, I will merge it. Thank you.

tetsuo-cpp:
@bvpvamsikrishna @keithbostic 
Just FYI, I was contacted by a developer from server a little while ago who told me that pinning to a specific Clang Format version is important since they sometimes format differently even for the same settings (this is what prompted me to add the version check in my script). He also linked me to a Python Clang Format script that they use in server which has functionality for downloading their "agreed" version for either Linux or Darwin.

Since we're

dist/s_clang-format
```diff
@@ -0,0 +1,38 @@
+#! /bin/sh
+
+# Installation of the clang development package isn't standard, list a
+# couple of the places we're using.
+export PATH=$PATH:/usr/local/clang50/bin:/usr/local/llvm-devel/bin
```
keithbostic: Why is this `clang50`, when we required Clang 6.0?

dist/s_clang-format
```diff
@@ -0,0 +1,38 @@
...
+while test "$p" != "/" ; do
+	if test -d "$p/build_posix"; then
+		break;
+	fi
+	p=`dirname $p`
+done
+test "$p" != "/" || {
+	echo "$0: cannot find the WiredTiger top-level directory"
+	exit 1
+}
```
keithbostic: Maybe `git rev-parse --show-cdup`.

dist/s_clang-format.list
```diff
@@ -0,0 +1,2 @@
+src/support/hash_city.c
+src/support/hash_fnv.c
```
keithbostic: The file is missing a terminating `newline`.

dist/s_clang-format
tetsuo-cpp: 8493ae7
Neat. Thanks for that.

dist/s_clang-format.list
tetsuo-cpp: 8493ae7
Fixed. Also added a few more third party files that I saw listed here:
https://source.wiredtiger.com/3.1.0/license.html

dist/s_clang-format
tetsuo-cpp: 8493ae7
Fixed.

tetsuo-cpp:
@keithbostic 
> What's the plan on enforcing this change in the tree?

Sorry, I ought to have left a comment about this earlier. My plan is to leave the the Clang Format script and settings file in the repo for a while and let others experiment with it. If everyone is happy with the usability of the script and the impact it's going to have on the source tree, then we can find a time when we don't have too many open PRs/branches and apply to `develop`.

Just FYI, Mongo has a more complex script that actually pulls in a Clang Format binary for the right platform (Linux, Darwin, etc), caches it and does a bunch of other stuff. I think that since the team is smaller and our development setups are more homogenous, we can probably get away with just having a version check and call it a day. But if someone in the team is running an unusual distro and is having difficulties getting that version, we have the option of repurposing that script for WiredTiger before applying the changes.

https://github.com/mongodb/mongo/blob/master/buildscripts/clang_format.py

> Here's a case statement: comments aren't aligned

`AlignTrailingComments` is already switched on but it only works for consecutive lines. I think it makes sense in the general case since the case statements could be far apart and you wouldn't expect those to be aligned. But yeah, I agree it is quite ugly.

> a `#define` that includes `case` is incorrectly placed

This one is a bit troublesome. I'll have to have a play around with that macro to see if there's a way to get around this.

> Multiple indents where it's not appropriate

I don't believe there's a setting to toggle this behaviour. The reason it indented like this is because the line break happened in the middle of a binary operator (rather than at the end of an argument) so it indents to 4 spaces past the beginning of that argument since it's still "in" that expression.

> Breaking before `?:` operators

Fixed in 8493ae7.

keithbostic:
@tetsuo-cpp, I suspect we should do a fairly serious review and see what the problems are going to be before we let people play with it? I can take a first run at that, if you like, try and identify things worth fixing?

>> a `#define` that includes `case` is incorrectly placed
> This one is a bit troublesome. I'll have to have a play around with that macro to see if there's a way to get around this.

I think we should make the changes we have to make to get a clang format working for the team without too much special-casing. If we have to change that macro to not include the `default` keyword, so be it.

>> Multiple indents where it's not appropriate
> I don't believe there's a setting to toggle this behaviour. 

Ditto this change. If that's the price we pay, so be it. Making every new engineer struggle with the current format is wasted effort, IMHO.

And, ideally, we can make this happen and get the adjusted line lengths we want at the same time.

The problem, of course, is that we can't diff anything after those changes, which is a little scary.

@agorrod, @bvpvamsikrishna, what's your position on all of this?

agorrod:
@keithbostic I'm a fan as long as we can find a format everyone is prepared to live with. Once this gets merged in, I think we should start an evaluation process with everyone on the team who is interested in contributing.

I'm keen enough for this auto-checking that I'd be OK with supplementing a standard tool with some custom scripts if we can't find a tool that does what we want out of the box.

keithbostic:
@agorrod,
> I'm a fan as long as we can find a format everyone is prepared to live with.

Let's commit to that -- I don't want to fail at this and there's got to be something we can all live with. I'm worried without some pressure we're going to let this languish, but if we're committed to merge in, say, a month, people will take a look.

keithbostic:
I'd also like to avoid any custom scripts: from the MongoDB server team:
>We experimented with customizing ClangFormat to match our existing rules. The ClangFormat code was easy to work with, but ultimately, the ongoing cost of maintaining a fork is not worth the effort to match our old format. Instead, we adapted our existing rules to match what ClangFormat was already capable of. We strongly recommend that you do the same.

https://engineering.mongodb.com/post/succeeding-with-clangformat-part-2-the-big-reformat

dist/s_clang-format
```diff
@@ -0,0 +1,27 @@
...
+echo "$0: running clang-format"
+
+# Get all source files that aren't in s_clang-format.list.
+search=`find src -name '*.[chi]'`
+for f in `cat dist/s_clang-format.list`; do
+	search=`echo "$search" | sed "\#$f#d"`
+done
+
+# Format each file inplace.
+clang-format -i $search
```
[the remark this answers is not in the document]
keithbostic: Would it be reasonable to allow running the script on a single file? Should at least help with experimentation:
```
-echo "$0: running clang-format"
-
-# Get all source files that aren't in s_clang-format.list.
-search=`find src -name '*.[chi]'`
-for f in `cat dist/s_clang-format.list`; do
-       search=`echo "$search" | sed "\#$f#d"`
-done
+case $# in
+0)
+       # Get all source files that aren't in s_clang-format.list.
+       search=`find src -name '*.[chi]'`
+       for f in `cat dist/s_clang-format.list`; do
+               search=`echo "$search" | sed "\#$f#d"`
+       done;;
+1)
+       search="$1";;
+*)
+       echo "usage: $0 [files]"
+       exit 1;;
+esac
```

keithbostic:
@tetsuo-cpp, another bad one. If you run on `src/btree/bt_io.c`, around line 120 format seriously messed up an else statement. The `if` clause is 8-indented, the `else` clause is 4-space indented.
```
        } else
            /*
             * If we uncompressed above, the page is in the correct buffer.
             * If we get here the data may be in the wrong buffer and the
             * buffer may be the wrong size.  If needed, get the page
             * into the destination buffer.
             */
            if (ip != NULL)
                WT_ERR(__wt_buf_set(session, buf, ip->data, dsk->mem_size));
```

keithbostic:
It would be nice if it didn't indent labels:
```
        if (0) {
-corrupt:       if (ret == 0)
+       corrupt:
+               if (ret == 0)
                        ret = WT_ERROR;
```

keithbostic:
It seems like sometimes it fills aggressively and sometimes it doesn't:
```
        /* Checkpoint calls are different than standard calls. */
-       WT_ASSERT(session,
-           (!checkpoint && addr != NULL && addr_sizep != NULL) ||
-           (checkpoint && addr == NULL && addr_sizep == NULL));
+       WT_ASSERT(
+           session, (!checkpoint && addr != NULL && addr_sizep != NULL) ||
+                        (checkpoint && addr == NULL && addr_sizep == NULL));
```
Is there a way to configure for consistency, always line-fill aggressively?

keithbostic:
Unexpected space after `TAILQ_FOREACH`:
```
@@ -783,7 +787,7 @@ __wt_cursor_cache_get(WT_SESSION_IMPL *session, const char *
uri,    
         * cursor that matches uri and configuration, use it.
         */     
        bucket = hash_value % WT_HASH_ARRAY_SIZE;
-       TAILQ_FOREACH(cursor, &session->cursor_cache[bucket], q) {
+       TAILQ_FOREACH (cursor, &session->cursor_cache[bucket], q) {
```

keithbostic:
Unexpected indentation when a function call including an assignment was broken:
```
@@ -956,7 +961,7 @@ __wt_cursor_reconfigure(WT_CURSOR *cursor, const char *confi
g)  
         */
        if (WT_CURSOR_RECNO(cursor)) {
                if ((ret = __wt_config_getones(
-                   session, config, "append", &cval)) == 0) {
+                        session, config, "append", &cval)) == 0) {
                        if (cval.val)
                                F_SET(cursor, WT_CURSTD_APPEND);
                        else
```

keithbostic:
Exceeded line length:
```
@@ -968,8 +973,8 @@ __wt_cursor_reconfigure(WT_CURSOR *cursor, const char *confi
g)
        /*       
         * overwrite
         */      
-       if ((ret = __wt_config_getones(
-           session, config, "overwrite", &cval)) == 0) {
+       if ((ret = __wt_config_getones(session, config, "overwrite", &cval)) ==
+           0) {
```

dist/s_clang-format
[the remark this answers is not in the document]
tetsuo-cpp: 561202e
Thanks!

tetsuo-cpp:
@keithbostic 
Unfortunately I don't have satisfactory answers to most of these.
> It would be nice if it didn't indent labels

Looks like this guy did some work but it never got merged in: http://lists.llvm.org/pipermail/cfe-dev/2015-September/045014.html.

Surprising that it does this by default and isn't configurable since it implies that the labels have a scope. I can try getting this change into Clang now but it won't be usable to us anytime soon.

> Unexpected space after `TAILQ_FOREACH`

This is because I've registered that as one of the `ForEachMacros` in our settings file. So it interprets that macro as a type of control flow (and places a space after it in the same way you would do with a keyword like `for`). This is necessary otherwise the braces/loop body doesn't get indented correctly.

> Unexpected indentation when a function call including an assignment was broken

This is a similar thing to the line you pointed out earlier where we get a line break within an argument and Clang Format indented from the beginning of that argument. It seems that if you have any nested expressions that span multiple lines, they get indented starting from their "parent" in the syntax tree.

This line from `conn_cache_pool.c` shows it more clearly. Since the `cval.val != 0` part of that `if` statement is the right hand side of the top level `&&` operator it gets outdented. Whereas the `0` on the second line gets indented since it's the part of the inner `==` operator that is nested within the `&&`.

```
        if (__wt_config_gets(session, &cfg[1], "shared_cache.reserve", &cval) ==
                0 &&
            cval.val != 0)
                reserve = (uint64_t)cval.val;
```

> Exceeded line length

Are you using Clang Format 6? I count 79 chars and `s_longlines` doesn't complain.

tetsuo-cpp:
Another thing to be aware of is that certain more "creative" macros will get butchered by Clang Format. Example from `btmem.h`.


```
#define WT_LAS_CONFIG                                                                   \
        "key_format=" WT_UNCHECKED_STRING(QIQu) ",value_"                               \
                                                "format=" WT_UNCHECKED_STRING(          \
                                                    QQQBBu) ",block_"                   \
                                                            "compressor"                \
                                                            "=" WT_LOOKASIDE_COMPRESSOR \
                                                            ",leaf_value_max="          \
                                                            "64MB"                      \
                                                            ",prefix_"                  \
                                                            "compression=true"
```
We'll likely need to turn off the formatter for these (can be done with a comment similar to the linter).

tetsuo-cpp:
retest this please

PR #4501, opened 2019-03-01
WT-4602 Clang Tidy and Clang Scan fixes

tetsuo-cpp:
This PR is to fix or suppress any warnings from Clang Tidy and Clang Scan.

### DIFF
--- a/src/btree/bt_huffman.c
+++ b/src/btree/bt_huffman.c
@@ -336,6 +336,7 @@ __wt_huffman_read(WT_SESSION_IMPL *session, WT_CONFIG_ITEM *ip,
 		WT_ERR(__wt_getline(session, fs, tmp));
 		if (tmp->size == 0)
 			break;
+		/* NOLINTNEXTLINE(cert-err34-c) */
 		n = sscanf(
 		    tmp->data, "%" SCNi64 " %" SCNi64, &symbol, &frequency);
 		/*
--- a/src/conn/conn_capacity.c
+++ b/src/conn/conn_capacity.c
@@ -306,6 +306,7 @@ __wt_capacity_throttle(WT_SESSION_IMPL *session, uint64_t bytes,
 
 	conn = S2C(session);
 	cap = &conn->capacity;
+	/* NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores) */
 	capacity = steal_capacity = 0;
 	reservation = steal = NULL;
 	switch (type) {
--- a/src/include/gcc.h
+++ b/src/include/gcc.h
@@ -11,6 +11,7 @@
 
 /* GCC-specific attributes. */
 #define	WT_PACKED_STRUCT_BEGIN(name)					\
+	/* NOLINTNEXTLINE(misc-macro-parentheses) */	\
 	struct __attribute__ ((__packed__)) name {
 #define	WT_PACKED_STRUCT_END						\
 	};
--- a/src/include/swap.h
+++ b/src/include/swap.h
@@ -47,6 +47,7 @@ static inline uint64_t
 __wt_bswap64(uint64_t v)
 {
 	return (
+	    /* NOLINTNEXTLINE(misc-redundant-expression) */
 	    ((v << 56) & 0xff00000000000000UL) |
 	    ((v << 40) & 0x00ff000000000000UL) |
 	    ((v << 24) & 0x0000ff0000000000UL) |
--- a/src/os_posix/os_fs.c
+++ b/src/os_posix/os_fs.c
@@ -140,7 +140,7 @@ __posix_directory_sync(WT_SESSION_IMPL *session, const char *path)
 
 	fd = 0;				/* -Wconditional-uninitialized */
 	WT_SYSCALL_RETRY((
-	    (fd = open(dir, O_RDONLY, 0444)) == -1 ? -1 : 0), ret);
+	    (fd = open(dir, O_RDONLY | O_CLOEXEC, 0444)) == -1 ? -1 : 0), ret);
 	if (ret != 0)
 		WT_ERR_MSG(session, ret, "%s: directory-sync: open", dir);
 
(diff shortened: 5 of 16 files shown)


src/btree/bt_huffman.c
```diff
@@ -336,6 +336,7 @@ __wt_huffman_read(WT_SESSION_IMPL *session, WT_CONFIG_ITEM *ip,
 		WT_ERR(__wt_getline(session, fs, tmp));
 		if (tmp->size == 0)
 			break;
+		/* NOLINTNEXTLINE(cert-err34-c) */
```
tetsuo-cpp: Clang Tidy doesn't like the use of sscanf to convert strings to integers as it supposedly won't report any conversion errors (it suggests to use strtoul instead).

Some of the sscanf usages seemed a bit complex so I wasn't confident replacing them without understanding the codebase better. So I've decided to just suppress the linter warnings for the time being.

src/conn/conn_capacity.c
```diff
@@ -306,6 +306,7 @@ __wt_capacity_throttle(WT_SESSION_IMPL *session, uint64_t bytes,
 
 	conn = S2C(session);
 	cap = &conn->capacity;
+	/* NOLINTNEXTLINE(clang-analyzer-deadcode.DeadStores) */
```
tetsuo-cpp: Clang Scan doesn't like that we're assigning capacity to 0 and then reassigning it immediately in the switch statement below. I tried removing the assignment but my compiler complains that we might be using capacity when it's uninitialised.

It seems that we can't win so I've just suppressed the linter here.

src/include/gcc.h
```diff
@@ -11,6 +11,7 @@
 
 /* GCC-specific attributes. */
 #define	WT_PACKED_STRUCT_BEGIN(name)					\
+	/* NOLINTNEXTLINE(misc-macro-parentheses) */	\
 	struct __attribute__ ((__packed__)) name {
```
tetsuo-cpp: Clang Tidy is complaining that ``name`` isn't wrapped in brackets (usually to avoid operator precedence issues) even though that's not allowed in a struct definition.

I'm assuming that this is an unusual use of preprocessor macros that they haven't considered so I've suppressed the linter here.

src/include/swap.h
```diff
@@ -47,6 +47,7 @@ static inline uint64_t
 __wt_bswap64(uint64_t v)
 {
 	return (
+	    /* NOLINTNEXTLINE(misc-redundant-expression) */
 	    ((v << 56) & 0xff00000000000000UL) |
```
tetsuo-cpp: Clang Tidy seems to think that this line doesn't do anything. The code looks sane to me so I've just suppressed the linter here.

src/txn/txn_timestamp.c
```diff
@@ -41,7 +41,8 @@ __wt_timestamp_to_hex_string(wt_timestamp_t ts, char *hex_timestamp)
 		return;
 	}
 	if (ts == WT_TS_MAX) {
-		(void)strcpy(hex_timestamp, "ffffffffffffffff");
+		(void)strncpy(
```
tetsuo-cpp: Clang Tidy doesn't like `strcpy`. It suggests `strlcpy` but that appears to be a BSD thing so I've gone for `strncpy` instead and it seems happy with that.

src/os_posix/os_fs.c
```diff
@@ -140,7 +140,7 @@ __posix_directory_sync(WT_SESSION_IMPL *session, const char *path)
 
 	fd = 0;				/* -Wconditional-uninitialized */
 	WT_SYSCALL_RETRY((
-	    (fd = open(dir, O_RDONLY, 0444)) == -1 ? -1 : 0), ret);
+	    (fd = open(dir, O_RDONLY | O_CLOEXEC, 0444)) == -1 ? -1 : 0), ret);
```
tetsuo-cpp: Clang Tidy recommends to use this flag for calls to `open`. To my understanding, it just causes any child processes created via `fork` to not have access to any open file descriptors from the parent. I don't see a `fork` or `exec` call between here and where we call `close` so I think this change should be safe.

src/support/hash_city.c
```diff
@@ -306,10 +306,10 @@ static inline uint64_t CityHash64(const char *s, size_t len) {
 	if (len <= 32) {
```
tetsuo-cpp: Is it ok to edit this file?

It appears to be based of Google's CityHash. It already has some WiredTiger functions in it so I figured that it's ok.

keithbostic: I'd toss in a comment just so the next person knows there's no real change here (or just turn off the linter for these lines, this isn't really "our" code).

src/txn/txn_timestamp.c
keithbostic: Static analyzers tend not to like `strncpy` because it doesn't always nul-terminate. Maybe `memcpy`?

```
--- txn_timestamp.c     2019-03-04 10:45:49.398128000 -0500
+++ -   2019-03-04 10:56:11.712206653 -0500
@@ -41,8 +41,9 @@
                return;
        }
        if (ts == WT_TS_MAX) {
-               (void)strncpy(
-                   hex_timestamp, "ffffffffffffffff", WT_TS_HEX_STRING_SIZE);
+#define        WT_TS_MAX_HEX_STRING    "ffffffffffffffff"
+               (void)memcpy(hex_timestamp,
+                   WT_TS_MAX_HEX_STRING, strlen(WT_TS_MAX_HEX_STRING) + 1));
                return;
        }
```
I don't feel strongly at all, just tossing that idea out if you like it.

tetsuo-cpp: faca49d
This seems like a better idea. I've switched over to `memcpy`.

src/support/hash_city.c
tetsuo-cpp: 772866d
I just decided to turn off the linter here. I can imagine that if we keep making code changes to this file it could become confusing when comparing to the original source.

tetsuo-cpp:
retest this please

lukech:
retest this please

PR #4502, opened 2019-03-01
WT-4615 Sync backup file before closing.

sueloverso:
Please review this change. It is from a SERVER ticket where a user did a hardware snapshot backup and the `WiredTiger.backup` file was zero length. We definitely were not syncing the backup file before returning from opening the backup cursor.


src/cursor/cur_backup.c
```diff
@@ -359,18 +359,18 @@ __backup_start(WT_SESSION_IMPL *session,
...
-		WT_TRET(__wt_fs_rename(session, WT_BACKUP_TMP, dest, false));
+		WT_TRET(__wt_sync_and_rename(session,
+		    &cb->bfs, WT_BACKUP_TMP, dest));
 		__wt_writelock(session, &conn->hot_backup_lock);
 		conn->hot_backup_list = cb->list;
 		__wt_writeunlock(session, &conn->hot_backup_lock);
 		F_SET(session, WT_SESSION_BACKUP_CURSOR);
-	}
+	} else if (cb->bfs != NULL)
+		WT_TRET(__wt_fclose(session, &cb->bfs));
```
keithbostic: I would do this as:
```
--- cur_backup.c        2019-03-04 10:30:08.246209000 -0500
+++ -   2019-03-04 10:35:11.202636339 -0500
@@ -369,7 +369,8 @@
                conn->hot_backup_list = cb->list;
                __wt_writeunlock(session, &conn->hot_backup_lock);
                F_SET(session, WT_SESSION_BACKUP_CURSOR);
-       } else if (cb->bfs != NULL)
+       }
+       if (cb->bfs != NULL)
                WT_TRET(__wt_fclose(session, &cb->bfs));

 done:
```
That way we won't depend on any particular implementation of `__wt_sync_and_rename()`, if it returns without closing the `cb.bfs` stream, we won't have a bug.

keithbostic: I'm also wondering about the fact that we ignore failures from `__wt_sync_and_rename()` for the purpose of setting `WT_CONNECTION.hot_backup_list`? I mean, why isn't it something like this:
```
err:    /* Close the hot backup file. */
        if (srcfs != NULL)
                WT_TRET(__wt_fclose(session, &srcfs));
        if (ret == 0)  
                ret = __wt_sync_and_rename(session,
                    &cb->bfs, WT_BACKUP_TMP, dest));
        if (ret == 0) {
                __wt_writelock(session, &conn->hot_backup_lock);
                conn->hot_backup_list = cb->list;
                __wt_writeunlock(session, &conn->hot_backup_lock);
                F_SET(session, WT_SESSION_BACKUP_CURSOR);
        }
        if (cb->bfs != NULL)
                WT_TRET(__wt_fclose(session, &cb->bfs));
```
I don't know how this code works, so if you're happy with it, fbm.

(I tossed `WT_ASSERT(session, dest != NULL);`. Not sure why it's there historically, but it can't fire in the current code.)

sueloverso:
Thanks @keithbostic. I like those suggestions. I'll merge once PR testing completes.

PR #4503, opened 2019-03-04
WT-4554 Enhance WT salvage to handle the case of corrupted WiredTiger turtle

nehakhatri5:
Check if the turtle file is corrupted before salvaging the metadata. If it's corrupted, delete it and create a new turtle.
@sueloverso Please let me know your feedback for this change.

### DIFF
--- a/src/meta/meta_turtle.c
+++ b/src/meta/meta_turtle.c
@@ -189,7 +189,7 @@ int
 __wt_turtle_init(WT_SESSION_IMPL *session)
 {
 	WT_DECL_RET;
-	char *metaconf, *value;
+	char *metaconf, *unused_value;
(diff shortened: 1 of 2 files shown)


src/meta/meta_turtle.c
```diff
@@ -189,11 +189,12 @@ int
 __wt_turtle_init(WT_SESSION_IMPL *session)
 {
 	WT_DECL_RET;
-	char *metaconf;
-	bool exist_backup, exist_incr, exist_isrc, exist_turtle, load;
+	char *metaconf, *value;
```
sueloverso: I suggest renaming `value` to `unused_value` to make it clear it is not otherwise used.

nehakhatri5:
autobuild: Test this please.

PR #4505, opened 2019-03-04
WT-4613 Processes stuck on 'WT_TEST.wt4333_handle_locks' files on Mac hosts

keithbostic:
@agorrod, @lukech, here are changes to skip the `wt4333_handle_locks` test on OS X, I see no evidence there's a real problem here.

I can't reproduce the problem on any of the 3 test systems I tried, and I don't think it's worth any additional effort to figure out.

### DIFF
--- a/test/csuite/wt4333_handle_locks/main.c
+++ b/test/csuite/wt4333_handle_locks/main.c
@@ -252,7 +252,7 @@ sweep_stats(void)
 }
 
 static void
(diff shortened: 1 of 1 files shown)


keithbostic:
@lukech, please merge if you're happy with this -- thanks!

lukech:
Thanks for investigating and making the change @keithbostic. I will merge.

PR #4509, opened 2019-03-06
WT-4620 Assert a uint64_t can hold a time_t, maintain seconds as uint64_t

keithbostic:
@sueloverso, would you please do the detailed review?
@michaelcahill, I can't think of any reason not to make this change, but wanted to run it by you for sanity.   Thanks!

### DIFF
--- a/src/conn/conn_sweep.c
+++ b/src/conn/conn_sweep.c
@@ -18,7 +18,7 @@
  *	handles.
  */
 static void
-__sweep_mark(WT_SESSION_IMPL *session, time_t now)
+__sweep_mark(WT_SESSION_IMPL *session, uint64_t now)
 {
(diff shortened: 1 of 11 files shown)


keithbostic:
@sueloverso, I took a look and I don't see any errors in our `time_t` handling in the test programs, and there isn't much simplification available, it's a translation in those cases.

That said, I agree it's awkward to leave test programs doing "more portable but less fiddly" things than the library itself is prepared to do. I'll merge this ticket, but if you feel strongly just say so and I'll open a new ticket and fix them.

sueloverso:
I'm happy with this being merged. I only commented on the tests in case you missed them and the mild awkwardness. Thanks for looking into it.

PR #4516, opened 2019-03-09
WT-3929 Changes to allow python3 to be used to run the Python test suite.

ddanderson:


ddanderson:
This is not ready for review/merge, but shows a path to getting Python3 to work with the test suite.
Many tests run, but many do not. I believe the SWIG changes are pretty complete.  The packing code is mostly fixed (though format 8t is noticeably broken) and now transmits Python3 bytes to the SWIG layer.

One possible goal will be to have the code base usable with both Python2 and Python3 before merging to ease the transition (at least for a short time).  This branch does not try to achieve that yet, but I think it should be possible (and straightforward) once we have just Python3 running.

agorrod:
Thanks @ddanderson this looks like a reasonable change to me - definitely worth pushing further. I think it would be good if we can at least start out supporting both python2 and python3 - so we don't need to co-ordinate developers upgrading their systems. If it's not straightforward, it's probably not worth the effort.

ddanderson:
retest this please

ddanderson:
retest this please

ddanderson:
Lesson learned: having errors in the tearDown code is bad: bogus and very long pathnames get constructed as test directories: `build_posix/WT_TEST/.../.../.../...` , and then subsequent runs fail because  one of the first build steps (`git clean -fdqx -e '*.tgz'`) fails.  The recourse is to go into the build machine and remove the `build_posix/WT_TEST` directory.

ddanderson:
I've cleaned up the build machine, and now:
retest this please

agorrod:
I added the locked label until the Windows and modify parts of the change are finalized.

ddanderson:
I'm removing this pull request and branch, it is superseded by #4548 under the ticket WT-4695.

PR #4517, opened 2019-03-13
WT-4494 Support standalone recovery of WiredTiger files.

keithbostic:


src/block/block_open.c
```diff
@@ -274,75 +276,145 @@ __wt_block_close(WT_SESSION_IMPL *session, WT_BLOCK *block)
...
+	 * Get a copy of where the application's file configuration differs from
+	 * the base defaults. The block manager stores a minimal configuration
+	 * so we can read standalone files. Standalone files have limited space
+	 * when first created, so explicitly strip the app_metadata field since
+	 * that's the field that's likely to be large: we'll get a copy of it on
+	 * the first checkpoint (when the standalone file can grow to store as
+	 * much configuration information as we have).
+	 */
+	WT_ERR(__wt_config_discard_defaults(session,
+	    filecfg, metadata, "app_metadata=", &min_config));
```
agorrod: I'd assumed we would go the other way and only include the information necessary to decode the rest of the file in the header. Did you try that?
I'd quite like to fail creation of files with metadata greater than a certain length, probably with a flag to force creation. Which is more appealing if the header block gets a known set of configuration options.

keithbostic: I didn't try that.

The difficulty is ensuring we don't skip anything we need, and of course, remembering the block manager layer when we add new items. I'm sure we could get all that right, but it's trickier. It doesn't save us on code complexity because including only the information we need doesn't ensure the information will fit, we still have to do all the work of checking because a 4KB path to a compressor will blow out the space.

An argument more to your point (because we do have to store the path to the compressor at the block layer), 4KB of column names would also blow out the space and we don't actually need column names in the block layer.

There's something else we also briefly discussed. I think it's possible to grab additional blocks at the start of the file and simply hide them from previous releases. It's going to require  a bit of magic code to pull it apart and stitch it all back together, but I think it's doable. If we did that, then this additional information is just always available, end of story. Let me take a run at that and see how deep that hole is.

keithbostic:
FTR, commits 4de7828 and fc2aedf contain bug fixes for the develop branch. If we don't merge this PR in the next release, the changes in that commit need to be merged into develop separately.

sueloverso:
While I'm not sure of the expected timeline for these changes, I'd be happy with a new ticket getting opened and those two changes for `develop` bug fixes getting in sooner than later.

src/block/block_open.c
keithbostic: @agorrod 
> I'd assumed we would go the other way and only include the information necessary to decode the rest of the file in the header.

I'm going to have to go with your approach -- I hadn't solved the encryption problem (parts of the file configuration string require encryption, for example `columns=`), and I can't find a way to make encryption work. So, I will rework things to write only the information that's required to tear the file apart.

agorrod: That makes me a bit sad, but thanks.

src/session/session_api.c
```diff
@@ -1114,6 +1114,73 @@ __session_drop_readonly(
...
+		WT_ERR_MSG(session, EINVAL,
+		    "WT_SESSION.import requires %s be an absolute path",
+		    path);
+
+	ret = __wt_metadata_search(session, uri, &value);
+	if (ret == 0)
+		WT_ERR_MSG(session, EINVAL,
+		    "an object named \"%s\" already exists in the database",
+		    uri);
+	WT_ERR_NOTFOUND_OK(ret);
```
keithbostic: We don't do anything to prevent importing a file that's already in the database, which would mean you could possible have a single file appear multiple times, by different names, in the database metadata. Nothing good is going to happen after that, but I couldn't think of any easy way to detect it.

src/block/block_ckpt.c
```diff
@@ -686,7 +669,16 @@ __ckpt_process(WT_SESSION_IMPL *session, WT_BLOCK *block, WT_CKPT *ckptbase)
...
+			/*
+			 * Configure final checkpoint recovery magic by flagging
+			 * the checkpoint information to the write routine.
+			 */
+			block->final_ckpt = ckpt;
+			block->final = true;
+			ret = __ckpt_update(session, block, ckpt, ci);
+			block->final = false;
+			block->final_ckpt = NULL;
+			WT_ERR(ret);
```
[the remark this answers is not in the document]
sueloverso: How about a `WT_WITH_FINAL_CKPT` macro? We have a lot of `WT_WITH` macros that are things other than locks like dhandles, btrees, page indexes, etc.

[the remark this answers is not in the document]
keithbostic: I wouldn't object to a `WT_WITH_FINAL_CKPT` macro, but that really just calls out the layering violation involved -- I'm sort of hoping for something more natural than `set flag; operation; clear flag`, but I couldn't think of one myself.

src/btree/bt_import.c
```diff
@@ -0,0 +1,121 @@
...
+		WT_ERR(__wt_buf_set(session, a, v.str, v.len));
+		WT_ERR(__wt_buf_grow(session, b, a->size + 1));
+		((uint8_t *)a->data)[a->size] = '\0';
+	} else {
+		WT_ERR(__wt_buf_set(session, a, v.str, v.len));
+		WT_ERR(__wt_buf_grow(session, b, a->size));
+		WT_ERR(__wt_nhex_to_raw(session, a->data, a->size, b));
+		WT_ERR(__wt_decrypt(session, kencryptor->encryptor, 0, b, a));
+		WT_ERR(__wt_buf_grow(session, b, a->size + 1));
+		((uint8_t *)a->data)[a->size] = '\0';
```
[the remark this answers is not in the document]
keithbostic: This has been cleaned up, so I'm deleting the comment.

src/block/block_ckpt.c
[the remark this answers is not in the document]
sueloverso: Pretty much the definition of `WT_WITH_<something>` is `set something; operation; clear something`. So at least it is consistently wrapped and named with the other things that do that.

src/block/block_ckpt_scan.c
```diff
@@ -0,0 +1,345 @@
...
+ * Copyright (c) 2008-2014 WiredTiger, Inc.
+ *	All rights reserved.
+ *
+ * See the file LICENSE for redistribution information.
+ */
+
+#include "wt_internal.h"
+
+/*
+ * In historic WiredTiger files, it wasn't possible to open standalone files,
```
sueloverso: The first `files` should be `databases`?

src/block/block_ckpt_scan.c
```diff
@@ -0,0 +1,345 @@
...
+ * when scanning the file for the avail lists, we're figuring out exactly the
+ * information needed to fix up the checkpoint information we wrote, that is,
+ * the avail list's offset, size and checksum triplet. As for the final file
+ * size, we allocate all space in the file before we calculate block checksums,
+ * so we can do that space allocation, then fill in the final file size before
+ * calculating the checksum and writing the actual block.
+ *	The second problem is we have to be able to find the avail lists that
+ * include checkpoint information (ignoring previous files created by previous
+ * releases, and, of course, making upgrade/downgrade work seamlessly). Extent
+ * list are written to their own pages, and we could version this change using
```
sueloverso: `list` -> `lists`

src/btree/bt_import.c
```diff
@@ -0,0 +1,121 @@
...
+	 */
+	WT_ERR(__wt_config_getones(
+	    session, metadata, "block_metadata_encrypted", &v));
+	WT_ERR(__wt_btree_config_encryptor(session, filecfg, &kencryptor));
+	if ((kencryptor == NULL && v.val != 0) ||
+	    (kencryptor != NULL && v.val == 0))
+		WT_ERR_MSG(session, EINVAL,
+		    "%s: loaded object's encryption configuration doesn't "
+		    "match the database's encryption configuration",
+		    source);
```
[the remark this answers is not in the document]
sueloverso: To make sure I understand how this applies to encryption. The metadata string is retrieved from the file itself. Inside that metadata would be the encryption information. That would contain the `name` and `keyid` that this table used. This then assumes that the same names and keyids are configured and available in this database. I think that is reasonable. There is no other alternative I can think of.

src/btree/bt_slvg.c
```diff
@@ -157,11 +157,77 @@ static int  __slvg_trk_ovfl(WT_SESSION_IMPL *,
...
+
+	btree = S2BT(session);
+	dhandle = session->dhandle;
+	config = NULL;
+
+	/*
+	 * XXX
+	 * The salvage process reads and discards previous checkpoints, so the
+	 * underlying block manager has to ignore any previous checkpoint
+	 * entries when creating a new checkpoint, in other words, we can't use
```
sueloverso: You can end the sentence after `new checkpoint. In other words,...`

src/btree/bt_slvg.c
```diff
@@ -157,11 +157,77 @@ static int  __slvg_trk_ovfl(WT_SESSION_IMPL *,
...
+	 * the metadata checkpoint list, it has all of those checkpoint listed
```
sueloverso: I think that should be `all of those checkpoints...`

src/btree/bt_slvg.c
```diff
@@ -157,11 +157,77 @@ static int  __slvg_trk_ovfl(WT_SESSION_IMPL *,
...
+	 * and we don't care about them.  Build a clean checkpoint list and use
+	 * it instead.
+	 *
+	 * Don't first clear the metadata checkpoint list and call the function
+	 * to get a list of checkpoints: a crash between clearing the metadata
+	 * checkpoint list and creating a new checkpoint list would look like a
+	 * create or open of a file without a checkpoint to roll-forward from,
+	 * and the contents of the file would be discarded.
```
sueloverso: I think this is fine to point out but it would be good to talk about what the code does rather than what we can't do. I'd prefer to see this comment more concise and then add in `Instead we...`

src/btree/bt_import.c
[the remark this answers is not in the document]
keithbostic: @sueloverso 
> To make sure I understand how this applies to encryption. 

Yes; the `encryption=()` configuration is stored in the clear, everything else is stored encrypted. The `encryption=()` information must be sufficient to decrypt the metadata, further, the database into which we're importing would have to be configured sufficiently similarly to the database in which the object was created, to permit decryption.

I would appreciate a close review of the encrypt/decrypt code, it's not code I'm familiar with. In addition, I abstracted code out of `btree/bt_handle.c:__btree_conf()` (see `__wt_btree_config_encryptor()`), which I'm using to decide if the metadata needs to be encrypted, and I'd like confirmation of that change as well.

Thank you!

src/block/block_ckpt_scan.c
```diff
@@ -0,0 +1,345 @@
...
+	ci.file_size = (wt_off_t)saved->file_size;
+	WT_RET(__wt_buf_extend(
+	    session, checkpoint, WT_BLOCK_CHECKPOINT_BUFFER));
+	endp = checkpoint->mem;
+	WT_RET(__wt_block_ckpt_to_buffer(session, block, &endp, &ci, false));
+	checkpoint->size = WT_PTRDIFF(endp, checkpoint->mem);
+
+	if (WT_VERBOSE_ISSET(session, WT_VERB_CHECKPOINT))
+		__wt_ckpt_verbose(
+		    session, block, "import replace", NULL, checkpoint->mem);
```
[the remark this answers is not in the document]
keithbostic: What's happening in this function is only in the service of import. While we're updating checkpoint information, I don't want this to be confused with actions taken during a checkpoint.

src/block/block_ckpt_scan.c
```diff
@@ -0,0 +1,345 @@
...
+	memset(&saved, 0, sizeof(saved));
+
+	ext_off = 0;			/* [-Werror=maybe-uninitialized] */
+	ext_size = 0;
+	len = write_gen = 0;
+
+	F_SET(session, WT_SESSION_QUIET_CORRUPT_FILE);
+
+	WT_ERR(__wt_scr_alloc(session, 64 * 1024, &tmp));
+
```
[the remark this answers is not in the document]
keithbostic: They're 4 different logical chunks: (1) set side-effect returns, (2) initialize local variables, (3) set a session flag, (4) allocate temporary memory. I did remove one line that separated two things that were both part of #2.

I'm gonna buy you a bigger monitor... ;)

src/btree/bt_slvg.c
keithbostic: @sueloverso 
> it would be good to talk about what the code does rather than what we can't do.

I think the first part of that comment talks about what we do -- can you me know what you think it's missing?

src/block/block_ckpt_scan.c
keithbostic: Fixed in c55b7c2.

src/block/block_ckpt_scan.c
keithbostic: Fixed in c55b7c2.

src/block/block_ckpt_scan.c
```diff
@@ -0,0 +1,345 @@
...
+	 * checkpoint, we're the only writer. Plus, because we use the write
+	 * generation as a database connection generation, it's guaranteed to
+	 * move forward and never repeat.
+	 *	It's a layering violation though, this is the only place the
+	 * block manager uses the write generation. The alternative would be to
+	 * add our own write-generation scheme in the block manager, storing a
+	 * value and recovering it when we open the file. We could do that, as
+	 * reading the final avail list when a file is opened is unavoidable,
+	 * so we can retrieve the value written here when we open the file, but
+	 * this approach is simpler.
```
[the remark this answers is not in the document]
keithbostic: I added a lot of wording here because it's a reasonably bad layering violation.

src/block/block_ckpt_scan.c
```diff
@@ -0,0 +1,345 @@
...
+
+	WT_ERR(__wt_scr_alloc(session, 64 * 1024, &tmp));
+
+	/*
+	 * Scan the file, starting after the descriptor block, looking for
+	 * pages. The minimum block size in WiredTiger is 512B, use that as
+	 * our minimum scan chunk.
+	 */
+	fh = block->fh;
+	for (nblocks = 0, offset = 512; offset < block->size; offset += size) {
```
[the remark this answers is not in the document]
keithbostic: Agreed, fixed in c55b7c2.

src/btree/bt_slvg.c
keithbostic: Fixed in c55b7c2.

src/btree/bt_slvg.c
keithbostic: Reworded for clarity in c55b7c2.

keithbostic:
@michaelcahill, I ended up going down the `evergreen.yml` rabbit-hole today. There are extensive changes, centered around  making `top_srcdir` and `top_builddir` the same for all test programs, which required making the `test/csuite` scripts pay attention to those values in all cases, and test programs work when called from other directories.  Feel free to throw this all away if you don't like it or have a better approach (but then we'll need to make `import` build & run in Evergreen in some other way). I don't understand `evergreen.yml` all that well, I may have missed a better fix.

@lukech, depending on your willingness to leave the `working_dir` variable in the `evergreen.yml` recipe alone, this work may cover WT-4846.

keithbostic:
@lukech, I've added you as a reviewer, if you can please take a look at the `evergreen.yml` and associated test changes and let me know if you approve.

The command `git diff 5c1b667 a7c4910` should give you a pretty clean diff of just those changes.

test/csuite/import/smoke.sh
```diff
@@ -0,0 +1,56 @@
+#! /bin/sh
+
+set -e
+
+# Bypass this test for slow machines, valgrind
+test "$TESTUTIL_SLOW_MACHINE" = "1" && exit 0
+test "$TESTUTIL_BYPASS_VALGRIND" = "1" && exit 0
+
+# If $top_builddir/$top_srcdir aren't set, default to running in test/csuite.
```
lukech: I guess the idea is to assume the default location of calling smoke.sh is from within `build_posix/test/csuite`. The `test/csuite` in comment makes me think of a different directory outside of `build_posix`.

test/csuite/wt2909_checkpoint_integrity/main.c
```diff
@@ -541,17 +540,25 @@ subtest_main(int argc, char *argv[], bool close_test)
 	testutil_check(__wt_snprintf(
 	    filename, sizeof(filename), "%s/%s", opts->home, STDOUT_FILE));
 	testutil_assert(freopen(filename, "a", stdout) != NULL);
+
+	/*
+	 * Use $top_builddir if it's available, otherwise assume we're running
+	 * from the test/csuite directory.
```
lukech: Can `test/csuite` be replaced with `build_posix/test/csuite` here to avoid confusion?

test/csuite/import/smoke.sh
keithbostic: `build_posix` is not required. I personally never build in `build_posix`, I always build in the top-level directory.

test/csuite/wt2909_checkpoint_integrity/main.c
keithbostic: Same as above -- there's no reason to assume we're building in `build_posix`.

test/csuite/import/smoke.sh
lukech: That's a good point. Though in evergreen.yml it's set to build under `build_posix`, other directories  are supported as well. Ignore my comment then.

michaelcahill:
@keithbostic, thanks for pushing this through to a conclusion.  I'm going to go ahead and merge: there were a few comments that may warrant follow up, but I'm keen to see this big branch merged, so we can deal with any test fallout.  Hopefully any follow up will cause minimal code change.

PR #4520, opened 2019-03-14
WT-4585 Add WT_WITH_HOTBACKUP_LOCK macro

tetsuo-cpp:
This PR is adding a convenience macro for managing the hot backup lock as well as a corresponding session flag. The macro is only for acquiring the lock for reading since this is the more common case.


tetsuo-cpp:
retest this please

src/include/schema.h
```diff
@@ -256,6 +256,25 @@ struct __wt_table {
...
+ */
+#define	WT_WITH_HOTBACKUP_LOCK(session, op) do {		\
+	WT_CONNECTION_IMPL *__conn = S2C(session);			\
+	if (F_ISSET(session, WT_SESSION_HOTBACKUP_LOCKED)) {		\
+		op;							\
+	} else {							\
+		__wt_readlock(session, &__conn->hot_backup_lock);	\
+		F_SET(session, WT_SESSION_HOTBACKUP_LOCKED);		\
+		do {							\
+			op;						\
```
tetsuo-cpp: I've wrapped this in a loop so that when we decide to jump out of the code block with `break`, we unlock cleanly.

src/schema/schema_util.c
```diff
@@ -26,24 +26,24 @@ __wt_schema_backup_check(WT_SESSION_IMPL *session, const char *name)
...
+	WT_WITH_HOTBACKUP_LOCK(session, {
+		/*
+		 * There is a window at the end of a backup where the list has
+		 * been cleared from the connection but the flag is still set.
+		 * It is safe to drop at that point.
+		 */
+		if (!conn->hot_backup ||
+		    (backup_list = conn->hot_backup_list) == NULL) {
+			ret = 0;
 			break;
```
tetsuo-cpp: Instead of explicitly unlocking and returning here, I'm setting `ret` and breaking which will unlock on the way out.

src/log/log.c
```diff
@@ -1517,17 +1520,20 @@ __log_truncate_file(WT_SESSION_IMPL *session, WT_FH *log_fh, wt_off_t offset)
...
-		else {
-			ret = __wt_ftruncate(session, log_fh, offset);
-			__wt_readunlock(session, &conn->hot_backup_lock);
+		WT_WITH_HOTBACKUP_LOCK(session, {
+			if (!conn->hot_backup) {
+				truncate = true;
+				ret = __wt_ftruncate(session, log_fh, offset);
+			}
+		});
+		if (truncate) {
```
tetsuo-cpp: There were a few cases like this where I couldn't easily reproduce the sequence of locks/unlocks. So I had to introduce a variable to know whether to continue with the `else` block after I've released the lock.

sueloverso:
This change is unlike the other uses of `WT_WITH_*LOCK` macros. I suggest the following:
* It does not do the write lock.
* The name does not reflect it is only for the read lock.
* To address both of those I suggest looking at the definition and usage of the `WT_WITH_HANDLE_LIST_READ_LOCK` and `WT_WITH_HANDLE_LIST_WRITE_LOCK`.
* No other uses of any of the locking macros entail multi-line statements. Please refactor the code into wrapper functions that can be called in a single like all the other uses. It is likely that will clean up the special cases you had to add for error handling and unlocking.

tetsuo-cpp:
@sueloverso 
Thanks for the review. I believe I've addressed your comments now:

- There is now a separate macro for the write lock.
- The naming of the macro and the corresponding bit flag reflects this.
- No more multi-line statements within the macro invocation.
- Some of the functions that I refactored out were a bit hard for me to describe in a way that isn't simply "critical section of the function calling it". I've tried to give them proper names where possible but, in those cases, I've called them `<parent_function_name>_requires_hot_backup_lock` to indicate that it's just a helper function containing code that needs the lock to be acquired.
- In the case where the body is an `if` statement guarding another function call, I have not refactored into a separate function since it is still a single statement. Let me know if this is still pushing it and I can refactor them into separate functions. I just felt that it'd be a bit overkill to have functions like this littered around the codebase:
```
static void
__func_helper()
{
    if (cond)
        __func();
}
```
- Likewise in `cur_backup.c`, we need to do two assignments with the hot backup lock. I've just done them in a single statement using the comma operator but I can extract this into a function if necessary.

src/txn/txn_ckpt.c
```diff
@@ -1271,6 +1271,93 @@ __drop_to(WT_CKPT *ckptbase, const char *name, size_t len)
 	}
 }
 
+/*
+ * __checkpoint_lock_dirty_tree_requires_hot_backup_lock --
```
sueloverso: This naming is rather verbose. It really is the main content of the original function, I just call it `__checkpoint_lock_dirty_tree_int` as the internal version of the function with the lock.

src/schema/schema_util.c
```diff
@@ -9,41 +9,56 @@
...
- * this is done.
+ * __schema_backup_check_requires_hot_backup_lock --
+ *	Helper for __wt_schema_backup_check.  Intended to be called while
+ *	holding the hot backup read lock.
  */
-int
-__wt_schema_backup_check(WT_SESSION_IMPL *session, const char *name)
+static int
+__schema_backup_check_requires_hot_backup_lock(
+    WT_SESSION_IMPL *session, const char *name)
```
sueloverso: Like the others, just call this the `_int` version of its parent function.

src/log/log.c
```diff
@@ -1504,6 +1522,27 @@ __wt_log_acquire(WT_SESSION_IMPL *session, uint64_t recsize, WT_LOGSLOT *slot)
 	return (0);
 }
 
+/*
+ * __log_truncate_file_requires_hot_backup_lock --
+ *	Helper for __log_truncate_file. Intended to be called while holding
+ *	the hot backup read lock.
```
sueloverso: And here too for the `_int` naming suggestion.

src/include/schema.h
```diff
@@ -256,6 +259,39 @@ struct __wt_table {
...
+ * WT_WITH_HOTBACKUP_READ_LOCK, WT_WITH_HOTBACKUP_WRITE_LOCK --
+ * 	Acquire the hot backup lock, perform an operation, drop the lock.
+ */
+#define	WT_WITH_HOTBACKUP_READ_LOCK(session, op) do {			\
+	WT_CONNECTION_IMPL *__conn = S2C(session);			\
+	if (F_ISSET(session, WT_SESSION_LOCKED_HOTBACKUP)) {		\
+		op;							\
+	} else {							\
+		__wt_readlock(session, &__conn->hot_backup_lock);	\
+		F_SET(session, WT_SESSION_LOCKED_HOTBACKUP_READ);	\
```
sueloverso: If you made the following line `if (!conn->hot_backup)` then almost every multi-line issue and lots of wrapper functions go away.

src/schema/schema_util.c
tetsuo-cpp: bc1914f

src/log/log.c
tetsuo-cpp: bc1914f

src/include/schema.h
tetsuo-cpp: This approach won't work for `__checkpoint_lock_dirty_tree_int` from `txn_ckpt.c`.
I also feel that it'd be misleading to have a macro that doesn't necessarily execute its given statement.

I could have a macro just for that check and call it like:
```
WT_WITH_HOTBACKUP_READ_LOCK(session,
    WT_IFN_HOTBACKUP(ret = __wt_ftruncate(session, block->fh, len)));
```
But it seems to be not much of a win over just having the condition there.

src/txn/txn_ckpt.c
tetsuo-cpp: bc1914f

tetsuo-cpp:
retest this please

src/txn/txn_ckpt.c
```diff
@@ -1378,92 +1461,20 @@ __checkpoint_lock_dirty_tree(WT_SESSION_IMPL *session,
...
-			    "cannot be deleted during a hot backup",
-			    ckpt->name);
-		}
-
-	/*
-	 * Mark old checkpoints that are being deleted and figure out which
-	 * trees we can skip in this checkpoint.
-	 */
-	WT_ERR(__checkpoint_mark_skip(session, ckptbase, force));
+	WT_WITH_HOTBACKUP_READ_LOCK_CKPT(session,
```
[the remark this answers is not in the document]
tetsuo-cpp: de8f83f

src/conn/conn_log.c
```diff
@@ -389,19 +415,10 @@ __log_archive_once(WT_SESSION_IMPL *session, uint32_t backup_file)
...
-			WT_ERR(__wt_log_extract_lognum(
-			    session, logfiles[i], &lognum));
-			if (lognum < min_lognum)
-				WT_ERR(__wt_log_remove(
-				    session, WT_LOG_FILENAME, lognum));
-		}
-	}
-	__wt_readunlock(session, &conn->hot_backup_lock);
-	locked = false;
+	WT_WITH_HOTBACKUP_READ_LOCK(session,
```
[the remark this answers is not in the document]
tetsuo-cpp: de8f83f

src/schema/schema_util.c
```diff
@@ -9,41 +9,54 @@
...
-	if (!conn->hot_backup)
-		return (0);
-	__wt_readlock(session, &conn->hot_backup_lock);
+
 	/*
 	 * There is a window at the end of a backup where the list has been
 	 * cleared from the connection but the flag is still set.  It is safe
 	 * to drop at that point.
 	 */
 	if (!conn->hot_backup ||
```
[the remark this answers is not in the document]
tetsuo-cpp: de8f83f

This is the part that was breaking the build. We actually want to execute the code here if there is a backup in progress (we return out immediately if `!conn->hot_backup` or `backup_list == NULL`). So I can't use the macro here.

I decided to use the specialized checkpoint macro here and renamed the macro to `WT_WITH_HOTBACKUP_READ_LOCK_UNCOND` to reflect that its being used for things other than checkpointing.

src/include/schema.h
```diff
@@ -256,6 +259,76 @@ struct __wt_table {
...
+/*
+ * WT_WITH_HOTBACKUP_READ_LOCK --
+ *	Acquire the hot backup read lock and perform an operation provided that
+ *	there is no hot backup in progress.  The skipp parameter can be used to
+ *	check whether the operation got skipped or not.
+ */
+#define	WT_WITH_HOTBACKUP_READ_LOCK(session, op, skipp) do {		\
+	WT_CONNECTION_IMPL *__conn = S2C(session);			\
+	bool *__skipp = skipp;						\
+	if (__skipp)							\
```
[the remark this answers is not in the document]
tetsuo-cpp: de8f83f

Had to do a bit of casting to get around compiler warnings about redundant `NULL` checks.

src/conn/conn_log.c
```diff
@@ -342,6 +342,34 @@ __wt_logmgr_reconfig(WT_SESSION_IMPL *session, const char **cfg)
...
+__log_archive_once_int(WT_SESSION_IMPL *session,
+    uint32_t backup_file, char **logfiles, u_int logcount, uint32_t min_lognum)
+{
+	WT_CONNECTION_IMPL *conn;
+	uint32_t lognum;
+	u_int i;
+
+	conn = S2C(session);
+
+	if (!conn->hot_backup || backup_file != 0) {
```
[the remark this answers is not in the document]
tetsuo-cpp: de8f83f

src/include/schema.h
```diff
@@ -256,6 +259,76 @@ struct __wt_table {
...
+		op;							\
+		F_CLR(session, WT_SESSION_LOCKED_HOTBACKUP_WRITE);	\
+		__wt_writeunlock(session, &__conn->hot_backup_lock);	\
+	}								\
+} while (0)
+
+/*
+ * WT_WITH_HOTBACKUP_READ_LOCK_CKPT --
+ *	Acquire the hot backup read lock and perform an operation.  This is a
+ *	specialised macro for the checkpoint logic.  Code that wishes to acquire
```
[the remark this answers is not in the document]
tetsuo-cpp: de8f83f

PR #4521, opened 2019-03-15
WT-4324 Checkpoints must rewrite pages with data in the future.

michaelcahill:
Sometimes we attempt eviction of a page by writing the most recent versions of updates.  If such a page remains dirty in cache, it must be visited by the next checkpoint to make sure the data that appears in the checkpoint is a consistent cut of versions.


michaelcahill:
I don't love this solution: it's just one more special case.

This ticket has also made me realize that we need some developer documentation describing what "clean" and "dirty" really mean, and all of the optimizations we have in place that try to minimize the work checkpoints need to do.

That said, I believe this change fixes the bug, so I've pushed it absent a more holistic solution.

PR #4523, opened 2019-03-18
WT-4643 Adding commit timestamp validitiy check into txn commit

luke-pearson:
This PR fixes the scenario described in WT-4643, it additionally fixes a bug where the commit timestamp can be set at the stable timestamp. These fixes are tested in the provided python test file.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -776,6 +776,7 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	wt_timestamp_t prev_commit_timestamp, ts;
 	uint32_t fileid;
 	u_int i;
+	char ts_string[2][WT_TS_INT_STRING_SIZE];
 	bool locked, prepare, readonly, update_timestamp;
 
 	txn = &session->txn;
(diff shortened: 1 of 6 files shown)


src/txn/txn_timestamp.c
```diff
@@ -612,7 +612,7 @@ __wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 		    "%s timestamp %.*s older than oldest timestamp %s",
 		    name, (int)cval->len, cval->str, ts_string[0]);
 	}
-	if (durable_ts && has_stable_ts && ts < stable_ts) {
+	if (durable_ts && has_stable_ts && ts <= stable_ts) {
```
luke-pearson: Fixes scenario where commit timestamp can be set at the stable timestamp

luke-pearson:
retest this please

luke-pearson:
retest this please

src/txn/txn.c
```diff
@@ -837,6 +838,23 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		    session, "durable", ts, &cval, true));
 	}
 
+	/*
+	 * Confirm if the first commit timestamp is valid
+	 * with both prepared and unprepared transactions.
+	 */
+	if (txn_global->has_stable_timestamp &&
```
luke-pearson: I'm not sure if this is the best method to do this but it does avoid moving it into the assert function (which is presumably for when asserting is turned on) and it also avoids calling the much more expensive validate function.

bvpvamsikrishna: For prepared transactions, first commit timestamp will not be set. The condition '(prepare && !F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))' is redundant.

bvpvamsikrishna: Better to move this check to '__wt_txn_commit_timestamp_validate'

luke-pearson: > For prepared transactions, first commit timestamp will not be set. The condition '(prepare && !F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))' is redundant.

Does this mean that my test scenarios 4 and 5 are invalid? The test will fail if I remove this check, implying the first commit timestamp is set.

luke-pearson:
retest this please

luke-pearson:
Closing as ticket is currently blocked

PR #4524, opened 2019-03-18
WT-4645 Add isolation level tests

tetsuo-cpp:
These situations are indirectly tested in a number of existing tests. They typically use the `make_scenarios` helper to create a list of operations and then make some assertions to periodically check how many records are visible at each isolation level.

I thought it'd be nice to have some tests that check the isolation levels in detail and ensure that the expected value is visible (along with some comments to aid understanding).

### DIFF
--- a/test/suite/test_txn20.py
+++ b/test/suite/test_txn20.py
@@ -1,0 +1,115 @@
+#!/usr/bin/env python
+#
+# Public Domain 2014-2019 MongoDB, Inc.
+# Public Domain 2008-2014 WiredTiger, Inc.
+#
+# This is free and unencumbered software released into the public domain.
+#
+# Anyone is free to copy, modify, publish, use, compile, sell, or
+# distribute this software, either in source code form or as a compiled
+# binary, for any purpose, commercial or non-commercial, and by any
+# means.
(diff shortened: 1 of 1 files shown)


test/suite/test_txn20.py
```diff
@@ -0,0 +1,121 @@
...
+        self.session.commit_transaction()
+        self.assertEqual(cursor['key: aaa'], 'value: aaa')
+
+        # Cleanup.
+        self.session.close()
+        s.rollback_transaction()
+        s.close()
+
+if __name__ == '__main__':
+    wttest.run()
```
bvpvamsikrishna: Between different test cases, it seems one or two lines are different. In that case we can use make_scenario's utility to reduce the test code. If not clear how to do it, we can discuss.

tetsuo-cpp: How does this help? It looks like `make_scenarios` generates a series of operations whereas I have a specific operation that I want to test.

I can put some of the code in some `setUp` and `tearDown` functions if that's helpful.

tetsuo-cpp: cc6c6fe

PR #4525, opened 2019-03-18
WT-4646 Coverity false positive on fopen calls

keithbostic:
@sueloverso, I don't see the problem, so I just tried to quiet the complaint.

Please check me!

### DIFF
--- a/test/salvage/salvage.c
+++ b/test/salvage/salvage.c
@@ -160,7 +160,7 @@ run(int r)
 
 	testutil_make_work_dir(HOME);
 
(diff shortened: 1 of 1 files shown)


keithbostic:
> I agree this is a strange warning. Did either the salvage test or the assert code change recently? Why did this just come up now? Or did Coverity itself change?

My guess is Coverity changed, it's happened before. (Salvage changed a few months ago, though, we may have been having difficulty getting our runs to execute since then, that's also happened before.)

PR #4528, opened 2019-03-21
WT-4631 Setting read timestamp to WT_TS_NONE in clear readtimestamp

luke-pearson:
…unction

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -150,6 +150,20 @@ __wt_txn_parse_timestamp(WT_SESSION_IMPL *session, const char *name,
 }
 
 /*
+ * __txn_get_read_timestamp --
+ *	Get the read timestamp from the transaction. Additionally
+ *	return bool to specify whether the transaction has set
+ *	clear read queue flag.
+ */
+static void
+__txn_get_read_timestamp(
+    WT_TXN *txn, wt_timestamp_t *read_timestampp, bool *validp)
+{
+	WT_ORDERED_READ(*read_timestampp, txn->read_timestamp);
+	*validp = !txn->clear_read_q;
+}
+
+/*
  * __txn_get_pinned_timestamp --
  *	Calculate the current pinned timestamp.
  */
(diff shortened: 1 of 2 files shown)


luke-pearson:
I think this results in a very low chance race in txn_timestamp.c at line 199, which could possibly be fixed by storing txn->read_timestamp in a temp variable prior to accessing it.

michaelcahill:
@luke-pearson, I think WT-4621 added another similar case here: https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn_timestamp.c#L783-L789

michaelcahill:
Maybe consider something like this:

```static void
__txn_get_read_timestamp(WT_TXN *txn, wt_timestamp_t *read_timestampp, boolean *validp)
{
    *read_timestampp = txn->read_timestamp;
    WT_ORDERED_READ(*validp, !txn->clean_read_q);
}```

File: src/txn/txn_timestamp.c [hunk 1]
```diff
@@ -187,6 +187,8 @@ __txn_get_pinned_timestamp(
 	/* Look for the oldest ordinary reader. */
 	__wt_readlock(session, &txn_global->read_timestamp_rwlock);
 	TAILQ_FOREACH(txn, &txn_global->read_timestamph, read_timestampq) {
+		/* Copy value out to prevent possible race condition. */
+		tmp_read_timestamp = txn->read_timestamp;
```
agorrod: The compiler can optimize this assignment away (similarly below). You could either add an explicit read barrier or mark the variable declaration volatile (https://barrgroup.com/Embedded-Systems/How-To/C-Volatile-Keyword).

This is subtle - I'd be happy to talk through the reasons and alternatives tomorrow if you'd like.

luke-pearson: I've pushed up a change which uses read barriers, let me know what you think. I'd be interested in having a discussion about them though tomorrow.

luke-pearson:
retest this please

PR #4529, opened 2019-03-21
WT-4644 column-store missing run-length encoding opportunities

keithbostic:
@michaelcahill, this change means we check for `updates that have timestamps but are still globally visible`, earlier in reconciliation than before. Before, we checked during cell layout, but that's too late for variable-length column store, because it's already selected the RLE for the item and we can't go back and fix it up if the timestamp changes mean that we could RLE collapse a pair of adjacent items.

It wouldn't be an error to just write the two adjacent items with identical timestamps, but I **really** don't  want to give up the verification checks as to whether or not adjacent column-store items should have been collapsed by RLE (and there are probably workloads where that's the wrong thing anyway and we'd write more data by doing that).

I tried to separate the column-store and row-store reconciliation paths so we can check for row-store updates later, but not check column-store updates, but that was ugly and complex.


agorrod:
@keithbostic is this ready for review?

keithbostic:
@agorrod 
> is this ready for review?

Yes,  I think so.

@michaelcahill should sign off on where I plan to do the final "do we need timestamps for this value", test but otherwise it's pretty straight-forward.

Thanks!

src/reconcile/rec_write.c
```diff
@@ -1407,6 +1406,20 @@ __rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
...
+		 *
+		 * TIMESTAMP-FIXME
+		 * Values (presumably) have associated transaction IDs, but we
+		 * haven't yet decided how to handle them. For now, ignore them
+		 * in determining value durability.
+		 */
+		if (upd_select->start_ts != WT_TS_NONE &&
+		    __wt_txn_visible_all(
+		    session, WT_TXN_NONE, upd_select->start_ts))
+			upd_select->start_ts = WT_TS_NONE;
```
michaelcahill: I'm uneasy about reconciliation (or any code) changing anything in a committed update structure.  We have a lot of assumptions in many places of the code that committed updates are immutable.

That said, there is already a FIXME here, and I expect the real solution will be to create a `WT_UPDATE_VALIDITY` struct (or similar) that will hold the relevant information (start and stop timestamps, the latter should be calculated here and removed from `WT_UPDATE`, along with transaction IDs).  That should mean reconciliation is not making any modifications to updates.

PR #4530, opened 2019-03-21
WT-4317 Zero the end of the file if we don't have ftruncate.

sueloverso:
Here's a branch that avoids the salvage problem and zeroes out the end of a table that we wanted to ftruncate but it's not available.

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -627,6 +627,7 @@ dsrc_stats = [
     BlockStat('block_minor', 'minor version number', 'max_aggregate,no_scale'),
     BlockStat('block_reuse_bytes', 'file bytes available for reuse', 'no_scale,size'),
     BlockStat('block_size', 'file size in bytes', 'no_scale,size'),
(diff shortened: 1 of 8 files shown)


sueloverso:
This branch should not be merged as-is. It is not the solution for WT-4317 as the problem could happen having the same data in a different part of the file.

However, doing this branch caused me to make the `log_zero` function more general and move it to the `os_common` directory for generic zero'ing of a file at some offset for some length. @agorrod is that code refactoring worth keeping? Anyone have an opinion? I would need to remove the call to it from `block_write.c` but the function would be available in the future.

agorrod:
@sueloverso I've taken a look at the change, and am pretty neutral about the code movement. I agree that zero'ing a range makes some more sense in the filesystem code, but there are some optimizations in there that are pretty specific to logging. I'm happy to take your recommendation when you return.

sueloverso:
Closing this PR in favor of #4555

PR #4535, opened 2019-03-26
WT-4658 Review Clang Format changes

tetsuo-cpp:
This PR is a preview of what `develop` looks like after applying `s_clang-format`. I've opened this up to review/discuss the changes.


tetsuo-cpp:
@keithbostic 
When you suggested to open a PR to review/discuss the changes, did you intend for it to be merged eventually? I think that if this review takes a while, it will be hard to merge. So my idea was to:

* Discuss in this PR what changes we want to make to our code (e.g. removing problematic macros, adding more formatting scripts, etc). This PR will stick around for a while but there is no intention to merge.
* Make the discussed changes in a separate PR without applying Clang Format.
* Follow up with one final PR to apply the Clang Format changes for real. This one is intended to be merged and will only be open for a moment.

However, one of the things you brought up was that you wanted to check for any correctness issues introduced by the formatter. So I'm thinking that if we review for correctness in this PR and then blindly apply the formatter later (when other changes have gone into `develop`) then this may not be ok.

src/btree/bt_io.c
```diff
@@ -115,21 +109,19 @@ __wt_bt_read(WT_SESSION_IMPL *session,
...
-		if (ip != NULL)
-			WT_ERR(__wt_buf_set(
-			    session, buf, ip->data, dsk->mem_size));
+	    /*
+	     * If we uncompressed above, the page is in the correct buffer.
+	     * If we get here the data may be in the wrong buffer and the
+	     * buffer may be the wrong size.  If needed, get the page
+	     * into the destination buffer.
+	     */
+	    if (ip != NULL)
```
tetsuo-cpp: This indentation is a known issue from the last PR. I can wrap the inner `if` statement with braces to solve this.

src/btree/bt_curnext.c
```diff
@@ -660,14 +649,12 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
...
-				    cbt, newpage, restart);
+				ret = __cursor_fix_append_next(cbt, newpage, restart);
 				break;
 			case WT_PAGE_COL_VAR:
-				ret = __cursor_var_append_next(
-				    cbt, newpage, restart);
+				ret = __cursor_var_append_next(cbt, newpage, restart);
 				break;
-			WT_ILLEGAL_VALUE_ERR(session, page->type);
+				WT_ILLEGAL_VALUE_ERR(session, page->type);
```
tetsuo-cpp: Known issue. This just looks like a function call so it's understandable that it gets indented under the `case` statement above.

I think the solution is to just remove this macro entirely since it doesn't provide much value.

src/async/async_api.c
```diff
@@ -43,7 +43,7 @@ __async_get_format(WT_CONNECTION_IMPL *conn, const char *uri,
 	 * is a possibility a duplicate entry might be inserted, but
 	 * that is not harmful.
 	 */
-	TAILQ_FOREACH(af, &async->formatqh, q) {
+	TAILQ_FOREACH (af, &async->formatqh, q) {
```
tetsuo-cpp: I've listed these macros under the `ForEachMacros` list in Clang Format (which is why it indents nicely).
Unfortunately, a side effect of this is what it treats `TAILQ_FOREACH` like `for` and puts a space after it.

src/btree/bt_io.c
```diff
@@ -146,23 +138,23 @@ __wt_bt_read(WT_SESSION_IMPL *session,
 	WT_STAT_CONN_INCRV(session, cache_bytes_read, dsk->mem_size);
 	WT_STAT_DATA_INCRV(session, cache_bytes_read, dsk->mem_size);
 	WT_STAT_SESSION_INCRV(session, bytes_read, dsk->mem_size);
-	(void)__wt_atomic_add64(
-	    &S2C(session)->cache->bytes_read, dsk->mem_size);
+	(void)__wt_atomic_add64(&S2C(session)->cache->bytes_read, dsk->mem_size);
 
 	if (0) {
-corrupt:	if (ret == 0)
+	corrupt:
```
tetsuo-cpp: `goto` labels get indented which isn't great. However it should be trivial to write a script that flushes these labels to the left.

src/include/btmem.h
```diff
@@ -226,33 +222,31 @@ struct __wt_ovfl_reuse {
...
-#define	WT_LOOKASIDE_COMPRESSOR	"none"
+#define WT_LOOKASIDE_COMPRESSOR "none"
 #endif
-#define	WT_LAS_CONFIG							\
-    "key_format=" WT_UNCHECKED_STRING(QIQu)				\
-    ",value_format=" WT_UNCHECKED_STRING(QQQBBu)			\
-    ",block_compressor=" WT_LOOKASIDE_COMPRESSOR			\
-    ",leaf_value_max=64MB"						\
-    ",prefix_compression=true"
+#define WT_LAS_CONFIG                                                                              \
```
tetsuo-cpp: The formatter doesn't handle macros well if they don't look like real C.
The server had this problem too but WiredTiger gets more mileage out of the preprocessor so it'll be more of a problem.

src/block/block_addr.c
```diff
@@ -190,9 +186,8 @@ __wt_block_buffer_to_ckpt(WT_SESSION_IMPL *session,
  * version.
  */
 int
-__wt_block_ckpt_decode(WT_SESSION *wt_session,
-    size_t allocsize, const uint8_t *p, WT_BLOCK_CKPT *ci)
-    WT_GCC_FUNC_ATTRIBUTE((visibility("default")))
+__wt_block_ckpt_decode(WT_SESSION *wt_session, size_t allocsize, const uint8_t *p,
+    WT_BLOCK_CKPT *ci) WT_GCC_FUNC_ATTRIBUTE((visibility("default")))
```
tetsuo-cpp: More issues with macros. Ideally it'd be on its own line but the formatter removes the line break.

src/block/block_read.c
```diff
@@ -113,18 +109,16 @@ __wt_bm_read(WT_BM *bm, WT_SESSION_IMPL *session,
...
-    WT_ITEM *buf, wt_off_t offset, uint32_t size, uint32_t checksum)
-    WT_GCC_FUNC_ATTRIBUTE((cold))
+__wt_bm_corrupt_dump(WT_SESSION_IMPL *session, WT_ITEM *buf, wt_off_t offset, uint32_t size,
+    uint32_t checksum) WT_GCC_FUNC_ATTRIBUTE((cold))
 {
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	size_t chunk, i, nchunks;
 
-#define	WT_CORRUPT_FMT	"{%" PRIuMAX ", %" PRIu32 ", %#" PRIx32 "}"
```
tetsuo-cpp: `#define`s are not tabbed out.

src/btree/bt_curnext.c
michaelcahill: @tetsuo-cpp do you mean remove the macro and replace uses with explicit `default:` labels?  There is value in this kind of error checking, even if the macro causes problems.

tetsuo-cpp: Yeah sorry, I should have clarified. I was suggesting to remove the macro and just write the contents inline. It won't be much more verbose since it just contains this:

```
#define WT_ILLEGAL_VALUE(session, v)                                    \
        default:                                                        \
                return (__wt_illegal_value(session, v))
```

keithbostic:
> When you suggested to open a PR to review/discuss the changes, did you intend for it to be merged eventually? I think that if this review takes a while, it will be hard to merge.

As you suggested, I think the key is to make "real" changes in other pull requests, then we update this branch as those changes are merged, in which case this branch eventually becomes the final branch you talked about.

It might be useful to do those merges and updates, because then we can see the specific changes caused by other changes in the system. That said, I don't have any experience here, so I'm hesitant to push any particular approach, keeping it simpler until we need something more complex is probably better.

Regardless, I agree, the changes in this PR are auto-generated.

> However, one of the things you brought up was that you wanted to check for any correctness issues introduced by the formatter. So I'm thinking that if we review for correctness in this PR and then blindly apply the formatter later (when other changes have gone into develop) then this may not be ok.

I think that's OK.

src/btree/bt_io.c
keithbostic: Wrapping the else in braces seems like something we could live with.

src/btree/bt_io.c
keithbostic: Lint prefers them on separate lines, too.

src/include/btmem.h
keithbostic: In some places we may be able to fix this by switching to inlined functions. That's better for debugging, anyway.

keithbostic:
This is totally going to mess my brain up. :)

lukech:
evergreen retest

lukech:
evergreen retry

lukech:
evergreen retry

tetsuo-cpp:
We've come to an agreement on what needs to happen here.
I'll work on some changes to our formatting style, scripts, etc and will open another PR shortly.

PR #4537, opened 2019-03-26
WT-4636 Fix strace test

tetsuo-cpp:
There are sort of two fixes rolled into one here as I had to jump through some hoops to fix the issue described in the ticket.

The actual Jenkins failure originates from my changes to accomodate Clang Tidy. Clang Tidy recommends that for calls to `open`, that you supply the `O_CLOEXEC` flag which prevents forked processes from inheriting file descriptors from the parent. I think the reasoning for making the change is correct so my fix was to change the expected `strace` output to include the new flag.

My original comment about this change is here: https://github.com/wiredtiger/wiredtiger/pull/4501#discussion_r261496029

This failure was reproducible on `tinderbox` but locally, I was getting a different failure. On my Ubuntu dev box, `strace` doesn't pick up any `open` calls. After a bit of investigation, I found that newer versions of glibc (later than 2.26) implement `open` in terms of `openat`. So I've modified the test to watch for `openat` calls as well and added some conditionals to let our test know which calls it ought to be expecting.


tetsuo-cpp:
retest this please

agorrod:
@ddanderson please take a look at this change.

PR #4538, opened 2019-03-27
WT-4614 Trigger sweep immediately to reclaim space after a drop

sulabhM:


sulabhM:
@agorrod, I did create an evergreen patch and tested the original issue. It is fixed. This change can be merged now.

PR #4539, opened 2019-03-28
WT-4668 lint

keithbostic:
@bvpvamsikrishna, can you please look at c29e5c1?

Just in case we should be doing something with the `ts` assignment.

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -655,12 +655,13 @@ __wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
 	 * Compare against the prepare timestamp of the current transaction.
 	 * Return an error if the given timestamp is older than the prepare
 	 * timestamp.
+	 *
+	 * If roundup timestamps is configured, the commit timestamp will be
+	 * rounded up to the prepare timestamp.
 	 */
-	if (F_ISSET(txn, WT_TXN_PREPARE) && ts < txn->prepare_timestamp) {
-		/*
-		 * If roundup timestamps is configured, the commit timestamp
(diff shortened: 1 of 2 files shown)


src/txn/txn_timestamp.c
```diff
@@ -655,22 +655,18 @@ __wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
...
+	 * If roundup timestamps is configured, the commit timestamp will be
+	 * rounded up to the prepare timestamp.
 	 */
-	if (F_ISSET(txn, WT_TXN_PREPARE) && ts < txn->prepare_timestamp) {
-		/*
-		 * If roundup timestamps is configured, the commit timestamp
-		 * will be rounded up to the prepare timestamp.
-		 */
-		if (F_ISSET(txn, WT_TXN_TS_ROUND_PREPARED))
-			ts = txn->prepare_timestamp;
```
bvpvamsikrishna: @keithbostic, is not the statement ```ts = txn->prepare_timestamp``` missing in code change ?.

keithbostic: Yes, that is the heart of the change. As far as I can tell, the assignment was never used.

bvpvamsikrishna: @keithbostic, it is used in the caller. without that assignment, the ```ts``` holds the specified ```commit_timestamp```. with the ```WT_TXN_TS_ROUND_PREPARED``` option, if the ```commit_timestamp``` is less than ```prepare_timestamp```, commit timestamp will be rounded to the prepare timestamp

bvpvamsikrishna: @keithbostic, I have not checked that ```ts``` is passed by value. the original code has a bug.

PR #4540, opened 2019-03-29
WT-4623 Remove barriers from transaction ID allocation.

michaelcahill:
Also try yielding after spinning, evidence suggests this can help on some single-socket servers.

### DIFF
--- a/src/include/txn.h
+++ b/src/include/txn.h
@@ -129,8 +129,6 @@ struct __wt_txn_global {
 	bool oldest_is_pinned;
 	bool stable_is_pinned;
 
-	WT_SPINLOCK id_lock;
-
 	/* Protects the active transaction states. */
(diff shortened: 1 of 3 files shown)


michaelcahill:
Somewhat experimental: performance tests so far haven't shown dramatic improvements over WT-4571.

bvpvamsikrishna:
@luke-pearson, please run the performance test, which showed regression and you narrowed the commit to txn id change on this branch.

luke-pearson:
My current recommendation is to keep the changes in WT-4571, I can elaborate on this but will be best explained in person.

michaelcahill:
After reviewing automated perf test results with @luke-pearson, I agree that the impact of this change is inconclusive and not worth keeping.  Closing.

PR #4544, opened 2019-04-01
WT-4579 Track the newest durable timestamp for each page.

keithbostic:
I'm still debugging, but this is the basic shape of the change I expect to merge (there are still some interesting cases around fast-delete I need to solve).


keithbostic:
autobuild: retest this please

src/reconcile/rec_write.c
```diff
@@ -4242,15 +4250,18 @@ __rec_col_var(WT_SESSION_IMPL *session,
 
 	/* Set the "last" values to cause failure if they're not set. */
 	last.value = r->last;
-	last.start_ts = last.stop_ts = WT_TS_NONE;
+	last.start_ts = WT_TS_MAX;
+	last.durable_ts = last.stop_ts = WT_TS_NONE;
```
agorrod: This change surprises me a bit. It's a change to the `start_ts`, and in other places `durable_ts` matches `start_ts`, not `stop_ts`. Similarly below.

agorrod:
@bvpvamsikrishna I've added you as a reviewer here. Looking through this change made me think that we need a pretty strong way to ensure that commit and durable are always set if we don't go down the path of WT-4550 (handling durable only when we need to, rather than using it as the default most places).

src/reconcile/rec_write.c
keithbostic: @agorrod, owed you a response:
> This change surprises me a bit. It's a change to the `start_ts`, and in other places `durable_ts` matches `start_ts`, not `stop_ts`. Similarly below.

This is unrelated to the rest of the changes. It's not a bug fix, either, it's just adding some measure of safety. The reason is so we (hopefully) fail if there's a bug and `last.start_ts` isn't set appropriately before being used, that is, a start timestamp of `WT_TS_NONE` is OK, one of `WT_TS_MAX` will fail.

keithbostic:
@agorrod, I think this one is ready for a real review.

michaelcahill:

PR #4546, opened 2019-04-02
WT-4693 Allow reconfigure of other settings with open transactions while downgraded.

sueloverso:


sueloverso:
@agorrod The automated scripts fixed whitespace in a few tests.

PR #4547, opened 2019-04-03
WT-4652 Return error if no durable timestamp for prepared txn.

bvpvamsikrishna:
1)__wt_txn_commit_timestamp_validate is currently overloaded. It is segregated as two:
          __wt_txn_commit_timestamp_validate_and_roundup
          __wt_txn_durable_timestamp_validate.
2) Return error if durable timestamp is not set for a prepared transaction. Also return error if durable timestamp is set for a non-prepared transaction.
3)Support setting the prepare timestamp through timestamp_transaction API.


bvpvamsikrishna:
retest this please

src/txn/txn.c
```diff
@@ -638,6 +638,7 @@ __wt_txn_release(WT_SESSION_IMPL *session)
 
 	/* Ensure the transaction flags are cleared on exit */
 	txn->flags = 0;
+	txn->prepare_timestamp = 0;
```
agorrod: Should this be WT_TS_NONE?

src/txn/txn.c
```diff
@@ -826,45 +827,52 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+		F_SET(txn, WT_TXN_HAS_TS_DURABLE);
+	}
+
+	if (prepare) {
+		if (!F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))
+			WT_ERR_MSG(session, EINVAL,
+			    "durable_timestamp is required for a "
+			    "prepared transaction");
+	} else {
+		if (F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))
```
agorrod: This could be an "else if"

src/txn/txn.c
```diff
@@ -826,45 +827,52 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
-		F_SET(txn, WT_TXN_HAS_TS_DURABLE);
 		txn->durable_timestamp = ts;
-		WT_ERR(__wt_txn_commit_timestamp_validate(
-		    session, "durable", ts, &cval, true));
+		WT_ERR(__wt_txn_durable_timestamp_validate(session));
+		F_SET(txn, WT_TXN_HAS_TS_DURABLE);
+	}
+
+	if (prepare) {
+		if (!F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))
```
agorrod: I believe this could be folded into the `if` above, if you make a change to the following else as well - and I think that would be an improvement.

src/txn/txn.c
```diff
@@ -826,45 +827,52 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+			WT_ERR_MSG(session, EINVAL,
+			    "durable_timestamp should not be specified for "
+			    "non-prepared transaction");
 	}
 
 	WT_ERR(__txn_commit_timestamps_assert(session));
 
+	if (prepare)
+		WT_ASSERT(session,
+		    !(txn->prepare_timestamp > txn->commit_timestamp));
```
agorrod: Could you change this to `<=` rather than not greater than? Similarly below.

src/txn/txn_timestamp.c
```diff
@@ -600,14 +600,10 @@ set:	__wt_writelock(session, &txn_global->rwlock);
...
- *	Validate a timestamp to be not older than running transaction commit
- *	timestamp and running transaction prepare timestamp. Validate a durable
- *	timestamp to be not older than the global oldest and global stable
- *	timestamp.
+ *	Validate the commit timestamp of a transaction.
  */
 int
-__wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
-    wt_timestamp_t ts, WT_CONFIG_ITEM *cval, bool durable_ts)
+__wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session)
```
agorrod: Given that this function is assigning values to the content of the `WT_TXN` structure, I think it should be renamed to reflect that it's more than `validate`

src/txn/txn_timestamp.c
```diff
@@ -630,60 +626,138 @@ __wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
...
 	}
+	return (0);
+}
+
+/*
+ * __wt_txn_durable_timestamp_validate --
+ *	Validate the durable timestamp of a transaction.
+ */
+int
+__wt_txn_durable_timestamp_validate(WT_SESSION_IMPL *session)
```
agorrod: I like that this is separated now.

src/txn/txn_timestamp.c
```diff
@@ -716,40 +790,32 @@ __wt_txn_set_timestamp(WT_SESSION_IMPL *session, const char *cfg[])
...
-		WT_RET_NOTFOUND_OK(ret);
-		if (ret == 0 && cval.len != 0) {
-			WT_TRET(__wt_txn_context_check(session, true));
-			WT_RET(__wt_txn_parse_timestamp(
-			    session, "durable", &ts, &cval));
-			txn->durable_timestamp = ts;
-			F_SET(txn, WT_TXN_HAS_TS_DURABLE);
-			prepare_allowed = true;
-		}
+	/* look for a durable timestamp. */
```
agorrod: You should capitalize `Look`

test/format/format.h
```diff
@@ -290,6 +290,8 @@ typedef struct {
 
 	WT_RAND_STATE rnd;			/* thread RNG state */
 
+	bool preparetxn;			/* is prepare transaction */
```
agorrod: Please add an `_` to make `prepare_txn`

src/txn/txn.c
```diff
@@ -826,43 +827,49 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
 
+	if (prepare) {
+		if (!F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))
+			WT_ERR_MSG(session, EINVAL,
+			    "durable_timestamp is required for a "
+			    "prepared transaction");
+		else
+			WT_ASSERT(session,
+			   txn->prepare_timestamp <= txn->commit_timestamp);
+
```
agorrod: This newline should be removed

src/txn/txn.c
```diff
@@ -826,43 +827,49 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		 * For prepared transactions commit timestamp could be earlier
 		 * than stable timestamp.
 		 */
-		WT_ERR(__wt_txn_commit_timestamp_validate(
```
agorrod: Could this function call `__wt_txn_set_timestamp` instead of parsing the timestamps out here and duplicating the logic? Let's talk about that tomorrow.

bvpvamsikrishna: Thank you @agorrod for quick review. I would like your suggestion on whether commit should return an error if prepare timestamp is set for a non-prepared transaction, or should we clear the prepare timestamp? I am noting down here so as not to forget to discuss this.

src/txn/txn_timestamp.c
```diff
@@ -630,166 +629,164 @@ __wt_txn_commit_timestamp_validate(WT_SESSION_IMPL *session, const char *name,
...
+				WT_RET_MSG(session, EINVAL,
+				    "commit timestamp %s is less than the "
+				    "prepare timestamp %s for this transaction",
+				    ts_string[0], ts_string[1]);
+			}
+			commit_ts = txn->prepare_timestamp;
+		}
 	}
-
+	txn->durable_timestamp = txn->commit_timestamp = commit_ts;
```
agorrod: Could you add a `WT_ASSERT` above the assignment that either `txn->durable_timestamp == WT_TS_NONE` or `txn->durable_timestamp == txn->commit_timestamp`?

i.e: that we've never explicitly set the durable timestamp, before overriding it.

src/txn/txn_timestamp.c
```diff
@@ -845,140 +842,182 @@ __wt_txn_parse_prepare_timestamp(
...
+	/* Look for a commit timestamp. */
+	ret = __wt_config_gets_def(session, cfg, "commit_timestamp", 0, &cval);
+	WT_RET_NOTFOUND_OK(ret);
+	if (ret == 0 && cval.len != 0) {
+		WT_RET(__wt_txn_parse_timestamp(session, "commit", &ts, &cval));
+		WT_RET(__wt_txn_set_commit_timestamp(session, ts));
+		__wt_txn_publish_commit_timestamp(session);
+	}
+
+	/* Look for a durable timestamp. */
```
agorrod: I'd like this comment to mention that it's important for it to come after setting the commit timestamp.

src/txn/txn.c
```diff
@@ -810,59 +818,43 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
-	    session, cfg, "durable_timestamp", 0, &cval));
-	if (cval.len != 0) {
-		if (!prepare)
+		if (!F_ISSET(txn, WT_TXN_HAS_TS_DURABLE))
+			WT_ERR_MSG(session, EINVAL,
+			    "durable_timestamp is required for a prepared "
+			    "transaction");
+
+		WT_ASSERT(session,
+			   txn->prepare_timestamp <= txn->commit_timestamp);
```
agorrod: The re-indenting doesn't look right here.

PR #4551, opened 2019-04-04
WT-4703 Extend test/checkpoint to do removes and online checking

michaelcahill:


test/checkpoint/workers.c
```diff
@@ -126,13 +126,22 @@ worker_op(WT_CURSOR *cursor, uint64_t keyno, u_int new_val)
...
-		if (ret == WT_ROLLBACK)
-			return (WT_ROLLBACK);
-		return (log_print_err("cursor.insert", ret, 1));
+	/* Roughly 5% removes. */
+	if (new_val % 19 == 0) {
+		if ((ret = cursor->remove(cursor)) != 0) {
+			if (ret == WT_ROLLBACK)
+				return (WT_ROLLBACK);
+			return (log_print_err("cursor.remove", ret, 1));
+		}
```
sueloverso: The removes are also new functionality. AFAICT they're not related to online consistency checking. Should this change be part of this ticket/commit or is this part of some other work you were doing?

agorrod: This change was attempting to reproduce an issue from MongoDB that involves inconsistency between collection and indexes, which involved removes. I'll update the ticket and title to reflect that.

agorrod:
I'm going to go ahead and merge this in - I'd like to be able to try the new test case against some different versions of WiredTiger, and merging makes that easier.

@sueloverso  if you'd like any follow on changes please open a new ticket.

PR #4552, opened 2019-04-05
WT-4705 column-store no longer needs to handle WT_COL page offsets of 0

keithbostic:


src/btree/bt_stat.c
```diff
@@ -160,21 +160,17 @@ __stat_page_col_var(
 	 * we see.
 	 */
 	WT_COL_FOREACH(page, cip, i) {
-		if ((cell = WT_COL_PTR(page, cip)) == NULL) {
+		cell = WT_COL_PTR(page, cip);
+		__wt_cell_unpack(session, page, cell, unpack);
+		if (unpack->type == WT_CELL_DEL)
 			orig_deleted = true;
-			++deleted_cnt;
```
michaelcahill: What happened to the `deleted_cnt` increment: the `orig_deleted` logic at line 183 and 191 seems to expect it...

keithbostic: I didn't look at that (all I did in this change was remove code that was never being executed).

I think the bug here is historic, and records of type `WT_CELL_DEL` should increment `deleted_cnt`, and didn't.

Additionally, records of type `WT_CELL_DEL` have associated RLE counts, and so we should increment `deleted_cnt` by any associated RLE value.

I've pushed a change to do that in 3edfd41, let me know what you think.

michaelcahill:

PR #4553, opened 2019-04-08
WT-4708 Coverity reported copy-paste error in WiredTiger error message

keithbostic:


bvpvamsikrishna:
@keithbostic, Thank you.

PR #4555, opened 2019-04-08
WT-4317 Remove portion of test not guaranteed to work with salvage.

sueloverso:
Also code movement to refactor zero'ing part of a file into fs code.

### DIFF
--- a/src/os_common/os_fhandle.c
+++ b/src/os_common/os_fhandle.c
@@ -500,3 +500,53 @@ __wt_close_connection_close(WT_SESSION_IMPL *session)
 	} WT_TAILQ_SAFE_REMOVE_END
 	return (ret);
 }
+
+/*
+ * __wt_file_zero --
+ *	Zero out the file from offset for size bytes.
+ */
+int
+__wt_file_zero(WT_SESSION_IMPL *session,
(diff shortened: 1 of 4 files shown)


src/os_common/os_fhandle.c
```diff
@@ -500,3 +500,53 @@ __wt_close_connection_close(WT_SESSION_IMPL *session)
...
+	zerobuf = NULL;
+	bufsz = WT_MIN((uint64_t)size, WT_MEGABYTE);
+	/*
+	 * For now logging is the only type and statistic. This needs
+	 * updating if block manager decides to use this function.
+	 */
+	type = WT_THROTTLE_LOG;
+	WT_STAT_CONN_INCR(session, log_zero_fills);
+	WT_RET(__wt_scr_alloc(session, bufsz, &zerobuf));
+	memset(zerobuf->mem, 0, zerobuf->memsize);
```
keithbostic: We don't need to zero out the entire scratch buffer, we only need to zero out the chunk we'll write, and scratch buffers can be large. I don't know how much this function gets called, so no idea if that might make a difference.

sueloverso: We are allocating the size we will use. This function is rarely configured so I don't think it makes a difference.

PR #4557, opened 2019-04-09
WT-4598 Enable durable timestamp assertions.

bvpvamsikrishna:
Enabled durable timestamp assertions.


agorrod:
@bvpvamsikrishna I'd like an Evergreen patch build of the server test suite before merging this - could you start a patch build?

PR #4558, opened 2019-04-09
WT-4711 Dist tools should ignore generated file wiredtiger.py.

ddanderson:
Fixed error in fixlinks, called by pyfilter.

### DIFF
--- a/dist/s_docs
+++ b/dist/s_docs
@@ -146,7 +146,7 @@ EOF
 	}
 
 	# Add optional extras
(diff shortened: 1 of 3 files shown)


ddanderson:
The change to tools/fixlinks.py, and the fix in s_docs exposes an odd issue, I discuss it in the ticket, and I opened Wt-4713 to follow up.

ddanderson:
@lukech can you take a look at this?

lukech:
@ddanderson Thanks for fixing the errors. This branch solved the build issues I encountered yesterday. I'll merge.

PR #4559, opened 2019-04-10
WT-4704 Add statistic tracking oldest active read timestamp

sulabhM:


sulabhM:
@agorrod, I have also put this on WT-4704: For the purpose of this statistic, do we want to consider the checkpoint read timestamp as one of the active reads in the system?

sulabhM:
@agorrod , this is ready for review

dist/stat_data.py
```diff
@@ -586,6 +586,7 @@ def __init__(self, name, desc, flags=''):
     TxnStat('txn_snapshots_created', 'number of named snapshots created'),
     TxnStat('txn_snapshots_dropped', 'number of named snapshots dropped'),
     TxnStat('txn_sync', 'transaction sync calls'),
+    TxnStat('txn_timestamp_oldest_active_read', 'transaction oldest active read timestamp', 'no_clear,no_scale'),
```
agorrod: I'd prefer to name this `txn_pinned_timestamp_reader`, and have the description similar to `txn_pinned_timestamp_oldest` ie:
`transaction range of timestamps pinned by the oldest active read timestamp`

sulabhM:
@agorrod , added "txn_timestamp_oldest_active_read', 'transaction read timestamp of the oldest active reader". The description should put it along the 'transaction range of timestamps .. "  statistics.

PR #4562, opened 2019-04-12
WT-4699 Save a copy of column list for JSON dump.

ddanderson:
Added a test showing the original problem.

### DIFF
--- a/test/csuite/wt4699_json/main.c
+++ b/test/csuite/wt4699_json/main.c
@@ -1,0 +1,84 @@
+/*-
+ * Public Domain 2014-2019 MongoDB, Inc.
+ * Public Domain 2008-2014 WiredTiger, Inc.
+ *
+ * This is free and unencumbered software released into the public domain.
+ *
+ * Anyone is free to copy, modify, publish, use, compile, sell, or
+ * distribute this software, either in source code form or as a compiled
+ * binary, for any purpose, commercial or non-commercial, and by any
+ * means.
+ *
+ * In jurisdictions that recognize copyright laws, the author or authors
+ * of this software dedicate any and all copyright interest in the
+ * software to the public domain. We make this dedication for the benefit
+ * of the public at large and to the detriment of our heirs and
+ * successors. We intend this dedication to be an overt act of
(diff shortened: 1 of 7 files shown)


ddanderson:
@sueloverso, can you please review?

test/csuite/wt4699_json/main.c
```diff
@@ -0,0 +1,84 @@
...
+	    "key_format=i,value_format=ii,columns=(k,v0,v1)"));
+
+	/* Create the single record. */
+	testutil_check(
+	    session->open_cursor(session, opts->uri, NULL, NULL, &c));
+	c->set_key(c, 1);
+	c->set_value(c, 1, 1);
+	testutil_check(c->insert(c));
+	testutil_check(c->close(c));
+	strcpy(projection, opts->uri);
```
sueloverso: Some in-the-test comments about what you're setting up and then what you're doing would be helpful here.

test/csuite/wt4699_json/main.c
```diff
@@ -0,0 +1,84 @@
...
+	testutil_check(c->next(c));
+	/* Overwrite the projection, with not enough columns */
+	strcpy(projection, opts->uri);
+	strcat(projection, "(aaa,bbb)");
+	testutil_check(c->get_key(c, &jsonkey));
+	testutil_check(c->get_value(c, &jsonvalue));
+	testutil_assert(strstr(jsonvalue, "aaa") == NULL);
+	printf("KEY: %s\n", jsonkey);
+	printf("VALUE: %s\n", jsonvalue);
+	testutil_assert(c->next(c) == WT_NOTFOUND);
```
sueloverso: Is this the configuration parse error you refer to in the test description? It does not seem like it is as that would return EINVAL I'd expect but this is the only expected error.

ddanderson: @sueloverso, it's actually the call to get_value() that gets a parse error.  That's because in the JSON output we need to use the names of the columns.  These names come from the projection, which is re-parsed each time we generate a set of values in JSON.

Thanks, I've added comments.

PR #4564, opened 2019-04-12
WT-4698 Fixes for modify tests to work with Python3.

ddanderson:
- clear Python error code if we guess wrong on what type we are decoding
- fix malloc failure condition
- fixed modify and calc_modify tests to use bytes type for 'u'
- changed calc_modify test to test both 'u' and 'S' types.


ddanderson:
@agorrod, could you review this please?

agorrod:
Thanks @ddanderson lgtm

PR #4565, opened 2019-04-12
WT-4720 Coverity complaint #111697: handle tiny sizes in block manager reads

keithbostic:


src/block/block_read.c
```diff
@@ -252,6 +252,16 @@ __wt_block_read_off(WT_SESSION_IMPL *session, WT_BLOCK *block,
 	WT_RET(__wt_read(session, block->fh, offset, size, buf->mem));
 	buf->size = size;
 
+	/*
+	 * Ensure we don't read information that isn't there. It shouldn't ever
+	 * happen, but it's a cheap test.
+	 */
+	if (size < block->allocsize)
```
tetsuo-cpp: Could you please explain this check for my benefit? I can't make much sense of it.

From reading the surrounding code, my understanding is:
`size`: The size of the data that we want to read from the block. So we want to read beginning from `offset` bytes into the block until `offset + size` bytes into the block.
`block->allocsize`: The size of the block file on disk.

So how come we'd expect that our read (which I'd have thought is a portion of the full block) is larger than the block itself that we're reading from?

tetsuo-cpp:
@keithbostic 
In WiredTiger, what is the relationship between blocks and pages? I saw this line from the wiki.
> A component of WiredTiger called Block Manager divides the on-disk pages into smaller chunks called blocks, which then get written to the disk.

So it sounds to me like a page is written to disk over multiple blocks. Can a block have the end of one page and the beginning of a different one? And how come a page is broken up into smaller files?

src/block/block_read.c
keithbostic: In this code, I believe `size` is the size in bytes of the block that we're reading from the file.

I think that Coverity is complaining about us reading `size` bytes, then setting `blk` to an offset of about 30 bytes into the bytes we read, and looking at stuff, even if size could have been smaller than the offset. I'm not 100% sure of that, but I think that's the complaint.

I could either test size against `WT_PAGE_HEADER_SIZE` (which is what `WT_BLOCK_HEADER_REF` is using), or against `block.allocsize`, which is minimum size chunk we're allowed to allocate from this file. I couldn't decide which was better, so I just picked one, let me know if that makes things less clear.

keithbostic:
@tetsuo-cpp 
> In WiredTiger, what is the relationship between blocks and pages?

It's a little messy, probably as things moved between layers, over time.

Generally, I think of WiredTiger "blocks" as being a single, contiguous chunk of data laid out in the file, owned by the block manager. The btree layer reads/writes "blocks" to the block manager. When the btree layer reads a "block", it then adds other stuff around it (like `WT_UPDATE` lists), and that's a "page", the in-memory representation. Then, the btree reconciliation process takes the in-memory "page", splits it if necessary into multiple "blocks", which are written by the block manager.

> So it sounds to me like a page is written to disk over multiple blocks. Can a block have the end of one page and the beginning of a different one?

The life-cycle is there's an in-memory "page". In the simple case, it maps one-to-one to a block manager "block" which is a chunk of data in the file. If the page grows, though, it might need to be split, in which case, reconciliation will create two blocks from the page. The next time the page is read into memory, though, it will be two pages, not one. So, in general, it's a one-to-one mapping.

> And how come a page is broken up into smaller files?

Pages are broken up into smaller blocks because if a page gets too large, single point updates can trash your cache because you're reading and writing a huge chunk of data in service of a single update.

src/block/block_read.c
milkie: Yes, Coverity is noticing that you are reading bytes out of a file, and those bytes could technically be anything (say, file corruption or a malicious actor).  So the bytes need to be bounds-checked before using them to calculate addresses from which to read or write memory.

keithbostic:
> This is a pretty hot code path, is it not?

It's at most a cache miss, and we're about to do I/O which will overshadow anything we do here.

PR #4567, opened 2019-04-15
WT-4575 Review usages of __wt_errx vs __wt_err

tetsuo-cpp:
I did a quick review of places where we use `__wt_errx` and switched them over to `__wt_err` and supplied an error code when applicable.

### DIFF
--- a/src/os_common/os_fhandle.c
+++ b/src/os_common/os_fhandle.c
@@ -299,9 +299,9 @@ __handle_close(WT_SESSION_IMPL *session, WT_FH *fh, bool locked)
 	conn = S2C(session);
 
 	if (fh->ref != 0) {
-		__wt_errx(session,
-		    "Closing a file handle with open references: %s", fh->name);
 		WT_TRET(EBUSY);
+		__wt_err(session, ret,
+		    "Closing a file handle with open references: %s", fh->name);
 	}
 
 	/* Remove from the list. */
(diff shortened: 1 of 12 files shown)


src/os_win/os_fs.c
```diff
@@ -17,7 +17,7 @@
 			ret = __wt_map_windows_error(windows_error);	\
 			if (windows_error == ERROR_ACCESS_DENIED) {	\
 				if (__retry == 0)			\
-					__wt_errx(session,		\
+					__wt_err(session, ret,		\
```
sueloverso: This is the only one you've changed that I disagree with. This is an informational message that is printed on the first encounter of this and then we reset `ret` and retry a bunch of times. If the user eventually gets an access denied error, then an error will be returned to them after all the retries. This situation is often transient.

tetsuo-cpp: ca5b267

That makes sense. Thanks for catching that.

src/os_common/os_fhandle.c
```diff
@@ -299,9 +299,9 @@ __handle_close(WT_SESSION_IMPL *session, WT_FH *fh, bool locked)
 	conn = S2C(session);
 
 	if (fh->ref != 0) {
-		__wt_errx(session,
-		    "Closing a file handle with open references: %s", fh->name);
 		WT_TRET(EBUSY);
+		__wt_err(session, ret,
```
bvpvamsikrishna: This change is not required. 'ret' will still have '0'.

tetsuo-cpp: 5e0c171

As discussed, that line wasn't doing anything so I've removed it and changed the call back to `wt_errx`.

PR #4568, opened 2019-04-15
WT-4723 Split rec_write.c into different files

agorrod:
Move the structure definitions into reconcile.h
Add a new rec_visibility.c that contains rec_upd_select.


agorrod:
@keithbostic could you take a look at this and let me know what you think?

keithbostic:
@agorrod 
> could you take a look at this and let me know what you think?

This seems fine to me, and there is some other code we could move out of `rec_write.c`, too (the dictionary lookup code, for example).

The change I never really got working last time we looked at this, was to split column-store from row-store where they share some support routines. Maybe we should quit trying to solve the problem in one big change, and nibble away at it until we get it done.

I'd like to wait until WT-4642 is done, though.

agorrod:
I applied the locked label - we don't want to merge this until after WT-4642 is merged.

keithbostic:
Rather than merge develop into this branch, I'm going to reproduce the changes in a new branch, I think that looks easier and less likely to break.

PR #4571, opened 2019-04-16
WT-4732 Remove the bias from the eviction policy on the dhandle order

modithah:


wtbuild:

michaelcahill:
autobuild: ok to test

agorrod:
@sulabhM could you please help by getting this PR to pass testing?

sulabhM:
@modithah, thanks for making this change. At the moment, picking the next dhandle is simply walking the dhandle list. With this proposed change, we will end up walking the whole list again to the picked dhandle. We frequently get customer workloads with a lot of dhandles (as many as 400 - 500k), if we were to walk the list each time, it might be a lot of work.

Alternatively, dhandles are also kept in a hash table, where there are multiple buckets and each bucket has a list of dhandles. I am exploring the idea to randomly pick a bucket first and then to pick a dhandle randomly in that bucket.

Since I can't propose changes to your fork, I am experimenting with that change in the following pull request. I am still working on it, but I thought I will let you know and get your feedback:
https://github.com/wiredtiger/wiredtiger/pull/4575

modithah:
@sulabhM : I was doing some experiments with this. Maybe this is not the best idea. There is a possibility that the same dhandle to be picked twice and the queue will be filled by the same b-tree twice.

sulabhM:
@modithah , I looked at the change you suggested as part of the last commit, the one described as "The current target calculations resulted in having more targets than the available slots for trees that come in the beginning resulting no eviction of other trees that come later in the loop."

The change suggested is effectively replacing total_slots with remaining_slots,  `remaining_slots = max_entries - *slotp`. I think the idea of keeping `remaining_slots` constant for the eviction walk going over multiple dhandles is to make sure that we calculate slots per dhandle proportional to the cache being consumed by a particular table. If we use remaining_slots instead of the total, the calculated target will not be proportional to the cache used by that dhandle.

That said, the eviction server code is complicated and the reported issue needs further looking into. The results from your tests in https://github.com/wiredtiger/wiredtiger/pull/4575 are also interesting and need to be understood. I have filed another ticket WT-4802 to do these investigations. It will be helpful to separate this issue from choosing dhandles randomly so that we can make incremental progress.

I am going to close this pull request and work with you in improving eviction with WT-4802

PR #4572, opened 2019-04-16
WT-4712 Add diagnostic logging.

sueloverso:
Not yet ready for review.


sueloverso:
Here's a first cut at the diagnostic functionality. This ended up as basically alternative 4. I'll write up the path to get here in the ticket. Please review and offer any feedback, additions, etc.

sueloverso:
I couple things I want to note and specific feedback:
* This is only minimally tested. I turned on the logging diagnostics in `ex_sync` just for simple smoke testing. It needs real testing via `timestamp_abort` for a good start.
* Are there other things it would be useful to add to this new operation record? Recording the current time? Or do we care about the timestamp's origin (i.e. whether it was set via `timestamp_transaction` or `commit_transaction`)?
* Do we want to add in the durable and prepared timestamps? I think yes but I didn't add them now in case we needed to backport to 4.0 right away.

sueloverso:
This is ready for review. The only outstanding question is if anything else should be added to the operation record, like a timestamp. I've added testing and added `durable` and `prepare` timestamps in a separate commit so they're easy to undo if we wanted.

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: Retest this please

sueloverso:
While this is ready for review for general concepts, it is not ready to merge. If this is the path we want to take, this branch needs log versioning changes added for the format change before merging. I will work on getting those done for this branch.

sueloverso:
autobuild: retest this please

dist/api_data.py
```diff
@@ -476,6 +476,23 @@ def __ge__(self, other):
             above 0 configures periodic checkpoints''',
             min='0', max='100000'),
         ]),
+    Config('diagnostic', '', r'''
```
agorrod: I don't love the name {{diagnostic}} here - it overlaps with the HAVE_DIAGNOSTIC flag in an uncomfortable way. I'm not sure it's any better but maybe {{debug_mode}}?

src/conn/conn_api.c
```diff
@@ -1816,6 +1816,48 @@ __conn_single(WT_SESSION_IMPL *session, const char *cfg[])
...
+	txn_global = &conn->txn_global;
+
+	WT_RET(__wt_config_gets(session,
+	    cfg, "diagnostic.rollback_error", &cval));
+	txn_global->diag_rollback = (uint64_t)cval.val;
+
+	WT_RET(__wt_config_gets(session,
+	    cfg, "diagnostic.table_logging", &cval));
+	if (cval.val)
+		FLD_SET(conn->log_flags, WT_CONN_LOG_DIAGNOSTICS);
```
agorrod: This flag seems a bit generic to me - I can think of lots of things diagnostic log might enable - maybe this will become not relevant if you make a name change suggestion I made above. I'm going to hold off on further review for now.

dist/api_data.py
sueloverso: I can live with `debug_mode` or something similar. I am mildly concerned it may be confusing with the `debug_info` API. But maybe not. Perhaps it fits into a "family" of debugging related things even if they're config strings versus method names. I will make the naming change and see how it looks.

src/conn/conn_api.c
sueloverso: I will think on a more specific name for this flag and its effects.

sueloverso:
@agorrod, @ddanderson, @michaelcahill  and @dgottlieb this is ready for final review.

@dgottlieb please look at the API change and see if it seems reasonable to you as a user in MongoDB.
@ddanderson please look at the tests, in particular the changes to allow wildcard patterns in `wttest.py` and any copy/paste errors between the 3 new tests. Of course you can fully review the entire change also!

Additional considerations: 
* This is a log format change without bumping the log format version since the change is only in debugging mode. That was intentional but review carefully to make sure it is correct.
* I overload `WT_LOGOP_IGNORE` to be used both in an operation type and in the fileid of a known operation type.

dist/api_data.py
```diff
@@ -476,6 +476,23 @@ def __ge__(self, other):
...
+            of checkpoints. Zero or one means perform normal archiving.''',
+            min='0', max='1024'),
+        Config('rollback_error', '0', r'''
+            return a WT_ROLLBACK error from a transaction operation about
+            every Nth operation to simulate a collision''',
+            min='0', max='10M'),
+        Config('table_logging', 'false', r'''
+            if true, write transaction related information to the log for all
+            operations, even operations for tables with logging turned off.
+            These operations are informational and skipped in recovery''',
```
agorrod: I'm inclined to be paranoid and mention that enabling this changes the log file format and breaks recovery with older versions of WiredTiger.

agorrod:
I wasn't clear: if you are happy to make the suggested API change, I'm happy for you to go ahead and merge.

test/suite/test_debug_mode02.py
```diff
@@ -0,0 +1,109 @@
...
+    def conn_config(self):
+        return 'log=(enabled=true,file_max=100K),debug_mode=(checkpoint_retention=%d)' % self.retain
+
+    def check_archive(self, cur):
+        cur_set = set(cur)
+        for i in range(1,90):
+            # Sleep and then see if archive ran. We do this in a loop
+            # for slow machines. Max out at 90 seconds.
+            time.sleep(2.0)
+            new_logs = fnmatch.filter(os.listdir(self.home), "*gerLog*")
```
ddanderson: This `fnmatch` idiom is used several times in this file, could be a method.

test/suite/test_debug_mode02.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def test_checkpoint_retain(self):
+        self.session.create(self.uri, 'key_format=i,value_format=i')
+        # No log files should be archived while we have fewer than the
+        # retention number of logs. Make sure each iteration the new
+        # logs are a proper superset of the previous time.
+        for i in range(1, self.retain):
+            cur_logs = fnmatch.filter(os.listdir(self.home), "*gerLog*")
+            self.advance_log_checkpoint()
+            # We don't accomodate slow machines here because we don't expect
```
ddanderson: If some of these tests run a long time we might want to only run them occasionally.  They can be tagged with this before the `def` of a `test_...` function:
```
    @wttest.longtest("Checkpoint retain test waits to create lots of log files")
```

PR #4573, opened 2019-04-17
WT-4450 Implementing python test for timestamping updates in the future

luke-pearson:


test/suite/test_timestamp17.py
```diff
@@ -0,0 +1,172 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+# OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+# ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+# test_timestamp17.py
+#   Test using an incorrect timestamp on an update and ensure behaviour
```
dgottlieb: To me, this statement makes sense. I'm not sure if others coming across this will know what it means to incorrectly timestamp things, and what the intended behavior is in those cases.

The test itself does shed light on this (which to me is more important), so it's up to you on whether you think this warrants changing.

dgottlieb:
I think this testing of the API is necessary and thorough. I still believe there is room for exercising WT code that keeps miscellaneous timestamp values on structures correct in the face of update chains modified by concurrent clients not being in increasing timestamp order (but otherwise satisfying snapshot isolation write-write conflict constraints). An example: https://github.com/wiredtiger/wiredtiger/blob/81305b4ade4b92bde247931334c99c32fb148c44/src/include/btmem.h#L239-L241

That other sort of testing is better suited for a different work item, or maybe we feel existing WT testing has that well covered at this point. If it turns out we want something, I'm happy filing a new ticket that's more prescriptive about what we want to accomplish.

test/suite/test_timestamp17.py
luke-pearson: I'll update the wording to make it a bit clearer

luke-pearson:
> 
> 
> I think this testing of the API is necessary and thorough. I still believe there is room for exercising WT code that keeps miscellaneous timestamp values on structures correct in the face of update chains modified by concurrent clients not being in increasing timestamp order (but otherwise satisfying snapshot isolation write-write conflict constraints). An example:
> 
> [wiredtiger/src/include/btmem.h](https://github.com/wiredtiger/wiredtiger/blob/81305b4ade4b92bde247931334c99c32fb148c44/src/include/btmem.h#L239-L241)
> 
> Lines 239 to 241 in [81305b4](/wiredtiger/wiredtiger/commit/81305b4ade4b92bde247931334c99c32fb148c44)
>  WT_DECL_TIMESTAMP(min_timestamp)/* Min timestamp in lookaside */ 
>  				/* Max timestamp on page */ 
>  WT_DECL_TIMESTAMP(onpage_timestamp) 
> 
> That other sort of testing is better suited for a different work item, or maybe we feel existing WT testing has that well covered at this point. If it turns out we want something, I'm happy filing a new ticket that's more prescriptive about what we want to accomplish.

@dgottlieb Could you please create a new jira for this with an example, thanks.

I'm merging this PR.

PR #4574, opened 2019-04-17
WT-4631 Always clear the read timestamp for transactions (#4528)

luke-pearson:
Set the read timestamp to WT_TS_NONE during transaction cleanup - to avoid reuse in the future.

(cherry picked from commit ecbe60ef4f5b4e48898ed54554471bf86a879aa7)


luke-pearson:
This backport wasn't straight forward but is largely the same as what I had committed originally.

PR #4575, opened 2019-04-17
WT-4732 Eviction walk should pick dhandles randomly

sulabhM:


sulabhM:
@autobuild retest this please

modithah:
@sulabhM Maybe its better to only pick random tree at the beginning as there is a possibility to pick the same tree twice and also the loop guarantees visiting all the trees opposed to the random walk. I have made some other changes on some of the calculations [4571](https://github.com/wiredtiger/wiredtiger/pull/4571) that would aid the eviction.

sulabhM:
@modithah, the problem with picking a dhandle randomly and then iterating is that we will end up favouring the dhandles at the end of the list much more than the ones in front of the list. I am trying to get the original change working.

Thanks for the suggestions in https://github.com/wiredtiger/wiredtiger/pull/4571, I will look at it

sulabhM:
@autobuild: retest this please

sulabhM:
@agorrod , @modithah ,

I think I was able to identify all the areas of code where picking up dhandles randomly was causing issues. Pull request tests are passing on this change now. I also ran perf tests on this change and do not see a regression.

There are probably further improvements needed before we can check-in, especially in the retry logic around scanning dhandles. But this would be a good time to have a look.

@modithah,
can you run tests on this change and check if this resolves the issue you were noticing in WT-4732.

modithah:
@sulabhM I ran the experiments on the issue and the issue seem to be mostly mitigated. But, in terms of stability, it can be better. (all graphs are average of 7 runs)

This was before,
image

This one is after
image

As you can see the index is no longer being evicted more. However, the ups and downs are somewhat hard to explain.

With the calculations that I mentioned in [4571](https://github.com/wiredtiger/wiredtiger/pull/4571) the result is as follows which i think is quite more stable.

image

sulabhM:
@modithah , thanks for running the tests. I am taking a look at your suggested improvement in https://github.com/wiredtiger/wiredtiger/pull/4571

src/evict/evict_lru.c
```diff
@@ -1346,6 +1346,60 @@ err:	WT_TRACK_OP_END(session);
...
+	WT_CONNECTION_IMPL *conn;
+	WT_DATA_HANDLE *dhandle;
+	u_int dh_bucket_count, rnd_bucket, rnd_dh;
+
+	conn = S2C(session);
+	*dhandle_p = NULL;
+
+	WT_ASSERT(session, __wt_rwlock_islocked(session, &conn->dhandle_lock));
+
+	// Nothing to do if the dhandle list is empty
```
agorrod: C++ style comment here. More comments below.

sulabhM:
@agorrod , I ran sys-perf and did not see any regressions. Please proceed with the review.

sulabhM:
@agorrod , please have a final look.

PR #4576, opened 2019-04-18
WT-4697 Fixes to workgen and runner scripts to work with Python3

ddanderson:
- fix integer division, print, and exception syntax
- fix "FrozenClass" implementation for Python3
- added comments for usage of extensions_config.
- in runner scripts, use new property sample_interval_ms of Workload
  (corresponding to a previous workgen change).


ddanderson:
@agorrod, can you please review?  I've made sure that all the runner scripts run, a few had stopped working before (with Python2).

agorrod:
@tetsuo-cpp and @lukech please review this change. If you are happy with it and comfortable to do so please go ahead and merge.

bench/workgen/runner/compress_ratio.py
```diff
@@ -121,7 +128,7 @@ def operations(optype, tables, key, value = None, ops_per_txn = 0, logtable = No
 workload = Workload(context, threads)
 workload.options.run_time = 60
 workload.options.report_interval = 1
-workload.options.sample_interval = 1
+workload.options.sample_interval_ms = 1000
```
tetsuo-cpp: I don't see any usages of this in the diff. Is that ok?

lukech: `workgen.cxx` and `wtperf.py` have code to check the `sample_interval_ms` value. And in below code segement ms and sec intervals are used together to control the sleep time:
>     // The whole and fractional part of sample_interval are separated,
>     // we don't want to sleep longer than a second.
>     int sample_secs = ms_to_sec(options->sample_interval_ms);
>     useconds_t sample_usecs =
>       ms_to_us(options->sample_interval_ms) - sec_to_us(sample_secs);
Not sure if the change here makes any logic difference though.

bench/workgen/workgen.swig
```diff
@@ -105,11 +105,10 @@ import numbers
 %extend workgen::classname {
 %pythoncode %{
     def __setattr__(self, attr, val):
-        if getattr(self, attr) == None:
+        if attr != 'this' and getattr(self, attr) == None:
             raise AttributeError("'" + #classname +
               "' object has no attribute '" + attr + "'")
-        f = _workgen.__dict__[#classname + '_' + attr + '_set']
-        f(self, val)
+        object.__setattr__(self, attr, val)
```
lukech: Is this change for code refactoring?

bench/workgen/runner/compress_ratio.py
ddanderson: Thanks @tetsuo-cpp and @lukech, you are right, the change to use sample_interval_ms is a no-op.  I mistakenly thought I had deprecated sample_interval earlier, I hadn't.  So I'll change back to use sample_interval for clarity.

bench/workgen/workgen.swig
ddanderson: @lukech, this change is a clean up or clarification.  SWIG generates a function named something like `_workgen.ThreadOptions_throttle_set` to set the `throttle` attribute of `ThreadOptions`.  But that function always just sets the attribute.  I hadn't realized that previously, and once I did, I made this change.  Rather than look up the function by name and calling it, I just do what it does: set the attribute.

lukech: Got it, thanks!

PR #4577, opened 2019-04-19
WT-4731 Add new retries to the checkpoint integrity test.

ddanderson:
If the test fails to find a range to run a valid test,
issue a warning, flag it as a soft (retryable) error, and rerun the
entire test so that a new calibration for the range is done.
Only if that process fails repeated efforts do we abort the test.


ddanderson:

PR #4581, opened 2019-04-23
WT-4706 Add a statistic to track the lookaside table size

tetsuo-cpp:
This PR is adding a new `lookaside_bytes` connection level statistic to track the size of the lookaside file.
I do this by inspecting the underlying block manager's `size`.


src/cache/cache_las.c
```diff
@@ -787,6 +788,11 @@ __wt_las_insert_block(WT_CURSOR *cursor,
 		__las_insert_block_verbose(session, btree, multi);
 	}
 
+	las_btree = S2BT_SAFE(session);
+	if (las_btree != NULL)
+		WT_STAT_CONN_SET(session, lookaside_bytes,
+		    las_btree->bm->block->size);
```
tetsuo-cpp: There is a parameter called `btree` in this function. To my understanding, this BTree is associated with the table owning the page that is being evicted and written to the lookaside file rather than the actual lookaside file itself. So I think setting the statistic to `btree->bm->block->size` would be wrong.

Instead, I've taken the BTree associated with `connection` which comes from the lookaside cursor which seems like it should give me the right value.

src/cache/cache_las.c
```diff
@@ -787,6 +788,11 @@ __wt_las_insert_block(WT_CURSOR *cursor,
 		__las_insert_block_verbose(session, btree, multi);
 	}
 
+	las_btree = S2BT_SAFE(session);
```
tetsuo-cpp: I've had to use `S2BT_SAFE` here because the `dhandle` member of session is `NULL` in a few of the Python suite tests. I wasn't sure whether this was ok/expected so I figured I'd just open this up and gauge reactions before investigating further. I've tried moving this block before the `err` tag to check whether it's the cases where we jump with `WT_ERR` and fail to write to the LAS that is causing this but that doesn't seem to be the what's happening.

sueloverso:
There is a lot of overlap between this and the design for `Mitigate operational issue with pinned history`. There are changes needed for that, that could be moved forward into this change. That design requires getting the file size exactly like this PR requires.

dist/stat_data.py
```diff
@@ -158,6 +158,7 @@ def __init__(self, name, desc, flags=''):
     ConnStat('cond_wait', 'pthread mutex condition wait calls'),
     ConnStat('file_open', 'files currently open', 'no_clear,no_scale'),
     ConnStat('fsync_io', 'total fsync I/Os'),
+    ConnStat('lookaside_bytes', 'bytes in lookaside file'),
```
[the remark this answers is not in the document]
sueloverso: There already exists a very similar statistics `cache_bytes_lookaside`. I suggest calling this one `CacheStat('cache_lookaside_ondisk'...)` and put it with the other `cache_lookaside_*` related statistics in that section.

src/cache/cache_las.c
tetsuo-cpp: 8150be1

Ok done. I've also wrapped the `__wt_block_manager_named_size` call with `WT_RET`. Not 100% sure whether it warrants returning non-zero since it's just for incrementing a statistic. But I imagine if this doesn't work then something has gone horribly wrong.

If it's easy to explain, what are the layering violations you're talking about? I see you mentioned something similar in a comment on the technical design doc for "mitigate operational issues with pinned history". I had a talk to Vamsi about this before writing this code and we thought that it'd be preferable to get this info from the block manager if it already exists instead of going to the operating system for it.

src/cache/cache_las.c
tetsuo-cpp: 8150be1

dist/stat_data.py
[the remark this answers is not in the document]
tetsuo-cpp: 8150be1

Renamed this statistic to `cache_lookaside_ondisk`.

src/cache/cache_las.c
keithbostic: The block manager is the layer between the access methods and the OS.

It should be possible to replace block managers — for example with an NVM friendly engine.

src/cache/cache_las.c
```diff
@@ -787,6 +788,9 @@ __wt_las_insert_block(WT_CURSOR *cursor,
 		__las_insert_block_verbose(session, btree, multi);
 	}
 
+	WT_RET(__wt_block_manager_named_size(session, WT_LAS_FILE, &las_size));
+	WT_STAT_CONN_SET(session, cache_lookaside_ondisk, las_size);
+
```
sueloverso: Do we want to do this in the case of an error? I think we probably do not. I'd move it above the `err` label and change the `WT_RET` to `WT_ERR`.

dist/stat_data.py
```diff
@@ -274,6 +274,7 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_lookaside_cursor_wait_internal', 'cache overflow cursor internal thread wait time (usecs)'),
     CacheStat('cache_lookaside_entries', 'cache overflow table entries', 'no_clear,no_scale'),
     CacheStat('cache_lookaside_insert', 'cache overflow table insert calls'),
+    CacheStat('cache_lookaside_ondisk', 'the ondisk size of the cache overflow table'),
```
sueloverso: We want to keep the earlier text similar so it appears in t2 and such together with other similar items. I suggest changing the description to `cache overflow table on-disk size`.

tetsuo-cpp: 1573b8f

Ah, so it gets sorted by name then by description.
Thanks, I'll note this in the doc also.

src/cache/cache_las.c
tetsuo-cpp: 1573b8f

Makes sense. I've moved it down and used `WT_ERR`.

tetsuo-cpp:
@sueloverso 
> There is a lot of overlap between this and the design for `Mitigate operational issue with pinned history`. There are changes needed for that, that could be moved forward into this change. That design requires getting the file size exactly like this PR requires.

I had a look at the technical design again. I think it should be easy to use the lookaside file size for this purpose.

The only thing is that you may want to have is to get the lookaside file size and do the check after every `cursor->update` rather than right at the end when we're finished updating like I'm doing now. But it should be easy to change either way.

src/cache/cache_las.c
tetsuo-cpp: @keithbostic 
Ah, understood. So "layering violation" refers to relying on some detail that should be abstracted away at that layer.

PR #4582, opened 2019-04-23
 WT-4647 Avoid splitting words when wrapping lines in wiredtiger.in

ddanderson:
To fit lines within 80 chars, we allow line splits after a set of
non-breaking space indicators.


ddanderson:

lukech:
@ddanderson Thanks for pushing the change!

PR #4584, opened 2019-04-24
WT-4712 A version of diagnostics that only uses log_printf.

sueloverso:
@dgottlieb @agorrod and @michaelcahill Here's a different approach that uses only log printf.

It does suffer from the issues I mentioned in the ticket about using this approach in that all the individual operations are sprinkled throughout the log. I did add an in-message LSN to allow manual walking backward, but it is not easy.

However, this does not change or introduce any new log records. So it is fully compatible to backport to 4.0 or any earlier release. Let me know what you think.


michaelcahill:
@sueloverso, you seem to be relying on this version for diagnosing issues.  Should we go ahead and merge it into develop (and throw away the other approach)?

sueloverso:
@michaelcahill actually I've been primarily using the other branch that contains the log records in the log for the other debugging. I first tried this branch, and it is okay. But seeing the actual transactions, as the other branch does, was *far* more useful. So we still have the pros/cons of both of these, sadly.

sueloverso:
Close in favor of PR #4572

PR #4585, opened 2019-04-26
WT-4582 Respect prepared updates in lookaside

michaelcahill:


michaelcahill:
@keithbostic, can you please try the repro against this change to see whether it fixes the bug?

src/cache/cache_las.c
```diff
@@ -424,22 +424,23 @@ __wt_las_page_skip_locked(WT_SESSION_IMPL *session, WT_REF *ref)
 	/*
 	 * Skip lookaside pages if reading as of a timestamp, we evicted new
 	 * versions of data and all the updates are in the past.
-	 */
-	if (ref->page_las->skew_newest &&
-	    txn->read_timestamp > ref->page_las->unstable_durable_timestamp)
-		return (true);
```
michaelcahill: This is the main fix here: include the timestamp from prepapred updates in the calculated `unstable_timestamp`.

src/reconcile/rec_write.c
```diff
@@ -1260,6 +1260,10 @@ __rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 		if (WT_TXNID_LT(max_txn, txnid))
 			max_txn = txnid;
 
+		/* Track the first update with non-zero timestamp. */
+		if (first_ts_upd == NULL && upd->start_ts != WT_TS_NONE)
+			first_ts_upd = upd;
```
michaelcahill: This is the main fix here: include the timestamp from prepapred updates in the calculated `unstable_timestamp`.

src/cache/cache_las.c
```diff
@@ -424,22 +424,23 @@ __wt_las_page_skip_locked(WT_SESSION_IMPL *session, WT_REF *ref)
...
-		 * unstable updates during checkpoint would end up reading more
-		 * content from lookaside than necessary.
-		 */
+	if (ref->page_las->skew_newest) {
+		if (WT_SESSION_IS_CHECKPOINT(session) &&
+		    txn->read_timestamp > ref->page_las->unstable_durable_timestamp)
+			return (true);
+
+		if (txn->read_timestamp > ref->page_las->unstable_timestamp)
+			return (true);
```
michaelcahill: Here is the other part of the fix: only checkpoint looks at `unstable_durable_timestamp` -- other reads respect `unstable_timestamp`, which (now) includes prepared timestamps.

keithbostic:
@michaelcahill, this change hits an assert in `__rec_split_write_supd()`:
```
(gdb) where
#0  0x00007f1c2ccb7c37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f1c2ccbb028 in __GI_abort () at abort.c:89
#2  0x0000000000457457 in __wt_abort (session=0x7f1c2de8c9f8)
    at src/os_common/os_abort.c:32
#3  0x00000000004745d0 in __rec_split_write_supd (session=0x7f1c2de8c9f8, 
    r=0x7f1bac4f8710, chunk=0x7f1bac4f8790, multi=0x7f1bac53cba0, 
    last_block=true) at src/reconcile/rec_write.c:3038
#4  0x0000000000474e52 in __rec_split_write (session=0x7f1c2de8c9f8, 
    r=0x7f1bac4f8710, chunk=0x7f1bac4f8790, compressed_image=0x0, 
    last_block=true) at src/reconcile/rec_write.c:3323
#5  0x0000000000474004 in __rec_split_finish (session=0x7f1c2de8c9f8, 
    r=0x7f1bac4f8710) at src/reconcile/rec_write.c:2916
#6  0x000000000047861f in __rec_col_var (session=0x7f1c2de8c9f8, 
    r=0x7f1bac4f8710, pageref=0xc75f60, salvage=0x0)
    at src/reconcile/rec_write.c:4782
#7  0x000000000046f2fa in __wt_reconcile (session=0x7f1c2de8c9f8, 
    ref=0xc75f60, salvage=0x0, flags=138, lookaside_retryp=0x0)
    at src/reconcile/rec_write.c:465
#8  0x0000000000447173 in __evict_review (session=0x7f1c2de8c9f8, 
    ref=0xc75f60, closing=false, inmem_splitp=0x7f1c2b9dcd58)
    at src/evict/evict_page.c:684
#9  0x000000000044584c in __wt_evict (session=0x7f1c2de8c9f8, ref=0xc75f60, 
    closing=false, previous_state=5) at src/evict/evict_page.c:174
#10 0x000000000044280a in __evict_page (session=0x7f1c2de8c9f8, 
    is_server=false) at src/evict/evict_lru.c:2320
#11 0x000000000043efc2 in __evict_lru_pages (session=0x7f1c2de8c9f8, 
    is_server=false) at src/evict/evict_lru.c:1173
#12 0x000000000043d1b5 in __wt_evict_thread_run (session=0x7f1c2de8c9f8, 
    thread=0xbacfa0) at src/evict/evict_lru.c:318
#13 0x00000000004b2de6 in __thread_run (arg=0xbacfa0)
    at src/support/thread_group.c:31
#14 0x00007f1c2d685184 in start_thread (arg=0x7f1c2b9dd700)
    at pthread_create.c:312
#15 0x00007f1c2cd7f03d in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
(gdb) frame 3
#3  0x00000000004745d0 in __rec_split_write_supd (session=0x7f1c2de8c9f8, 
    r=0x7f1bac4f8710, chunk=0x7f1bac4f8790, multi=0x7f1bac53cba0, 
    last_block=true) at src/reconcile/rec_write.c:3038
3038			WT_ASSERT(session, r->all_upd_prepare_in_prog == true ||
(gdb) l
3033			multi->page_las.max_txn = r->max_txn;
3034			multi->page_las.unstable_txn = r->unstable_txn;
3035			WT_ASSERT(session, r->unstable_txn != WT_TXN_NONE);
3036			multi->page_las.max_timestamp = r->max_timestamp;
3037	
3038			WT_ASSERT(session, r->all_upd_prepare_in_prog == true ||
3039			    r->unstable_durable_timestamp >= r->unstable_timestamp);
3040	
3041			multi->page_las.unstable_timestamp = r->unstable_timestamp;
3042			multi->page_las.unstable_durable_timestamp =
(gdb) p r->all_upd_prepare_in_prog
$1 = false
(gdb) p r->unstable_durable_timestamp
$2 = 0
(gdb) p r->unstable_timestamp
$3 = 337461
```

keithbostic:
@michaelcahill, the CONFIG that triggers that assert is column-store (I haven't yet seen a row-store failure with this change, but the WT-4582 failing row-store CONFIG normally takes a lot longer to fail).
```
#  RUN PARAMETERS
abort=0
alter=0
auto_throttle=1
backups=0
bitcnt=4
bloom=1
bloom_bit_count=12
bloom_hash_count=15
bloom_oldest=0
cache=94
cache_minimum=20
checkpoints=on
checkpoint_log_size=160
checkpoint_wait=38
checksum=uncompressed
chunk_size=4
compaction=0
compression=zlib
data_extend=0
data_source=table
delete_pct=4
dictionary=0
direct_io=0
encryption=none
evict_max=3
file_type=variable-length column-store
firstfit=0
huffman_key=0
huffman_value=0
independent_thread_rng=1
in_memory=0
insert_pct=10
internal_key_truncation=0
internal_page_max=17
isolation=random
key_gap=1
key_max=107
key_min=13
leaf_page_max=13
leak_memory=0
logging=0
logging_archive=1
logging_compression=zlib
logging_file_max=36901
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=4
memory_page_max=1
merge_max=9
mmap=1
modify_pct=32
ops=0
prefix_compression=1
prefix_compression_min=0
prepare=1
quiet=1
read_pct=48
rebalance=1
repeat_data_pct=35
reverse=0
rows=1000000
runs=1
salvage=1
split_pct=75
statistics=0
statistics_server=0
threads=20
timer=4
timing_stress_aggressive_sweep=0
timing_stress_checkpoint=0
timing_stress_lookaside_sweep=0
timing_stress_split_1=0
timing_stress_split_2=0
timing_stress_split_3=0
timing_stress_split_4=0
timing_stress_split_5=0
timing_stress_split_6=0
timing_stress_split_7=0
timing_stress_split_8=0
transaction_timestamps=1
transaction-frequency=65
truncate=1
value_max=884
value_min=8
verify=1
wiredtiger_config=
write_pct=6
```

michaelcahill:
Thanks @keithbostic, I'll try to reproduce.  That assertion is definitely wrong with the change as it stands because `unstable_timestamp` includes prepared updates but `unstable_durable_timestamp` doesn't.  It's probably better to include prepared updates in the calculation of `unstable_durable_timestamp` and maintain the relationship, so I'll try that.

michaelcahill:
I reproduced and looked more closely and I've pushed a different approach.  I think the only time it is safe to skip prepared updates in lookaside is if the read is guaranteed to be older than the prepare timestamp.  Otherwise, reading a prepared update should cause a `WT_PREPARE_CONFLICT` error until the prepared transaction is resolved.

keithbostic:
@michaelcahill, I'm not a great reviewer here, my understanding of chunks of this code is minimal to be kind. That said, this last set of changes passes the 3 different test configurations I'd seen fail, it runs correctly for me now in all cases.

src/cache/cache_las.c
```diff
@@ -422,30 +422,29 @@ __wt_las_page_skip_locked(WT_SESSION_IMPL *session, WT_REF *ref)
...
-		    ref->page_las->unstable_durable_timestamp)
+	if (ref->page_las->skew_newest) {
+		if (!ref->page_las->has_prepares &&
+		    txn->read_timestamp > (WT_SESSION_IS_CHECKPOINT(session) ?
+		    ref->page_las->unstable_durable_timestamp :
+		    ref->page_las->unstable_timestamp))
 			return (true);
-
-		if (txn->read_timestamp < ref->page_las->unstable_timestamp)
+	} else {
```
agorrod: This could be an `else if()`, rather than a nested `if`.

I'd prefer to save the result of this ternary into a local variable - since we are doing it twice, and I suspect we might need similar logic again.

src/reconcile/rec_write.c
```diff
@@ -1538,27 +1532,31 @@ __rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
...
-			 * Don't set the unstable durable timestamp with the
-			 * durable timestamp of an in-progress prepared update.
 			 * An in-progress prepared update will always have a
-			 * zero durable timestamp.
+			 * zero durable timestamp.  Checkpoints can only skip
+			 * reading lookaside history if all updates are in the
+			 * future, including the prepare, so including the
+			 * prepare timestamp instead.
 			 */
-			if (upd->prepare_state != WT_PREPARE_INPROGRESS &&
```
agorrod: How did you find this one? It'd be nice to have a repeatable process for finding this sort of thing..

src/cache/cache_las.c
michaelcahill: Done.

src/reconcile/rec_write.c
michaelcahill: I started down the path of including prepare timestamps in all of the "unstable" calculations, but then realized it isn't sufficient for `skew_newest` (because any later reader needs to get a prepare conflict).

I don't know any repeatable way to find these things other than increasing test coverage :-(

PR #4586, opened 2019-04-28
WT-4690 During checkpoints, make sure eviction does not split.

ddanderson:


ddanderson:
retest this please

ddanderson:
@keithbostic, can you please review this?  In my testing of several hours running the configuration both with and without the fix, I've seen the problem happen without the fix once, and never with the fix.

ddanderson:
To add some context: the problem occurs because reader threads doing tree walks are waiting for pages that are in `WT_REF_LOCKED`.  Meanwhile, checkpoint is trying to do a reverse split.  The checkpointing thread has been given a `WT_READ_NO_SPLIT` so it really shouldn't be splitting.

The fix involves pushing the flag further down into `__wt_page_release_evict` so it can be seen when `__wt_evict` is called.  `__wt_evict` already has a `closing` flag, which would avoid splitting, but that has some particular other semantics that didn't seem appropriate.  So this change modifies `__wt_evict` to have a flags parameter to track both the "closing" and "no split" conditions.

PR #4589, opened 2019-04-30
WT-4725 Fix miscellaneous scripts to work with Python3.

ddanderson:


ddanderson:
@agorrod, can you please review or delegate?

PR #4590, opened 2019-04-30
WT-4746 Document that transactions without timestamps clear history.

keithbostic:


src/docs/transactions.dox
```diff
@@ -182,6 +182,10 @@ single transaction.
 Setting a read timestamp in WT_SESSION::begin_transaction forces a transaction
 to run at snapshot isolation and ignore any commits with a newer timestamp.
 
+Committing an update without a timestamp truncates the update's timestamp
+history and limits repeatable reads: no earlier version of the update will
+be returned regardless of the setting of the read timestamp.
+
```
michaelcahill: This seems like a special case of "commits to a particular data item must be performed in timestamp order" -- maybe reorder, or add this as a follow up sentence to that?

keithbostic: Thanks.
I've pushed that in 994e8b5, and I'll merge/close.

PR #4591, opened 2019-04-30
WT-4688 Fix test and docs for correct log cursor usage.

sueloverso:
@ddanderson can you review the test changes.
@keithbostic can you review the doc changes. The doc heavily in this area 5 years ago by you and it appears it was not part of any PR or issue that I could find. Some of the content was not really accurate which I have tried to fix.


src/docs/cursor-log.dox
```diff
@@ -31,12 +31,16 @@ uint32_t
...
-one-to-one to a transaction (in other words, the returned log record has
-the only database operation the transaction made), the operation counter
-returned for the key will be zero.
+to a record that is not a transaction, the operation counter
+returned for the key will be zero. When the log record maps to a
+transaction, the first record returned, with an operation counter of
+zero, will be the entire record. Then the WT_CURSOR::next call will
+step into the transaction and return the first individual operation
+within that transaction. So, a transaction with a single operation will
+return two records related to that transaction.
```
keithbostic: I don't understand "a record that is not a transaction". Should this be "When a log record maps to a record that is not `part of` a transaction"?

Are we talking about records inserted outside of WiredTiger's transactional mechanisms? If an update is wrapped in a transaction by WiredTiger, rather than an explicit begin/commit in the application, then the log records are still part of a transaction. So, I'm confused.

keithbostic: Should we note the operation counter is then incremented for each individual record?

src/docs/cursor-log.dox
```diff
@@ -78,17 +82,20 @@ Here is an example of getting the log cursor's value:
...
-at 5.  The final return from the log cursor will again have a unique log
-ID, a unique transaction ID, and an operation counter of 0.
+log ID, no transaction ID, and an operation counter of 0.  The
+next six returns from the log cursor will have a common log ID, a
+common transaction ID, and operation counters starting at 0, returning
+the whole record and then starting at 1 and ending at 5 for each
+of the five individual operations. The next return from the log cursor
+will again have a unique log ID, a unique transaction ID, and an
+operation counter of 0. And the final return from the log cursor
+will have an operation counter of 1.
```
keithbostic: Not sure what "an internal operation" means?

keithbostic: I'd add a numbered list, that way each of the elements is explicitly listed, something like:
```
1. log ID=5, no transaction ID, operation counter=0, value of the internal operation
2. log ID=6, transaction ID=30, operation counter=0, all records in transaction 30
3. log ID=6, transaction ID=30, operation counter=1, first record in transaction 30
4. log ID=6, transaction ID=30, operation counter=2, second record in transaction 30
5. log ID=6, transaction ID=30, operation counter=3, third record in transaction 30
6. log ID=6, transaction ID=30, operation counter=4, fourth record in transaction 30
7. log ID=6, transaction ID=30, operation counter=5, fifth record in transaction 30
8. log ID=7, transaction ID=31, operation counter=0, all records in transaction 30
9. log ID=7, transaction ID=31, operation counter=1, first record in transaction 31
```

keithbostic: Another thing, is it possible to have `no transaction ID`? Doesn't that translate to a transaction ID of 0?

src/docs/cursor-log.dox
sueloverso: "a record that is not a transaction" refers to a log record that is not a transaction, not a key/value pair. So something like a log_printf message log record, or a checkpoint log record, etc. I will add the word `log` in front of each lone `record` for clarity.

src/docs/cursor-log.dox
sueloverso: Yes. A transaction id will be 0. I'll note that.

PR #4592, opened 2019-04-30
WT-4748 internal eviction flag handling could be simplified

keithbostic:
@michaelcahill, would you comment on whether or not  you like this change?

When I was reviewing Don's changes for WT-4690, I thought about simplifying our eviction flag handling and this is the result.

The root change is to not pass `WT_READ_XXX` flags to the `__wt_page_release()` and `__wt_page_release_evict()` functions, replacing the flag handling from there down through `__wt_evict() and `__evict_review()` with boolean arguments.

I can think of reasons this is a bad idea:

* We might want to pay attention to additional `WT_READ_XXX` flags in the eviction code in the future, as we have in the past, and we'd have to go back to passing the read flags through the `__wt_page_release` calls.
* If someone adds `WT_READ_NO_SPLIT` to a tree-walk flag set in the future, they can't rely on that automatically being passed through to eviction, they'll need to pay attention to the `__wt_page_release` calls, so this change might be an invitation to add a subtle bug.

There were two places we passed the `WT_READ_NO_EVICT` flags to `__wt_page_release()`, and I can't see that was having any effect at all.

Anyway, don't hesitate to discard this change without further discussion if this isn't a way you want to go.

@ddanderson, if Michael approves the change, can you please do the review?

In eb0e8b0 I replaced the two flags you added with two booleans -- I don't feel strongly about that change, feel free to discard that change if you prefer the flags.

### DIFF
--- a/src/btree/bt_compact.c
+++ b/src/btree/bt_compact.c
@@ -224,7 +224,7 @@ __wt_compact(WT_SESSION_IMPL *session)
 	}
 
 err:	if (ref != NULL)
-		WT_TRET(__wt_page_release(session, ref, 0));
+		WT_TRET(__wt_page_release(session, ref, false));
 
 	return (ret);
 }
--- a/src/btree/bt_delete.c
+++ b/src/btree/bt_delete.c
@@ -81,7 +81,7 @@ __wt_delete_page(WT_SESSION_IMPL *session, WT_REF *ref, bool *skipp)
 		}
 
 		(void)__wt_atomic_addv32(&S2BT(session)->evict_busy, 1);
-		ret = __wt_evict(session, ref, previous_state, 0);
+		ret = __wt_evict(session, ref, previous_state, false, false);
 		(void)__wt_atomic_subv32(&S2BT(session)->evict_busy, 1);
 		WT_RET_BUSY_OK(ret);
 		ret = 0;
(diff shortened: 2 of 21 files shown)


keithbostic:
autobuild: retest this please

michaelcahill:
@keithbostic, generally I'd prefer flags over booleans -- I do expect more in future.

Would a unified set of flags covering both reads and eviction address your concern?  I'm not really clear on what is being simplified.

keithbostic:
@michaelcahill 
> I'm not really clear on what is being simplified.

I should have articulated that better:

1. I find the current set of `WT_READ_XXX` flags confusing. The code in `__wt_page_in_func()` is not clear, and I'd rather those flags stayed out of eviction;
1. The only read flag eviction cares about is `WT_READ_NO_SPLIT`, and passing the entire set of read flags into eviction feels like overkill and an unnecessary complication;
1. Much of the code passing read flags into eviction isn't actually passing `WT_READ_NO_SPLIT`, so we're passing around flags we can know we'll never need and it makes it harder to understand the code;
1. I have a mild preference for booleans because they provide more information ("nosplit" vs. "flags").

I do think of these flags as "read" flags, and I don't see why we should pass read flags into eviction. And, page-release seemed like a natural place to limit their reach.

Anyway, I'm not going to fight for this change: if that argument changes your mind we can resurrect the branch.

PR #4593, opened 2019-05-01
WT-4745 Update upgrading doc for 3.2.0 WiredTiger release

lukech:
The changes of below tickets were identified to worth a description in the upgrading doc, which had been covered by the WT tickets. Just need to replace 3.1.1 with 3.2.0 since we decided to use 3.2.0 as the new release version. 
- WT-4192 Remove WiredTiger raw compression support
- WT-4427 Make WiredTiger timestamps always on and 8 bytes
- WT-4193 Make WT_CURSOR.modify require explicit transactions

Please note WT-4283 (together with WT-4156) introduced a new top level API (salvage), which did not exist in the previous WT release 3.1.0. Won't mention it in the upgrading doc, instead will cover it in the release note.

### DIFF
--- a/src/docs/upgrading.dox
+++ b/src/docs/upgrading.dox
@@ -1,12 +1,12 @@
 /*! @page upgrading Upgrading WiredTiger applications
 
 </dl><hr>
-@section version_311 Upgrading to Version 3.1.1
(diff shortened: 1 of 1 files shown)


lukech:
Since it's a simple change, I will go ahead to merge.

PR #4594, opened 2019-05-01
WT-4738 Tighten rules around use of timestamps.

michaelcahill:
* Check (in diagnostic mode, as per docs) that commit timestamps are after any active readers.
* Make the prepare check also diagnostic-only.
* Require snapshot isolation when timestamps are set (because lower isolation levels allow out-of-order commits to the same key).
* Update tests to comply with these rules.


michaelcahill:
I started looking into a `format` failure and noticed that it is doing some unexpected things, like setting commit timestamps out of order.  This happens because it uses timestamps with weaker isolation levels than snapshot, which is not guaranteed to give repeatable reads.

Mostly for discussion at this stage: I think tightening up the rules in this way makes sense, but it is worth talking through.

keithbostic:
autobuild: retest this please

src/txn/txn_timestamp.c
```diff
@@ -592,6 +592,64 @@ set:	__wt_writelock(session, &txn_global->rwlock);
...
+ *	Assert that commit and prepare timestamps are greater than the latest
+ *	active read timestamp, if any.
+ */
+static int
+__txn_assert_after_reads(
+    WT_SESSION_IMPL *session, const char *op, wt_timestamp_t ts, WT_TXN **prevp)
+{
+#ifdef HAVE_DIAGNOSTIC
+	WT_TXN *prev, *txn = &session->txn;
+	WT_TXN_GLOBAL *txn_global = &S2C(session)->txn_global;
```
sueloverso: Move the initializations out of the declarations please.

src/txn/txn_timestamp.c
```diff
@@ -769,7 +837,7 @@ __wt_txn_set_prepare_timestamp(
 {
 	WT_TXN *prev, *txn = &session->txn;
 	WT_TXN_GLOBAL *txn_global = &S2C(session)->txn_global;
```
sueloverso: These too, even though you didn't modify them.

test/suite/test_timestamp01.py
```diff
@@ -37,6 +37,8 @@ def timestamp_str(t):
     return '%x' % t
 
 class test_timestamp01(wttest.WiredTigerTestCase, suite_subprocess):
+    session_config = 'isolation=snapshot'
+
```
sueloverso: I think this test is the place to perhaps not make that change global in the test so that this test can add an error case of using timestamps without snapshot isolation.

agorrod:
@bvpvamsikrishna I've taken my name off this - if you are happy please go ahead and merge.

PR #4596, opened 2019-05-01
WT-4749 Fix error case for prepare_transaction

ddanderson:
- fixed reference to uninitialized variable.
- removed stray spaces from messages.


ddanderson:
@bvpvamsikrishna, can you please review? I hope you don't mind me jumping in here, I spotted the error while debugging failures with Python3. The change allows the prepare tests to run without error with Python3.

bvpvamsikrishna:
Thank you @ddanderson

PR #4597, opened 2019-05-01
WT-4723 Restructure the reconciliation code

keithbostic:
@agorrod, here's my attempt at splitting up the reconciliation code. The result:
```
 LOC     Name                    Function
 199     rec_dictionary.c        Leaf-page dictionary value lookup
 317     reconcile.h             Shared #defines, structures
 319     reconcile.i             Shared inline functions
 329     rec_child.c             Internal page child WT_REF handling
 500     rec_visibility.c        Update/visibility window code
 602     rec_track.c             Overflow tracking (unchanged)
1111     rec_row.c               Row-store reconciliation
1255     rec_col.c               Column-store reconciliation
2655     rec_write.c             Shared reconciliation code
                                    (mostly initialization, split management, tear-down)

7287 total  
```

The first 2 commits (175b8fb and 8cbafe7) are similar to what you did in the last branch, except I didn't put the `WT_CELL` information into `reconcile.h`, I put that work into a new file, `cell.h`. (I feel pretty strongly the cell information should not be included in `reconcile.h`, cells are verified and read as well as written, they aren't limited to reconciliation.)

The next commit (444d977), splits out the dictionary lookup code, I don't expect that to be controversial.

The next commit (056ffb8), is just some minor cleanup.

The final commit (f6a481c), is a big change, it splits the row- and column-store reconciliation out  into their own files.

It's arguably better than what we have -- there's no way we can argue these files are modular, but at least it's a step in a better direction.


keithbostic:
Everybody: if it's not obvious, these changes have a short shelf-life, it's going to be painful to work around them and merges are out of the question. I don't think there are any open PRs that are an issue right now, but that's not going to last. We should decide quickly if we want this change or not.

src/include/reconcile.h
```diff
@@ -0,0 +1,317 @@
...
+typedef struct {
+	WT_UPDATE *upd;			/* Update to write (or NULL) */
+
+					/* Transaction IDs, timestamps */
+	wt_timestamp_t start_ts, durable_ts, stop_ts;
+	uint64_t start_txn, stop_txn;
+
+	bool upd_saved;			/* Updates saved to list */
+
+} WT_UPDATE_SELECT;
```
sueloverso: Minor nit: I don't find having all that whitespace (blank lines) in the struct is more readable. I suggest getting rid of the 3 blank lines.

src/include/reconcile.i
```diff
@@ -0,0 +1,319 @@
...
+#define	WT_CROSSING_SPLIT_BND(r, next_len) ((next_len) > (r)->space_avail)
+#define	WT_CHECK_CROSSING_BND(r, next_len)				\
+	(WT_CROSSING_MIN_BND(r, next_len) || WT_CROSSING_SPLIT_BND(r, next_len))
+
+/*
+ * __rec_need_split --
+ *	Check whether adding some bytes to the page requires a split.
+ */
+static inline bool
+__rec_need_split(WT_RECONCILE *r, size_t len)
```
sueloverso: I would like to see all the functions in here prepended with `__wt_` since they're all called from many files. It at least gives an indication to the reader that it may be in a different file. If I see a call without the `__wt` prefix, I always look for it in the current file. (Like `txn.i` uses lots of `__wt` inlined functions.)

src/include/reconcile.h
keithbostic: Sure, pushed in d29ac8f.

src/include/reconcile.i
keithbostic: Sure, pushed in f9fb748.

PR #4599, opened 2019-05-02
WT-4750 Improve coverage of sweep, lookaside with consistency checking

michaelcahill:
Make worker threads periodically pause and close cursors so that sweep has a chance to run.  Sweep is supposed to only close handles after all of the changes are on disk and the history is no longer needed, so this should have no effect on consistency.


src/btree/bt_sync.c
```diff
@@ -267,13 +267,19 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+				    mod->rec_max_timestamp >
+				    btree->rec_max_timestamp)
 					btree->rec_max_timestamp =
 					    mod->rec_max_timestamp;
+				if (walk->page_las != NULL &&
+				    btree->rec_max_txn < walk->page_las->max_txn)
+					btree->rec_max_txn = walk->page_las->max_txn;
+				if (walk->page_las != NULL &&
+				    btree->rec_max_timestamp < walk->page_las->max_timestamp)
+					btree->rec_max_timestamp = walk->page_las->max_timestamp;
```
michaelcahill: Trying to find more places where version information in the tree needs to be rolled up during a checkpoint.  This is clearly not sufficient, because `__wt_tree_walk` will completely skip over some references (e.g., if they were deleted as of the checkpoint but may have lookaside history after that point).

src/cache/cache_las.c
```diff
@@ -717,6 +717,8 @@ __wt_las_insert_block(WT_CURSOR *cursor,
 				las_value.size = upd->size;
 				break;
 			case WT_UPDATE_BIRTHMARK:
+				WT_ASSERT(session, upd != first_upd || multi->page_las.skew_newest);
+				/* FALLTHROUGH */
```
michaelcahill: Just sanity checking that we only don't birthmarks at the beginning of a list when `skew_newest = false`, because the sweep server relies on this when discarding data from lookaside.  I have never seen this one fire.

src/evict/evict_lru.c
```diff
@@ -1980,7 +1980,7 @@ __evict_walk_tree(WT_SESSION_IMPL *session,
 		 * The goal here is that if trees become completely idle, we
 		 * eventually push them out of cache completely.
 		 */
-		if (WT_PAGE_IS_INTERNAL(page)) {
+		if (false && WT_PAGE_IS_INTERNAL(page)) {
```
michaelcahill: Mix it up: include internal pages regularly, and (with the `__evict_lru_cmp` change above), don't weight in favor of leaf pages.

src/include/btree.i
```diff
@@ -922,6 +922,7 @@ __wt_row_leaf_key_set_cell(WT_PAGE *page, WT_ROW *rip, WT_CELL *cell)
 	 */
 	v = WT_CELL_ENCODE_OFFSET(WT_PAGE_DISK_OFFSET(page, cell)) |
 	    WT_CELL_FLAG;
+	WT_ASSERT(NULL, WT_ROW_SLOT(page, rip) < page->entries);
```
michaelcahill: These were attempts to catch some segfaults.  Not relevant to the corruptions, plus they've never fired for me.

src/reconcile/rec_write.c
```diff
@@ -343,6 +343,35 @@ __rec_write_page_status(WT_SESSION_IMPL *session, WT_RECONCILE *r)
...
+			btree->rec_max_txn = mod->rec_max_txn;
+		if (btree->rec_max_timestamp < mod->rec_max_timestamp)
+			btree->rec_max_timestamp = mod->rec_max_timestamp;
+	}
+
+	if (WT_TXNID_LT(btree->tmp_rec_max_txn, mod->rec_max_txn))
+		btree->tmp_rec_max_txn = mod->rec_max_txn;
+	if (btree->tmp_rec_max_timestamp < mod->rec_max_timestamp)
+		btree->tmp_rec_max_timestamp = mod->rec_max_timestamp;
+
```
michaelcahill: I doubt this change is safe: it was an attempt to capture (modulo races) the largest txn ID / timestamp ever seen during reconciliation in a tree.  Without something like this, we're relying on checkpoint visiting everything to calculate the maximum, otherwise it isn't safe for sweep to discard a tree.

src/btree/bt_split.c
```diff
@@ -861,6 +861,10 @@ __split_parent(WT_SESSION_IMPL *session, WT_REF *ref, WT_REF **ref_new,
 			}
 		}
 
+		/* Check that we are not discarding active history. */
+		WT_ASSERT(session, (next_ref == ref && discard) ||
+		    !__wt_page_las_active(session, next_ref));
```
michaelcahill: This is a fixed version of the previous assertion I had here.  When a page is splitting, that's because it has been just been reconciled, so it's okay if there was an active `ref->page_las` -- we must have read that data into cache before reconciling.

Otherwise, we're splitting into a parent page and discarding a sibling that appeared empty to reconciliation but had active history in lookaside.

agorrod:
I'm going to close this PR, we have merged a different branch with a fix for this associated with WT-4750, and have split off WT-4793 for a version of the testing changes that are in this branch.

PR #4600, opened 2019-05-02
WT-4751 __wt_timestamp_to_string can return a pointer to simplify verbose message handling

keithbostic:


src/txn/txn_timestamp.c
```diff
@@ -12,12 +12,13 @@
...
  *	Convert a timestamp to the MongoDB string representation.
  */
-void
+char *
 __wt_timestamp_to_string(wt_timestamp_t ts, char *ts_string)
 {
 	WT_IGNORE_RET(__wt_snprintf(ts_string, WT_TS_INT_STRING_SIZE,
 	    "(%" PRIu32 ",%" PRIu32 ")",
 	    (uint32_t)((ts >> 32) & 0xffffffff), (uint32_t)(ts & 0xffffffff)));
+	return (ts_string);
```
sueloverso: It is a shame that we cannot get rid of the 2nd argument as it detracts from the readability, IMO. But we need it for memory management.

keithbostic: I agree, but I don't see a path to that either.

PR #4605, opened 2019-05-02
Add Michael's debugging to the log printf branch. Turn on diagnostic

sueloverso:
@michaelcahill Here's your branch merged against my diagnostic/log printf branch with the additional calls in where the btree fields are updated and diagnostic turned on in test/checkpoint.


sueloverso:
No longer needed

PR #4606, opened 2019-05-02
WT-4728 remove the stop timestamp from the WT_UPDATE structure

keithbostic:


keithbostic:
autobuild: retest this please

PR #4607, opened 2019-05-02
WT-4670 Remove support for WiredTiger LevelDB APIs

keithbostic:


src/docs/upgrading.dox
```diff
@@ -1,5 +1,16 @@
 /*! @page upgrading Upgrading WiredTiger applications
 
+</dl><hr>
+@section version_321 Upgrading to Version 3.2.1
```
lukech: Hi @keithbostic We are in the mid of WiredTiger 3.2.0 release (have not cut the release yet). I'd prefer to include this change as part of 3.2.0.

keithbostic: That's entirely up to you and @agorrod, let me know how I can assist.

lukech: I just pushed a commit to move the upgrading doc description for this ticket into 3.2.0 release section.

PR #4610, opened 2019-05-03
WT-4733 change format to do longer-term repeatable read testing

keithbostic:
@michaelcahill, I believe this branch has all of your changes from #4594.


keithbostic:
This branch, with this CONFIG, fails pretty quickly:
```
#assert_read_timestamp=1
assert_commit_timestamp=1
backups=0
cache_minimum=20
file_type=row
huffman_key=0
huffman_value=0
isolation=snapshot
rebalance=0
rebalance=1
rows=1000000
runs=100
salvage=0
salvage=1
timer=4
transaction_timestamps=1
verify=1
```
The failure is here:
```
(gdb) where
#0  0x00007ffd5a8fbc37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007ffd5a8ff028 in __GI_abort () at abort.c:89
#2  0x000000000045773a in __wt_abort (session=0x7ffd5bad2140)
    at src/os_common/os_abort.c:32
#3  0x00000000004b71d6 in __txn_commit_timestamps_assert (session=0x7ffd5bad2140)
    at src/txn/txn.c:756
#4  0x00000000004b75dd in __wt_txn_commit (session=0x7ffd5bad2140,
    cfg=0x7ffd537fdd50) at src/txn/txn.c:859
#5  0x0000000000498424 in __session_commit_transaction (
    wt_session=0x7ffd5bad2140, config=0x0) at src/session/session_api.c:1664
#6  0x00000000004090b9 in commit_transaction (tinfo=0x2910e40,
    session=0x7ffd5bad2140, prepared=false, commit_tsp=0x7ffd537fde98)
    at ops.c:795
#7  0x000000000040a6b9 in ops (arg=0x2910e40) at ops.c:1321
#8  0x00007ffd5b2c9184 in start_thread (arg=0x7ffd537fe700)
    at pthread_create.c:312
#9  0x00007ffd5a9c303d in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
(gdb) frame 3
#3  0x00000000004b71d6 in __txn_commit_timestamps_assert (session=0x7ffd5bad2140)
    at src/txn/txn.c:756
756                                     __wt_abort(session);
(gdb) p *op.u.op_row.upd
$11 = {
  txnid = 1047572,
  durable_ts = 0,
  start_ts = 0,
  stop_ts = 0,
  next = 0x7ffd3cd2f320,
  size = 8,
  type = 4 '\004',
  prepare_state = 0 '\000',
  data = 0x7ffd3c67480e "00005841"
}
(gdb) frame 6
#6  0x00000000004090b9 in commit_transaction (tinfo=0x2910e40,
    session=0x7ffd5bad2140, prepared=false, commit_tsp=0x7ffd537fde98)
    at ops.c:795
795             testutil_check(session->commit_transaction(session, NULL));
(gdb) list 763,796
763     commit_transaction(TINFO *tinfo,
764         WT_SESSION *session, bool prepared, uint64_t *commit_tsp)
765     {
766             uint64_t ts;
767             char buf[64];
768
769             if (commit_tsp != NULL)
770                     *commit_tsp = WT_TS_NONE;
771
772             ++tinfo->commit;
773
774             if (g.c_txn_timestamps) {
775                     /* Lock out the oldest timestamp update. */
776                     testutil_check(pthread_rwlock_wrlock(&g.ts_lock));
777
778                     ts = __wt_atomic_addv64(&g.timestamp, 1);
779                     testutil_check(__wt_snprintf(
780                         buf, sizeof(buf), "commit_timestamp=%" PRIx64, ts));
781                     testutil_check(session->timestamp_transaction(session, buf
));
782
783                     if (prepared) {
784                             testutil_check(__wt_snprintf(buf, sizeof(buf),
785                                 "durable_timestamp=%" PRIx64, ts));
786                             testutil_check(
787                                 session->timestamp_transaction(session, buf));
788                     }
789
790                     testutil_check(pthread_rwlock_unlock(&g.ts_lock));
791
792                     if (commit_tsp != NULL)
793                             *commit_tsp = ts;
794             }
795             testutil_check(session->commit_transaction(session, NULL));
796     }
(gdb) p g.c_txn_timestamps
$13 = 1
(gdb) p prepared
$14 = false
(gdb) p ts
$15 = 39782
```
It appears to me we're failing because there's an update without a start timestamp, but as far as I can tell, we're calling into this code from a format routine that has just set the commit timestamp for the update. I haven't yet gotten any further, but I'll keep digging next chance I get.

keithbostic:
@michaelcahill, @bvpvamsikrishna, I should flag one change in here: the current WiredTiger API outputs a verbose error message if a read timestamp is older than the oldest timestamp. This version of format does that on a regular basis, so I changed the error to not include a verbose message unless `WT_VERB_TIMESTAMP` is set.

That seems like a reasonable change to me (and I don't think it's out of the question that a normal application might attempt a read using a timestamp that has timed out of the system), but wanted to bring it to your attention.

keithbostic:
@michaelcahill, @bvpvamsikrishna, I think I understand this failure a little better.

We're in this code in `__txn_commit_timestamps_assert()`:
```
/*      
 * Skip over any aborted update structures or ones 
 * from our own transaction.
 */      
while (upd != NULL && (upd->txnid == WT_TXN_ABORTED ||
    upd->txnid == txn->id))
        upd = upd->next;

/*      
 * Check the timestamp on this update with the
 * first valid update in the chain. They're in
 * most recent order.
 */      
if (upd == NULL)
        continue;
/*      
 * Check for consistent per-key timestamp usage.
 * If timestamps are or are not used originally then
 * they should be used the same way always. For this
 * transaction, timestamps are in use anytime the
 * commit timestamp is set. 
 * Check timestamps are used in order.
 */      
op_zero_ts = !F_ISSET(txn, WT_TXN_HAS_TS_COMMIT);
upd_zero_ts = upd->start_ts == WT_TS_NONE;
if (op_zero_ts != upd_zero_ts)
        WT_RET_MSG(session, EINVAL,                   <<<< FAILING HERE
            "per-key timestamps used inconsistently");
```
We've got an update chain where the update at the tail of the list's transaction ID doesn't match the update at the start of the list's transaction ID, and so we fall into the test for mismatching timestamps.

The update at the tail of the list is a tombstone with no transaction ID:
```
(gdb) p *txn.mod[4].u.op_row.upd
$37 = {
  txnid = 3876, 
  durable_ts = 0, 
  start_ts = 0, 
  next = 0x7f63a89297b0, 
  size = 20, 
  type = 4 '\004', 
  prepare_state = 0 '\000', 
  data = 0x7f63a8782426 "0000562697/LMNOPQRST"
}
(gdb) p *$37.next
$38 = {
  txnid = 0, 
  durable_ts = 0, 
  start_ts = 0, 
  next = 0x0, 
  size = 0, 
  type = 5 '\005', 
  prepare_state = 0 '\000', 
  data = 0x7f63a89297d6 ""
}
```
So, I'm pretty suspicious we're coming out of the `__rec_append_orig_value()` code:
```
/*      
 * We need the original on-page value for some reader: get a copy and
 * append it to the end of the update list with a transaction ID that 
 * guarantees its visibility.
 *       
 * If we don't have a value cell, it's an insert/append list key/value
 * pair which simply doesn't exist for some reader; place a deleted 
 * record at the end of the update list.
 */      
append = NULL;                  /* -Wconditional-uninitialized */
size = 0;                       /* -Wconditional-uninitialized */
if (unpack == NULL || unpack->type == WT_CELL_DEL)
        WT_RET(__wt_update_alloc(session,
            NULL, &append, &size, WT_UPDATE_TOMBSTONE));
else {  
        WT_RET(__wt_scr_alloc(session, 0, &tmp)); 
        WT_ERR(__wt_page_cell_data_ref(session, page, unpack, tmp));
        WT_ERR(__wt_update_alloc(
            session, tmp, &append, &size, WT_UPDATE_STANDARD));
}       
```
which appends a tombstone with a transaction ID of 0.

Thoughts?

bvpvamsikrishna:
@keithbostic, may be we can modify `__txn_commit_timestamps_assert()` to handle this case?

keithbostic:
@bvpvamsikrishna 
> maybe we can modify `__txn_commit_timestamps_assert()` to handle this case?

That would be my inclination, but I don't know what that fix would look like, was hoping you or @michaelcahill did.

keithbostic:
@bvpvamsikrishna, @agorrod, I should also mention this ticket increases our timestamp test percentage from 10% to 70%, that is, we'll be doing far more timestamp runs in the future. That seemed right, given our increasing dependency on timestamps.

agorrod:
@keithbostic I'm conscious that this branch has stalled a bit.

Are there still known failures when these changes are enabled, or is it good to merge as far as you know?

keithbostic:
@agorrod 
> Are there still known failures when these changes are enabled, or is it good to merge as far as you know?

This branch can't be merged, there's a known issue with the `__txn_commit_timestamps_assert()` code that would need to be fixed for it to be merged.

This branch also includes some other testing changes which should be approved: `format` no longer uses timestamps except when also doing snapshot-isolation mode, it increases timestamp (vs. transaction ID), testing from 10% to 70%, and adds timestamp assertions on 5% of `format` runs.

This branch wasn't scheduled work and has repeatedly stalled, so feel free to kill it without further discussion.

In that case, please resurrect WT-4777, that's a case where `format` fails configuring timestamps with an isolation mode other than snapshot.

bvpvamsikrishna:
@keithbostic, I pushed a commit, with a trivial bug fix. With these changes, it is not crashing to me as quickly as it was before. I will run it longer and will check.

keithbostic:
@bvpvamsikrishna 
> With these changes, it is not crashing to me as quickly as it was before. I will run it longer and will check.

Thanks, Vamsi! That change looks reasonable to me (although I'm a poor reviewer for it).

keithbostic:
@bvpvamsikrishna, I merge develop into this branch since the merge of WT-4844 (#4669) included a different fix for the `read timestamp older than oldest timestamp` messages.

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest

keithbostic:
autobuild: retest this please

keithbostic:
@agorrod, I believe this branch is ready for review & merge.

@bvpvamsikrishna, I did not push the change you mentioned, wrapping the "bulk-load" code in a transaction when it's not actually using a bulk load cursor (when the reverse collator is specified). I've never seen a failure so I'm not sure if it's needed, or was just confusing? Please let me know, I'm happy to make that change.

This branch reworks significant chunks of the `test/format` program, here are list of the changes that might be worth review:

* This PR reworks @michaelcahill's changes from #4594, but the idea is that using timestamps requires snapshot-isolation and explicit transactions, we no longer mix-and-match non-timestamp or non-transactional operations with timestamp operations. Michael can comment on this, I'm just following his lead. The function to review is `test/format/config.c:config_transaction()`, to confirm the flow and requirements make sense.

* This PR adds `assert_commit_timestamp=true` and/or `assert_read_timestamp=true` configurations on 5% of the runs. Michael said something about one of them not making sense for `format`: Michael, if you can remind me which of those you didn't want, I'll remove it.

* This PR switches us from configuring 10% of our runs with `transaction_timestamps=true` to configuring 70% of our runs with timestamps. That's a big change, but since our MongoDB use has switched to timestamps, it seems like our testing should follow.

* This PR adds repeatable read testing, that is, when running with timestamps, after a transaction is resolved, we repeat some number of the operations in the transaction based on their read or commit timestamps. The code to review is `test/format/snap.c`, to confirm the rules around which operations will repeat are neither too loose or to restrictive, that is, that we're not skipping operations we can reasonably repeat, or attempting to repeat operations that might fail, and that we're using the correct timestamps for that purpose. There's also the code in `test/format/ops.c:begin_transaction_ts()`, where we select a read timestamp when we begin a timestamped transaction, , but I think you've already reviewed that.

test/format/bulk.c
```diff
@@ -71,16 +112,25 @@ wts_load(void)
 	key_gen_init(&key);
 	val_gen_init(&value);
 
+	if (g.c_txn_timestamps)
+		bulk_begin_transaction(session);
```
agorrod: I'm a little surprised you can do a bulk load in a timestamp transaction - since I don't think the timestamp will be applied, but I guess it won't hurt to have the transaction running.

PR #4611, opened 2019-05-05
WT-4756 Fix a comment in integer packing/unpacking code.

ddanderson:


ddanderson:
self-reviewed simple comment change.

PR #4612, opened 2019-05-06
WT-3929 Fix "pip install" to work with Python3.

ddanderson:
- avoid the use of relative imports, it doesn't work well with
  installed packages.
- introduce packutil.py, to remove a circular dependency that causes
  trouble with installed packages.
- minor changes to setup_pip.py for Python3.
- add a new library directory to check for standard compression libs.


ddanderson:
Not ready for merge yet, needs to be tested using the process to create a Python release and then using "pip install" for both Python2 and Python3.

ddanderson:
@agorrod, this is ready for review.  I put some information about how I tested it in the ticket.

agorrod:
Thanks @ddanderson lgtm

PR #4616, opened 2019-05-07
Debug against log format branch

sueloverso:
@michaelcahill and @dgottlieb this branch is (clearly) not for merge. But I applied Michael's changes to the 4712 branch with the log format change so that it is easier to see commits to non-logged tables and added some log printf calls to see where they fall in the stream of operations. There are a few changes to `test/checkpoint` as well:
* I've made the key format always 'S' for easier printlog viewing.
* I modified the test `verify` function to keep going after it hits a mismatch error so that we can see all of the errors not just the first one.

### DIFF
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -1409,6 +1409,19 @@ err:	if (parent != NULL)
 	return (0);
 }
 
+static void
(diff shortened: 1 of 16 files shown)


sueloverso:
No longer needed.

PR #4617, opened 2019-05-07
WT-4765 Remove support for Helium/Levyx data sources in WiredTiger.

keithbostic:


agorrod:
Thanks @keithbostic lgtm

PR #4618, opened 2019-05-07
WT-4766 Make wt_ckpt_decode script work with both python2 and python3

ddanderson:
The change is needed because the WT unpack routine now takes a bytes object.  It was working previously for Python3 because a bytearray object behaves like a bytes object in Python3.  In Python2, they are not the same at all, but an explicit conversion will solve the problem.


ddanderson:
@agorrod, this is ready for review, just a one line change.  Thanks!

PR #4619, opened 2019-05-07
WT-4750 Sweep can remove active lookaside records when files are closed and re-opened

keithbostic:


keithbostic:
`test_timestamp04` fails pretty reliably with this branch... could we have somehow been depending on the previous behavior? I'll chase, but let me know if anybody has ideas.

michaelcahill:
@keithbostic, the rollback_to_stable changes here aren't correct -- the bitmap there needs to cover `las_id`.

keithbostic:
> the rollback_to_stable changes here aren't correct -- the bitmap there needs to cover las_id.

Thanks, Michael -- I'm not 100% sure of `conn->stable_rollback_maxfile`, but I think a straight-forward translation of `log_id` to `las_id` works.

keithbostic:
autobuild: retest this please

sueloverso:
I am pulling the `test/checkpoint` test changes to this branch and running the repro to see if it fixes most/all of the mismatches.

src/include/cache.h
```diff
@@ -202,14 +202,15 @@ struct __wt_cache {
 	uint64_t las_remove_count;      /* Count of removes from lookaside */
 	uint64_t las_pageid;		/* Lookaside table page ID counter */
 
-	bool las_reader;		/* Indicate an LAS reader to sweep */
+	bool las_reader;		/* Indicate a LAS reader to sweep */
```
sueloverso: In my head I say the letter names, not "lass". So in that case `an ELL` is correct.

src/btree/bt_handle.c
```diff
@@ -337,9 +337,12 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
...
 	}
 
-	/* Get the file ID. */
+	/* Get the logging/permanent file ID. */
 	WT_RET(__wt_config_gets(session, cfg, "id", &cval));
-	btree->id = (uint32_t)cval.val;
+	btree->log_id = (uint32_t)cval.val;
+
+	/* Set the temporary file ID. */
+	btree->las_id = __wt_atomic_add32(&conn->cache->las_id, 1);
```
sueloverso: Given the subtlety and magnitude of the problem, and this one line is the crux of the fix, it deserves a healthy comment.

michaelcahill:
Closing in favor of #4620.

PR #4620, opened 2019-05-07
WT-4750 Remove our id from the LAS dropped list on reopen if it's on it.

sueloverso:
@keithbostic and @michaelcahill for consideration.


michaelcahill:
Thanks @sueloverso (and @keithbostic and @dgottlieb and @agorrod).  After looking at both branches again and reading the discussion about their pros and cons, I'm going to merge this one.  In the end, keeping a single, fixed ID for each tree (and the associated benefit to future diagnostics) swayed me in this direction.  This is also closer to what I expect we will want in future when history becomes durable.

lgtm, I'll merge.

PR #4622, opened 2019-05-09
WT-4768 Diagnostics looking for page split bug

sueloverso:
@michaelcahill New pull request after renaming the branch for the new ticket number.


sueloverso:
@michaelcahill you should be able to just run the same test reproducer with this branch to see the same failures.

agorrod:
@sueloverso I think it can be closed now - since some of the diagnostic information has been merged into develop and we have a plan to pull the rest in as well.

PR #4623, opened 2019-05-10
WT-4769 Keep empty page references with active history.

michaelcahill:
It is only okay to discard a reference once its history is no longer required, even if reconciliation created an empty page image.

### DIFF
--- a/src/txn/txn_rollback_to_stable.c
+++ b/src/txn/txn_rollback_to_stable.c
@@ -248,15 +248,29 @@ __txn_abort_newer_updates(
 	 */
 	local_read = false;
 	read_flags = WT_READ_WONT_NEED;
-	if (ref->page_las != NULL && ref->page_las->skew_newest &&
-	    rollback_timestamp < ref->page_las->unstable_durable_timestamp) {
-		/* Make sure get back a page with history, not limbo page */
-		WT_ASSERT(session,
-		    !F_ISSET(&session->txn, WT_TXN_HAS_SNAPSHOT));
-		WT_RET(__wt_page_in(session, ref, read_flags));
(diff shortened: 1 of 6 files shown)


michaelcahill:
@sueloverso, I am running more tests, but this branch has prevented the new assertion in `bt_split.c` triggering.

I'm running with this change on top of #4622, but without the `file_manager` settings to `wiredtiger_open`.  So far that configuration hasn't hit any errors.  If that holds over the weekend, I think the remaining issues we're seeing are related to sweep.

src/txn/txn_rollback_to_stable.c
```diff
@@ -248,15 +248,29 @@ __txn_abort_newer_updates(
...
-	    rollback_timestamp < ref->page_las->unstable_durable_timestamp) {
-		/* Make sure get back a page with history, not limbo page */
-		WT_ASSERT(session,
-		    !F_ISSET(&session->txn, WT_TXN_HAS_SNAPSHOT));
-		WT_RET(__wt_page_in(session, ref, read_flags));
-		WT_ASSERT(session, ref->state != WT_REF_LIMBO &&
-		    ref->page != NULL && __wt_page_is_modified(ref->page));
-		local_read = true;
+	if (ref->page_las != NULL) {
+		if (ref->page_las->skew_newest && rollback_timestamp <
```
sueloverso: This is a fairly complicated set of conditionals. Someone (like me) not intimately familiar with this can be easily confused. A comment would be helpful here.

michaelcahill: I added more description in dfb95ae: does that help?

PR #4625, opened 2019-05-10
WT-4747 Change optrack scripts to run under Python3.

ddanderson:
- various simple syntax changes
- rewrote the color palette selection. In python3, colors from matplotlib
  are sorted by name alphabetically, which has some bad effects (lots of
  "light-" colors together, for example). Using numbered rather than
  named colors allows us to space color selections.


ddanderson:
retest this please

PR #4628, opened 2019-05-13
WT-4776 Modify operations should be equivalent to updates.

michaelcahill:
If modifies are created with out-of-order timestamps, WiredTiger should use apply all previous modifications until it finds a complete copy of the document. Once a modify is identified as the most recent visible version for a value WT should not apply normal visibility rules to the prior versions - they were visible when the modify was performed, so should be included in the read.


agorrod:
autobuild: retest this

agorrod:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

PR #4630, opened 2019-05-13
WT-4768 Save the id for sweep dropping in btree_close instead of eviction.

sueloverso:
With this fix, I do not get any data mismatch failures with the test case. Prior to this fix, I was getting 100% failure with 20 parallel processes.

@michaelcahill I'm a bit concerned this undoes some of the fixes from WT-3715 and dropping the LAS entries for a dropped file with `force=true` by moving it. I also considered some path that would remove or avoid saving the id with the sweep dead versus discard phases if we reopened, but that felt messy pretty quickly.

### DIFF
--- a/src/btree/bt_handle.c
+++ b/src/btree/bt_handle.c
@@ -270,6 +270,16 @@ __wt_btree_close(WT_SESSION_IMPL *session)
 	F_SET(btree, WT_BTREE_CLOSED);
 
 	/*
+	 * If closing a tree let sweep drop lookaside entries for it.
+	 */
+	if (F_ISSET(S2C(session), WT_CONN_LOOKASIDE_OPEN) &&
+	    btree->lookaside_entries) {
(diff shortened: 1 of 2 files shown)


sueloverso:
I think this is one downside of keeping the `id` the same for the btree as opposed to the alternative fix last week for WT-4750. We may not have hit this if the new dhandle had a new id.

sueloverso:
autobuild: retest this please

sueloverso:
autobuild: retest this please

michaelcahill:
@sueloverso, this change lgtm.  Where we were storing `btree->id` on close previously, the handle could be marked dead (and reopened) before the old ID was saved.  This way around makes sense to me and avoids a race.

On the question of whether we should switch to per-dhandle IDs (so reopening a btree would get a new ID), I agree that would have avoided this issue, but I still think keeping a unique ID across logging, metadata and lookaside is preferable (and better matches where things are going with durable history).

PR #4631, opened 2019-05-14
WT-4779 Disable an assert that commit timestamp is valid

agorrod:
We added an assertion that non-prepared transactions commit
timestamp is newer than the oldest, which triggered a test failure.


agorrod:
I'm going to go ahead and merge this - testing was clean, and @michaelcahill approved.

PR #4632, opened 2019-05-16
WT-4794 Mark lookaside history resolved in all read paths.

michaelcahill:
When the `resolved` flag was first added to `ref->page_las`, it was only being set in the common case where lookaside history was re-instantiated in memory.  This change sets it consistently in all (currently both) paths.


agorrod:
autobuild: retest this please

agorrod:
This has passed pretty much all the testing - I'm going to go ahead and merge.

PR #4633, opened 2019-05-16
WT-4796 Enhance diagnostics that track ref state transitions

agorrod:
Include atomic CAS transitions as well as setting state.


agorrod:
@michaelcahill I'd appreciate if you can take a look and let me know if you can think of a better structure - getting __LINE__ and __FILE__ to work involved more indirection than would be ideal.
@sulabhM please do a full review.

agorrod:
autobuild: retest this please

michaelcahill:

PR #4635, opened 2019-05-17
WT-4793 Extend test/checkpoint to use timestamps and more

agorrod:
Add options to encourage seeing conditions related to lookaside usage,
handle sweep, checkpoints and different types of data access patterns.

This is the first step, subsequently will make the new functionality
configurable.

### DIFF
--- a/dist/s_string.ok
+++ b/dist/s_string.ok
@@ -1370,6 +1370,7 @@ vw
 vxr
 vxz
 vz
+wW
 waitpid
 waker
 wakeup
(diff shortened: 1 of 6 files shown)


agorrod:
This isn't ready for merge yet - we need to parameterize much of the new functionality See WT-4793 for a suggestion about how that might work.

sueloverso:
@agorrod you should review this. It does not let me add you as a reviewer because you opened the original PR. I randomized and parameterized several things. I think a follow-on ticket could be adding in more `timing_stress_for_test` settings that does not need to be part of the epic for data loss. For now I just added `aggressive_sweep` when the `sweep_stress` arg is specified.

I ran this version of the program against the tree before WT-4750 was merged to make sure it can still detect the failure.

PR #4636, opened 2019-05-17
WT-4771 Updating compatibility tests to include 3.1 and 3.2

luke-pearson:


lukech:
Thanks for making the test changes @luke-pearson.
I found these series of compatibility tests not so easy to understand. An overview of the test purpose, various min/max compat settings, and a guideline on how to add new versions in and take old versions out of the tests would be really helpful. I assume every time we bump the release version, there's a need of updating these series of compatibility tests, with nuances depending on if major or minor or patch version is bumped.

test/suite/test_compat02.py
```diff
@@ -56,20 +56,27 @@ class test_compat02(wttest.WiredTigerTestCase, suite_subprocess):
...
     # and without the patch number. Test one on release and the
     # required minimum just for testing of parsing.
-    #
+
     compat_create = [
         ('def', dict(create_rel='none', log_create=3)),
         ('32', dict(create_rel="3.2", log_create=3)),
+        ('31', dict(create_rel="3.1", log_create=3)),
         ('30', dict(create_rel="3.0", log_create=2)),
         ('26', dict(create_rel="2.6", log_create=1)),
```
lukech: Do we still want to support compatibility testing for 2.6? It feels more sustainable to remove the oldest version while introducing a newer version into the tests, otherwise we're running more and more permutations of tests.

sueloverso: As long as the code supports log version 1 compatibility then yes, we need to test a version that uses log version 1. We need to test moving up and down between them as long as the code allows that.

luke-pearson: @sueloverso, I've documented the process for updating the compatibility version for next release, Luke has given me feedback and I'll make some improvements to it. Could you please review this PR when you get a chance?

luke-pearson: Actually @lukech has pointed out a potential issue, so hold for the time being.

luke-pearson: In order to address @lukech  concerns I've updated the comment which explains how to add version to the compat max and compat min dictionaries. The tests as they are I think are correct. So this PR is ready to be reviewed again. Apologies for the confusion.

lukech:
@luke-pearson LGTM. Thanks for updating the comments to cover the exception cases for bumping patch version.

PR #4637, opened 2019-05-17
WT-4240 document that system crashes may result in files that need salvaging

ddanderson:


ddanderson:
@sueloverso, can you please review?  This is in reference to https://jira.mongodb.org/browse/WT-4233, if after a system crash, we get a partial write of a large log record, where the last part is written out and the first part is not, we'll flag that as needing salvage.  Since this can happen in normal operations, some documentation seems appropriate.

PR #4640, opened 2019-05-20
WT-4719 Modifying prepared update resolution logic

luke-pearson:


bvpvamsikrishna:
@luke-pearson, lgtm.

Let me know once the testing kicked on this branch is passed, I will merge it.

src/include/txn.i
```diff
@@ -218,37 +222,20 @@ __wt_txn_resolve_prepared_op(
...
 	}
-#endif
-
-	WT_STAT_CONN_INCR(session, txn_prepared_updates_resolved);
-
-	for (; upd != NULL; upd = upd->next) {
-		 if (upd->txnid != txn->id)
-			continue;
 
+	for (; upd != NULL && upd->txnid == txn->id; upd = upd->next) {
```
sulabhM: Looks like we are stopping at the update that doesn't correspond to this transaction id `txn->id`. Earlier we used to go past it and search for updates that might match `txn->id`. Are we saying that once we have updates from another transaction in the update chain, we can not have updates from the current transaction `txn->id`. I think this will be correct, wouldn't hurt double checking with another reviewer.

src/include/txn.i
```diff
@@ -171,17 +171,21 @@ __txn_resolve_prepared_update(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
  *      transaction commit/rollback.
+ *
+ *      If no updates are resolved throw an error. Return the count of
+ *      resolved updates.
  */
 static inline int
 __wt_txn_resolve_prepared_op(
-    WT_SESSION_IMPL *session, WT_TXN_OP *op, bool commit)
+    WT_SESSION_IMPL *session, WT_TXN_OP *op, bool commit,
+    u_int *resolved_update_countp)
```
sulabhM: I will suggest setting `*resolved_update_countp` to 0 in this function. This will guarantee the correct result even when the caller forgets to initialize the variable to 0.

src/txn/txn.c
```diff
@@ -944,6 +942,13 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		fileid = op->btree->id;
 		switch (op->type) {
 		case WT_TXN_OP_NONE:
+			/*
+			 * Increment the resolved update count for empty
+			 * operations to avoid thinking we've lost a prepared
+			 * update as we've got less resolved updates than
+			 * the txn mod count.
+			 */
```
sulabhM: What is the reason to have empty operations? A comment will help.
Also, I see now that since you increment `resolved_update_count` here, you do not want to zero out `resolved_update_count` inside `__txn_resolve_prepared_update()`. Perhaps this could have been a different local variable and the value obtained from `__txn_resolve_prepared_update()` gets added into it. I expected `__wt_txn_resolve_prepared_op()` to output number of prepared updates, not increment the value by it. Maybe it just needs improving comments.

src/include/txn.i
```diff
@@ -171,17 +171,21 @@ __txn_resolve_prepared_update(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
+ *      Resolve a transaction's operations indirect references.
  *
  *      In case of prepared transactions, the prepared updates could be evicted
  *      using cache overflow mechanism. Transaction operations referring to
  *      these prepared updates would be referring to them using indirect
  *      references (i.e keys/recnos), which need to be resolved as part of that
  *      transaction commit/rollback.
+ *
+ *      If no updates are resolved throw an error. Return the count of
+ *      resolved updates.
```
sulabhM: See below, this function seems to be incrementing `resolved_update_countp` with resolved updates, not necessarily set it to resolved updates for this invocation.

src/txn/txn.c
```diff
@@ -1136,6 +1142,8 @@ __wt_txn_prepare(WT_SESSION_IMPL *session, const char *cfg[])
 			WT_PUBLISH(upd->prepare_state, WT_PREPARE_INPROGRESS);
 			op->u.op_upd = NULL;
 			WT_STAT_CONN_INCR(session, txn_prepared_updates_count);
+			if (upd->next != NULL && upd->txnid == upd->next->txnid)
```
sulabhM: I think this deserves a comment

src/txn/txn.c
```diff
@@ -1204,6 +1212,13 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
 
 		switch (op->type) {
 		case WT_TXN_OP_NONE:
+			/*
+			 * Increment the resolved update count for empty
+			 * operations to avoid thinking we've lost a prepared
+			 * update as we've got less resolved updates than
+			 * the txn mod count.
+			 */
+			resolved_update_count++;
```
sulabhM: same as before

src/include/txn.h
```diff
@@ -240,6 +240,11 @@ struct __wt_txn_op {
 			} mode;
 		} truncate_row;
 	} u;
+
+/* AUTOMATIC FLAG VALUE GENERATION START */
+#define	WT_TXN_OP_KEY_REPEATED	0x1u
+/* AUTOMATIC FLAG VALUE GENERATION STOP */
+	uint32_t flags;
```
sulabhM: Seems a lot of space used for a single flag. I don't feel strongly about it, seems like this came out of a discussion about algorithm to follow.

src/include/txn.i
```diff
@@ -218,37 +222,20 @@ __wt_txn_resolve_prepared_op(
...
+	 * If we haven't found anything then there's an error.
 	 */
-	if (upd == NULL && commit) {
-		/*
-		 * FIXME:
-		 * WT_ASSERT(session, txn->multi_update_count > 0);
-		 */
-		--txn->multi_update_count;
+	if (upd == NULL) {
+		WT_ASSERT(session, upd != NULL);
```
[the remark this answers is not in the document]
luke-pearson: I'm fine with leaving it inside, it does look kinda odd but makes sense overall.

PR #4644, opened 2019-05-22
Wt 4811 fix upd durable ts

bvpvamsikrishna:
Fixed durable timestamp assignment for updates created from truncate.


bvpvamsikrishna:
unwanted commits are included.

PR #4646, opened 2019-05-23
WT-4710 Modify datafile verification test

luke-pearson:
I've basically just moved the dump section into the verbose block and added the call to unset as it prevents the fork out of memory failure we have been seeing.

### DIFF
--- a/test/evergreen/verify_wt_datafiles.sh
+++ b/test/evergreen/verify_wt_datafiles.sh
@@ -81,14 +81,14 @@ do
 		fi
 
 		if [ "${verbose}" == true ]; then
-			dump=$(${wt_binary} -h ${d} dump ${t})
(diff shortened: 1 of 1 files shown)


luke-pearson:
> 
> 
> I think we deliberately set `dump` as part of the verification - a supplement step besides `verify`. If the dump output is too large and takes too much memory, add a checking for the verbose flag and only store dump output is verbose flag is set.
> Another solution is to remove the verbose support from the script. I was thinking when a file dump fails the verbose setting could give people an easy way of debugging the problem, at the time I initially wrote this script.
> I personally prefer the 1st solution.

I've set it so that if verbose is set we let the dump call print to stdout, otherwise we pipe it to /dev/null, I think this is the best solution as we avoid taking up memory either way.

lukech:
LGTM.

PR #4647, opened 2019-05-23
WT-4804 Add support for ignore_prepare=force.

michaelcahill:


sueloverso:
Is there more to this? Otherwise `force` and `true` are identical in the code and there seems no reason to add this.

agorrod:
Yes - this is a precursor to a change that disallows writes in transactions that ignore prepare. The new force flag will retain the existing behaviour.
It's being staged so MongoDB can update to the new flag before we change the semantic.

PR #4648, opened 2019-05-23
WT-4775 Make the "bad file descriptor" test resilient against crashing.

ddanderson:
As part of this change, a function is made available that allows a specialized test method to run in a subprocess.

### DIFF
--- a/test/suite/test_bug018.py
+++ b/test/suite/test_bug018.py
@@ -101,10 +101,14 @@ class test_bug018(wttest.WiredTigerTestCase, suite_subprocess):
         # sync even though table 1 was successfully written and table 2
         # had an error on close.
         self.open_conn(new_home_dir)
-        c1 = self.session.open_cursor(self.uri1)
-        c2 = self.session.open_cursor(self.uri2)
-        results1 = list(c1)
-        results2 = list(c2)
+
+        # It's possible the second table can't even be opened,
+        # that's okay for our test.
+        results1 = list(self.session.open_cursor(self.uri1))
+        try:
+            results2 = list(self.session.open_cursor(self.uri2))
+        except:
+            results2 = []
         self.assertEqual(results1, results2)
 
 if __name__ == '__main__':
--- a/test/suite/test_bug018.py
+++ b/test/suite/test_bug018.py
@@ -92,7 +92,6 @@ class test_bug018(wttest.WiredTigerTestCase, suite_subprocess):
         subdir = 'SUBPROCESS'
         [ignore_result, new_home_dir] = self.run_subprocess_function(subdir,
             'test_bug018.test_bug018.subprocess_bug018')
-        #os.system('ls -R ' + new_home_dir + '>/dev/tty')
 
         # Make a backup for forensics in case something goes wrong.
         backup_dir = 'BACKUP'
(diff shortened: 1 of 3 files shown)


ddanderson:
@agorrod, can you please suggest someone to review this?

test/suite/test_bug018.py
```diff
@@ -82,17 +85,29 @@ def test_bug018(self):
...
-        self.open_conn()
-        c1 = self.session.open_cursor(baseuri + '01.wt')
-        c2 = self.session.open_cursor(baseuri + '02.wt')
-        self.assertEqual(list(c1), list(c2))
+        self.open_conn(new_home_dir)
+        c1 = self.session.open_cursor(self.uri1)
+        c2 = self.session.open_cursor(self.uri2)
+        results1 = list(c1)
+        results2 = list(c2)
+        #self.assertGreater(len(results1), 0)
```
tetsuo-cpp: Should this just get removed?

test/suite/test_bug018.py
```diff
@@ -82,17 +85,29 @@ def test_bug018(self):
...
                 self.conn = None
 
+    def test_bug018(self):
+        '''Test closing multiple tables'''
+
+        self.close_conn()
+        subdir = 'SUBPROCESS'
+        [ignore_result, new_home_dir] = self.run_subprocess_function(subdir,
+            'test_bug018.test_bug018.subprocess_bug018')
+        #os.system('ls -R ' + new_home_dir + '>/dev/tty')
```
tetsuo-cpp: This too.

test/suite/test_bug018.py
```diff
@@ -82,17 +85,29 @@ def test_bug018(self):
             except wiredtiger.WiredTigerError:
                 self.conn = None
 
+    def test_bug018(self):
+        '''Test closing multiple tables'''
+
+        self.close_conn()
+        subdir = 'SUBPROCESS'
+        [ignore_result, new_home_dir] = self.run_subprocess_function(subdir,
```
tetsuo-cpp: If some test assertion fails in the subprocess, is it going to fail the test?
I kind of assumed that you'd need to parse an exit code for the subprocess.

ddanderson: @sueloverso, can you weigh in on this, since you wrote this test?  As I understand it, if we're closing a file handle underneath WiredTiger, we don't really care how it exits, or what error is returned, or even if an error is detected, what we care about is how recovery works in the resulting directory.  Is that correct?

test/suite/test_bug018.py
ddanderson: Fixed.

test/suite/test_bug018.py
ddanderson: Fixed

test/suite/test_bug018.py
sueloverso: The test originally expected the an error, but not looking for a specific message. So a general test is probably sufficient.

tetsuo-cpp:
retest this please

ddanderson:
@sueloverso, I needed to tweak the test.  Can you look please and see if it still make sense to you?

PR #4649, opened 2019-05-23
WT-4817 Fix off-by-one error.

sueloverso:
@ddanderson can you do a quick review of this? All sanitizer builds will fail until this is merged. With this change it passes, prior to this sanitizer catches the bug all the time.


sueloverso:
autobuild: retest this please

PR #4650, opened 2019-05-24
WT-4816 don't disallow LSM configurations in test/format

keithbostic:


keithbostic:
@sueloverso, this change is to not set the `data_source` value when it's a "permanent" value, that is, configured in some way outside of rolling dice.

PR #4651, opened 2019-05-24
WT-4760 checkpoint should not read past stable entry

luke-pearson:


michaelcahill:
Thanks @luke-pearson, lgtm.  I'll merge...

PR #4652, opened 2019-05-26
WT-4819 wt utility list command drops core on empty objects

keithbostic:


keithbostic:
@ddanderson, you wrote this code a long time ago, so in the off chance you remember what was going on, I'm pushing the review to you.

In summary, I can't think of any reason the allocation size would be 0 (the default is 4KB and it should be impossible to set an allocation size of 0), and the checkpoint address may not be set in cases where the object never created a checkpoint.

The change in 23dd804 is just minor code shuffling for clarity.

PR #4653, opened 2019-05-27
WT-4822 Track max pre compression page size

tetsuo-cpp:
This PR is adding two new stats for the max pre compression size for leaf pages and internal pages.


dist/stat_data.py
```diff
@@ -730,6 +730,8 @@ def __init__(self, name, desc, flags=''):
     CompressStat('compress_write', 'compressed pages written'),
     CompressStat('compress_write_fail', 'page written failed to compress'),
     CompressStat('compress_write_too_small', 'page written was too small to compress'),
+    CompressStat('compress_precomp_leaf_maximum_page_size', 'pre-compressed maximum size for leaf pages', 'no_clear,no_scale,size'),
+    CompressStat('compress_precomp_intl_maximum_page_size', 'pre-compressed maximum size for internal pages', 'no_clear,no_scale,size'),
```
agorrod: It's nice in these cases if the text sorts together, even if the existing group of stats don't work that way. Could you figure out a way to re-word these so they start with the word compress, yet still make sense as a sentence?

agorrod: You could make the description a bit shorter as well by abbreviating `maximum` (I'm fickle with abbreviations, but this is a case where I think it's super obvious so OK): `compress_precomp_leaf_max_page_size`

tetsuo-cpp: fef619a
I also moved them slightly since the names seem to be in alphabetical order (sort of).

dist/stat_data.py
```diff
@@ -727,6 +727,8 @@ def __init__(self, name, desc, flags=''):
     # Compression statistics
     ##########################################
     CompressStat('compress_read', 'compressed pages read'),
+    CompressStat('compress_precomp_intl_max_page_size', 'compressed page pre-compressed internal max size', 'no_clear,no_scale,size'),
+    CompressStat('compress_precomp_leaf_max_page_size', 'compressed page pre-compressed leaf max size', 'no_clear,no_scale,size'),
```
agorrod: I should have made a concrete suggestion for the text, since I'm obviously feeling pedantic:

`compressed page maximum leaf page size prior to compression`

reads better to me.

tetsuo-cpp: 8321182
Done!
I was trying to keep as much of the prefix identical as possible with the previous one.

dist/stat_data.py
```diff
@@ -727,6 +727,8 @@ def __init__(self, name, desc, flags=''):
     # Compression statistics
     ##########################################
     CompressStat('compress_read', 'compressed pages read'),
+    CompressStat('compress_precomp_intl_max_page_size', 'compressed page maximum internal page size prior to compression', 'no_clear,no_scale,size'),
```
bvpvamsikrishna: I think "compressed internal page maximum size prior to compression" or "compressed maximum internal page size prior to compression" sounds better.  If changing please change the next one as well accordingly.

PR #4654, opened 2019-05-27
Add warning if the open file limit is too small.

fedorova:
After conversion to Python3, a script that visualizes operation tracking logs begin to fail on systems where the open file limit is set to 256. The culprit is the multiprocessing module, which appears to not clean up the open files right away after the processes exit. Others are reporting similar issues online. This is affecting us only on OSX. The workaround is simple: the user only needs to bump the open file limit. To keep things simple, I print a warning to the user and terminate the script if the open file limit is too small.


agorrod:
autobuild: retest this please

PR #4655, opened 2019-05-28
WT-4823 Add check for uninitialised lookaside resources

tetsuo-cpp:
For in-memory configurations, the lookaside session isn't allocated so we need to add a `NULL` check here.

I've amended the `test_las04` tests to include in-memory configuration. When I remove the check, this test causes a segmentation fault in the same spot as the reported issue so I believe that this is the correct fix.

After discussing with @agorrod, we elected to simply ignore lookaside configuration for in-memory setups.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -75,11 +75,15 @@ __wt_las_config(WT_SESSION_IMPL *session, const char **cfg)
 		    "max cache overflow size %" PRId64 " below minimum %d",
 		    cval.val, WT_LAS_FILE_MIN);
 
+	/* This is expected for in-memory configurations. */
+	las_session = S2C(session)->cache->las_session[0];
+	if (las_session == NULL)
+		return (0);
+
 	/*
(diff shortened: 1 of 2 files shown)


src/cache/cache_las.c
```diff
@@ -75,11 +75,15 @@ __wt_las_config(WT_SESSION_IMPL *session, const char **cfg)
 		    "max cache overflow size %" PRId64 " below minimum %d",
 		    cval.val, WT_LAS_FILE_MIN);
 
+	/* This is expected for in-memory configurations. */
+	las_session = S2C(session)->cache->las_session[0];
```
agorrod: Could you add an assertion in here:

`WT_ASSERT(session, las_session != NULL || F_SET(conn, WT_CONN_IN_MEMORY));`

tetsuo-cpp: 4d28e8b
Oops, forgot this one.
Also changed `F_SET` to `F_ISSET` since I think that's making the flag true.

PR #4656, opened 2019-05-28
WT-4830 Modify encrypt06 test to be clearer and match documentation

ddanderson:


ddanderson:
@keithbostic, I set this up as a pull request against your branch for you to review.  Encrypt06 should now pass with your changes.  However, the rules for whether metadata gets encrypted seem a little strange.  Perhaps I'm thinking about it the wrong way.  I've reworked the test, hopefully to be somewhat clearer.  Please take a look at `expect_encrypted_metadata` in `test_encrypt06.py` and see if it matches your expectations before merging.

ddanderson:
(Hmm, it won't pass the PR test without a proper comment - nonetheless, it passes locally for me).

ddanderson:
@keithbostic, I've made some changes to make this test more understandable.  It skip tests where encryption settings for column groups and indices don't match the table's settings. What happens currently in the API for this case doesn't make sense, and WT-4830 (to fix that) is proving tougher than it looks. I'm hoping this can unblock you.

ddanderson:
Made another improvement - we no longer skip testing the case where column group (or index) config is blank, but the tests are conservative: we at least ensure we won't leak info when asked to encrypt the table.

michaelcahill:
Thanks @ddanderson, this change makes sense to me -- I'll merge.  On @agorrod's request, I'll use WT-4830 for this change.

PR #4657, opened 2019-05-28
WT-4828 Fix the type exception in operation tracking.

fedorova:
Explicitly cast the value to int to avoid a type exception. This started showing up after we converted to Python3.


fedorova:

PR #4658, opened 2019-05-28
WT-4829 Fix broken filenames in operation tracking.

fedorova:
After conversion to python 3, file names were encoded as byte arrays and not decoded. As a result the filenames used at HTML links were not properly formed and we could not click on these links. Add proper decoding to fix that.

### DIFF
--- a/tools/optrack/find-latency-spikes.py
+++ b/tools/optrack/find-latency-spikes.py
@@ -864,7 +864,7 @@ def generateTSSlicesForBuckets():
                               numBuckets, "Generating timeline charts");
 
     for i, fname in returnValues.items():
-        bucketFilenames.append(str(fname.value));
+        bucketFilenames.append(fname.value.decode("ascii"));
     print(color.END);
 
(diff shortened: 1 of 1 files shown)


tools/optrack/find-latency-spikes.py
```diff
@@ -864,7 +864,7 @@ def generateTSSlicesForBuckets():
                               numBuckets, "Generating timeline charts");
 
     for i, fname in returnValues.items():
-        bucketFilenames.append(str(fname.value));
+        bucketFilenames.append(fname.value.decode("ascii"));
```
ddanderson: @fedorova, technically, should be `fname.value.decode()` or `fname.value.decode('utf-8')`.  The `fname.value` was generated by this statement at line 685: `retFilename.value = fileName.encode();` - and the default encoding is `utf-8`.  I realize it won't any difference in this case, as they are simple generated file names, but we should be in the habit of matching up encodings/decodings.

PR #4659, opened 2019-05-29
WT-4190 Decrease shutdown time by doing multi-threaded cache flush

tetsuo-cpp:
This PR is artificially bringing the dirty target down to encourage the eviction threads to help reconcile prior to shutdown.

I reproduced similar results to those recorded by @agorrod in the ticket. Bringing the target value down even further doesn't seem to yield any real improvement. And again, this doesn't help much for when we have a single table.

### DIFF
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1064,6 +1064,13 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
 			F_SET(conn, WT_CONN_CLOSING_TIMESTAMP);
 	}
 
+	/*
+	 * Ramp the eviction dirty target down to encourage eviction threads to
+	 * clear dirty content out of cache.
+	 */
+	conn->cache->eviction_dirty_trigger = 1.0;
+	conn->cache->eviction_dirty_target = 0.1;
+
 err:	/*
 	 * Rollback all running transactions.
 	 * We do this as a separate pass because an active transaction in one


agorrod:
@tetsuo-cpp there is one other change I'm interested in testing here. I'd like to know how much cache (clean and dirty) is still in use before and after the call to `__wt_evict_destroy` in `__wt_connection_close`.

I'm interested because, the subsequent call to `__wt_conn_dhandle_discard` will go through and clean up any content that remains in cache. That's single threaded which is why ramping down the eviction server is effective (it lets us do the work multi-threaded on shutdown). It may be that waiting a while for the targets to be reached before discarding data handles would be worthwhile.

tetsuo-cpp:
@agorrod 
Ok so I did some testing to see:
- How much of the cache is dirty when we begin reduce the target and when we begin discarding the dhandle.
- How much time is spent discarding the dhandle.

My results look like this:

```
May 30 01:37:38: Ramping down target, inuse=13992339339, dirty=13600363532
May 30 01:38:06: About to perform dhandle discard, inuse=5505717479, dirty=0
May 30 01:38:16: Finished dhandle discard
```

So there seems to be nothing dirty prior to discarding the dhandle and yet we spend roughly a quarter of our shutdown time here. I'm using `__wt_cache_bytes_inuse` and `__wt_cache_dirty_inuse` for these stats.

Looks like there is quite a bit of freeing going on here even without the single threaded disk IO part when we have dirty bytes in cache so perhaps that could explain it? If this result doesn't make sense to you, I'm happy to investigate further with `perf` to figure out where we're spending our time.

PR #4661, opened 2019-05-29
WT-4608 Skip verbose cache dump of files opened exclusively.

sueloverso:


sueloverso:
@agorrod Don't merge this yet. @keithbostic had questions about why verify pages are not able to be evicted.

sueloverso:
I reverted the change to turn off timing stress during verify. I think it is better to have it stop with the stuck cache than to run for many hours, until WT-4834 is done. I will merge the piece that avoids the segfault.

sueloverso:
autobuild: retest this please

PR #4664, opened 2019-06-03
WT-4719 Improve validation of prepared update resolution

luke-pearson:


src/include/txn.i
```diff
@@ -218,37 +222,27 @@ __wt_txn_resolve_prepared_op(
...
-
+		/*
+		 * Aborted updates can exist in the update chain of our TXN
+		 * generally this will occur due to a reserved update.
+		 * As such we should skip over these updates, if the txn
+		 * id is then different and not aborted we know we've
+		 * reached the end of our update chain and can exit.
+		 */
+		if (upd->txnid == WT_TXN_ABORTED) continue;
+		if (upd->txnid != txn->id) break;
```
michaelcahill: whitespace: our style is to always put the body on a separate line.

src/txn/txn.c
```diff
@@ -978,8 +977,27 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
-				    session, op, true));
+				/*
+				 * Decrement the resolved update count as we
+				 * as we would either resolve at least 1 update
+				 * or have previously resolved the update linked
+				 * to this modification when we resolved the
+				 * prepared updates of a previous modification
+				 * in this txn.
+				 */
+				resolved_update_count--;
```
michaelcahill: I'm finding this hard to wrap my head around.  I *think* the idea is that we're adding to `resolved_update_count` inside `__wt_txn_resolve_prepared_op` for every update that gets resolved.  And here we're decrementing it for every op we visit.  So it should end up at zero?

But then we use the resulting count to increment a statistic, so if it's supposed to be zero at the end, I'm missing something.

Would it lose something to maintain two counts, `update_count` (which gets incremented here) and `resolved_update_count` that is incremented in `__wt_txn_resolve_prepared_op`, assert that `resolved_update_count >= update_count` in the loop, then at the end assert that `resolved_update_count == update_count`?

That way there is something meaningful to use for statistics, and I think the check that all updates get resolved at the end is important.

src/include/txn.i
```diff
@@ -218,37 +222,27 @@ __wt_txn_resolve_prepared_op(
...
-
-	WT_STAT_CONN_INCR(session, txn_prepared_updates_resolved);
 
 	for (; upd != NULL; upd = upd->next) {
-		 if (upd->txnid != txn->id)
-			continue;
-
+		/*
+		 * Aborted updates can exist in the update chain of our TXN
+		 * generally this will occur due to a reserved update.
```
sueloverso: Minor nits: 'TXN' should lowercase. Also, the first line is a complete sentence. Please end it with a period and capitalize the 'G' in 'generally' on the next line.

src/include/txn.i
```diff
@@ -218,37 +222,27 @@ __wt_txn_resolve_prepared_op(
...
+		 * As such we should skip over these updates, if the txn
```
sueloverso: The comma should be a period.

src/txn/txn.c
```diff
@@ -978,8 +977,27 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 
 				__wt_txn_op_set_timestamp(session, op);
 			} else {
-				WT_ERR(__wt_txn_resolve_prepared_op(
-				    session, op, true));
+				/*
+				 * Decrement the resolved update count as we
+				 * as we would either resolve at least 1 update
```
sueloverso: 'as we' is at both the end of the first line and beginning of the 2nd line.

src/txn/txn.c
```diff
@@ -1218,10 +1241,29 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
...
 			 * operation, in case of prepared transaction.
 			 */
-			if (F_ISSET(txn, WT_TXN_PREPARE))
-				WT_RET(__wt_txn_resolve_prepared_op(
-				    session, op, false));
-			else {
+			if (F_ISSET(txn, WT_TXN_PREPARE)) {
+				/*
+				 * Decrement the resolved update count as we
+				 * as we would either resolve at least 1 update
```
sueloverso: Same comment as before. `as we` is repeated. 
Since this is a copy of the earlier code, @michaelcahill's comment also applies here. Maybe consider refactoring into a function to have one copy.

michaelcahill:
Thanks @luke-pearson, this lgtm now.  Fingers crossed it stays merged this time...

PR #4665, opened 2019-06-03
WT-4494  Change import API to work directly on a file URI

michaelcahill:


michaelcahill:
Just for discussion while I'm getting to grips with the main change.

@keithbostic, I'll try to catch you tomorrow to discuss: I think I understand how `import` works for "file:" URIs, but I couldn't see how they would work for anything else...

michaelcahill:
Also, as you predicted, I'm concerned about a "file:XXX" URI where the matching file XXX doesn't exist -- I suspect that's going to cause a bunch of problems (e.g., backup for starters).

keithbostic:
@michaelcahill 
>  I think I understand how import works for "file:" URIs, but I couldn't see how it would work for anything else...

I mostly agree. We haven't solved the problem of how to import objects from outside a database, and this work potentially has applicability to that problem. Arguably, this isn't the way we'd do it --  when importing objects we'd reasonably have existing metadata to work with, and for complex objects we'd need some way to describe the file relationships as well, which means multiple rows of metadata at least.

Perhaps this work is the way we'd get metadata for objects that didn't already have it, as a precursor to importing them for real? Anyway, my only plan here was to make sure you and @agorrod had thought through the specific functionality this work should cover, and hopefully the API will fall out of that.

> Also, as you predicted, I'm concerned about a "file:XXX" URI where the matching file XXX doesn't exist -- I suspect that's going to cause a bunch of problems (e.g., backup for starters).

Yes. For import to work, we either need to copy files into the database home (which seems like a terrible idea), or have some way to reference external files. Again, we're just poking at the import problem here, but I think we'll need some kind of solution eventually, so it would be good to think about it now, even if it's just to declare "this isn't it".

michaelcahill:
After discussion offline, I think we're more or less in sync about these changes.  We confirmed that `file:/path/to/file.wt` works, which I believe handles @keithbostic's main use case.  And we both agreed that we should have a story for exporting higher level metadata, but that this change (supporting just "file" URIs) is an important step.

agorrod:
@keithbostic are you happy to pull this into your main branch for this change?

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this

PR #4666, opened 2019-06-03
WT-4671 Remove limbo pages and associated statistics

sulabhM:


michaelcahill:
Sorry for causing trouble here: I have a script that deletes branches that are "fully merged into develop", and because the base branch here is empty, it was accidentally closed.

agorrod:
I'm going to go ahead and merge - this is going into a feature branch.

PR #4667, opened 2019-06-03
WT-4839 Shuffle python tests and execute every Nth test

sulabhM:


test/suite/run.py
```diff
@@ -117,6 +117,8 @@ def usage():
   -h      | --help               show this message\n\
   -j N    | --parallel N         run all tests in parallel using N processes\n\
   -l      | --long               run the entire test suite\n\
+  -o N    | --Nth N              randomly sort scenarios to be run, then\n\
```
luke-pearson: Option here might be better as -N? so you'd do something like run.py -N 20? Could also be -r potentially for random.

sulabhM:
@agorrod, have a look, please. Appreciated a suggestion for the name/letter for the option.

test/suite/run.py
agorrod: I agree re matching `-n` for the short version. How about leaving out a short option altogether and calling it `--random-sample`. I'd be happy with either of those choices.

sulabhM:
@agorrod we already have -n for a dry run that lists all the scenarios. I put a change to have -r or --random-sample.
I think this looks good, I will checkin after the pull req tester passes.

PR #4668, opened 2019-06-04
WT-4827 Apply commit timestamps to deletion operations

luke-pearson:


src/include/txn.i
```diff
@@ -442,6 +442,45 @@ __wt_txn_op_apply_prepare_state(
...
+	txn = &session->txn;
+	ts = txn->commit_timestamp;
+
+	/*
+	 * Lock the ref to ensure we don't race with eviction freeing the page
+	 * deleted update list or with a page instantiate.
+	 */
+	for (;; __wt_yield()) {
+		previous_state = ref->state;
+		WT_ASSERT(session, previous_state != WT_REF_READING);
```
sulabhM: Can leave a comment explaining why the prev state cant be reading

src/include/txn.i
```diff
@@ -442,6 +442,45 @@ __wt_txn_op_apply_prepare_state(
...
+		if (previous_state != WT_REF_LOCKED && WT_REF_CAS_STATE(
```
sulabhM: consider restructring as:
```
if (previous_state != WT_REF_LOCKED &&
    WT_REF_CAS_STATE(session,
    ref, previous_state, WT_REF_LOCKED))
        break;
```

src/include/txn.i
```diff
@@ -442,6 +442,45 @@ __wt_txn_op_apply_prepare_state(
...
+		    session, ref, previous_state, WT_REF_LOCKED))
+			break;
+	}
+
+	for (updp = ref->page_del->update_list;
+	    updp != NULL && *updp != NULL; ++updp) {
+		(*updp)->start_ts = ts;
```
sulabhM: it would be more readable to replace ts with txn->commit_timestamp OR, save both txn->durable_timestamp and txn->commit_timestamp into local variables and assign to each update.

src/include/txn.i
luke-pearson: I don't think we need a comment here hopefully it's clear enough that we shouldn't be running this function if our current page is being read in.

michaelcahill:
Thanks @luke-pearson, lgtm.  I'll merge.

PR #4669, opened 2019-06-04
WT-4844 Only log an informational message when a set read-timestamp is older than the oldest timestamp.

keithbostic:
Alex, this is a change already queued up in WT-4733 (#4610), to not log an error message if the application specifies a timestamp older than the current oldest timestamp.

The argument is an application will hit this unless it explicitly synchronizes the read and oldest timestamps, and it's not an error so much as an alert the application's read timestamp likely needs to move forward.

The `test/format` PR #4610 requires this change, and MongoDB is running into it as well (SERVER-40591), which is why I'll pulled it out into its own PR.

@bvpvamsikrishna, @michaelcahill: is this OK with you?

### DIFF
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -928,6 +928,21 @@ __wt_txn_set_read_timestamp(
 			did_roundup_to_oldest = true;
 		} else {
 			__wt_readunlock(session, &txn_global->rwlock);
+
+			/*
+			 * In some cases, MongoDB sets a read timestamp older
+			 * than the oldest timestamp, relying on WiredTiger's
+			 * concurrency to detect and fail the set. In other
+			 * cases it's a bug and MongoDB wants error context to
+			 * make it easier to find those problems. Don't output
+			 * an error message because that logs a MongoDB error,
+			 * use an informational message to provide the context
+			 * instead.
+			 */
+			WT_RET(__wt_msg(session, "read timestamp "
+			    "%s less than the oldest timestamp %s",
+			    __wt_timestamp_to_string(read_ts, ts_string[0]),
+			    __wt_timestamp_to_string(ts_oldest, ts_string[1])));
 			return (EINVAL);
 		}
 	} else
--- a/src/txn/txn_timestamp.c
+++ b/src/txn/txn_timestamp.c
@@ -928,6 +928,8 @@ __wt_txn_set_read_timestamp(
 			did_roundup_to_oldest = true;
 		} else {
 			__wt_readunlock(session, &txn_global->rwlock);
+			if (!WT_VERBOSE_ISSET(session, WT_VERB_TIMESTAMP))
+				return (EINVAL);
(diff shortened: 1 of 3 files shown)


src/txn/txn_timestamp.c
```diff
@@ -928,6 +928,8 @@ __wt_txn_set_read_timestamp(
 			did_roundup_to_oldest = true;
 		} else {
 			__wt_readunlock(session, &txn_global->rwlock);
+			if (!WT_VERBOSE_ISSET(session, WT_VERB_TIMESTAMP))
+				return (EINVAL);
```
sueloverso: I was talking to Eric and Louis about this issue. This change rather a hack. At a minimum it needs a long comment about why we'd return EINVAL when the next line down is also returning EINVAL with more information and why does verbose factor into it at all? MongoDB is raising an error when receiving the original value and message, but they would prefer to have the actual content as an informational message I believe.

sueloverso: What about something like:
```
/* comment about it */
__wt_msg( information about timestamps);
return (EINVAL);
```
that is not predicated on verbose and still gives the information. That is tolerable but not great. It is impossible to know in general, which of our messages like this one are "informational" errors.

keithbostic: My preference is no message at all, simply `return (error-maybe-not-EINVAL);` and document that fact. In other words, what you're referring to as "informational errors" shouldn't be logged at all.

I'm not the go-to person on this stuff, but it seems reasonable to me that application read timestamps can age out, and it's OK when that happens.

I agree about the verbose check (I put in `WT_VERBOSE_ISSET` when I thought `test/format` was the only application that cared about this).

sueloverso: I'm in half-agreement. I think from the library perspective, detecting this error is valid. I also agree that `error-not-EINVAL` is better. I just spoke with @milkie. It seems like changing `EINVAL` to `WT_ROLLBACK` (in the `WT_RET_MSG` call) may solve the problem. The caller to `timestamp_transaction` can retry with another timestamp and they do get the informational message (which is desired).

keithbostic: I pushed a branch that discards the message and documents the error. At this point, maybe it's best if @bvpvamsikrishna, @agorrod, and @michaelcahill weigh in and make this call.

sueloverso: @milkie has said to me, several times, that they still want the message. Is there a problem with the suggestion of changing EINVAL to WT_ROLLBACK in WT_RET_MSG?

keithbostic:
autobuild: retest this please

agorrod:
I think logging something to the message callback makes sense in this case. I'm pretty ambivalent about the error code we return - I think EINVAL more accurately captures the type of error, but understand that MongoDB already has handling for WT_ROLLBACK that makes it more suitable for their use case here.

keithbostic:
@sueloverso, I believe this change is acceptable to @milkie, specifically returning EINVAL is fine.

PR #4670, opened 2019-06-04
WT-4837 Don't sleep in timing_stress if the cache is totally full.

sueloverso:
With this change the LSM + timing stress stuck cache failure does not happen. @keithbostic let me know what you think of these tweaks to the timing stress function.


sueloverso:
autobuild: retest this please

PR #4671, opened 2019-06-04
WT-4460 Optimize for in-order, non-overlapping modifications

keithbostic:


src/support/modify.c
```diff
@@ -172,6 +178,172 @@ __modify_apply_one(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
...
+	 * overwriting fixed-sized fields.
+	 *
+	 * If either the data size or replacement count extend past the end of
+	 * the current value, we have to deal with padding bytes. Don't try to
+	 * fast-path padding bytes; it's not common and adds branches to the
+	 * loop applying the changes.
+	 */
+	for (i = 0; i < nentries; ++i) {
+		if (entries[i].data.size != entries[i].size ||
+		    value->size < entries[i].offset + entries[i].size ||
```
sulabhM: Since comparison operators are short-circuiting, when the first condition is false, the next two conditions would end up being same, ie entries[i].size equals entries[i].data.size. So we can get rid of one of those?

src/support/modify.c
```diff
@@ -172,6 +178,172 @@ __modify_apply_one(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
...
+		memmove(to, from, n);
+		to += n;
+		memmove(to, entries[i].data.data, entries[i].data.size);
+		to += entries[i].data.size;
+
+		from += n + entries[i].size;
+	}
+
+	/* Copy any data remaining in the original buffer. */
+	n = value->size - WT_PTRDIFF(from, value->data);
```
sulabhM: When we copy the remaining data, we must be copying the trailing '\0' too? Unless that got replaced by a modify.
So if we restore the trailing null, there might be cases we are adding an extra '\0'?

src/support/modify.c
keithbostic: Thanks, fixed in a5fd572.

src/support/modify.c
keithbostic: Nice catch, thanks! Fixed in a0336bc.

keithbostic:
@sulabhM, thanks for the review. I also have three specific questions.

1. I couldn't think of any way to make the fast-path work in a single buffer. I think it would be possible if all of the modifications are growing or shrinking the buffer (in the case of growing the buffer, by building the new value from the end of the buffer back toward the beginning), but I don't think modifications that only grow or only shrink the buffer are likely, and I can't see any way to make it work in a single buffer if we're mixing modifications that grow the buffer with modifications that shrink the buffer. Can you think of a way that might work?

2. I think it would be possible to use the intermediate results of the fast-path work. So, the same way we can consume modifications that are simply replacements, then fall back to a slower method once we hit a modification that isn't a simple replacement, we could do the new fast-path for as long as it works and then fall back to a slower method once it no longer works. I didn't do that because there's a scratch buffer allocation and a buffer copy in that case, plus it would add branching to the loop that's applying the changes. On the other hand, it means we no longer have to walk the list of modifications twice, we just start applying changes and see how far it goes (although that implies we don't know the correct size for the buffer allocation and might not allocate a big enough one). Thoughts?

3. I don't expect padding bytes (bytes after the end of the existing value), are likely in a MongoDB workload. Do you (or @michaelcahill) know if that's true?

michaelcahill:
I'll try to find some time today to talk through 1 & 2 with @sulabhM.

On 3, no, MongoDB never tries to read/write past the end of a document.

michaelcahill:
Sorry, didn't get a chance to grapple with this today.  I'll try again tomorrow.

keithbostic:
@michaelcahill, thanks, there's no urgency as far as I know.

sulabhM:
@keithbostic , I couldn't get time to think of a solution that avoids an extra buffer allocation. I am away on vacation for the next few days. I will make some time when back if the pull request is still up.

sulabhM:
@keithbostic, I spent some time thinking about a solution that would not require allocating an extra buffer,  and it is not obvious to me how. If there is such a solution, do you think we can merge this change and work on it separately? I am assuming other than that all is resolved in this pull request and is ready to merge ?

keithbostic:
@sulabhM 
> If there is such a solution, do you think we can merge this change and work on it separately?

That's fine with me, whatever you & @agorrod think best.

agorrod:
@michaelcahill I know you wanted to think about optimizing this one as well. Is it OK with you if we merge this approach and open a follow on ticket?

agorrod:
I touched base with @michaelcahill about this change today. He is concerned that it will introduce a regression for the case where there is a large value (in excess of 1MB) and between 1-5 small updates to that value.

We are going to hold off merging until we've explored all optimization paths.

keithbostic:
FTR, talked to @michaelcahill, the concern is if we aren't going to get a scratch buffer because the value it too large (the scratch buffer code limits what we'll cache to 2MB), so we'll pay an allocation that will likely have to be resolved by the operating system.

On the other hand, the original fallback method, if the current buffer is big enough, will repeatedly shuffle memory, but won't have to do an allocation.

If we eventually decide the current code is going to be merged, we should probably add a check that if the current buffer is big enough for the updates and it's also large enough that the allocation won't be satisfied by an existing scratch buffer, we fall back to memory shuffling instead of doing the allocation.

michaelcahill:
I kicked off some performance tests with these changes:

https://evergreen.mongodb.com/version/5d3f9d2fe3c331413d4a3ea0
https://evergreen.mongodb.com/version/5d3f9d1be3c331413d4a3e11

There are no regressions, but also no big wins.  It's likely that the tests don't do operations that are complex enough to get the benefit of these changes (without being so complex that we fall back to the original implementation).

Let's take a look and decide whether we like this change for its own sake, otherwise I don't think we've got strong evidence that it makes any of the workloads we measure in automated testing better.

michaelcahill:
autobuild: retest this please

michaelcahill:
I should add that I like the test changes here regardless of whether we take the new implementation: they made it much easier for me to see when things were going wrong.

keithbostic:
@michaelcahill, I still like this change.

It can't hurt, and at least in some obvious cases, it's going to help. Merging & closing.

Thank you for the nice algorithmic work!

sueloverso:
I reverted the merge of this due to so many test failures. I have restored the branch.

PR #4672, opened 2019-06-05
WT-4712 Add diagnostic logging. (#4572)

sueloverso:
* WT-4712 Add diagnostic logging.

* We may not have a btree here, don't check it.

* Add bit field on fileid to ignore operation and remove ignore flag
from the record.

* Allocate a txn id in diagnostic mode if needed.

* Only call logging of timestamp setting if one is set.

* Skip timestamp operation in recovery. Add diagnostic to timestamp_abort.

* Need to unpack timestamp operation to move pointers to next operation.

* Correctly set bit to ignore the fileid. Remove verbose from test.

* Remove debug line.

* Conditionalize assertion that is no longer always true.

* Add support for durable and prepare timestamps.

* Whitespace and spelling.

* Add clock time to txn_timestamp log record.

* Add in support for checkpoint retention (i.e. archiving changes).

* Remove verbose setting from test.

* Move diagnostic free to general connection code.

* Ignore reconfigure while debugging.

* Fix off by one bug.

* Add special uint32_id type for fileids so they are printed in both
decimal and hex in printlog.

* Minor wordsmithing.

* Fix wording for long line.

* Commit auto-generated file.

* Reinstate code supporting reconfigure.

* Remove accidentally committed code.

* Check the flag, not just the bit value.

* Remove extra arg.

* Add diagnostic operation range usage to make it easier to ignore
those operation records.

* Add some comments to log_data.py file about diagnostic operations.

* Wordsmithing.

* Account for size of record we skip.

* Rename the config string to debug_mode and related changes.

* Fix editing error.

* Add test for rollback_error debug mode setting.

* Add test for checkpoint_retention.

* Fix minor test issue.

* Add debug print line.

* Use // for integer division for python3.

* Add test for table_logging.

* Whitespace

* Accomodate slow machines in sleep/archive test.

* Review comments.

* Fix typo

* Modify description.

* Turn on verbose=(log) on slow systems.

* Simplify archive checking to (hopefully) alleviate timing issues.

(cherry picked from commit a610f847393f5d21923f92e38a7e1aa7258f7d47)

### DIFF
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -456,6 +456,25 @@ connection_runtime_config = [
             above 0 configures periodic checkpoints''',
             min='0', max='100000'),
         ]),
+    Config('debug_mode', '', r'''
+        control the settings of various extended debugging features''',
+        type='category', subconfig=[
+        Config('checkpoint_retention', '0', r'''
+            adjust log archiving to retain the log records of this number
+            of checkpoints. Zero or one means perform normal archiving.''',
+            min='0', max='1024'),
+        Config('rollback_error', '0', r'''
+            return a WT_ROLLBACK error from a transaction operation about
+            every Nth operation to simulate a collision''',
+            min='0', max='10M'),
+        Config('table_logging', 'false', r'''
+            if true, write transaction related information to the log for all
+            operations, even operations for tables with logging turned off.
+            This setting introduces a log format change that may break older
+            versions of WiredTiger. These operations are informational and
+            skipped in recovery.''',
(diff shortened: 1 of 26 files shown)


sueloverso:
@agorrod or @luke-pearson please carefully review this backport PR to 4.0. It was not a clean cherry-pick. In particular `txn_timestamp.c` required some significant changes. I did test that the correct information is getting written to the log from `timestamp_abort`. But there was a lot of code refactoring in that one file that made this more complex than usual.

PR #4674, opened 2019-06-07
WT-4790 Coverity: Memory leak within __wt_curbulk_init

tetsuo-cpp:
This PR is just to free the `WT_RECONCILE` struct that we allocate earlier in the function.
I think it makes sense to free here if we're returning non-zero.


tetsuo-cpp:
Apologies, I missed the conditional where we don't always allocate in that function. Also the part about using the cleanup functions makes sense since those structs have pointers to stuff that has been separately malloc'd.

I've more or less done exactly what you suggested above with some slight modifications to get it to compile and use the right return value.

PR #4676, opened 2019-06-07
WT-4850 __slvg_checkpoint() metadata memory leak

keithbostic:
Memory allocated into `config` was never freed.

I also explicitly initialized `ckptbase`, just for clarity.


sueloverso:
LGTM. Merge when PR testing completes.

PR #4678, opened 2019-06-07
WT-4852 Evergreen runs of test/format require shared library extensions

keithbostic:


lukech:
It looks Jenkins jobs (e.g. wiredtiger-test-check-long) runs `make check` under `build_posix`, which triggered a `cd test/csuite` and calling of test program under `build_posix/test/csuite`. See WT-4860 for one of the build failures.

PR #4679, opened 2019-06-08
WT-4854 wt utility list command has a memory leak and minor error handling bugs.

keithbostic:


keithbostic:
@luke-pearson, I need to merge this PR (#4679) because the sanitizer builds are hitting it on every run.

Please do a review when you're back next week, and I'm happy to fix any problems you find then.

PR #4681, opened 2019-06-09
WT-4853 standalone recovery code cannot handle deleted checkpoints

keithbostic:
@agorrod, @michaelcahill, this is a significant change to the standalone recovery code Michael merged. In summary, the previous code tried to repair the most recent checkpoint found in the metadata, and that doesn't work in all cases. The reason I didn't see it was I had a bug in my test harness, and was only testing the import process, instead of doing a full verification after the import.

In this branch, instead of just storing the base file metadata and the corrected checkpoint information in the file, the block manager now constructs the complete, almost-correct, file checkpoint metadata entry, and copies it into the file's avail list along with the base metadata and the corrected checkpoint information.

Additionally, I've pushed changes into the test program to do a full verify after the import.

The new layering is a little uglier than before (but not that much), and almost all of the changes are in the block-scan and btree-import code, so the added risk in merging this code is minimal.

### DIFF
--- a/src/block/block_ckpt.c
+++ b/src/block/block_ckpt.c
@@ -9,8 +9,8 @@
 #include "wt_internal.h"
 
 static int __ckpt_process(WT_SESSION_IMPL *, WT_BLOCK *, WT_CKPT *);
-static int __ckpt_update(
-	WT_SESSION_IMPL *, WT_BLOCK *, WT_CKPT *, WT_BLOCK_CKPT *);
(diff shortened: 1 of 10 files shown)


keithbostic:
@agorrod, I'm going to go ahead and merge the new standalone recovery PR (#4681), and see if I can get our testing back on track. I'm fine with reverting the original PR and this PR if you want to initiate a longer review process, but given the changes in this PR are almost exclusively in the new import code, I believe merging this PR doesn't make anything worse.

PR #4682, opened 2019-06-09
WT-4853 update WT_SESSION.import example code and documentation.

keithbostic:


keithbostic:
@lukech, this is the fix for the current `wiredtiger-doc-build` failures.

PR #4685, opened 2019-06-10
WT-4856 format should test LSM and timestamp configurations.

keithbostic:
@agorrod, @sueloverso, this turns LSM and timestamp configurations back on in test/format.

I'm running LSM with timestamps on a test box, I'll let it run for a few hours and see how it goes.


keithbostic:
We can't merge this change, LSM can't handle the current `test/format` behavior of repeatedly attempting `WT_CURSOR.search/search-near` operations in the face of `WT_PREPARE_CONFLICT` errors.

PR #4687, opened 2019-06-11
WT-4859 Failed to find test/format program in endian tests

keithbostic:


lukech:
The little-endian generation test passed. Will go ahead and merge.

PR #4688, opened 2019-06-12
WT-4776 Modify operations should be equivalent to updates (v4.0 backport)

lukech:
If modifies are created with out-of-order timestamps, WiredTiger should use apply all previous modifications until it finds a complete copy of the document. Once a modify is identified as the most recent visible version for a value WT should not apply normal visibility rules to the prior versions - they were visible when the modify was performed, so should be included in the read.

(cherry picked from commit 8eb2d9c36c5c23da09f75e16dff982f0be52d34f)


lukech:
@michaelcahill It's a clean cherry-pick. Thanks.

michaelcahill:
Thanks @lukech, lgtm.  I'll merge.

PR #4692, opened 2019-06-12
WT-4694 Throttle reads in wtperf to make readonly workload perform more consistently

tetsuo-cpp:
In our Jenkins testing, IO performance seems to vary a lot. This change is simply throttling the read workload so that variation in IO performance is less likely to affect the `500m-btree-rdonly` workload.

### DIFF
--- a/bench/wtperf/runners/500m-btree-rdonly.wtperf
+++ b/bench/wtperf/runners/500m-btree-rdonly.wtperf
@@ -17,5 +17,5 @@ report_interval=10
 run_time=7200
 sample_interval=10
 sample_rate=1
-threads=((count=20,reads=1))
+threads=((count=20,reads=1,throttle=10000))
 warmup=120
--- a/bench/wtperf/runners/500m-btree-rdonly.wtperf
+++ b/bench/wtperf/runners/500m-btree-rdonly.wtperf
@@ -17,5 +17,5 @@ report_interval=10
 run_time=7200
 sample_interval=10
 sample_rate=1
-threads=((count=20,reads=1,throttle=10000))
+threads=((count=20,reads=1,throttle=500))
 warmup=120
--- a/bench/wtperf/runners/500m-btree-rdonly.wtperf
+++ b/bench/wtperf/runners/500m-btree-rdonly.wtperf
@@ -17,5 +17,5 @@ report_interval=10
 run_time=7200
 sample_interval=10
 sample_rate=1
-threads=((count=20,reads=1,throttle=500))
+threads=((count=20,reads=1,throttle=625))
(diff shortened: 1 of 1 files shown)


sueloverso:
I'm not sure this is a good change to make. I agree it will flatten the plots and avoid variations. But it simply moves the bottleneck to sleeping. If the read-only plots are pegged at 72M or 100K/10sec always, which is at the very low end of its wide range, what information will the plots give us? We can never notice any improvements nor degradations.

tetsuo-cpp:
@sueloverso 
Do you feel that the number is too low? Or that throttling in general is a bad idea.

I chose 100K/10 sec because it's somewhere between the max and min that we're seeing at the moment. I've skewed the number towards the min because I want to make sure it gets flattened out but if it seems to extreme, we could do 125K/10 sec or 150K/10 sec.

@agorrod 
Do you have thoughts on what seems like a reasonable number?

agorrod:
@tetsuo-cpp this change makes sense to me. We aren't interested in chasing performance variations that are due to hardware variability - which is what we were doing here. Limiting the test so that it's close to but within the bounds of the hardware capacity makes sense to me.

bench/wtperf/runners/500m-btree-rdonly.wtperf
```diff
@@ -17,5 +17,5 @@ report_interval=10
 run_time=7200
 sample_interval=10
 sample_rate=1
-threads=((count=20,reads=1))
+threads=((count=20,reads=1,throttle=10000))
```
agorrod: I'm a little surprised that this is the right throttle config - I thought the number was applied to each thread, so 10k would result in 20x10k ops/second. If that isn't how it works we should update the wtperf configuration option description to match reality (https://github.com/wiredtiger/wiredtiger/blob/develop/bench/wtperf/wtperf_opt.i#L192)

tetsuo-cpp: My mistake. I misunderstood this configuration and have fixed it now to be divided by the number of threads. I ran it just then with 500 for 12 min and got:
```
Executed 7196882 read operations (100%) 9995 ops/sec
```

sueloverso:
I still fear that throttling is going to render the test not very useful and we'll never see a change. I'd suggest bumping it up to be more mid-range. We can see how that goes.

tetsuo-cpp:
@sueloverso 
Ok, I've brought it up to 125K/10 sec. This is about halfway between the max and min for most of the runs so I think this is pretty reasonable.

```
Executed 8996000 read operations (100%) 12494 ops/sec
```

PR #4642, opened 2019-05-21
WT-4352 Backport data corruption fixes

agorrod:
Tagged with a ticket to get PR testing started.

PR #4693, opened 2019-06-13
WT-4352 Backport data corruption fixes to 3.6

lukech:


lukech:
@sueloverso I was trying to backport the series of data inconsistency fixes to 3.6. There were some merge conflicts I fixed manually, and the pull request tester picked up the `size of array ‘__check_WT_REF’ is negative` failure. Not quite sure what changes I missed that are required to pass the size checking. Could you help to take a look? Thanks.
The series of commits I cherry-picked from 4.0 are in #4642

lukech:
@bvpvamsikrishna helped to confirm the problem was about the setting of `WT_REF_SIZE`. 

The cherrypick (from mongodb-4.0) brought the block of definitions of `__wt_ref_hist` and `histoff` (under the HAVE_DIAGNOSTIC ifdef control) into `__wt_ref` structure, but the related `WT_REF_SIZE` define change (with HAVE_DIAGNOSTIC ifdef) was missing. The ticket that introduced the `__wt_ref` structure and related `WT_REF_SIZE` change is WT-3898, which was backported to 4.0 but not 3.6. 

I checked the mongodb-3.6 src directory and did not find any instance that is using the `__wt_ref_hist` or `histoff` structure. 

2 solutions:
1. Keep the addition of `__wt_ref_hist` in `__wt_ref` structure, and add related `WT_REF_SIZE` define change (with HAVE_DIAGNOSTIC ifdef) to reach consistency
2. Remove the definition of `__wt_ref_hist` and `histoff`

@michaelcahill can you help to justify above analysis? If it's right solution 2 feels cleaner to me.

agorrod:
@lukech I think we need to take a different approach here - I can see several issues with the backport.

I think we shouldn't backport the changes for WT-4796 - they are the ones that are causing the WT_REF size problem, and they are designed to facilitate live debugging of issues, which we don't really do with 3.6 any more.

We need to understand one of the changes in txn_rollback_to_stable.c as well:
https://github.com/wiredtiger/wiredtiger/pull/4693/files#diff-1441b97f8a34662f4c2645a06d0871acL257
There is a call to __wt_timestamp_cmp in the 3.6 code, which is different in 4.0. We need to understand when that change was made in 4.0, and what we should do in 3.6. This was associated with the fix in WT-4352.

The original conditional associated with the change to btree.i from WT-4769 is different. We need to figure out whether the change is relevant there as well.

In summary: Let's not backport WT-4796. We need to do more investigation of WT-4352 and WT-4769 to figure out if other tickets need to be backported as well in order for the changes to work as expected.

lukech:
@agorrod Thanks for the suggestions. I reverted WT-4796, and wrap the timestamp comparison lines with `__wt_timestamp_cmp`. It was WT-4427 that removed the definition and reference to `__wt_timestamp_cmp` in 4.0. The build and PR testing are passing now. 

Not sure if the addition of `page_las->resolved` checking logic in WT-4769 is applicable in 3.6. Any suggestion?

agorrod:
@lukech the difference in 3.6 and 4.0 to this line:
https://github.com/wiredtiger/wiredtiger/commit/11106cb8bc718a46b663e51685552db58272fbaa#diff-b2442811cb1095a2949962a9b1fa9c24L1193

Was introduced as part of WT-3898 - I suspect we decided not to backport that change since MongoDB doesn't use prepared transactions in 3.6. I think this diff shouldn't be pulling in the unrelated change.

lukech:
@agorrod I see. Removed the checking for `page_las->has_prepares`. Can you help to have another look? Cheers.

agorrod:
Can we close this one now @lukech?

lukech:
Yes. Closing it as we re-did the backporting work in #4705

PR #4695, opened 2019-06-13
WT-4842 Adjust lookaside score when walking obsolete.

sueloverso:
Some comment fixes I saw along the way too.

@agorrod Please review this to see if it is what you had in mind. It took me a while to find a location that didn't really have an impact. The obsolete code is one of the few places we're walking the updates already in order to see/detect "saving due to history".

Please closely look at the conditional to determine an update is unstable to verify it is correct.

I did write a test to see the value move.


agorrod:
Thanks @sueloverso this approach makes sense to me. I'll reach out and get confirmation that it addresses issues with the previous approach.

agorrod:
I've updated to develop and switched perf testing over to this branch.

PR #4699, opened 2019-06-14
WT-4869 When eviction isn't keeping up, stop adding pages to cache.

michaelcahill:


luke-pearson:
I've got a sys-perf patch build for this here: https://evergreen.mongodb.com/version/5d0721310305b94476a2e8bb

luke-pearson:
The patch looks good. @sulabhM would you be able to review this PR?

src/evict/evict_lru.c
```diff
@@ -613,32 +616,38 @@ __evict_update_work(WT_SESSION_IMPL *session)
...
-	if (bytes_inuse < (uint64_t)((target + trigger) * bytes_max) / 200 &&
-	    dirty_inuse <
-	    (uint64_t)((dirty_target + dirty_trigger) * bytes_max) / 200)
-		F_SET(cache, WT_CACHE_EVICT_SCRUB);
+	if (bytes_inuse < (uint64_t)((target + trigger) * bytes_max) / 200) {
+		if (dirty_inuse < (uint64_t)
+		    ((dirty_target + dirty_trigger) * bytes_max) / 200)
+			LF_SET(WT_CACHE_EVICT_SCRUB);
+	} else
+		LF_SET(WT_CACHE_EVICT_NOKEEP);
```
sulabhM: Are we saying that by default with MongoDB when cache usage exceeds (80 + 5)/2 = 42.5%, we will not keep new pages being read into the cache?

src/meta/meta_ckpt.c
```diff
@@ -662,6 +662,7 @@ __wt_meta_ckptlist_free(WT_SESSION_IMPL *session, WT_CKPT **ckptbasep)
 
 	WT_CKPT_FOREACH(ckptbase, ckpt)
 		__wt_meta_checkpoint_free(session, ckpt);
+	__wt_free(session, ckpt->name);
```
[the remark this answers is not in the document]
michaelcahill: Sorry, unrelated change attempting to keep coverity quiet.  This should be discarded.

src/evict/evict_lru.c
michaelcahill: It's when the cache usage exceeds (80 + 95) / 2 == 87% by default.  That is, background eviction threads are not able to keep up with the read load (which would keep the cache at 80%).

PR #4700, opened 2019-06-14
WT-4658 Clang Format Preview

tetsuo-cpp:
Here's a preview of the Clang Format output after applying the changes we agreed to.
I'll include the extra script under `dist/` to flush goto labels left shortly.

I don't intend to actually merge this branch since it will quickly accumulate merge conflicts. Once we agree to merging, I'll run Clang Format over a fresh branch off of develop and use that instead.

I've also chosen not to bring over server's formatting script as it has dependencies on other scripts under `buildscripts/` that each need to be adapted for WiredTiger making it a maintenance burden.


.clang-format
```diff
@@ -67,7 +67,7 @@ IncludeCategories:
   - Regex:           '.*'
     Priority:        1
 IndentCaseLabels: false
-IndentWidth:     8
+IndentWidth:     4
```
kommiharibabu: 8 to 4 space indentation is the major change that I observed as the major change in the diff, is it possible to keep it as 8 for one commit and later change only the indent to 4 and run the clang and provide those changes as another commit?

src/txn/txn_timestamp.c
```diff
@@ -28,117 +27,101 @@ __wt_timestamp_to_string(wt_timestamp_t ts, char *ts_string)
...
-		if (hex_val < 0)
-			WT_RET_MSG(session, EINVAL,
-			    "Failed to parse %s timestamp '%.*s'",
-			    name, (int)cval->len, cval->str);
-		ts = (ts << 4) | (uint64_t)hex_val;
-	}
-	*timestamp = ts;
-
-	return (0);
+    static const int8_t hextable[] = {-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
```
luke-pearson: 😢 Should we avoid reformatting thing like this?

src/txn/txn_timestamp.c
```diff
@@ -658,318 +603,298 @@ __txn_assert_after_reads(
...
+    bool did_roundup_to_oldest;
+
+    WT_RET(__wt_txn_context_prepare_check(session));
+
+    /* Read timestamps imply / require snapshot isolation. */
+    if (!F_ISSET(txn, WT_TXN_RUNNING))
+        txn->isolation = WT_ISO_SNAPSHOT;
+    else if (txn->isolation != WT_ISO_SNAPSHOT)
+        WT_RET_MSG(session, EINVAL, "setting a read_timestamp"
+                                    " requires a transaction running at snapshot"
```
luke-pearson: String re-indents seem kinda weird?

src/txn/txn_log.c
```diff
@@ -240,64 +231,62 @@ err:		__wt_logrec_free(session, &logrec);
...
+
+    uint32_t fileid;
+
+    conn = S2C(session);
+    txn = &session->txn;
+
+    if (!FLD_ISSET(conn->log_flags, WT_CONN_LOG_ENABLED) ||
+      F_ISSET(session, WT_SESSION_NO_LOGGING) ||
+      (F_ISSET(S2BT(session), WT_BTREE_NO_LOGGING) &&
+        !FLD_ISSET(conn->log_flags, WT_CONN_LOG_DEBUG_MODE)))
```
luke-pearson: This if looks a bit weird with the 3rd line of re-indent being extra indented

src/conn/conn_reconfig.c
```diff
@@ -13,315 +13,272 @@
...
+                               "cannot be larger than compatibility release %" PRIu16 ".%" PRIu16,
+          min_major, min_minor, rel_major, rel_minor);
+
+    /*
+     * On a reconfigure, check the new release version against any
+     * required maximum version set on open.
+     */
+    if (reconfig && conn->req_max_major != WT_CONN_COMPAT_NONE &&
+      (conn->req_max_major < rel_major ||
+          (conn->req_max_major == rel_major && conn->req_max_minor < rel_minor)))
```
luke-pearson: More weird indentation

.clang-format
```diff
@@ -121,5 +98,5 @@ SpacesInParentheses: false
 SpacesInSquareBrackets: false
 Standard:        Cpp11
 TabWidth:        8
```
agorrod: Should TabWidth be set to 4 just for consistency? Or should we remove this explicit setting if we aren't using it?

.clang-format
agorrod: I'm with Hari - I actually wonder if we should make two separate changes here. It's going to be pretty much impossible to notice a functional change in this enormous patch, but even harder if it touches literally every line because of the indenting change.

tetsuo-cpp: Ok. I've rewritten history here so there's a commit where I've applied the original Clang Format style and then another commit where I've applied our agreed changes (mostly just indentation).

Before applying agreed changes: d66812e
After applying agreed changes: f3fa720

@agorrod @kommiharibabu Is this just for the purposes of this pull request? Or do you want the actual commits that go into `develop` to look like this? To me, it just creates yet another inconvenient commit that gets in the way when blaming a line.

.clang-format
tetsuo-cpp: Sure. I'll most likely set it to 4 since I'd rather have the full list of settings so it's easier to edit/compare.

src/txn/txn_timestamp.c
tetsuo-cpp: d7e9a08

Literals always get aligned like this. I made a change to always break before multi line strings so hopefully this looks less weird.

src/txn/txn_log.c
tetsuo-cpp: As discussed before, this is default behaviour for line breaks within a nested expression. Related LLVM discussion here: https://reviews.llvm.org/D42787

src/conn/conn_reconfig.c
tetsuo-cpp: As discussed, it's indented even further to avoid aligning with the body of the `if` statement.

src/txn/txn_timestamp.c
tetsuo-cpp: Probably. I'm not 100% sure what i'm looking at but it seems important for it to be displayed as a table.

.clang-format
kommiharibabu: @tetsuo-cpp I don't have any problem in squashing the commits before merging.
With seperate commit in pull request will help the reviewers to find out exactly what is changed by the clang other than indentation.

kommiharibabu: @tetsuo-cpp @agorrod I just want to hear your opinions on the backport tickets after the changes of tabs from 8 to 4, this may introduce some additional conflicts?

tetsuo-cpp: Sorry for taking a long time to respond to this.
The way the server teams did this was to apply the formatter to each of the release branches. This way, you will have a window in time where merge conflicts can happen (backporting a pre-format `develop` change to `mongodb-4.0`). But any new post-format changes going into develop, which will be the majority of backports going forwards, shouldn't be affected.

dist/s_goto.py
```diff
@@ -0,0 +1,18 @@
+# Trim all trailing whitespace in front of goto labels.
+# This is a workaround for a Clang Format limitation where goto labels are
+# automatically indented according to nesting.
+import re, sys
+
+# 1. Zero or more whitespace characters.
+# 2. One or more lowercase ASCII characters.
+# 3. Colon character.
+p = re.compile('^\s*[a-z]+:$')
```
kommiharibabu: There are some goto statements in the code as "err: /*", the clang format is not bringing the comments into the next line like statements, so the newly added goto script is missing to adjust these gotos. IMO, it is better to move the comment into the new line, so that this s_goto can work properly.

.clang-format
kommiharibabu: @tetsuo-cpp I am happy that if we take care of applying the clang format to all the supported branches. If we can merge the clang format changes when there are no planned minor releases,  we can delay the backport fixes until the clang format changes are done in supported branches also.

dist/s_comment.py
```diff
@@ -0,0 +1,44 @@
...
+        in_multiline = True
+        # Figure out how far we need to indent.
+        for c in line:
+            if c == ' ':
+                indentation += 1
+            else:
+                break
+    # End of a comment. If we were in the middle of a multi-line comment then
+    # write it.
+    elif sline.endswith('*/'):
```
tetsuo-cpp: No need to look at this file yet. It's on the right track but needs a bit of refinement and cleanup.

dist/s_goto.py
tetsuo-cpp: Thanks for picking up on that. I agree that the comment should just go underneath the tag.
Before applying this change into `develop`, I'm going to do a PR to some preliminary cleanup like remove some low value macros. I'll also include this change at the same time.

src/btree/bt_curnext.c
```diff
@@ -488,16 +478,15 @@ err:	__wt_scr_free(session, &a);
...
+__wt_cursor_key_order_check(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, bool next)
 {
 	switch (cbt->ref->page->type) {
 	case WT_PAGE_COL_FIX:
 	case WT_PAGE_COL_VAR:
 		return (__cursor_key_order_check_col(session, cbt, next));
 	case WT_PAGE_ROW_LEAF:
 		return (__cursor_key_order_check_row(session, cbt, next));
-	WT_ILLEGAL_VALUE(session, cbt->ref->page->type);
+		WT_ILLEGAL_VALUE(session, cbt->ref->page->type);
```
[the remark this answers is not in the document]
tetsuo-cpp: Yes, this is one of the macros that is going to be removed. It's just a wrapper around a function so there's no harm in getting rid of these.

src/block/block_ext.c
```diff
@@ -777,37 +750,36 @@ __block_ext_overlap(WT_SESSION_IMPL *session,
...
-	if (a->off == b->off) {				/* Case #1, #4, #9 */
-		if (a->size == b->size) {		/* Case #1 */
-			/*
-			 * Move caller's A and B to the next element
-			 * Add that A and B range to the avail list
-			 * Delete A and B
-			 */
+	if (a->off == b->off) {           /* Case #1, #4, #9 */
+		if (a->size == b->size) { /* Case #1 */
+			                  /*
```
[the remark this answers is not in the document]
tetsuo-cpp: I don't think this can be helped. We want to have the `AlignTrailingComments` on and it gives us this behaviour.

src/btree/bt_sync.c
```diff
@@ -54,8 +54,7 @@ __sync_checkpoint_can_skip(WT_SESSION_IMPL *session, WT_PAGE *page)
 	 * the lock.
 	 */
 	if (mod->rec_result == WT_PM_REC_MULTIBLOCK)
-		for (multi = mod->mod_multi,
-		    i = 0; i < mod->mod_multi_entries; ++multi, ++i)
+		for (multi = mod->mod_multi, i = 0; i < mod->mod_multi_entries; ++multi, ++i)
```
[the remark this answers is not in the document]
tetsuo-cpp: This is not possible with Clang Format so I'm going to consider this out of scope for the time being. I believe this is possible with Clang Tidy but getting this setup up properly is a separate body of work.

src/btree/bt_debug.c
```diff
@@ -960,18 +934,20 @@ __debug_page_col_int(WT_DBG *ds, WT_PAGE *page, uint32_t flags)
...
 	if (LF_ISSET(WT_DEBUG_TREE_WALK))
-		WT_INTL_FOREACH_BEGIN(session, page, ref) {
+		WT_INTL_FOREACH_BEGIN (session, page, ref) {
 			if (ref->state == WT_REF_MEM) {
 				WT_RET(ds->f(ds, "\n"));
 				WT_RET(__debug_page(ds, ref, flags));
 			}
-		} WT_INTL_FOREACH_END;
+		}
+	WT_INTL_FOREACH_END;
```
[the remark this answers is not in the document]
tetsuo-cpp: It's because the `if` doesn't have braces. So it interprets the the `WT_INTL_FOREACH_BEGIN` as the only statement.
The solution is just to put braces around this stuff. If we feel strongly about it, I can include these changes in the aforementioned PR.

src/block/block_ext.c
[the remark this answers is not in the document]
kommiharibabu: on a further check, I feel that not every comment in the if block is right aligned. Only if there exist any comments like below before.

if (a->size == b->size) { /* Case #1 */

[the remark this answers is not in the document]
tetsuo-cpp: My understanding is it just aligns comments on consecutive lines.

dist/s_comment.py
```diff
@@ -0,0 +1,67 @@
+import re, sys
```
tetsuo-cpp: @agorrod @keithbostic 
This is the requested script to fill comments to the new line length.

It works well for the most part but I doubt we'll ever get it to the point where it'll be fine to run this unsupervised. For example:
* We don't always paragraph comments in the same way.
* Some comments contain numbered lists/bullet points.
* Comments within macros.

If you both agree, I'd rather just use this script as a one time thing to do most of the legwork rather than making it first class and keeping it under `dist/`.

keithbostic:
@tetsuo-cpp 
> This is the requested script to fill comments to the new line length.

When I first thought about it, I was thinking the script would only touch simple comments, that is, ignore any comment where any line of the comment had anything other than a leading asterisk and a single space immediately followed by `[A-Za-z]`. Skimming the sources, it seemed like that was most of the comments, and avoided anything that was formatted in any interesting way.

Did you find that not to be the case?

Regardless, running this script one time with review, or on demand, is fine with me.

tetsuo-cpp:
> Did you find that not to be the case?

The main exceptions are the comments describing each function and block comments with paragraphs. Though, I can make the script ignore that and get most of the benefit. It feels like a job half done though...

My idea was to use this script as a one time thing (and fix manually) straight after landing the Clang Format changes. Since I'm applying the Clang Format changes to each of the release branches, I'll need to repeat this process for each of them.

keithbostic:
> The main exceptions are the comments describing each function and block comments with paragraphs.

Should be easy (??) to pick up function comments by looking for the '--$\n\t' pattern?

Block comments with paragraphs, yeah, shrug, there aren't that many, are there? We can fix them over time.

> My idea was to use this script as a one time thing (and fix manually) straight after landing the Clang Format changes.

I wouldn't fix anything manually because you don't want to do that to every release, do you? Plus, anything you do manually carries a bit more risk, I'd think. I would try hard to keep this as a "crank turn", one run of clang and a script or two, on each release.

keithbostic:
@tetsuo-cpp, it would be good to know the recipe you end up using so if people have branches they want to bring forward, they can run the recipe on their branch, take a diff, and apply it to a new tree? Would that work?

PR #4701, opened 2019-06-16
WT-4832 closing WiredTiger connection corrupts LSM handles

keithbostic:
@agorrod, @michaelcahill, I think this problem has been around for a long time and just surfaced recently because of the changes in the `checkpoint` test program.

In summary, `WT_CONNECTION.close` sets `WT_CONN_CLOSING_TIMESTAMP` while LSM is still opening and writing file handles, and nothing good happens after that.

Like @sueloverso, I inevitably saw one of two failure modes: either a dead handle is written because LSM re-opened a handle (and because `WT_CONN_CLOSING_TIMESTAMP` was set it got discarded out from under us), or, a bloom filter didn't have the expected contents. I didn't chase that one down, so I don't know the details of that failure, but I'm suspicious of how we're evicting dirty pages when `WT_CONN_CLOSING_TIMESTAMP` has been set.

I tried a couple of different solutions that left shutting down the LSM and async servers in their current place, after we've done the final checkpoint, but they all got messy fast. I thought this change might be OK (because I think of LSM and async servers as running on top of the rest of WiredTiger,) but that may have led me into a bad change. Specifically, while I don't see any place I've broken the final checkpoint code by closing the LSM and async servers first, I wouldn't bet a lot I'm correct.

Anyway, this branch at least shows where the problem is, I'm happy to work on implementing a different solution if you prefer something else.


keithbostic:
Passing the test suite with a faster locking subsystem requires a little more extensive rearrangement.

Broadly restructure __conn_close() into a few different parts: first, configure general stuff like pushing the cache to be more aggressive; second, close external transactions and sessions and wait for stuff to drain; third, close down high-level servers (async & LSM), then perform the final checkpoint.

Two notes: if we throw this change away, we should preserve the move of the `err:` lable in `__conn_close()`, it's almost certainly wrong. Second, why don't we move the final checkpoint into `__wt_txn_global_shutdown()`, isn't it reasonably part of shutting down the global transaction system?

agorrod:
Thanks @keithbostic I'm nervous about changing the shutdown order - as are you based on your comments. I've taken a look at how WT_CONN_CLOSING is used and don't see any obvious issues with changing the ordering based on that, but I'm going to do some more thinking about this and try to figure out if there are other issues.

src/conn/conn_api.c
```diff
@@ -1047,22 +1047,13 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
 	const char *ckpt_cfg;
 
 	conn = (WT_CONNECTION_IMPL *)wt_conn;
-	ckpt_cfg = "use_timestamp=false";
 
 	CONNECTION_API_CALL(conn, session, close, config, cfg);
-
-	/* The default session is used to access data handles during close. */
-	F_CLR(session, WT_SESSION_NO_DATA_HANDLES);
+err:
```
sueloverso: Having the `err:` label at the start of the function is quite unusual and worthy of a comment.

src/conn/conn_open.c
```diff
@@ -80,28 +80,12 @@ __wt_connection_close(WT_CONNECTION_IMPL *conn)
 	F_SET(conn, WT_CONN_CLOSING);
```
sueloverso: Unrelated to your change specifically, but in the ballpark: `wt_connection_close` is only called from 2 functions in `conn_api.c`. It should move to that file and be a static in there instead of here.

src/conn/conn_open.c
```diff
@@ -80,28 +80,12 @@ __wt_connection_close(WT_CONNECTION_IMPL *conn)
...
 	 * exit before files are closed.
 	 */
-	WT_TRET(__wt_lsm_manager_destroy(session));
-
-	/*
-	 * Once the async and LSM threads exit, we shouldn't be opening any
-	 * more files.
-	 */
-	F_SET(conn, WT_CONN_CLOSING_NO_MORE_OPENS);
-	WT_FULL_BARRIER();
```
sueloverso: I am concerned about removing these lines from `wt_connection_close`. That function is called on the error path out of `wiredtiger_open` and if those threads were started and we later got an error, we would no longer call the `destroy` functions here. At a minimum, I'd like to see test cases added that (somehow) force an error in `wiredtiger_open` to convince ourselves this is okay.

sueloverso: Maybe have a new function `connection_close_early` that calls the `wt_async_destroy` and `wt_lsm_manager_destroy` and sets the flags that can be called in `conn_close` as well as in the `wiredtiger_open` error path?

keithbostic:
Thank you, Sue.

@agorrod, are we committed to making these changes work, shall I address @sueloverso's review specifics?

keithbostic:
@sueloverso, 
> I'm concerned about error path with this change. I also think we need explicit testing with async and/or LSM and error path on wiredtiger_open added.

What do you have in mind, how would that testing work?

sueloverso:
One idea for testing would be to force an error during `wiredtiger_open`, after the LSM and async threads have started. One simple (but a bit hacky since the test would depend on the current order in the code) would be to have a forced API configuration error for a subsystem after those two systems have started. They're started in `conn_open.c:wt_connection_workers`. LSM is started very early, async near the end. Something like an illegal `io_capacity` setting would cause an error after both are started.

src/conn/conn_api.c
keithbostic: I don't object to a comment if you want, feel free to add what you like.

But, it's not that unusual, we do it everywhere there's a API macro (XXX_API_) and we're not jumping out of the function on error, usually found in some API "close" function )(`__curbackup_close()`, `__curconfig_close()`, `__curds_close()`, `__curdump_close()`, `__curfile_close()`, `__curindex_close()`, `__curjoin_close()`, `__curlog_close()`, `__curmetadata_close()`, `__curstat_close()`, `__curtable_close()`).

src/conn/conn_open.c
keithbostic: I don't disagree, but every function in `conn_open.c` is only called from `conn_api.c`, by that argument, we'd merge `conn_open.c` entirely into `conn_api.c`.

And, that's true for `conn_handle.c` as well. The files `conn_handle.c` and `conn_open.c` are 263 and 141 lines respectively. The file `conn_api.c` is 2838 lines, so I guess I see how we got here, somebody was trying to split things up.

You sure?

@agorrod, @michaelcahill, any ideas how to cleanly split up the `WT_CONNECTION` code into multiple files?

src/conn/conn_open.c
keithbostic: I just reviewed this again, and please tell me if I'm missing something.

I think there are 3 questions we have to answer:

* Do we start either the LSM or async services before `wiredtiger_open()` returns? If so, there's a problem because `wiredtiger_open()` errors call `__wt_connection_close()` to do cleanup, and the proposed change removes LSM and async shutdown from that cleanup.

* Do the LSM or async services care about the `WT_CONN_CLOSING` flag? If so, there's a problem because removing LSM and async shutdown from `__wt_connection_close()` means LSM and async must correctly shut down absent that flag being set.

* Does the code in `__wt_connection_close()`, executed after LSM and async shutdown, ever open files? If so, there's a problem because the proposed change sets the `WT_CONN_CLOSING_NO_MORE_OPENS` flag earlier than before.

I think the answer to all 3 questions is "No", and so the change is safe.

Do you see a path I'm missing?

keithbostic:
@sueloverso 
> One idea for testing

I added an assert that neither LSM or async are running in `__wt_connection_close()`.

src/conn/conn_open.c
agorrod: In my ideal world conn_api.c would just contain the API entry points. Simple entry points could be implemented there directly , but otherwise the implementations should be in a different file.

i.e: I'd rather not move the functions into conn_api.c.

sueloverso: Thanks for the bigger picture. Stand down, we'll leave it as-is.

src/conn/conn_open.c
sueloverso: The answer to #1 is yes. At a minimum for async:
`wt_connection_workers -> wt_async_create -> async_start -> wt_thread_create`

keithbostic:
@sueloverso, it turned out to be pretty straight-forward to start the async server on demand.

I've requested another review from you, let me know what you think.

A couple of notes:
* In `__wt_async_reconfig()`, I'm passing its `cfg` argument to `__wt_async_create()`. I believe that's correct, because the `WT_CONNECTION.reconfigure` constructed `cfg` is the final `wiredtiger_open()` connection configuration plus the configuration passed into `WT_CONNECTION.reconfigure`.

* In `__wt_async_new_op()`, I'm constructing the same pair to pass to `__wt_async_create()`, that is, the final `wiredtiger_open()` connection configuration plus the configuration passed into `__wt_async_new_op()`. Note that previously, `__conn_async_new_op()` passed `__wt_async_new_op()` the BASE configuration plus the configuration passed into `__conn_async_new_op()`, which is, I think, incorrect. It worked because (1) `__wt_async_new_op()` doesn't care about anything in the base configuration, and (2) `__wt_async_new_op()` calls `__async_runtime_config()`, which uses `__wt_config_gets_def()`, ignoring any values from `cfg[0]`. We could more simply write `__async_runtime_config()` to retrieve values from `config` string by itself rather than constructing a `cfg[]`, but shrug, I didn't see any point in rewriting it, it's not going to be any faster.

keithbostic:
@agorrod, @sueloverso and I believe this set of changes fixes the problem and is safe.

Can you sign off on the general shutdown reconfiguration?

I'd also suggest moving the final checkpoint into `__wt_txn_global_shutdown()`. It seems to me like it's part of shutting down the global transaction system, but you have a better feel for that structure than I do.

PR #4702, opened 2019-06-16
WT-4863 Prototype WiredTiger inline functions

keithbostic:
@agorrod, the problem here is we don't prototype inline functions and so we're not asking the compiler to complain when return values are ignored.

We're also missing out on some other compiler checking:
* because `__wt_snprintf` is inlined, we don't get argument type checking because the compiler doesn't realize it's a printf-style function,
* we're not noticing if we ignore return values from any function returning something other than `bool` or `int`.

I rewrote `dist/s_prototype` in Python, and then prototyped the inline functions.

This change has a bad side-effect: we can no longer use `(void)` to ignore the return value of a function, we have to use `WT_IGNORE_RET` (and `WT_IGNORE_RET` now has two new flavors, `WT_IGNORE_BOOL` and `WT_IGNORE_PTR`, to handle different function return types).


keithbostic:
There's another bad side-effect: we can no longer put declared type information (that's used as a function return or function argument type), in `XXX.i` files. We did that in one place, we had this in `include/txn.i`:

```
typedef enum {
       WT_VISIBLE_FALSE=0,     /* Not a visible update */
       WT_VISIBLE_PREPARE=1,   /* Prepared update */   
       WT_VISIBLE_TRUE=2       /* A visible update */
} WT_VISIBLE_TYPE;
```
and it had to move.

keithbostic:
@luke-pearson, feel free to modify the Python w/o discussion,my Python is pretty poor.

PR #4707, opened 2019-06-18
WT-4827 Apply commit timestamps when a truncated page is read (#4668) (backport-4.0)

kommiharibabu:
(cherry picked from commit 39d5985b0db70d601a5adeedbbed964c13d61a24)

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -425,6 +425,42 @@ __wt_txn_op_apply_prepare_state(
 }
 
 /*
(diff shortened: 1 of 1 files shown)


kommiharibabu:
There are no start_ts and durable_ts fields in 4.0 version in WT_UPDATE structure and also there is no durable_timestamp field in WT_TXN structure. The mapping is done to timestamp field of WT_UPDATE structure to commit_timestamp field in WT_TXN structure.

PR #4708, opened 2019-06-18
WT-4758 Create a workload that stresses eviction by running periodic scans

ddanderson:
To wtperf, add scan_table_count, scan_interval, scan_pct, scan_icount to
support threads that do scanning, using their own separate tables or the
regular set of tables.

Records are not scanned in order, but using a pattern that eventually visits
all records, and using tables in round robin.

Added printing for more fail conditions, so we can tell better why threads fail.

When an operation fails with rollback error when running with log_like tables
enabled and ops per txn at 0, do retries, instead of exiting.

Added a new wtperf workload using the new scan features, tuned to show
failures similar to ones seen in the field.

### DIFF
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -1724,8 +1724,7 @@ execute_populate(WTPERF *wtperf)
 	lprintf(wtperf, 0, 1,
 	    "Starting %" PRIu32
 	    " populate thread(s) for %" PRIu64 " items",
-	    opts->populate_threads,
-	    (uint64_t)opts->icount + (uint64_t)opts->scan_icount);
+	    opts->populate_threads, max_key);
 
 	/* Start cycling idle tables if configured. */
 	start_idle_table_cycle(wtperf, &idle_table_cycle_thread);
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -41,6 +41,7 @@ static WT_THREAD_RET monitor(void *);
 static WT_THREAD_RET populate_thread(void *);
 static void	 randomize_value(WTPERF_THREAD *, char *);
 static void	 recreate_dir(const char *);
+static WT_THREAD_RET scan_worker(void *);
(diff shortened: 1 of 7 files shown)


ddanderson:
@luke-pearson and @tetsuo-cpp, can you please hold off from merging for a moment?  I ran this again after pushing the branch and the run crashed yesterday.  I had just merged with latest develop, so I assume some bug got exposed.  Unfortunately, it takes an hour and a half to run.

ddanderson:
@luke-pearson and @tetsuo-cpp, I've done some retesting.  Under gdb, everything runs fine, and I got the same results that I reported in the ticket.  The crashes I saw, not under gdb, just reported "Killed" to my shell.  I was thinking it might be an out of memory kind of error.  I'll run it some more not under gdb.

ddanderson:
@luke-pearson, please go ahead and review and merge if you're satisfied.  I ran a few more times and did not see the "Killed".  This test uses most of the physical memory available on the machine, so it's possible it will sometimes die with OOM error.  A future improvement may be to try to refine the wtperf script to work with a smaller cache, yet exhibit the same symptoms. I was trying to get it as close to the customer use case as possible as a first cut.  I think this test, as cumbersome as it is, is still useful as a first cut.

bench/wtperf/wtperf.c
```diff
@@ -1556,18 +1712,20 @@ execute_populate(WTPERF *wtperf)
...
 	opts = wtperf->opts;
+	max_key = (uint64_t)opts->icount + (uint64_t)opts->scan_icount;
 
 	lprintf(wtperf, 0, 1,
 	    "Starting %" PRIu32
-	    " populate thread(s) for %" PRIu32 " items",
-	    opts->populate_threads, opts->icount);
+	    " populate thread(s) for %" PRIu64 " items",
+	    opts->populate_threads,
+	    (uint64_t)opts->icount + (uint64_t)opts->scan_icount);
```
luke-pearson: Might be better to just use max_key here?

ddanderson:
retest this please

PR #4709, opened 2019-06-19
WT-4803 Implement file_max configuration for Cache Overflow mechanism (v4.0 backport)

tetsuo-cpp:
Clean cherrypick of 1be63c5 off `develop`.


tetsuo-cpp:
Actually, this one was not a clean cherrypick. Not sure why I wrote that.
It was trivial to fix though so I don't anticipate problems.

PR #4712, opened 2019-06-19
WT-4877 Uninitialized memory being written during checkpoints

keithbostic:


michaelcahill:
Thanks @keithbostic, lgtm -- I'll merge.

PR #4713, opened 2019-06-20
WT-4733 Enhance format diagnostic logging to include timestamps.

agorrod:


agorrod:
@keithbostic and @bvpvamsikrishna I went ahead and changed the operation logging in format to include timestamp information. I pushed this into another PR because it's not the cleanest implementation - but I couldn't see an obviously better way.

With this set of changes, I can get more information about the failure. For example:
Running `./t -c c4733 -l`

```
snapshot-isolation 0000449018.00/opqr search mismatch
	expected {0000449018/LMNOPQRSTUV}
	   found {000044}
```

The log contains:

```
$ grep 449018 log
0x55dde26a6f90:bulk K     449018 {0000449018.00/opqr}
0x55dde26a6f90:bulk V     449018 {0000449018/LMNOPQRSTUV}
0x55dde26a7da0:WT_CURSOR.next{0000449018.00/opqr}, {0000449018/LMNOPQRSTUV}, read_ts: 147e1, commit_ts: 147e5, repeatable read: yes
0x55dde26a7440:WT_CURSOR.prev{0000449018.00/opqr}, {0000449018/LMNOPQRSTUV}, read_ts: 17307, commit_ts: 17306, repeatable read: no
0x55dde26a7440:WT_CURSOR.prev{0000449018.00/opqr}, {0000449018/LMNOPQRSTUV}, read_ts: 1d02f, commit_ts: 1d02e, repeatable read: no
0x55dde26a7440:put       {0000449018.00/opqr}, {000044}, read_ts: 1ebf6, commit_ts: 1ebf5, repeatable read: no
0x55dde26a7da0:read      449018, read_ts: 1fcca, commit_ts: 1fccd, repeatable read: yes
0x55dde26a7da0:snapshot-isolation 0000449018.00/opqr search mismatch, read_ts: 1fd22, commit_ts: 1fd35, repeatable read: no
```

The core shows the following search:
```
(gdb) p *snap
$1 = {op = READ, keyno = 449018, ts = 130250, repeatable = true, last = 0, kdata = 0x0, ksize = 0, kmemsize = 0, vdata = 0x7efbe433a5b0,
  vsize = 22, vmemsize = 23}
(gdb) p /x snap->ts
$2 = 0x1fcca
```

What surprises me here is that the `snap->ts` differs from the `tinfo->read_ts` - that doesn't seem to be expected. If the `snap->ts` had the timestamp from `tinfo` it would be seeing the expected result.

keithbostic:
@agorrod, I think that I've gotten these changes into the tree -- it's a bit differently than you did, but I think it's all there.

Basically, the output of format's logging option is now groups of lines, each group starting with the transaction begin, then ending with the transaction commit/abort, and each operation in-between has a separate line in the output, as before.

What I've been doing to debug is locating the change I'm interested in, then I can find the thread's wrapping begin and end transaction lines by searching for the session address that appears in the output line's prefix.

PR #4714, opened 2019-06-20
WT-4460 Fast path for in-place modifications.

michaelcahill:
Avoid memory allocation in common paths (both for the buffer and for the array of modifications).  Always pack modifies before applying: that is the common case (during reads), so don't complicate the implementation by supporting a separate implementation for arrays of WT_MODIFY structs.  During WT_CURSOR::modify, we're going to pack them anyway in most cases.

Add macros to walk packed modifies (forward and reverse) rather than repeating the unpack logic in multiple places.

While in the area, switch to using value->mem for writes and value->data for reads.  Remove redundant session handle from methods that take a cursor.


michaelcahill:
autobuild: retest this please

michaelcahill:
evergreen retry

src/support/modify.c
```diff
@@ -8,28 +8,58 @@
...
+	for (__i = 0; __i < nentries; ++__i) {				\
+		memcpy(&mod.size, --__p, sizeof(size_t));		\
+		memcpy(&mod.offset, --__p, sizeof(size_t));		\
+		memcpy(&mod.data.size, --__p, sizeof(size_t));		\
+		mod.data.data = (__data -= mod.data.size);
+
+#define	WT_MODIFY_FOREACH_END						\
+	}								\
+} while (0)
+
```
keithbostic: Why did you decide to copy this information twice? Unless we expect large numbers of modifications in the common case (where we'd have to allocate a `WT_MODIFY` structure), why not allocate an array on the stack as the previous version did?

src/support/modify.c
```diff
@@ -179,165 +221,129 @@ __modify_apply_one(
...
-	WT_ITEM *value;
-	int i;
-	bool success;
-
-	value = &cursor->value;
+	WT_MODIFY current;
+	size_t sz;
+	const uint8_t *from;
+	uint8_t *to;
+	int delta;
```
keithbostic: Why isn't `delta` a `size_t`?

src/support/modify.c
```diff
@@ -179,165 +221,129 @@ __modify_apply_one(
...
+	WT_MODIFY_FOREACH_REVERSE(current, p, nentries, datasz) {
+		/* Move the current unmodified block into place if necessary. */
+		sz = WT_PTRDIFF(to, value->mem) -
+		    (current.offset + current.data.size);
+		from -= sz;
+		to -= sz;
+		WT_ASSERT(session, from >= (const uint8_t *)value->data &&
+		    to >= (uint8_t *)value->mem);
+		WT_ASSERT(session,
+		    from + sz <= (const uint8_t *)value->data + value->size);
```
keithbostic: I think the previous version of this code should remain. If we're overwriting fixed-size fields, we don't need to do anything other than a copy, and the check is worth doing because we can simply consume any number of changes until we hit one that isn't fixed size.

src/support/modify.c
```diff
@@ -347,86 +353,56 @@ __modify_apply(WT_SESSION_IMPL *session,
...
-	ret = __modify_apply(session, cursor, entries, (int)nentries);
+	if (fast_path && !sformat)
+		return (__modify_apply_no_overlap(
+		    session, value, p, nentries, datasz, destsz));
 
-	if (entries != _entries)
-		__wt_free(session, entries);
+	WT_MODIFY_FOREACH_BEGIN(mod, p, nentries) {
+		WT_RET(__modify_apply_one(session, value, &mod, sformat));
+	} WT_MODIFY_FOREACH_END;
```
keithbostic: I don't understand why we can't fast-path `sformat` values.

If we can't, is there any point in calling `__modify_check_fast_path()`?

michaelcahill:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

michaelcahill:
autobuild: retest this please

src/support/modify.c
michaelcahill: I'm copying it less ;-)

Seriously, the inner loop of this macro turns into 4 instructions:
```
       addq    $24, %r12       #, p
        movq    -24(%r12), %r13 # MEM[base: p_130, offset: 0B], D.42154
        movq    -16(%r12), %rbx # MEM[base: p_130, offset: 8B], D.42154
        movq    -8(%r12), %rcx  # MEM[base: p_195, offset: -8B], size
```

Previously, this was first copied to the stack, then copied out of there when needed.  There was a threshold size where `malloc` would kick in and the performance would change.  This version doesn't make a copy and so never needs to allocate extra space to hold it.

src/support/modify.c
michaelcahill: Ack, fixed.

src/support/modify.c
michaelcahill: Done.  It's a little messy to keep track of a partially applied modify, but in the common case where everything applies in place there is now a single pass.

src/support/modify.c
michaelcahill: Fixed, by moving that logic up to the calling function.

michaelcahill:
I've made the suggested changes, so I'll merge this into the original branch so we have one place to look at for final review.

PR #4715, opened 2019-06-20
WT-4716 Add recovery_compressor option for logging

tetsuo-cpp:
This PR adds some functionality to try each available compressor after failing to decompress a log record with the currently used compressor.

### DIFF
--- a/src/conn/conn_log.c
+++ b/src/conn/conn_log.c
@@ -254,6 +254,11 @@ __logmgr_config(
 		WT_RET(__wt_compressor_config(
 		    session, &cval, &conn->log_compressor));
 
+		WT_RET(__wt_config_gets_none(
+		    session, cfg, "log.recovery_compressor", &cval));
+		WT_RET(__wt_recovery_compressor_config(session, &cval,
+		    conn->log_compressor, &conn->recovery_log_compressor));
+
 		WT_RET(__wt_config_gets(session, cfg, "log.path", &cval));
 		WT_RET(__wt_strndup(
 		    session, cval.str, cval.len, &conn->log_path));
--- a/src/include/connection.h
+++ b/src/include/connection.h
@@ -372,6 +372,8 @@ struct __wt_connection_impl {
 	bool		 log_wrlsn_tid_set;/* Log write lsn thread set */
 	WT_LOG		*log;		/* Logging structure */
 	WT_COMPRESSOR	*log_compressor;/* Logging compressor */
+	/* Recovery logging compressor */
+	WT_COMPRESSOR	*recovery_log_compressor;
 	uint32_t	 log_cursors;	/* Log cursor count */
 	wt_off_t	 log_dirty_max;	/* Log dirty system cache max size */
 	wt_off_t	 log_file_max;	/* Log file max size */
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -703,7 +703,7 @@ __log_decompress(WT_SESSION_IMPL *session, WT_ITEM *in, WT_ITEM *out)
 	WT_NAMED_COMPRESSOR *ncomp;
 	size_t result_len, skip;
 	uint32_t uncompressed_size;
-	bool skipped_active_compressor;
+	bool successful;
 
 	conn = S2C(session);
 	logrec = (WT_LOG_RECORD *)in->mem;
@@ -735,16 +735,16 @@ __log_decompress(WT_SESSION_IMPL *session, WT_ITEM *in, WT_ITEM *out)
 		if (WT_STREQ(ncomp->name, "nop"))
 			nop_compressor = ncomp->compressor;
 
-	skipped_active_compressor = true;
-
 	/* Try the active compressor provided that it's not nop. */
+	successful = false;
 	if (compressor != NULL && compressor->decompress != NULL &&
 	    compressor != nop_compressor) {
 		ret = compressor->decompress(compressor, &session->iface,
 		    (uint8_t *)in->mem + skip, in->size - skip,
 		    (uint8_t *)out->mem + skip,
 		    uncompressed_size - skip, &result_len, false);
-		skipped_active_compressor = false;
+		if (ret == 0)
+			successful = true;
 	}
 
 	/*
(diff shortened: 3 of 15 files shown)


src/log/log.c
```diff
@@ -712,10 +714,28 @@ __log_decompress(WT_SESSION_IMPL *session, WT_ITEM *in, WT_ITEM *out)
...
+	    uncompressed_size - skip, &result_len);
+
+	/*
+	 * If our chosen compressor doesn't work, we may have been running with
+	 * a different compressor last time. Cycle through the available ones
+	 * and see if one of them succeeds.
+	 */
+	if (ret != 0)
+		TAILQ_FOREACH(ncomp, &conn->compqh, q) {
+			ret = ncomp->compressor->decompress(compressor,
```
tetsuo-cpp: Each failed call results in error messages. Do we want to suppress this?

test/suite/test_log04.py
```diff
@@ -0,0 +1,101 @@
...
+
+        self.conn = self.wiredtiger_open(
+            self.homedir, self.make_config_string(self.init_compress))
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(1))
+        self.session = self.conn.open_session('isolation=snapshot')
+        self.populate()
+        self.session.close()
+        self.conn.close()
+
+        with self.expectedStderrPattern(""):
```
tetsuo-cpp: TODO: Fill this in with an actual pattern.

tetsuo-cpp:
I'm having some issues running the tests with compressor extensions. It seems that the testing scripts expect to see a shared lib for each extension however, I get a `.o` for each extension that gets statically linked into `libwiredtiger.so`.

As you can see from the `ldd` output, it dynamically links to `libsnappy` whereas I think I need to it to link to something like `libwiredtiger_snappy.so` (if I'm understanding this script correctly).
```
alexc@kitty:~/work/wiredtiger/build_posix/.libs$ ldd libwiredtiger.so
	linux-vdso.so.1 (0x00007ffddabf1000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fbb5ed62000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fbb5eb5e000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fbb5e93f000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fbb5e722000)
	libsnappy.so.1 => /usr/lib/x86_64-linux-gnu/libsnappy.so.1 (0x00007fbb5e51a000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fbb5e129000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fbb5f384000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fbb5dda0000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fbb5db88000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fbb5d7ea000)
```

Do you know how to get it to build in this way? I've had a look at Jenkins and it's running `configure` the same way I am so I assume it's skipping extension specific tests. And I can't see anything in the `configure` script that allows me to specify this. To my knowledge, `--enable-shared` and `--disable-static` refers to whether it generates a `libwiredtiger.so` or `libwiredtiger.a` not the extension libraries.

src/log/log.c
keithbostic: > Each failed call results in error messages. Do we want to suppress this?

Yeah, I think so, it's "expected" behavior.

Of course, we'll want an error on our final failure.

test/suite/test_log04.py
tetsuo-cpp: No longer relevant.

tetsuo-cpp:
@keithbostic Sorry, things have gotten pretty ugly since you LGTM'd. Could you please take another look when you get the chance? I've done some work to suppress decompression errors and added a workaround for the nop compressor (which doesn't really behave like a real compressor).

tetsuo-cpp:
Unfortunately, the checksum attached to the log record is from the on-disk data (after compression and encryption). So it doesn't help us determine whether we've decrypted properly.

From looking at the code, I think that decision was arbitrary and there isn't any reason that I can't just generate the checksum from the uncompressed and unencrypted data.

src/log/log.c
```diff
@@ -705,7 +705,7 @@ __log_decompress(WT_SESSION_IMPL *session, WT_ITEM *in, WT_ITEM *out)
 	conn = S2C(session);
 	logrec = (WT_LOG_RECORD *)in->mem;
 	skip = WT_LOG_COMPRESS_SKIP;
-	compressor = conn->log_compressor;
+	compressor = conn->recovery_log_compressor;
```
sueloverso: This code should use the correct compressor based on whether or not we're in recovery. Something like a `printlog` would be decompressing log records and should not use the recovery one.

src/conn/conn_log.c
```diff
@@ -254,6 +254,11 @@ __logmgr_config(
 		WT_RET(__wt_compressor_config(
 		    session, &cval, &conn->log_compressor));
 
+		WT_RET(__wt_config_gets_none(
```
sueloverso: I would add a comment that this code must be called after the above code so that the `log_compressor` is known to be set up.

src/include/connection.h
```diff
@@ -372,6 +372,8 @@ struct __wt_connection_impl {
 	bool		 log_wrlsn_tid_set;/* Log write lsn thread set */
 	WT_LOG		*log;		/* Logging structure */
 	WT_COMPRESSOR	*log_compressor;/* Logging compressor */
+	/* Recovery logging compressor */
+	WT_COMPRESSOR	*recovery_log_compressor;
```
sueloverso: I'd prefer naming this `log_recovery_compressor` to keep the `log_` prefix.

test/suite/test_log04.py
```diff
@@ -0,0 +1,104 @@
...
+    init_compression_values = [
+        ('none', dict(init_compress=None)),
+        ('nop', dict(init_compress='nop')),
+        ('lz4', dict(init_compress='lz4')),
+        ('snappy', dict(init_compress='snappy')),
+        ('zlib', dict(init_compress='zlib')),
+        ('zstd', dict(init_compress='zstd'))
+    ]
+    after_compression_values = [
+        ('none', dict(after_compress=None)),
```
sueloverso: This doesn't test the explicit `none` string as opposed to the empty string to the API. I think we need that too.

tetsuo-cpp:
As discussed in the ticket, we're not going ahead with this change right now.

PR #4716, opened 2019-06-21
WT-4878 Disable random dhandle selection for eviction

sulabhM:
Disable random selection till we get more testing and confidence with
that change.

Fix handle lock test to reset cursors before caching them. Otherwise,
for configurations with many threads it can end up in a state
where there is a cursor on every page, preventing the pages from
splitting, and ultimately blocking eviction.  This was exposed by
changes to eviction, but is a structural problem with the test case.


michaelcahill:
Thanks @sulabhM, this looks like a reasonable step to me.  I'll merge so we get more testing (including perf tests).

PR #4719, opened 2019-06-22
WT-4881 don't allow reconciliation to be re-entered

keithbostic:


keithbostic:
I'm pretty sure this is wrong, closing and discarding the branch.

PR #4720, opened 2019-06-24
WT-4867 Skip assert when we see reserved updates

luke-pearson:
This fixes the failure seen in test/format where our update chain has the following pattern:
update2 -> reserved -> update1. Originally we were not setting the key repeated flag on update2 as we saw txnid = WT_TXN_ABORTED on the reserved update as it was getting aborted during the prepare call. The fix I have implemented is to introduce a new flag which specifies that we have seen a reserved update on the modification associated with update2 and as such we skip the assertion in commit and rollback that checks if resolved_update_count == visited_update_count.


luke-pearson:
retest this please

src/txn/txn.c
```diff
@@ -1122,8 +1128,14 @@ __wt_txn_prepare(WT_SESSION_IMPL *session, const char *cfg[])
...
 
-	/* Prepare updates. */
-	for (i = 0, op = txn->mod; i < txn->mod_count; i++, op++) {
+	/*
+	 * Prepare updates, traverse the modification array in reverse order
+	 * so that we visit the update chain in newest to oldest order
+	 * allowing us to set the key repeated flag with reserved updates in
+	 * the chain.
+	 */
+	for (i = txn->mod_count; i > 0; i--, op--) {
```
sulabhM: dont need "op--".

src/txn/txn.c
```diff
@@ -803,13 +803,13 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	int64_t resolved_update_count, visited_update_count;
 	uint32_t fileid;
 	u_int i;
-	bool locked, prepare, readonly, update_timestamp;
+	bool locked, prepare, readonly, update_timestamp, skip_update_assert;
```
sulabhM: Small nit pick: Ideally, variable names should be in the alphabetic order. Sorry I did not notice before.

agorrod:
@bvpvamsikrishna please pick up the code review here.

luke-pearson:
@bvpvamsikrishna has request I add a python unit test to test this functionality. I'll lock it until that is implemented.

bvpvamsikrishna:
@luke-pearson, can we discuss the logic to account for reserved updates instead of skipping the assert.

PR #4721, opened 2019-06-24
WT-4878 Disable random dhandle selection for eviction (#4716)

sulabhM:
This mostly goes back to the state before data handles were chosen randomly to search for eviction candidates.  It also changes some of the calculations for how many pages to enqueue per tree.

Also fix handle lock test to reset cursors before caching them. Otherwise, for configurations with many threads it can end up in a state where there is a cursor on every page, preventing the pages from splitting, and ultimately blocking eviction.  This was exposed by changes to eviction, but is a structural problem with the test case.

(cherry picked from commit 859c527b01143c0f3d1a064b4ae14592c65628d1)


sulabhM:
4.2 backports are being manually merged by Luke outside the pull-req process, so closing this one.

PR #4724, opened 2019-06-25
WT-4881 Don't allow reconciliation to be re-entered

keithbostic:


keithbostic:
@agorrod 
> Does it make sense to switch stress testing over here before merging?

I think this is a relatively safe merge, as these things go. I'll kick off a local stress run, if it doesn't fail in the next 30 minutes, I think we're good to go.

PR #4732, opened 2019-06-30
WT-4892 forced eviction statistics are misleading

keithbostic:
@agorrod, this branch continues to track forced evictions caused by large numbers of deleted items, but doesn't track the time those forced evictions required separately from the time other forced evictions required.


dist/stat_data.py
```diff
@@ -222,7 +222,6 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_eviction_fail', 'pages selected for eviction unable to be evicted'),
     CacheStat('cache_eviction_force', 'pages evicted because they exceeded the in-memory maximum count'),
     CacheStat('cache_eviction_force_delete', 'pages evicted because they had chains of deleted items count'),
-    CacheStat('cache_eviction_force_delete_time', 'pages evicted because they had chains of deleted items time (usecs)'),
```
agorrod: I think this change is worthwhile, but I'm inclined to approach it a bit differently. I'd like to end up with:

`pages queued for eviction because they had chains of deleted items` (I believe this matches the code change you made).
`pages evicted immediately that were clean` (along with matching time spent stat)
`pages evicted immediately that were dirty` (along with matching time spent stat)

I'd be fine with replacing `immediately` with `force`, or something more verbose describing `due to reasons other than normal cache management`

The reason I care is that tracking all pages that are evicted inline is valuable.

keithbostic: @agorrod, I've pushed this change in c1e7a87.

It seemed cleanest to move the forcible eviction statistics out of `__wt_page_release_evict` and into `__wt_evict` because the latter function already knows if the evicted page was clean or dirty, and updates statistics based on that fact. The one thing this changes is we won't include forcible eviction requests resulting in in-memory splits as failed or successful forcible evictions.

Let me know if you want further changes or I've misunderstood something.

src/evict/evict_page.c
```diff
@@ -73,47 +67,20 @@ __wt_page_release_evict(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
...
 
-	/*
-	 * Track how long the call to evict took. If eviction is successful then
-	 * we have one of two pairs of stats to increment.
-	 */
+	(void)__wt_atomic_addv32(&btree->evict_busy, 1);
 	ret = __wt_evict(session, ref, previous_state, evict_flags);
-	time_stop = __wt_clock(session);
-	if (ret == 0) {
-		if (too_big) {
```
agorrod: Did you have to push this down a level to know reliably whether the page was clean or dirty?

agorrod:
I hope that you don't mind that I pushed a change directly here - it seemed more efficient than using words to describe a change to some text.

src/evict/evict_page.c
```diff
@@ -171,6 +139,17 @@ __wt_evict(WT_SESSION_IMPL *session,
 		__wt_evict_list_clear_page(session, ref);
 	}
 
+	/*
+	 * Track how long forcible eviction took. Immediately increment the
+	 * forcible eviction statistics, we might do an in-memory split and
+	 * not an eviction, which skips the other statistics.
+	 */
+	time_start = 0;			/* [-Wconditional-uninitialized] */
+	if (LF_ISSET(WT_EVICT_CALL_URGENT)) {
```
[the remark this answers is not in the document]
keithbostic: IIRC, I vaguely recall thinking that since it was likely to fail (especially in the new world of immediate evictions), it wasn't part of the cost of eviction. I don't feel strongly at all, feel free to simply include that work in the timing without further discussion.

src/evict/evict_page.c
keithbostic: We can figure it out at a higher level (by calling `__wt_page_evict_clean()`), but that function is a kluge, not excited about using it in more places.

More importantly, all the other statistics live in `__wt_evict()` and there's no advantage in in-lining that code in a bunch of places.

In summary, it just seemed cleaner to handle all the statistics together in the same place.

keithbostic:
> I hope that you don't mind that I pushed a change directly here - it seemed more efficient than using words to describe a change to some text.

Oh, no, not at all. This is your change, I'm just helping out. :)

dist/stat_data.py
```diff
@@ -220,13 +220,14 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_eviction_dirty', 'modified pages evicted'),
     CacheStat('cache_eviction_empty_score', 'eviction empty score', 'no_clear,no_scale'),
     CacheStat('cache_eviction_fail', 'pages selected for eviction unable to be evicted'),
-    CacheStat('cache_eviction_force', 'pages evicted because they exceeded the in-memory maximum count'),
-    CacheStat('cache_eviction_force_delete', 'pages evicted because they had chains of deleted items count'),
-    CacheStat('cache_eviction_force_delete_time', 'pages evicted because they had chains of deleted items time (usecs)'),
-    CacheStat('cache_eviction_force_fail', 'failed eviction of pages that exceeded the in-memory maximum count'),
-    CacheStat('cache_eviction_force_fail_time',  'failed eviction of pages that exceeded the in-memory maximum time (usecs)'),
```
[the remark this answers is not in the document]
keithbostic: Sure, pushed in 9cd8c8b

keithbostic:
> Is there a reason why you chose to exclude the time spent getting exclusive access?

Adding `cache_eviction_force_fail_time` back in made it a lot easier to include the time spent getting exclusive access, so that's what I did.

PR #4737, opened 2019-07-02
WT-4900 Implement all_durable timestamp

tetsuo-cpp:
This PR is to add an `all_durable` timestamp to the WiredTiger API.
This timestamp is designed to replace `all_committed`. We'll keep `all_committed` for the time and then remove it after MongoDB has made the corresponding server change to use `all_durable`.


src/txn/txn_timestamp.c
```diff
@@ -1151,6 +1219,124 @@ __wt_txn_clear_commit_timestamp(WT_SESSION_IMPL *session)
 	WT_PUBLISH(txn->flags, flags);
 }
 
+/*
```
tetsuo-cpp: This section is more or less copied verbatim from the `commit` versions of these functions.

test/suite/test_timestamp14.py
```diff
@@ -238,6 +257,72 @@ def test_pinned_oldest(self):
...
+        # with a lower durable timestamp than has previously been committed.
+        session1.begin_transaction()
+        cur1[1] = 4
+        session1.prepare_transaction('prepare_timestamp=3')
+        session1.timestamp_transaction('commit_timestamp=4')
+        session1.timestamp_transaction('durable_timestamp=5')
+        self.assertTimestampsEqual(
+            self.conn.query_timestamp('get=all_committed'), '3')
+        self.assertTimestampsEqual(
+            self.conn.query_timestamp('get=all_durable'), '4')
```
tetsuo-cpp: As you can see, the API still allows `all_durable` to move backwards which is the same issue that server has with `all_committed`.

Not sure who can answer this, but is this a problem? I was talking to @bvpvamsikrishna earlier and we were thinking that perhaps the way server uses timestamps may mean that `all_durable` will only go forwards even though the WiredTiger API doesn't enforce that.

michaelcahill: It's worth double checking with Judah, but I think this is okay.  The differences between this test case and the server are (a) the server allocates durable timestamps monotonically, and (b) even if it didn't, we constrain the durable timestamp for a transaction to be newer than the system-wide stable timestamp, which will also prevent the jumps that are problematic `all_committed`.

dist/api_data.py
```diff
@@ -1569,7 +1569,7 @@ def __ge__(self, other):
         stable checkpoint taken prior to a shutdown and \c stable returns the
         most recent \c stable_timestamp set with WT_CONNECTION::set_timestamp.
         See @ref transaction_timestamps''',
-        choices=['all_committed','last_checkpoint',
+        choices=['all_committed','all_durable','last_checkpoint',
```
sueloverso: I looked at the ticket and Eric asked for the naming to be something different than `all_durable`. What that request considered?

src/include/txn.h
```diff
@@ -135,10 +135,12 @@ struct __wt_txn_global {
 	wt_timestamp_t pinned_timestamp;
 	wt_timestamp_t recovery_timestamp;
 	wt_timestamp_t stable_timestamp;
+	wt_timestamp_t durable_timestamp;
 	bool has_commit_timestamp;
 	bool has_oldest_timestamp;
 	bool has_pinned_timestamp;
 	bool has_stable_timestamp;
+	bool has_durable_timestamp;
```
sueloverso: Please alphabetize both of these new fields in their respective groupings.

src/include/txn.h
```diff
@@ -160,6 +162,11 @@ struct __wt_txn_global {
 	TAILQ_HEAD(__wt_txn_rts_qh, __wt_txn) read_timestamph;
 	uint32_t read_timestampq_len;
 
+	/* List of transactions sorted by durable timestamp. */
+	WT_RWLOCK durable_timestamp_rwlock;
+	TAILQ_HEAD(__wt_txn_dts_qh, __wt_txn) durable_timestamph;
+	uint32_t durable_timestampq_len;
```
sueloverso: Same alphabetize comment here.

src/include/txn.h
```diff
@@ -310,8 +317,11 @@ struct __wt_txn {
 
 	TAILQ_ENTRY(__wt_txn) commit_timestampq;
 	TAILQ_ENTRY(__wt_txn) read_timestampq;
+	TAILQ_ENTRY(__wt_txn) durable_timestampq;
 	bool clear_commit_q;	/* Set if need to clear from the commit queue */
 	bool clear_read_q;	/* Set if need to clear from the read queue */
+	/* Set if need to clear from the durable queue */
+	bool clear_durable_q;
```
sueloverso: And for both fields here too.

src/txn/txn.c
```diff
@@ -800,11 +801,13 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_TXN_GLOBAL *txn_global;
 	WT_TXN_OP *op;
 	WT_UPDATE *upd;
-	wt_timestamp_t prev_commit_timestamp;
+	wt_timestamp_t candidate_durable_timestamp, prev_commit_timestamp;
+	wt_timestamp_t prev_durable_timestamp;
```
sueloverso: I'd probably split these to put the 2 `prev_` together on the same line and leave the `candidate` one solo.

src/txn/txn.c
```diff
@@ -1068,6 +1083,31 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
+	if (candidate_durable_timestamp != 0) {
+		prev_durable_timestamp = txn_global->durable_timestamp;
+		update_durable_ts =
+		    candidate_durable_timestamp > prev_durable_timestamp;
+	}
+
+	/*
+	 * If it looks like we'll need to move the global durable timestamp,
+	 * write lock and re-check.
+	 */
```
sueloverso: The comment refers to taking a write lock but the code does not do that.

src/txn/txn.c
```diff
@@ -1458,6 +1498,8 @@ __wt_txn_stats_update(WT_SESSION_IMPL *session)
 	    stats, txn_commit_queue_len, txn_global->commit_timestampq_len);
 	WT_STAT_SET(session,
 	    stats, txn_read_queue_len, txn_global->read_timestampq_len);
+	WT_STAT_SET(session,
+	    stats, txn_durable_queue_len, txn_global->durable_timestampq_len);
```
sueloverso: Again alphabetize this line with its related ones.

src/txn/txn.c
```diff
@@ -1520,6 +1562,10 @@ __wt_txn_global_init(WT_SESSION_IMPL *session, const char *cfg[])
 	    &txn_global->read_timestamp_rwlock, read_timestamp);
 	TAILQ_INIT(&txn_global->read_timestamph);
 
+	WT_RWLOCK_INIT_TRACKED(session,
+	    &txn_global->durable_timestamp_rwlock, durable_timestamp);
+	TAILQ_INIT(&txn_global->durable_timestamph);
```
sueloverso: And here too.

src/txn/txn.c
```diff
@@ -1553,6 +1599,7 @@ __wt_txn_global_destroy(WT_SESSION_IMPL *session)
 	__wt_rwlock_destroy(session, &txn_global->rwlock);
 	__wt_rwlock_destroy(session, &txn_global->commit_timestamp_rwlock);
 	__wt_rwlock_destroy(session, &txn_global->read_timestamp_rwlock);
+	__wt_rwlock_destroy(session, &txn_global->durable_timestamp_rwlock);
```
sueloverso: And here...

src/txn/txn.c
```diff
@@ -1684,6 +1731,9 @@ __wt_verbose_dump_txn(WT_SESSION_IMPL *session)
 	    __wt_timestamp_to_string(txn_global->pinned_timestamp, ts_string)));
 	WT_RET(__wt_msg(session, "stable timestamp: %s",
 	    __wt_timestamp_to_string(txn_global->stable_timestamp, ts_string)));
+	WT_RET(__wt_msg(session, "durable timestamp: %s",
+	    __wt_timestamp_to_string(
+	    txn_global->durable_timestamp, ts_string)));
```
sueloverso: And here.

dist/api_data.py
tetsuo-cpp: Not yet. I wasn't confident writing this code and anticipated that I might be told to go back to the drawing board.

Since it seems to be going somewhat in the right direction, now's probably a good time to have that discussion. The rename itself should be trivial. I'll leave a comment on the ticket.

src/include/txn.h
tetsuo-cpp: 955effc

src/include/txn.h
tetsuo-cpp: 955effc

src/include/txn.h
tetsuo-cpp: 955effc

src/txn/txn.c
tetsuo-cpp: 955effc

src/txn/txn.c
tetsuo-cpp: 955effc

Changed to "attempt atomic cas" and also fixed the corresponding comment for the commit timestamp above.

src/txn/txn.c
tetsuo-cpp: 955effc

src/txn/txn.c
tetsuo-cpp: 955effc

src/txn/txn.c
tetsuo-cpp: 955effc

src/txn/txn.c
tetsuo-cpp: 955effc

src/txn/txn_timestamp.c
```diff
@@ -273,6 +273,73 @@ __txn_global_query_timestamp(
 		 */
 		if (ts == 0)
 			return (WT_NOTFOUND);
+	} else if (WT_STRING_MATCH("all_durable", cval.str, cval.len)) {
```
tetsuo-cpp: Any opinions on how I've done this?

I first check the queue of transactions ordered by durable timestamp which contains only transactions that explicitly have a durable timestamp set. Then I go and check the queue ordered by commit timestamp and look for the first one that doesn't have a durable timestamp (and is therefore non-prepared).

The other alternative is to put everything in the durable queue and have it ordered by `x` where `x` is `durable_timestamp` for prepared transactions and `commit_timestamp` for non-prepared transactions. That way, I can just check the one queue and avoid multiple rounds of locking. I didn't go for that approach since it seems more troublesome to maintain the queues and that setting a commit_timestamp will mean locking and inserting into two different queues.

tetsuo-cpp:
evergreen retry

dist/api_data.py
michaelcahill: I followed up on the ticket, but absent someone's great idea, I'm inclined to stick with "all_durable".  It made sense to the replication team, and since we're returning the maximum timestamp that is durable (one less than the first hole), everything up to and including this timestamp is durable.

src/txn/txn_timestamp.c
michaelcahill: Do we need the commit timestamp queue for anything other than answering "all_committed" queries?  I'm assuming that will go away once we've updated MongoDB to use "all_durable"...

tetsuo-cpp: Yeah thanks for pointing that out. The commit timestamp queue is there purely for `all_committed`. In that case, it makes sense to not rely on that queue so it can be removed later.

tetsuo-cpp: 713350c

Again, I'm unsure about a lot of this code.
So if you could take a detailed look that'd be great.

bvpvamsikrishna:
Below approach is simpler. 
When the durable timestamp is set for a transaction, add the durable timestamp to commit Q, and remove the commit timestamp of that transaction from the commit Q. This change will satisfy the requirement of all_durable. At this point query_timestamp("all_committed") and query_timestamp("all_durable") will return the same value. 
Then we find & replace the "all_committed" with the "all_durable".
Remove the support of "all_committed".

michaelcahill:
@bvpvamsikrishna, I think that's fine -- the only issue would be if MongoDB calls `timestamp_transaction` with a commit timestamp, then calls it again with a durable timestamp.  If `all_durable` is queried concurrently, we never want it to include the commit timestamp.

tetsuo-cpp:
@bvpvamsikrishna 
My solution will be to make `all_committed` use the `all_durable` logic. This way I can remove the commit queue in this PR and have everything using the durable queue.

@michaelcahill 
Not sure if this helps but I can make an exception where if someone specifies a commit timestamp, I deliberately don't insert it into the queue if it's prepared (knowing that in the future, I'm going to get a durable timestamp). Does that help the issue at all?

michaelcahill:
@tetsuo-cpp I think that ignoring commit timestamps for prepared transactions is correct.

tetsuo-cpp:
retest this please

tetsuo-cpp:
@michaelcahill @bvpvamsikrishna 
Apologies for the delay. I got a clean MongoDB patch build from this latest set of changes.
This is ready to go in whenever you're happy with it.

src/txn/txn.c
```diff
@@ -644,7 +644,12 @@ __wt_txn_release(WT_SESSION_IMPL *session)
 	/* Ensure the transaction flags are cleared on exit */
 	txn->flags = 0;
 	txn->prepare_timestamp = WT_TS_NONE;
-	txn->durable_timestamp = WT_TS_NONE;
+
+	/*
+	 * Purposely do NOT clear the commit and durable timestamps on release.
+	 * Other readers may still find these transactions in the durable queue
+	 * and will need to see those timestamps.
+	 */
```
agorrod: It's odd to have a comment not associated with any code block. I'd join this with the comment about clearing flags at line 644.

src/txn/txn_timestamp.c
```diff
@@ -1047,7 +1105,15 @@ __wt_txn_publish_commit_timestamp(WT_SESSION_IMPL *session)
...
 
-	if (F_ISSET(txn, WT_TXN_PUBLIC_TS_COMMIT))
+	/*
+	 * If we know for a fact that this is a prepared transaction, don't
+	 * add to the durable queue. If we poll all_durable after setting the
+	 * commit timestamp of a prepared transaction, that prepared transaction
+	 * should NOT be visible.
+	 */
+	if (F_ISSET(txn, WT_TXN_PREPARE))
+		return;
```
agorrod: Do we need to remove an entry from the durable queue when preparing a transaction?

src/txn/txn_timestamp.c
```diff
@@ -1056,98 +1122,207 @@ __wt_txn_publish_commit_timestamp(WT_SESSION_IMPL *session)
...
 	ts = txn->commit_timestamp;
+	txn->first_commit_timestamp = ts;
 
-	__wt_writelock(session, &txn_global->commit_timestamp_rwlock);
+	/*
+	 * Provided that we don't already have an explicit durable timestamp set
+	 * (indicated by WT_TXN_HAS_TS_DURABLE) and we haven't published a
+	 * durable timestamp yet (indicated by WT_TXN_PUBLIC_TS_DURABLE), let's
+	 * insert the txn in the durable queue using its commit timestamp as an
+	 * implied durable timestamp.
```
agorrod: This comment doesn't seem to match the code - since the code isn't checking for `WT_TXN_PUBLIC_TS_DURABLE`. It's also generally preferable to avoid referencing variable names in comments - they should be standalone description of the functionality not a restatement of the code logic.

src/txn/txn_timestamp.c
```diff
@@ -1056,98 +1122,207 @@ __wt_txn_publish_commit_timestamp(WT_SESSION_IMPL *session)
...
+
+	txn = &session->txn;
+	txn_global = &S2C(session)->txn_global;
+
+	if (F_ISSET(txn, WT_TXN_PUBLIC_TS_DURABLE))
+		return;
+
+	/*
+	 * Copy the current durable timestamp (which can change while the
+	 * transaction is running).
```
agorrod: Is that true? I'm surprised that another thread could change the durable timestamp for a transaction. I guess it could when it's being cleared from the durable queue - but I doubt that should race with this publish?

src/txn/txn_timestamp.c
```diff
@@ -1056,98 +1122,207 @@ __wt_txn_publish_commit_timestamp(WT_SESSION_IMPL *session)
...
+	F_SET(txn, WT_TXN_PUBLIC_TS_DURABLE);
+	__wt_writeunlock(session, &txn_global->durable_timestamp_rwlock);
+}
+
+/*
+ * __wt_txn_publish_durable_timestamp --
+ *	Publish a transaction's durable timestamp.
+ */
+void
+__wt_txn_publish_durable_timestamp(WT_SESSION_IMPL *session)
```
agorrod: Should this be removed and __wt_txn_publish_commit_timestamp renamed instead?

dist/api_data.py
```diff
@@ -1582,6 +1583,14 @@ def __ge__(self, other):
         where the application is rolling back locally committed transactions.
         The supplied value must not be older than the current oldest and
         stable timestamps.  See @ref transaction_timestamps'''),
+    Config('durable_timestamp', '', r'''
```
agorrod: I know the plan is to remove commit_timestamp from here - but should we mark it as deprecated and update the documentation to say that it's a synonym for durable_timestamp, in case we don't end up doing that soon?

src/txn/txn_timestamp.c
```diff
@@ -743,7 +796,7 @@ __wt_txn_set_commit_timestamp(
 		}
 	}
 
-	WT_ASSERT(session, txn->durable_timestamp == WT_TS_NONE ||
+	WT_ASSERT(session, !F_ISSET(txn, WT_TXN_HAS_TS_DURABLE) ||
 	    txn->durable_timestamp == txn->commit_timestamp);
 	txn->durable_timestamp = txn->commit_timestamp = commit_ts;
```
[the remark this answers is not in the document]
tetsuo-cpp: 3e0eab0

src/txn/txn_timestamp.c
```diff
@@ -420,22 +459,36 @@ __wt_txn_update_pinned_timestamp(WT_SESSION_IMPL *session, bool force)
...
-	    cfg, "commit_timestamp", 0, &commit_cval));
-	has_commit = commit_cval.len != 0;
-	if (has_commit)
-		WT_STAT_CONN_INCR(session, txn_set_ts_commit);
+	    cfg, "commit_timestamp", 0, &durable_cval));
+	has_durable = durable_cval.len != 0;
+	if (has_durable)
+		WT_STAT_CONN_INCR(session, txn_set_ts_durable);
+
+	if (!has_durable) {
```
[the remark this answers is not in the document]
tetsuo-cpp: As discussed, it's just a temporary fix to allow either `commit_timestamp` or `durable_timestamp` to be set on the connection level.

src/txn/txn_timestamp.c
```diff
@@ -221,6 +221,40 @@ __wt_txn_get_pinned_timestamp(
...
+	 * explicitly set as opposed to simply being copied over from the commit
+	 * timestamp value.
+	 */
+	if (txn->durable_timestamp != WT_TS_NONE &&
+	    txn->durable_timestamp != txn->commit_timestamp)
+		return (txn->durable_timestamp);
+	if (txn->first_commit_timestamp != WT_TS_NONE)
+		return (txn->first_commit_timestamp);
+
+	return (0);
```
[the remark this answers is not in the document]
tetsuo-cpp: 3e0eab0

Added assertion.

dist/api_data.py
tetsuo-cpp: 3e0eab0

Marked as deprecated in docs.

src/txn/txn_timestamp.c
tetsuo-cpp: 3e0eab0

Merged into a single function `__wt_txn_publish_timestamp`.

src/txn/txn_timestamp.c
tetsuo-cpp: 3e0eab0

That was a leftover from when I copied from the `commit` variant of this function.
Since I've merged the two, this doesn't exist anymore.

src/txn/txn.c
tetsuo-cpp: 3e0eab0

Moved the comment.

src/txn/txn_timestamp.c
tetsuo-cpp: 3e0eab0

This part was refactored and the comment itself was removed as it was no longer relevant.

src/txn/txn_timestamp.c
tetsuo-cpp: The sequence for a prepared transaction must be:
- Set prepare ts.
- Set commit ts.
- Set durable ts.
- Set more commit ts (optional).

This is to protect us from inserting into the queue at either commit timestamp. The first one will hit this condition. Any that come after the durable timestamp will hit the `WT_TXN_TS_PUBLISHED` check further up.

tetsuo-cpp:
@bvpvamsikrishna @agorrod @michaelcahill
Pushed suggested fixes from Alex and Vamsi. Once we're happy to merge, I'll run another patch build since there's been a slight refactor since the last one.

src/txn/txn.c
```diff
@@ -1047,25 +1053,28 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
...
 
-	/* First check if we've already committed something in the future. */
-	if (update_timestamp) {
-		prev_commit_timestamp = txn_global->commit_timestamp;
-		update_timestamp =
-		    txn->commit_timestamp > prev_commit_timestamp;
+	/* First check if we've made something durable in the future. */
+	update_durable_ts = false;
+	prev_durable_timestamp = 0;
+	if (candidate_durable_timestamp != 0) {
```
agorrod: Should use WT_TS_NONE

src/txn/txn_timestamp.c
```diff
@@ -240,36 +278,38 @@ __txn_global_query_timestamp(
...
+		 * Compare with the least recently durable transaction.
+		 */
+		__wt_readlock(session, &txn_global->durable_timestamp_rwlock);
+		TAILQ_FOREACH(txn, &txn_global->durable_timestamph,
+		    durable_timestampq) {
+			if (txn->clear_durable_q)
 				continue;
 
-			tmpts = txn->first_commit_timestamp;
-			WT_ASSERT(session, tmpts != 0);
```
agorrod: This seems like a worthwhile check - we are returning an unsigned value and decrementing below 0 will mean that we return something wrapped to max instead. I know it was only being done in diagnostic mode before - we should probably make this more robust for release builds as well.

src/txn/txn.c
tetsuo-cpp: a6a10f4

src/txn/txn_timestamp.c
tetsuo-cpp: This assertion still exists but it has been moved within the `__txn_get_published_timestamp` function.

As for making it more robust in release mode, what would you expect WiredTiger to do if it sees something like this?

agorrod: I think it'd be better to return 0 rather than UINT_MAX.

tetsuo-cpp: As discussed, `all_durable` won't return UINT_MAX here. It'll return `txn_global->durable_timestamp`.

It's hard to make this robust since we call this function when inserting into the queue also and there's no reasonable way to recover from that. I think the only possibilities are to leave it as an assertion or to invoke a panic here.

tetsuo-cpp:
retest this please

agorrod:
autobuild: retest this please

PR #4738, opened 2019-07-02
WT-4791 Guard against deference null return (coverity)

kommiharibabu:
CID 112861 (#1 of 1): Deference null return

The reported coverity problem is not possible, but in the same
flow with a different branch flow the issue can happen, so added
a NULL check.

### DIFF
--- a/src/btree/bt_curprev.c
+++ b/src/btree/bt_curprev.c
@@ -49,8 +49,7 @@ restart:
 	 * with a search.
 	 */
 	recno = WT_INSERT_RECNO(cbt->ins);
-	while (((current = cbt->ins) != PREV_INS(cbt, 0)) &&
-	    (current != NULL)) {
+	while ((current = cbt->ins) != PREV_INS(cbt, 0)) {
 		if (cbt->btree->type == BTREE_ROW) {
 			key.data = WT_INSERT_KEY(current);
 			key.size = WT_INSERT_KEY_SIZE(current);
--- a/src/btree/bt_curprev.c
+++ b/src/btree/bt_curprev.c
@@ -47,7 +47,8 @@ restart:
 	 * If the search stack does not point at the current item, fill it in
 	 * with a search.
 	 */
-	while (((current = cbt->ins) != PREV_INS(cbt, 0)) && (current != NULL)) {
+	while (((current = cbt->ins) != PREV_INS(cbt, 0)) &&
+	    (current != NULL)) {
 		if (cbt->btree->type == BTREE_ROW) {
 			key.data = WT_INSERT_KEY(current);
 			key.size = WT_INSERT_KEY_SIZE(current);
--- a/src/btree/bt_curprev.c
+++ b/src/btree/bt_curprev.c
@@ -38,6 +38,7 @@ __cursor_skip_prev(WT_CURSOR_BTREE *cbt)
 	WT_INSERT *current, *ins;
 	WT_ITEM key;
 	WT_SESSION_IMPL *session;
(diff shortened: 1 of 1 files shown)


agorrod:
@keithbostic could you please do a review of this branch when it's convenient?

kommiharibabu:
@keithbostic, During my analysis as per the above suggestion, passing always the first `WT_INSERT_RECNO(cbt->ins)` to __col_insert_search() doesn't it have any problem? I find a code chunk as below that works based on the passed `recno`. Sometimes it may return another `ret_ins` that have a different 'recno'.

		ins_recno = WT_INSERT_RECNO(ret_ins);
		cmp = (recno == ins_recno) ? 0 : (recno < ins_recno) ? -1 : 1;

keithbostic:
@kommiharibabu, 
> During my analysis as per the above suggestion, passing always the first `WT_INSERT_RECNO(cbt->ins)` to `__col_insert_search()` doesn't it have any problem?

I don't think so. We already have a record number (our current location), and it isn't going away or change.

I believe this change would resolve Coverity's warning without any cost.

For clarity, here's the change I propose:
```
diff --git a/src/btree/bt_curprev.c b/src/btree/bt_curprev.c
index 5ac53cb..e0deb72 100644
--- a/src/btree/bt_curprev.c
+++ b/src/btree/bt_curprev.c
@@ -38,6 +38,7 @@ __cursor_skip_prev(WT_CURSOR_BTREE *cbt)
        WT_INSERT *current, *ins;
        WT_ITEM key;
        WT_SESSION_IMPL *session;
+       uint64_t recno;
        int i;
 
@@ -47,6 +48,7 @@ restart:
         * If the search stack does not point at the current item, fill it in
         * with a search.
         */
+       recno = WT_INSERT_RECNO(cbt->ins);
        while ((current = cbt->ins) != PREV_INS(cbt, 0)) {
                if (cbt->btree->type == BTREE_ROW) {
                        key.data = WT_INSERT_KEY(current);
@@ -55,8 +57,7 @@ restart:
                            session, cbt, cbt->ins_head, &key));
                } else
                        cbt->ins = __col_insert_search(cbt->ins_head,
-                           cbt->ins_stack, cbt->next_stack,
-                           WT_INSERT_RECNO(current));
+                           cbt->ins_stack, cbt->next_stack, recno);
        }
 
        /*
```
cc: @michaelcahill

src/btree/bt_curprev.c
```diff
@@ -47,16 +48,17 @@ __cursor_skip_prev(WT_CURSOR_BTREE *cbt)
 	 * If the search stack does not point at the current item, fill it in
 	 * with a search.
 	 */
-	while ((current = cbt->ins) != PREV_INS(cbt, 0)) {
+	recno = WT_INSERT_RECNO(cbt->ins);
+	while (((current = cbt->ins) != PREV_INS(cbt, 0)) &&
+	    (current != NULL)) {
```
keithbostic: My hope is that adding the `recno` assignment outside the loop will clear up the Coverity warning because we'll no longer be indirecting through `current` after its assignment. So, the test for `current != NULL` should no longer be necessary.

kommiharibabu: Yes, the NULL check is not required, I missed to remove during the modifications. 
Fixed in d1d819e.

keithbostic:
@kommiharibabu, I've merged WT-4791 and closed the JIRA ticket.

I've also marked the Coverity report "False Positive", and "Fix Submitted", hopefully Coverity will be happier now.

PR #4739, opened 2019-07-03
WT-4805 Modify test verification of error messages

kommiharibabu:
Modified the test to ensure that it verifies the returned
error messages for the operations

### DIFF
--- a/test/suite/test_prepare04.py
+++ b/test/suite/test_prepare04.py
@@ -96,8 +96,8 @@ class test_prepare04(wttest.WiredTigerTestCase, suite_subprocess):
         c.set_value(2)
         c.update()
         self.session.prepare_transaction('prepare_timestamp=' + self.prepare_ts)
-        conflictmsg = '/conflict between concurrent operations/'
-        preparemsg = '/conflict with a prepared update/'
+        conflictmsg = "WT_ROLLBACK: conflict between concurrent operations"
+        preparemsg = "WT_PREPARE_CONFLICT: conflict with a prepared update"
 
         #'''
         # Verify data visibility from a different session/transaction.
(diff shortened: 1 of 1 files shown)


test/suite/test_prepare04.py
```diff
@@ -96,8 +96,8 @@ def test_prepare_conflict(self):
         c.set_value(2)
         c.update()
         self.session.prepare_transaction('prepare_timestamp=' + self.prepare_ts)
-        conflictmsg = '/conflict between concurrent operations/'
-        preparemsg = '/conflict with a prepared update/'
+        conflictmsg = "WT_ROLLBACK: conflict between concurrent operations"
+        preparemsg = "WT_PREPARE_CONFLICT: conflict with a prepared update"
```
sueloverso: I think you don't need the change above. The form with `/string/` will match on that as a substring of the message so I don't think you need to make the message that much more detailed. It should be fine as-is.

test/suite/test_prepare04.py
```diff
@@ -106,14 +106,17 @@ def test_prepare_conflict(self):
         s_other.begin_transaction(self.txn_config + self.ignore_config)
         c_other.set_key(1)
         if self.ignore == False and self.after_ts == True:
-            self.assertRaises(wiredtiger.WiredTigerError, lambda:c_other.search())
+            # WT_PREPARE_CONFLICT error returned when a read encounters a prepared transaction update
```
sueloverso: I would make this comment more general like:
`# Make sure we get the expected prepare conflict message.`

test/suite/test_prepare04.py
```diff
@@ -106,14 +106,17 @@ def test_prepare_conflict(self):
...
+            self.assertRaisesException(wiredtiger.WiredTigerError, lambda:c_other.search(), preparemsg)
         else:
             c_other.search()
             self.assertTrue(c_other.get_value() == 1)
+
         c_other.set_value(3)
-        self.assertRaises(wiredtiger.WiredTigerError, lambda:c_other.update())
+
+        # WT_ROLLBACK error returned due to conflict with concurrent operations
+        self.assertRaisesException(wiredtiger.WiredTigerError, lambda:c_other.update(), conflictmsg)
```
sueloverso: Similar comment here.
`# Make sure we detect the conflict between operations.`

kommiharibabu: updated according to your comment.

test/suite/test_prepare04.py
kommiharibabu: updated according to your comment.

test/suite/test_prepare04.py
kommiharibabu: @sueloverso Thanks, changed the error strings as original substrings

PR #4740, opened 2019-07-03
WT-4858 Clang 8 and GCC 9 compatibility

tetsuo-cpp:
This PR is to get WiredTiger to compile with the newest Clang and GCC releases.

I had a look at the new flags introduced in both releases but all of the interesting ones are either on by default or are part of `-Wextra`.

The main change here is to move away from the legacy `__sync` compiler intrinsics and towards the `__atomic` ones. These new functions allow us to specify memory order so we could potentially use one of the weaker orderings for performance reasons. I think for our purposes, we want the strongest ordering (sequentially consistent ordering which is what you get by default with a C++ `std::atomic`).

### DIFF
--- a/src/include/gcc.h
+++ b/src/include/gcc.h
@@ -135,11 +135,6 @@ __wt_atomic_fetch_add##name(vp_arg, v_arg)				\
 {									\
 	return (__atomic_fetch_add(vp, v, __ATOMIC_SEQ_CST));		\
 }									\
-static inline void							\
-__wt_atomic_store##name(vp_arg, v_arg)					\
-{									\
-	__atomic_store_n(vp, v, __ATOMIC_SEQ_CST);			\
-}									\
 static inline ret							\
 __wt_atomic_sub##name(vp_arg, v_arg)					\
 {									\
--- a/src/include/gcc.h
+++ b/src/include/gcc.h
@@ -90,11 +90,11 @@
  */
 
 #define	WT_ATOMIC_CAS(ptr, oldp, new)					\
-	__atomic_compare_exchange_n(						\
+	__atomic_compare_exchange_n(					\
 	    ptr, oldp, new, false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)
-#define	WT_ATOMIC_CAS_FUNC(name, vp_arg, old_arg, new_arg) \
+#define	WT_ATOMIC_CAS_FUNC(name, vp_arg, old_arg, new_arg)		\
 static inline bool							\
-__wt_atomic_cas##name(vp_arg, old_arg, new_arg)			\
+__wt_atomic_cas##name(vp_arg, old_arg, new_arg)				\
 {									\
 	return (WT_ATOMIC_CAS(vp, &old, new));				\
 }
--- a/src/include/gcc.h
+++ b/src/include/gcc.h
@@ -89,6 +89,17 @@
  * swap) operations.
  */
 
+/*
+ * We've hit optimization bugs with Clang 3.5 in the past when using the atomic
+ * builtins. See http://llvm.org/bugs/show_bug.cgi?id=21499 for details.
+ */
+#if defined(__clang__) && \
+	defined(__clang_major__) && defined(__clang_minor__) && \
+	(((__clang_major__ == 3) && (__clang_minor__ <= 5)) || \
+	 (__clang_major__ < 3))
+#error "Clang versions 3.5 and earlier are unsupported by WiredTiger"
+#endif
+
 #define	WT_ATOMIC_CAS(ptr, oldp, new)					\
 	__atomic_compare_exchange_n(					\
 	    ptr, oldp, new, false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)
--- a/src/include/gcc.h
+++ b/src/include/gcc.h
@@ -89,38 +89,29 @@
  * swap) operations.
  */
 
-#ifdef __clang__
-/*
- * We avoid __sync_bool_compare_and_swap with due to problems with optimization
- * with some versions of clang. See http://llvm.org/bugs/show_bug.cgi?id=21499
- * for details.
- */
-#define	WT_ATOMIC_CAS(ptr, old, new)					\
-	(__sync_val_compare_and_swap(ptr, old, new) == (old))
-#else
-#define	WT_ATOMIC_CAS(ptr, old, new)					\
-	__sync_bool_compare_and_swap(ptr, old, new)
-#endif
-#define	WT_ATOMIC_CAS_FUNC(name, vp_arg, old_arg, new_arg)		\
+#define	WT_ATOMIC_CAS(ptr, oldp, new)					\
+	__atomic_compare_exchange_n(						\
+	    ptr, oldp, new, false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)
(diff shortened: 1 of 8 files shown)


src/include/gcc.h
```diff
@@ -89,38 +89,29 @@
...
 {									\
-	return (WT_ATOMIC_CAS(vp, old, new));				\
+	return (WT_ATOMIC_CAS(vp, &old, new));				\
 }
 WT_ATOMIC_CAS_FUNC(8, uint8_t *vp, uint8_t old, uint8_t new)
 WT_ATOMIC_CAS_FUNC(16, uint16_t *vp, uint16_t old, uint16_t new)
 WT_ATOMIC_CAS_FUNC(32, uint32_t *vp, uint32_t old, uint32_t new)
 WT_ATOMIC_CAS_FUNC(v32,							\
-    volatile uint32_t *vp, volatile uint32_t old, volatile uint32_t new)
+    volatile uint32_t *vp, uint32_t old, volatile uint32_t new)
```
tetsuo-cpp: The "overloads" for these `__atomic` functions don't support a pointer to a volatile type for the expected argument. I understand why you'd need to preserve the volatile specifier for the value that is getting written to but I think it's fine to remove it for the expected value.

src/include/gcc.h
```diff
@@ -130,29 +121,29 @@ WT_ATOMIC_CAS_FUNC(size, size_t *vp, size_t old, size_t new)
...
 {									\
-	return (__sync_fetch_and_add(vp, v));				\
+	return (__atomic_fetch_add(vp, v, __ATOMIC_SEQ_CST));		\
 }									\
-static inline ret							\
+static inline void							\
 __wt_atomic_store##name(vp_arg, v_arg)					\
 {									\
-	return (__sync_lock_test_and_set(vp, v));			\
+	__atomic_store_n(vp, v, __ATOMIC_SEQ_CST);			\
```
tetsuo-cpp: This is the only change where the actual behaviour may be different.

`__sync_lock_test_and_set` doesn't always use the value passed in. On some architectures, the value just gets set to some implementation defined "on" value. It also uses a weaker "acquire" memory order rather than the sequentially consistent one I'm using here. The analogous function in the `__atomic` series of intrinsics is `__atomic_test_and_set`. However, this doesn't allow us to specify a value that "might" be used like the `__sync` version.

I assumed we probably didn't intend for that subtlety so I've just gone for `__atomic_store_n` with a stronger ordering.

src/include/gcc.h
```diff
@@ -89,38 +89,29 @@
  * swap) operations.
  */
 
-#ifdef __clang__
-/*
- * We avoid __sync_bool_compare_and_swap with due to problems with optimization
```
tetsuo-cpp: @keithbostic 
Spooky stuff right here...

Ideally, I'd like to remove this special case since we're not using that set of intrinsics anymore. But to my understanding, the `__sync` series of builtins is implemented in terms of the more flexible `__atomic` ones. So it may suffer from the same problem.

If you can remember the story behind this, could you please fill me in on:
- Whether this was "obvious" and caused build failures often.
- What part of WiredTiger exposed this bug. I see that in the issue tracker, the LLVM dev who responded couldn't reproduce so I'm assuming this "optimisation" doesn't always trigger. If I'm going to look at what's being emitted by Clang then I'll need to narrow down to a usage that is known to cause this.

If the bug was obvious in testing, I'd be tempted to just merge this and see if it breaks anything in Jenkins.

src/include/gcc.h
keithbostic: I agree, I think this change is correct.

src/include/gcc.h
keithbostic: I'm kind of confused why we're using `__wt_atomic_store` at all. In the only two places we're using it, we're discarding any return value, which means all it's doing is a store followed by a flush? Do you understand what that's buying us?

src/include/gcc.h
keithbostic: While it's been 4.5 years and 5 major clang releases since we saw the problem, I'm hesitant to remove the workaround. Further, we don't know when (well, honestly, "if"), the bug has been fixed, so we can't even detect versioning problems at compile time. (As an example, MongoDB/Evergreen allows you to spawn Ubuntu 1404, which has clang 3.4 installed.) 

I don't know how difficult it was to find or reproduce, @michaelcahill did that work.

cc: @agorrod

tetsuo-cpp: The workaround is a blocker since it is Clang who is noisily complaining about our use of `__sync` functions. So I can't have:
```
#ifdef __clang__
/* __sync versions of the functions */
#else
/* __atomic versions of the functions */
#endif
```
I can `#ifdef` based on the Clang version but it doesn't help if we're afraid that this bug could exist even in Clang 8.

keithbostic: This is an old bug, I'm much less concerned that it's fixed by now than that someone will use an older compiler and we don't know where in the clang release cycle the bug was fixed. On the other hand, we can only push this into the future for so long, at some point, we're going to have to move forward.

@michaelcahill, is there a reasonably easy way we could check the assembly and figure out under which compiler versions this fails?

michaelcahill: It's going to be hard to check the assembly: it's all moved around so much since the original bug fired.  That said, the symptom was dramatic: we were evicting pages even though the atomic change of `ref->state` to `WT_REF_LOCKED` had failed.

I'm inclined to make the change we want and run stress testing against a `clang` build for a day: given how much our test coverage has improved, if that doesn't show failure I think we're safe.

src/include/gcc.h
michaelcahill: Likewise, this doesn't make a lot of sense to me.  We assume all over the place that we can do ordinary assignments to variables that are being read and atomically incremented concurrently: I don't see what `__wt_atomic_store` is buying us.

tetsuo-cpp: I can remove it if you'd like but I don't get why we don't need it.
Don't you need to use an atomic store to avoid a reader thread observing a partially written value? In one of the cases it's a store on a 64 bit value so, to my understanding, on some architectures it might get compiled down to multiple separate stores.

keithbostic: WiredTiger requires the architecture not use multiple separate stores for integral or pointer values, there's some discussion here: https://github.com/wiredtiger/wiredtiger/blob/master/src/include/gcc.h#L27-L90

tetsuo-cpp: fefe1b6

Thanks, that was an interesting read.
Removed them.

src/include/gcc.h
tetsuo-cpp: @keithbostic @michaelcahill 
Ok, I plan to compile this change with the offending `clang` version and if it passes testing then we can probably conclude that the `__atomic` intrinsics don't suffer from the same issue.

It's difficult if it fails but we can cross that bridge if/when we come to it.

lukech:
retest this please

lukech:
retest this please

src/include/gcc.h
tetsuo-cpp: 9f3ca3b

I've added a check to stop us from compiling with `clang-3.5` or earlier.

I've run some stress testing with `clang-3.9` (that's the oldest one on our Jenkins). I've tried twice and it runs fine for ~8 hours without issues but eventually terminates with a stuck cache (this has been an issue even in `develop` for the past few days). It's a bit difficult to judge since `develop` is not passing test format stress testing right now but I'd expect that this compiler bug would cause us to fail faster than that.

@michaelcahill @keithbostic 
If you're satisfied, I think this is ready to go.

PR #4743, opened 2019-07-04
WT-4910 Python 3 Windows support

markbenvenuto:
Two sets of trivial changes. 
1. print statement to print function
2. Fix a bad indent

Python 2 will be still supported since this such a simple file.


lukech:
evergreen retry

PR #4748, opened 2019-07-04
WT-4792 Add stat to track pages queued for eviction after LRU sorting

tetsuo-cpp:
At the moment, we have a stat called `cache_eviction_pages_queued` which tracks the amount of pages initially queued. Afterwards, we do an LRU sort of the pages and trim some entries off the end of the queue. This PR is adding a statistic to track the size of the queue after this processing.

### DIFF
--- a/dist/stat_data.py
+++ b/dist/stat_data.py
@@ -235,6 +235,7 @@ connection_stats = [
     CacheStat('cache_eviction_maximum_page_size', 'maximum page size at eviction', 'no_clear,no_scale,size'),
     CacheStat('cache_eviction_pages_queued', 'pages queued for eviction'),
     CacheStat('cache_eviction_pages_queued_oldest', 'pages queued for urgent eviction during walk'),
+    CacheStat('cache_eviction_pages_queued_post_lru', 'pages queued for eviction post lru sorting'),
     CacheStat('cache_eviction_pages_queued_urgent', 'pages queued for urgent eviction'),
     CacheStat('cache_eviction_pages_seen', 'pages seen by eviction walk'),
     CacheStat('cache_eviction_queue_empty', 'eviction server candidate queue empty when topping up'),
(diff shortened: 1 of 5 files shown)


src/evict/evict_lru.c
```diff
@@ -1313,6 +1313,8 @@ __evict_lru_walk(WT_SESSION_IMPL *session)
 		__evict_list_clear(session, &queue->evict_queue[--entries]);
 
 	queue->evict_entries = entries;
+	WT_STAT_CONN_INCRV(
+	    session, cache_eviction_pages_queued_post_lru, entries);
```
tetsuo-cpp: @sulabhM 
I see that we have `queue->evict_candidates` to track the number of things to be evicted whereas I'm tracking `queue->evict_entries` which is the overall size of the queue. Do you want me to also add a candidates statistic? Or is it not that interesting.

sulabhM: I think `entries` here is not interesting. I meant to track `queue->evict_candidates` after we make adjustments with the if/else block that starts at 1332. Those adjustments are that bite us at times with FTDC when we do not know how many pages we actually decided to evict.

tetsuo-cpp: 2b0df56

tetsuo-cpp:
retest this please

dist/stat_data.py
```diff
@@ -235,6 +235,7 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_eviction_maximum_page_size', 'maximum page size at eviction', 'no_clear,no_scale,size'),
     CacheStat('cache_eviction_pages_queued', 'pages queued for eviction'),
```
sulabhM: Ideally, we will call this statistic as "pages considered for LRU sorting" or something similar and the new statistic as "pages queued for eviction". But the changes to this statistic might add to the confusion with FTDC within releases and we will have to educate a lot of people around it. @agorrod , what do you think? Any suggestions for naming the new statistic that is really "pages queued".

tetsuo-cpp:
evergreen retry

dist/stat_data.py
agorrod: I agree that ideally we'd change the name of this statistic, but also that the fallout is non-trivial. I'm inclined to leave as-is, I think the new stat will sort in a way that makes it clear enough (at least for experts).

PR #4749, opened 2019-07-05
WT-4893 internal page child-page eviction checks can race with cursors in the tree

keithbostic:


src/include/btmem.h
```diff
@@ -897,37 +905,44 @@ struct __wt_ref {
...
 	(ref)->hist[(ref)->histoff].name = session->name;		\
-	(ref)->hist[(ref)->histoff].file = (f);				\
+	(ref)->hist[(ref)->histoff].func = (f);				\
 	(ref)->hist[(ref)->histoff].line = (uint16_t)(l);		\
 	(ref)->hist[(ref)->histoff].state = (uint16_t)(s);		\
 	(ref)->histoff =						\
 	    ((ref)->histoff + 1) % WT_ELEMENTS((ref)->hist);		\
 } while (0)
 #define	WT_REF_SET_STATE(ref, s) do {					\
-	WT_REF_SAVE_STATE(ref, s, __FILE__, __LINE__);			\
```
agorrod: I *think* we've been using `__FILE__` instead of `__func__ `for compatability reasons - Windows MSVC doesn't support `__func__`:
https://stackoverflow.com/questions/7008485/func-or-function-or-manual-const-char-id

src/support/hazard.c
```diff
@@ -328,6 +328,10 @@ __wt_hazard_check(WT_SESSION_IMPL *session,
 	WT_SESSION_IMPL *s;
 	uint32_t i, j, hazard_inuse, max, session_cnt, walk_cnt;
 
+	/* If a file can never be evicted, hazard pointers aren't required. */
```
agorrod: Is this safe? Do we need hazard pointers to manage when it's OK to do in-memory splits - which are part of in memory.

agorrod:
I've switch http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-zseries over to run on this branch

src/include/btmem.h
keithbostic: MSVC supports `__FUNCTION__`, and has since around 2003. We have:
```
/* MSVC Doesn't provide __func__, it has __FUNCTION__ */
#ifdef _MSC_VER
#define __func__        __FUNCTION__
#endif
```
in `include/msvc.h`, and we've been using it in the standard error handling function calls for awhile now.

src/support/hazard.c
keithbostic: That same check is already in `__wt_hazard_set()` and `__wt_hazard_clear()`, the hazard pointer set/clear functions, this just adds the check to the eviction path.

agorrod: I thought there was a check somewhere, but couldn't spot it in 2 minutes.

src/include/btmem.h
agorrod: Ahh - I was surprised it didn't fail compilation.

agorrod:
@keithbostic I'm good to go with this change now - should I go ahead and merge?

keithbostic:
@agorrod, yes, I think this is ready to merge.

agorrod:
Will do. @michaelcahill I'd appreciate if you can take a look at this one as well (even in retrospect).

keithbostic:
@michaelcahill, one question, I convinced myself no barrier was necessary to ensure he loops happen exactly as intended, it would be good to have someone else agree. 

Cc: @agorrod

michaelcahill:
I think I understand the problem this change is solving.  That said, it now walks each internal page 3 times: forwards, then backwards, then forwards again (the main loop).  Would the backwards pass then the main loop be sufficient?

While we're in the area, can I check my understanding of what stops a thread hazard coupling from an adjacent internal page to the first / last child of this internal page after we've check it?  I think the `__ref_ascend` / `descend` code in `bt_walk.c` means we will couple from the last child of internal page 1 to internal page 2 before coupling from internal page 2 to its first child.  In other words, two calls to `__ref_ascend` then `__wt_page_swap` from a leaf to an internal page (which would fail here because the internal page is locked).  lmk if this is wrong or doesn't make sense.

keithbostic:
> it now walks each internal page 3 times: forwards, then backwards, then forwards again (the main loop).  Would the backwards pass then the main loop be sufficient?

Yes, I think that would be sufficient. The reason I did 3 passes was because I saw lots of failures from this function in immediate-eviction workloads (we're constantly trying to evict internal pages, and in the case of cursors in the page, it's going to fail), and I thought the first two checks would be cheap relative to the 3rd pass (which includes visibility checks). That calculation assumes deleted or lookaside children, of course. All that said, I would have no objection to switching to 2 walks.

> While we're in the area, can I check my understanding of what stops a thread hazard coupling from an adjacent internal page to the first / last child of this internal page after we've check it?

I believe that's exactly correct.

PR #4751, opened 2019-07-05
WT-4836 Lower scheduling frequency for endianness compatibility tests

lukech:
Lower the scheduling frequency for endianness compatibility tests from **once per day** to **once per week**, to avoid overwhelming little-endian (zseries/s390x) Evergreen distro queue.


lukech:
@agorrod Can you help to take a look? Cheers.

PR #4752, opened 2019-07-05
WT-4918 LSM allocated mutexes leaked in readonly tests

keithbostic:


src/lsm/lsm_manager.c
```diff
@@ -220,6 +219,16 @@ __wt_lsm_manager_start(WT_SESSION_IMPL *session)
 	/* We need at least a manager, a switch thread and a generic worker. */
 	WT_ASSERT(session, manager->lsm_workers_max > 2);
 
+	/* Allocate serialization resources. */
```
agorrod: I don't think this is safe. There could be multiple handles being opened simultaneously - each calls `__wt_lsm_manager_start`. If that happens all calls to `__wt_lsm_manager_start` will return early except for the winner of the above atomic cas. Those other opens can get to the point where a work unit is pushed onto the queue before the below locks are initialized.
It's an unlikely path, but also a failure that would be a pain to diagnose.

PR #4753, opened 2019-07-08
WT-4612 Improve test coverage for prepare updates older than the oldest.

bvpvamsikrishna:
Added a scenario to improve coverage when the oldest timestamp advances past a prepared update.

### DIFF
--- a/test/suite/test_prepare07.py
+++ b/test/suite/test_prepare07.py
@@ -49,7 +49,7 @@ class test_prepare07(wttest.WiredTigerTestCase):
         self.conn.set_timestamp('stable_timestamp=' + timestamp_str(100))
 
         # Commit some updates.
-        value_b = "bbbbb" * 100
+        value_b = b"bbbbb" * 100
         cursor = self.session.open_cursor(uri)
         self.session.begin_transaction('isolation=snapshot')
         cursor.set_key(ds.key(nrows + 1))
@@ -145,7 +145,7 @@ class test_prepare07(wttest.WiredTigerTestCase):
         nrows = 100
         ds = SimpleDataSet(self, uri, nrows, key_format="S", value_format='u')
         ds.populate()
-        value_a = "aaaaa" * 100
+        value_a = b"aaaaa" * 100
 
         # Initially load huge data
         cursor = self.session.open_cursor(uri)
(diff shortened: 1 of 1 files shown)


test/suite/test_prepare07.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertEquals(cursor_b.get_value(),value_b)
+        cursor_b.set_key(ds.key(nrows + 5))
+        self.assertEquals(cursor_b.search(), 0)
+        self.assertEquals(cursor_b.get_value(),value_b)
+
+        # Committed transactions newer to the stable timestamp should not be seen.
+        cursor_b.set_key(ds.key(nrows + 6))
+        self.assertEquals(cursor_b.search(), 0)
+        self.assertEquals(cursor_b.get_value(),value_a)
+
```
kommiharibabu: Adding an additional test scenario of validating the prepared updates after moving the stable timestamp? Is it worth?

bvpvamsikrishna: I think yes, the reason being ```__wt_txn_visible_all``` uses the oldest txn id, and as part of prepare_transaction we release transaction id. The visibility of a prepared transaction is controlled by the timestamp and state rather than txn id, till its resolution. So, when the prepared transaction is the oldest transaction, eviction/checkpoint should not make prepared updates durable based on the txn id, but should consider the timestamp and state.

test/suite/test_prepare07.py
```diff
@@ -0,0 +1,165 @@
...
+import fnmatch, os, shutil, time
+from helper import copy_wiredtiger_home
+import wiredtiger, wttest
+from wtdataset import SimpleDataSet
+
+def timestamp_str(t):
+    return '%x' % t
+
+# test_prepare07.py
+# test to ensure prepared updates older than oldest timestamp are not visible.
```
luke-pearson: Should use capitals at the start of your sentences.

test/suite/test_prepare07.py
```diff
@@ -0,0 +1,165 @@
...
+        cursor.set_key(ds.key(nrows + 5))
+        cursor.set_value(value_b)
+        self.assertEquals(cursor.update(), 0)
+        self.session.commit_transaction('commit_timestamp=' + timestamp_str(150))
+
+        # Move the oldest and the stable timestamp to the latest.
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(155))
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(155))
+
+        # Commit an update newer to the stable timestamp.
```
luke-pearson: newer than the

PR #4756, opened 2019-07-08
WT-4916 Pre-formatting fixes

tetsuo-cpp:
This PR is to address some formatting issues identified in the past two Clang Format review PRs. I'd like to merge this before applying Clang Format to make the output as nice as possible.
The changes are:
- Remove `WT_ILLEGAL_VALUE` macros. We expect these to be indented at the same level as `case` label however, the formatter has no way of knowing it's not just a regular function call.
- Don't allow comments to trail after a `goto` label. It means we can use `s_goto` in its current state and also it is more consistent since Clang Format will not let code trail after a `goto` label like we do now.
- Wrap a few things with braces that don't get formatted well.

### DIFF
--- a/src/reconcile/rec_col.c
+++ b/src/reconcile/rec_col.c
@@ -273,8 +273,8 @@ __wt_rec_col_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_REF *pageref)
 				addr = &child->modify->mod_replace;
 				break;
 			default:
-			ret = __wt_illegal_value(
-			    session, child->modify->rec_result);
+				ret = __wt_illegal_value(
+				    session, child->modify->rec_result);
 			goto err;
 			}
 			break;
--- a/src/reconcile/rec_col.c
+++ b/src/reconcile/rec_col.c
@@ -804,7 +804,8 @@ __wt_rec_col_var(WT_SESSION_IMPL *session,
 		WT_ERR(__wt_dsk_cell_data_ref(
 		    session, WT_PAGE_COL_VAR, vpack, orig));
 
-record_loop:	/*
+record_loop:
+		/*
 		 * Generate on-page entries: loop repeat records, looking for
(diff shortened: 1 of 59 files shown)


src/reconcile/rec_col.c
```diff
@@ -272,8 +272,10 @@ __wt_rec_col_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_REF *pageref)
 			case WT_PM_REC_REPLACE:
 				addr = &child->modify->mod_replace;
 				break;
-			WT_ILLEGAL_VALUE_ERR(
+			default:
+			ret = __wt_illegal_value(
```
kommiharibabu: Indentation?

src/reconcile/rec_row.c
```diff
@@ -456,8 +456,10 @@ __wt_rec_row_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
 				 */
 				addr = &child->modify->mod_replace;
 				break;
-			WT_ILLEGAL_VALUE_ERR(
+			default:
+			ret = __wt_illegal_value(
```
kommiharibabu: indentation?

src/reconcile/rec_col.c
tetsuo-cpp: 4faf83c

src/reconcile/rec_row.c
tetsuo-cpp: 4faf83c

keithbostic:
@tetsuo-cpp, I've pushed a couple of cleanups, and I'm ready to merge.

Please review 01bc3a6 and eaebd1f, and if you approve of them, let's merge and close.

tetsuo-cpp:
@keithbostic 
Thanks for the commits. As discussed, that static analysis failure is unrelated to this PR so I'm going to go ahead and merge.

PR #4757, opened 2019-07-08
WT-4857 Use a local variable to access the insert structure.

sueloverso:
This is the simple change that has (so far) fixed the column store mismatch issue. Please review.

### DIFF
--- a/src/include/column.i
+++ b/src/include/column.i
@@ -29,7 +29,7 @@ __col_insert_search_gt(WT_INSERT_HEAD *ins_head, uint64_t recno)
 	 * go as far as possible at each level before stepping down to the next.
 	 */
 	ins = NULL;
-	for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;)
+	for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;) {
 		/*
 		 * Use a local variable to access the insert because the skip
 		 * list can change across references.
@@ -42,6 +42,7 @@ __col_insert_search_gt(WT_INSERT_HEAD *ins_head, uint64_t recno)
 			--i;		/* LT: drop down a level */
 			--insp;
 		}
+        }
 
 	/*
 	 * If we didn't find any records smaller than the target, we never set
@@ -83,7 +84,7 @@ __col_insert_search_lt(WT_INSERT_HEAD *ins_head, uint64_t recno)
 	 * The insert list is a skip list: start at the highest skip level, then
 	 * go as far as possible at each level before stepping down to the next.
 	 */
-	for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;)
+	for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;) {
 		/*
 		 * Use a local variable to access the insert because the skip
 		 * list can change across references.
@@ -96,6 +97,7 @@ __col_insert_search_lt(WT_INSERT_HEAD *ins_head, uint64_t recno)
 			--i;		/* LTE: drop down a level */
 			--insp;
 		}
+        }
 
 	return (ins);
 }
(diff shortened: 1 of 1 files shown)


src/include/column.i
```diff
@@ -122,13 +122,18 @@ __col_insert_search_match(WT_INSERT_HEAD *ins_head, uint64_t recno)
...
+		ret_ins = *insp;
+		ins_recno = WT_INSERT_RECNO(ret_ins);
 		cmp = (recno == ins_recno) ? 0 : (recno < ins_recno) ? -1 : 1;
 
 		if (cmp == 0)			/* Exact match: return */
-			return (*insp);
+			return (ret_ins);
 		if (cmp > 0)			/* Keep going at this level */
-			insp = &(*insp)->next[i];
+			insp = &ret_ins->next[i];
```
keithbostic: On the surface this looks correct to me. Two comments:

1. We're still reading `*insp` twice and presumably it can change between the two reads, so I'd suggest:
```
/*
 * Use a local variable to access the insert because the skip
 * list can change across references.
 */
if ((ret_ins = *insp) == NULL) {
        --i;
        --insp;
        continue;
}

ins_recno = WT_INSERT_RECNO(ret_ins);
cmp = (recno == ins_recno) ? 0 : (recno < ins_recno) ? -1 : 1;

if (cmp == 0)                   /* Exact match: return */
```

2. The two functions in the same file above this one, `__col_insert_search_lt()` and `__col_insert_search_gt()` were probably copied from `__col_insert_search_match()`, and appear to need the same fix? The function `__col_insert_search()` seems fine, but the revision history doesn't indicate that it was modified to fix a race.

sueloverso: Fixed. If only the other 3 functions had the same fix applied 7 years ago!

src/include/column.i
```diff
@@ -30,9 +30,11 @@ __col_insert_search_gt(WT_INSERT_HEAD *ins_head, uint64_t recno)
...
 	ins = NULL;
 	for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;)
-		if (*insp != NULL && recno >= WT_INSERT_RECNO(*insp)) {
-			ins = *insp;	/* GTE: keep going at this level */
-			insp = &(*insp)->next[i];
+		if ((local_ins = *insp) != NULL &&
+		    recno >= WT_INSERT_RECNO(*insp)) {
+			/* GTE: keep going at this level */
+			ins = local_ins;
+			insp = &local_ins->next[i];
```
keithbostic: This doesn't look right to me, we're still reading `*insp` twice. I'd suggest getting rid of `local_ins`, I think `ins` is all you need (but it wouldn't hurt my feelings to have all functions use the same naming pattern).

src/include/column.i
```diff
@@ -79,9 +81,11 @@ __col_insert_search_lt(WT_INSERT_HEAD *ins_head, uint64_t recno)
...
 	 */
 	for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;)
-		if (*insp != NULL && recno > WT_INSERT_RECNO(*insp)) {
-			ins = *insp;	/* GT: keep going at this level */
-			insp = &(*insp)->next[i];
+		if ((local_ins = *insp) != NULL &&
+		    recno > WT_INSERT_RECNO(*insp)) {
+			/* GT: keep going at this level */
+			ins = local_ins;
+			insp = &local_ins->next[i];
```
keithbostic: Ditto my previous comment.

keithbostic: I'm also a little concerned that in all 3 cases there's nothing to prevent the compiler from reading `*insp` twice on its own. Maybe:
```
ins = NULL;
for (i = WT_SKIP_MAXDEPTH - 1, insp = &ins_head->head[i]; i >= 0;)
        WT_ORDERED_READ(ins, *insp);
        if (ins != NULL && recno >= WT_INSERT_RECNO(ins)) {
                /* GTE: keep going at this level */
                insp = &local_ins->next[i];
        } else {
                --i;            /* LT: drop down a level */
                --insp;
        }
```

keithbostic:
@michaelcahill, I think this is ready to merge. Can you please review and merge it if you agree?

Thanks.

agorrod:
Thanks @sueloverso, @keithbostic  and @michaelcahill nice team work. This lgtm, I'm going to merge.

PR #4758, opened 2019-07-10
WT-4913 Fix the Windows CRC32 on blocks that aren't 8B aligned and/or multiples of 8B

keithbostic:
@agorrod, this is only lightly tested, but seems to work.

I can't think of any way to test it (other than hacking the code to exercise it, which seems like cheating). Let me know if either of you have a better idea for testing.

### DIFF
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -45,26 +45,18 @@ static bool
 __log_checksum_match(WT_ITEM *buf, uint32_t reclen)
 {
 	WT_LOG_RECORD *logrec;
-	uint32_t checksum_calculate, checksum_tmp;
-	bool alt_checksum;
+	uint32_t checksum_saved, checksum_tmp;
+	bool checksum_matched;
 
-	alt_checksum = false;
-
-	logrec = (WT_LOG_RECORD *)buf->mem;
-	checksum_tmp = logrec->checksum;
-	logrec->checksum = 0;
-retry_checksum:
-	checksum_calculate = alt_checksum ?
-	    __wt_checksum_alt(logrec, reclen) : __wt_checksum(logrec, reclen);
+	logrec = buf->mem;
+	checksum_saved = checksum_tmp = logrec->checksum;
 #ifdef WORDS_BIGENDIAN
-	checksum_calculate = __wt_bswap32(checksum_calculate);
+	checksum_tmp = __wt_bswap32(checksum_tmp);
 #endif
-	if (logrec->checksum != checksum_calculate && !alt_checksum) {
-		alt_checksum = true;
-		goto retry_checksum;
-	}
-	logrec->checksum = checksum_tmp;
-	return (logrec->checksum == checksum_calculate);
+	logrec->checksum = 0;
+	checksum_matched = __wt_checksum_match(logrec, reclen, checksum_tmp);
+	logrec->checksum = checksum_saved;
+	return (checksum_matched);
 }
 
 /*
--- a/src/log/log.c
+++ b/src/log/log.c
@@ -42,19 +42,27 @@ __wt_log_printf(WT_SESSION_IMPL *session, const char *format, ...)
  *	Given a log record, return whether the checksum matches.
  */
 static bool
-__log_checksum_match(WT_SESSION_IMPL *session, WT_ITEM *buf, uint32_t reclen)
+__log_checksum_match(WT_ITEM *buf, uint32_t reclen)
 {
 	WT_LOG_RECORD *logrec;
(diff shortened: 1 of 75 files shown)


agorrod:
@keithbostic thanks for putting this together. It's not quite what I expected, but I can think of reasons why the layering was complex. Could we talk this through?

agorrod:
I chatted to @keithbostic and now understand better. The reason we can't hide this inside `wiredtiger_crc32c_func` is that we don't know there if the checksum is going to succeed because we don't have the expected value available.
We talked about possibly adding a new API that was `wiredtiger_crc32c_compare_func` or similar that takes the comparator as an argument, but that doesn't layer nicely with how WiredTiger uses the checksum code internally (specifically in logging and __desc_read where byteswapping happens in between calculating and checking the checksum).

We also talked a bit about whether it's possible that we would encounter this problem. We don't know of any places where WiredTiger doesn't adhere to the 8-byte alignment full 8-byte boundaries conditions that mean this can't happen, but since this code relies on block reads and memory allocations, it's possible that not all checksums are in multiples of 8-bytes. I wondered if we could "wait and see" since this would only become apparent after a user with such a checksum was upgrading an existing database. That would be reasonable if we could expect to see a failure immediately - in such a scenario. We can't - the failure could be encountered at some random time after restarting.
I had a further doubt about whether we can handle this case regardless. In the case where there are 10 bytes to be checksummed, and aligned on an 8 byte boundary. The first 8 bytes are handled by the "normal path". The code is then calculating the next part of the checksum by looking at the next 4 bytes, but in this case only 2 of those bytes have valid information in them - the other 2 are garbage (and likely to be different after a buffer is created with content that is re-read).

markbenvenuto:
@agorrod In terms of whether it checksums garbage, the code in question that makes the call to the `crc32l` instruction, always reads 1 byte at a time and then zero extends it.

```
142397fb6:      0f b6 00        movzbl  (%rax), %eat
142397fb9:      8b 0c 24        movl    (%rsp), %ecx
142397fbc:      f2 0f 38 f1 c8  crc32l  %eax, %ecx     <=== 32 bits crc
```

The `movzbl` in this case will read a byte `XX` and then zero extend it. This means that we are always checksumming `0x000000XX` which is deterministic. It was an excellent question.

agorrod:
Thanks @keithbostic this change looks good to me now. I was hoping we could get rid of the `_alt` code, but I can see why you've kept it. I don't think MongoDB needs to know about it (since they only started using WiredTiger checksums in 4.2, so there aren't any supported versions that use the old checksum), so it doesn't need to be part of the public API.

I *think* the PR test failure is due to an issue in develop, not this branch.

@markbenvenuto would you mind doing a review as well?

keithbostic:
autobuild: retest this please

keithbostic:
autobuild: retest this please

src/log/log.c
```diff
@@ -42,21 +42,21 @@ __wt_log_printf(WT_SESSION_IMPL *session, const char *format, ...)
...
-	WT_UNUSED(session);
-	logrec = (WT_LOG_RECORD *)buf->mem;
-	checksum_tmp = logrec->checksum;
-	logrec->checksum = 0;
-	checksum_calculate = __wt_checksum(logrec, reclen);
+	logrec = buf->mem;
+	checksum_saved = checksum_tmp = logrec->checksum;
 #ifdef WORDS_BIGENDIAN
-	checksum_calculate = __wt_bswap32(checksum_calculate);
+	checksum_tmp = __wt_bswap32(checksum_tmp);
```
agorrod: You've swapped the things being byte-swapped here. It used to be swapping the calculated checksum, now it's byte-swapping the saved checksum. I *think* that results in the same checking, but am pointing it out because I'm not confident enough in my endian knowledge.

keithbostic: I <i>think</i> it's the same, too...

@lukech, does our BE/LE testing include log file portability tests?

keithbostic:
I'm going to lock this PR for a few days: I want to take a run at pushing the `__wt_checksum_match()` API down into the checksum code, instead of the current change which pushes the `__wiredtiger_crc32c_alt_func()` API down.

src/log/log.c
lukech: @keithbostic Current BE/LE tests verify data files generated by below test/format command. I don't think log files are generated.
`./t -1 -h "WT_TEST.$i" -c $(pwd)/../../../test/format/CONFIG.endian`

michaelcahill: fwiw, I think this change is fine.

PR #4759, opened 2019-07-10
WT-4674 Initialize birthmarks using in-memory data instead of lookaside.

sulabhM:
When doing lookaside eviction keep birthmark data for each record in
memory. Use this data to instantiate the page back in memory. Also,
cross-reference with the lookaside to assert if there is a mismatch.


src/btree/bt_read.c
```diff
@@ -528,9 +556,12 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
 	 * Don't free WT_REF.page_las, there may be concurrent readers.
 	 */
 	if (final_state == WT_REF_MEM && ref->page_las != NULL &&
-	    (!ref->page_las->skew_newest || ref->page_las->has_prepares))
+	    (!ref->page_las->skew_newest || ref->page_las->has_prepares)) {
 		WT_ERR(__wt_las_remove_block(
 		    session, ref->page_las->las_pageid));
+		__wt_free(session, ref->page_las->birthmarks);
+		__wt_free(session, ref->page_las);
```
sulabhM: This is not being done at the moment in develop. What happens is if this page gets evicted again and we keep this lookside reference, when the page gets re-instantiated lookaside search is done to locate records with this las_pageid. No record is found because this block got removed. But searching for the record is an inefficiency we could fix.

sulabhM:
@agorrod , @tetsuo-cpp this is ready for review

src/btree/bt_read.c
sulabhM: Removing eviction_to_lookaside resolved this issue in this branch. Also, I couldn't reproduce the mentioned case in develop. So now this comment can be ignored.

src/btree/bt_read.c
```diff
@@ -530,9 +557,12 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
 	 * Don't free WT_REF.page_las, there may be concurrent readers.
 	 */
 	if (final_state == WT_REF_MEM && ref->page_las != NULL &&
-	    (!ref->page_las->skew_newest || ref->page_las->has_prepares))
+	    (!ref->page_las->skew_newest || ref->page_las->has_prepares)) {
 		WT_ERR(__wt_las_remove_block(
 		    session, ref->page_las->las_pageid));
+		__wt_free(session, ref->page_las->birthmarks);
```
tetsuo-cpp: I don't get this part.
Isn't this array of birthmarks supposed to live as long as the `WT_REF` is alive? How come we're freeing it after instantiating.

src/cache/cache_las.c
```diff
@@ -838,12 +875,23 @@ __wt_las_insert_block(WT_CURSOR *cursor,
...
 	__las_restore_isolation(session, saved_isolation);
 
+	if (ret == 0 && insert_cnt > 0)
+		ret = __wt_calloc(session, insert_cnt,
+		    sizeof(WT_BIRTHMARK_DETAILS), &multi->page_las.birthmarks);
+
 	if (ret == 0 && insert_cnt > 0) {
 		multi->page_las.las_pageid = las_pageid;
 		multi->page_las.has_prepares = prepared_insert_cnt > 0;
+		memcpy(multi->page_las.birthmarks, birthmarks->mem,
```
tetsuo-cpp: Is there any reason we can't just do:

```
multi->page_las.birthmarks = birthmarks->mem;
```

Seems weird that we allocate a new buffer, `memcpy` into it and free the old one immediately.

src/cache/cache_las.c
```diff
@@ -699,6 +702,11 @@ __wt_las_insert_block(WT_CURSOR *cursor,
 	WT_ERR(__wt_txn_begin(session, NULL));
 	local_txn = true;
 
+	/* Ensure enough room for a column-store key without checking. */
+	WT_ERR(__wt_scr_alloc(session, WT_INTPACK64_MAXSIZE, &key));
+
+	WT_ERR(__wt_scr_alloc(session, 0, &birthmarks));
```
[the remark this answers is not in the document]
sulabhM: This ensures that there is a scratch buffer allocated that is of size >= 0. There is more to scratch buffer allocation than just allocating memory of the given size. The logic here is to allocate a scratch buffer of a minimum size and then expand it as needed. Scratch buffer allocation and expansion follows a logic to minimize memory allocations.

src/cache/cache_las.c
sulabhM: That is a good question, I had tried to only allocate only one buffer before till I realized that scratch buffer is tied up with a session and here the allocated memory is tied up with the page structure that gets accessed through multiple sessions. There is certainly scope of improvement here, and we will likely work on it as a separate ticket.

src/evict/evict_page.c
```diff
@@ -334,18 +332,15 @@ __evict_page_clean_update(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
...
-		WT_REF_SET_STATE(ref, WT_REF_LOOKASIDE);
-	} else if (ref->addr == NULL) {
+	if (ref->addr == NULL) {
 		WT_WITH_PAGE_INDEX(session,
 		    ret = __evict_delete_ref(session, ref, flags));
 		WT_RET_BUSY_OK(ret);
 	} else
 		WT_REF_SET_STATE(ref, WT_REF_DISK);
 
+	WT_UNUSED(closing);
```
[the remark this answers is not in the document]
sulabhM: We did use "closing" as a part of WT_ASSERT, but since non-debug builds exclude asserts, the compiler complained about "closing" not being used.

test/suite/test_timestamp04.py
```diff
@@ -109,6 +109,10 @@ def ConnectionOpen(self, cacheSize):
         self.session = wttest.WiredTigerTestCase.setUpSessionOpen(self, self.conn)
 
     def test_rollback_to_stable(self):
+
+        # Disable this test until we deal with rollback for the relevant history project
```
sulabhM: Note: I have added a comment in WT-4680 to re-enable this test when we work on rollback to stable for this project.

sulabhM:
autobuild: retest this again

sulabhM:
autobuild: retest this please

sulabhM:
@keithbostic , could you review this one, please.

This is the initial change for the project to keep only relevant history in the cache. The aim is to be able to search through the lookaside file directly for an old record rather than having to instantiate lookaside contents as updates on the page. Since we need to read birthmarks from the lookaside to obtain the timestamps for the on-disk value, this pull request is the first step in the direction to keep birthmark information in-memory and use that to create birthmark records when the page gets instantiated with lookaside:
1. Keep birthmark information in the memory when doing lookaside eviction
2. Use the in-memory information to instantiate birthmark-updates when reading a page from the lookaside. Cross-reference the data from the lookaside to make sure that birthmark information matches.

This pull request covers the above two points. There will be followup work (I have put comments in the review to point these out) which would be to:
1. Optimize the amount of memory consumed for birthmarks by only allocating as much memory as birthmarks. In the current pull request, I store one birthmark per key even when that record doesn't need a birthmark.
2. Optimize scratch buffer memory allocation for birthmarks, to reduce the number of times we allocate memory

Let me know if we should have a call regarding this change.

src/btree/bt_read.c
```diff
@@ -189,10 +198,29 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 		WT_ERR(__wt_update_alloc(
 		    session, &las_value, &upd, &incr, upd_type));
 		total_incr += incr;
-		upd->txnid = las_txnid;
-		upd->durable_ts = durable_timestamp;
-		upd->start_ts = las_timestamp;
-		upd->prepare_state = prepare_state;
+		if (upd_type == WT_UPDATE_BIRTHMARK) {
+			birthmarkp =
+			    (WT_BIRTHMARK_DETAILS *)ref->page_las->birthmarks +
```
keithbostic: This cast isn't necessary.

src/btree/bt_split.c
```diff
@@ -1788,6 +1790,10 @@ __wt_multi_to_ref(WT_SESSION_IMPL *session,
 
 		WT_RET(__wt_calloc_one(session, &ref->page_las));
 		*ref->page_las = multi->page_las;
+		ref->page_las->birthmarks = multi->page_las.birthmarks;
+#ifdef HAVE_DIAGNOSTIC
+		ref->page_las->birthmarks_cnt = multi->page_las.birthmarks_cnt;
+#endif
```
keithbostic: @sulabhM, I don't understand this code, can you please explain why it's a diagnostic?

src/cache/cache_las.c
keithbostic: This is right -- we use a scratch buffer and grow it fast if necessary (`__wt_buf_extend()` allocates chunks of memory, not just what's a requirement at the moment), and this copies into just what we need. To just save off the scratch buffer's memory might waste a huge chunk of space, if the scratch buffer is big.

src/evict/evict_page.c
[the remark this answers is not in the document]
keithbostic: I don't think there's any need here for a local `closing` variable, why not just check the flag set in the `WT_ASSERT`?

src/btree/bt_read.c
sulabhM: We are supposed to keep the birthmarks as long as there is any history for the page. In this case, here we have discarded the history and hence do not need it anymore. Later tickets might get rid of the code to read lookaside in the first place, but for now, we need to free both birthmarks and page_las when we discard the history.

src/btree/bt_split.c
sulabhM: The birthmarks_cnt is diagnostic because it is not really needed at the moment for a functional reason. It is just to test the code for correction. At the moment, birthmarks_cnt is always going to be the same as the number of keys on the page, which we determine by going over the LAS file. Certainly, this will change when we stop initialising from LAS, also when we add optimisations around not keeping birthmark per key.

src/evict/evict_page.c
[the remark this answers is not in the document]
sulabhM: fixed

sulabhM:
autobuild: retest this please

sulabhM:
autobuild: retest this please

sulabhM:
autobuild: retest this please

PR #4761, opened 2019-07-11
WT-4927 Allow updates with timestamps to be evicted to lookaside.

michaelcahill:


src/reconcile/rec_visibility.c
```diff
@@ -218,9 +211,8 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 		if (upd_select->upd == NULL && r->las_skew_newest)
 			upd_select->upd = upd;
 
-		if ((F_ISSET(r, WT_REC_VISIBLE_ALL) ?
+		if (F_ISSET(r, WT_REC_VISIBLE_ALL) ?
 		    !__wt_txn_upd_visible_all(session, upd) :
-		    !__wt_txn_upd_visible(session, upd)) ||
 		    !__wt_txn_upd_durable(session, upd)) {
```
michaelcahill: I changed `__wt_txn_upd_visible_all` to respect durable timestamps, in part to simplify this condition.  (I also think it's required in other places).  It's probably time to discuss the `visible_all` naming...

src/evict/evict_lru.c
```diff
@@ -634,10 +634,6 @@ __evict_update_work(WT_SESSION_IMPL *session)
 	    LF_ISSET(WT_CACHE_EVICT_CLEAN_HARD))
 		LF_SET(WT_CACHE_EVICT_DIRTY);
 
-	/* When we stop looking for dirty pages, reduce the lookaside score. */
-	if (!LF_ISSET(WT_CACHE_EVICT_DIRTY))
-		__wt_cache_update_lookaside_score(session, 1, 0);
-
```
michaelcahill: This code should never have been pushed (by me) -- it was an experiment to reduce the lookaside score when no longer considering dirty eviction.

src/btree/bt_read.c
```diff
@@ -862,14 +867,13 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
...
-		 * substitute that for a sleep.
+		 * check if the cache needs help evicting clean pages (don't
+		 * force a read to do dirty eviction).  If we do work for the
+		 * cache, substitute that for a sleep.
 		 */
 		if (!LF_ISSET(WT_READ_IGNORE_CACHE_SIZE)) {
 			WT_RET(__wt_cache_eviction_check(
-			    session, true,
-			    !F_ISSET(&session->txn, WT_TXN_HAS_ID),
-			    &cache_work));
```
michaelcahill: Some subtlety here: I don't think readers should pay to do dirty eviction while waiting for a page read.

src/btree/bt_read.c
```diff
@@ -784,8 +784,13 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
...
 				++force_attempts;
 				ret = __wt_page_release_evict(session, ref, 0);
-				/* If forced eviction fails, stall. */
-				if (ret == EBUSY) {
+				/*
+				 * If forced eviction succeeded, don't retry.
+				 * If it failed, stall.
+				 */
+				if (ret == 0)
+					evict_skip = true;
```
michaelcahill: I don't think it was causing the current issue, but once we have successfully evicted a page (which might be an update/restore), we shouldn't keep retrying.

PR #4763, opened 2019-07-11
WT-4899 Write out no more than a single birthmark for a key in the LAS file

ddanderson:
@tetsuo-cpp, It appears that WT-4899 and WT-4922 are probably the same.  I would like do some more testing of this before merging(\*), but I wanted to put this out for consideration.  (\*) my test scenario is a little convoluted because I'm using an old version of test/format that exposes this problem, but it has an incompatibility with the current develop branch at the moment.


ddanderson:
Failed test/checkpoint.  I guess the approach was a little too simple-minded.  Will take another look tonight/tomorrow.

agorrod:
Thanks for taking a stab at this @ddanderson - I don't think your solution is the right one. We need to keep those updates around - but they need to be resolved and a full record written to lookaside.

That resolution has to happen earlier in the code, since the original page image isn't accessible from `__wt_las_insert_block`. There is code at the end of `__wt_rec_upd_select` which aims to resolve all birthmark updates that could be needed later. That code is called in service of reconciliation from `__wt_rec_row_leaf` (in the case of test_las02 which fails regularly for me, though not every time).

The next step is trying to figure out why reconciliation isn't resolving that birthmark (which should be done in `__rec_append_orig_value`). Since it's Skunkworks time at the moment - I'm happy to take ownership of this one, let me know if you'd like me to do that.

ddanderson:
@agorrod, I'd be happy if you took ownership of this!  Thanks.

michaelcahill:
I pushed a change here to deal with the most recent failure: folks should stare at it before merging because it's fragile for lookaside inserts to rely on this...

agorrod:
@ddanderson it would be great if you can run this branch against the reproducer you were using.

agorrod:
I've switched the sanitizer job over to this branch:

agorrod:
> folks should stare at it before merging because it's fragile for lookaside inserts to rely on this...

I thought about possible alternatives, but didn't come up with anything I like. We could possibly change `__rec_append_orig_value` to resolve birthmarks regardless of visibility checks, but that isn't very appealing because it's used for multiple purposes and it would generally be wasted work.

michaelcahill:
Thanks @agorrod and @ddanderson, this change lgtm now assuming it passes all of our testing.

agorrod:
autobuild: retest this please

agorrod:
This passed testing, and I reconstructed the case @ddanderson mentioned on the JIRA ticket and it failed regularly without this fix and hasn't failed with it. I'm going to go ahead and merge.

PR #4764, opened 2019-07-12
WT-4908 Add new flag for recovery of metadata.

luke-pearson:


luke-pearson:
So there's 8 places in __wt_log_scan that this change will impact:
1. log.c:2269, harmless should just return more error msg's instead of just error.
2. log.c:2284, again more error logging.
3. log.c:2344, this has an impact where we might truncate the log in this pass of __wt_log_scan as opposed to a later pass. I think this is fine.
4. log.c:2369, harmless
5. log.c:2385, this one potentially triggers an assertion and also breaks out of the recovery loop. I think this is fine.
6. log.c:2398, more verbose logging and additionally breaking out of the loop but this only happens if our version is equal to WT_LOG_VERSION_SYSTEM which I don't think it should be in most cases. Again I think this is relatively harmless.
7. log.c:2589, like #3 we truncate early in certain cases.
8. log.c:2615, additional error logging, harmless.

luke-pearson:
retest this please

kommiharibabu:
>log.c:2344, this has an impact where we might truncate the log in this pass of __wt_log_scan as >opposed to a later pass. I think this is fine.

@luke-pearson,
In the first recovery pass, the recovery is happening only for the metadata. It only recovers the log records that are specific to the metadata. If we truncate the logs, we lost the data logs that are not yet recovered. I feel we shouldn't pass the "WT_LOGSCAN_RECOVER" during the metadata restore. Better to add a new macro that doesn't truncate the logs.

PR #4765, opened 2019-07-12
WT-4913 Checksum restructure attempt to keep Windows-specific changes local.

michaelcahill:


michaelcahill:
@keithbostic, I'm relying on automated builds for Windows, so this may be completely broken -- is this obviously misguided?

PR #4768, opened 2019-07-14
WT-4930 review and fix likely coding errors of error returns after jump to error

keithbostic:


src/btree/bt_read.c
```diff
@@ -579,7 +579,7 @@ __page_read(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
 		    session, ref->page_las->las_pageid));
 
 	WT_REF_SET_STATE(ref, final_state);
-	return (ret);
```
agorrod: I don't see how we could come here with ret anything other than `0`, and I think it would be bad if we did. Could you add an assert that `ret == 0` instead or as well?

src/log/log.c
```diff
@@ -2272,7 +2277,7 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
 				    PRIu32 "/%" PRIu32,
 				    lsnp->l.file, lsnp->l.offset);
 			else
-				return (WT_NOTFOUND);
```
agorrod: The comment makes these returns sound deliberate. @sueloverso or @luke-pearson could you chime in as to whether this change is correct?

src/btree/bt_read.c
keithbostic: > Could you add an assert that ret == 0 instead or as well?

Sure, pushed in 9d65fc85a.

src/log/log.c
keithbostic: If the returns are deliberate (and correct), then the preceding `WT_ERR_MSG` calls are wrong.

My suspicion is the code preceding the [general resource allocation, here](https://github.com/wiredtiger/wiredtiger/blob/develop/src/log/log.c#L2303-L2316) could be rewritten to not need a jump to the `err:` lable at all, but I didn't do that work. It should be possible, though, there's just a bit of trickery around the possible failure call to `__wt_fs_directory_list_free()`.

luke-pearson: I think this change is okay but also benign, happy to wait for @sueloverso to get back and share her opinion.

src/log/log.c
```diff
@@ -2272,7 +2277,7 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
...
+				WT_ERR (WT_NOTFOUND);
```
sueloverso: The change is fine as long as we're leaving the function and the caller gets `WT_NOTFOUND` here. 

Minor nit: Please remove the space to be: `WT_ERR(...)`

src/log/log.c
```diff
@@ -2288,7 +2293,7 @@ __wt_log_scan(WT_SESSION_IMPL *session, WT_LSN *lsnp, uint32_t flags,
 				    " larger than biggest log file %" PRIu32,
 				    lsnp->l.file, lsnp->l.offset, lastlog);
 			else
-				return (WT_NOTFOUND);
+				WT_ERR (WT_NOTFOUND);
```
sueloverso: Remove space here too.

src/log/log.c
keithbostic: Fixed in 5251f40.

src/log/log.c
keithbostic: Fixed in 5251f40.

PR #4769, opened 2019-07-14
WT-4931 Fix clang testing in Jenkins Pull Request testing

keithbostic:


dist/s_clang-scan
```diff
@@ -44,9 +44,11 @@ $scan $args make -j 8 wt 1>/dev/null 2>$t
 
 make -j 8 distclean > /dev/null
 
-grep 'No bugs found' $t > /dev/null && exit 0
+(grep 'No bugs found' $t ||
+    diff $t dist/s_clang-scan.diff) > /dev/null && exit 0
```
agorrod: @keithbostic do you think we could strip the line numbers out from the base file and the thing being compared? I think that would make this check significantly less noisy.

keithbostic: Sure, done, pushed in ff7e663.

PR #4770, opened 2019-07-15
WT-4932 Add logging files to WiredTiger big/little-endian testing

keithbostic:


keithbostic:
@lukech 
> Question: why abort=0 is set in the config - is it used to avoid core generation?
Yes, it should simplify testing in this case.

PR #4777, opened 2019-07-18
WT-4940 Reconciliation should set prepared/uncommitted for each update.

michaelcahill:


michaelcahill:
(bug found by @agorrod, I just typed faster)

PR #4778, opened 2019-07-19
WT-4943 Track uncommitted/prepared status through reconciliation.

michaelcahill:
Don't skip selecting an update after seeing an uncommitted / prepared entry in an update list, but remember whether one has been seen when determining whether the page can be evicted.

### DIFF
--- a/src/include/reconcile.h
+++ b/src/include/reconcile.h
@@ -51,8 +51,9 @@ struct __wt_reconcile {
 	u_int updates_seen;		/* Count of updates seen. */
 	u_int updates_unstable;		/* Count of updates not visible_all. */
 
-	bool update_uncommitted;	/* An update was uncommitted */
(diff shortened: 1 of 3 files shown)


michaelcahill:
WT-4940 was a partial fix (don't discriminate against stable updates after seeing one that is uncommitted / prepared) but went too far by forgetting whether an unstable update was seen when deciding whether everything in an update list was committed.

agorrod:
Thanks @michaelcahill this change makes sense to me, and I think removing the local variables makes the code cleaner and less error prone.

michaelcahill:
Thanks @agorrod, but I had second thoughts about this change while it was testing.  In particular, I think the local variables were different in an important way because they were tracking the current update chain, where `r->*` applies to the whole page...  I'll push a follow up change.

PR #4779, opened 2019-07-19
WT-4943 Separate tracking per update list vs per page.

michaelcahill:
We've tried a few times to get this right, so I've renamed the variables to clarify.  Reconciliation as a whole needs to know whether there are any uncommitted updates on a page (to prevent lookaside eviction): that's tracked in `r->updates_uncommitted`.  Selecting a single update needs to know whether any updates *in the list* are prepared or uncommitted.  That's now tracked in `list_prepared` and `list_uncommitted`.


agorrod:
Thanks @michaelcahill - sorry for being trigger happy before.

PR #4780, opened 2019-07-21
WT-4658 Make dist scripts cooperate with Clang Format

tetsuo-cpp:
This is the final PR that needs to be merged in before applying Clang Format. I did some work to get the existing scripts under `dist/` to play nicely with Clang Format.

I've disabled `s_all` since I don't want to reformat the codebase yet. Locally, it works perfectly and after 2 runs, it finishes cleanly and returns 0. We need 2 runs for the initial reformat since some earlier steps in `s_all` see more opportunities to cleanup after later steps have run for the first time (kind of hard to explain in a comment but it's nothing to be alarmed about).


tetsuo-cpp:
retest this please

dist/s_clang-format
```diff
@@ -1,24 +1,39 @@
...
+		wget https://s3.amazonaws.com/boxes.10gen.com/build/clang-format-3.8-rhel55.tar.gz -O dist/clang-format.tar.gz
+		tar --strip=2 -C dist/ -xf dist/clang-format.tar.gz build/bin/clang-format && rm dist/clang-format.tar.gz
+	elif [ `uname` = "Darwin" ]; then
+		wget https://s3.amazonaws.com/boxes.10gen.com/build/clang%2Bllvm-3.8.0-x86_64-apple-darwin.tar.xz -O dist/clang-format.tar.gz
+		tar --strip=2 -C dist/ -xf dist/clang-format.tar.gz clang+llvm-3.8.0-x86_64-apple-darwin/bin/clang-format && rm dist/clang-format.tar.gz
+	else
+		echo "$0: unsupported environment"
+		exit 1
+	fi
+}
```
keithbostic: Does this mean we can't support developers using Windows? (How does server cope with that?)

dist/s_clang-format
```diff
@@ -1,24 +1,39 @@
...
+download_clang_format() {
+	echo "$0: downloading clang-format for $(uname)"
+	if [ `uname` = "Linux" ]; then
+		wget https://s3.amazonaws.com/boxes.10gen.com/build/clang-format-3.8-rhel55.tar.gz -O dist/clang-format.tar.gz
+		tar --strip=2 -C dist/ -xf dist/clang-format.tar.gz build/bin/clang-format && rm dist/clang-format.tar.gz
+	elif [ `uname` = "Darwin" ]; then
+		wget https://s3.amazonaws.com/boxes.10gen.com/build/clang%2Bllvm-3.8.0-x86_64-apple-darwin.tar.xz -O dist/clang-format.tar.gz
+		tar --strip=2 -C dist/ -xf dist/clang-format.tar.gz clang+llvm-3.8.0-x86_64-apple-darwin/bin/clang-format && rm dist/clang-format.tar.gz
+	else
+		echo "$0: unsupported environment"
```
keithbostic: I'd suggest dumping the environment in the error.

src/block/block_ext.c
```diff
@@ -1174,9 +1174,9 @@ __wt_block_extlist_read(WT_SESSION_IMPL *session,
 	WT_DECL_ITEM(tmp);
 	WT_DECL_RET;
 	wt_off_t off, size;
+	const uint8_t *p;
```
keithbostic: This was unexpected, I would have thought `uint8_t` would sort after `int`, but if you think it's right, fbm.

src/btree/bt_import.c
```diff
@@ -23,10 +23,10 @@ __wt_import(WT_SESSION_IMPL *session, const char *uri)
 	WT_DECL_ITEM(checkpoint);
 	WT_DECL_RET;
 	WT_KEYED_ENCRYPTOR *kencryptor;
-	const char *filename;
+	char *checkpoint_list, *fileconf, *metadata, fileid[64];
 	const char *filecfg[] = {
 	   WT_CONFIG_BASE(session, file_meta), NULL, NULL, NULL, NULL, NULL };
-	char *checkpoint_list, *fileconf, *metadata, fileid[64];
+	const char *filename;
```
keithbostic: This was unexpected, I would have expected `const type` to sort either after or before `type`, rather than inter-mixed, but if you think it's right, fbm.

dist/s_clang-format
tetsuo-cpp: a9ddfa2

dist/s_clang-format
tetsuo-cpp: The script that server uses doesn't support Windows unfortunately.
So people developing on Windows are responsible for installing the correct version and adding it to their `PATH` (sorry @luke-pearson).

src/block/block_ext.c
tetsuo-cpp: Yeah it doesn't make much sense to me but that's the order specified in `function.py`.
The reason why I've made this change is because the new line length means that some multiline declarations (which previously caused `function.py` to stop parsing) have been collapsed into a single line. So after formatting, we're able to detect these ordering issues for the first time.

src/btree/bt_import.c
tetsuo-cpp: In `function.py`, `const`, `static` and `volatile` get stripped before applying the sort order.

keithbostic: In retrospect, that seems wrong, but I'm probably the person that wrote it that way.

@agorrod, should we group `const` variables together, or inter-mix them with non-const variables?

src/block/block_ext.c
keithbostic: That one (in retrospect), continues to feel right -- "int" is an undefined size, so shrug.

tetsuo-cpp:
retest this please

src/btree/bt_import.c
agorrod: I feel like they should not be inter-mixed.

keithbostic: @tetsuo-cpp, I know it's a pain, but maybe we could sort based on the type and then re-sort based on the leading static/const/volatile?

I haven't thought about this a lot, but that feels right? Types grouped together, and a secondary sort including the qualifiers?

tetsuo-cpp: 2e77263

tetsuo-cpp:
@lukech 
When's a good time to go ahead with this? Is it too close to release?

tetsuo-cpp:
retest this please

tetsuo-cpp:
retest this please

lukech:
retest this please

tetsuo-cpp:
retest this please

PR #4781, opened 2019-07-22
WT-4906 wtperf.c integer overflow and dead code fixes (Coverity)

kommiharibabu:
CID 113115, CID 113116, CID 113117 and CID 113119 (integer overflow)
Typecasted one of the variable to uint64_t to evaluate the arithmetic
using 64 bits instead of 32 bits.

CID 113118 (dead code)
Removed dead code


kommiharibabu:
retest this please

PR #4783, opened 2019-07-23
WT-4946 Tighten up failure checks in join code and add an assertion to clarify.

ddanderson:


ddanderson:
retest this please

lukech:
retest this please

lukech:
retest this please

ddanderson:
@lukech, thanks for pushing the tests along!

src/cursor/cur_join.c
```diff
@@ -633,15 +636,22 @@ __curjoin_entry_member(WT_SESSION_IMPL *session, WT_CURSOR_JOIN_ENTRY *entry,
...
+		/*
+		 * If we have a subordinate join, the membership
+		 * check is delegated to it.
+		 */
 		WT_ASSERT(session,
 		    iter == NULL || entry->subjoin == iter->child->cjoin);
 		ret = __curjoin_entries_in_range(session, entry->subjoin,
 		    key, iter == NULL ? NULL : iter->child);
+		if (ret != WT_NOTFOUND)
+			WT_ERR(ret);
```
keithbostic: I would worry someone will translate that to `WT_ERR_NOTFOUND_OK`, which isn't safe because it would set `ret = 0`. I'd add an explicit comment out of sheer paranoia, something like "leave ret set to WT_NOTFOUND".

I'm also not clear why we call `__curjoin_iter_bump` if we're going to return `WT_NOTFOUND`, isn't the cursor essentially dead at that point?

ddanderson:
retest this please

keithbostic:
retest this please

PR #4784, opened 2019-07-24
WT-4958 Temporarily never skip checkpoints with a timestamp

agorrod:
It causes failures in MongoDB.


agorrod:
@michaelcahill here is the branch we discussed, I checked and it resolves the failure I was reproducing. Could you please review once it passes testing?

agorrod:
autobuild: retest this please

michaelcahill:
Thanks @agorrod, I'll go ahead and merge because this fixes various MongoDB build failures.  I am convinced that the Jenkins timeout was due to an overloaded test machine.

PR #4786, opened 2019-07-25
WT-4961 Checkpoints with cache overflow must keep history for reads.

michaelcahill:
When cache overflow has been active, checkpoints may need to pull pages into cache in order to rewrite the correct versions.  These pages must be evicted as soon as possible to avoid checkpoint filling the cache with pages that other threads cannot evict (until checkpoint has moved on to another tree).

This leads to the unusual case of attempting eviction of pages while writing disk images that are correct as of the checkpoint's transaction.  In other words, `__wt_txn_visible` is used to select the versions of updates written to the disk image.  However, it is not okay in this case to discard history because everything happens to be visible to a checkpoint: snapshot readers may need an earlier version (back to the oldest timestamp).

### DIFF
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -247,16 +247,12 @@ struct __wt_ovfl_reuse {
 struct __wt_page_lookaside {
 	uint64_t las_pageid;		/* Page ID in lookaside */
 	uint64_t max_txn;		/* Maximum transaction ID */
-	uint64_t unstable_txn;		/* First transaction ID not on page */
-	wt_timestamp_t max_timestamp;	/* Maximum timestamp */
-	wt_timestamp_t unstable_timestamp;/* First timestamp not on page */
-	wt_timestamp_t unstable_durable_timestamp;
-					/* First durable timestamp not on
-					 * page */
+	wt_timestamp_t max_onpage_ts;	/* Maximum timestamp on page */
+	wt_timestamp_t max_ts;		/* Maximum timestamp seen */
+	wt_timestamp_t min_newer_ts;	/* Update newer than on-page */
 	bool eviction_to_lookaside;	/* Revert to lookaside on eviction */
 	bool has_prepares;		/* One or more updates are prepared */
 	bool resolved;			/* History has been read into cache */
-	bool skew_newest;		/* Page image has newest versions */
 };
 
 /*
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -242,13 +242,28 @@ struct __wt_ovfl_reuse {
 
 /*
  * WT_PAGE_LOOKASIDE --
- *	Related information for on-disk pages with lookaside entries.
+ *	Information for on-disk pages with lookaside entries.
+ *
+ * This information is used to decide whether history evicted to lookaside is
+ * needed for a read, and when it is no longer needed at all. We track the
+ * newest update written to the disk image in `max_ondisk_ts`, and the oldest
+ * update skipped to choose the on-disk version in `min_skipped_ts`.  If no
+ * updates were skipped, then the disk image contains the newest versions of
+ * all updates and `min_skipped_ts == WT_TS_MAX`.
+ *
+ * For reads without a timestamp, we check that there are no skipped updates
+ * and that the reader's snapshot can see everything on disk.
+ *
+ * For readers with a timestamp, it is safe to ignore lookaside if either
+ * (a) there are no skipped updates and everything on disk is visible, or
+ * (b) everything on disk is visible, and the minimum skipped update is in
+ * the future of the reader.
  */
 struct __wt_page_lookaside {
 	uint64_t las_pageid;		/* Page ID in lookaside */
 	uint64_t max_txn;		/* Maximum transaction ID */
-	wt_timestamp_t max_ondisk_ts;	/* Maximum timestamp on page */
-	wt_timestamp_t min_skipped_ts;	/* Update newer than on-page */
+	wt_timestamp_t max_ondisk_ts;	/* Maximum timestamp on disk */
+	wt_timestamp_t min_skipped_ts;	/* Skipped in favor of disk version */
 	bool eviction_to_lookaside;	/* Revert to lookaside on eviction */
 	bool has_prepares;		/* One or more updates are prepared */
 	bool resolved;			/* History has been read into cache */
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -247,9 +247,8 @@ struct __wt_ovfl_reuse {
 struct __wt_page_lookaside {
 	uint64_t las_pageid;		/* Page ID in lookaside */
 	uint64_t max_txn;		/* Maximum transaction ID */
-	wt_timestamp_t max_onpage_ts;	/* Maximum timestamp on page */
-	wt_timestamp_t max_ts;		/* Maximum timestamp seen */
-	wt_timestamp_t min_newer_ts;	/* Update newer than on-page */
+	wt_timestamp_t max_ondisk_ts;	/* Maximum timestamp on page */
+	wt_timestamp_t min_skipped_ts;	/* Update newer than on-page */
 	bool eviction_to_lookaside;	/* Revert to lookaside on eviction */
 	bool has_prepares;		/* One or more updates are prepared */
 	bool resolved;			/* History has been read into cache */
--- a/src/reconcile/rec_visibility.c
+++ b/src/reconcile/rec_visibility.c
@@ -398,9 +398,8 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 	if (all_stable)
 		goto check_original_value;
 
-	if (upd != first_txn_upd ||
-	    list_prepared || list_uncommitted ||
-		F_ISSET(r, WT_REC_VISIBLE_ALL) ||
+	if (upd != first_txn_upd || list_prepared || list_uncommitted ||
+	    F_ISSET(r, WT_REC_VISIBLE_ALL) ||
 	    !__wt_txn_visible(session, max_txn, max_ts))
 		r->leave_dirty = true;
 
--- a/src/reconcile/rec_visibility.c
+++ b/src/reconcile/rec_visibility.c
@@ -396,7 +396,9 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 
 	if (upd != first_txn_upd || list_prepared || list_uncommitted ||
 	    F_ISSET(r, WT_REC_VISIBLE_ALL) ||
-	    !__wt_txn_visible(session, max_txn, max_ts))
+	    !(WT_SESSION_BTREE_SYNC(session) ?
+	    __wt_txn_visible_all(session, max_txn, max_ts) :
+	    __wt_txn_visible(session, max_txn, max_ts)))
 		r->leave_dirty = true;
 
 	if (F_ISSET(r, WT_REC_VISIBILITY_ERR))
--- a/src/reconcile/rec_visibility.c
+++ b/src/reconcile/rec_visibility.c
@@ -394,9 +394,10 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 	if (all_stable)
 		goto check_original_value;
 
-	if (upd != first_txn_upd || list_prepared || list_uncommitted ||
+	if (!r->leave_dirty &&
+	    (upd != first_txn_upd || list_prepared || list_uncommitted ||
 	    F_ISSET(r, WT_REC_VISIBLE_ALL) ||
-	    !__wt_txn_visible(session, max_txn, max_ts))
+	    !__wt_txn_visible(session, max_txn, max_ts)))
 		r->leave_dirty = true;
 
 	if (F_ISSET(r, WT_REC_VISIBILITY_ERR))
--- a/src/reconcile/rec_visibility.c
+++ b/src/reconcile/rec_visibility.c
@@ -172,8 +172,6 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
 		 */
 		if (first_txn_upd == NULL)
 			first_txn_upd = upd;
-
-		/* Track the largest transaction ID seen. */
 		if (WT_TXNID_LT(max_txn, txnid))
 			max_txn = txnid;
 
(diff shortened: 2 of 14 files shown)


michaelcahill:
@keithbostic, this change makes sense to me, and seems to fix the original issue found in MongoDB testing.  It's also coming in hot... :-(

That said, while running MongoDB with this change I've hit the assertion at `reconcile.i:191` that `addr->size != 0`, which reconciling a row store internal page (at `rec_row.c:480` as part of ordinary writes for a checkpoint, I think, but I'm trying to reproduce with a build where I can get reliable stacks and examine local variables -- at the moment I can't make it fail in a debug build).

Can you please take a look at this change, and also see if you can figure out how we end up with an empty address that reconciliation doesn't expect?

michaelcahill:
Possibly related: http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-custom-branch/117/console

```
++ nice catchsegv ./t -1 -c ../../../test/format/CONFIG.stress
[1564041086:208763][2806:0x7f50fe5c6740], t, file:wt.wt, WT_SESSION.verify: __wt_rec_child_modify, 212: false
[1564041086:208789][2806:0x7f50fe5c6740], t, file:wt.wt, WT_SESSION.verify: __wt_abort, 30: aborting WiredTiger library
Aborted (core dumped)
```

michaelcahill:
Another version (row store, but inside MongoDB): https://logkeeper.mongodb.org/lobster/build/57a41da39fbecd1929b5b8cc869e864c/test/5d3962c890413065e2e9da35#bookmarks=0%2C3242&l=1

keithbostic:
@michaelcahill, did you review `__wt_page_las_active()`? Should there be any changes there, based on the changes to `__wt_rec_upd_select()`

lukech:
@agorrod with the recent branching of `mongodb-4.4`, we started to have the flexibility of vendor-ing different contents to mongo master and v4.2. I will pay special attention to make sure WT-4961 NOT landing to 4.2 until we gain confidence from testing on master.

lukech:
retest this please

src/reconcile/rec_visibility.c
```diff
@@ -179,22 +193,24 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
...
 				r->update_uncommitted = list_uncommitted = true;
 				continue;
 			}
+			if (upd->prepare_state == WT_PREPARE_LOCKED ||
+			    upd->prepare_state == WT_PREPARE_INPROGRESS) {
+				list_prepared = true;
+				if (upd->start_ts > max_ts)
+					max_ts = upd->start_ts;
+				continue;
+			}
```
michaelcahill: I'd like to talk this part through.  My reasoning for this reordering is that we should consistently check that the transaction ID is committed, then look at `upd->prepare_state`, then look at `start_ts` and/or `durable_ts`.

src/reconcile/rec_visibility.c
```diff
@@ -363,17 +391,18 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins,
...
-	if (all_visible)
+	if (all_stable)
 		goto check_original_value;
 
-	r->leave_dirty = true;
+	if (upd != first_txn_upd ||
+	    list_prepared || list_uncommitted ||
+		F_ISSET(r, WT_REC_VISIBLE_ALL) ||
+	    !__wt_txn_visible(session, max_txn, max_ts))
+		r->leave_dirty = true;
```
michaelcahill: This is the bug fix for the original failure that sparked all of this.  If a checkpoint is doing eviction, it runs with a timestamp, so `WT_REC_VISIBLE_ALL` is not set, but it's not okay to evict a page and discard its history when everything is visible to checkpoint -- there may be an older reader active (or equivalently we may be required to maintain history back to the application-specified `oldest_timestamp`).

With this change, if everything on a page is visible to checkpoint eviction but something is not `visible_all`, the page will still be left clean, but the updates will be saved, which will trigger saving the update to lookaside.

src/include/btmem.h
```diff
@@ -247,16 +247,12 @@ struct __wt_ovfl_reuse {
...
 	uint64_t max_txn;		/* Maximum transaction ID */
-	uint64_t unstable_txn;		/* First transaction ID not on page */
-	wt_timestamp_t max_timestamp;	/* Maximum timestamp */
-	wt_timestamp_t unstable_timestamp;/* First timestamp not on page */
-	wt_timestamp_t unstable_durable_timestamp;
-					/* First durable timestamp not on
-					 * page */
+	wt_timestamp_t max_onpage_ts;	/* Maximum timestamp on page */
+	wt_timestamp_t max_ts;		/* Maximum timestamp seen */
+	wt_timestamp_t min_newer_ts;	/* Update newer than on-page */
```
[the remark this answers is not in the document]
michaelcahill: @bvpvamsikrishna's feedback (in person): better naming would be `max_ondisk_ts` and `min_skipped_ts`.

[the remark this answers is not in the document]
michaelcahill: Further idea from the in-person review: a better check for "no updates newer than onpage" might be `min_skipped_ts == WT_TS_MAX`.

It's possible that some future eviction mode could choose an update with timestamp 100 for key A, then skip over an update with timestamp 90 and choose timestamp 80 for key B.  That would have `max_ts == max_ondisk_ts == 100` but `min_skipped_ts == 90`.

src/txn/txn_rollback_to_stable.c
```diff
@@ -276,14 +274,10 @@ __txn_abort_newer_updates(
...
 		}
-		if (ref->page_las->max_timestamp > rollback_timestamp)
-			ref->page_las->max_timestamp = rollback_timestamp;
-		if (ref->page_las->unstable_durable_timestamp >
-		    rollback_timestamp)
-			ref->page_las->unstable_durable_timestamp =
-			    rollback_timestamp;
-		if (ref->page_las->unstable_timestamp > rollback_timestamp)
-			ref->page_las->unstable_timestamp = rollback_timestamp;
+		if (page_las->max_ondisk_ts > rollback_timestamp)
```
bvpvamsikrishna: Please merge this condition with the above one as both are same.
Also if condition at line 279, better write in reverse order ( i.e if (rollback_timestamp < page_las->min_skipped_ts)), for better readability.

agorrod:
autobuild: retest this please

lukech:
retest this please

lukech:
retest this please

agorrod:
I've switched [stress|http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/configure] testing over to this branch.

If it runs green for a while, and there aren't any 4.2 critical bug fixes coming in I'll merge tomorrow.

lukech:
There's 1 failed run (of the total 4 runs on the feature branch overnight) on the job:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/24594/console
```
++ nice ./t -1 -c ../../../test/format/CONFIG.stress
[1565107478:431093][24285:0x7f0904fcc7c0], t, file:wt.wt, WT_SESSION.salvage: lz4 error: LZ4 decompress error: -19500
t: process 24285
   1: table, row-store (292 seconds)
+ for i in {1..10}
+ eval nice ./t -1 -c ../../../test/format/CONFIG.stress
++ nice ./t -1 -c ../../../test/format/CONFIG.stress
[1565107524:302553][24415:0x7f3c01105700], t, file:wt.wt, WT_SESSION.checkpoint: __wt_rec_cell_build_addr, 191: addr->size != 0
[1565107524:302587][24415:0x7f3c01105700], t, file:wt.wt, WT_SESSION.checkpoint: __wt_abort, 30: aborting WiredTiger library
/tmp/jenkins5374665303989270135.sh: line 24: 24415 Aborted                 (core dumped) nice ./t -1 -c ../../../test/format/CONFIG.stress
```

lukech:
Another hit (with the same issue signature) on the sanitizer job:
http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/24599/console

keithbostic:
@agorrod, @bvpvamsikrishna: what's going on with this branch is that we're checkpointing an internal page, and one of its child pages is in an in-memory state, has never been written to disk (no disk address), and has no visible content:
```
0x2f05010: row-store leaf
        disk (nil), entries 0, clean, replaced, page-state=0, memory-size 3938
        insert {0000089380.00/opqrstuvwxyz}
        value {tombstone}
        txn id 628912, start_ts (0,0)
        value {0000089380/LMNOPQRSTUVWXYZABCD ... }
        txn id 0, start_ts (0,0)
```
At some point the page was evicted (state set to WT_REF_LOOKASIDE), then read in by checkpoint (I'm guessing this is checkpoint writing the leaf page, state set to WT_REF_MEM), then the checkpoint reconciles its parent, and hits the code in `__wt_rec_child_modify()`:
```
mod = ref->page->modify;
if (mod != NULL && mod->rec_result != 0)
        *statep = WT_CHILD_MODIFIED;
else if (ref->addr == NULL) { 
        *statep = WT_CHILD_IGNORE;
        WT_CHILD_RELEASE(session, *hazardp, ref);
}       
```
sets the result of the child review to WT_CHILD_MODIFIED, which means we need to write an address, we don't have one, and things go badly after that.

I don't see anything in the branch that changes how the checkpoint's reconciliation of the leaf page would have otherwise proceeded, but I'm suspicious of the changes to `__las_page_instantiate()` where we decide whether or not this page can be marked clean. I don't see anything on this page that needs to be included in a checkpoint, so perhaps marking it clean is the fix.
```
if (page_las->min_skipped_ts == WT_TS_MAX &&
    !page_las->has_prepares &&
    !S2C(session)->txn_global.has_stable_timestamp &&
    __wt_txn_visible_all(
    session, page_las->max_txn, page_las->max_ondisk_ts)) {
        page->modify->rec_max_txn = page_las->max_txn;
        page->modify->rec_max_timestamp =
            page_las->max_ondisk_ts;
}

(gdb) p *ref.page_las
$10 = {
  las_pageid = 662, 
  max_txn = 17956, 
  max_ondisk_ts = 0, 
  min_skipped_ts = 18446744073709551615, 
  eviction_to_lookaside = false, 
  has_prepares = false, 
  resolved = true
}
(gdb) S2C
$5 = (WT_CONNECTION_IMPL *) 0x126e860
(gdb) p $5.txn_global.has_stable_timestamp
$6 = false
(gdb) p __txn_visible_all_id(session, 17956)
$7 = false
(gdb) p __wt_txn_oldest_id(session)
$8 = 14489
```
So, it's not getting marked clean because `max_txn` isn't globally visible, which doesn't sound correct for a checkpoint?

agorrod:
> I don't see anything on this page that needs to be included in a checkpoint, so perhaps marking it clean is the fix.

I feel like that's not OK. Wouldn't marking the page clean mean that a future checkpoint that should include the update would not look at the page either? It's entirely possible that I'm wrong here - and checkpoints use a different mechanism than clean to determine whether they need to write a page.

I'm curious about what `mod->rec_result` is - I wonder if the code in `__wt_rec_child_modify` should ignore the case where `mod->rec_result == WT_PM_REC_EMPTY` in the first `if` statement at line 322?

Don't we want something that says there is no work for the checkpoint to do for the page, but it isn't safe to free the internal page and it isn't safe for future checkpoints to skip looking at the page?

I'm not an expert on how empty pages work with lookaside and checkpoint. I'd be happy to get on a call and discuss this if that would lead to finding the best solution.

keithbostic:
>> I don't see anything on this page that needs to be included in a checkpoint, so perhaps marking it clean is the fix.

> I feel like that's not OK. Wouldn't marking the page clean mean that a future checkpoint that should include the update would not look at the page either? It's entirely possible that I'm wrong here - and checkpoints use a different mechanism than clean to determine whether they need to write a page.

Since a checkpoint timestamp can't go backwards in time, isn't it a requirement a future timestamp won't need the page's information absent the page being modified?

> I'm curious about what `mod->rec_result` is - I wonder if the code in `__wt_rec_child_modify` should ignore the case where `mod->rec_result == WT_PM_REC_EMPTY` in the first `if` statement at line 322?

`mod.rec_result` is in either `WT_PM_REC_EMPTY` or `WT_PM_REC_REPLACE`, I think, I don't think we can get here with a multi-block replacement. I messed around a bit with adjusting that test, but it always ended up being too broad a check, I always ended up not writing pages that needed to be written for the checkpoint.

> I'm not an expert on how empty pages work with lookaside and checkpoint. I'd be happy to get on a call and discuss this if that would lead to finding the best solution.

Me either -- who's the go-to person on this topic?

Let's talk when you're online.

keithbostic:
FTR, this CONFIG reproduces the problem fastest and most consistently for me.
```
#  RUN PARAMETERS
abort=0
alter=0
assert_commit_timestamp=0
assert_read_timestamp=0
auto_throttle=1
backups=0
bitcnt=5
bloom=1
bloom_bit_count=41
bloom_hash_count=29
bloom_oldest=0
cache=270
cache_minimum=20
checkpoints=on
checkpoint_log_size=87
checkpoint_wait=66
checksum=on
chunk_size=2
compaction=0
compression=snappy
data_extend=0
data_source=table
delete_pct=10
dictionary=0
direct_io=0
encryption=rotn-7
evict_max=1
file_type=row-store
firstfit=0
huffman_key=0
huffman_value=0
independent_thread_rng=1
in_memory=0
insert_pct=11
internal_key_truncation=1
internal_page_max=10
isolation=read-committed
key_gap=1
key_max=74
key_min=14
leaf_page_max=10
leak_memory=0
logging=0
logging_archive=0
logging_compression=none
logging_file_max=435460
logging_prealloc=0
long_running_txn=0
lsm_worker_threads=4
memory_page_max=9
merge_max=12
mmap=1
modify_pct=0
ops=0
prefix_compression=0
prefix_compression_min=4
prepare=0
quiet=1
read_pct=10
rebalance=1
repeat_data_pct=75
reverse=0
rows=1000000
runs=1
salvage=1
split_pct=55
statistics=0
statistics_server=0
threads=15
timer=4
timing_stress_aggressive_sweep=0
timing_stress_checkpoint=1
timing_stress_lookaside_sweep=0
timing_stress_split_1=0
timing_stress_split_2=0
timing_stress_split_3=0
timing_stress_split_4=0
timing_stress_split_5=0
timing_stress_split_6=0
timing_stress_split_7=0
timing_stress_split_8=0
transaction_timestamps=0
transaction-frequency=22
truncate=1
value_max=2569
value_min=20
verify=1
wiredtiger_config=
write_pct=69
```

tetsuo-cpp:
Superseded by #4835.
I'm closing this PR but will keep the branch around. Once you're happy with the migrated branch, please delete this one.

PR #4787, opened 2019-07-25
WT-4956 Handle the case where 4 billion updates are made to a page without eviction

tetsuo-cpp:
This PR is to handle the issue where we consistently have at least one update that doesn't get written to the data file by a checkpoint causing the `write_gen` to never get reset to 0. This type of workload causes the `write_gen` to continually increase until we hit the check in `__page_write_gen_wrapped_check` and begin stalling.

This new scheme involves capping the `write_gen` to 2 since we only need this value to determine whether the page is clean or dirty (or has been dirtied during reconciliation).

### DIFF
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -498,10 +498,9 @@ struct __wt_page_modify {
 	 * WT_PAGE_DIRTY --
 	 *	The page has 2 or more unreconciled updates.
 	 *
-	 * When we reconcile, we artificially set the state to
-	 * single and check afterwards that it hasn't changed
-	 * to determine whether reconciliation
-	 * was able to render the page clean.
+	 * When we reconcile, we artificially set the state to single and check
+	 * afterwards that it hasn't changed to determine whether reconciliation was
+	 * able to render the page clean.
 	 */
 #define	WT_PAGE_CLEAN		0
 #define	WT_PAGE_DIRTY_FIRST	1
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -500,7 +500,7 @@ struct __wt_page_modify {
 	 * to determine whether reconciliation was able to render the page
 	 * clean.
 	 *
-	 * Using write generation is a simple counter of the number of updates
+	 * Using write generation as a simple counter of the number of updates
 	 * is not acceptable as some workloads will get close to hitting
 	 * UINT32_MAX.
 	 */
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -505,7 +505,7 @@ struct __wt_page_modify {
 #define	WT_PAGE_CLEAN		0
 #define	WT_PAGE_DIRTY_FIRST	1
 #define	WT_PAGE_DIRTY		2
-	uint8_t page_state;
+	uint32_t page_state;
 
 #define	WT_PM_REC_EMPTY		1	/* Reconciliation: no replacement */
 #define	WT_PM_REC_MULTIBLOCK	2	/* Reconciliation: multiple blocks */
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -489,22 +489,28 @@ struct __wt_page_modify {
 	WT_SPINLOCK page_lock;		/* Page's spinlock */
 
 	/*
-	 * The write generation is incremented when a page is modified.
+	 * The mod state is incremented when a page is modified.
 	 *
-	 * 0: The page is clean.
-	 * 1: The page has 1 unreconciled update.
-	 * 2: The page has 2 or more unreconciled updates.
+	 * WT_MOD_STATE_CLEAN --
+	 *	The page is clean.
+	 * WT_MOD_STATE_SINGLE --
+	 *	The page has 1 unreconciled update.
+	 * WT_MOD_STATE_MANY --
+	 *	The page has 2 or more unreconciled updates.
 	 *
-	 * When we reconcile, we artificially set the write generation to 1 and
-	 * check afterwards that the write generation hasn't changed (incr to 2)
-	 * to determine whether reconciliation was able to render the page
-	 * clean.
+	 * When we reconcile, we artificially set the mod state to
+	 * WT_MOD_STATE_SINGLE and check afterwards that it hasn't changed
+	 * (increased to WT_MOD_STATE_MANY) to determine whether reconciliation
+	 * was able to render the page clean.
 	 *
-	 * Using write generation as a simple counter of the number of updates
-	 * is not acceptable as some workloads will get close to hitting
-	 * UINT32_MAX.
+	 * Using mod state as a simple counter of the number of updates is not
+	 * acceptable as some workloads will get close to hitting the max size
+	 * of whatever unsigned type we use.
 	 */
-	uint32_t write_gen;
+#define	WT_MOD_STATE_CLEAN	0
+#define	WT_MOD_STATE_SINGLE	1
+#define	WT_MOD_STATE_MANY	2
+	uint8_t mod_state;
 
 #define	WT_PM_REC_EMPTY		1	/* Reconciliation: no replacement */
 #define	WT_PM_REC_MULTIBLOCK	2	/* Reconciliation: multiple blocks */
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -495,12 +495,14 @@ struct __wt_page_modify {
 	 * 1: The page has 1 unreconciled update.
 	 * 2: The page has 2 or more unreconciled updates.
 	 *
-	 * When we reconcile, we artifically set the write generation to 1 and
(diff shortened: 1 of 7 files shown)


src/include/btmem.h
```diff
@@ -489,8 +489,20 @@ struct __wt_page_modify {
...
+	 * 2: The page has 2 or more unreconciled updates.
+	 *
+	 * When we reconcile, we artificially set the write generation to 1 and
+	 * check afterwards that the write generation hasn't changed (incr to 2)
+	 * to determine whether reconciliation was able to render the page
+	 * clean.
+	 *
+	 * Using write generation as a simple counter of the number of updates
+	 * is not acceptable as some workloads will get close to hitting
+	 * UINT32_MAX.
```
sueloverso: If this is the direction that this change is going in, then I'd like to see it move that way all the way. That is, `write_gen` is really not a generation number anymore. It is a state now and should be coded as such. So the field could be named `mod_state` or `page_state` or something similar and the hard-coded numbers should be defined as `..._STATE_CLEAN`, `..._STATE_SINGLE` and `..._STATE_MANY`. I think it will be a simpler concept and the comments will read clearer once it has a slant of state instead of write generation.

tetsuo-cpp: 8a50aa7

tetsuo-cpp:
retest this please

src/reconcile/rec_write.c
```diff
@@ -602,13 +607,20 @@ __rec_init(WT_SESSION_IMPL *session,
...
-	WT_ORDERED_READ(r->orig_write_gen, page->modify->write_gen);
+
+	/*
+	 * Set the mod state to WT_MOD_STATE_SINGLE prior to reconciliation.
+	 * After reconciliation, we want to atomic cas
+	 * WT_MOD_STATE_SINGLE => WT_MOD_STATE_CLEAN to potentially mark it as
+	 * clean. If the mod state got incremented to WT_MOD_STATE_MANY, by a
+	 * concurrent update, it'll remain marked as dirty.
+	 */
+	page->modify->mod_state = WT_MOD_STATE_SINGLE;
```
keithbostic: Don't we have to flush this write?

keithbostic:
FTR, I prefer @michaelcahill's change, it's potentially doing less work.

src/reconcile/rec_write.c
tetsuo-cpp: 5f7e966

src/include/btree.i
```diff
@@ -492,23 +494,42 @@ static inline void
...
 	 *
 	 * Every time the page transitions from clean to dirty, update the cache
 	 * and transactional information.
 	 */
-	if (__wt_atomic_add32(&page->modify->write_gen, 1) == 1) {
+	made_dirty = false;
+	prev_mod_state = page->modify->mod_state;
+	while (prev_mod_state < WT_MOD_STATE_MANY) {
+		if (__wt_atomic_cas8(&page->modify->mod_state,
+		    prev_mod_state, prev_mod_state + 1)) {
```
michaelcahill: We'd typically write this loop as `while ((prev_mod_state = page->modify->mod_state) < WT_MOD_STATE_MANY)` and save the separate reads before and at the end of the loop body.

That said, this code is already doing arithmetic on something we're trying to treat as an `enum` to calculate the new value, so why not use `__wt_atomic_add` (with a type larger than `uint8_t`)?

tetsuo-cpp: Doesn't the `page->modify->mod_state < WT_MOD_STATE_MANY` race with the atomic increment and potentially result in a number greater than 2? The reason I'm doing the cas is to avoid this.

michaelcahill: It does potentially race: several threads could increment concurrently and the value could go over 2.  That's why I'm suggesting a type larger than `uint8_t`.  I don't think it changes the logic anywhere else, though: we only care if it's CLEAN, DIRTY or >=MANY.

tetsuo-cpp: Understood. I've switched over to doing an increment now.

src/include/btree.i
```diff
@@ -496,19 +498,21 @@ __wt_page_only_modify_set(WT_SESSION_IMPL *session, WT_PAGE *page)
 	WT_ASSERT(session, !F_ISSET(session->dhandle, WT_DHANDLE_DEAD));
 
 	last_running = 0;
-	if (page->modify->write_gen == 0)
+	if (page->modify->page_state == WT_PAGE_CLEAN)
 		last_running = S2C(session)->txn_global.last_running;
 
 	/*
-	 * We depend on atomic-add being a write barrier, that is, a barrier to
+	 * We depend on atomic-cas being a write barrier, that is, a barrier to
```
agorrod: This should go back to being `atomic-add` or I'd prefer `the atomic operation` - then we won't need to keep changing it.

I'd like a comment in here about wrapping. Something like:

```
The page state can only ever be incremented above dirty by the number of concurrently running threads, so the counter will never approach the point where it would wrap.
```

src/include/btmem.h
```diff
@@ -489,10 +489,23 @@ struct __wt_page_modify {
...
 
 	/*
-	 * The write generation is incremented when a page is modified, a page
-	 * is clean if the write generation is 0.
+	 * The page state is incremented when a page is modified.
+	 *
+	 * WT_PAGE_CLEAN --
+	 *	The page is clean.
+	 * WT_PAGE_DIRTY_FIRST --
+	 *	The page has 1 unreconciled update.
```
agorrod: I'd prefer `The page is in this state after the first operation that marks a page dirty, or when reconciliation checking to see if it's done enough work to be able to mark the page clean`.

src/include/btmem.h
```diff
@@ -489,10 +489,23 @@ struct __wt_page_modify {
...
+	 * WT_PAGE_DIRTY --
+	 *	The page has 2 or more unreconciled updates.
```
agorrod: I'd prefer `Two or more updates have been added to the page. `

tetsuo-cpp: 4c3cd34

src/include/btmem.h
tetsuo-cpp: 4c3cd34

I removed the block comment below it since it was effectively saying the same thing.

src/include/btree.i
tetsuo-cpp: 4c3cd34

tetsuo-cpp:
retest this please

lukech:
retest this please

src/include/btree.i
```diff
@@ -579,7 +587,7 @@ __wt_page_modify_clear(WT_SESSION_IMPL *session, WT_PAGE *page)
 	 * Allow the call to be made on clean pages.
 	 */
 	if (__wt_page_is_modified(page)) {
-		page->modify->write_gen = 0;
+		page->modify->page_state = WT_PAGE_CLEAN;
```
sueloverso: Do we need a barrier here to make sure it is visible?

agorrod: No - the timing of when other threads see this change of state isn't important

sueloverso: Maybe a comment about why that is here would be good since every other use of the `page_state` field either uses atomic ops or a barrier and comments about it.

agorrod: I think you, Keith Alex and I should get on a call to talk about the need for barriers in this code early next week

tetsuo-cpp: 52269c9

src/reconcile/rec_write.c
```diff
@@ -602,13 +608,22 @@ __rec_init(WT_SESSION_IMPL *session,
...
+
+	/*
+	 * Update the page state to indicate that all currently installed
+	 * updates will be included in this reconciliation if it would mark the
+	 * page clean.
+	 *
+	 * Add a write barrier to make it more likely that a thread adding an
+	 * update will see this state change.
+	 */
+	page->modify->page_state = WT_PAGE_DIRTY_FIRST;
```
tetsuo-cpp: @keithbostic 
I think as per our discussion, this one might need `WT_FULL_BARRIER`.

src/include/btree.i
```diff
@@ -579,7 +587,17 @@ __wt_page_modify_clear(WT_SESSION_IMPL *session, WT_PAGE *page)
...
+		 * The only part where ordering matters is during
+		 * reconciliation where updates on other threads are performing
+		 * writes to the page state that need to be visible to the
+		 * reconciliation thread.
+		 *
+		 * Since clearing of the page state is not going to be happening
+		 * during reconciliation on a separate thread, there's no write
+		 * barrier needed here.
+		 */
+		page->modify->page_state = WT_PAGE_CLEAN;
```
tetsuo-cpp: @keithbostic 
And maybe this one?

src/reconcile/rec_write.c
tetsuo-cpp: 15aacfe

PR #4789, opened 2019-07-26
WT-4965 Avoid eviction during checkpoints that causes amnesia.

michaelcahill:
Testing has uncovered a situation where a checkpoint can discard history required for snapshot readers.  Disable this eviction for now when history can be lost, to make sure snapshot transactions get correct results.


michaelcahill:
This change avoids the situation where eviction during checkpoint can lose history needed by readers.  It isn't a long-term solution: it should mean we get correct results, but may reintroduce some stalls during checkpoints (when checkpoint pulls in lots of pages with history in lookaside, filling the cache with dirty pages that cannot be evicted by other threads).

I've confirmed locally that the MongoDB test case that sparked this work is fixed by this change.

agorrod:
Thanks @michaelcahill this lgtm as well.

PR #4790, opened 2019-07-29
WT-4935 worst case performance of auto calcuating modify API

kommiharibabu:
wiredtiger_calc_modify API is added to cover the update operations
to modify operations to reduce the overhead. This test is to find out
the overhead that this API can cause when the modifications reach
the maximum limit in either maximum size of modification or
maximum number of modifications. The maximum size of modification
and maximum number of modifications are taken from the MongoDB
usage of this API.


kommiharibabu:
retest this please

PR #4793, opened 2019-07-30
WT-4917 fix format bug where truncate operations not correctly compared with previous updates

keithbostic:


keithbostic:
@sueloverso 
> the old code had the same logic, but was actually modifying the `keyno` in the `snap_list` entry itself. The fix is to leave that entry unchanged, and use the same logic on a local variable in `snap_verify`.

Not exactly. Previously, in `snap_repeat_txn()`, we modified `keyno` in the `snap_list` entry itself, yes, and that screwed up the subsequent test in `snap_repeat_ok_match()` because it didn't look like the `snap_list` entry overlapped with a previous update operation, that is, a truncate of `0-N` that overlapped became `N-N` without any overlap. The point of the modification was so we would only test one record in a truncate range, either the first or last one. Since we're no longer doing that modification in `snap_repeat_txn()`, the underlying `snap_verify()` function has to do the adjustment locally every time it's called.

PR #4795, opened 2019-07-30
WT-5036 fix format bug where tracking buffer wrap causes repeatable operation failure

keithbostic:


sueloverso:
autobuild: retest this please

PR #4802, opened 2019-08-05
WT-5038 format heap use after free in __txn_commit_timestamps_assert

tetsuo-cpp:
This PR is to address the use after free issue exposed by WT-4957. Since `__txn_commit_timestamps_assert` actually populates the update list with pointers to data from a given cursor, that cursor needs to stay open until we're finished with the update list.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -797,6 +797,27 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session, WT_CURSOR **cursors)
 }
 
 /*
+ * __txn_commit_free_cursors --
+ *	Close an array of cursors and free the underlying buffer.
+ */
+static void
+__txn_commit_free_cursors(WT_SESSION_IMPL *session, WT_CURSOR **cursors)
+{
+	WT_CURSOR *cursor;
+	WT_TXN *txn;
+	u_int i;
+
+	txn = &session->txn;
+
+	for (i = 0; i < txn->mod_count; ++i) {
+		cursor = cursors[i];
+		if (cursor != NULL)
+			WT_IGNORE_RET(cursor->close(cursor));
+	}
+	__wt_free(session, cursors);
+}
+
+/*
  * __wt_txn_commit --
  *	Commit the current transaction.
  */
@@ -805,7 +826,6 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 {
 	WT_CONFIG_ITEM cval;
 	WT_CONNECTION_IMPL *conn;
-	WT_CURSOR *cursor;
 	WT_CURSOR **cursors;
 	WT_DECL_RET;
 	WT_TXN *txn;
@@ -1040,12 +1060,7 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_STAT_CONN_INCRV(session, txn_prepared_updates_resolved,
 	    resolved_update_count);
 
-	for (i = 0; i < txn->mod_count; ++i) {
-		cursor = cursors[i];
-		if (cursor != NULL)
-			WT_IGNORE_RET(cursor->close(cursor));
-	}
-	__wt_free(session, cursors);
+	__txn_commit_free_cursors(session, cursors);
 
 	txn->mod_count = 0;
 
@@ -1106,6 +1121,8 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	return (0);
 
 err:
+	__txn_commit_free_cursors(session, cursors);
+
 	/*
 	 * If anything went wrong, roll back.
 	 *
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -655,12 +655,6 @@ __wt_txn_release(WT_SESSION_IMPL *session)
 /*
  * __txn_commit_timestamps_assert --
  *	Validate that timestamps provided to commit are legal.
- *
- *	For prepared transactions, this function will open cursors to mutate the
- *	update list with pointers to uncommitted data. After they are finished
- *	with the update list, the caller is responsible for closing every
- *	non-null cursor pointer in the cursors array and freeing the underlying
- *	buffer.
  */
 static inline int
 __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -655,9 +655,15 @@ __wt_txn_release(WT_SESSION_IMPL *session)
 /*
  * __txn_commit_timestamps_assert --
  *	Validate that timestamps provided to commit are legal.
+ *
+ *	For prepared transactions, this function will open cursors to mutate the
+ *	update list with pointers to uncommitted data. After they are finished
+ *	with the update list, the caller is responsible for closing every
+ *	non-null cursor pointer in the cursors array.
  */
 static inline int
-__txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
+__txn_commit_timestamps_assert(
+    WT_SESSION_IMPL *session, WT_CURSOR **cursors, size_t cursors_len)
 {
 	WT_CURSOR *cursor;
 	WT_DECL_RET;
@@ -665,6 +671,7 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
 	WT_TXN_OP *op;
 	WT_UPDATE *upd;
 	wt_timestamp_t op_timestamp;
+	size_t cursor_index;
 	u_int i;
 	const char *open_cursor_cfg[] = {
 	    WT_CONFIG_BASE(session, WT_SESSION_open_cursor), NULL };
@@ -706,6 +713,7 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
 	 * Error on any valid update structures for the same key that
 	 * are at a later timestamp or use timestamps inconsistently.
 	 */
+	cursor_index = 0;
 	for (i = 0, op = txn->mod; i < txn->mod_count; i++, op++)
 		if (op->type == WT_TXN_OP_BASIC_COL ||
 		    op->type == WT_TXN_OP_BASIC_ROW) {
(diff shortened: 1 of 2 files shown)


src/txn/txn.c
```diff
@@ -816,6 +823,10 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	txn_global = &conn->txn_global;
 	locked = skip_update_assert = false;
 	resolved_update_count = visited_update_count = 0;
+	cursors = NULL;
+	if (txn->mod_count != 0)
+		WT_ERR(__wt_calloc(
```
tetsuo-cpp: Unsure how perf sensitive this is. Is it fine for me to allocate here? Or should I make some attempt to use the stack first like we're doing here: https://github.com/wiredtiger/wiredtiger/blob/develop/src/btree/bt_ret.c#L195

src/txn/txn.c
```diff
@@ -716,7 +721,8 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
 			if (F_ISSET(txn, WT_TXN_PREPARE)) {
 				WT_RET(__wt_open_cursor(session,
 				    op->btree->dhandle->name, NULL,
```
tetsuo-cpp: A potential optimisation could be to check if we've already got a cursor opened for that dhandle. In that case I think we can just reuse the cursor. But the bookkeeping logic is going to become more complex.

Does anyone think this is a good idea?

src/txn/txn.c
kommiharibabu: commit is definitely a performance flow function, whether to use the stack instead of heap depends on the cost of the new change? If you feel that it adds overhead, better to move it to stack.

agorrod:
@sueloverso could you please do a review of this change?

src/txn/txn.c
tetsuo-cpp: I have a hunch that the opening/closing of cursors + the repeated calls to `__wt_btcur_search_uncommitted` are going to dwarf this. I'm thinking of just leaving this as is unless there is an observable perf degradation after merging.

@kommiharibabu To my understanding, this strategy will still fallback to the heap for longer update lists. It will just leave enough space on the stack for the "common case", kind of like how SSO is done.

kommiharibabu: I am also fine to go with the current change until it has major performance impact.

sueloverso: This checking in `txn_commit_timestamps_assert` is only active when `assert=(...)` is configured so no customers/users will ever have that turned on. We do not want to penalize all users with the memory allocation/free for a very rare debugging case. Couldn't this allocation be done down in `txn_commit_timestamps_assert` after the return if no debugging is turned on? Then only users of the debugging feature pay the price and `cursors` will just stay NULL in the common case.

src/txn/txn.c
```diff
@@ -1029,6 +1060,8 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 	WT_STAT_CONN_INCRV(session, txn_prepared_updates_resolved,
 	    resolved_update_count);
 
+	__txn_commit_free_cursors(session, cursors);
```
sueloverso: Then here you'd need to either only call this `if (cursors != NULL)` (which will be the rare case) or have that function return immediately `if (cursors == NULL)`. The former avoids the function call overhead and would be my preference but a function call will not impact performance.

src/txn/txn.c
sueloverso: This is a rarely used debug path. I would do what is easiest and most readable.

src/txn/txn.c
tetsuo-cpp: 6be3417

The function should be safe to call even on `NULL` but I've added a check anyway.

src/txn/txn.c
tetsuo-cpp: 6be3417

Only allocating in the debugging case now.

src/txn/txn.c
tetsuo-cpp: Understood. I'll keep it as is for the time being.

tetsuo-cpp:
retest this please

sueloverso:
Have you pointed some of the failing Jenkins sanitizer jobs at this branch to make sure the problem is fixed, or otherwise reproduced and verified the solution?

tetsuo-cpp:
> Have you pointed some of the failing Jenkins sanitizer jobs at this branch to make sure the problem is fixed, or otherwise reproduced and verified the solution?

I've been able to reproduce this pretty consistently locally and this fix solves the problem for me. I haven't run the Jenkins jobs on this branch though.

tetsuo-cpp:
retest this please

lukech:
retest this please

src/txn/txn.c
```diff
@@ -655,11 +655,18 @@ __wt_txn_release(WT_SESSION_IMPL *session)
...
+ *	For prepared transactions, this function will open cursors to mutate the
+ *	update list with pointers to uncommitted data. After they are finished
+ *	with the update list, the caller is responsible for closing every
+ *	non-null cursor pointer in the cursors array and freeing the underlying
+ *	buffer.
  */
 static inline int
-__txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
+__txn_commit_timestamps_assert(
+    WT_SESSION_IMPL *session, WT_CURSOR ***cursor_buf)
```
tetsuo-cpp: http://wiki.c2.com/?ThreeStarProgrammer

tetsuo-cpp:
retest this please

tetsuo-cpp:
retest this please

tetsuo-cpp:
retest this please

tetsuo-cpp:
Just FYI, this is ready to go. We've had issues with Jenkins PR tests timing out over the past few days.

src/txn/txn.c
agorrod: For the record. It wouldn't be OK to re-use a cursor for another op, if we need to keep the op in cache. The cursor needs to remain positioned on the page until it would be OK for the page to be free'd, and re-using the cursor might leave it positioned on another page.

agorrod:
@tetsuo-cpp sorry for the long delay in reviewing this. Let's chat on Monday and get it over the line.

src/txn/txn.c
sueloverso: Ha ha! We don't have a lot of them, but we do have a few. `os_posix/os_dir.c, btree/bt_split.c` for a couple examples. The comment is very helpful.

src/txn/txn.c
```diff
@@ -702,6 +709,13 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
 	if (!F_ISSET(txn, WT_TXN_TS_COMMIT_KEYS | WT_TXN_TS_DURABLE_KEYS))
 		return (0);
 
+	cursors = NULL;
+	if (txn->mod_count != 0) {
+		WT_RET(__wt_calloc(
+		    session, sizeof(WT_CURSOR *), txn->mod_count, cursor_buf));
+		cursors = *cursor_buf;
```
sueloverso: A more typical usage (used in the files I noted above) is to allocate into the local variable `&cursors` (maybe use `cursor_list` everywhere?) and then set the return variable after that is successful. The word `buf` doesn't help me though. Maybe something like `cursor_listp`?

tetsuo-cpp: 3009132

I've fixed the naming of these cursor variables to be more consistent.

As for the suggestion about only assigning to the return variable on success, I think that in this instance we should stray from the conventional usage. The reason is that the assertion function actually populates the update list with pointers to uncommitted data. With the suggested approach, if some assertion fails, we'll close the cursors and free that data making the update list unusable to the caller (walking will invoke UB).

Right now the difference is immaterial because if this assertion function returns non-zero, the caller won't be doing anything with the update lists. But it seems more error prone to me and I can see someone getting hit by this in the future.

sueloverso:
While I know that I said the efficiency didn't matter, allocating a cursor per mod, I've been reconsidering that. It may be worthwhile running the test in `bench/wtperf/runners/long-txn-btree.wtperf` with and without  the assert mode that would get us into this code and see what the effect is. That test does 100K `ops_per_txn`.

src/txn/txn.c
```diff
@@ -702,6 +709,13 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
 	if (!F_ISSET(txn, WT_TXN_TS_COMMIT_KEYS | WT_TXN_TS_DURABLE_KEYS))
 		return (0);
 
+	cursors = NULL;
+	if (txn->mod_count != 0) {
+		WT_RET(__wt_calloc(
+		    session, sizeof(WT_CURSOR *), txn->mod_count, cursor_buf));
```
sulabhM: Is "sizeof(WT_CURSOR *)" correct here. Should it be "sizeof(WR_CUROSR **)". I know a pointer of any kind would be of the same size so this is purely semantics I think.

src/txn/txn.c
```diff
@@ -864,7 +901,7 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 		WT_ASSERT(session,
 		    txn->commit_timestamp <= txn->durable_timestamp);
 
-	WT_ERR(__txn_commit_timestamps_assert(session));
+	WT_ERR(__txn_commit_timestamps_assert(session, &cursors));
```
sulabhM: I am unable to follow the change. We do not use "cursors" anywhere after this point and free at the end of the function. Can we in that case instead just have cursors as a local variable in {{__txn_commit_timestamps_assert}}. I am clearly missing something here.

src/txn/txn.c
tetsuo-cpp: I think it's ok but I had to double check when I wrote this so maybe I'm wrong. I want an array of `WT_CURSOR *`. So the size is `sizeof(WT_CURSOR *)` and the number of elements is `txn->mod_count`.

src/txn/txn.c
tetsuo-cpp: The trouble comes from this part here: https://github.com/wiredtiger/wiredtiger/blob/develop/src/txn/txn.c#L735
So the cursors need to stay open for any subsequent traversal of the update lists. I'm going to discuss with @agorrod later as maybe there's a way to avoid this.

src/txn/txn.c
```diff
@@ -751,6 +755,21 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
 			 */
 			if (upd == NULL)
```
tetsuo-cpp: @agorrod @bvpvamsikrishna 
I think we need to close the cursor here too.

tetsuo-cpp: c464c44

agorrod: I agree. I'd prefer to wrap the assignments at line 759 and 760 in an `if (upd != NULL)`, and then move the continue below the cursor close.

tetsuo-cpp: Sounds good. If I'm understanding your comment, I think that's what I did in the commit above.

tetsuo-cpp:
@sueloverso 
I had an in-person discussion with @agorrod and @bvpvamsikrishna today so there's some missing context here on why we made these changes.

The only reason why I was managing this buffer of cursors with a weird lifetime was because were later inspecting the txn op's update list which was populated with pointers to data belonging to those cursors. It turns out that the spots where we were using the update lists in the prepare path are avoidable so we can avoid the use of this cursor buffer.

The main usage was a line in `__wt_txn_resolve_prepared_op` which we decided should never be getting hit and is safe to remove. The other usages in `__txn_commit_timestamps_assert` are more trivial and just involve copying into temporaries so the actual `WT_UPDATE` doesn't need to remain alive.

sueloverso:
Ok! Where this ended up is certainly a lot cleaner and simpler.

PR #4803, opened 2019-08-05
WT-4935 Add a perf test to find out the wiredtiger_calc_modify overhead

kommiharibabu:
The wiredtiger_calc_modify function is used to find out the modify vector
based on the old and new record values. This function may fail when the
changed modifications of the new record reaches the allowed limit of
modification for the MODIFY operation, instead of UPDATE.

This perf test is to find out the extra overhead because of
wiredtiger_calc_modify function where it fails to find out the modify
vectors because of reach of the limits configured for the MODIFY operation
and at the end the operations is carried out as normal update.

### DIFF
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -220,6 +220,33 @@ config_threads(WTPERF *wtperf, const char *config, size_t len)
                     goto err;
                 continue;
             }
+            if (STRING_MATCH("modify", k.str, k.len)) {
+                if ((workp->modify = v.val) < 0)
+                    goto err;
+                continue;
+            }
+            if (STRING_MATCH("modify_delta", k.str, k.len)) {
+                if (v.type == WT_CONFIG_ITEM_ID || v.type == WT_CONFIG_ITEM_STRING) {
+                    if (strncmp(v.str, "rand", 4) != 0)
+                        goto err;
+                    /* Special random value */
+                    workp->modify_delta = INT64_MAX;
+                    F_SET(wtperf, CFG_GROW);
+                } else {
+                    workp->modify_delta = v.val;
+                    if (v.val > 0)
+                        F_SET(wtperf, CFG_GROW);
+                    if (v.val < 0)
+                        F_SET(wtperf, CFG_SHRINK);
+                }
+                continue;
+            }
+            if (STRING_MATCH("modify_force_update", k.str, k.len)) {
+                if (v.type != WT_CONFIG_ITEM_BOOL)
+                    goto err;
+                workp->modify_force_update = v.val;
+                continue;
+            }
             if (STRING_MATCH("ops_per_txn", k.str, k.len)) {
                 if ((workp->ops_per_txn = v.val) < 0)
                     goto err;
@@ -302,7 +329,8 @@ config_threads(WTPERF *wtperf, const char *config, size_t len)
         scan = NULL;
         if (ret != 0)
             goto err;
-        if (workp->insert == 0 && workp->read == 0 && workp->update == 0 && workp->truncate == 0)
+        if (workp->insert == 0 && workp->modify == 0 && workp->read == 0 && workp->truncate == 0 &&
+          workp->update == 0)
             goto err;
         /* Why run with truncate if we don't want any truncation. */
         if (workp->truncate != 0 && workp->truncate_pct == 0 && workp->truncate_count == 0)
@@ -312,7 +340,8 @@ config_threads(WTPERF *wtperf, const char *config, size_t len)
         /* Truncate should have its own exclusive thread. */
         if (workp->truncate != 0 && workp->threads > 1)
             goto err;
-        if (workp->truncate != 0 && (workp->insert > 0 || workp->read > 0 || workp->update > 0))
+        if (workp->truncate != 0 &&
+          (workp->insert > 0 || workp->modify > 0 || workp->read > 0 || workp->update > 0))
             goto err;
         wtperf->workers_cnt += (u_int)workp->threads;
     }
@@ -765,11 +794,12 @@ config_sanity(WTPERF *wtperf)
 
     if (wtperf->workload != NULL)
         for (i = 0, workp = wtperf->workload; i < wtperf->workload_cnt; ++i, ++workp) {
-            if (opts->readonly &&
-              (workp->insert != 0 || workp->update != 0 || workp->truncate != 0)) {
+            if (opts->readonly && (workp->insert != 0 || workp->modify != 0 ||
+                                    workp->truncate != 0 || workp->update != 0)) {
                 fprintf(stderr,
-                  "Invalid workload: insert, update or "
-                  "truncate specified with readonly\n");
+                  "Invalid workload: insert, modify, "
+                  "truncate or update specified with "
+                  "readonly\n");
                 return (EINVAL);
             }
             if (workp->insert != 0 && workp->table_index != INT32_MAX) {
@@ -783,6 +813,13 @@ config_sanity(WTPERF *wtperf)
                 fprintf(stderr,
                   "Workload table index %" PRId32 " is larger than table count %" PRIu32,
                   workp->table_index, opts->table_count);
+                return (EINVAL);
+            }
+            if (workp->modify != 0 && workp->ops_per_txn == 0) {
+                fprintf(stderr,
+                  "Modify operations must be executed with "
+                  "explicit transaction, specify "
+                  "ops_per_txn.");
                 return (EINVAL);
             }
         }
--- a/bench/wtperf/config.c
+++ b/bench/wtperf/config.c
@@ -797,8 +797,7 @@ config_sanity(WTPERF *wtperf)
             if (opts->readonly && (workp->insert != 0 || workp->modify != 0 ||
                                     workp->truncate != 0 || workp->update != 0)) {
                 fprintf(stderr,
-                  "Invalid workload: insert, modify, "
-                  "truncate or update specified with "
+                  "Invalid workload: insert, modify, truncate or update specified with "
                   "readonly\n");
                 return (EINVAL);
             }
@@ -817,8 +816,7 @@ config_sanity(WTPERF *wtperf)
             }
             if (workp->modify != 0 && workp->ops_per_txn == 0) {
                 fprintf(stderr,
-                  "Modify operations must be executed with "
-                  "explicit transaction, specify "
+                  "Modify operations must be executed with explicit transaction, specify "
                   "ops_per_txn.");
                 return (EINVAL);
             }
--- a/bench/wtperf/runners/modify-force-update-large-record-btree.wtperf
+++ b/bench/wtperf/runners/modify-force-update-large-record-btree.wtperf
@@ -1,0 +1,14 @@
+# wtperf options file: btree with updates. The goal here is to have a workload of
+# large documents that uses a lot of transaction IDs for modify with forced update
+# to ensure that update operation happened instead of modify.
+conn_config="cache_size=20G"
+sess_config="isolation=snapshot"
+table_config="type=file"
+key_sz=40
+value_sz=100000
+icount=500000
+report_interval=5
+random_value=true
+run_time=120
+populate_threads=1
+threads=((count=4,modify=1,ops_per_txn=1,modify_force_update=true))
--- a/bench/wtperf/runners/modify-large-record-btree.wtperf
+++ b/bench/wtperf/runners/modify-large-record-btree.wtperf
@@ -1,4 +1,4 @@
-# wtperf options file: btree with updatesy. The goal here is to have a workload
+# wtperf options file: btree with updates. The goal here is to have a workload
 # of large documents that uses a lot of transaction IDs for modify operations
 conn_config="cache_size=20G"
 sess_config="isolation=snapshot"
--- a/bench/wtperf/runners/modify-large-record-btree.wtperf
+++ b/bench/wtperf/runners/modify-large-record-btree.wtperf
@@ -1,0 +1,13 @@
+# wtperf options file: btree with updatesy. The goal here is to have a workload
+# of large documents that uses a lot of transaction IDs for modify operations
+conn_config="cache_size=20G"
+sess_config="isolation=snapshot"
+table_config="type=file"
+key_sz=40
+value_sz=100000
+icount=500000
+report_interval=5
+random_value=true
+run_time=120
+populate_threads=1
+threads=((count=4,modify=1,ops_per_txn=1))
--- a/bench/wtperf/runners/wtperf_run.sh
+++ b/bench/wtperf/runners/wtperf_run.sh
@@ -39,18 +39,18 @@ rm -f $outfile
 echo "Parsed $# args: test: $wttest runmax: $runmax args: $wtarg" >> $outfile
 
 # Each of these has an entry for each op in ops below.
-avg=(0 0 0 0)
-max=(0 0 0 0)
-min=(0 0 0 0)
-sum=(0 0 0 0)
+avg=(0 0 0 0 0)
+max=(0 0 0 0 0)
+min=(0 0 0 0 0)
+sum=(0 0 0 0 0)
 # Load needs floating point and bc, handle separately.
-loadindex=5
+loadindex=6
 avg[$loadindex]=0
 max[$loadindex]=0
 min[$loadindex]=0
 sum[$loadindex]=0
-ops=(read insert update truncate)
-outp=("Read count:" "Insert count:" "Update count:" "Truncate count:")
+ops=(insert modify read truncate update)
+outp=("Insert count:" "Modify count:" "Read count:" "Truncate count:" "Update count:"  )
 outp[$loadindex]="Load time:"
 
 # getval min/max val cur
--- a/bench/wtperf/track.c
+++ b/bench/wtperf/track.c
@@ -118,6 +118,11 @@ sum_insert_ops(WTPERF *wtperf)
     return (sum_ops(wtperf, offsetof(WTPERF_THREAD, insert)));
 }
 uint64_t
+sum_modify_ops(WTPERF *wtperf)
+{
+    return (sum_ops(wtperf, offsetof(WTPERF_THREAD, modify)));
+}
+uint64_t
 sum_read_ops(WTPERF *wtperf)
 {
     return (sum_ops(wtperf, offsetof(WTPERF_THREAD, read)));
@@ -185,11 +190,11 @@ latency_op(WTPERF *wtperf, size_t field_offset, uint32_t *avgp, uint32_t *minp, 
     }
 }
 void
-latency_read(WTPERF *wtperf, uint32_t *avgp, uint32_t *minp, uint32_t *maxp)
+latency_insert(WTPERF *wtperf, uint32_t *avgp, uint32_t *minp, uint32_t *maxp)
 {
     static uint32_t last_avg = 0, last_max = 0, last_min = 0;
 
-    latency_op(wtperf, offsetof(WTPERF_THREAD, read), avgp, minp, maxp);
+    latency_op(wtperf, offsetof(WTPERF_THREAD, insert), avgp, minp, maxp);
 
     /*
      * If nothing happened, graph the average, minimum and maximum as they were the last time, it
@@ -206,11 +211,32 @@ latency_read(WTPERF *wtperf, uint32_t *avgp, uint32_t *minp, uint32_t *maxp)
     }
 }
 void
-latency_insert(WTPERF *wtperf, uint32_t *avgp, uint32_t *minp, uint32_t *maxp)
+latency_modify(WTPERF *wtperf, uint32_t *avgp, uint32_t *minp, uint32_t *maxp)
 {
     static uint32_t last_avg = 0, last_max = 0, last_min = 0;
 
-    latency_op(wtperf, offsetof(WTPERF_THREAD, insert), avgp, minp, maxp);
+    latency_op(wtperf, offsetof(WTPERF_THREAD, modify), avgp, minp, maxp);
+
+    /*
+     * If nothing happened, graph the average, minimum and maximum as they were the last time, it
+     * keeps the graphs from having discontinuities.
+     */
+    if (*minp == 0) {
+        *avgp = last_avg;
+        *minp = last_min;
+        *maxp = last_max;
+    } else {
+        last_avg = *avgp;
+        last_min = *minp;
+        last_max = *maxp;
+    }
+}
+void
+latency_read(WTPERF *wtperf, uint32_t *avgp, uint32_t *minp, uint32_t *maxp)
+{
+    static uint32_t last_avg = 0, last_max = 0, last_min = 0;
+
+    latency_op(wtperf, offsetof(WTPERF_THREAD, read), avgp, minp, maxp);
 
     /*
      * If nothing happened, graph the average, minimum and maximum as they were the last time, it
@@ -286,6 +312,11 @@ sum_insert_latency(WTPERF *wtperf, TRACK *total)
     sum_latency(wtperf, offsetof(WTPERF_THREAD, insert), total);
 }
 static void
+sum_modify_latency(WTPERF *wtperf, TRACK *total)
+{
+    sum_latency(wtperf, offsetof(WTPERF_THREAD, modify), total);
+}
+static void
 sum_read_latency(WTPERF *wtperf, TRACK *total)
 {
     sum_latency(wtperf, offsetof(WTPERF_THREAD, read), total);
@@ -344,6 +375,8 @@ latency_print(WTPERF *wtperf)
 
     sum_insert_latency(wtperf, &total);
     latency_print_single(wtperf, &total, "insert");
+    sum_modify_latency(wtperf, &total);
+    latency_print_single(wtperf, &total, "modify");
     sum_read_latency(wtperf, &total);
     latency_print_single(wtperf, &total, "read");
     sum_update_latency(wtperf, &total);
--- a/bench/wtperf/wtperf.c
+++ b/bench/wtperf/wtperf.c
@@ -39,7 +39,7 @@ static int execute_workload(WTPERF *);
 static int find_table_count(WTPERF *);
 static WT_THREAD_RET monitor(void *);
 static WT_THREAD_RET populate_thread(void *);
-static void randomize_value(WTPERF_THREAD *, char *);
+static void randomize_value(WTPERF_THREAD *, char *, int64_t);
 static void recreate_dir(const char *);
 static WT_THREAD_RET scan_worker(void *);
 static int start_all_runs(WTPERF *);
@@ -48,7 +48,7 @@ static void start_threads(
   WTPERF *, WORKLOAD *, WTPERF_THREAD *, u_int, WT_THREAD_CALLBACK (*)(void *));
 static void stop_threads(u_int, WTPERF_THREAD *);
 static WT_THREAD_RET thread_run_wtperf(void *);
-static void update_value_delta(WTPERF_THREAD *);
+static void update_value_delta(WTPERF_THREAD *, int64_t);
 static WT_THREAD_RET worker(void *);
 
 static uint64_t wtperf_rand(WTPERF_THREAD *);
@@ -65,10 +65,11 @@ get_next_incr(WTPERF *wtperf)
 
 /*
  * Each time this function is called we will overwrite the first and one other element in the value
- * buffer.
+ * buffer. The delta can be zero for insert operations and may have a value for update or modify
+ * workloads.
  */
 static void
-randomize_value(WTPERF_THREAD *thread, char *value_buf)
+randomize_value(WTPERF_THREAD *thread, char *value_buf, int64_t delta)
 {
     CONFIG_OPTS *opts;
     uint8_t *vb;
@@ -78,13 +79,13 @@ randomize_value(WTPERF_THREAD *thread, char *value_buf)
 
     /*
      * Limit how much of the buffer we validate for length, this means that only threads that do
-     * growing updates will ever make changes to values outside of the initial value size, but
-     * that's a fair trade off for avoiding figuring out how long the value is more accurately in
-     * this performance sensitive function.
+     * growing updates/modifies will ever make changes to values outside of the initial value size,
+     * but that's a fair trade off for avoiding figuring out how long the value is more accurately
+     * in this performance sensitive function.
      */
-    if (thread->workload == NULL || thread->workload->update_delta == 0)
+    if (delta == 0)
         max_range = opts->value_sz;
-    else if (thread->workload->update_delta > 0)
+    else if (delta > 0)
         max_range = opts->value_sz_max;
     else
         max_range = opts->value_sz_min;
@@ -134,21 +135,20 @@ map_key_to_table(CONFIG_OPTS *opts, uint64_t k)
 }
 
 /*
- * Figure out and extend the size of the value string, used for growing updates. We know that the
- * value to be updated is in the threads value scratch buffer.
+ * Figure out and extend the size of the value string, used for growing updates/modifies. Delta is
+ * the value to be updated according to the thread operation.
  */
 static inline void
-update_value_delta(WTPERF_THREAD *thread)
+update_value_delta(WTPERF_THREAD *thread, int64_t delta)
 {
     CONFIG_OPTS *opts;
     WTPERF *wtperf;
     char *value;
-    int64_t delta, len, new_len;
+    int64_t len, new_len;
 
     wtperf = thread->wtperf;
     opts = wtperf->opts;
     value = thread->value_buf;
-    delta = thread->workload->update_delta;
     len = (int64_t)strlen(value);
 
     if (delta == INT64_MAX)
@@ -296,6 +296,8 @@ op_name(uint8_t *op)
         return ("insert");
     case WORKER_INSERT_RMW:
         return ("insert_rmw");
+    case WORKER_MODIFY:
+        return ("modify");
     case WORKER_READ:
         return ("read");
     case WORKER_TRUNCATE:
@@ -385,14 +387,15 @@ worker_async(void *arg)
             goto op_err;
         case WORKER_INSERT:
             if (opts->random_value)
-                randomize_value(thread, value_buf);
+                randomize_value(thread, value_buf, 0);
             asyncop->set_value(asyncop, value_buf);
             if ((ret = asyncop->insert(asyncop)) == 0)
                 break;
             goto op_err;
         case WORKER_UPDATE:
             if (opts->random_value)
-                randomize_value(thread, value_buf);
+                randomize_value(
+                  thread, value_buf, thread->workload ? thread->workload->update_delta : 0);
             asyncop->set_value(asyncop, value_buf);
             if ((ret = asyncop->update(asyncop)) == 0)
                 break;
@@ -504,14 +507,16 @@ worker(void *arg)
     WTPERF_THREAD *thread;
     WT_CONNECTION *conn;
     WT_CURSOR **cursors, *cursor, *log_table_cursor, *tmp_cursor;
+    WT_ITEM newv, oldv;
+    WT_MODIFY entries[MAX_MODIFY_NUM];
     WT_SESSION *session;
-    size_t i;
-    uint32_t total_table_count;
-    int64_t ops, ops_per_txn;
+    size_t i, iter, modify_offset, modify_size, total_modify_size;
+    int64_t delta, ops, ops_per_txn;
     uint64_t log_id, next_val, usecs;
+    uint32_t rand_val, total_table_count;
     uint8_t *op, *op_end;
-    int measure_latency, ret, truncated;
-    char *value_buf, *key_buf, *value;
+    int measure_latency, nmodify, ret, truncated;
+    char *key_buf, *value, *value_buf;
     char buf[512];
 
     thread = (WTPERF_THREAD *)arg;
@@ -607,12 +612,17 @@ worker(void *arg)
             else
                 next_val = opts->icount + get_next_incr(wtperf);
             break;
+        case WORKER_MODIFY:
+            trk = &thread->modify;
+        /* FALLTHROUGH */
         case WORKER_READ:
-            trk = &thread->read;
+            if (*op == WORKER_READ)
+                trk = &thread->read;
         /* FALLTHROUGH */
         case WORKER_UPDATE:
             if (*op == WORKER_UPDATE)
                 trk = &thread->update;
+
             next_val = wtperf_rand(thread);
 
             /*
@@ -681,7 +691,7 @@ worker(void *arg)
         /* FALLTHROUGH */
         case WORKER_INSERT:
             if (opts->random_value)
-                randomize_value(thread, value_buf);
+                randomize_value(thread, value_buf, 0);
             cursor->set_value(cursor, value_buf);
             if ((ret = cursor->insert(cursor)) == 0)
                 break;
@@ -697,6 +707,7 @@ worker(void *arg)
                 break;
             }
             goto op_err;
+        case WORKER_MODIFY:
         case WORKER_UPDATE:
             if ((ret = cursor->search(cursor)) == 0) {
                 if ((ret = cursor->get_value(cursor, &value)) != 0) {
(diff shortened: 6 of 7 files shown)


bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+						/*
+						 * Pass the old and new data
+						 * to find out the modify
+						 * vectors, according to the
+						 * passed input, this function
+						 * fail as the modifications
+						 * count reaches the maximum
+						 * number of modifications
+						 * that are allowed for MODIFY
+						 * operation.
```
ddanderson: Why do we care that the calc_modify function fails?  If we need it to fail, it would be good to list that in the comment.  If we don't need it to fail, perhaps let it succeed?

I assume we are just calling it to incur the performance overhead of the call?  If that's the case, it would be good to just mention that in the comment.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+				 * whole document
+				 */
+				if (opts->distribute_modify)
+				{
+					int nentries = 16;
+					size_t iter;
+					size_t total_modify_size;
+					size_t modify_size;
+					size_t modify_offset;
+
```
ddanderson: I was initially worried that this code didn't handle the case of small values of `value_sz` (like values less than 10 or 16).  I see that `modify_size` will be zero unless `value_sz >= 160`.  Maybe a comment to that effect: "We don't need to worry about small value sizes (less than 10 * nentries), as modify_size will end up zero, and we won't do any copying at all."

bench/wtperf/wtperf.c
ddanderson: I should have read the ticket first... I have a slightly better understanding of wanting a failure case (off by one).  I still think the code should explain this.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+						    (iter * modify_offset)
+						    - modify_size],
+						 modify_size);
+
+					/*
+					 * Induce the modify function overhead
+					 * if configured
+					 */
+					if (opts->induce_modify_overhead)
+					{
```
sueloverso: Style: the brace (there are a few in this diff) should be at the end of the previous line.

bench/wtperf/wtperf_opt.i
```diff
@@ -122,6 +124,9 @@ DEF_OPT_AS_UINT32(idle_table_cycle, 0,
     "Default 0 which means disabled.")
 DEF_OPT_AS_BOOL(index, 0,
     "Whether to create an index on the value field.")
+DEF_OPT_AS_BOOL(induce_modify_overhead, 0,
+    "Induce the overhead of modify vector generation, because of failure in "
+    "reaching the maximum configured limits")
```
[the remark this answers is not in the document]
sueloverso: If `modify` becomes a thread workload indicator, then overhead could become a workload modifier.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+						WT_MODIFY entries[16];
+						WT_ITEM oldv, newv;
```
sueloverso: Style: Again all local variables should be declared at the top of the function. The `WT_` structures should be declared in alphabetical order going down, the variable names on the line should be alphabetical left to right.

bench/wtperf/wtperf.c
sueloverso: Style: All local variables should be declared at the top of the function. We never initialize local variables in the declaration.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+
+					/* 10% of total size of the record */
+					total_modify_size = strlen(value) / 10;
+
+					/*
+					 * maximum modification size per
+					 * modify
+					 */
+					modify_size = (total_modify_size
+							/ (size_t)nentries);
```
sueloverso: Continuation lines are indented 4 spaces.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+
+					/*
+					 * offset location difference between
+					 * modifications
+					 */
+					modify_offset = (strlen(value)
+							/ (size_t)nentries);
```
sueloverso: 4 spaces.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+						 * to find out the modify
+						 * vectors, according to the
+						 * passed input, this function
+						 * fail as the modifications
+						 * count reaches the maximum
+						 * number of modifications
+						 * that are allowed for MODIFY
+						 * operation.
+						 */
+						(void)wiredtiger_calc_modify(
```
sueloverso: All of these should be 4 spaces.

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
 					value_buf[0] = 'a';
 				if (opts->random_value)
 					randomize_value(thread, value_buf);
+
+				/*
+				 * Distribute the modifications across
+				 * whole document
+				 */
+				if (opts->distribute_modify)
+				{
```
sueloverso: If `modify` becomes a workload thread operation type a lot of these indented conditionals would go away I think.

bench/wtperf/wtperf_opt.i
```diff
@@ -108,6 +108,8 @@ DEF_OPT_AS_UINT32(database_count, 1,
     "number of WiredTiger databases to use. Each database will execute the"
     " workload using a separate home directory and complete set of worker"
     " threads")
+DEF_OPT_AS_BOOL(distribute_modify, 0,
+    "Distribute the changes of modifications across the document for update")
```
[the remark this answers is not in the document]
kommiharibabu: @sueloverso, Thanks for the review.
I thought of adding the Modify operation as a test, but I want to find out the overhead of finding out the modify vectors from wiredtiger_calc_modify function for cursor modify operation when that function fails and it leads to cursor update operation because of configured limits of the number of modifications or the total size of modification.

Current update operation has the modifications are very small (1 to 2 bytes) and the number of modifications is around 2 to 3 places, these changes are not good candidates for finding out the overhead from wiredtiger_calc_modify function, so I introduced a distributed modify for 16 modifications of around 10% of record size (limits followed by MongoDB).

I am not sure whether the modify operation is correct for the above operation?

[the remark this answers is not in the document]
sueloverso: I think what I suggest below would get you what you want and allow an easy comparison. You could have a "clean" modify workload, and then have one with the `induce_overhead` or `force_update` workload modifier that forces the failures and then compare runs of one with the others (similar to the shrink/grow workloads).

I think what you want to do is the mostly the same but differently specified and it is easier to see in the configuration as a workload.

kommiharibabu:
retest this please

bench/wtperf/wtperf.c
```diff
@@ -762,6 +762,90 @@ worker(void *arg)
...
+						&value_buf[
+						    (iter * modify_offset)
+						    - modify_size],
+						 modify_size);
+
+					/*
+					 * Induce the modify function overhead
+					 * if configured
+					 */
+					if (opts->induce_modify_overhead)
```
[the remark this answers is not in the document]
kommiharibabu: I changed the approach of the perf test as with new modify workload option, so I removed the above options.

bench/wtperf/wtperf.c
kommiharibabu: I changed the approach of the variable with the new workload option.

bench/wtperf/wtperf.c
kommiharibabu: Corrected the coding style in the new approach and also added a comment to explain for the sizes less than 160

bench/wtperf/wtperf_opt.i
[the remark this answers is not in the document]
kommiharibabu: I changed the approach as per the discussion to modify workload. Please do a complete review again as most of the code is changed.

bench/wtperf/wtperf_opt.i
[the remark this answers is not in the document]
kommiharibabu: Removed those options.

bench/wtperf/wtperf.c
kommiharibabu: Corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: Most of the indentation issues are fixed (maybe some are not proper as the coding guidelines)

bench/wtperf/wtperf.c
kommiharibabu: fixed.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

kommiharibabu:
@ddanderson, @sueloverso, The approach to find the overhead is completely changed with modify workload option, please do the review again.

kommiharibabu:
retest this please

kommiharibabu:
evergreen retry

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+					strncpy(
+					    &value_buf[(iter * modify_offset) -
+					    modify_offset],
+					    &value_buf[(iter * modify_offset) -
+					    modify_size],
+					    modify_size);
+
+				/*
+				 * Increase the number of modifications, so
+				 * that normal modify operations gets
```
ddanderson: ...`so that normal modify operations succeed.`

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				 * otherwise modifications don't spread
+				 * properly.
+				 */
+				if (modify_offset < modify_size)
+				    modify_offset = modify_size + 1;
+
+				for (iter = (size_t)nmodifications;
+					 iter > 0;
+					 iter--)
+					strncpy(
```
ddanderson: I think `strncpy` here may have overlapping values for source and destination, which is not defined.  (There may be an overlap for example when `rand_val` is a small value).  Better to use `memmove`, which does the right thing for overlapping values.

bench/wtperf/wtperf.c
```diff
@@ -1397,19 +1583,21 @@ monitor(void *arg)
 
 		(void)fprintf(fp,
```
ddanderson: When you add a field to this `fprintf`, you'll need to followup with a change to a couple of Jenkins jobs that "know" what the fields mean.  See https://jira.mongodb.org/browse/WT-4967 for the last time this happened.  It's probably best to open a separate ticket to fix the Jenkins part, and just plan to do that as soon as this ticket merges.

bench/wtperf/config.c
```diff
@@ -315,7 +343,8 @@ config_threads(WTPERF *wtperf, const char *config, size_t len)
 		if (ret != 0)
 			goto err;
 		if (workp->insert == 0 && workp->read == 0 &&
-		    workp->update == 0 && workp->truncate == 0)
+		    workp->update == 0 && workp->truncate == 0 &&
+		    workp->modify == 0)
```
sueloverso: Alphabetize this into the conditional.

bench/wtperf/config.c
```diff
@@ -328,7 +357,8 @@ config_threads(WTPERF *wtperf, const char *config, size_t len)
 		if (workp->truncate != 0 && workp->threads > 1)
 			goto err;
 		if (workp->truncate != 0 &&
-		    (workp->insert > 0 || workp->read > 0 || workp->update > 0))
+		    (workp->insert > 0 || workp->read > 0 ||
+		     workp->update > 0 || workp->modify > 0))
```
sueloverso: Alphabetize.

bench/wtperf/config.c
```diff
@@ -805,10 +835,11 @@ config_sanity(WTPERF *wtperf)
 		    i < wtperf->workload_cnt; ++i, ++workp) {
 			if (opts->readonly &&
 			    (workp->insert != 0 || workp->update != 0 ||
-			    workp->truncate != 0)) {
+			    workp->modify != 0 || workp->truncate != 0)) {
```
sueloverso: Alphabetize.

bench/wtperf/runners/modify-force-update-large-record-btree.wtperf
```diff
@@ -0,0 +1,14 @@
...
+sess_config="isolation=snapshot"
+table_config="type=file"
+key_sz=40
+value_sz=100000
+icount=500000
+report_interval=5
+random_value=true
+run_time=120
+populate_threads=1
+threads=((count=4,modify=1,ops_per_txn=1,modify_force_update=true))
```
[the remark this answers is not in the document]
sueloverso: Or, as an alternative, if `modify_delta` is just something nice to have, keep this change specific to the test case at hand and open a new ticket for it as a new feature.

bench/wtperf/runners/wtperf_run.sh
```diff
@@ -39,18 +39,18 @@ rm -f $outfile
...
-loadindex=5
+loadindex=6
 avg[$loadindex]=0
 max[$loadindex]=0
 min[$loadindex]=0
 sum[$loadindex]=0
-ops=(read insert update truncate)
-outp=("Read count:" "Insert count:" "Update count:" "Truncate count:")
+ops=(read insert update modify truncate)
+outp=("Read count:" "Insert count:" "Update count:" "Modify count:" "Truncate count:")
```
sueloverso: Alphabetize (including fixing the earlier miss of that).

bench/wtperf/track.c
```diff
@@ -132,6 +132,11 @@ sum_update_ops(WTPERF *wtperf)
 {
 	return (sum_ops(wtperf, offsetof(WTPERF_THREAD, update)));
 }
+uint64_t
+sum_modify_ops(WTPERF *wtperf)
+{
+	return (sum_ops(wtperf, offsetof(WTPERF_THREAD, modify)));
+}
```
sueloverso: Make the location of all of the functions added here alphabetized with the other similar functions.

bench/wtperf/track.c
```diff
@@ -355,4 +386,6 @@ latency_print(WTPERF *wtperf)
 	latency_print_single(wtperf, &total, "read");
 	sum_update_latency(wtperf, &total);
 	latency_print_single(wtperf, &total, "update");
+	sum_modify_latency(wtperf, &total);
+	latency_print_single(wtperf, &total, "modify");
```
sueloverso: Same here.

bench/wtperf/wtperf.h
```diff
@@ -58,11 +58,15 @@ typedef struct __truncate_queue_entry TRUNCATE_QUEUE_ENTRY;
 #define	ZSTD_EXT							\
 	EXT_PFX EXTPATH "zstd/.libs/libwiredtiger_zstd.so" EXT_SFX
 
+#define	MAX_NUM_OF_MODIFICATIONS 16
+#define	MAX_SIZE_OF_MODIFICATION 10
```
sueloverso: I'd suggest `MAX_MODIFY_NUM` and `MAX_MODIFY_PCT`. That way if there are more or defines for other operations they sort together based on the operation type. I'd add a comment about what these numbers means too.

bench/wtperf/wtperf.c
```diff
@@ -142,21 +143,20 @@ map_key_to_table(CONFIG_OPTS *opts, uint64_t k)
 
 /*
  * Figure out and extend the size of the value string, used for growing
- * updates. We know that the value to be updated is in the threads value
- * scratch buffer.
+ * updates/modifies, delta is the value to be updated according to the thread
```
sueloverso: This is two complete sentences. It should be `updates/modifies. Delta is the...`.

bench/wtperf/wtperf.c
```diff
@@ -316,6 +316,8 @@ op_name(uint8_t *op)
 		return ("truncate");
 	case WORKER_UPDATE:
 		return ("update");
+	case WORKER_MODIFY:
+		return ("modify");
```
sueloverso: Alphabetize.

bench/wtperf/wtperf.c
```diff
@@ -402,14 +404,17 @@ worker_async(void *arg)
...
 			if ((ret = asyncop->insert(asyncop)) == 0)
 				break;
 			goto op_err;
 		case WORKER_UPDATE:
 			if (opts->random_value)
-				randomize_value(thread, value_buf);
+				randomize_value(thread,
+				    value_buf,
+				    thread->workload ?
+					thread->workload->update_delta : 0);
```
sueloverso: Line continuation style. Don't break up each item.

bench/wtperf/wtperf.c
```diff
@@ -524,13 +529,19 @@ worker(void *arg)
...
 	WT_CURSOR **cursors, *cursor, *log_table_cursor, *tmp_cursor;
+	WT_ITEM newv, oldv;
+	WT_MODIFY entries[MAX_NUM_OF_MODIFICATIONS];
 	WT_SESSION *session;
 	size_t i;
-	uint32_t total_table_count;
+	size_t iter;
+	size_t modify_offset;
+	size_t modify_size;
+	size_t total_modify_size;
```
sueloverso: Combine these onto a single declaration line.

bench/wtperf/wtperf.c
```diff
@@ -524,13 +529,19 @@ worker(void *arg)
...
+	uint32_t rand_val, total_table_count;
```
sueloverso: Move this to below the 64-bit declarations for decreasing size order.

bench/wtperf/wtperf.c
```diff
@@ -524,13 +529,19 @@ worker(void *arg)
...
 	int64_t ops, ops_per_txn;
 	uint64_t log_id, next_val, usecs;
 	uint8_t *op, *op_end;
-	int measure_latency, ret, truncated;
+	int nmodifications, measure_latency, ret, truncated;
```
sueloverso: Alphabetize across the line.

bench/wtperf/wtperf.c
```diff
@@ -644,8 +656,11 @@ worker(void *arg)
 			trk = &thread->read;
 			/* FALLTHROUGH */
 		case WORKER_UPDATE:
+		case WORKER_MODIFY:
 			if (*op == WORKER_UPDATE)
 				trk = &thread->update;
+			if (*op == WORKER_MODIFY)
+				trk = &thread->modify;
```
sueloverso: Alphabetize the case to sort above `WORKER_UPDATE` and alphabetize the conditional in the same way.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
 				if ((ret = cursor->update(cursor)) == 0)
 					break;
 				goto op_err;
 			}
 
+			/*
+			 * Reads can fail with WT_NOTFOUND: we may be searching
+			 * in a random range, or an insert thread might have
+			 * updated the last record in the table but not yet
+			 * finished the actual insert.  Count failed search in
```
sueloverso: "Count a failed search"...

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				if (opts->random_value)
+					randomize_value(thread,
+					    value_buf,
+					    workload->modify_delta);
+
+				/*
+				 * Distribute the modifications across whole
+				 * document.
+				 *
+				 * The distribution limits are same as MongoDB
```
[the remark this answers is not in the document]
sueloverso: I agree with this and was going to make the same suggestion. I also suggest removing the reference to MongoDB. I suggest maybe having the comment read something like this:
```
/*
 * Distribute the modifications across the whole
 * document. We randomly choose up to the maximum
 * number of modifications and modify up to the
 * maximum percent of the record size.
 */
```

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				 * that are imposed for an update operation to
+				 * be considered as modify or not?
+				 *
+				 * up to 10% of total modification size
+				 * 16 number of modifications per record
+				 *
+				 * NOTE: for records of size less than 160,
+				 * there is no distributed modifications.
+				 */
+
```
sueloverso: Remove blank line.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+					value_buf[0] = 'b';
+				else
+					value_buf[0] = 'a';
+				if (opts->random_value)
+					randomize_value(thread,
+					    value_buf,
+					    workload->modify_delta);
+
+				/*
+				 * Distribute the modifications across whole
```
sueloverso: "across the whole"

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+
+				nmodifications = MAX_NUM_OF_MODIFICATIONS;
+
+				/* 10% of total size of the record */
+				total_modify_size = strlen(value) /
+				    MAX_SIZE_OF_MODIFICATION;
+
+				/*
+				 * Randomize the maximum modification size
+				 * and offset per modify. With randomization
```
sueloverso: I think that only the first sentence of this comment is necessary.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+
+				/*
+				 * Offset location difference between
+				 * modifications
+				 */
+				modify_offset =
+				    strlen(value) / (size_t)nmodifications;
+				if (modify_offset != 0)
+					modify_offset =
+					    (size_t)rand_val % modify_offset;
```
sueloverso: I find these a bit difficult to follow. Both of these could be simplified to something like:
```
if (total_modify_size != 0)
        modify_size = rand_val %
            (total_modify_size / MAX_MODIFY_PCT);
else
        modify_size = 0;
/* And similar for modify_offset */
```

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+
+				/*
+				 * Make sure the offset is more than size,
+				 * otherwise modifications don't spread
+				 * properly.
+				 */
+				if (modify_offset < modify_size)
+				    modify_offset = modify_size + 1;
```
sueloverso: This should be incorporated into the simplified conditional also. And style: the conditional line should be tabbed in 8.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+
+				for (iter = (size_t)nmodifications;
+					 iter > 0;
+					 iter--)
```
sueloverso: Style for the lines above.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				 * succeeded.
+				 */
+				if (!workload->modify_force_update)
+					nmodifications++;
+
+				oldv.data = value;
+				oldv.size = strlen(value);
+
+				newv.data = value_buf;
+				newv.size = strlen(value_buf);
```
sueloverso: These could be combined to `oldv.data = newv.data = value_buf;`, and similar for size.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+
+				/*
+				 * Pass the old and new data to find out the
+				 * modify vectors, according to the passed
+				 * input, this function may fail when the
```
sueloverso: "...input. This function..."

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				 * modifications count reaches the maximum
+				 * number of modifications that are allowed
+				 * for MODIFY operation.
```
sueloverso: "for the modify operation."

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				 */
+				ret = wiredtiger_calc_modify(session,
+				    &oldv,
+				    &newv,
+				    total_modify_size,
+				    entries,
+				    &nmodifications);
```
sueloverso: Line continuation style.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+
+				if (ret == WT_NOTFOUND ||
+					workload->modify_force_update) {
+					cursor->set_value(cursor, value_buf);
```
sueloverso: Line continuation style. Spaced in 4.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+						cursor->update(cursor)) == 0)
+						break;
+				}
+				else {
+					if ((ret =
+						cursor->modify(cursor,
+						entries,
+						nmodifications)) == 0)
+						break;
+				}
```
sueloverso: Lots of line continuation style changes needed above.

bench/wtperf/wtperf.c
```diff
@@ -970,7 +1134,8 @@ run_mix_schedule(WTPERF *wtperf, WORKLOAD *workp)
 
 	if (workp->truncate != 0) {
 		if (workp->insert != 0 ||
-		    workp->read != 0 || workp->update != 0) {
+		    workp->read != 0 || workp->update != 0 ||
+			workp->modify != 0) {
```
sueloverso: Add this alphabetically, and fix the line continuation spacing.

bench/wtperf/wtperf.c
```diff
@@ -979,28 +1144,36 @@ run_mix_schedule(WTPERF *wtperf, WORKLOAD *workp)
 		return (0);
 	}
 
-	/* Confirm reads, inserts and updates cannot all be zero. */
-	if (workp->insert == 0 && workp->read == 0 && workp->update == 0) {
+	/* Confirm reads, inserts, modifies and updates cannot all be zero. */
+	if (workp->insert == 0 && workp->read == 0 && workp->update == 0 &&
+	    workp->modify == 0) {
```
sueloverso: Alphabetical (in pretty much all the places here and below).

bench/wtperf/wtperf.h
```diff
@@ -58,11 +58,15 @@ typedef struct __truncate_queue_entry TRUNCATE_QUEUE_ENTRY;
...
+
 typedef struct {
 	int64_t threads;		/* Thread count */
 	int64_t insert;			/* Insert ratio */
 	int64_t read;			/* Read ratio */
 	int64_t update;			/* Update ratio */
+	int64_t modify;			/* Modify ratio */
```
sueloverso: Alphabetize its location.

bench/wtperf/wtperf.h
```diff
@@ -73,12 +77,15 @@ typedef struct {
...
 	int64_t update_delta;		/* Value size change on update */
+	int64_t modify_delta;		/* Value size change on modify */
+	bool	modify_force_update; /* Do force update instead of modify */
 
 #define	WORKER_INSERT		1	/* Insert */
 #define	WORKER_INSERT_RMW	2	/* Insert with read-modify-write */
 #define	WORKER_READ		3	/* Read */
 #define	WORKER_TRUNCATE		4	/* Truncate */
 #define	WORKER_UPDATE		5	/* Update */
+#define	WORKER_MODIFY		6   /* Modify */
```
sueloverso: Alphabetize. And fix the spacing for the comment.

bench/wtperf/runners/modify-force-update-large-record-btree.wtperf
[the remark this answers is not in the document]
kommiharibabu: @ddanderson, @sueloverso, I will raise a new ticket to handle the modify_delta perf scenario.

bench/wtperf/wtperf.c
kommiharibabu: Fixed.

bench/wtperf/wtperf.c
```diff
@@ -545,6 +556,7 @@ worker(void *arg)
 	ops_per_txn = workload->ops_per_txn;
 	session = NULL;
 	trk = NULL;
+	nmodifications = MAX_NUM_OF_MODIFICATIONS;
```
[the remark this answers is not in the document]
kommiharibabu: Removed.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+			 * Reads can fail with WT_NOTFOUND: we may be searching
+			 * in a random range, or an insert thread might have
+			 * updated the last record in the table but not yet
+			 * finished the actual insert.  Count failed search in
+			 * a random range as a "read".
+			 */
+			if (ret == WT_NOTFOUND)
+				break;
+			goto op_err;
+		case WORKER_MODIFY:
```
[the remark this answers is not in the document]
kommiharibabu: Code is refactored to reduce the duplicate code.

bench/wtperf/wtperf.c
```diff
@@ -755,19 +770,168 @@ worker(void *arg)
...
+				 * up to 10% of total modification size
+				 * 16 number of modifications per record
+				 *
+				 * NOTE: for records of size less than 160,
+				 * there is no distributed modifications.
+				 */
+
+				nmodifications = MAX_NUM_OF_MODIFICATIONS;
+
+				/* 10% of total size of the record */
```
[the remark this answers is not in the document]
kommiharibabu: changed as per the above.

bench/wtperf/wtperf.c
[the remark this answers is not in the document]
kommiharibabu: Changed as per the above.

bench/wtperf/wtperf.c
kommiharibabu: changed as per the above.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: I will create another Jira ticket similar to WT-4957 after it is merged.

bench/wtperf/config.c
```diff
@@ -304,6 +304,34 @@ config_threads(WTPERF *wtperf, const char *config, size_t len)
 				}
 				continue;
 			}
+			if (STRING_MATCH("modify", k.str, k.len)) {
```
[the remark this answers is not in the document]
kommiharibabu: Corrected.

bench/wtperf/config.c
kommiharibabu: corrected.

bench/wtperf/config.c
kommiharibabu: corrected.

bench/wtperf/config.c
kommiharibabu: corrected.

bench/wtperf/runners/wtperf_run.sh
kommiharibabu: corrected.

bench/wtperf/track.c
kommiharibabu: changed according to the other functions. Alphabetized.

bench/wtperf/track.c
kommiharibabu: corrected.

bench/wtperf/wtperf.h
kommiharibabu: Changed as per your suggestion.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: changed accordingly.

bench/wtperf/wtperf.c
kommiharibabu: Changed along the similar lines, not exactly as above.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: corrected.

bench/wtperf/wtperf.h
kommiharibabu: corrected.

bench/wtperf/wtperf.h
kommiharibabu: corrected.

bench/wtperf/wtperf.c
kommiharibabu: Actually **oldv** is pointing to **value** and the **newv** is pointing to **value_buf**. In the current test, both the sizes are the same, but in case if there is any difference because of modify_delta, we cannot assign the same in the single line.

kommiharibabu:
@ddanderson, @sueloverso, I fixed all the review comments and also changed most of the code to Alphabetize where modify workload option is added.

bench/wtperf/config.c
```diff
@@ -765,11 +794,12 @@ config_sanity(WTPERF *wtperf)
 
     if (wtperf->workload != NULL)
         for (i = 0, workp = wtperf->workload; i < wtperf->workload_cnt; ++i, ++workp) {
-            if (opts->readonly &&
-              (workp->insert != 0 || workp->update != 0 || workp->truncate != 0)) {
+            if (opts->readonly && (workp->insert != 0 || workp->modify != 0 ||
+                                    workp->truncate != 0 || workp->update != 0)) {
```
sueloverso: Whitespace. This should not be spaced all the way over there.

bench/wtperf/config.c
```diff
@@ -785,6 +815,13 @@ config_sanity(WTPERF *wtperf)
                   workp->table_index, opts->table_count);
                 return (EINVAL);
             }
+            if (workp->modify != 0 && workp->ops_per_txn == 0) {
+                fprintf(stderr,
+                  "Modify operations must be executed with "
+                  "explicit transaction, specify "
+                  "ops_per_txn.");
```
sueloverso: With the new format these lines can probably be joined.

bench/wtperf/wtperf.h
```diff
@@ -143,6 +150,7 @@ struct __wtperf {         /* Per-database structure */
     uint64_t read_ops;     /* read operations */
     uint64_t truncate_ops; /* truncate operations */
     uint64_t update_ops;   /* update operations */
+    uint64_t modify_ops;   /* modify operations */
```
sueloverso: Alphabetize this to above `read_ops`.

bench/wtperf/wtperf.h
```diff
@@ -245,6 +253,7 @@ struct __wtperf_thread {    /* Per-thread structure */
     TRACK read;           /* Read operations */
     TRACK scan;           /* Scan operations */
     TRACK update;         /* Update operations */
+    TRACK modify;         /* Modify operations */
```
sueloverso: Alphabetize this.

bench/wtperf/wtperf.h
```diff
@@ -263,6 +272,7 @@ void latency_insert(WTPERF *, uint32_t *, uint32_t *, uint32_t *);
 void latency_print(WTPERF *);
 void latency_read(WTPERF *, uint32_t *, uint32_t *, uint32_t *);
 void latency_update(WTPERF *, uint32_t *, uint32_t *, uint32_t *);
+void latency_modify(WTPERF *, uint32_t *, uint32_t *, uint32_t *);
```
sueloverso: Alphabetize.

bench/wtperf/wtperf.h
```diff
@@ -277,6 +287,7 @@ uint64_t sum_pop_ops(WTPERF *);
 uint64_t sum_read_ops(WTPERF *);
 uint64_t sum_truncate_ops(WTPERF *);
 uint64_t sum_update_ops(WTPERF *);
+uint64_t sum_modify_ops(WTPERF *);
```
sueloverso: Alphabetize.

bench/wtperf/wtperf.c
```diff
@@ -1307,23 +1416,27 @@ monitor(void *arg)
...
               ",\"update\":{\"ops per sec\":%" PRIu64 ",\"average latency\":%" PRIu32
               ",\"min latency\":%" PRIu32 ",\"max latency\":%" PRIu32 "}",
               cur_updates, update_avg, update_min, update_max);
             fprintf(jfp, "}}\n");
         }
 
-        if (latency_max != 0 &&
-          (read_max > latency_max || insert_max > latency_max || update_max > latency_max)) {
+        if (latency_max != 0 && (insert_max > latency_max || modify_max > latency_max ||
+                                  read_max > latency_max || update_max > latency_max)) {
```
ddanderson: Whitespace on the continuation indent??  Sue flagged something like this earlier, but I know the style and indentation rules have recently changed.  If this file passes the style checker, then I'm good.

bench/wtperf/runners/modify-large-record-btree.wtperf
```diff
@@ -0,0 +1,13 @@
+# wtperf options file: btree with updatesy. The goal here is to have a workload
```
ddanderson: Typo.  should be `btree with updates.`

bench/wtperf/wtperf.c
```diff
@@ -707,24 +718,103 @@ worker(void *arg)
...
+                 * Distribute the modifications across the whole document. We randomly choose up to
+                 * the maximum number of modifications and modify up to the maximum percent of the
+                 * record size.
+                 */
+                nmodify = MAX_MODIFY_NUM;
+
+                /*
+                 * The maximum that will be modified is a fixed percentage of the total record size.
+                 */
+                total_modify_size = strlen(value) / MAX_MODIFY_PCT;
```
ddanderson: I think we should pre-calculate strlen(value) into a local variable. It is called 4 times (or 5 times if you include strlen(value_buf) which should be identical), which may be significant if the value buffer is configured to be very large.

bench/wtperf/config.c
kommiharibabu: This is because of new clang format, continuation indentation is applying after the nested brackets. I even tried to bring the entire second block down, still, the format is returning to the same way.

bench/wtperf/wtperf.c
kommiharibabu: This is because of new clang format, continuation indentation is applying after the nested brackets. I even tried to bring the entire second block down, still, the format is returning to the same way.

bench/wtperf/runners/modify-large-record-btree.wtperf
kommiharibabu: Fixed

bench/wtperf/wtperf.c
kommiharibabu: Fixed with a local variable for strlen(value), I left strlen(value_buf) as it is, because it is possible that value and value_buf sizes may differ.

bench/wtperf/runners/modify-large-record-btree.wtperf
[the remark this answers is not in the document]
kommiharibabu: Fixed.

bench/wtperf/config.c
kommiharibabu: corrected.

bench/wtperf/wtperf.h
kommiharibabu: Fixed.

bench/wtperf/wtperf.h
kommiharibabu: Fixed.

bench/wtperf/wtperf.h
kommiharibabu: Fixed.

bench/wtperf/wtperf.h
kommiharibabu: Fixed.

bench/wtperf/wtperf.c
kommiharibabu: Created a new Jira ticket to correct Jenkins fail jobs 
https://jira.mongodb.org/browse/WT-5062

PR #4805, opened 2019-08-07
WT-4678 Write the test workload for relevant history in cache

tetsuo-cpp:
This PR is adding a test that exercises the new lookaside behaviour and makes some assertions on how the implementation should behave from a memory usage + perf perspective.

### DIFF
--- a/test/suite/test_las06.py
+++ b/test/suite/test_las06.py
@@ -1,0 +1,83 @@
+import wiredtiger, wttest
+from wiredtiger import stat
+
+def timestamp_str(t):
+    return '%x' % t
+
+# test_las06.py
+# Verify that triggering lookaside usage does not cause a spike in memory usage
+# to form an update chain from the lookaside contents.
+#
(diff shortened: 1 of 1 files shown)


test/suite/test_las06.py
```diff
@@ -0,0 +1,100 @@
...
+        # entry for a given value to get read in. Now that we're only reading
+        # the lookaside entry that we need, we expect to see comparatively fewer
+        # reads from lookaside.
+        #
+        # TODO: Set the upper bound to something smaller once the project work
+        # is done. At the time of writing, I get ~1500 lookaside reads for this
+        # workload.
+        las_usage = self.get_lookaside_usage()
+        self.assertNotEqual(las_usage, 0)
+        self.assertLess(las_usage, 2000)
```
tetsuo-cpp: One of the things we discussed is checking for stalling due to excessive lookaside reads. My idea is to use this statistic as an indicator for that behaviour.

tetsuo-cpp:
retest this please

PR #4806, opened 2019-08-07
WT-5050 Don't allow checkpoints to evict without a snapshot

agorrod:
Checkpoint gives up it's snapshot prior to checkpointing the metadata.
Once the snapshot is given up, it's not OK for checkpoint to try eviction prior to doing reconciliation.


agorrod:
autobuild: retest this please

PR #4807, opened 2019-08-07
WT-4676 Write a search function to locate an update in the lookaside file.

sulabhM:
Also, match the record found with the record in-memory to verify that
the search function is behaving correctly.


sulabhM:
Note to reviewers:
The key functionality being added here is the ```__wt_find_lookaside_upd()``` function. At the moment to test this functionality, I am allocating an update out of ```__wt_find_lookaside_upd()``` and comparing to the one found in the in-memory update chain. I am also letting the update created by ```__wt_find_lookaside_upd()``` to be copied to the cursor value item.

The proposal is to keep the lookaside cursor positioned on the relevant record and then copy directly to the user cursor value. This will avoid any changes to WT_UPDATE and unnecessary copies of the data. I am yet to create tickets to do this.

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+	 * updates till we locate the one we want.
+	 */
+	WT_PUBLISH(cache->las_reader, true);
+	__wt_readlock(session, &cache->las_sweepwalk_lock);
+	WT_PUBLISH(cache->las_reader, false);
+	locked = true;
+
+	for (ret = __wt_las_cursor_position(cursor, las_pageid);
+	    ret == 0;
+	    ret = cursor->next(cursor)) {
```
tetsuo-cpp: Don't you need something at the beginning like:

```
WT_ERR(ret);
```

If your `next` call or the initial `__wt_las_cursor_position` fails, don't you want to get out of here?

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+
+		/*
+		 * Keys are sorted in an order, skip the ones before the desired
+		 * key, and bail out if we have crossed over the desired key and
+		 * not found the record we are looking for.
+		 */
+		WT_ERR(__wt_compare(session,
+		    S2BT(session)->collator, &las_key, key, &cmp));
+		if (cmp < 0)
+			continue;
```
tetsuo-cpp: Is this fine? The terminating cond for this loop is `ret == 0`.

It seems that `ret` could be zero up until here and then the continue will cause you to terminate even though you haven't found the right key.

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+
+		/*
+		 * It is safe to assume that the updates are sorted newest to
+		 * the oldest. We can quit searching after finding the newest
+		 * visible record.
+		 */
+		upd_visible =
+		    __wt_txn_visible(session, las_txnid, las_timestamp);
+		if (!upd_visible)
+			continue;
```
tetsuo-cpp: Same here. If `ret` is 0 here then we're going to break out of the loop which sounds like not what you're intending.

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+	/*
+	 * The lookaside records are in key and update order, that is, there
+	 * will be a set of in-order updates for a key, then another set of
+	 * in-order updates for a subsequent key. We scan through all of the
+	 * updates till we locate the one we want.
+	 */
+	WT_PUBLISH(cache->las_reader, true);
+	__wt_readlock(session, &cache->las_sweepwalk_lock);
+	WT_PUBLISH(cache->las_reader, false);
+	locked = true;
```
keithbostic: This looks wrong to me: `WT_PUBLISH` ensures that all previous writes have completed before the current write, and it's unclear to me what previous writes are being flushed.

keithbostic: The `locked` variable isn't really needed (it never has a value other than `true`), and I suspect one of the static analysis tools is likely to notice that.

src/btree/bt_read.c
keithbostic: +1, `__wt_find_lookaside_upd()` can fail for a variety of reasons, as can `cursor->next` and we're not handling those errors.

Maybe `WT_ERR_NOTFOUND_OK(ret);` at the end of the loop? That has a problem with the places where we do
```
ret = WT_NOTFOUND;
break;
```
but we don't need to set `ret` at all there, do we? Won't
```
/* Couldn't find a record */
if (upd == NULL && ret == 0)
        ret = WT_NOTFOUND;
```
do what we need?
```

File: src/btree/bt_read.c [hunk 5]
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+	int cmp;
+	bool locked, upd_visible;
+
+	cache = S2C(session)->cache;
+	cursor = NULL;
+	session_flags = 0;		/* [-Werror=maybe-uninitialized] */
+	ref = cbt->ref;
+	las_pageid = ref->page_las->las_pageid;
+	key = &cbt->iface.key;
+	*updp = upd = NULL;
```
keithbostic: Function side-effects should happen first of anything in the function, so life `*updp = NULL` to the top.

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+	WT_UPDATE *upd;
+	wt_timestamp_t durable_timestamp, las_timestamp;
+	size_t incr;
+	uint64_t las_counter, las_pageid, las_txnid;
+	uint32_t las_id, session_flags;
+	uint8_t prepare_state, upd_type;
+	int cmp;
+	bool locked, upd_visible;
+
+	cache = S2C(session)->cache;
```
keithbostic: It helps people figure out if there's any missing initializations if local variable initialization is roughly in the same order as the declarations. Not a rule, but a guideline. :)

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
...
+ */
+int
+__wt_find_lookaside_upd(WT_SESSION_IMPL *session,
+    WT_CURSOR_BTREE *cbt, WT_UPDATE **updp, bool ignore_birthmark)
+{
+	WT_CACHE *cache;
+	WT_CURSOR *cursor;
+	WT_DECL_RET;
+	WT_ITEM *key;
+	WT_ITEM las_key, las_value;
```
keithbostic: No reason to have two lines of `WT_ITEM`s?

src/btree/bt_read.c
keithbostic: @tetsuo-cpp, not following:`ret` must be 0 at the continue, the continue means we move to the next record?

src/btree/bt_ret.c
```diff
@@ -155,9 +155,16 @@ __wt_value_return_upd(WT_SESSION_IMPL *session,
...
+		if (upd->ext) {
+			/* Copy an external update, and delete after using it */
+			ret = __wt_buf_set(session,
+			    &cursor->value, upd->data, upd->size);
+			__wt_free_update_list(session, upd);
+		} else {
+			cursor->value.data = upd->data;
+			cursor->value.size = upd->size;
+		}
+		return (ret);
```
keithbostic: In general, I think we're pretty consistent about immediate return on error. I'd go with:
```
if (upd->ext) {
        /* Copy an external update, and delete after using it */
        WT_RET(__wt_buf_set(session,
            &cursor->value, upd->data, upd->size));
        __wt_free_update_list(session, upd);
} else {
        cursor->value.data = upd->data;
        cursor->value.size = upd->size;
}       
return (0);  
```

src/include/btmem.h
```diff
@@ -1140,7 +1143,7 @@ struct __wt_update {
  * WT_UPDATE_SIZE is the expected structure size excluding the payload data --
  * we verify the build to ensure the compiler hasn't inserted padding.
  */
-#define	WT_UPDATE_SIZE	38
+#define	WT_UPDATE_SIZE	39
```
keithbostic: I think we might get into trouble here because a boolean isn't a fixed size, and so it might be padded? Should we make that a `uint8_t`?

src/include/txn.i
```diff
@@ -896,6 +898,34 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_UPDATE *upd, WT_UPDATE **updp)
...
+		/* It is okay to locate an older update in lookaside */
+		WT_ASSERT(session, upd->start_ts >= tmp_upd->start_ts);
+
+		if (upd->start_ts == tmp_upd->start_ts) {
+			/*
+			 * If we located the correct record, the contents should
+			 * be the same and we can use the record from lookaside.
+			 */
+			WT_ASSERT(session, (upd->size == tmp_upd->size &&
+			    memcmp(upd->data, tmp_upd->data, upd->size) == 0));
```
keithbostic: There's no need for the outer set of parenthesis.

src/include/txn.i
```diff
@@ -896,6 +898,34 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_UPDATE *upd, WT_UPDATE **updp)
...
+	 * in-memory record. The following conditions will be refined in the
+	 * future once we make the change to stop initializing a page using
+	 * lookaside.
+	 */
+	if (upd != NULL && upd->txnid != 0 && cbt->ref->page_las != NULL &&
+	    __wt_find_lookaside_upd(session, cbt, &tmp_upd, true) == 0) {
+		/* It is okay to locate an older update in lookaside */
+		WT_ASSERT(session, upd->start_ts >= tmp_upd->start_ts);
+
+		if (upd->start_ts == tmp_upd->start_ts) {
```
keithbostic: Why not pass the `start_ts` into `__wt_find_lookaside_upd()`? It felt a little odd to check lots of things in that function, but make that test in the caller instead?

src/btree/bt_read.c
```diff
@@ -68,6 +68,129 @@ __row_instantiate(WT_SESSION_IMPL *session,
 	return (0);
 }
 
+/*
+ * __wt_find_lookaside_upd --
```
keithbostic: Why put `__wt_find_lookaside_upd()` in `bt_read.c`? Happy to believe it should be there, but I would have guessed `cache_las.c`?

src/include/txn.i
```diff
@@ -896,6 +898,34 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_UPDATE *upd, WT_UPDATE **updp)
...
+			WT_ASSERT(session, (upd->size == tmp_upd->size &&
+			    memcmp(upd->data, tmp_upd->data, upd->size) == 0));
+			upd = tmp_upd;
+			/*
+			 * Mark this update as external and to be discarded
+			 * when not needed.
+			 */
+			upd->ext = true;
+		} else
+			__wt_free_update_list(session, tmp_upd);
```
keithbostic: If `start_ts is passed into `__wt_find_lookaside_upd()`, then `upd.ext` can be set there, and there's no need to free the update list (in short, this work gets simpler)?

src/btree/bt_read.c
tetsuo-cpp: Oops. Nevermind me. Not sure what I was thinking.

src/btree/bt_read.c
sulabhM: I was not sure we need to publish here myself. I based this code on ```__las_page_instantiate()```, are we publishing there for a reason (maybe related to generation number updates in ```__las_page_instantiate_verbose()```) ?

Can you please have a look, I will remove the publish then, it does seem they aren't needed.

src/btree/bt_read.c
sulabhM: I based this function on ```__las_page_instantiate()``` which is in bt_read.c. I have moved ```__wt_find_lookaside_upd()``` into cache_las.c

src/btree/bt_read.c
sulabhM: good catch. Fixed

src/include/btmem.h
sulabhM: I did think about bool not having a standard defined size, thought this being a throw away code it should be fine. I decided to make in uint8_t

src/include/txn.i
sulabhM: I was trying to avoid passing anything into ```__wt_find_lookaside_upd()``` from an in-memory update because the next ticket will do away with initializing updates from lookaside into memory. On second thought, this simplifies the code a lot. Also these asserts and conditions will go away when I do the next ticket anyways.

I have fixed this as you suggested, looks much cleaner.

sulabhM:
@keithbostic and @tetsuo-cpp , I have addressed review comments. Please have another look.

src/btree/bt_read.c
keithbostic: @sueloverso, in https://github.com/wiredtiger/wiredtiger/commit/f29d2b9a20662d60baadda142c2283d7afc18b33 you and @michaelcahill wrapped a lock in two `WT_PUBLISH` statements in `__las_page_instantiate()`:
```
WT_PUBLISH(cache->las_reader, true);
__wt_readlock(session, &cache->las_sweepwalk_lock);
WT_PUBLISH(cache->las_reader, false);
```
I don't see why either of the `WT_PUBLISH` statements would be necessary, and I can't find any discussion. Can you reconstruct what's going on here?

(Sorry, I know this change is old, but @sulabhM copied in his new code, and I noticed it there.)

src/cache/cache_las.c
```diff
@@ -1396,3 +1396,132 @@ err:		__wt_buf_free(session, sweep_key);
...
+		if (prepare_state == WT_PREPARE_INPROGRESS &&
+		    !F_ISSET(&session->txn, WT_TXN_IGNORE_PREPARE))
+			break;
+
+		if (upd_type == WT_UPDATE_BIRTHMARK && ignore_birthmark)
+			break;
+
+		/* It is okay to locate an older update in lookaside,
+		 * but we proceed only if we find a matching record. */
+		WT_ASSERT(session, start_ts >= las_timestamp);
```
tetsuo-cpp: I'm sure you're right since the code works but I don't understand it. Why can't we see a lower timestamp? Is it not ordered from oldest to newest? I was expecting something like:
```
/* Keep going. */
if (starts_ts > las_timestamp)
    continue;
/* Too far. */
else if (start_ts < las_timestamp)
    break;
/* Got a match. */
```

sulabhM: The code is doing somewhat similar. The updates are ordered from newest to oldest. We traverse the list stopping at the first record that meets our visibility criterion. Timestamp is one of the criterion,  ```__wt_txn_visible()``` tells us if a certain update is visible to this transaction or not.

Since at the moment I am walking through lookaside if there has been a lookaside eviction ever for this page, there is a possibility that the lookaside contents are obsolete or there has been newer content in-memory. So we might end up finding a record that is older than being asked for, but it can never be newer. Hence the assert.

Does that make sense ?

tetsuo-cpp: In that case shouldn't the call to `__wt_txn_visible` be using `start_ts` as the timestamp argument?

sulabhM: It does takes ```las_timestamp``` as argument, which is the timestamp of the current lookaside record being tested for visibility. It compares it with ```txn->read_timestamp```. If the transaction is reading later than the record's start timestamp (```las_timestamp```), the record is visible otherwise it is not.

src/btree/bt_read.c
sueloverso: It has been a long time. My best guess is that the intent is that the value of `cache->las_reader` needs to be immediately visible to other threads. That field will cause a long LAS sweep to break out of its sweep-work if it knows a reader is waiting.

keithbostic: OK, here's my argument:

The first `WT_PUBLISH` would flush any writes before setting `las_reader`. I can't identify any writes before the set of `las_reader` that need to be flushed.

The second `WT_PUBLISH` would flush any writes after the `wt_readlock()` call before setting `las_reader`. There aren't any, so again, it's not doing anything useful.

@sulabhM, unless you or @sueloverso see something I'm missing, I think we should change both instances to:
```
cache->las_reader = true;
__wt_readlock(session, &cache->las_sweepwalk_lock);
cache->las_reader = false;
```

sueloverso: @keithbostic and I discussed and agreed to this.

sulabhM:
autobuild: retest this please

PR #4813, opened 2019-08-08
WT-5040 Add dumping of cursor's page on a mismatch.

sueloverso:
@keithbostic for any suggestions of other/additional places to call this new function. Not yet ready for merge.

### DIFF
--- a/src/btree/bt_debug.c
+++ b/src/btree/bt_debug.c
@@ -658,30 +658,6 @@ __wt_debug_tree_all(
 }
 
 /*
- * __wt_debug_tree_cursor --
- *	Dump the in-memory information for a tree, including leaf pages
- *	starting from a user's cursor and the root of the tree.
- */
-int
-__wt_debug_tree_cursor(WT_CURSOR *cursor, const char *ofile)
-    WT_GCC_FUNC_ATTRIBUTE((visibility("default")))
-{
-	WT_CURSOR_BTREE *cbt;
-	WT_DECL_RET;
-	WT_SESSION_IMPL *session;
-
-	session = (WT_SESSION_IMPL *)cursor->session;
-	cbt = (WT_CURSOR_BTREE *)cursor;
-
-	/*
-	 * Dump only the cursor referenced page.
-	 */
-	WT_WITH_BTREE(session, cbt->btree, ret = __wt_debug_page(
-	    session, NULL, cbt->ref, ofile));
-	return (ret);
-}
-
-/*
  * __wt_debug_tree --
  *	Dump the in-memory information for a tree, not including leaf pages.
  */
@@ -742,6 +718,7 @@ __wt_debug_page(
  */
 int
 __wt_debug_cursor_page(void *cursor_arg, const char *ofile)
+    WT_GCC_FUNC_ATTRIBUTE((visibility("default")))
 {
 	WT_CURSOR *cursor;
 	WT_CURSOR_BTREE *cbt;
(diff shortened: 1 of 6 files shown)


test/format/format.h
```diff
@@ -73,6 +73,7 @@ typedef struct {
 	char *home_rand;			/* RNG log file path */
 	char *home_salvage_copy;		/* Salvage copy command */
 	char *home_stats;			/* Statistics file path */
+	char *home_treedump;			/* Treedump filename */
```
keithbostic: Possible name change to reflect we're not dumping a tree, we're dumping a page?

src/btree/bt_debug.c
```diff
@@ -657,6 +657,30 @@ __wt_debug_tree_all(
...
+	cbt = (WT_CURSOR_BTREE *)cursor;
+
+	/*
+	 * Dump only the cursor referenced page.
+	 */
+	WT_WITH_BTREE(session, cbt->btree, ret = __wt_debug_page(
+	    session, NULL, cbt->ref, ofile));
+	return (ret);
+}
+
```
keithbostic: Isn't this the same as `wt_debug_cursor_page()`?

test/format/snap.c
```diff
@@ -178,6 +178,11 @@ snap_verify(WT_CURSOR *cursor, TINFO *tinfo, SNAP_OPS *snap)
 	}
 
 	/* Things went pear-shaped. */
+#ifdef HAVE_DIAGNOSTIC
+	fprintf(stderr,
+	    "snapshot-isolation error: Dumping tree to %s\n", g.home_treedump);
+	(void)__wt_debug_tree_cursor(cursor, g.home_treedump);
+#endif
```
keithbostic: This is going to hit some of the "function return value unchecked" complaints, should have `testutil_check()`.

test/format/snap.c
```diff
@@ -181,7 +182,9 @@ snap_verify(WT_CURSOR *cursor, TINFO *tinfo, SNAP_OPS *snap)
 #ifdef HAVE_DIAGNOSTIC
 	fprintf(stderr,
 	    "snapshot-isolation error: Dumping tree to %s\n", g.home_treedump);
-	(void)__wt_debug_tree_cursor(cursor, g.home_treedump);
+	t_ret = __wt_debug_tree_cursor(cursor, g.home_treedump);
+	if (ret == 0)
+		ret = t_ret;
```
keithbostic: I'd go with `testutil_check()`, else `WT_TRET`.

src/btree/bt_debug.c
sueloverso: Yes! After I made the change, I didn't realize the same thing was just a bit further down. I'll fix.

test/format/format.h
sueloverso: I like that.

test/format/snap.c
sueloverso: Fixed.

test/format/util.c
```diff
@@ -348,6 +348,12 @@ path_setup(const char *home)
 	testutil_check(__wt_snprintf(
 	    g.home_stats, len, "%s/%s", g.home, "stats"));
 
+	/* Tree dump file. */
+	len = strlen(g.home) + strlen("pagedump") + 2;
+	g.home_pagedump = dmalloc(len);
+	testutil_check(__wt_snprintf(
+	    g.home_pagedump, len, "%s/%s", g.home, "pagedump"));
+
```
keithbostic: Not sure the comment makes sense any more.

sueloverso: Fixed

PR #4816, opened 2019-08-12
WT-4899 Fix bugs that could allow more than one birthmark in an update chain (#4763) (v4.0 backport)

lukech:
Specifically there were cases where writing a page to lookaside multiple times would result in an update chain with multiple birthmarks, which isn't right. A birthmark refers to the on-page value, and there should only ever be one update in a chain that refers to it.

There were two issues:
* Sometimes birthmarks weren't being resolved during reconciliation. This would mean that they would be re-written in subsequent lookaside evictions. This would only be a problem if reconciliation had chosen a different record to put in the data file the second time, and it should not have been doing that.
* Sometimes obsolete updates were being written to lookaside, and we never resolve values for obsolete updates. A birthmark could be written, but it would never have been visible.

(cherry picked from commit a9a9857e32181befc4cba4541325dc36008749d6)

### DIFF
--- a/dist/s_void
+++ b/dist/s_void
@@ -57,6 +57,7 @@ func_ok()
 	    -e '/int __wt_block_write_size$/d' \
 	    -e '/int __wt_buf_catfmt$/d' \
 	    -e '/int __wt_buf_fmt$/d' \
+	    -e '/int __wt_count_birthmarks$/d' \
 	    -e '/int __wt_curjoin_joined$/d' \
 	    -e '/int __wt_cursor_noop$/d' \
 	    -e '/int __wt_epoch$/d' \
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -27,7 +27,8 @@ __col_instantiate(WT_SESSION_IMPL *session,
 	 * Just free the memory: it hasn't been accounted for on the page yet.
 	 */
 	if (updlist->next != NULL &&
-	    (upd = __wt_update_obsolete_check(session, page, updlist)) != NULL)
+	    (upd = __wt_update_obsolete_check(
+	    session, page, updlist, false)) != NULL)
 		__wt_free_update_list(session, upd);
 
 	/* Search the page and add updates. */
@@ -56,7 +57,8 @@ __row_instantiate(WT_SESSION_IMPL *session,
 	 * Just free the memory: it hasn't been accounted for on the page yet.
 	 */
 	if (updlist->next != NULL &&
-	    (upd = __wt_update_obsolete_check(session, page, updlist)) != NULL)
+	    (upd = __wt_update_obsolete_check(
+	    session, page, updlist, false)) != NULL)
 		__wt_free_update_list(session, upd);
 
 	/* Search the page and add updates. */
@@ -238,7 +240,8 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 	WT_ERR_NOTFOUND_OK(ret);
 
 	/* Insert the last set of updates, if any. */
-	if (first_upd != NULL)
+	if (first_upd != NULL) {
+		WT_ASSERT(session, __wt_count_birthmarks(first_upd) <= 1);
 		switch (page->type) {
 		case WT_PAGE_COL_FIX:
 		case WT_PAGE_COL_VAR:
@@ -253,6 +256,7 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
 			break;
 		WT_ILLEGAL_VALUE_ERR(session, page->type);
 		}
+	}
 
 	/* Discard the cursor. */
 	WT_ERR(__wt_las_cursor_close(session, &cursor, session_flags));
--- a/src/btree/bt_split.c
+++ b/src/btree/bt_split.c
@@ -1408,12 +1408,12 @@ err:	if (parent != NULL)
 
 #ifdef HAVE_DIAGNOSTIC
 /*
- * __check_upd_list --
+ * __wt_count_birthmarks --
  *	Sanity check an update list.
  *	In particular, make sure there no birthmarks.
  */
-static void
-__check_upd_list(WT_SESSION_IMPL *session, WT_UPDATE *upd)
+int
+__wt_count_birthmarks(WT_UPDATE *upd)
 {
 	int birthmark_count;
 
@@ -1421,7 +1421,7 @@ __check_upd_list(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 		if (upd->type == WT_UPDATE_BIRTHMARK)
 			++birthmark_count;
 
-	WT_ASSERT(session, birthmark_count <= 1);
+	return (birthmark_count);
 }
 #endif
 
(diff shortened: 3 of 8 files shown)


lukech:
This is not a clean cherry-pick. Resolved the merge conflict by removing the `upd->start_ts` checking and `upd_unstable` counter bump code segment, which was introduced by WT-4842. Code compiled but failed the Python prepare lookaside test:
```
ERROR: test_prepare_lookaside01.test_prepare_lookaside01.test_prepare_lookaside (subunit.RemotedTestCase)
test_prepare_lookaside01.test_prepare_lookaside01.test_prepare_lookaside
----------------------------------------------------------------------
_StringException: lost connection during test 'test_prepare_lookaside01.test_prepare_lookaside01.test_prepare_lookaside'
```
@ddanderson Can you help to understand this failure? There might be some dependent changes missed in this 4.0 backport. Thanks.

lukech:
I got a stack trace from the core:
```(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007f48844df801 in __GI_abort () at abort.c:79
#2  0x00007f48826fc163 in __wt_abort (session=session@entry=0x55ba1fd1a3a0) at ../src/os_common/os_abort.c:32
#3  0x00007f488266e017 in __wt_las_insert_block (cursor=0x55ba1fd99d40, btree=0x55ba1fdc5510, page=0x7f48741bc960, multi=multi@entry=0x7f4874080470, key=0x7f487400a340) at ../src/cache/cache_las.c:754
#4  0x00007f4882725ac3 in __rec_las_wrapup (r=0x7f48743bcbe0, session=0x55ba1fd1a3a0) at ../src/reconcile/rec_write.c:2482
#5  __rec_write_wrapup (session=session@entry=0x55ba1fd1a3a0, r=r@entry=0x7f48743bcbe0, page=page@entry=0x7f48741bc960) at ../src/reconcile/rec_write.c:2301
#6  0x00007f4882727162 in __wt_reconcile (session=session@entry=0x55ba1fd1a3a0, ref=ref@entry=0x7f48780334c0, salvage=salvage@entry=0x0, flags=flags@entry=138, lookaside_retryp=lookaside_retryp@entry=0x0)
    at ../src/reconcile/rec_write.c:179
#7  0x00007f48826ce57e in __evict_review (inmem_splitp=<synthetic pointer>, evict_flags=0, ref=0x7f48780334c0, session=0x55ba1fd1a3a0) at ../src/evict/evict_page.c:710
#8  __wt_evict (session=session@entry=0x55ba1fd1a3a0, ref=ref@entry=0x7f48780334c0, previous_state=previous_state@entry=5, flags=flags@entry=0) at ../src/evict/evict_page.c:180
#9  0x00007f48826c6ef4 in __evict_page (session=session@entry=0x55ba1fd1a3a0, is_server=is_server@entry=false) at ../src/evict/evict_lru.c:2334
#10 0x00007f48826c7583 in __evict_lru_pages (session=session@entry=0x55ba1fd1a3a0, is_server=is_server@entry=false) at ../src/evict/evict_lru.c:1192
#11 0x00007f48826c9c92 in __wt_evict_thread_run (session=0x55ba1fd1a3a0, thread=0x55ba1fd6a3a0) at ../src/evict/evict_lru.c:337
#12 0x00007f488275fdd1 in __thread_run (arg=0x55ba1fd6a3a0) at ../src/support/thread_group.c:31
#13 0x00007f48842876db in start_thread (arg=0x7f487fe56700) at pthread_create.c:463
#14 0x00007f48845c088f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
The [newly added assertion](https://github.com/wiredtiger/wiredtiger/pull/4816/files#diff-2ae7a96a0871e2331a5670d3a812602dR750-R754) in `__wt_las_insert_block()` was triggered.

sueloverso:
@lukech While I don't know if your issue is related to the backport specifically, BF-14345 showed the same assertion failure and stack yesterday. Maybe this provides a reproducer case for that BF failure.

ddanderson:
@lukech, I took a close look at what you did for the cherry-pick, it looks right to me.  I think the new birthmark checking that is part of this commit exposed an existing problem in 4.0.  Don't know if it is fixed in the current code base.  I'll continue looking.

sueloverso:
Between this and the BF ticket, it seems worthwhile to open a WT ticket with a repro case if you have one. The BF was on mongoDB master so not 4.0.

lukech:
@sueloverso thanks for pointing out the BF ticket. With this 4.0 backport branch it reproed consistently with the `test_prepare_lookaside01` Python test. I'll leave to @ddanderson to open a WT ticket to address the BF.

ddanderson:
I still don't know if this is the same as the BF, but after several runs, here what I see during debugging just before the failure (with only slight variation):
```
(gdb) p first_upd
$22 = (WT_UPDATE *) 0x555556e18700
(gdb) p *first_upd
$23 = {txnid = 29200, timestamp = 4677, next = 0x555555cfd750, size = 500, type = 4 '\004', prepare_state = 0 '\000', 
  data = 0x555556e1871e 'd' <repeats 200 times>...}
(gdb) p *first_upd->next
$24 = {txnid = 23620, timestamp = 2, next = 0x555556321eb0, size = 500, type = 4 '\004', prepare_state = 1 '\001', 
  data = 0x555555cfd76e 'c' <repeats 200 times>...}
(gdb) p *first_upd->next->next
$25 = {txnid = 14778, timestamp = 4674, next = 0x555555eee7f0, size = 0, type = 1 '\001', prepare_state = 0 '\000', 
  data = 0x555556321ece ""}
(gdb) p *first_upd->next->next->next
$26 = {txnid = 0, timestamp = 0, next = 0x0, size = 500, type = 4 '\004', prepare_state = 0 '\000', 
  data = 0x555555eee80e 'a' <repeats 200 times>...}
(gdb) 
```
That is, in this update chain of length 4, only the last update in the chain is visible (I'm counting going down the chain following the next).  The 3rd is a birthmark, the 2nd is a prepare.  The code for `__wt_update_obsolete_check` determines that 4th (visible) update should be the new "first", but it's already "first" (in time) on the list, being at the end, so nothing changes.  @agorrod, is this data already in a situation that isn't supposed to happen?

I don't know why this same data situation doesn't occur for the same test in the develop tree.  Are there visibility changes that have not been backported?

agorrod:
Thanks for the diagnostic effort @ddanderson - this has hallmarks of other issues we solved at as follow ons to this work. The ones I could find were WT-4940 and WT-4943, and back ports for those pre-dated the test failure here, so they should be included.

Your diagnosis is spot on. The intent of this change was to make sure that reconciliation always resolved birthmarks before it's updates are written to the lookaside file. That's done as part of `__rec_append_orig_value`, the fact that the assertion is firing means that a page gets through reconciliation without having all of the birthmark updates resolved.

I don't think this is the same as the failure in the BF - this one seems more straight forward to reproduce. You've taken this investigation far enough for now, I'll find someone in Sydney to pick up the investigation work and move it forward.

tetsuo-cpp:
Superseded by #4838.
I'm closing this PR but will keep the branch around. Once you're happy with the migrated branch, please delete this one.

PR #4817, opened 2019-08-12
WT-4840 WT_CURSOR.modify must require explicit, snapshot-isolation transaction (v4.0 backport)(2nd attempt after revert)

lukech:
Disallow WT_CURSOR.modify in anything other than an explicit transaction
at snapshot isolation.

Add tests that other transaction configurations fail.

(cherry picked from commit 7c486eb1d4aecb67a635e06cdd74d179d69638f1)


lukech:
The same branch was reviewed & merged before, and later reverted in favor of 4.0.12 release. As the PR testing passed I will go ahead and merge again.

PR #4818, opened 2019-08-12
WT-5046 Prepared transactions aren't properly cleared from global table with WT_CONN_LOG_DEBUG_MODE enabled

tetsuo-cpp:
This PR is adding some additional logic to clear txn IDs from the global table when we have table logging on. To give a bit of background on what this is solving:

The lifecycle for a normal prepared txn is:
1. Begin txn.
2. Do operations. The first one will alloc a txn ID and a write to the log.
3. Prepare the txn. Here we clear the txn ID from the global table.
4. Commit the txn. We purposely don't clear the txn ID since we did it at the prepare stage.

When a txn has no operations, it gets neither a txn ID nor a log record allocated for it. When we have table logging turned on, an empty prepared txn will have its log record initialised (and a txn ID allocated) at commit time in order to write it to the log. With our existing logic, this causes the txn ID to not get cleared from the global table and prevents rollback to stable.

The lifecycle for an empty prepared txn is:
1. Begin txn.
2. Do no operations. We won't get a txn ID allocated or any log initialised.
3. Prepare the txn. We have no txn ID to clear from the global table so nothing special happens in that regard.
4. Commit the txn. Since we haven't initialised the log yet, we do it here and allocate a txn ID as part of this process. We purposely don't clear the txn ID since we supposedly did it at the prepare stage (!!!).

So this PR is modifying the clearing logic at the commit stage to clear the txn ID from the global table for not only non-prepared txns but ALSO prepared txns that have no modifications if table logging is on.

### DIFF
--- a/test/suite/test_debug_mode05.py
+++ b/test/suite/test_debug_mode05.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 #
-# Public Domain 2034-2039 MongoDB, Inc.
-# Public Domain 2008-2034 WiredTiger, Inc.
+# Public Domain 2014-2019 MongoDB, Inc.
+# Public Domain 2008-2014 WiredTiger, Inc.
 #
 # This is free and unencumbered software released into the public domain.
 #
--- a/test/suite/test_debug_mode05.py
+++ b/test/suite/test_debug_mode05.py
@@ -62,9 +62,9 @@ class test_debug_mode05(wttest.WiredTigerTestCase):
         self.conn.rollback_to_stable()
 
         # The original bug happened when we had a txn that:
-        # 1). Was prepared.
-        # 2). Did not cause anything to be written to the log before committing.
-        # 3). Was the last txn before the rollback to stable call.
+        # 1. Was prepared.
+        # 2. Did not cause anything to be written to the log before committing.
+        # 3. Was the last txn before the rollback to stable call.
         # Therefore, we're specifically not doing any operations here.
         self.session.begin_transaction()
         self.session.prepare_transaction(
@@ -75,6 +75,10 @@ class test_debug_mode05(wttest.WiredTigerTestCase):
             'durable_timestamp=' + timestamp_str(400))
         self.session.commit_transaction()
 
+        # The aforementioned bug resulted in a failure in rollback to stable.
+        # This is because we failed to clear out a txn id from our global state
+        # which caused us to think that we had a running txn.
+        # Verify that we can rollback to stable without issues.
         self.conn.rollback_to_stable()
 
         self.session.begin_transaction()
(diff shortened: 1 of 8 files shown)


test/suite/test_debug_mode05.py
```diff
@@ -0,0 +1,86 @@
...
+        # Therefore, we're specifically not doing any operations here.
+        self.session.begin_transaction()
+        self.session.prepare_transaction(
+            'prepare_timestamp=' + timestamp_str(300))
+        self.session.timestamp_transaction(
+            'commit_timestamp=' + timestamp_str(350))
+        self.session.timestamp_transaction(
+            'durable_timestamp=' + timestamp_str(400))
+        self.session.commit_transaction()
+
```
sueloverso: It would be good to add a comment here about what happens during this call when the bug existed versus what we expect now. I.e. the nature of the original bug. Did the call hang? Take an assertion failure? Return incorrect data?

test/suite/test_debug_mode05.py
```diff
@@ -0,0 +1,86 @@
...
+        self.conn.rollback_to_stable()
+
+        self.session.begin_transaction()
+        for i in range(1, 50):
+            cursor[i] = b'b' * 100
+        self.session.commit_transaction(
+            'commit_timestamp=' + timestamp_str(450))
+
+        self.conn.rollback_to_stable()
```
sueloverso: Doesn't this need:
```
if __name__ == '__main__':
    wttest.run()
```

tetsuo-cpp: cf2c797

test/suite/test_debug_mode05.py
tetsuo-cpp: cf2c797

test/suite/test_debug_mode05.py
```diff
@@ -0,0 +1,86 @@
...
+        self.session.timestamp_transaction(
+            'durable_timestamp=' + timestamp_str(250))
+        self.session.commit_transaction()
+
+        self.conn.rollback_to_stable()
+
+        # The original bug happened when we had a txn that:
+        # 1). Was prepared.
+        # 2). Did not cause anything to be written to the log before committing.
+        # 3). Was the last txn before the rollback to stable call.
```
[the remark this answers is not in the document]
tetsuo-cpp: cf2c797

sueloverso:
I did make a very general comment in the ticket about a different approach. You should read it, maybe talk it over with @agorrod and decide which direction to go.

tetsuo-cpp:
retest this please

tetsuo-cpp:
@sueloverso 
FYI, I've pushed the changes to just not make empty prepared txns log at all as you suggested. So this is ready to go. I mentioned this to @agorrod and I believe he's fine with it.

The other change was to add some diagnostics that @luke-pearson used when debugging the failure to begin with.

PR #4722, opened 2019-06-24
WT-4882 Allow metadata pages to split under load.

michaelcahill:

PR #4819, opened 2019-08-13
WT-4882 Allow metadata pages to split under load. (#4722) (v4.0 backport)

lukech:
Also add a wtperf workload that demonstrates the problem.

(cherry picked from commit 8313783f309cdfc61818607b8714ab0fe3bdec41)


lukech:
@luke-pearson This is a clean cherry-pick. Could you help to take a look?

PR #4823, opened 2019-08-13
WT-4810 Improve WT_ASSERT

luke-pearson:


src/include/error.h
```diff
@@ -128,9 +128,34 @@
...
+	}								\
+} while (0)
+#define	WT_RET_ASSERT(session, exp, v, ...) do {			\
+	if (!(exp))							\
+	{								\
+		int __ret = (v);					\
+		__wt_err(session, __ret, __VA_ARGS__);			\
+		__wt_abort(session);					\
+	}								\
+} while (0)
```
keithbostic: I don't think we need the `ret` or `__ret` temporary variables, and the names are a little odd (why `ret` in one case, and `__ret` in the other?)

keithbostic: The open brace in `WT_RET_ASSERT` is in the wrong place.

src/include/error.h
```diff
@@ -128,9 +128,34 @@
...
+#define	WT_ERR_ASSERT(session, exp, v, ...) do {			\
+	if (!(exp)) {							\
+		WT_ERR_MSG(session, v, __VA_ARGS__);			\
+	}								\
+} while (0)
+#define	WT_RET_ASSERT(session, exp, v, ...) do {			\
+	if (!(exp)) {							\
+		WT_RET_MSG(session, v, __VA_ARGS__);			\
+	}								\
+} while (0)
```
keithbostic: Open and closing braces aren't needed.

src/include/txn.i
```diff
@@ -216,8 +216,8 @@ __wt_txn_resolve_prepared_op(
 
 	/* If we haven't found anything then there's an error. */
 	if (upd == NULL) {
-		WT_ASSERT(session, upd != NULL);
-		WT_ERR(WT_NOTFOUND);
+		WT_ERR_ASSERT(session, upd != NULL, WT_NOTFOUND, "Unable to"
+		    " locate update associated with a prepared operation.");
```
keithbostic: We don't put trailing periods after messages because messages are usually concatenated together with separator characters, and a period in the middle of the error message looks odd. So, this is going to get the WiredTiger string associated with `WT_NOTFOUND` concatenated, so we'd get:
```
Unable to locate update associated with a prepared operation.: WT_NOTFOUND: item not found
```
and `<period><colon>` looks bad.

src/txn/txn.c
```diff
@@ -1331,8 +1333,10 @@ __wt_txn_rollback(WT_SESSION_IMPL *session, const char *cfg[])
 
 		__wt_txn_op_free(session, op);
 	}
-	WT_ASSERT(session, skip_update_assert ||
-	    resolved_update_count == visited_update_count);
+	WT_RET_ASSERT(session, skip_update_assert ||
+	    resolved_update_count == visited_update_count, EINVAL,
+	    "Expected number of resolved prepared updates does not"
+	    " match actual");
```
sueloverso: If we have a varargs here, why are we not giving the user the expected and actual values that failed the assert?

src/txn/txn.c
```diff
@@ -1024,8 +1024,11 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
 
 		__wt_txn_op_free(session, op);
 	}
-	WT_ASSERT(session, skip_update_assert ||
-	    resolved_update_count == visited_update_count);
+	WT_ERR_ASSERT(session, skip_update_assert ||
+	    resolved_update_count == visited_update_count, EINVAL,
+	    "Number of resolved prepared updates: %" PRId64 " does not match"
```
luke-pearson: @sueloverso, I've updated the message to include counts. I'm concerned the message is a bit cryptic thought but I'm not really sure how I could make it more meaningful. What do you think?

sueloverso: I think if/when the message triggers it will be helpful to the debuggers. It is no more cryptic than the same message but without the values. The first thing someone debugging this assertion failure is going to want to see is what those two values are. Thanks for adding it.

lukech:
retest this please

luke-pearson:
retest this please

PR #4824, opened 2019-08-15
WT-5052 Enhance lookaside search functionality to handle WT_MODIFY records

tetsuo-cpp:
This PR is adding support for modify records with the new lookaside logic. When we search lookaside and find a modify update, we keep walking backwards until we find a regular update and squash all the modifies in between on top of that base update.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -1414,10 +1414,11 @@ __wt_find_lookaside_upd(
 	WT_REF *ref;
 	WT_UPDATE *upd, *list[1000];
 	wt_timestamp_t durable_timestamp, las_timestamp;
+	wt_timestamp_t _durable_timestamp, _las_timestamp;
 	size_t incr;
-	uint64_t las_counter, las_pageid, las_txnid;
+	uint64_t las_counter, las_pageid, las_txnid, _las_txnid;
 	uint32_t las_id, session_flags;
-	uint8_t prepare_state, upd_type;
+	uint8_t prepare_state, upd_type, _prepare_state;
 	int cmp;
 	u_int i;
 	bool upd_visible;
@@ -1429,6 +1430,7 @@ __wt_find_lookaside_upd(
 	ref = cbt->ref;
 	las_pageid = ref->page_las->las_pageid;
 	session_flags = 0;		/* [-Werror=maybe-uninitialized] */
+	i = 0;
 
 	/* Open a lookaside table cursor. */
 	__wt_las_cursor(session, &cursor, &session_flags);
@@ -1503,7 +1505,6 @@ __wt_find_lookaside_upd(
 		 * Once we get to that point, squash the updates together.
 		 */
 		if (upd_type == WT_UPDATE_MODIFY) {
-			i = 0;
 			while (upd_type == WT_UPDATE_MODIFY) {
 				/*
 				 * TODO: Add asserts later.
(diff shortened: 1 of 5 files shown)


src/cache/cache_las.c
```diff
@@ -1397,6 +1397,15 @@ err:		__wt_buf_free(session, sweep_key);
...
 }
 
+/*
+ * When threads race modifying a record, we can end up with more than the max
+ * number of modifications in an update list.
+ *
+ * Leave space for double the maximum number on the stack as a best attempt to
+ * avoid heap allocation.
+ */
+#define	WT_MODIFY_ARRAY_SIZE	(WT_MAX_MODIFY_UPDATE * 2)
```
tetsuo-cpp: My approach towards allocation is pretty much identical to https://github.com/wiredtiger/wiredtiger/blob/develop/src/btree/bt_ret.c#L137.
There might be some potential for code reuse there.

sulabhM: We can explore code reuse once we start copying directly from the las cursor into the user cursor. I suggest leave a todo comment so that we do not forget.

src/cache/cache_las.c
```diff
@@ -1428,6 +1440,9 @@ __wt_find_lookaside_upd(
 	ref = cbt->ref;
 	las_pageid = ref->page_las->las_pageid;
 	session_flags = 0;		/* [-Werror=maybe-uninitialized] */
+	i = 0;
```
sulabhM: initialization is suggested in the same order as the declaration.

src/cache/cache_las.c
```diff
@@ -1489,12 +1504,78 @@ __wt_find_lookaside_upd(
 		if (upd_type == WT_UPDATE_BIRTHMARK && ignore_birthmark)
 			break;
 
-		/* It is okay to locate an older update in lookaside,
-		 * but we proceed only if we find a matching record. */
-		WT_ASSERT(session, start_ts >= las_timestamp);
+		/*
```
sulabhM: I suggest removing this comment. I do not think it is relevant after removing the assert

src/cache/cache_las.c
```diff
@@ -1489,12 +1504,78 @@ __wt_find_lookaside_upd(
...
+			while (upd_type == WT_UPDATE_MODIFY) {
+				WT_ERR(__wt_update_alloc(session,
+				    &las_value, &upd, &incr, upd_type));
+				if (i >= WT_MODIFY_ARRAY_SIZE) {
+					if (i == WT_MODIFY_ARRAY_SIZE)
+						listp = NULL;
+					WT_ERR(__wt_realloc_def(session,
+					    &allocated_bytes, i + 1, &listp));
+					if (i == WT_MODIFY_ARRAY_SIZE)
+						memcpy(
```
sulabhM: can this be reformatted as:
```
memcpy(listp,
    list, sizeof(list));
```

src/cache/cache_las.c
```diff
@@ -1489,12 +1504,78 @@ __wt_find_lookaside_upd(
...
+			}
+			WT_ASSERT(session, upd_type == WT_UPDATE_STANDARD);
+
+			/*
+			 * This value is a pointer/size pair pointing at a spot
+			 * in the cursor's buffer. Our modify function needs to
+			 * know to allocate its own buffer when applying
+			 * modifications so explicitly unset these.
+			 */
+			las_value.mem = NULL;
```
sulabhM: If we do ```WT_CLEAR(las_value)``` at the start of this function with the rest of initializations, will we need to zero them here? Maybe that would be cleaner ?

src/cache/cache_las.c
```diff
@@ -1489,12 +1504,78 @@ __wt_find_lookaside_upd(
...
+			las_value.memsize = 0;
+			while (i > 0) {
+				WT_ERR(__wt_modify_apply_item(
```
sulabhM: instead of ```cursor->session``` you can pass ```session``` that is locally available to this function. They are the same session I imagine.

src/cache/cache_las.c
```diff
@@ -1516,6 +1597,12 @@ __wt_find_lookaside_upd(
 
 err:	__wt_readunlock(session, &cache->las_sweepwalk_lock);
 	WT_TRET(__wt_las_cursor_close(session, &cursor, session_flags));
+	if (listp == NULL)
```
sulabhM: Are we leaking memory when ```listp``` is not null, in that case we do not free ```list``` that we copied to ```listp```. Can that be potentially an issue with the original code in ```__wt_value_return_upd ``` as well?

test/suite/test_las06.py
```diff
@@ -104,5 +104,67 @@ def test_las_reads_workload(self):
...
+            cursor[i] = value2
+        self.session.commit_transaction('commit_timestamp=' + timestamp_str(5))
+
+        # Now the latest version will get written to the data file.
+        self.session.checkpoint()
+
+        expected = list(value1)
+        expected[100] = 'B'
+        expected[200] = 'C'
+        expected = str().join(expected)
```
sulabhM: haha! I do not enjoy immutable strings

src/cache/cache_las.c
tetsuo-cpp: 067612a

Added a todo.

src/cache/cache_las.c
tetsuo-cpp: 067612a

Reordered.

src/cache/cache_las.c
tetsuo-cpp: 067612a

src/cache/cache_las.c
tetsuo-cpp: 067612a

src/cache/cache_las.c
tetsuo-cpp: I think that `session` is the one that's doing the read txn whereas the one I'm passing in is a LAS session.

src/cache/cache_las.c
tetsuo-cpp: That null check is for if we jump to `err` due to a failed `__wt_realloc_def` in which case `listp` will be `NULL` and won't have a heap allocated list assigned to it. So I just assign it back to `list` and free the updates in `list`.

I check `allocated_bytes` to see whether I exceeded the capacity of `list` and had to do the realloc. In that case, I free the underlying list also.

src/cache/cache_las.c
tetsuo-cpp: Done.

src/cache/cache_las.c
sulabhM: Ahh! ya, you are right this is las cursor and hence not the same user session. I think It is better to call with the user session, might be useful when logging to know what session called modify.

src/cache/cache_las.c
sulabhM: You are right, everything seems accounted for.

src/cache/cache_las.c
tetsuo-cpp: 24f2eea

Makes sense. Switched to the user session.

PR #4827, opened 2019-08-16
WT-5066 Extend test_las06 to support column store.

tetsuo-cpp:
I've just added a scenario to use the column store in `test_las06`.
This PR is pointing at my other branch, `wt-5052-las-modify` since that branch is also touching this test. Once that gets merged, I'll point this back to `relevant_history_in_cache`.


sulabhM:
Why merge into wt-5052-las-modify and not relevant_history_in_cache ?

sulabhM:
Ahh! got it. I think it is okay to merge into wt-5052-las-modify

PR #4829, opened 2019-08-19
WT-5063 cursor modify not supported error

kommiharibabu:


kommiharibabu:
retest this please

kommiharibabu:
evergreen retry

kommiharibabu:
@sueloverso, I went with the new function approach compared to adding a duplicate code in a bunch of functions to provide a proper error message. I tried to rename the function that is good for the readability and also added a comment in the code in explaining why we choose two default functions instead of one.

kommiharibabu:
@agorrod, Following is the error message output of cursor modify from unsupported cursor types.

Python test:
WiredTigerError: Operation not supported

C Test:

PR #4830, opened 2019-08-19
WT-4658 Clang Format Preparation (v3.6 backport)

tetsuo-cpp:


agorrod:
I'm inclined to hold off on back-porting to 3.6. We need to draw the line somewhere - I'm at least inclined to see how much fallout this change brings before deciding if it should be backported to 3.6

tetsuo-cpp:
Sure. I'll leave the branch around in case we want to use it later.

PR #4835, opened 2019-08-19
WT-4961 Checkpoints with cache overflow must keep history for reads. (Clang Format migration)

tetsuo-cpp:
This is a migration of #4786.

### DIFF
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -299,21 +299,9 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
              * cache clean but with history that cannot be
              * discarded), that is not wasted effort because
              * checkpoint doesn't need to write the page again.
-             *
-             * Once the transaction has given up it's snapshot it
-             * is no longer safe to reconcile pages. That happens
-             * prior to the final metadata checkpoint.
-             *
-             * XXX Only attempt this eviction when there are no
-             * readers older than the checkpoint.  Otherwise, a bug
-             * in eviction can mark the page clean and discard
-             * history, causing those reads to incorrectly see
-             * newer versions of data than they should.
              */
             if (!WT_PAGE_IS_INTERNAL(page) && page->read_gen == WT_READGEN_WONT_NEED &&
-              !tried_eviction && F_ISSET(&session->txn, WT_TXN_HAS_SNAPSHOT) &&
-              (!F_ISSET(txn, WT_TXN_HAS_TS_READ) ||
-                txn->read_timestamp == conn->txn_global.pinned_timestamp)) {
+              !tried_eviction) {
                 WT_ERR_BUSY_OK(__wt_page_release_evict(session, walk, 0));
                 walk = prev;
                 prev = NULL;
--- a/dist/api_data.py
+++ b/dist/api_data.py
@@ -1436,8 +1436,8 @@ methods = {
         dropped while a hot backup is in progress or if open in
         a cursor''', type='list'),
     Config('force', 'false', r'''
-        by default, checkpoints may be skipped if the underlying object
-        has not been modified, this option forces the checkpoint''',
+        if false (the default), checkpoints may be skipped if the underlying object has not been
+        modified, if true, this option forces the checkpoint''',
         type='boolean'),
     Config('name', '', r'''
         if set, specify a name for the checkpoint (note that checkpoints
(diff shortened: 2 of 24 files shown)


src/btree/bt_sync.c
```diff
@@ -299,21 +299,9 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
-             * readers older than the checkpoint.  Otherwise, a bug
-             * in eviction can mark the page clean and discard
-             * history, causing those reads to incorrectly see
-             * newer versions of data than they should.
              */
             if (!WT_PAGE_IS_INTERNAL(page) && page->read_gen == WT_READGEN_WONT_NEED &&
-              !tried_eviction && F_ISSET(&session->txn, WT_TXN_HAS_SNAPSHOT) &&
-              (!F_ISSET(txn, WT_TXN_HAS_TS_READ) ||
-                txn->read_timestamp == conn->txn_global.pinned_timestamp)) {
+              !tried_eviction) {
```
tetsuo-cpp: This was the only conflict involved. Someone will need to take a look at this.

keithbostic: @agorrod should confirm, but this looks correct to me, that is, the current PR branch looks right.

lukech:
It looks `wiredtiger-test-format-stress-sanitizer` job is still running on the `wt-4961-checkpoint-eviction-keep-history` branch and failing periodically. While I'm not sure if it deserves more time to run the feature branch (considering there were known failures), after clang format merge and a new migration branch created, it makes sense at least to switch the branch over to the new one in the Jenkins job?

keithbostic:
@agorrod, @bvpvamsikrishna, this branch ran successfully overnight with a mixture of `CONFIG.stress` and the CONFIG that was failing on my local machine, so if you approve it, I think we could merge.

We're going to need a lot of burn-in on this change, I think, before we let it get downstream.

lukech:
FYI - I updated `wiredtiger-test-format-stress-sanitizer` job to test against `wt-4961-checkpoint-eviction-keep-history-migration` branch (used in this PR).

tetsuo-cpp:
Is this PR just waiting on testing?

agorrod:
@bvpvamsikrishna is working on diagnosing some Evergreen test failures, that appear to be related to the checkpoint timestamp being reset (at least, that's what I remember from the last standup).

bvpvamsikrishna:
@agorrod, @keithbostic  could you review the last commit. Currently, there is one failure in the evergreen patch build with this change.

agorrod:
Nice work figuring this failure out Vamsi - it's nice to see the Evergreen tests green.

I'm not very comfortable using 1 as a special timestamp value - I think we should move the code away from using a WT_TS_NONE value as an indicator and keep a separate variable to track that if we can.

bvpvamsikrishna:
@agorrod , @keithbostic , I have reverted the if condition in rollback_to_stable_check as I think we can rollback even if the first time checkpoint is not complete.  request for a review.

keithbostic:
@bvpvamsikrishna, can you please explain what it means when we rollback-to-stable if we don't have a checkpoint in the system? I guess I'm (at least), confused about what the word "stable" means in that situation. Regardless, I'll defer to you & @agorrod on that question.

Also, the changes to require a checkpoint before rollback-to-stable was allowed were more pervasive than this: there were changes to the documentation, example code and test code as well, and all that needs to be rolled back.

bvpvamsikrishna:
> @bvpvamsikrishna, can you please explain what it means when we rollback-to-stable if we don't have a checkpoint in the system? I guess I'm (at least), confused about what the word "stable" means in that situation. Regardless, I'll defer to you & @agorrod on that question.

I am considering rollback to stable as "discarding data from cache overflow" rather than in relation with checkpoint. So, after the system started and before first checkpoint:
1. some operations are performed.
2. Stable timestamp is set.
3. some more operations are performed.
4. rollback to stable is invoked.
In this scenario, the data till stable timestamp to be retained.

agorrod:
Let's talk this through when I'm back in Australia on Tuesday - I haven't been following along and I think it'll be easier in person.

bvpvamsikrishna:
> Let's talk this through when I'm back in Australia on Tuesday - I haven't been following along and I think it'll be easier in person.

Sure.

src/btree/bt_sync.c
agorrod: @keithbostic good question about backing out the workaround. This is the change that does that, and it looks correct to me

bvpvamsikrishna:
@agorrod , @keithbostic I have updated the branch with the changes we have discussed in the morning. This is link to [evergreen patch build](https://evergreen.mongodb.com/version/5d8ab0a72fbabe60735e120c).

Request your review.

PR #4782, opened 2019-07-23
WT-4955 Extend backup API to allow for efficient incremental copy

keithbostic:

PR #4836, opened 2019-08-20
WT-4955 Extend backup API to allow for efficient incremental copy (Clang Format migration)

tetsuo-cpp:
This is a migration of #4782.


tetsuo-cpp:
@keithbostic This Jenkins build failure is the same as the one in the original PR.

PR #4843, opened 2019-08-23
WT-4831 Add ignore unexpected output python test suite flag

luke-pearson:


luke-pearson:
The only thing here I'm not too sure about is the naming

PR #4844, opened 2019-08-23
WT-4535 Setting readonly mode on various command line tools

luke-pearson:


src/utilities/util_main.c
```diff
@@ -217,25 +220,28 @@ main(int argc, char *argv[])
             goto done;
         } else if (strcmp(command, "create") == 0) {
             func = util_create;
-            config = "create";
+            config = "create," READONLY;
```
keithbostic: I don't think `create` is readonly, it's creating an object in the database.

luke-pearson:
@keithbostic I'm not sure how much sense it makes to continue with this change as we frequently hit this assertion __wt_open, 247: lock_file || !LF_ISSET(WT_FS_OPEN_CREATE), list, dump etc when the turtlefile is not present which would get created if we turned off readonly. Perhaps more work is required to rewrite the tools to not expect a turtle file or detect a missing turtle file with ./wt and log a "please recover your turtle file" esq message?

luke-pearson:
> 
> 
> @luke-pearson,
> 
> > I'm not sure how much sense it makes to continue with this change as we frequently hit this assertion __wt_open, 247: lock_file || !LF_ISSET(WT_FS_OPEN_CREATE), list, dump etc when the turtlefile is not present which would get created if we turned off readonly. Perhaps more work is required to rewrite the tools to not expect a turtle file or detect a missing turtle file with ./wt and log a "please recover your turtle file" esq message?
> 
> I strongly agree we should be asserting in expected paths (especially because `WT_ASSERT` doesn't do anything unless compiled in `#HAVE_DIAGNOSTIC` mode).
> 
> It feels like a bug to me to run the `wt` command in any database that doesn't have a turtle file, that's a corrupted database by definition, isn't it?
> 
> I'm pretty sure I don't understand completely the problem you're seeing, so let's talk about it next time we're both on line.

I think for the most part the issue is with list / backup when you call it on a database missing a .turtle, as we automatically think we've got a new connection and therefore cannot be readonly. `conn->is_new = exist ? 0 : 1;` in conn_api.c is the reason for this. For a quick repro you can just run test_backup01 in test/suite. So again perhaps logging an error might be the best path forward? Additionally backup is not readonly as it creates temp objects as such I've removed it's flag, Alex G and I discussed this and we think its best to create a follow up ticket to investigate the work required to make backup actually readonly for the original database.

luke-pearson:
> Ideally, that would fail, relatively hard. My point-of-view is a missing turtle file isn't a special case, it's a corrupted database. In other words, is there a "normal" situation where the turtle file is missing, and that's OK?

Not that I can think of, I think that'd warrant printing a message in stderr and exiting wt with non zero?

agorrod:
bq. Not that I can think of, I think that'd warrant printing a message in stderr and exiting wt with non zero?

As long as using any readonly `wt` command line on an empty or non-database directory replies with a reasonable error message, I agree.

luke-pearson:
So when we run backup we don't create a .turtle and create a .backup file and then obviously if you try and run list or dump in the backup directory it'll fail. The current plan is to add a command line argument which will allow commands that are normally readonly to be non-readonly additionally an error will be logged when we run a readonly command on an empty or non database directory. This will return non-zero exit code. Does everyone think this is a reasonable path forward?

src/conn/conn_api.c
```diff
@@ -1739,8 +1739,8 @@ __conn_single(WT_SESSION_IMPL *session, const char *cfg[])
     if (conn->is_new) {
         if (F_ISSET(conn, WT_CONN_READONLY))
             WT_ERR_MSG(session, EINVAL,
-              "Creating a new database is incompatible with "
-              "read-only configuration");
+              "The database metadata file is missing and a read-only connection cannot continue; "
+              "consider running recovery.");
```
sueloverso: I don't feel the new message here is an improvement. The most common case is creating a new database, not that someone removed the turtle file or metadata file. At a minimum I suggest removing the "consider running recovery" because that is not helpful when creating a new database. @keithbostic and @agorrod - you were both involved more in the message discussion. If you prefer this, I'll go with it.

src/utilities/util_main.c
```diff
@@ -289,6 +297,9 @@ main(int argc, char *argv[])
         goto err;
     }
 
+    if (recover || salvage)
```
sueloverso: I'd like a comment here talking about how salvage and recovery override the readonly option. It is not immediately obvious.

test/suite/test_backup01.py
```diff
@@ -64,7 +64,7 @@ def populate(self, skiplsm):
     # Compare the original and backed-up files using the wt dump command.
     def compare(self, uri):
         self.runWt(['dump', uri], outfilename='orig')
-        self.runWt(['-h', self.dir, 'dump', uri], outfilename='backup')
+        self.runWt(['-h', self.dir, '-R', 'dump', uri], outfilename='backup')
```
sueloverso: Please add a comment here (to all the tests needed) about why this is now necessary.

src/conn/conn_api.c
keithbostic: I don't see how create is the likely/common path. You'd have to explicitly set the read-only flag on an operation that requires database creation, and that's not something someone is likely to do? What am I missing?

I don't care if we mention `metadata file is missing`, it feels like an internal detail, happy to see that go away.

I don't like error messages that guess at what the user is doing and so can point the user in the wrong direction as they try to fix the problem. (In other words, don't mention `create` unless we know without question that the user is trying to create a database.)

@luke-pearson, as far as I'm concerned, feel free to just decide without further input from me.

PR #4850, opened 2019-08-29
WT-5074 Fix "make check" on exotic architectures

tetsuo-cpp:
This PR is removing code generation from `make check`. Since we began using `clang-format` as part of our code generation process, this has become a problem on architectures such as ZSeries and PowerPC since we don't have the appropriate binaries for them (and maintaining them all would be untenable).

I think it's reasonable to remove this from the build process since I don't believe it belongs here anyway. To my way of thinking, this is unnecessarily making a development dependency also a build dependency. Just as an example, if you have a project that uses code generated by Protobuf, I wouldn't expect to need `protoc` to build it (although developers would need it).

Please let me know your thoughts on this.

### DIFF
--- a/build_posix/Make.base
+++ b/build_posix/Make.base
@@ -53,13 +53,6 @@ $(srcdir)/Makefile.am: $(srcdir)/build_posix/Make.base $(srcdir)/build_posix/mak
 libtool: $(LIBTOOL_DEPS)
 	$(SHELL) ./config.status libtool
 
-$(srcdir)/src/include/extern.h: auto-includes.chk
-$(srcdir)/src/include/wt_internal.h: auto-includes.chk
-
-auto-includes.chk: $(libwiredtiger_la_SOURCES)
-	@(cd $(srcdir)/dist && \
-	    python prototypes.py && sh s_typedef -b) && touch $@
-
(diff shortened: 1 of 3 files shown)


lukech:
I agree to the point that a development dependency (e.g dist/s_*) is not necessarily making a build/testing dependency. And we do have some jobs run `dist/s_*` in case some bugs slips through the development process.

@tetsuo-cpp would you take this chance to replace `wget` with some more reliable approach (e.g. curl) to download the clang_format binary? A known issue is the Evergreen MacOS hosts do not have wget preinstalled and attempts of running `s_clang_format` script would fail.

tetsuo-cpp:
@lukech I think the only reason that failure happens is due to `make check` which will now no longer require `clang-format`. But I think it's a good idea to switch to `curl` given that @sueloverso had issues installing `wget`. I'll follow up with a commit shortly.

sueloverso:
@tetsuo-cpp I agree with the approach here. I have no idea if the dependency changes are correct or not.

However regarding the Mac, I did pull your branch on my old, personal Mac, where I had not yet done anything regarding `wget` or `clang-format`, and confirmed that I am able to build out-of-the-box with `curl`!

PR #4853, opened 2019-08-29
WT-5088 Refine clearing and setting of debug timer.

sueloverso:
Not yet ready for review.


sueloverso:
@keithbostic I have run some of the failing CONFIGs with this branch and they're passing. A lot of Jenkins is hitting this. Right now this tracking code is initialized at the start of eviction, so only eviction issues should trigger it and this hopefully eliminates all the rest of the false positives.

PR #4855, opened 2019-09-02
WT-5092 Modify lookaside schema to add timestamp and transaction id to the key

tetsuo-cpp:
This PR is to replace the `las_counter` part of the key in the lookaside schema and replace it with `las_timestamp` and `las_txnid` for more efficient searches.

Since those values are now embedded in the key, I've removed them from the value.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -998,14 +998,14 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
     uint32_t las_btree_id, saved_btree_id, session_flags;
     uint8_t prepare_state, upd_type;
     int cmp, notused;
-    bool local_txn, locked;
+    bool local_txn, locked, removing_key_block;
 
     cache = S2C(session)->cache;
     cursor = NULL;
     sweep_key = &cache->las_sweep_key;
     remove_cnt = 0;
     session_flags = 0; /* [-Werror=maybe-uninitialized] */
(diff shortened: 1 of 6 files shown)


src/cache/cache_las.c
```diff
@@ -827,13 +823,16 @@ __wt_las_cursor_position(
          * There may be no block of lookaside entries if they have been
```
sulabhM: There aren't blocks anymore, maybe edit this comment to say new entries can be inserted in between.

src/cache/cache_las.c
```diff
@@ -1073,20 +1073,23 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
         }
 
         /*
-         * We only want to break between key blocks. Stop if we've processed enough entries either
-         * all we wanted or enough and there is a reader waiting and we're on a key boundary.
+         * Stop if we've processed enough entries either all we wanted or enough and there is a
+         * reader waiting.
          */
         ++visit_cnt;
-        if (!removing_key_block &&
```
tetsuo-cpp: @sulabhM I've just removed this part of the check. To my understanding, it's just there to avoid breaking out midway through removing a block and having half of an old block in the las.

tetsuo-cpp: Can't just remove this. We're going to be dealing with sweep properly later.

PR #4856, opened 2019-09-02
WT-5058 Stop loading content from LAS when initializing a lookaside evicted page

keithbostic:
@sulabhM, the code snippet I passed along hadn't actually been tested -- this change passes testing for me.


sulabhM:
Thanks Keith, I created https://github.com/wiredtiger/wiredtiger/pull/4858 to get this change in. It is cleaner that I have merged wt-5058-no-las already.

I will close this pull request and use the other on instead.

PR #4857, opened 2019-09-03
WT-4893 Fix a race between internal page child-page eviction checks a…

luke-pearson:
…nd cursors in the tree(#4749)

When checking if an internal page's children will block the internal page's
eviction, there may be cursors in the tree walking the list of child pages.
The parent is locked, so all we care about is cursors already in the child
pages, no thread can enter them. Any cursor moving through the child pages
must be hazard pointer coupling between pages, where the page on which it
currently has a hazard pointer must be in a state other than on-disk. Walk
the child list forward, then backward, to ensure we don't race with a cursor
walking in the opposite direction from our check.

Also:
* Skip work in __wt_hazard_check() and __wt_hazard_check_assert()
if we aren't maintaining hazard pointers on the tree.
* Don't clear the WT_REF history when we overwrite the WT_REF on
discard, it's useful diagnostic information.

(cherry picked from commit 82082debc5983bfc8e096bec4d812cb085cc026d)

### DIFF
--- a/src/evict/evict_page.c
+++ b/src/evict/evict_page.c
@@ -446,11 +446,47 @@ __evict_child_check(WT_SESSION_IMPL *session, WT_REF *parent)
     WT_REF *child;
     bool active;
 
+    /*
+     * There may be cursors in the tree walking the list of child pages. The parent is locked, so
+     * all we care about is cursors already in the child pages, no thread can enter them. Any cursor
+     * moving through the child pages must be hazard pointer coupling between pages, where the page
+     * on which it currently has a hazard pointer must be in a state other than on-disk. Walk the
+     * child list forward, then backward, to ensure we don't race with a cursor walking in the
+     * opposite direction from our check.
+     */
+    WT_INTL_FOREACH_BEGIN (session, parent->page, child) {
+        switch (child->state) {
+        case WT_REF_DISK:      /* On-disk */
+        case WT_REF_DELETED:   /* On-disk, deleted */
+        case WT_REF_LOOKASIDE: /* On-disk, lookaside */
+            break;
+        default:
+            return (__wt_set_return(session, EBUSY));
+        }
+    }
+    WT_INTL_FOREACH_END;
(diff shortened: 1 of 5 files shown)


src/evict/evict_page.c
```diff
@@ -446,11 +446,47 @@ __evict_child_check(WT_SESSION_IMPL *session, WT_REF *parent)
     WT_REF *child;
     bool active;
 
+    /*
+     * There may be cursors in the tree walking the list of child pages. The parent is locked, so
+     * all we care about is cursors already in the child pages, no thread can enter them. Any cursor
+     * moving through the child pages must be hazard pointer coupling between pages, where the page
+     * on which it currently has a hazard pointer must be in a state other than on-disk. Walk the
```
tetsuo-cpp: Is there not a more direct way of checking this? I see a `__wt_hazard_check` that seems to be trying something similar. I'm sure there's a reason it has been done like this but I'd be interested to know what it is.

agorrod: @keithbostic could you please answer this question?

keithbostic: tl;dr: Would absolutely work,  but would be a lot slower.

I agree we could use `__wt_hazard_check()` to accomplish the same check.

The way to do it would be to loop through the parent page's list of child `WT_REFs`, and call `__wt_hazard_check()` on each one, which would return if there was an active `WT_SESSION` holding a hazard pointer on that `WT_REF`.

That would be a lot slower because `__wt_hazard_check()` requires underlying read/write barriers (the size of the `WT_SESSION` hazard pointer array can grow which implies magic), and we'd have to walk the complete list of active `WT_SESSION` handles for every `WT_REF` in the list instead of walking the `WT_REF` list once, so it's MxN more expensive operations versus N cheaper operations.

And, we'd still have to do both the forward & backward passes, the race between our check and a cursor moving in the opposite direction is still possible when calling `__wt_hazard_check()`.

PR #4859, opened 2019-09-03
WT-5100 Update test format to pass -R to disable readonly in wt util

luke-pearson:
The statistics_log configuration used by wiredtiger is not readonly, as such when we close the connection and then open it with any readonly wiredtiger util. e.g. dump it reads the config from the WiredTiger.basecfg file which tells it to begin statistic logging, which triggers this assertion:

`[1567488766:358842][30124:0x7f1ed3d3d700], wt, statlog-server: __wt_open, 248: lock_file || !LF_ISSET(WT_FS_OPEN_CREATE)`

### DIFF
--- a/test/format/rebalance.c
+++ b/test/format/rebalance.c
@@ -40,9 +40,9 @@ wts_rebalance(void)
 
     track("rebalance", 0ULL, NULL);
 
-    /* Dump the current object. */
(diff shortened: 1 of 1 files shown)


lukech:
@keithbostic if this change looks good to you, can you help to merge it? We hope the fix will address the failures we saw in those test/format jobs. Cheers

PR #4860, opened 2019-09-03
WT-5068 WT_CURSOR.random unnecessarily returns duplicate records

keithbostic:
@agorrod, this branch reworks the "pick a random page and return a random key/value item from it" version of `WT_CURSOR.next` configured for random key/value returns.

It appears to fix the [SERVER-42840](https://jira.mongodb.org/browse/SERVER-42840) failure.

I ended up with something that I think works better, but the change is chock-full of magic constants I can't defend, there's an awful lot of guesses about what is most likely to return a better random selection.

### DIFF
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -61,6 +61,8 @@ __random_skip_entries(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_INSERT_
     uint32_t entries;
     int level;
 
+    entries = 0; /* [-Wconditional-uninitialized] */
+
     if (ins_head == NULL)
         return (0);
 
--- a/src/btree/bt_random.c
+++ b/src/btree/bt_random.c
@@ -9,147 +9,330 @@
 #include "wt_internal.h"
 
 /*
- * __wt_row_random_leaf --
- *     Return a random key from a row-store leaf page.
+ * __random_insert_valid --
+ *     Check if the inserted key/value pair is valid.
  */
-int
-__wt_row_random_leaf(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt)
+static int
+__random_insert_valid(
+  WT_CURSOR_BTREE *cbt, WT_INSERT_HEAD *ins_head, WT_INSERT *ins, WT_UPDATE **updp, bool *validp)
 {
-    WT_INSERT *ins, **start, **stop;
-    WT_INSERT_HEAD *ins_head;
-    WT_PAGE *page;
-    uint64_t samples;
-    uint32_t choice, entries, i;
-    int level;
-
-    page = cbt->ref->page;
-    start = stop = NULL; /* [-Wconditional-uninitialized] */
-    entries = 0;         /* [-Wconditional-uninitialized] */
+    *updp = NULL;
+    *validp = false;
 
     __cursor_pos_clear(cbt);
+    cbt->slot = 0;
+    cbt->ins_head = ins_head;
+    cbt->ins = ins;
+    cbt->compare = 0;
 
-    /* If the page has disk-based entries, select from them. */
-    if (page->entries != 0) {
-        cbt->compare = 0;
-        cbt->slot = __wt_random(&session->rnd) % page->entries;
+    return (__wt_cursor_valid(cbt, updp, validp));
+}
 
-        /*
-         * The real row-store search function builds the key, so we have to as well.
-         */
-        return (__wt_row_leaf_key(session, page, page->pg_row + cbt->slot, cbt->tmp, false));
-    }
+/*
+ * __random_slot_valid --
+ *     Check if the slot key/value pair is valid.
+ */
+static int
+__random_slot_valid(WT_CURSOR_BTREE *cbt, uint32_t slot, WT_UPDATE **updp, bool *validp)
+{
+    *updp = NULL;
+    *validp = false;
 
-    /*
-     * If the tree is new (and not empty), it might have a large insert
-     * list.
-     *
-     * Walk down the list until we find a level with at least 50 entries,
-     * that's where we'll start rolling random numbers. The value 50 is
-     * used to ignore levels with only a few entries, that is, levels which
-     * are potentially badly skewed.
-     */
-    F_SET(cbt, WT_CBT_SEARCH_SMALLEST);
-    if ((ins_head = WT_ROW_INSERT_SMALLEST(page)) == NULL)
-        return (WT_NOTFOUND);
+    __cursor_pos_clear(cbt);
+    cbt->slot = slot;
+    cbt->compare = 0;
+
+    return (__wt_cursor_valid(cbt, updp, validp));
+}
+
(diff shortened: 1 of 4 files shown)


agorrod:
@sueloverso and @sulabhM  it would be great if you can do a review of this change. I talked it through at a high level with Keith, and it makes sense to me.

src/btree/bt_random.c
```diff
@@ -9,147 +9,332 @@
...
+}
 
-    /* If the page has disk-based entries, select from them. */
-    if (page->entries != 0) {
-        cbt->compare = 0;
-        cbt->slot = __wt_random(&session->rnd) % page->entries;
+/* Magic constant: 50 entries in a skip list is enough to predict the size. */
+#define WT_RANDOM_SKIP_PREDICT 50
+/* Magic constant: 5000 entries in a skip list is enough to forcibly evict. */
+#define WT_RANDOM_SKIP_EVICT_SOON 5000
```
sueloverso: Can all of these be moved to the top of the file (or an appropriate header file)? Having them scattered many times throughout the source file, even though I know you put them near the function using them, is difficult to read and since they all have a similar or the same prefix, there's risk of overlapping names since you cannot visually see them together.

src/btree/bt_random.c
```diff
@@ -9,147 +9,332 @@
...
 
-        /*
-         * The real row-store search function builds the key, so we have to as well.
-         */
-        return (__wt_row_leaf_key(session, page, page->pg_row + cbt->slot, cbt->tmp, false));
-    }
+/*
+ * __random_skip_entries --
+ *     Return a guess of how many entries are in a skip list.
```
sueloverso: Minor nit: I'd prefer "a guess" to be "an estimate". The word `guess` has connotations of `without basis` to me.

src/btree/bt_random.c
```diff
@@ -9,147 +9,332 @@
...
+    if (entries >= WT_RANDOM_SKIP_INSERT_ENOUGH) {
+        WT_RET(__random_leaf_skip(session, cbt, ins_head, entries, updp, validp));
+        if (*validp)
+            return (0);
+    }
+    return (0);
+}
+
+/* Magic constant: retry 10 times in the disk-based entries before giving up. */
+#define WT_RANDOM_DISK_RETRY 10
```
sueloverso: There is also `WT_RETRY_MAX` as a general retry macro, used in several places, that is also 10. You could just use that here.

src/btree/bt_random.c
```diff
@@ -9,147 +9,332 @@
...
+            return (__cursor_kv_return(session, cbt, upd));
+    }
+
+    /*
+     * We don't have many disk-based entries, we didn't find any large insert lists. Where we get
+     * into trouble is a small number of pages with large numbers of deleted items. Try and move out
+     * of the problematic namespace into something we can use by cursoring forward or backward. On a
+     * page with a sufficiently large group of deleted items that the randomly selected entries were
+     * all deleted, simply moving to the next record likely means returning the same record every
+     * time, so cursor a random number of items instead of doing just moving to the next item.
```
sueloverso: This line is very awkwardly worded. I suggest:
` * time, so move the cursor a random number of items instead of just moving to the next item.`

src/include/cursor.h
```diff
@@ -152,6 +152,8 @@ struct __wt_cursor_btree {
      *
      * A temporary buffer for building on-page keys when searching row-store
      * files.
+     *
+     * A temporary buffer for tracking random key return values to avoid repetition.
```
sueloverso: This is unclear to me. Having such a similar comment right after the lines above with no fields in between makes me think the new comment is supposed to supercede the one on line 153. Maybe:
`A temporary buffer for building on-page keys when searching row-store files or for tracking random key return values.`

test/suite/test_cursor_random.py
```diff
@@ -95,48 +95,58 @@ def test_cursor_random_single_record(self):
...
+    def test_cursor_random_multiple_page_records_reopen_small(self):
+        self.cursor_random_multiple_page_records(2000, True)
+    def test_cursor_random_multiple_page_records_reopen_large(self):
+        self.cursor_random_multiple_page_records(10000, True)
+    def test_cursor_random_multiple_page_records_small(self):
+        self.cursor_random_multiple_page_records(2000, False)
+    def test_cursor_random_multiple_page_records_large(self):
+        self.cursor_random_multiple_page_records(10000, False)
+
+    # Check that next_random succeeds in the presence of a set of values, some of
```
sueloverso: Not related to this specifically: Do we have a test to test against a tiny tree to make sure we get the "alternating" keys and get into the tiny tree code above?

src/btree/bt_random.c
keithbostic: I'd prefer not to move them (as you note, they appear where they're used), and they're only used once. Overlapping names would get flagged by the compiler, right?

We can give @sulabhM the vote. :)

src/btree/bt_random.c
keithbostic: Sure, fixed in 7bbad1497.

src/btree/bt_random.c
keithbostic: I'd prefer not using an "outside" macro here (that's the OS system call retry max), which is wildly different in meaning from this one.

src/btree/bt_random.c
keithbostic: Sure, fixed in 7bbad1497.

src/include/cursor.h
keithbostic: Sure, fixed in 7bbad1497 (I numbered the paragraphs so it's clear there are multiple uses).

test/suite/test_cursor_random.py
keithbostic: Yes, the `Check that next_random succeeds in the presence of a set of values, some of which are deleted.` will hit that code hard because all but 10 records in the 10K record set are deleted.

src/btree/bt_random.c
sueloverso: There really isn't much precedent for this form. Similar usage is in `conn/conn_cache_pool.c` has them at the top. In any case, @sulabhM can be the tiebreaker. At a minimum, if they stay, please alphabetize them in their little groups.

src/btree/bt_random.c
sueloverso: Eh, I don't care much but `WT_RETRY_MAX` is not OS related is not wildly different at all. It is used in eviction as well. All the place it's used it is simply as a "retry a few times" which is exactly what the comment below says. So I think their meanings are aligned.

src/btree/bt_random.c
keithbostic: Sorted in 2611383.

sulabhM: I don't like the idea of a separate header file. I don't really ever scroll the file from top to the bottom. I search for a function and scroll from there and it is useful to have related functionality clubbed together. I vote for leaving them as they are today. I have nothing against putting them together at the top, just that it would not help me in the way I search through the code. I believe most others do the same.

PR #4861, opened 2019-09-03
WT-4966 Fix valgrind detected memory leak from test/csuite/import/smoke.sh

sueloverso:


sueloverso:
@keithbostic Here's a simple workaround for valgrind. While the PR tests have run, I have also pointed the Jenkins valgrind job at this branch for testing as well.

test/csuite/import/smoke.sh
```diff
@@ -1,4 +1,8 @@
-#! /bin/sh
+#! /bin/dash
+# XXX - This should be /bin/sh but there is a bug in bash on many Linux systems that results
+# in a memory leak detection on immediate exit after a 'test' statement like below.
+# This is reported in https://bugzilla.redhat.com/show_bug.cgi?id=1746101
+# When that bug is fixed, this script should revert back on /bin/sh.
```
keithbostic: `/bin/dash` isn't standard on lots of systems, it would be nice to do something else (if it's easy to test, maybe just running a command like `date > /dev/null` before exiting would avoid the problem)?

Anyway, this is fine with me if there's nothing simpler.

sueloverso: I like this idea better. Fixed.

PR #4864, opened 2019-09-06
WT-5104 Add readonly flag to wt util

luke-pearson:


src/docs/command-line.dox
```diff
@@ -34,7 +36,10 @@ The \c wt tool supports several commands.  If configured in the underlying
 database, some commands will run recovery when opening the database.  If
 the user wants to force recovery on any command, use the \c -R option.
 In general, commands that modify the database or tables will run recovery
-by default and commands that only read data will not run recovery.
+by default and commands that only read data will not run recovery. It is
+recommended when attempting to diagnose a corrupt database the -r flag
+is supplied. This will prevent utility commands from writing to any of the
```
sueloverso: Minor wordsmithing. I suggest:
`It is recommended when attempting to diagnose a corrupt database, that the -r flag be used. This flag will open the connection read-only and prevent utility commands from writing...`

test/suite/test_backup03.py
```diff
@@ -95,8 +95,7 @@ def populate(self):
     # Compare the original and backed-up files using the wt dump command.
     def compare(self, uri):
         self.runWt(['dump', uri], outfilename='orig')
-        # Pass -R flag as we need to run recovery on the backup dir before calling dump.
-        self.runWt(['-h', self.dir, '-R', 'dump', uri], outfilename='backup')
+        self.runWt(['-h', self.dir, 'dump', uri], outfilename='backup')
```
sueloverso: Not related to this change, but I think it would be good to have unit tests (in the `test_util##.py` category) that tests using the `-r` flag correctly on some database, and using it with recovery or salvage and checking for the error message.

src/docs/command-line.dox
```diff
@@ -142,7 +147,7 @@ which can be re-loaded into a new table using the \c load command.
 See @subpage dump_formats for details of the dump file formats.
 
 @subsection util_dump_synopsis Synopsis
-<code>wt [-RVv] [-C config] [-E secretkey ] [-h directory] dump [-jrx] [-c checkpoint] [-f output] uri</code>
+<code>wt [-RrVv] [-C config] [-E secretkey ] [-h directory] dump [-jrx] [-c checkpoint] [-f output] uri</code>
```
sueloverso: Seeing all of these and the change above for `LRrVv` in the strings my question is: Should all of these synopsis lines have the `-L` added to them? Was that missed long ago?

luke-pearson: The command line documentation is definitely outdated as the whole thing is also missing -S. I'd prefer to not make refreshing the documentation part of this ticket and instead put it in in separate ticket.

test/suite/test_backup03.py
luke-pearson: Agreed, I'll create a follow-on ticket once we merge this change.

sueloverso:
I still think dropping core when diagnostics is enabled could be a surprising outcome. But if it becomes a problem we can visit it at a later date.

PR #4869, opened 2019-09-11
WT-5122 the sweep server can close handles the final checkpoint is in the process of using

keithbostic:
@agorrod, the interesting work is in 1231279 (fbb894c just moves the final checkpoint code out of the connection-close source and into the transaction subsystem source).

I moved the shutdown of the capacity, checkpoint and sweep servers to earlier in the shutdown process (it didn't seem useful to let them continue to run during shutdown), but that means we shut them down in two places rather than one, which is obviously more complicated. An alternative would be to not set the `WT_CONN_CLOSING_NO_MORE_OPENS` flag before doing the final checkpoint. That means the sweep server can race with the final checkpoint, but that's been true for a long time without any adverse effects.

### DIFF
--- a/src/conn/conn_api.c
+++ b/src/conn/conn_api.c
@@ -1010,7 +1010,6 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
     WT_SESSION *wt_session;
     WT_SESSION_IMPL *s, *session;
     uint32_t i;
-    const char *ckpt_cfg;
 
     conn = (WT_CONNECTION_IMPL *)wt_conn;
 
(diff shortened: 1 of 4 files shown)


keithbostic:
autobuild: retest this please

src/conn/conn_api.c
```diff
@@ -1074,47 +1073,24 @@ __conn_close(WT_CONNECTION *wt_conn, const char *config)
     WT_TRET(__wt_lsm_manager_destroy(session));
 
     /*
-     * After the async and LSM threads have exited, we shouldn't opening any more files.
+     * After the async and LSM threads have exited, we won't open more files for the application.
+     * However, the sweep server is still running and it can close file handles at the same time the
+     * final checkpoint is reviewing open data handles (forcing checkpoint to reopen handles). Shut
+     * down the sweep server and then flag the system should not open anything new.
      */
+    WT_TRET(__wt_sweep_destroy(session));
```
agorrod: @keithbostic  I've been sitting on the review of this because I'm nervous about this change. How confident are you that we don't need the sweep server to close handles after the checkpoint has finished? Maybe we can talk that through.

keithbostic: Happy to talk it through: my initial pov is sweep only does operations based on a timer, that is, it can't be depended on to close any handle at any particular time, which means that if it is required by checkpoint, there's a bug in there by definition.

agorrod: You are right, and if it failed we would notice.

PR #4873, opened 2019-09-17
WT-5128 Add script to run wtperf with XRay profiling

tetsuo-cpp:
This is a script to run `wtperf` with XRay profiling enabled and generate some interesting information using the logs. I'm invoking like so:
```
XRAY_BINARY=llvm-xray-8 FLAME_GRAPH_PATH=~/work/FlameGraph/flamegraph.pl bash ../../../bench/wtperf/runners/wtperf_xray.sh ../../../bench/wtperf/runners/small-btree.wtperf
```
`wtperf_account.txt`: A breakdown of the top 10 functions where the workload is spending most of its time.
```
Functions with latencies: 308
   funcid      count [      min,       med,       90p,       99p,       max]       sum  function
      258        504 [ 0.001050,  0.731710,  7.316153,  7.316191,  7.316230] 939.480654  os_mtx_cond.c:58:0: __wt_cond_wait_signal
      471         16 [ 5.624216, 88.403520, 88.404453, 88.404574, 88.404574] 752.231184  thread_group.c:17:0: __thread_run
       43          8 [87.809180, 87.809375, 87.809592, 87.809592, 87.809592] 702.474963  wtperf.c:501:0: worker
      898   20483363 [ 0.000011,  0.000030,  0.000031,  0.000042,  0.013131] 610.765654  cur_file.c:184:0: __curfile_search
      578   20483363 [ 0.000009,  0.000028,  0.000030,  0.000040,  0.013127] 582.309584  bt_cursor.c:509:0: __wt_btcur_search
      801   20483369 [ 0.000005,  0.000020,  0.000021,  0.000028,  0.013104] 412.661116  row_srch.c:213:0: __wt_row_search
      166        234 [ 0.000006,  0.731754,  0.731768,  0.731808,  0.731854] 147.538713  evict_lru.c:271:0: __wt_evict_thread_run
       21          1 [94.929063, 94.929063, 94.929063, 94.929063, 94.929063] 94.929063  wtperf.c:2543:0: main
       22          1 [94.689410, 94.689410, 94.689410, 94.689410, 94.689410] 94.689410  wtperf.c:2298:0: start_all_runs
       24          1 [94.689407, 94.689407, 94.689407, 94.689407, 94.689407] 94.689407  wtperf.c:2371:0: start_run
```
`wtperf_stack.txt`: The top 10 stack traces per thread. Here are a few stacks for a given thread.
```
Sum: 150608443030
lvl   function                                                            count              sum
#0    worker                                                                  1     263428775618
#1    __curfile_search                                                  2556560     229082313916
#2    __wt_btcur_search                                                 2556564     218432977048
#3    __wt_row_search                                                   2519298     150608443030

Sum: 738365164
lvl   function                                                            count              sum
#0    worker                                                                  1     263428775618
#1    __curfile_search                                                  2556560     229082313916
#2    __wt_btcur_search                                                 2556564     218432977048
#3    __wt_row_search                                                     37266       4124255246
#4    __wt_lex_compare_skip                                               21889        738365164
```
`wtperf_graph.svg`: A function call graph showing which functions call which. The edges are coloured and labelled to show what ratio of time is spent in each function.
Screen Shot 2019-09-17 at 6 52 07 pm
`wtperf_flame.svg`: A flame graph. You've all seen this before since we were generating these with `perf_events` data previously.

These commands are mostly copied from the documentation. I think it will take a bit of trial and error to see what sticks so this is just a starting point! @fedorova has mentioned the possibility of parsing the XRay logs into the operation tracking format or something that `t2` can understand.

### DIFF
--- a/bench/wtperf/runners/wtperf_xray.sh
+++ b/bench/wtperf/runners/wtperf_xray.sh
@@ -1,22 +1,22 @@
 #!/bin/bash
 
 if test "$#" -lt "1"; then
-    echo "$0: must specify wtperf test to run"
-    exit 1
+	echo "$0: must specify wtperf test to run"
+	exit 1
 fi
 
 # Check symbols to ensure we've compiled with XRay.
 objdump_out=$(objdump -h -j xray_instr_map ./wtperf)
 if test -z "$objdump_out"; then
-    echo "$0: wtperf not compiled with xray"
-    exit 1
+	echo "$0: wtperf not compiled with xray"
+	exit 1
 fi
 
 rm xray-log.wtperf.* \
-   wtperf_account.txt \
-   wtperf_stack.txt \
-   wtperf_graph.svg \
-   wtperf_flame.svg
+	wtperf_account.txt \
+	wtperf_stack.txt \
+	wtperf_graph.svg \
+	wtperf_flame.svg
 
 export XRAY_OPTIONS="patch_premain=true xray_mode=xray-basic verbosity=1"
 wtperf_out=$(./wtperf -O "$@")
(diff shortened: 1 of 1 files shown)


keithbostic:
@tetsuo-cpp, I'm probably the wrong guy to do a review, I mean, I know nothing at all about XRay profiling. If you just want someone to review the shell script, sure, happy to do that.

That said -- this looks very, very interesting to me, I'm seriously interested in playing with this. If there's a Wiki page on this, I'd like to know about it.

sueloverso:
Although I don't know anything about this, I will review it because this is really interesting and I'll probably have thoughts on the script itself. @fedorova should do the real review.

tetsuo-cpp:
@keithbostic 
Understood! There's no wiki page right now but it's on my todo list. I'll spend some time on it when I get back if we see potential value on this.

@sueloverso 
Thanks for deciding to review. Much appreciated!

sueloverso:
@tetsuo-cpp I pushed a new version of the script. Mostly I updated a few comments (I could type them there once as easily as typing them here for you to then type). There are a couple places I put `XXX` in the script for you to add a bit more information.

One thought I had, and feel free to ignore if it is too much of a pain to do it, is that if the user specifies a home directory, using that as the destination directory for the files generated by the script. I.e. if I said `wtperf_xray.sh eviction.wtperf -h WT_EVICT` then the files like `wtstack.txt` etc would end up in the `WT_EVICT` directory. I could definitely see running variations of configurations into different home directories and wanting to not overwrite all those files or remember to move them before running again.

agorrod:
@tetsuo-cpp is this good to go, or does it need something more?

tetsuo-cpp:
@agorrod Mostly but I had a few things I was planning to do. I need to use `getopt` to parse a `-h` flag in the script to tell it where to put the artifacts as per @sueloverso's suggestion.

Also I wanted to have a quick look at whether there is a simple way to constrain which functions/stack traces get analysed (what we discussed a few days ago).

tetsuo-cpp:
@agorrod @sueloverso I've added a `-h` option as suggested. There doesn't seem to be an easy way to constrain data to a set of code paths so I'm leaving this idea for the time being. This is ready for merging when you're happy.

sueloverso:
@tetsuo-cpp I think I wasn't entirely clear on my comment regarding the home directory. I hope it is an easy and minor fix. Basically, in an example like `wtperf_xray.sh eviction.wtperf -h WT_EVICT` the `-h WT_EVICT` is serving double-duty. I want it to be the home directory passed to `./wtperf` but also have the script use it as the destination for the xray files. In the script's current form I'd instead have to do:
`wtperf_xray.sh -h WT_EVICT eviction.wtperf -h WT_EVICT`
which is a bit cumbersome.

I think it would be fine/expected to have the usage specify concretely that `$2` must be `-h` if it is going to be used. The likely makes parsing easier and avoids the need for shifting args. So usage would be:
wtperf_xray.sh <wtperf-config-file> [-h output-directory] [other wtperf args]

and probably add a comment about `-h` serving both purposes clearly in the script.

@agorrod do you have any usability issues with overloading `-h` here?

agorrod:
I agree that overriding `-h` for something other than home isn't ideal - could we choose a different name for the option?

tetsuo-cpp:
@sueloverso How does this look? I think this does what you're describing. Apologies for the misunderstanding.

sueloverso:
@tetsuo-cpp that is nicely simplified and LGTM. I'll merge.

PR #4876, opened 2019-09-20
WT-5094 Switch to windows-64-vs2017-test distro and fix scons version

quchenhao:


lukech:
@quchenhao I assume you had a green evergreen patch build. Do you mind to share it?

PR #4877, opened 2019-09-20
WT-5093 Switch to rhel80-build distro for million-collections-test

quchenhao:


quchenhao:
evergreen retry

quchenhao:
evergreen retry

quchenhao:
https://evergreen.mongodb.com/task/wiredtiger_large_scale_test_million_collection_test_patch_90f3f9de951b4a51b2548ce27ee8ef98f5b4d849_5d8a70513627e01f73f75f17_19_09_24_19_37_10 The workaround to run with root does work.

test/evergreen.yml
```diff
@@ -1033,14 +1033,15 @@ tasks:
   - name: million-collection-test
     depends_on: []
     run_on:
-      - rhel62-large
+      - rhel80-build
```
lukech: Not related to the change directly - above 3 lines seem to be no-ops and could be cleaned up, as there's no dependent task and the build variant level has the same distro setting.

PR #4881, opened 2019-09-23
WT-5072 Consider updates from lookaside when checkpointing

sulabhM:


src/reconcile/rec_visibility.c
```diff
@@ -124,23 +223,80 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        default:
+            WT_ERR(__wt_illegal_value(session, page->type));
+        }
+
+        /* Open a lookaside cursor, position at the latest update for this key */
+        /* Remember to unlock */
+        __wt_las_cursor(session, &las_cursor, &session_flags);
+        las_cursor_open = true;
+        cache->las_reader = true;
+        //__wt_readlock(session, &cache->las_sweepwalk_lock);
```
sulabhM: Taking and releasing a lock here throws an error. Need to figure why

test/suite/test_las06.py
```diff
@@ -84,9 +84,16 @@ def test_las_reads_workload(self):
...
 
-        # Now the latest version will get written to the data file.
+        # Write a version of the data to disk.
+        self.conn.set_timestamp('stable_timestamp=' + timestamp_str(2))
         self.session.checkpoint()
 
+        # Check the checkpoint wrote the expected values.
+        cursor2 = self.session.open_cursor(uri, None, 'checkpoint=WiredTigerCheckpoint')
+        for key, value in cursor2:
+            self.assertEqual(value, value1)
```
sulabhM: Even though I checked and the code always picks the correct update (value1 = 'aaaa...') for the checkpoint, we only are not able to find all the 9999 keys here and some values end up being incorrect (value2 = 'bbbb....')

sulabhM:
Closing to create a new pull request based on these changes.

PR #4884, opened 2019-09-24
WT-5132 Fix buffer overflow caused by fscanf range check

quchenhao:


test/csuite/random_abort/main.c
```diff
@@ -72,11 +72,11 @@ static bool inmem;
...
 
-#define STR(X) #X
-#define XSTR(X) STR(X)
-
 #define MAX_VAL 4096
-#define STR_MAX_VAL XSTR(MAX_VAL)
+/*
+ * STR_MAX_VAL is set to MAX_VAL - 1 to account for the extra null character.
+ */
+#define STR_MAX_VAL "4095"
```
sulabhM: Do you think MAX_VAL_STR_LEN would be a better name for this variable?

sulabhM:
Consider running the sanitizer locally to check if the issue is gone.

PR #4890, opened 2019-09-29
WT-4677 Re-initialize prepared updates

tetsuo-cpp:
This PR is making it such that we initialise prepared updates into the update list if we happen to read them from the lookaside. This way, if we commit we can rewrite it into the lookaside with the new commit timestamp as the `start_ts` and if we rollback, we can simply avoid writing it back instead of having to search the lookaside to delete each prepared update.


src/cache/cache_las.c
```diff
@@ -1423,10 +1433,50 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
...
+            WT_ASSERT(session, cbt->compare == 0);
+            upd_entry = NULL;
+            switch (page->type) {
+            case WT_PAGE_COL_FIX:
+            case WT_PAGE_COL_VAR:
+                WT_ASSERT(session, cbt->ins != NULL);
+                upd_entry = &cbt->ins->upd;
+                break;
+            case WT_PAGE_ROW_LEAF:
+                upd_entry =
```
tetsuo-cpp: @kommiharibabu My understanding is that col/row instantiation has already happened so I can rely on these here. Does this look right to you?

src/cache/cache_las.c
```diff
@@ -1423,10 +1433,50 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
...
+                  cbt->ins == NULL ? &page->modify->mod_row_update[cbt->slot] : &cbt->ins->upd;
+                break;
+            }
+
+            WT_ASSERT(session, upd_entry != NULL);
```
tetsuo-cpp: My initial design was to use `__wt_row_modify` and `__wt_col_modify` for this. However I felt that those functions were getting a bit overloaded with different cases so I decided to just inline the relevant code here.

I think that this is a pretty specific case without too many exceptions (famous last words) so this seemed reasonable to me. Let me know if you disagree with what I've done here.

src/cache/cache_las.c
```diff
@@ -1423,10 +1433,50 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
...
+             */
+            tmp_upd = upd;
+            WT_ERR(__wt_update_serial(session, page, upd_entry, &tmp_upd, upd_size, false));
+
+            /*
+             * After reading in the prepared update, we need to delete it from lookaside. If it gets
+             * committed, the timestamp in the las key may differ so it's easier if we just get rid
+             * of it now and rewrite on eviction/commit/rollback.
+             */
+            WT_ERR(cursor->remove(cursor));
```
tetsuo-cpp: We're in a pickle if we fail here since we've got the prepared update in memory and will rewrite on eviction and potentially duplicate the update with a different timestamp. I'm thinking this should maybe be a fatal error?

src/cache/cache_las.c
kommiharibabu: Yes, the page has instantiated with all the keys by this time.

src/cache/cache_las.c
kommiharibabu: I agree that failure in the remove is a rare scenario. Raising a fatal error could be good.

src/cache/cache_las.c
```diff
@@ -1264,41 +1264,45 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
  *     the record and return to the caller.
  */
 int
-__wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE **updp)
+__wt_find_lookaside_upd(
+  WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE **updp, bool allow_prepare)
```
kommiharibabu: Can you add a comment in explaining the new parameter and it's usage or when this parameter is passed as true or etc.

src/cache/cache_las.c
```diff
@@ -825,7 +825,7 @@ __wt_las_cursor_position(WT_SESSION_IMPL *session, WT_CURSOR *cursor, uint32_t b
      */
     if (btree_id == 0) {
         WT_RET(cursor->reset(cursor));
-        return (cursor->next(cursor));
+        return (cursor->prev(cursor));
```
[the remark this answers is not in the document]
tetsuo-cpp: I think `reset` means it has no position. And you can use `next` to get to the beginning and `prev` to get to the end.

src/btree/bt_cursor.c
```diff
@@ -465,8 +465,9 @@ __wt_btcur_search_uncommitted(WT_CURSOR_BTREE *cbt, WT_UPDATE **updp)
 {
     WT_BTREE *btree;
     WT_CURSOR *cursor;
+    WT_DECL_RET;
     WT_SESSION_IMPL *session;
-    WT_UPDATE *upd;
+    WT_UPDATE *upd, *tmp_upd;
```
sulabhM: Can alphabetize the variable declarations

src/btree/bt_cursor.c
```diff
@@ -481,19 +482,32 @@ __wt_btcur_search_uncommitted(WT_CURSOR_BTREE *cbt, WT_UPDATE **updp)
...
-    if (cbt->compare != 0)
-        return (0);
+    if (cbt->compare == 0) {
+        /*
+         * Get the uncommitted update from the cursor. For column store there will be always a
+         * insert structure for updates irrespective of fixed length or variable length.
+         */
+        if (cbt->ins != NULL)
+            upd = cbt->ins->upd;
+        else if (cbt->btree->type == BTREE_ROW)
```
sulabhM: Is there any reason to remove the assert and add the if conditions. Optionally consider the following alternative:
```
        if (cbt->ins != NULL)
            upd = cbt->ins->upd;
        else {
            WT_ASSERT(session, cbt->btree->type == BTREE_ROW && cbt->ref->page->modify != NULL &&
              cbt->ref->page->modify->mod_row_update != NULL);
            upd = cbt->ref->page->modify->mod_row_update[cbt->slot];
        }
```

src/btree/bt_cursor.c
```diff
@@ -481,19 +482,32 @@ __wt_btcur_search_uncommitted(WT_CURSOR_BTREE *cbt, WT_UPDATE **updp)
...
+     * than those in the update list.
      */
-    if (cbt->ins != NULL)
-        upd = cbt->ins->upd;
-    else if (cbt->btree->type == BTREE_ROW) {
-        WT_ASSERT(session, cbt->btree->type == BTREE_ROW && cbt->ref->page->modify != NULL &&
-            cbt->ref->page->modify->mod_row_update != NULL);
-        upd = cbt->ref->page->modify->mod_row_update[cbt->slot];
+    if (__wt_page_las_active(session, cbt->ref) && cbt->ref->page_las->has_las &&
+      cbt->ref->page_las->has_prepares) {
```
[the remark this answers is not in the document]
sulabhM: I will say move `has_las` check inside `__wt_page_las_active()`, I think it used to be there at some point. Also, I think you need to explicitly perform `has_prepares` here to be sure that 1. there is las 2. there are prepared

src/btree/bt_cursor.c
```diff
@@ -481,19 +482,32 @@ __wt_btcur_search_uncommitted(WT_CURSOR_BTREE *cbt, WT_UPDATE **updp)
...
+        ret = __wt_find_lookaside_upd(session, cbt, &tmp_upd, true);
+        WT_RET_NOTFOUND_OK(ret);
+        if (ret == 0) {
+            if (upd == NULL || upd->start_ts < tmp_upd->start_ts)
+                upd = tmp_upd;
```
sulabhM: Are we assuming that the update we found is definitely going to be in the prepared state? I think it is a fair assumption as we "known" that for this key we have a prepared update and if the update we find in lookaside is the newest it must be in the prepared state. Let me know if my understanding is wrong.

I think we need to assert our assumption though: Once we have found a newer update in the lookaside, we should assert that this update is in the prepared state.

src/cache/cache_las.c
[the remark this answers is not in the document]
sulabhM: I think @tetsuo-cpp is correct.

src/cache/cache_las.c
```diff
@@ -1423,10 +1434,50 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
...
         upd->start_ts = las_timestamp;
         upd->prepare_state = prepare_state;
-        /*
-         * Mark this update as external and to be discarded when not needed.
-         */
-        upd->ext = 1;
+
+        if (prepare_state == WT_PREPARE_INPROGRESS) {
+            WT_ASSERT(session, !modify);
+            WT_ASSERT(session, cbt->compare == 0);
```
sulabhM: Can you put a comment why we are asserting `cbt->compare == 0`

src/cache/cache_las.c
```diff
@@ -1423,10 +1434,50 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
         upd->durable_ts = durable_timestamp;
         upd->start_ts = las_timestamp;
         upd->prepare_state = prepare_state;
-        /*
-         * Mark this update as external and to be discarded when not needed.
-         */
-        upd->ext = 1;
+
+        if (prepare_state == WT_PREPARE_INPROGRESS) {
```
sulabhM: Can you put a comment regarding what this block of code is expected to do

src/cache/cache_las.c
sulabhM: I suggest getting Keith's opinion on this.

src/cache/cache_las.c
sulabhM: Maybe sweep can detect such left-over prepared updates. If there is a committed update with the same or a newer timestamp than a prepared-update in lookaside, that prepare can be removed?

src/btree/bt_cursor.c
tetsuo-cpp: fd82c8e

Your understanding is correct. There is a comment indicating above that it's specifically for the prepare case. I've added an assertion to check that if we end up using the lookaside update that it truly is a prepare.

src/cache/cache_las.c
tetsuo-cpp: fd82c8e

I've added a comment and moved the assertion to the top of the function since it actually applies everywhere.

src/btree/bt_cursor.c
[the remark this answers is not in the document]
tetsuo-cpp: @kommiharibabu That function is not enough since it can pass for reasons that have nothing to do with prepare. I've condensed the check down to:
```
cbt->ref->page_las != NULL && cbt->ref->page_las->has_prepares
```
@sulabhM That variable was never part of the check. I added it to replace `las_pageid` since we were comparing it to 0 as a proxy for checking the existence of lookaside entries.

src/cache/cache_las.c
tetsuo-cpp: fd82c8e

Added to the function documentation.

src/btree/bt_cursor.c
tetsuo-cpp: fd82c8e

src/cache/cache_las.c
tetsuo-cpp: fd82c8e

The relevant comment was above the `remove` call. I've moved most of that comment to above the block.

src/btree/bt_cursor.c
tetsuo-cpp: 1449dd2

Thanks for picking that up. Reverted a few of these changes.

This was from before I knew of @kommiharibabu's change so it was accounting for when there's only a prepared update for a given key in lookaside but nothing in the update list. As I understand it, that's no longer possible.

src/cache/cache_las.c
```diff
@@ -1336,7 +1348,8 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
      * timestamp is part of the key, our cursor needs to go from the newest record (further in the
      * las) to the oldest (earlier in the las) for a given key.
      */
-    for (ret = __wt_las_cursor_position(session, cursor, las_btree_id, key, txn->read_timestamp);
+    for (ret = __wt_las_cursor_position(session, cursor, las_btree_id, key,
+           allow_prepare ? txn->prepare_timestamp : txn->read_timestamp);
```
keithbostic: This is harder to read and doesn't gain us space, I'd pull the function call out:
```
ret = __wt_las_cursor_position(session, cursor, las_btree_id, key,
  allow_prepare ? txn->prepare_timestamp : txn->read_timestamp;
for (;  ret == 0; ret = cursor->prev(cursor)) {
```

src/cache/cache_las.c
keithbostic: If this ever happens, things have gone pretty far south. I wouldn't bother trying to keep going.

src/cache/cache_las.c
```diff
@@ -1261,44 +1261,56 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
     uint32_t las_btree_id, session_flags;
     uint8_t prepare_state, _prepare_state, *p, upd_type;
     u_int i;
     int cmp;
-    bool sweep_locked;
+    bool modify, sweep_locked;
 
     WT_UNUSED(_las_timestamp);
     WT_UNUSED(_las_txnid);
+    WT_UNUSED(modify);
```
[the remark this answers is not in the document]
tetsuo-cpp: It is only used for an assertion. So release builds complain that it is written to but never read.

src/cache/cache_las.c
```diff
@@ -1423,10 +1442,49 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
         upd->durable_ts = durable_timestamp;
         upd->start_ts = las_timestamp;
         upd->prepare_state = prepare_state;
+
```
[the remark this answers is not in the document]
tetsuo-cpp: 3c5ff13

Moved it up a bit since it also applies to the diagnostic code.

src/cache/cache_las.c
tetsuo-cpp: 3c5ff13

Panicking here.

src/cache/cache_las.c
tetsuo-cpp: 721875b

src/cache/cache_las.c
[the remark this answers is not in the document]
tetsuo-cpp: 721875b

src/cache/cache_las.c
```diff
@@ -1423,10 +1442,49 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
...
-         * Mark this update as external and to be discarded when not needed.
+         * When we find a prepared update in lookaside, we should add it to our update list and
+         * subsequently delete the corresponding lookaside entry. If it gets committed, the
+         * timestamp in the las key may differ so it's easier if we get rid of it now and rewrite
+         * the entry on eviction/commit/rollback.
          */
-        upd->ext = 1;
+        if (prepare_state == WT_PREPARE_INPROGRESS) {
+            WT_ASSERT(session, !modify);
+            upd_entry = NULL;
```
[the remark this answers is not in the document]
tetsuo-cpp: My compiler complains about it since there are other possible values for `type`.

src/cache/cache_las.c
tetsuo-cpp: @keithbostic Ok, I'm using `__wt_row_modify` and `__wt_col_modify`. Let me know how this looks.

src/cache/cache_las.c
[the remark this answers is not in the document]
keithbostic: Got it.

`WT_UNUSED` is likely to be insufficient for some compilers/static test engines, try `WT_NOT_READ`.

src/btree/row_modify.c
```diff
@@ -64,6 +64,17 @@ __wt_row_modify(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, const WT_ITEM *k
...
+    /*
+     * We should have EITHER:
+     * - A full update list to instantiate with.
+     * - An update to append the existing update list with.
+     * - A key/value pair to create an update with and append to the update list.
+     *
+     * A full update list is distinguished from an update by checking whether it has any "next"
+     * update.
+     */
+    WT_ASSERT(session, (value != NULL) ^ (upd_arg != NULL));
```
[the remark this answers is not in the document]
tetsuo-cpp: 7cda52d

Not quite. I want to ensure that only one has been provided (not zero, or both).
I've removed the xor and switched to:
```
(value == NULL) != (upd_arg == NULL)
```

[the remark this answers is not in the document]
keithbostic: How about `(value == NULL && upd_arg != NULL) || (value != NULL && upd_arg == NULL)`?

The performance isn't going to matter and it's simpler?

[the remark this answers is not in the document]
tetsuo-cpp: 04a4529

src/cache/cache_las.c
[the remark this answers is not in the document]
tetsuo-cpp: 04a4529

PR #4894, opened 2019-10-02
WT-5146 Fix record miss for variable length columnar table

kommiharibabu:
If the ref contains any LAS page, the scan searches the LAS file
for the satisfied record before returning the disk image record
to the application. During this process, cursor btree tmp buffer
is used to store the key information that is used to search the
LAS file, but using this cursor btree tmp, corrupts the memory
that is already presents in the buffer that holds the records.
Due to this, the key itself is returned to the application instead
of the actual record.

Use another temp buffer las to store the key information that is
required for scanning the LAS file instead of using the existing
tmp buffer.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -1233,14 +1233,14 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
     WT_DECL_ITEM(las_key);
     WT_DECL_ITEM(las_value);
     WT_DECL_RET;
-    WT_ITEM *key;
+    WT_ITEM key;
     WT_TXN *txn;
     WT_UPDATE *upd, *list[WT_MODIFY_ARRAY_SIZE], **listp;
     wt_timestamp_t durable_timestamp, _durable_timestamp, las_timestamp, _las_timestamp;
     size_t allocated_bytes, incr;
     uint64_t las_txnid, _las_txnid;
     uint32_t las_btree_id, session_flags;
-    uint8_t prepare_state, _prepare_state, *p, upd_type;
+    uint8_t prepare_state, _prepare_state, *p, recno_key[WT_INTPACK64_MAXSIZE], upd_type;
     u_int i;
     int cmp;
     bool sweep_locked;
@@ -1252,7 +1252,7 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
 
     cache = S2C(session)->cache;
     cursor = NULL;
-    key = NULL;
+    WT_CLEAR(key);
     upd = NULL;
     listp = list;
     txn = &session->txn;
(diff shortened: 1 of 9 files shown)


src/cache/cache_las.c
```diff
@@ -1269,11 +1269,11 @@ __wt_find_lookaside_upd(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDAT
...
     case BTREE_COL_VAR:
-        WT_RET(__wt_buf_grow(session, cbt->tmp, WT_INTPACK64_MAXSIZE));
-        p = cbt->tmp->mem;
+        WT_RET(__wt_buf_grow(session, cbt->las, WT_INTPACK64_MAXSIZE));
+        p = cbt->las->mem;
         WT_RET(__wt_vpack_uint(&p, 0, cbt->recno));
-        cbt->tmp->size = WT_PTRDIFF(p, cbt->tmp->mem);
-        key = cbt->tmp;
+        cbt->las->size = WT_PTRDIFF(p, cbt->las->mem);
+        key = cbt->las;
```
keithbostic: There are a lot of cursors in the system (thousands and thousands), we don't want to grow that structure. We only need a small buffer here. I think something like this would be sufficient:
```
*** ../cache/cache_las.c        Wed Oct  2 01:30:49 2019
--- -   Wed Oct  2 01:36:35 2019
***************  
*** 1233,1246 ****
      WT_DECL_ITEM(las_key);
      WT_DECL_ITEM(las_value);
      WT_DECL_RET;
!     WT_ITEM *key;
      WT_TXN *txn; 
      WT_UPDATE *upd, *list[WT_MODIFY_ARRAY_SIZE], **listp;
      wt_timestamp_t durable_timestamp, _durable_timestamp, las_timestamp, _las_timestamp;
      size_t allocated_bytes, incr;
      uint64_t las_txnid, _las_txnid;
      uint32_t las_btree_id, session_flags;
!     uint8_t prepare_state, _prepare_state, *p, upd_type;
      u_int i;
      int cmp;
      bool sweep_locked;
--- 1233,1246 ----
      WT_DECL_ITEM(las_key);
      WT_DECL_ITEM(las_value);
      WT_DECL_RET;
!     WT_ITEM key;
      WT_TXN *txn;
      WT_UPDATE *upd, *list[WT_MODIFY_ARRAY_SIZE], **listp;
      wt_timestamp_t durable_timestamp, _durable_timestamp, las_timestamp, _las_timestamp;
      size_t allocated_bytes, incr;
      uint64_t las_txnid, _las_txnid;
      uint32_t las_btree_id, session_flags;
!     uint8_t prepare_state, _prepare_state, *p, recno_key[WT_INTPACK64_MAXSIZE], upd_type;
      u_int i;
      int cmp;
      bool sweep_locked;
***************
*** 1269,1279 ****
          break;
      case BTREE_COL_FIX:
      case BTREE_COL_VAR:
!         WT_RET(__wt_buf_grow(session, cbt->las, WT_INTPACK64_MAXSIZE));
!         p = cbt->las->mem;
          WT_RET(__wt_vpack_uint(&p, 0, cbt->recno));
!         cbt->las->size = WT_PTRDIFF(p, cbt->las->mem);
!         key = cbt->las;
      }

      /* Allocate buffers for the lookaside key/value. */
--- 1269,1278 ----
          break;
      case BTREE_COL_FIX:
      case BTREE_COL_VAR:
!         p = recno_key;
          WT_RET(__wt_vpack_uint(&p, 0, cbt->recno));
!         key.data = recno_key;
!         key.size = WT_PTRDIFF(p, recno_key);
      }

      /* Allocate buffers for the lookaside key/value. */
```
That's untested, but I think it's close.

kommiharibabu: Thanks for the suggestion, yes it is a very good improvement. I changed the code to use the local variable as you suggested.

keithbostic:
@kommiharibabu, I just pushed a change in 35dfaf6, changing my suggested fix.

Basically, there's nothing wrong with what you & I did, but I hate ever having two copies of allocated memory (the underlying key found in the cursor), I worry it will lead to bugs in the future somehow if we decide to free `key` or something else. And, well, it's a tiny bit faster for row-store, not that it's going to matter. :)

Feel free to just discard that commit if you don't like it.

kommiharibabu:
@keithbostic, Thanks for the changes. That commit makes the fix better.

PR #4900, opened 2019-10-03
WT-5141 Handle multiple modifies for the same key on lookaside eviction

tetsuo-cpp:
As it stands, if we open a transaction and apply multiple modifies at the same timestamp, lookaside eviction will cause all but one of these modifies to be lost. This is because they occupy the same key in lookaside (`btree_id, key, timestamp, txnid`).

This PR adds some logic to detect this situation and squash the modifies into a single standard update upon lookaside eviction.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -792,6 +792,14 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
             ++insert_cnt;
             if (upd->prepare_state == WT_PREPARE_INPROGRESS)
                 ++prepared_insert_cnt;
+
+            /*
+             * If we encounter subsequent updates with the same timestamp, skip them to avoid
+             * clobbering the entry that we just wrote to lookaside. Only the last thing in the
+             * update list for that timestamp matters anyway.
+             */
+            while (upd->next != NULL && upd->start_ts == upd->next->start_ts)
+                upd = upd->next;
         } while ((upd = upd->next) != NULL);
     }
 
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -543,6 +543,46 @@ __las_insert_updates_verbose(WT_SESSION_IMPL *session, WT_BTREE *btree, WT_MULTI
 }
 
 /*
+ * __las_squash_modifies --
+ *     Squash multiple modify operations for the same key and timestamp into a standard update and
+ *     insert it into lookaside. This is necessary since multiple updates on the same key/timestamp
+ *     and transaction will clobber each other in the lookaside which is problematic for modifies.
+ */
+static int
+__las_squash_modifies(WT_SESSION_IMPL *session, WT_CURSOR *cursor, WT_UPDATE **updp)
+{
+    WT_DECL_RET;
+    WT_ITEM las_value;
+    WT_UPDATE *modify_list[100], *next_upd, *start_upd, *upd;
+    u_int i;
+
+    WT_CLEAR(las_value);
+    start_upd = upd = *updp;
+    next_upd = start_upd->next;
+    i = 0;
+
+    while (upd->type == WT_UPDATE_MODIFY) {
+        WT_ASSERT(session, i < 100);
+        modify_list[i++] = upd;
+        upd = upd->next;
+        if (start_upd->start_ts == upd->start_ts && start_upd->txnid == upd->txnid)
+            next_upd = upd;
+        WT_ASSERT(session, upd != NULL);
+    }
+    WT_ERR(__wt_buf_set(session, &las_value, upd->data, upd->size));
+    while (i > 0)
+        WT_ERR(__wt_modify_apply_item(session, &las_value, modify_list[--i]->data, false));
+    cursor->set_value(
+      cursor, start_upd->durable_ts, start_upd->prepare_state, WT_UPDATE_STANDARD, &las_value);
+    WT_ERR(cursor->insert(cursor));
+    *updp = next_upd;
+
+err:
+    __wt_buf_free(session, &las_value);
+    return (ret);
+}
+
+/*
  * __wt_las_insert_updates --
  *     Copy one set of saved updates into the database's lookaside table.
  */
@@ -558,7 +598,7 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
     WT_SAVE_UPD *list;
     WT_SESSION_IMPL *session;
     WT_TXN_ISOLATION saved_isolation;
-    WT_UPDATE *first_upd, *upd;
+    WT_UPDATE *first_upd, *tmp_upd, *upd;
     wt_off_t las_size;
     uint64_t insert_cnt, max_las_size, prepared_insert_cnt;
     uint32_t birthmarks_cnt, btree_id, i, slot;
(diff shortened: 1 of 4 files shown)


src/cache/cache_las.c
```diff
@@ -699,13 +770,36 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
             ++insert_cnt;
             if (upd->prepare_state == WT_PREPARE_INPROGRESS)
                 ++prepared_insert_cnt;
+
+            /*
+             * If we encounter subsequent updates with the same timestamp, skip them to avoid
+             * clobbering the entry that we just wrote to lookaside. Only the last thing in the
+             * update list for that timestamp matters anyway.
+             */
+            while (upd->next != NULL && upd->start_ts == upd->next->start_ts)
```
tetsuo-cpp: I realized when working on this change that a variant of this problem can occur with regular updates. This problem is much easier to deal with since we can just ignore subsequent updates instead of squashing them.

kommiharibabu: I agree that skipping them is good. Don't we need to consider the transaction id's also as part of the check? Otherwise only once LAS record is inserted with one transaction id, the may not be visible to some ongoing transactions?

src/cache/cache_las.c
```diff
@@ -699,13 +770,36 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
+             * If we encounter a modify with preceding updates that have the same timestamp and
+             * transaction id, they need to be squashed into a single lookaside entry to avoid
+             * displacing each other.
+             */
+            if (upd->type == WT_UPDATE_MODIFY && upd->next != NULL &&
+              upd->start_ts == upd->next->start_ts && upd->txnid == upd->next->txnid) {
+                tmp_upd = upd;
+                WT_ERR(__las_squash_modifies(session, cursor, &upd));
+                if (tmp_upd->prepare_state == WT_PREPARE_INPROGRESS)
+                    ++prepared_insert_cnt;
```
keithbostic: Why do we need `tmp_upd`, can't we do the increment of `prepared_insert_cnt` before calling `__las_squash_modifies`?

src/cache/cache_las.c
keithbostic: I agree with @kommiharibabu, I think transaction ID is relevant.

keithbostic: Around line 860, there's a line of code `WT_UNUSED(first_upd);` that no longer seems necessary?

tetsuo-cpp: 65e671e

src/cache/cache_las.c
```diff
@@ -542,6 +542,77 @@ __las_insert_updates_verbose(WT_SESSION_IMPL *session, WT_BTREE *btree, WT_MULTI
...
+static int
+__las_squash_modifies(WT_SESSION_IMPL *session, WT_CURSOR *cursor, WT_UPDATE **updp)
+{
+    WT_DECL_RET;
+    WT_ITEM las_value;
+    WT_UPDATE *list[WT_MODIFY_ARRAY_SIZE], **listp, *next_upd, *start_upd, *upd;
+    size_t allocated_bytes;
+    u_int i;
+
+    WT_CLEAR(las_value);
```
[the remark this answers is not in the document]
tetsuo-cpp: 65e671e

src/cache/cache_las.c
```diff
@@ -542,6 +542,77 @@ __las_insert_updates_verbose(WT_SESSION_IMPL *session, WT_BTREE *btree, WT_MULTI
...
+         * modifies for a given timestamp and transaction id. So we want to resume insertions right
+         * where a new timestamp and transaction id pairing begins.
+         */
+        if (start_upd->start_ts == upd->start_ts && start_upd->txnid == upd->txnid)
+            next_upd = upd;
+        /*
+         * If we've spotted a modify, there should be a standard update later in the update list. If
+         * we hit the end and still haven't found one, something is not right.
+         */
+        WT_ASSERT(session, upd != NULL);
```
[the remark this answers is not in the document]
tetsuo-cpp: 65e671e

Using `WT_PANIC_ERR` here now.

src/cache/cache_las.c
```diff
@@ -542,6 +542,77 @@ __las_insert_updates_verbose(WT_SESSION_IMPL *session, WT_BTREE *btree, WT_MULTI
...
+    WT_ITEM las_value;
+    WT_UPDATE *list[WT_MODIFY_ARRAY_SIZE], **listp, *next_upd, *start_upd, *upd;
+    size_t allocated_bytes;
+    u_int i;
+
+    WT_CLEAR(las_value);
+    listp = list;
+    start_upd = upd = *updp;
+    next_upd = start_upd->next;
+    allocated_bytes = i = 0;
```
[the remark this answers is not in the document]
tetsuo-cpp: 65e671e

src/cache/cache_las.c
tetsuo-cpp: If we fail to squash, I don't want to increment `prepared_insert_cnt`.

src/cache/cache_las.c
```diff
@@ -542,6 +542,77 @@ __las_insert_updates_verbose(WT_SESSION_IMPL *session, WT_BTREE *btree, WT_MULTI
...
+/*
+ * When threads race modifying a record, we can end up with more than the max
+ * number of modifications in an update list.
+ *
+ * Leave space for double the maximum number on the stack as a best attempt to
+ * avoid heap allocation.
+ *
+ * TODO: Look into reusing code from __wt_value_return_upd.
+ */
+#define WT_MODIFY_ARRAY_SIZE (WT_MAX_MODIFY_UPDATE * 2)
```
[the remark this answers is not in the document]
tetsuo-cpp: No reason in particular. It wasn't visible in `cache_las.c` and I chose double since that seemed less arbitrary than 10 more.

I can either change them to be the same or I could put the `#define` in `btmem.h` where it's visible to everyone.

src/cache/cache_las.c
tetsuo-cpp: Since these updates are remaining in memory, the lookaside content only matters if we evict and then subsequently read the page back into the cache right? In that case, I don't see how it's possible to have an ongoing transaction from before the eviction.

src/cache/cache_las.c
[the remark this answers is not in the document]
keithbostic: We should add `WT_PANIC_ASSERT`.

src/cache/cache_las.c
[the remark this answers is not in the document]
keithbostic: Both seem pretty arbitrary to me. :)

I guess I have a mild preference for `btmem.h`, next to the `WT_MAX_MODIFY_UPDATE` define.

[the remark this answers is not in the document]
keithbostic: There was a messed-up comment in `btmem.h` I wanted to fix, so I pushed that change in 6ea4acbc6.

src/cache/cache_las.c
keithbostic: Under what conditions will we fail to squash?

I think we've decided there is something to squash before we call `__las_squash_modifies`, right?

And on error we'll ignore `prepared_insert_cnt`?

Even more simply, could we yank the increment of `prepared_insert_cnt` above the clause that tests for modifies? I think it applies to all clauses after the birthmark test?

```
/* NEEDS A COMMENT */
if (upd->prepare_state == WT_PREPARE_INPROGRESS)
    ++prepared_insert_cnt;

/*  
 * If we encounter a modify with preceding updates that have the same timestamp and 
 * transaction id, they need to be squashed into a single lookaside entry to avoid
 * displacing each other.
 */  
if (upd->type == WT_UPDATE_MODIFY && upd->next != NULL &&
  upd->start_ts == upd->next->start_ts && upd->txnid == upd->next->txnid) {
    WT_ERR(__las_squash_modifies(session, cursor, &upd));
    ++insert_cnt;
    continue;
}   

cursor->set_value(cursor, upd->durable_ts, upd->prepare_state, upd->type, &las_value);

/* Using insert so we don't keep the page pinned longer than necessary. */
WT_ERR(cursor->insert(cursor));
++insert_cnt;
```

src/cache/cache_las.c
keithbostic: I'm not following, why can't transactional readers span page eviction?

But, more simply, what if we're running w/o timestamps, this will discard all updates because the timestamps will all be `WT_TS_NONE`, won't they?

src/cache/cache_las.c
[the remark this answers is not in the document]
tetsuo-cpp: Thanks!

src/cache/cache_las.c
tetsuo-cpp: 189dfc2

My mistake. I missed that `prepared_insert_cnt` was ignored in the error case.

src/cache/cache_las.c
tetsuo-cpp: 6253dde

PR #4902, opened 2019-10-04
WT-4685 Add new statistics for relevant history in cache

tetsuo-cpp:
This PR is adding some statistics to track some of the new functionality in our relevant history branch.


dist/stat_data.py
```diff
@@ -282,8 +282,13 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_lookaside_insert', 'cache overflow table insert calls'),
     CacheStat('cache_lookaside_ondisk', 'cache overflow table on-disk size', 'no_clear,no_scale,size'),
     CacheStat('cache_lookaside_ondisk_max', 'cache overflow table max on-disk size', 'no_clear,no_scale,size'),
+    CacheStat('cache_lookaside_read', 'cache overflow table update reads'),
```
tetsuo-cpp: I've renamed this to separate it from the rest of the `cache_read_<blah>` stats. The meaning of this stat has changed because it's no longer tracking when we read a page into cache which has lookaside content. Instead, we're counting how many updates we had to reach into lookaside for.

dist/stat_data.py
```diff
@@ -282,8 +282,13 @@ def __init__(self, name, desc, flags=''):
...
+    CacheStat('cache_lookaside_read_miss', 'cache overflow table update reads missed'),
```
tetsuo-cpp: This is tracking how often we call the `__wt_find_lookaside_upd` and don't get anything useful. Even though we're searching the las relatively efficiently it will still be expensive.

dist/stat_data.py
```diff
@@ -282,8 +282,13 @@ def __init__(self, name, desc, flags=''):
...
+    CacheStat('cache_lookaside_read_squash', 'cache overflow table update reads requiring modify squashing'),
```
tetsuo-cpp: This is tracking how often we have to read multiple entries from lookaside and squash them into a value. This happens to any modify operation that gets lookaside evicted so this shouldn't be that uncommon.

dist/stat_data.py
```diff
@@ -282,8 +282,13 @@ def __init__(self, name, desc, flags=''):
...
+    CacheStat('cache_lookaside_read_wasted', 'cache overflow table update reads wasted'),
```
tetsuo-cpp: This is tracking how often we read an update from lookaside only to find that the in-memory candidate is more recent (meaning that it is wasted work).

dist/stat_data.py
```diff
@@ -282,8 +282,13 @@ def __init__(self, name, desc, flags=''):
...
     CacheStat('cache_lookaside_remove', 'cache overflow table remove calls'),
     CacheStat('cache_lookaside_score', 'cache overflow score', 'no_clear,no_scale'),
+    CacheStat('cache_lookaside_write_squash', 'cache overflow table update writes requiring modify squashing'),
```
tetsuo-cpp: This is tracking how often we have to squash modifies on lookaside eviction to avoid displacing each other in the case where they have the same timestamp and transaction id.

src/cache/cache_las.c
```diff
@@ -1534,9 +1536,16 @@ __wt_find_lookaside_upd(
...
-    if (upd == NULL && ret == 0)
-        ret = WT_NOTFOUND;
+    if (ret == 0) {
+        /* Couldn't find a record. */
+        if (upd == NULL) {
+            ret = WT_NOTFOUND;
+            WT_STAT_CONN_INCR(session, cache_lookaside_read_miss);
+        } else {
+            WT_STAT_CONN_INCR(session, cache_lookaside_read);
+            WT_STAT_DATA_INCR(session, cache_lookaside_read);
```
tetsuo-cpp: I heard that we should add data source statistics sparingly since they occupy memory for every `dhandle` in the system. This stat is easily the most important so I added this one but I wasn't sure about the others.

PR #4903, opened 2019-10-04
WT-4681 Fix the las sweep to clean the obsolete records

kommiharibabu:
The Sweep of LAS works by scanning the cursor from start of the LAS
till the maximum record that is found before the scan start. The records
with las_timestamp that is less than oldest_timestamp are considered
as obsolete. The identified obsolete records are not removed only when
the next record of the scan is satisfies the following conditions.
	- The next LAS record is from a different key.
	- The next LAS record is not of UPDATE MODIFY type.

In case if the next record is UPDATE MODIFY record type, the search
continues till the next record doesn't match the key or not of type
UPDATE MODIFY. If any one of the record in update chain is not obsolete
the removal is skipped.


src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+                WT_ERR(
+                  cursor->get_key(cursor, &las_btree_id, &las_key, &las_timestamp, &las_txnid));
+                WT_ERR(cursor->get_value(
+                  cursor, &durable_timestamp, &prepare_state, &upd_type, &las_value));
+                ++remove_cnt;
+            }
+            skip_update_modify_cnt = 0;
+        }
 
+        /* Don't remove the non obsolete entry. */
```
kommiharibabu: Is this check is enough for detecting the record is obsolete? And also is the `oldest_timestamp` is available even if it not set by the application? or a check for its valid value is required?

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+              memcmp(saved_key->data, las_key.data, las_key.size) != 0) {
+                remove_entry = true;
+                saved_btree_id = las_btree_id;
+                WT_ERR(__wt_buf_set(session, saved_key, las_key.data, las_key.size));
+            } else {
+                if (upd_type != WT_UPDATE_MODIFY)
+                    remove_entry = true;
+                else
+                    skip_update_modify_cnt++;
+            }
```
keithbostic: I'd write
```
if (prev_entry_can_remove) {
    if (las_btree_id != saved_btree_id || saved_key->size != las_key.size ||
      memcmp(saved_key->data, las_key.data, las_key.size) != 0) {
        remove_entry = true;
        saved_btree_id = las_btree_id;
        WT_ERR(__wt_buf_set(session, saved_key, las_key.data, las_key.size));
    } else {
        if (upd_type != WT_UPDATE_MODIFY)
            remove_entry = true;
        else
            skip_update_modify_cnt++;
    }   
}   
```
as:
```
    }
    else if (upd_type != WT_UPDATE_MODIFY)  
        remove_entry = true;
    else
        skip_update_modify_cnt++;
}   
```
there's no need for an additional indented block?

src/cache/cache_las.c
```diff
@@ -972,19 +976,21 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+      WT_CONFIG_BASE(session, WT_SESSION_begin_transaction), "isolation=snapshot", NULL};
+    bool local_txn, locked, prev_entry_can_remove, remove_entry;
 
     cache = S2C(session)->cache;
     cursor = NULL;
     sweep_key = &cache->las_sweep_key;
     remove_cnt = 0;
     session_flags = 0; /* [-Werror=maybe-uninitialized] */
-    local_txn = locked = removing_key_block = false;
+    local_txn = locked = remove_entry = false;
```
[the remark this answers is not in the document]
keithbostic: I don't think `saved_btree_id` needs to be initialized here, I'd initialize it down next to the loop like the other loop-specific variables.

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+                cursor->remove(cursor);
+                cursor->next(cursor);
+                WT_ERR(
+                  cursor->get_key(cursor, &las_btree_id, &las_key, &las_timestamp, &las_txnid));
+                WT_ERR(cursor->get_value(
+                  cursor, &durable_timestamp, &prepare_state, &upd_type, &las_value));
+                ++remove_cnt;
+            }
+            skip_update_modify_cnt = 0;
+        }
```
keithbostic: This loop is going to be unnecessarily expensive in order to support `WT_VERB_LOOKASIDE_ACTIVITY` (which is never configured in the field).

I would suggest only retrieving the key  and value in the case of 
```
if (WT_VERBOSE_ISSET(session, WT_VERB_LOOKASIDE_ACTIVITY))
```
I think we could also only retrieve the key/value once, at the start of the `for` loop, rather than twice, before the for loop and at the end of the for loop?

And maybe include `remove_cnt` as part of the for loop?

keithbostic: This seems like the only useful `__wt_verbose()` call (the one earlier in the loop doesn't add much, does it?)

Can we enhance this one to include all of the relevant information, including things like `prepare_state` and the timestamp, and get rid of the other one?

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+            /* Remove each individual LAS record till the current record. */
+            for (i = 0; i < skip_update_modify_cnt + 1; i++) {
+                __wt_verbose(session, WT_VERB_LOOKASIDE_ACTIVITY,
+                  "Sweep removing lookaside entry with "
+                  "btree ID: %" PRIu32 " saved key size: %" WT_SIZET_FMT ", record type: %" PRIu8
+                  " transaction ID: %" PRIu64,
+                  las_btree_id, saved_key->size, upd_type, las_txnid);
+
+                cursor->remove(cursor);
+                cursor->next(cursor);
```
keithbostic: The functions aren't being checked for error returns.

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
-          " transaction ID: %" PRIu64,
-          las_btree_id, saved_key->size, upd_type, las_txnid);
-        WT_ERR(cursor->remove(cursor));
-        ++remove_cnt;
+        prev_entry_can_remove = true;
+        /* Save the first key. */
+        if (saved_btree_id == 0) {
+            saved_btree_id = las_btree_id;
+            WT_ERR(__wt_buf_set(session, saved_key, las_key.data, las_key.size));
+        }
```
keithbostic: There's some code around line 1201:
```
/*  
 * If the loop terminates after completing a work unit, we will continue the table sweep next
 * time. Get a local copy of the sweep key, we're going to reset the cursor; do so before
 * calling cursor.remove, cursor.remove can discard our hazard pointer and the page could be
 * evicted from underneath us. 
 */  
if (ret == 0) {
    WT_ERR(__wt_cursor_get_raw_key(cursor, sweep_key));
    if (!WT_DATA_IN_ITEM(sweep_key))
        WT_ERR(__wt_buf_set(session, sweep_key, sweep_key->data, sweep_key->size));
}
```
I'm not sure what `cursor.remove` that comment is talking about and looking through old versions of the code, it seems like there was one near here awhile ago, but it's long gone. If you agree with that analysis, I'd suggest something like:
```
 * If the loop terminates after completing a work unit, we will continue the table sweep next
 * time. Get a local copy of the sweep key to initialize the cursor position.
```

test/suite/test_las07.py
```diff
@@ -0,0 +1,142 @@
...
+
+        # Pin oldest and stable to timestamp 400.
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(400) +
+            ',stable_timestamp=' + timestamp_str(400))
+
+        # Check that the new updates are only seen after the update timestamp
+        self.check(bigvalue2, uri, nrows, 400)
+
+if __name__ == '__main__':
+    wttest.run()
```
keithbostic: Is there any reason we couldn't read the LAS table from the test and check values explicitly? (I honestly don't know, I'm just wondering if that would be a useful thing to do.)

And are we testing chains of modify records anywhere? That seems like the most complicated case sweep has to deal with.

src/cache/cache_las.c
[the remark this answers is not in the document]
tetsuo-cpp: I'm not sure what the rest of sweep looks like but at least for the changes here, scanning with `prev` should be easier. There just needs to be a flag to track whether the last record we saw was a modify. And if so, a standard update should be protected from removal even if it is expired.

There's probably some nuance that I'm missing but I still imagine it's simpler than this. Not sure whether it's worth the effort to try this but that's my initial thought.

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+            }
+        }
 
-            WT_ASSERT(session,
-              (prepare_state == WT_PREPARE_INPROGRESS || durable_timestamp >= las_timestamp));
+        /* Remove the obsolete entries */
+        if (remove_entry) {
+            /* Rewind the cursor to a point of remove record. */
+            for (i = 0; i < skip_update_modify_cnt + 1; i++)
+                cursor->prev(cursor);
```
tetsuo-cpp: Also no check here.

test/suite/test_las07.py
```diff
@@ -0,0 +1,142 @@
...
+            self.assertEqual(v, check_value)
+            count += 1
+        session.rollback_transaction()
+        self.assertEqual(count, nrows)
+
+    def test_las(self):
+        nrows = 10000
+
+        # Create a table without logging to ensure we get "skew_newest" lookaside eviction behavior.
+        uri = "table:las02_main"
```
tetsuo-cpp: Nit: Should be called `las07_main`.

test/suite/test_las07.py
```diff
@@ -0,0 +1,142 @@
...
+        ds = SimpleDataSet(
+            self, uri, 0, key_format="S", value_format="S", config='log=(enabled=false)')
+        ds.populate()
+
+        uri2 = "table:las02_extra"
```
tetsuo-cpp: Nit: And here.

test/suite/test_las07.py
```diff
@@ -0,0 +1,142 @@
...
+        # Check that the new updates are only seen after the update timestamp
+        self.check(bigvalue, uri, nrows, 350)
+        self.check(bigvalue2, uri, nrows, 400)
+
+        # Pin oldest and stable to timestamp 400.
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(400) +
+            ',stable_timestamp=' + timestamp_str(400))
+
+        # Check that the new updates are only seen after the update timestamp
+        self.check(bigvalue2, uri, nrows, 400)
```
tetsuo-cpp: This doesn't check that sweep is actually removing entries, does it? Is that possible to test?

agorrod: It's historically been hard to test that without introducing noise in the test suite.

src/cache/cache_las.c
```diff
@@ -920,9 +920,10 @@ __las_sweep_init(WT_SESSION_IMPL *session, WT_CURSOR *las_cursor)
      * If no files have been dropped and the lookaside file is empty, there's nothing to do.
      */
     if (cache->las_dropped_next == 0) {
-        if (__wt_las_empty(session))
+        if (__wt_las_empty(session)) {
             ret = WT_NOTFOUND;
-        goto err;
+            goto err;
+        }
```
agorrod: This change really should be in a different ticket that's merged into develop. We want to get this change backported to MongoDB 4.2 and earlier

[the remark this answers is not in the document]

src/cache/cache_las.c
```diff
@@ -952,6 +953,9 @@ __las_sweep_init(WT_SESSION_IMPL *session, WT_CURSOR *las_cursor)
     /* Clear the list of btree IDs. */
     cache->las_dropped_next = 0;
 
+    /* Reset the cursor to start from first record */
+    WT_ERR(las_cursor->reset(las_cursor));
+
```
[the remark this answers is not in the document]
kommiharibabu: It is possible to resume the `__wt_las_sweep()` function sometimes, so adding a reset call before `cursor.next` may not be proper. If you feel that `reset` should be in `__wt_las_sweep()` function then add the `reset` only when the `__las_sweep_init()` call occurs in `__wt_las_sweep()` function.

src/cache/cache_las.c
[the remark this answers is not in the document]
kommiharibabu: I fix the removal when it is travesing backwards.

@tetsuo-cpp, I understand your comment as traverse from backwards instead of forward, is it correct?
I thought of writing the code using the `prev` instead of `next`, but what I felt is that using of `next` there is a greater chance of removing the records than using the `prev` when there is a hight amount writes that are happening to the LAS.

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
-         * It's important to check for page boundaries explicitly
-         * because it is possible for the same key to be at the start
-         * of the next block. See WT-3982 for details.
+         * Expect an update entry with:
+         *	1. not in a prepare locked state
+         *	2. durable timestamp as not max timestamp.
+         *	3. for an in-progress prepared update, durable
+         *	timestamp should be zero.
+         *	4. no restriction on durable timestamp value
+         *	for other updates.
```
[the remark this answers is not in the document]
kommiharibabu: Corrected them and also updated them according to ASSERT statements.

src/cache/cache_las.c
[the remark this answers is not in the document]

src/cache/cache_las.c
```diff
@@ -1068,8 +1082,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
          * all we wanted or enough and there is a reader waiting and we're on a key boundary.
          */
         ++visit_cnt;
-        if (!removing_key_block &&
-          (cnt == 0 || (visit_cnt > WT_LAS_SWEEP_ENTRIES && cache->las_reader)))
+        if (cnt == 0 || (visit_cnt > WT_LAS_SWEEP_ENTRIES && cache->las_reader))
```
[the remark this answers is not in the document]
kommiharibabu: >   __wt_buf_free(session, sweep_key);
    ret = WT_NOTFOUND;
    break;

Using WT_ERR there generates an error, but that scenario is a success scenario, later the return value is modified to 0. So I changed to `goto err;`

src/cache/cache_las.c
```diff
@@ -1101,65 +1110,85 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
-            /* If we've examined enough entries, give up. */
-            if (cnt == 0)
-                break;
+        WT_ASSERT(session, prepare_state != WT_PREPARE_LOCKED && durable_timestamp != WT_TS_MAX &&
+            (prepare_state != WT_PREPARE_INPROGRESS || durable_timestamp == 0));
 
-            saved_btree_id = las_btree_id;
-            WT_ERR(__wt_buf_set(session, saved_key, las_key.data, las_key.size));
+        WT_ASSERT(
+          session, (prepare_state == WT_PREPARE_INPROGRESS || durable_timestamp >= las_timestamp));
```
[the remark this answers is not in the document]
kommiharibabu: Fixed and used `WT_ERR_ASSERT()`. Hope the error strings are meaningful.

src/cache/cache_las.c

src/cache/cache_las.c
kommiharibabu: Thanks. That really improves. Fixed.

kommiharibabu: Removed the first `__wt_versbose` and updated the second one to include additional fields.

src/cache/cache_las.c

src/cache/cache_las.c

test/suite/test_las07.py
kommiharibabu: The sweep server cannot be triggered from the test to let it clean the obsolete records. And also reading from the LAS for correctness is difficult as we are not sure what records it should have cleaned in the previous sweep cycle. There are no stats available currently to find out how many LAS records are available. I feel those are needed.

And regarding `modify` records, I modified the test to include them.

src/cache/cache_las.c

test/suite/test_las07.py
kommiharibabu: Fixed.

test/suite/test_las07.py
kommiharibabu: Fixed.

src/cache/cache_las.c
```diff
@@ -1002,7 +1008,13 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
     WT_ASSERT(session, cursor->session == &session->iface);
     __las_set_isolation(session, &saved_isolation);
-    WT_ERR(__wt_txn_begin(session, NULL));
+
+    /*
+     * Open a transaction with snapshot isolation. The LAS cursor moves it's position to either prev
+     * or next many times during the sweep process, so make sure that the cursor doesn't see newer
+     * updates.
+     */
+    WT_ERR(__wt_txn_begin(session, txn_cfg));
```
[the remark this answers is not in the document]
kommiharibabu: No. The sweep server during the processing of a sweep, it shouldn't see other inserts that are happened to the LAS in parallel. As we are cleaning the previous records based on the next record, just want to avoid scenarios where it can see a completely different record.

src/cache/cache_las.c
kommiharibabu: ok, I will fix the above change in the develop ticket and merge it here.

kommiharibabu:
retest this please

src/cache/cache_las.c
kommiharibabu: Done.

test/suite/test_las07.py
kommiharibabu: I modified the test to let the sweep server to get triggered during the running of the test, but it is difficult to verify it. Adding stats of how many records present in the LAS and the way to trigger the LAS sweep from test capability can provide the verification.

src/cache/cache_las.c
```diff
@@ -1086,8 +1090,14 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
      */
     __wt_las_cursor(session, &cursor, &session_flags);
     WT_ASSERT(session, cursor->session == &session->iface);
-    WT_ERR(__wt_txn_begin(session, NULL));
     __las_set_isolation(session, &saved_isolation);
+
```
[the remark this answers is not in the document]
agorrod: @kommiharibabu the lookaside cursors need to use `read uncommitted`. I think of it like this: we are writing content into the lookaside file that contains the information required to decide whether something is visible. Any lookaside cursor should return all available data, then the thing digesting the information uses the timestamp/txnID information stored into the lookaside file to make decisions about visibility.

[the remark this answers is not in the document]
kommiharibabu: @agorrod I agree with you regarding the LAS read with `read uncommitted`. It works even for the `sweep server` also in the `develop` because from the LAS records itself it is possible to identify whether the disk image is written the newest version or not? If the disk image is visible to everyone, the entire contents of the LAS records related to the key can be removed.

But with `relevant history` there is no way to identify whether the disk image record is the newest page record or not? So we cannot delete the entire key block without checking each individual record. And with `modify` records, we should not remove the complete update records until there is a complete update record is found for the same key. Because of this reason, the `prev` record can only be deleted once the `next` record is not of modify type or there is a complete record.

With `read uncommitted` isolation, there is a chance the LAS prev operation may get a completely new record that is inserted newly than the record that we visited previously. Because of this reason, I do the `snapshot` isolation to get the same record that it visited last time.

Please let me know your views on this problem, is it not possible at all? or any better solution?

src/cache/cache_las.c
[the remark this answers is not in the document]
keithbostic: Good catch. I think there still may be something here, because the `err` label is going to free `sweep_key`. Should we jump to `srch_notfound` instead?

src/cache/cache_las.c
```diff
@@ -1186,72 +1194,102 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+            WT_ERR(cursor->next(cursor));
+
+            /* Get the LAS key and value in case if their replaced. */
+            if (WT_VERBOSE_ISSET(session, WT_VERB_LOOKASIDE_ACTIVITY)) {
+                WT_ERR(
+                  cursor->get_key(cursor, &las_btree_id, &las_key, &las_timestamp, &las_txnid));
+                WT_ERR(cursor->get_value(
+                  cursor, &durable_timestamp, &prepare_state, &upd_type, &las_value));
+            }
```
keithbostic: I'm confused why this code is here? It doesn't seem like the retrieved key/value information is ever used?

src/cache/cache_las.c
```diff
@@ -1186,72 +1194,102 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
-             *	timestamp should be zero.
-             *	4. no restriction on durable timestamp value
-             *	for other updates.
-             */
-            WT_ASSERT(session, prepare_state != WT_PREPARE_LOCKED &&
-                durable_timestamp != WT_TS_MAX &&
-                (prepare_state != WT_PREPARE_INPROGRESS || durable_timestamp == 0));
+        /* Remove the obsolete entries */
+        if (remove_entry) {
+            /* Remove each individual LAS record till the current record. */
```
keithbostic: This comment no longer makes sense, "till the current record" isn't what we do now.

src/cache/cache_las.c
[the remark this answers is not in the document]
kommiharibabu: @keithbostic, I don't think the problem can occur for the `same key` values, as there is no way that a new record that can be inserted in-between `same key` data. The problem can happen only in the boundary cases of moving from one key to another.

Let us take this example, Following are the LAS records. 
`key1 (10), timestamp (10), txnid (10)` -> `key2(20), timestamp(10), txnid(10)`

The sweep server identifies that `key1(10), timestamp(10), txnid(10)` record can be removed from LAS as it is obsolete, but to do that, it must verify whether the next record is not a modify operation. So the scan proceeds to the next record `key2(20), timestamp(10), txnid(10)` and decides that the key doesn't match, so the previous record can be deleted. By that time, it is possible that a new record for the `key1(10)` can get inserted with data as `key1(10), timestamp(40), txnid (40)`.

In the above scenario, if the sweep server continues to remove the `prev` record, it will remove the newly inserted record `key1(10), timestamp(40), txnid(40)`.

With the `isolation` snapshot, I think that this problem doesn't occur. I am not sure exactly how internally it works. Am I correct?

Another alternative to this problem is doing a `backward` scan of the LAS records and remove the record whenever it found as obsolete. There is no need to worry about `isolation` and `modify` record handling. opinions?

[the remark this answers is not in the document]
keithbostic: I believe your analysis is correct.

I also agree with your comment to @tetsuo-cpp that we are better served with a cursor-next approach to sweeping the LAS file.

An alternative I think might work (while allowing a read-uncommitted configuration), is to used the saved key/timestamp/txnid triplet to walk backward through the LAS file, discarding the matches. That is, cursor.prev backward until the number of matching saved-key modify records have been deleted, and then resume forward traversal. There's a count of modify records to be deleted, so the cursor.prev movement is bounded. We've added some get-key/get-value comparisons, but that's about all.

I agree this is more complicated in one dimension, but it would simplify the existing (fairly complex) loop logic, and also allow us to continue to use a read-uncommitted configuration.

I will leave it up to you, Alex and Vamsi which approach is best.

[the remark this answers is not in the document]
agorrod: @keithbostic, @kommiharibabu moved this conversation into JIRA.

src/cache/cache_las.c
kommiharibabu: > I think there still may be something here, because the err label is going to free sweep_key. Should we jump to srch_notfound instead

Once we reach the limit, we just break. In that code flow, we didn't free the `sweep_key` because the logic of `sweep_key` free is in `if (0)` condition.

> I'm confused why this code is here? It doesn't seem like the retrieved key/value information is ever used?

I remove the above statements with new logic.

src/cache/cache_las.c
kommiharibabu: > That was what I was thinking (add the reset only when __las_sweep_init() called`), I'm OK with your preference if you disagree.

Done.

> This comment no longer makes sense, "till the current record" isn't what we do now.

Removed.

kommiharibabu:
@keithbostic, I try to reduce the complexity of the logic. If it still looks like complex, I will check anything that I can improve.

PR #4905, opened 2019-10-08
WT-5160 Stop requiring a checkpoint before calling rollback_to_stable

agorrod:
We recently introduced this constraint, but it's not acceptable.


agorrod:
@bvpvamsikrishna please review.

PR #4906, opened 2019-10-08
WT-4975 Migrate wiredtiger-pull-request-asan

quchenhao:


test/evergreen.yml
```diff
@@ -71,59 +71,64 @@ functions:
...
-      display_name: Artifacts
-      remote_file: wiredtiger/${build_variant}/${revision}/artifacts/${task_id}.tgz
+      script: |
+        rm -rf "wiredtiger"
+        rm -rf "wiredtiger.tgz"
+
+pre:
+  - func: "cleanup"
+post:
+  - func: "upload artifact"
```
lukech: I like the refactoring code here.

test/evergreen.yml
```diff
@@ -71,59 +71,64 @@ functions:
...
+  - func: "cleanup"
```
lukech: This may not be necessary as Evergreen usually takes care of the cleanup.

test/evergreen.yml
```diff
@@ -13,7 +13,7 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        remote_file: wiredtiger/${build_variant}/${revision}/artifacts/${build_id}.tgz
+        remote_file: wiredtiger/${build_variant}/${revision}/artifacts/${parent_task|compile}_${build_id}.tgz
```
lukech: The term `parent_task` sounds a bit confusing to me. Could we use something else e.g. `dependent_task`?

test/evergreen.yml
```diff
@@ -71,59 +71,64 @@ functions:
...
-          display_name: Artifacts_compile
-          remote_file: wiredtiger/${build_variant}/${revision}/artifacts/${build_id}.tgz
-      # Remove the artifacts here so the later post commands won't perform duplicated archiving.
-      - command: shell.exec
-        params:
-          script: |
-            rm -rf "wiredtiger"
-            rm -rf "wiredtiger.tgz"
+      - func: "upload artifact"
+      - func: "cleanup"
```
lukech: Ditto. If the line is called explicitly here to avoid extra post cleanup, please add a comment.

test/evergreen.yml
quchenhao: dependent task sounds it can be multiple but really we can only specify one.

test/evergreen.yml
quchenhao: This is to prevent the build artifact being uploaded twice.

test/evergreen.yml
quchenhao: Better to keep that here. In case some of the stage failed to run.

test/evergreen.yml
```diff
@@ -71,59 +71,64 @@ functions:
...
+
+  - name: compile-asan
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          smp_command: -j $(grep -c ^processor /proc/cpuinfo)
+          test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
```
lukech: Are they necessary considering the 2 variables are set as the same values as build variant level expansion?

test/evergreen.yml
```diff
@@ -71,59 +71,64 @@ functions:
...
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ ASAN_OPTIONS=detect_leaks=1:abort_on_error=1:disable_coredump=0 ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer CFLAGS=-fsanitize=address
+      - func: "upload artifact"
```
lukech: Add setting for `posix_configure_flag`

test/evergreen.yml
```diff
@@ -148,209 +153,154 @@ tasks:
...
+      - func: "fetch artifacts"
+        vars:
+          parent_task: compile-asan
+      - func: "compile wiredtiger"
+      - func: "make check directory"
+        vars:
+          directory: examples/c
+          smp_command: -j $(grep -c ^processor /proc/cpuinfo)
+          test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ ASAN_OPTIONS=detect_leaks=1:abort_on_error=1:disable_coredump=0 ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer CFLAGS=-fsanitize=address
```
lukech: Move to before line 178

quchenhao:
retest this please

quchenhao:
evergreen retry

PR #4908, opened 2019-10-08
WT-5150 Fix the check to find out whethere LAS is empty

kommiharibabu:
Sweep server should periodically clean the LAS obsolete
entries when there as any tables are dropped or if the LAS
not empty. Fix the check that leads to not to remove the
obsolete entries when there are no tables dropped.


kommiharibabu:
@keithbostic, Yes, your understanding is correct. Sweep server never removes obsolete entries when there are no objects dropped.

PR #4911, opened 2019-10-09
WT-5167 Fix the missing pieces of page instantiation from las

kommiharibabu:
Update various memory allocations that are done during the page
instantiation for the tracking stats variables.

Set the page modify restore_state with proper value once the page
is successfully instantiated.

If LAS updates are newer than the disk image, set the flag to include
these updates in the next checkpoint.

Otherwise, if the page image contains newest versions of data than
LAS updates, check whether the page that is marked as dirtly can
be cleared under some conditions.

### DIFF
--- a/src/btree/bt_read.c
+++ b/src/btree/bt_read.c
@@ -69,14 +69,14 @@ __row_instantiate(
  *     Create a birthmark update to be put on the page.
  */
 static int
-__create_birthmark_upd(WT_SESSION_IMPL *session, WT_BIRTHMARK_DETAILS *birthmarkp, WT_UPDATE **updp)
+__create_birthmark_upd(
+  WT_SESSION_IMPL *session, WT_BIRTHMARK_DETAILS *birthmarkp, size_t *sizep, WT_UPDATE **updp)
 {
     WT_UPDATE *upd;
-    size_t not_used;
 
     *updp = NULL;
 
-    WT_RET(__wt_update_alloc(session, NULL, &upd, &not_used, WT_UPDATE_BIRTHMARK));
+    WT_RET(__wt_update_alloc(session, NULL, &upd, sizep, WT_UPDATE_BIRTHMARK));
     upd->txnid = birthmarkp->txnid;
     upd->durable_ts = birthmarkp->durable_ts;
     upd->start_ts = birthmarkp->start_ts;
@@ -98,12 +98,14 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
     WT_DECL_RET;
     WT_PAGE_LOOKASIDE *page_las;
     WT_UPDATE *upd;
+    size_t incr, total_incr;
     uint64_t recno;
     uint32_t i;
     const uint8_t *p;
 
     page_las = ref->page_las;
     upd = NULL;
+    total_incr = 0;
 
     __wt_btcur_init(session, &cbt);
     __wt_btcur_open(&cbt);
@@ -113,7 +115,8 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
         if (birthmarkp->instantiated)
             continue;
 
-        WT_ERR(__create_birthmark_upd(session, birthmarkp, &upd));
+        WT_ERR(__create_birthmark_upd(session, birthmarkp, &incr, &upd));
+        total_incr += incr;
 
         switch (ref->page->type) {
         case WT_PAGE_COL_FIX:
@@ -137,6 +140,7 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
 
     page_las->birthmarks_cnt = 0;
     __wt_free(session, page_las->birthmarks);
+    __wt_cache_page_inmem_incr(session, ref->page, total_incr);
 
 err:
     for (i = 0, birthmarkp = page_las->birthmarks; i < page_las->birthmarks_cnt; i++, birthmarkp++)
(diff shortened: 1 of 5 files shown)


src/btree/bt_read.c
```diff
@@ -232,9 +255,10 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
     cache->las_reader = true;
     __wt_readlock(session, &cache->las_sweepwalk_lock);
     cache->las_reader = false;
+    incr = total_incr = 0;
```
keithbostic: `__create_birthmark_upd()` calls `__wt_update_alloc()`, and `__wt_update_alloc()` should be clearing `*sizep` out of sheer paranoia.

src/btree/bt_read.c
```diff
@@ -351,6 +377,16 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
     /* Instantiate the remaining birthmark entries. */
     WT_ERR(__instantiate_birthmarks(session, ref));
 
+    if (total_incr != 0)
+        __wt_cache_page_inmem_incr(session, page, total_incr);
```
keithbostic: Why don't we add:
```
if (size == 0)
    return;
```
into `__wt_cache_page_inmem_incr()` and then we'll catch any functions wasting work on useless increments. It's a register compare, so it's basically free. Then we don't to make any explicit checks.

src/cache/cache_las.c
```diff
@@ -557,6 +557,7 @@ __las_squash_modifies(WT_SESSION_IMPL *session, WT_CURSOR *cursor, WT_UPDATE **u
     size_t allocated_bytes;
     u_int i;
 
+    WT_CLEAR(las_value);
```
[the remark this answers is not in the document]
kommiharibabu: Yes, I agree with it, Earlier the initialization was close to the usage place, but it leads to clang scan-build warnings of uninitialized variable usage. The `las_value` is freed in the `err` path whenever there is any error occurs. The uninitialized warning is coming with an assumption of an error occurring before the variable is initialized. Because of that reason, I moved the initialization to a place before any error occurs.

src/btree/bt_read.c
```diff
@@ -194,8 +200,25 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
     locked = false;
 
     /* Check whether the disk image contains all the newest versions of the page. */
-    if (ref->page_las->min_skipped_ts == WT_TS_MAX)
-        return (__instantiate_birthmarks(session, ref));
+    if (page_las->min_skipped_ts == WT_TS_MAX) {
+        WT_RET(__instantiate_birthmarks(session, ref));
+
+        FLD_SET(page->modify->restore_state, WT_PAGE_RS_LOOKASIDE);
```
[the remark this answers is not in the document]
kommiharibabu: The disk image that is written already contains the newest version of the page, so there is nothing to be done by the next checkpoint, because of this reason I didn't set that here. Even setting it with `WT_TXN_FIRST` doesn't matter with next checkpoint, it is better to do it. Opinions?

src/btree/bt_read.c
```diff
@@ -351,6 +377,16 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
     WT_ERR(__instantiate_birthmarks(session, ref));
 
+    if (total_incr != 0)
+        __wt_cache_page_inmem_incr(session, page, total_incr);
+
+    /*
+     * If the updates in lookaside are newer than the versions on the page, it must be included in
+     * the next checkpoint.
+     */
+    page->modify->first_dirty_txn = WT_TXN_FIRST;
```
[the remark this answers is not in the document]
kommiharibabu: The page marking it as clean logic only happens when the disk image contains all the newer versions of the page. That logic is already moved into a separate section at the top of the function. So it is not required here.

src/btree/bt_read.c
```diff
@@ -194,8 +200,25 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
+        WT_RET(__instantiate_birthmarks(session, ref));
+
+        FLD_SET(page->modify->restore_state, WT_PAGE_RS_LOOKASIDE);
+
+        /*
+         * The page image contained the newest versions of data so the updates in lookaside are all
+         * older and we could consider marking it clean (i.e., the next checkpoint can use the
+         * version already on disk).
+         */
+        if (!page_las->has_prepares && !S2C(session)->txn_global.has_stable_timestamp &&
```
tetsuo-cpp: The conditional is not obvious to me. Could you explain why we're checking this?

test/suite/test_las03.py
```diff
@@ -89,9 +89,9 @@ def test_checkpoint_las_reads(self):
             # Now just update one record and checkpoint again.
             self.large_updates(self.session, uri, bigvalue2, ds, nrows, 1)
 
-            las_reads_start = self.get_stat(stat.conn.cache_read_lookaside)
```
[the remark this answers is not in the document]
kommiharibabu: Sure, I will raise some other ticket to modify the function.

src/btree/bt_read.c
kommiharibabu: Without `incr = 0` compiler is generating warnings as uninitialized variable usage even if it is not used as uninitialized. so I added the initialization.

src/btree/bt_read.c
kommiharibabu: Fixed.

src/btree/bt_read.c
kommiharibabu: When the disk image of page contains the newest versions of the page and the data is globally visible, there is nothing required by the checkpoint to take care of it during the checkpoint. So we can mark the page as clean, but only when the page doesn't have prepares and there is no stable timestamp is available. I bring this check from earlier code.

PR #4912, opened 2019-10-09
WT-5158 Remove assertion in __wt_find_lookaside_upd for cursor comparison

tetsuo-cpp:
This PR is to fix an assertion failure in our long-running branch that I've been chasing. It happens in `test_las02.py` specifically when we iterate over the entire contents of a table by repeatedly calling `next`.

```
    /*
     * Prior to calling this function, we should have successfully searched the correct key with our
     * cursor.
     */
    WT_ASSERT(session, cbt->compare == 0);
```

The point of the assertion was to ensure that we don't hit a situation where we don't find our desired key and enter this code block. In `__wt_find_lookaside_upd`, we use the search key on `cbt` and use it to read updates from lookaside (and in some cases insert them into the update list) which is quite serious if we're not pointing at the right key.

This has been challenging to debug because:
* It only reproduces on `saru`.
* My attempts to set breakpoints in the debugger between the opening of the cursor and the assertion failure causes it to no longer reproduce. I need to be able to do this in order to set a conditional for that particular cursor.
* My attempts at adding prints sometimes cause it to no longer reproduce.

The code path for calling `next` on a cursor doesn't involve setting the `compare` value which tells me that it's likely set by a previous cursor search call. I don't fully understand what race is causing this code path but zeroing out the `compare` value on each API call seems necessary regardless since previous calls shouldn't be able to interfere in this way.

It's an unsatisfying explanation but my feeling is that it's best to just merge this in and keep forging ahead to avoid sinking too much time into this.


tetsuo-cpp:
As discussed, since the `compare` value can't be set to 0 on every API call, this assertion is wrong. A missed search and a `next` call will cause this assertion to fire which doesn't make sense.

@keithbostic is taking a quick look to see where the failed search is happening. But in the absence of any further information, we're just going to remove this assertion and call it a day.

PR #4913, opened 2019-10-09
WT-5123 Fill multi-paragraph comments

tetsuo-cpp:
It's probably helpful to look at each commit separately in this PR. In this instance, I'd say that it'd be better not to squash these commits since we'll want to cherrypick the script and typo changes separately when backporting.

If this review yields some extra code changes, I'll rebase them into the initial commit before merging.

### DIFF
--- a/dist/s_comment.py
+++ b/dist/s_comment.py
@@ -63,14 +63,20 @@ for line in sys.stdin:
             indent_ws = ' ' * indentation
             sys.stdout.write('{}/*\n'.format(indent_ws))
             current_line = indent_ws + ' *'
-            for word in words:
+            for i in range(len(words)):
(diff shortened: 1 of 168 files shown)


dist/s_comment.py
```diff
@@ -89,6 +95,10 @@
         function_desc = False
     elif multiline:
         comment += line
+        # We want to preserve newlines for block comments that have multiple paragraphs.
+        if len(sline) == 1 and sline.startswith('*'):
```
lukech: Does it actually equal `sline = '*'`?

tetsuo-cpp: Ha! Good point.
I rebased that change into the original script commit.

PR #4915, opened 2019-10-11
WT-4980 Add conf-dump-test to evergreen

quchenhao:


test/evergreen.yml
```diff
@@ -923,6 +923,20 @@ tasks:
 
   # End of Python unit test tasks
 
+  - name: test-conf-dump
```
lukech: Would be great to follow the existing task naming convention of putting `test` to the end, i.e. `conf-dump-test`.

test/evergreen.yml
```diff
@@ -1128,6 +1142,7 @@ buildvariants:
     test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
     smp_command: -j $(grep -c ^processor /proc/cpuinfo)
     configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH
+    posix_configure_flags: --enable-diagnostic --enable-python --enable-zlib --enable-snappy --enable-strict --enable-static --prefix=$(pwd)/LOCAL_INSTALL
```
lukech: Should we add `CFLAGS="-ggdb" --enable-silent-rules`, which are currently in the `wiredtiger-pull-request-python` job? And do we need to preload `libeatmydata.so`?

test/evergreen.yml
```diff
@@ -923,6 +923,20 @@ tasks:
...
+    commands:
+      - func: "fetch artifacts"
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+            cd test
+            ${test_env_vars|} ${python_binary|python} ../../test/wtperf/test_conf_dump.py 2>&1
```
lukech: Not sure if these directory switchings are still necessary. Can we set the working_dir to  `wiredtiger/test/wtperf` on line 933 and call the script directly?

test/evergreen.yml
quchenhao: These flags are all about debugging information and doesn't change the build itself. I don't want to create a separate compile task just for this test. libeatmydata is currently not available and we have decided to support that when we have it.

quchenhao:
retest this please

PR #4917, opened 2019-10-11
WT-5182 Add s-all test to evergreen

quchenhao:


test/evergreen.yml
```diff
@@ -923,6 +923,19 @@ tasks:
 
   # End of Python unit test tasks
 
+  - name: s-all
```
lukech: Depends on if WT-5176 is merged before this one, we need to tag this task as `pull-request`.

test/evergreen.yml
```diff
@@ -923,6 +923,19 @@ tasks:
...
+    - name: compile
+    commands:
+      - func: "fetch artifacts"
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/dist"
+          script: |
+            set -o errexit
+            set -o verbose
+            sh s_all -A -E 2>&1
```
lukech: What's the benefit of `-A -E` options?

test/evergreen.yml
```diff
@@ -923,6 +923,19 @@ tasks:
 
   # End of Python unit test tasks
 
+  - name: s-all
+    depends_on:
+    - name: compile
```
lukech: Some special configure option, preloaded library and env variables currently set in `wiredtiger-pull-request-linux` job (pasted below) are not covered by the `compile` task. 
```../configure CFLAGS="-ggdb" --enable-silent-rules --enable-strict --enable-diagnostic --disable-static ; make -j2
[ -e /usr/local/lib/libeatmydata.so ] && export LD_PRELOAD=/usr/local/lib/libeatmydata.so
export TESTUTIL_DISABLE_LONG_TESTS=1
```

test/evergreen.yml
quchenhao: To make it fail if something is wrong otherwise the command will always exit with exit code 0.

quchenhao:
This is not to migrate the linux job actually. The s_all command in pull-request-linux will always succeed as there is no -A -E flags. This is to migrate part of the pull-request-builds job which runs s_all -A -E.
We have discussed libeatmydata package before. This is only to speed up the test and we will support that when it is available in the distro. I haven't checked TESTUTIL_DISABLE_LONG_TESTS this env variable as this is not intended to migrate the pull-request-linux-job

test/evergreen.yml
quchenhao: We can fist merge WT-5176 and rebase.

PR #4920, opened 2019-10-14
WT-5176 Group pull request tasks using Evergreen task tags

lukech:
Changes include:
- tagging all applicable tasks as `pull_request`
- tagging Windows only tasks as `windows`
- simplifying listing tasks in `ubuntu1804` build variant
- updating task templates to fit recent retro code

@quchenhao Can you help to take a look? Thanks.

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -982,7 +982,7 @@ tasks:
   # End of Python unit test tasks
 
   - name: compile-windows-alt
-    tags: ["pull_request", "windows"]
+    tags: ["pull_request", "windows_only"]
     depends_on:
     - name: compile
     commands:
@@ -1016,7 +1016,7 @@ tasks:
             fi
 
   - name: format
-    tags: ["pull_request", "windows"]
+    tags: ["pull_request", "windows_only"]
     depends_on:
     - name: compile
     commands:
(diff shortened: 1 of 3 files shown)


test/evergreen/csuite_test_evg_task.template
```diff
@@ -1,4 +1,5 @@
   - name: {{task_name}}
+    tags: ["pull_request"]
```
quchenhao: I don't know why we need these template files. They just duplicate the same code in the evergreen.yaml file. Are we using them correctly?

test/evergreen.yml
```diff
@@ -1124,70 +1185,12 @@ buildvariants:
...
-    - name: unit-test-bucket02
-    - name: unit-test-bucket03
-    - name: unit-test-bucket04
-    - name: unit-test-bucket05
-    - name: unit-test-bucket06
-    - name: unit-test-bucket07
-    - name: fops
-    - name: compile-asan
-    - name: examples-c-test-asan
+    - name: ".pull_request !.windows"
```
quchenhao: Are we plan only to run ubuntu1804 in pull request? What about the python3 tests?

quchenhao: I get it. I think it is OK. But may be better to add more tags such as, csuite, make_test, python_unit, etc... So we can use the same mechanism in other build variants.

test/evergreen.yml
```diff
@@ -956,6 +1016,7 @@ tasks:
             fi
 
   - name: format
+    tags: ["pull_request", "windows"]
```
quchenhao: May be better to explicitly use "windows_only" and in the future, we'll have "linux_only" tests such as "s_all"

test/evergreen/csuite_test_evg_task.template
lukech: When people add new make check directories or new csuite tests, it's expected some new code being added into evergreen.yml to cover those tests in Evergreen. The templates files are used to make people's life easier in crafting the evergreen config. Try run `s_evergreen` after adding a new test and you should see the output.

test/evergreen.yml
lukech: Yeah I agree, those will be improvements we could consider in the future, while not part of this ticket.

test/evergreen.yml
lukech: OK, will update the tag name as suggested.

PR #4927, opened 2019-10-18
WT-5193 Revert "WT-5150 Fix the check to find out whethere LAS is empty (#4908)"

sueloverso:
This reverts commit a74b4299f9cd5701c580cae502a3f81585a4d89f.


sueloverso:
@kommiharibabu I added you as reviewer. I thought @keithbostic might see it the other day. But now you're coming online soon.

PR #4928, opened 2019-10-20
WT-5191 LAS sweep to consider last checkpoint time

kommiharibabu:
Sometimes it is possible that the obsolete records in
lookaside are still not may be written to disk image. To avoid
removal of those records, consider the last_ckpt_timestamp also
into consideration before removing those records.

Checkpoint will ensure read all the pages in the tree that has
lookaside content for checkpoint, this ensures that proper version
of the lookaside records are written to disk image, so considering
the last checkpoint time also in obsolete record removal will avoid
of removing any records that don't have disk image.

### DIFF
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -1272,6 +1272,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 
         /* Don't remove the non obsolete entry. */
         if (las_timestamp > S2C(session)->txn_global.oldest_timestamp ||
+          las_timestamp > S2C(session)->txn_global.last_ckpt_timestamp ||
           prepare_state == WT_PREPARE_INPROGRESS) {
             /*
              * TODO: In case if there exists any pending remove count records that can be removed,
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -1082,7 +1082,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
     uint32_t las_btree_id, remove_btree_id, saved_btree_id, session_flags;
     uint8_t prepare_state, upd_type;
     int cmp, notused;
-    bool local_txn, key_change, locked, prev_rec_verified, globally_visible_ondisk_value;
+    bool globally_visible_ondisk_value, key_change, local_txn, locked, prev_rec_verified;
 
     cache = S2C(session)->cache;
     cursor = NULL;
--- a/src/cache/cache_las.c
+++ b/src/cache/cache_las.c
@@ -1082,7 +1082,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
     uint32_t las_btree_id, remove_btree_id, saved_btree_id, session_flags;
     uint8_t prepare_state, upd_type;
     int cmp, notused;
-    bool local_txn, key_change, locked, prev_rec_verified;
+    bool local_txn, key_change, locked, prev_rec_verified, special_ondisk_value;
 
     cache = S2C(session)->cache;
     cursor = NULL;
@@ -1146,6 +1146,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
     saved_btree_id = 0;
     saved_timestamp = WT_TS_NONE;
     saved_txnid = WT_TXN_NONE;
+    special_ondisk_value = false;
 
     /* Walk the file. */
     while ((ret = cursor->next(cursor)) == 0) {
@@ -1237,9 +1238,19 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
         /*
          * Remove the previous obsolete entries whenever the scan moves into next key or the current
          * LAS record is not a modify record.
+         *
+         * The special ondisk value that is present in the LAS is only removed whenever a next
+         * record of the same key gets removed. This is to protect the cases where we may still need
+         * the older ondisk value for some scenarios. The draw back is that this record never be
+         * removes until no updates happens on that key, but these scenarios are minimal.
+         *
+         * TODO: Better key/value pair design that let you know that on disk impage of the key is
+         * globally visible will simplify the logic of removing the entire LAS records for that key.
          */
-        if (pending_remove_cnt && (key_change || upd_type != WT_UPDATE_MODIFY)) {
+        if ((special_ondisk_value ? pending_remove_cnt > 1 : pending_remove_cnt > 0) &&
+          (key_change || upd_type != WT_UPDATE_MODIFY)) {
             prev_rec_verified = false;
+            special_ondisk_value = false;
             while (pending_remove_cnt > 0) {
                 WT_ERR(cursor->prev(cursor));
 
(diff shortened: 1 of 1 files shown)


src/cache/cache_las.c
```diff
@@ -1272,6 +1272,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
 
         /* Don't remove the non obsolete entry. */
         if (las_timestamp > S2C(session)->txn_global.oldest_timestamp ||
+          las_timestamp > S2C(session)->txn_global.last_ckpt_timestamp ||
```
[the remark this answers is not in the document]
kommiharibabu: I handled all the above scenarios

kommiharibabu:
@keithbostic, @sulabhM, @tetsuo-cpp I fixed some additional edge case scenarios of LAS having a previous on-disk image value.

src/cache/cache_las.c
```diff
@@ -1237,9 +1238,19 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
         /*
          * Remove the previous obsolete entries whenever the scan moves into next key or the current
          * LAS record is not a modify record.
+         *
+         * The special ondisk value that is present in the LAS is only removed whenever a next
+         * record of the same key gets removed. This is to protect the cases where we may still need
+         * the older ondisk value for some scenarios. The draw back is that this record never be
+         * removes until no updates happens on that key, but these scenarios are minimal.
```
sulabhM: removes until no updates --> removed until more updates

sulabhM: Can you also comment on what the special ondisk value is, ie one with 0 txnid and 0 timestamp that is globally visible and always the oldest record. It represents a previous on-disk value that is still needed for an old reader

src/cache/cache_las.c
```diff
@@ -1270,9 +1281,18 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
             WT_ERR(cursor->next(cursor));
         }
 
-        /* Don't remove the non obsolete entry. */
-        if (las_timestamp > S2C(session)->txn_global.oldest_timestamp ||
-          las_timestamp > S2C(session)->txn_global.last_ckpt_timestamp ||
+        /*
+         * There are several conditions that need to be met
```
sulabhM: Please change the comment such that lines wrap at 100 column width wherever you are not starting a new paragraph or writing separate points.

src/cache/cache_las.c
```diff
@@ -1270,9 +1281,18 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+         * before we choose to remove a lookaside entry:
+         *  * If there exists a last checkpoint timestamp, then the lookaside record timestamp
+         *    must be less than last checkpoint timestamp.
+         *  * The entry is globally visible.
+         *  * The entry wasn't from a prepared transaction.
+         */
+        if (((S2C(session)->txn_global.last_ckpt_timestamp != WT_TS_NONE) ?
```
sulabhM: I could be wrong here, is this if condition simplified to:
```
if (((S2C(session)->txn_global.last_ckpt_timestamp != WT_TS_NONE) &&
      (las_timestamp > S2C(session)->txn_global.last_ckpt_timestamp)) ||
     !__wt_txn_visible_all(session, las_txnid, durable_timestamp) ||
     prepare_state == WT_PREPARE_INPROGRESS)
```

kommiharibabu: Yes, Fixed.

src/cache/cache_las.c
kommiharibabu: Changed.

src/cache/cache_las.c
kommiharibabu: Changed accordingly.

src/cache/cache_las.c
```diff
@@ -1270,8 +1283,17 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
             WT_ERR(cursor->next(cursor));
         }
 
-        /* Don't remove the non obsolete entry. */
-        if (las_timestamp > S2C(session)->txn_global.oldest_timestamp ||
+        /*
+         * There are several conditions that need to be met before we choose to remove a lookaside
+         * entry:
+         *  * If there exists a last checkpoint timestamp, then the lookaside record timestamp must
```
tetsuo-cpp: Nit: Maybe use a hyphen instead of asterisk since we're already using an asterisk at the beginning of each line in a comment.

src/cache/cache_las.c
```diff
@@ -1237,9 +1238,21 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
         /*
          * Remove the previous obsolete entries whenever the scan moves into next key or the current
          * LAS record is not a modify record.
+         *
+         * The special ondisk value i.e both timestamp and transaction id values as 0 and is the
```
tetsuo-cpp: Is a "special" update one that is created by `__rec_append_orig_value`?

src/cache/cache_las.c
```diff
@@ -1146,6 +1146,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
     saved_btree_id = 0;
     saved_timestamp = WT_TS_NONE;
     saved_txnid = WT_TXN_NONE;
+    special_ondisk_value = false;
```
tetsuo-cpp: "Special" isn't very descriptive in my opinion. Maybe `special_ondisk_value` => `globally_visible_ondisk_value`? If that is accurate.

src/cache/cache_las.c
```diff
@@ -1237,9 +1238,21 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
...
+         * whenever a next record of the same key gets removed. This is to protect the cases where
+         * we may still need the older ondisk value for some scenarios. The draw back is that this
+         * record never be removed until more updates happens on that key, but these scenarios are
+         * minimal.
+         *
+         * TODO: Better key/value pair design that let you know that on disk image of the key is
+         * globally visible will simplify the logic of removing the entire LAS records for that key.
          */
-        if (pending_remove_cnt && (key_change || upd_type != WT_UPDATE_MODIFY)) {
+        if ((special_ondisk_value ? pending_remove_cnt > 1 : pending_remove_cnt > 0) &&
```
tetsuo-cpp: Just to clarify, is this because if we have a globally visible update we don't want to count that one? We need to have at least one more update getting removed before sweep is allowed to remove it?

src/cache/cache_las.c
kommiharibabu: Nice, changed.

src/cache/cache_las.c
kommiharibabu: Las sweep server has no information to identify the globally visible on disk value even if there is a disk image that is globally visible. So it will be removed only whenever any LAS record of the same key gets removed.

src/cache/cache_las.c
kommiharibabu: yes

src/cache/cache_las.c
kommiharibabu: Added numbers

src/cache/cache_las.c
```diff
@@ -1082,7 +1082,7 @@ __wt_las_sweep(WT_SESSION_IMPL *session)
     uint32_t las_btree_id, remove_btree_id, saved_btree_id, session_flags;
     uint8_t prepare_state, upd_type;
     int cmp, notused;
-    bool local_txn, key_change, locked, prev_rec_verified;
+    bool local_txn, key_change, locked, prev_rec_verified, globally_visible_ondisk_value;
```
keithbostic: Names should be sorted.

PR #4934, opened 2019-10-22
WT-5186 Handle reading of LAS modifies on top of a birthmark record

tetsuo-cpp:
The main work in this PR is to get the lookaside reading/instantiation logic to consider birthmark records and on-disk values however there are a few smaller fixes to get `las07` running properly. With this change, `las06` passes and `las07` gets a stuck cache (which is expected). `las07` gets much further in the test and hits the code paths I've added multiple times so I think this change is reasonable.

As part of fixing a bug in the squashing logic, I've done a light refactor by introducing a `WT_MODIFY_VECTOR` data structure to contain the small vector logic we've been using through the codebase. The actual logic for applying the modifies is still repeated however it is hard to abstract since they are subtly different. For example, sometimes we own the updates (because we alloc'd them from lookaside) and sometimes they belong to the update list so the cleanup is different. So I think this is the correct level to generalize. This logic will disappear when we reverse deltas so I don't think it's worth getting too hung up on this part.


src/cache/cache_las.c
```diff
@@ -1462,40 +1478,64 @@ __wt_find_lookaside_upd(
          */
         if (upd_type == WT_UPDATE_MODIFY) {
             WT_NOT_READ(modify, true);
+            __find_birthmark_update(cbt, &birthmark_upd);
             while (upd_type == WT_UPDATE_MODIFY) {
```
kommiharibabu: This loop exit in three conditions.
1. Whenever a UPDATE_STANDARD record type is found in LAS.
2. Whenever we found a LAS record of the same key that is less than a birthmark record.
3. Whenever we cross the key boundary.

I feel adding the above conditions in comments may be helpful.

src/btree/bt_read.c
```diff
@@ -152,6 +152,78 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
...
+ *     Free any resources associated with a modify vector. If we exceeded the allowed stack space on
+ *     the vector and had to fallback to dynamic allocations, we'll be doing a free here.
+ */
+void
+__wt_modify_vector_free(WT_MODIFY_VECTOR *modifies)
+{
+    if (modifies->capacity != 0)
+        __wt_free(modifies->session, modifies->listp);
+    WT_CLEAR(*modifies);
+    modifies->listp = modifies->list;
```
keithbostic: Call `__wt_modify_vector_init()` instead of repeating the code?

src/btree/bt_read.c
```diff
@@ -166,17 +238,18 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
     WT_PAGE_LOOKASIDE *page_las;
     WT_PAGE *page;
-    WT_UPDATE *list[WT_MODIFY_ARRAY_SIZE], **listp, *upd;
-    wt_timestamp_t durable_timestamp, las_timestamp;
-    size_t allocated_bytes, notused, size, total_incr;
-    uint64_t birthmark_cnt, las_txnid, recno;
+    WT_UPDATE *mod_upd, *upd;
+    wt_timestamp_t durable_timestamp, _durable_timestamp, las_timestamp, _las_timestamp;
+    size_t notused, size, total_incr;
+    uint64_t birthmark_cnt, las_txnid, _las_txnid, recno;
```
keithbostic: I really don't like the underscore prefixes here. Generally, we use underscores to flag things that are private to a namespace (because C forces that on us), and for structures that we'll never reference by that name:
```
WT_ITEM _foo, *foo;
foo = &_foo;    /* We never use _foo itself. */
```
but I don't see how either of those makes sense here? I'd strongly prefer something like `las_txnid_tmp`, or `las_txnid_notused`, or something, but underscore prefix doesn't tell me anything useful.

(To be fair, I was one of the people that started the whole `WT_ITEM _foo, *foo;` stuff in the code base, and I regret it now. )

src/btree/bt_ret.c
```diff
@@ -175,14 +183,7 @@ __wt_value_return_upd(
...
              */
-            if (i >= WT_MODIFY_ARRAY_SIZE) {
-                if (i == WT_MODIFY_ARRAY_SIZE)
-                    listp = NULL;
-                WT_ERR(__wt_realloc_def(session, &allocated_bytes, i + 1, &listp));
-                if (i == WT_MODIFY_ARRAY_SIZE)
-                    memcpy(listp, list, sizeof(list));
-            }
-            listp[i++] = upd;
+            WT_ERR(__wt_modify_vector_push(&modifies, upd));
```
keithbostic: The associated comment no longer makes sense.

src/btree/bt_read.c
```diff
@@ -332,30 +407,43 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
-                }
-                WT_ERR(__wt_update_alloc(
-                  session, &las_value, &listp[mod_counter++], &notused, upd_type));
-                WT_ERR(cursor->prev(cursor));
-                WT_ERR(cursor->get_value(
-                  cursor, &durable_timestamp, &prepare_state, &upd_type, &las_value));
+                WT_ERR(__wt_update_alloc(session, &las_value, &mod_upd, &notused, upd_type));
+                WT_ERR(__wt_modify_vector_push(&modifies, mod_upd));
+                mod_upd = NULL;
+                WT_ERR(las_cursor->prev(las_cursor));
```
keithbostic: Isn't the beginning of the table (`WT_NOTFOUND`) also a boundary condition here?

src/btree/bt_read.c
```diff
@@ -428,21 +516,21 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
-    while (mod_counter > 0)
-        __wt_free_update_list(session, listp[--mod_counter]);
-    if (allocated_bytes != 0)
-        __wt_free(session, listp);
+    __wt_free_update_list(session, mod_upd);
+    while (modifies.size > 0) {
+        __wt_modify_vector_pop(&modifies, &mod_upd);
+        __wt_free_update_list(session, mod_upd);
+    }
+    __wt_modify_vector_free(&modifies);
```
keithbostic: Would it be better for `__wt_modify_vector_free()` to free any remaining individual `WT_UPDATE` items as well?

src/btree/bt_read.c
```diff
@@ -152,6 +152,78 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
     return (ret);
 }
 
+/*
+ * __wt_modify_vector_init --
+ *     Initialise a modify vector.
+ */
+void
+__wt_modify_vector_init(WT_MODIFY_VECTOR *modifies, WT_SESSION_IMPL *session)
```
keithbostic: It would be nice if these weren't in `bt_read.c`;  maybe `support/modify.c` or `include/misc.i`?

Ignore if you like it where it is.

src/cache/cache_las.c
```diff
@@ -1462,40 +1478,64 @@ __wt_find_lookaside_upd(
...
-                    if (i == WT_MODIFY_ARRAY_SIZE)
-                        memcpy(listp, list, sizeof(list));
-                }
-                listp[i++] = upd;
-                WT_ERR(cursor->prev(cursor));
-#ifdef HAVE_DIAGNOSTIC
+                WT_ERR(__wt_update_alloc(session, las_value, &mod_upd, &notused, upd_type));
+                WT_ERR(__wt_modify_vector_push(&modifies, mod_upd));
+                mod_upd = NULL;
+                WT_ERR(las_cursor->prev(las_cursor));
```
keithbostic: Another spot where I'm wondering if `WT_NOTFOUND` is a boundary condition rather than an error.

src/btree/bt_read.c
```diff
@@ -152,6 +152,78 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
...
+            migrate_from_stack = true;
+            modifies->listp = NULL;
+        }
+        WT_ERR(__wt_realloc_def(
+          modifies->session, &modifies->capacity, modifies->size + 1, &modifies->listp));
+        if (migrate_from_stack)
+            memcpy(modifies->listp, modifies->list, sizeof(modifies->list));
+    }
+    modifies->listp[modifies->size++] = upd;
+    return (0);
```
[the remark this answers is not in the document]
tetsuo-cpp: I did it like this because you might have a case where you might have a `push`, `pop`, `push` pattern where you cross the `WT_MODIFY_VECTOR_STACK_SIZE` multiple times.

src/btree/bt_read.c
tetsuo-cpp: I considered this but there are some usages that store pointers to items in the update list. In that case, it's not ours to free.

src/btree/bt_read.c
tetsuo-cpp: abcd390

src/include/btmem.h
```diff
@@ -1120,17 +1120,31 @@ struct __wt_update {
...
+ * 	A resizable array for storing modify updates. The allocation strategy is similar to that of
+ *	llvm::SmallVector<T> where we keep space on the stack for the regular case but fall back to
+ *	dynamic allocation as needed.
+ */
+struct __wt_modify_vector {
+    WT_SESSION_IMPL *session;
+    WT_UPDATE *list[WT_MODIFY_VECTOR_STACK_SIZE];
+    WT_UPDATE **listp;
+    size_t capacity;
+    size_t size;
```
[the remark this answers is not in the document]
tetsuo-cpp: abcd390

src/btree/bt_read.c
```diff
@@ -152,6 +152,78 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
...
+int
+__wt_modify_vector_push(WT_MODIFY_VECTOR *modifies, WT_UPDATE *upd)
+{
+    WT_DECL_RET;
+    bool migrate_from_stack;
+
+    migrate_from_stack = false;
+
+    if (modifies->size >= WT_MODIFY_VECTOR_STACK_SIZE) {
+        if (modifies->capacity == 0 && modifies->size == WT_MODIFY_VECTOR_STACK_SIZE) {
```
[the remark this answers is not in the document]
tetsuo-cpp: abcd390

The name is hopefully more clear now.

src/btree/bt_read.c
```diff
@@ -152,6 +152,78 @@ __instantiate_birthmarks(WT_SESSION_IMPL *session, WT_REF *ref)
...
+        if (migrate_from_stack)
+            memcpy(modifies->listp, modifies->list, sizeof(modifies->list));
+    }
+    modifies->listp[modifies->size++] = upd;
+    return (0);
+
+err:
+    if (modifies->listp == NULL) {
+        modifies->listp = modifies->list;
+        modifies->capacity = 0;
```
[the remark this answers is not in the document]
tetsuo-cpp: I'm unsure what you mean. If we jump to `err` it means we failed to allocate. I just set `allocated_bytes` to 0 to ensure that we don't think we have heap memory to free when we call `__wt_modify_vector_free`. I don't think we need to do anything with `size` since we wouldn't have incremented it yet. The struct should look the way it did before the `push` call.

That was what I was aiming for when I wrote this. Does something seem fishy here?

src/btree/bt_ret.c
tetsuo-cpp: 71f63db

src/btree/bt_read.c
tetsuo-cpp: 885a5d8

src/btree/bt_read.c
tetsuo-cpp: eff57c7

I've given everything a `_tmp` suffix. I also changed `__wt_find_lookaside_upd` which is where I started getting into this habit.

src/btree/bt_read.c
tetsuo-cpp: 1dc2b0c

Thanks for picking up on that. I've added some logic to treat this case the same.

src/cache/cache_las.c
tetsuo-cpp: e03abe2

src/cache/cache_las.c
tetsuo-cpp: 428f200

tetsuo-cpp:
@kommiharibabu @keithbostic Thanks for the reviews. Could you please take a look at my responses to your comments? Anything that wasn't 100% trivial, I've left unresolved. If you're still happy can you mark them as resolved to indicate that your previous LGTM still stands?

src/btree/bt_read.c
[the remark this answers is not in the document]
keithbostic: > I don't think we need to do anything with size since we wouldn't have incremented it yet. 

But on the second & subsequent calls to realloc, won't size be non-zero? IOW, the 2nd time we call realloc, it fails.

src/btree/bt_read.c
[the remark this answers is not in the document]
keithbostic: I agree with your argument.

keithbostic:
@tetsuo-cpp, my only remaining issue is I believe the `err:` lable in `__wt_modify_vector_push` needs to fix `size`. (It's not an issue now, but might some day.)

I'm still good to go, lgtm.

src/btree/bt_read.c
```diff
@@ -166,17 +166,19 @@ __las_page_instantiate(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    WT_ITEM las_key, las_key_tmp, las_value, next_las_key;
+    WT_MODIFY_VECTOR modifies;
     WT_PAGE_LOOKASIDE *page_las;
     WT_PAGE *page;
-    WT_UPDATE *list[WT_MODIFY_ARRAY_SIZE], **listp, *upd;
-    wt_timestamp_t durable_timestamp, las_timestamp;
-    size_t allocated_bytes, notused, size, total_incr;
-    uint64_t birthmark_cnt, las_txnid, recno;
+    WT_UPDATE *mod_upd, *upd;
+    wt_timestamp_t durable_timestamp, durable_timestamp_tmp;
```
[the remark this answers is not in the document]
tetsuo-cpp: 5c62c4e

src/btree/bt_read.c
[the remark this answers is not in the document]
tetsuo-cpp: 96f3eeb

I still don't see the problem. I've added a comment explaining what I'm trying to do. If there's still an issue here let me know and I can fix it in a subsequent change.

PR #4935, opened 2019-10-22
WT-4986 Migrate wiredtiger-linux-directio

quchenhao:


test/evergreen.yml
```diff
@@ -1264,6 +1264,22 @@ tasks:
             set -o verbose
             sh dist/s_clang-scan 2>&1
 
+  - name: directio
```
lukech: Prefer retaining `linux-` in the task name.

test/evergreen.yml
```diff
@@ -1264,6 +1264,22 @@ tasks:
...
+    depends_on:
+    - name: compile
+    commands:
+      - func: "fetch artifacts"
+      - func: "compile wiredtiger"
```
lukech: Jenkins job has the configure options `../configure --enable-diagnostic --enable-strict`. Not sure if other options (as part of the standard `compile wiredtiger` func will make any difference for the test output. Adding @ddanderson for comment.

test/evergreen.yml
```diff
@@ -1264,6 +1264,22 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix/test/format"
+          script: |
+            set -o errexit
+            set -o verbose
+            for i in {1..3}; do
+              ./t -1 -C "direct_io=[data]" -c ../../../test/format/CONFIG.stress || cleanup
```
lukech: `|| cleanup` is a legacy for Jenkins, which is not needed here.

test/evergreen.yml
quchenhao: I think other things are just extras.

PR #4936, opened 2019-10-22
WT-5042 Avoid calling config functions during checkpoint.

sueloverso:


sueloverso:
@keithbostic please review and think about any bad effects of this. In particular, this change concatenates the `checkpoint=,checkpoint_lsn` onto the end of the metadata string which is now out of alphabetical order. I don't know if that will cause any problems in the config code and if it makes assumptions about the order of items.

The amount of code changes here are not reflective of the time and difficulty of getting this working.

src/meta/meta_ckpt.c
```diff
@@ -106,25 +106,17 @@ __wt_meta_checkpoint_clear(WT_SESSION_IMPL *session, const char *fname)
...
-    WT_ERR(__wt_metadata_search(session, fname, &config));
-
-    /* Replace the checkpoint entry. */
-    cfg[0] = config;
-    cfg[1] = v == NULL ? "checkpoint=()" : v;
-    cfg[2] = NULL;
-    WT_ERR(__wt_config_collapse(session, cfg, &newcfg));
-    WT_ERR(__wt_metadata_update(session, fname, newcfg));
+    WT_ERR(__wt_scr_alloc(session, 0, &tmp));
+    WT_ASSERT(session, strcmp(session->dhandle->name, fname) == 0);
```
keithbostic: I don't see the point of this assertion? It's not a requirement they be the same, is it?

src/meta/meta_ckpt.c
```diff
@@ -106,25 +106,17 @@ __wt_meta_checkpoint_clear(WT_SESSION_IMPL *session, const char *fname)
...
+    WT_ERR(__wt_buf_fmt(session, tmp, "%s", session->dhandle->meta_base));
+    WT_ERR(__wt_buf_catfmt(session, tmp, ",%s", v == NULL ? "checkpoint=()," : v));
```
keithbostic: There's no need for two calls, just do `"%s,%s"`.

src/conn/conn_dhandle.c
```diff
@@ -48,6 +51,9 @@ __conn_dhandle_config_set(WT_SESSION_IMPL *session)
             ret = __wt_set_return(session, ENOENT);
         WT_RET(ret);
     }
+    cfg[0] = metaconf;
+    cfg[1] = NULL;
+    cfg[2] = NULL;
```
keithbostic: We don't need this initialization in the case of `WT_DHANDLE_TYPE_TABLE`.

src/conn/conn_dhandle.c
```diff
@@ -64,16 +70,28 @@ __conn_dhandle_config_set(WT_SESSION_IMPL *session)
     switch (dhandle->type) {
     case WT_DHANDLE_TYPE_BTREE:
         WT_ERR(__wt_strdup(session, WT_CONFIG_BASE(session, file_meta), &dhandle->cfg[0]));
+        WT_ASSERT(session, dhandle->meta_base == NULL);
+        /* Overwrite any checkpoint info. */
+        cfg[1] = "checkpoint=()";
```
keithbostic: I'm not sure why this is necessary -- collapse doesn't discard anything, it just gets you a single set of items? IOW, doesn't the subsequent discard take care of any existing checkpoint info?

src/conn/conn_dhandle.c
```diff
@@ -64,16 +70,28 @@ __conn_dhandle_config_set(WT_SESSION_IMPL *session)
...
+        cfg[1] = NULL;
+        WT_ERR(__wt_config_merge(session, cfg, "checkpoint=,checkpoint_lsn=", &base));
+        __wt_free(session, tmp);
         break;
     case WT_DHANDLE_TYPE_TABLE:
         WT_ERR(__wt_strdup(session, WT_CONFIG_BASE(session, table_meta), &dhandle->cfg[0]));
         break;
     }
     dhandle->cfg[1] = metaconf;
+    dhandle->meta_base = base;
```
keithbostic: Have we confirmed that (except for periodic checkpoints), the file metadata never changes?

src/meta/meta_ckpt.c
```diff
@@ -236,37 +228,28 @@ __wt_meta_block_metadata(WT_SESSION_IMPL *session, const char *config, WT_CKPT *
...
      */
     WT_ERR(__wt_btree_config_encryptor(session, filecfg, &kencryptor));
     if (kencryptor == NULL) {
-        metadata = min_config;
-        metadata_len = strlen(min_config);
+        metadata = config;
+        metadata_len = strlen(config);
     } else {
-        WT_ERR(__wt_buf_set(session, a, min_config, strlen(min_config)));
+        WT_ERR(__wt_buf_set(session, a, config, strlen(config)));
```
keithbostic: Given the cost of encryption, it seems like we only we should only encrypt once -- perhaps `__conn_dhandle_config_set()` should call one routine that both fills in the base configuration value for the handle, and additionally encrypts a copy for later use here, and both strings are maintained in the dhandle for the life of the handle.

That also gets us away from using the current metadata configuration for anything other than retrieving checkpoint information -- in other words, if we believe the base configuration can't change, let's double-down on that belief.

src/conn/conn_dhandle.c
sueloverso: No. This specific piece of code was the most difficult to work out in fact. The `wt_config_merge` call below with `checkpoint=,...` in the `strip` arg will not actually strip out real checkpoint information. That code requires and only strips out specific elements when a subcategory is involved. So for an actual checkpoint record, that would need the exact checkpoint name and each piece. For example, `checkpoint=(WiredTigerCheckpoint.37=(addr=))`. That wasn't going to happen. I tried for a while to get the merge/strip code to remove the entire category, but it started getting messy in lower levels of the config parsing and I never got it working. Given how I can simply make it work with the two calls, those code changes did not seem worth debugging or taking more time.

src/conn/conn_dhandle.c
sueloverso: The only way I know of is via alter. Do you know of any other paths? Or do you know of any other single path where I could confirm that?

src/meta/meta_ckpt.c
sueloverso: I am using the metadata string from the dhandle. I wanted to be sure that the metadata I'm using comes from the file we are checkpointing.

dist/s_funcs.list
```diff
@@ -12,6 +12,7 @@ __wt_bloom_get
 __wt_bulk_insert_fix
 __wt_bulk_insert_row
 __wt_bulk_insert_var
+__wt_config_discard_defaults
```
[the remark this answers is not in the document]
sueloverso: Fixed.

src/meta/meta_ckpt.c
sueloverso: Fixed.

src/conn/conn_dhandle.c
sueloverso: Moved.

src/conn/conn_dhandle.c
keithbostic: I agree, the only way I know is via `alter`.

src/meta/meta_ckpt.c
sueloverso: In `ckpt_set` we (after debugging) retain the old path and only use the faster path for checkpoints. There are a number of times and places where we cannot use that base and need that.

But WRT encryption, my changes do not change anything regarding encryption, only the string we're encrypting. However, I'm not following your suggestion. The base string stored in the dhandle alone (i.e. without the checkpoint and checkpoint_lsn elements) would encrypt differently than the resulting string with that base and those items concatenated to its end. It's not like we can just encrypt the `checkpoint=...,checkpoint_lsn` piece separately and concatenate that to the encrypted base-only string. We need to encrypt the whole string at once.

sueloverso:
@agorrod and @keithbostic this is finally passing PR testing. Please review.

src/conn/conn_dhandle.c
```diff
@@ -63,17 +67,31 @@ __conn_dhandle_config_set(WT_SESSION_IMPL *session)
     WT_ERR(__wt_calloc_def(session, 3, &dhandle->cfg));
     switch (dhandle->type) {
     case WT_DHANDLE_TYPE_BTREE:
+        cfg[0] = metaconf;
+        cfg[1] = "checkpoint=()";
+        cfg[2] = NULL;
         WT_ERR(__wt_strdup(session, WT_CONFIG_BASE(session, file_meta), &dhandle->cfg[0]));
+        WT_ASSERT(session, dhandle->meta_base == NULL);
```
agorrod: I'd like a comment here describing what we are saving in meta_base and why. Something like: "Save the metadata string that we know won't change between subsequent checkpoints, to save calculating it repeatedly over the lifetime of the handle"

sueloverso: I pushed a large comment. Give it a read. If it's okay I'll merge this branch on Monday since I'm away the whole weekend, I don't want to possibly break the world.

PR #4939, opened 2019-10-23
WT-5157 Fix atomics usage in spinlock impl

tetsuo-cpp:
This bug was introduced in #4740. I didn't notice the deadlock at the time since I didn't run the Python test suite with `--with-spinlock=gcc`.

> The byte is set to some implementation defined nonzero “set” value and the return value is true if and only if the previous contents were “set”

The `__atomic_test_and_set` compiler intrinsic is similar to its `__sync_*` counterpart in that it returns a `bool` for the previous value before the call took place (not for success like I thought). So actually a successful acquisition should return `false` since no other thread should have set it to its "on" value.

### DIFF
--- a/src/include/mutex.h
+++ b/src/include/mutex.h
@@ -100,6 +100,7 @@ struct __wt_spinlock {
 #if SPINLOCK_TYPE == SPINLOCK_GCC
     WT_CACHE_LINE_PAD_BEGIN
     volatile bool lock;
+    uint8_t unused[7];
 #elif SPINLOCK_TYPE == SPINLOCK_PTHREAD_MUTEX || \
   SPINLOCK_TYPE == SPINLOCK_PTHREAD_MUTEX_ADAPTIVE || SPINLOCK_TYPE == SPINLOCK_MSVC
     wt_mutex_t lock;
--- a/src/include/mutex.h
+++ b/src/include/mutex.h
@@ -99,7 +99,7 @@ struct __wt_rwlock { /* Read/write lock */
 struct __wt_spinlock {
 #if SPINLOCK_TYPE == SPINLOCK_GCC
     WT_CACHE_LINE_PAD_BEGIN
-    volatile int lock;
+    volatile bool lock;
 #elif SPINLOCK_TYPE == SPINLOCK_PTHREAD_MUTEX || \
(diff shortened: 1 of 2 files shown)


src/include/mutex.h
```diff
@@ -99,7 +99,7 @@ struct __wt_rwlock { /* Read/write lock */
 struct __wt_spinlock {
 #if SPINLOCK_TYPE == SPINLOCK_GCC
     WT_CACHE_LINE_PAD_BEGIN
-    volatile int lock;
+    volatile bool lock;
```
tetsuo-cpp: From the GCC documentation:

> It should be only used for operands of type bool or char. For other types only part of the value may be set.

According to [this](https://github.com/wiredtiger/wiredtiger/pull/4740#discussion_r305622159) comment from @keithbostic , we probably don't need to worry about this in WiredTiger. But I don't believe there's any reason not to use `bool` so let's just follow their advice.

tetsuo-cpp:
retest this please

src/include/mutex.h
sueloverso: I'm neutral on this. It probably does not change anything because this is the beginning of the structure and we're using `WT_CACHE_LINE_PAD_BEGIN` to start it aligned. Then the next item after `lock` is a pointer, `const char *name` which will be 8 bytes and aligned anyway. So we still have that 7 bytes. If we care about bytes here or we think we'd ever want any other debug or info, you could add in a `uint8_t unused[7];` to make things very clear.

tetsuo-cpp: 91f4d23

Yeah agreed. It almost certainly isn't a problem as the functions are likely reading and writing the first byte of the `int`. But I figure we may as well follow their advice since it explicitly says it should only be used for `bool` or `char`.

tetsuo-cpp:
@sueloverso I was able to reproduce the hang and confirm that this patch fixes it. It happens very consistently since any thread that successfully acquires with `__wt_try_spinlock` will never let go of the lock since it thinks it failed to acquire and any other thread (including the one that successfully acquired it) that ends up spinning for that lock will spin indefinitely as a result.

sueloverso:
This still LGTM to me. Merge when ready.

PR #4940, opened 2019-10-23
WT-4976 Migrate wiredtiger-pull-request-builds

quchenhao:


test/evergreen.yml
```diff
@@ -154,6 +154,25 @@ tasks:
           configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ CFLAGS="-fsanitize=undefined -ggdb"
           posix_configure_flags: --enable-silent-rules --enable-strict --enable-diagnostic
 
+  # build with --disable-shared, without --enable-diagnostic
+  # the compiler builds turn on diagnostics
```
lukech: This comment confused me. Lets clarify it with team.

test/evergreen.yml
```diff
@@ -154,6 +154,25 @@ tasks:
...
+  - name: compile-disable-shared-without-enable-diagnostic
+    tags: ["pull_request"]
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-silent-rules --enable-strict --disable-shared
```
lukech: Considering we have many versions of compile tasks, I think it may worth investigating whether we can abstract out those common parts. I know mongo uses task templates, not sure if it helps here.

test/evergreen.yml
```diff
@@ -154,6 +154,25 @@ tasks:
...
+
+  # build with --disable-static, without --enable-diagnostic
+  # the compiler builds turn on diagnostics
+  - name: compile-disable-static-without-enable-diagnostic
```
lukech: I feel the `-without-enable-diagnostic` part could be saved for the task name, not strongly opinioned though.

test/evergreen.yml
quchenhao: updated

test/evergreen.yml
quchenhao: Looked into it. It's not helpful in this case because if you want to override the vars, you have to override the whole commands section which is the same here.

test/evergreen.yml
quchenhao: updated

PR #4941, opened 2019-10-24
WT-5195 Reduce Python unit test build time for Evergreen Windows build variant

lukech:
Reduce the runtime for Evergreen Windows build variant, unit test bucket 01 and 06 (the bottlenecks), by dedicating the 2 buckets for `compat` and `timestamp` tests, moving the other tests of `c` and `t` families into other buckets. 

After this change, bucket 01 and 06 both had less than 9m runtime for the Windows build variant. The ubuntu1804 build variant had a similar evenly distribution. 

Patch build: https://evergreen.mongodb.com/version/5db12ad15623437f20d25a7f

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -914,7 +914,8 @@ tasks:
             set -o errexit
             set -o verbose
 
-            ${test_env_vars|} ${python_binary|python} ../test/suite/run.py [c] -v 2 ${smp_command|} 2>&1
+            # Reserve this bucket only for compat tests, which take a long time to run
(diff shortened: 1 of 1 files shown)


lukech:
@quchenhao I abstracted the file query pipe command into a shell variable, and intentionally avoiding using `{}` while quoting the variable to avoid being treated as a build variant expansion by Evergreen. I think the change makes the code more organized. Would you mind to take a quick look?

The patch build: https://evergreen.mongodb.com/version/5db2221a32f4173e9267b64c

quchenhao:
lgtm

PR #4942, opened 2019-10-25
WT-5201 Update test format in mongodb-4.0 branch

luke-pearson:
So I had to cherry pick the WT-4844 commit as it affects the way wiredtiger logs a message for read timestamps being before oldest timestamp. This then gets caught and ignored in test/format.

I'll run this on a branch over the weekend to see that we don't get any unexpected failures. Once this is merged we can also close out WT-5188.


luke-pearson:
@keithbostic, I had to backport an additional fix for an assertion that is triggered by the newer test format. Additionally I turned off the assert timestamp functionality but left it in for potential future use. I generally wanted to avoid backporting fixes.

Please do another quick review, thanks.

PR #4943, opened 2019-10-28
WT-5120 checkpoint hangs when reconciliation doesn't release the eviction generation

keithbostic:
@bvpvamsikrishna, please feel free to change the reviewers on this ticket, I picked @sueloverso because she looked at a possibly related failure in WT-5043, and Alex because he's been in some of this code before.

Reviewers: I'm going to post some discussion in WT-5120.

The changes in 3c2a6a3 are semantic sugar, just getting ready for the change in 62391cc which is the real fix.

### DIFF
--- a/src/btree/bt_curnext.c
+++ b/src/btree/bt_curnext.c
@@ -159,7 +159,7 @@ __cursor_var_append_next(WT_CURSOR_BTREE *cbt, bool newpage, bool restart)
                 ++cbt->page_deleted_count;
             continue;
         }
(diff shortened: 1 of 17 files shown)


keithbostic:
> This was changed just a month ago from update to insert (by you) in WT-5135. Now it's going back to update. Just confirming that the use of the new flag allows use of the pinned page, which was not getting used in the previous update code due to using set_key (per WT-5135).

Yes, that is correct.

PR #4945, opened 2019-10-28
WT-5204 Add diagnostic logging code to the generation drain function.

keithbostic:


keithbostic:
> Let the WT-5120 get merged (with the debugging from WT-5043 that is still in there) and let it run for several days on Jenkins to make sure the problem is fixed and it isn't detecting any other issues. We've just been assuming every "operation timeout" was that problem. If all is well, then we merge this.

That seems simplest.

keithbostic:
@sueloverso, based on WT-5207, I've reverted so the WT-5043 debugging remains, and now this PR is just the addition of generation logging. I'll merge it if you approve.

src/support/generation.c
```diff
@@ -122,6 +148,23 @@ __wt_gen_drain(WT_SESSION_IMPL *session, int which, uint64_t generation)
...
+                minutes = 1;
+                __wt_seconds(session, &start);
+            } else {
+                __wt_seconds(session, &stop);
+                if (stop - start > minutes * WT_MINUTE) {
+                    WT_IGNORE_RET(__wt_msg(session, "%s generation drain waited %u minutes",
+                      __gen_name(which), minutes));
+                    ++minutes;
+                }
+                WT_ASSERT(session, stop - start < 3 * WT_MINUTE);
```
sueloverso: My only comment here is that the WT-5043 debugging fails after 5 minutes, this is 3 minutes. Both of those seem fairly random as "a long time for code but not super long for a person". Both should be the same and use a #define? If you agree, some comments and messages need to change from having the hard-coded number.

keithbostic: Since the WT-5043 debugging is going to go away, is there any need to create a shared #define?

sueloverso: I don't feel strongly. Just a point for thought. Feel free to merge as-is.

PR #4949, opened 2019-11-01
WT-5032 Migrate wiredtiger-syscall-linux to evergreen

quchenhao:


test/evergreen.yml
```diff
@@ -1420,6 +1434,19 @@ buildvariants:
     - name: ".unit_test"
     - name: conf-dump-test
 
+- name: syscall-linux
+  display_name: Syscall linux
```
lukech: Can we make the disaply name `Linux syscall`, or `Syscall Linux`?

PR #4950, opened 2019-11-01
WT-5031 Migrate wiredtiger-nightly-package to evergreen

quchenhao:


test/evergreen.yml
```diff
@@ -1384,6 +1384,17 @@ tasks:
               ./t -1 -C "direct_io=[data]" -c ../../../test/format/CONFIG.stress
             done
 
+  - name: release
```
lukech: I'd prefer to retain the original name `nightly-package`, or `package` in case we need to adjust the running frequency later on.

PR #4953, opened 2019-11-04
WT-5218 Add function to determine whether a page should be evicted

luke-pearson:


src/include/btree.i
```diff
@@ -1394,6 +1394,23 @@ __wt_page_can_evict(WT_SESSION_IMPL *session, WT_REF *ref, bool *inmem_splitp)
...
+/*
+ * __wt_should_evict_page --
+ *     Determine whether a page should be evicted.
+ */
+static inline bool
+__wt_should_evict_page(WT_PAGE *page)
+{
+    switch (page->read_gen) {
+    case WT_READGEN_OLDEST:
+        return false;
```
michaelcahill: I know I've been away, but we still require `return (x)`, right?

src/include/btree.i
```diff
@@ -1435,7 +1452,7 @@ __wt_page_release(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags)
      * evict it.
      */
     page = ref->page;
-    if (WT_READGEN_EVICT_SOON(page->read_gen) && btree->evict_disabled == 0 &&
```
michaelcahill: Can the `WT_READGEN_EVICT_SOON` macro go away?  It seems confusing and error prone to keep two different tests...

luke-pearson:
retest this please

src/include/btree.i
luke-pearson: It's used in at least 5 other places so I'll have to go through each scenario and determine if that's a valid use case for the new function.

luke-pearson: There are scenarios where we only have the readgen available, i.e. no page reference and use the macro. In which case the function wouldn't suit, I could change it to handle that case too but I think it would be best to keep both?

src/include/btree.i
```diff
@@ -489,6 +489,8 @@ __wt_page_only_modify_set(WT_SESSION_IMPL *session, WT_PAGE *page)
     if (page->modify->page_state < WT_PAGE_DIRTY &&
       __wt_atomic_add32(&page->modify->page_state, 1) == WT_PAGE_DIRTY_FIRST) {
         __wt_cache_dirty_incr(session, page);
+        if (page->read_gen == WT_READGEN_WONT_NEED)
+            __wt_cache_read_gen_new(session, page);
```
luke-pearson: There was a choice here to move it to WT_READGEN_NOTSET or give it a new readgen immediately, I think this way is better as generally we expect pages to have a valid readgen?

agorrod: Did you consider using __wt_cache_read_gen_bump instead? I'm really only asking because you said there is a 10% regression remaining, and I wonder if setting the read generation to current rather than middle of the road would help.

src/include/btree.i
```diff
@@ -489,6 +489,8 @@ __wt_page_only_modify_set(WT_SESSION_IMPL *session, WT_PAGE *page)
     if (page->modify->page_state < WT_PAGE_DIRTY &&
       __wt_atomic_add32(&page->modify->page_state, 1) == WT_PAGE_DIRTY_FIRST) {
         __wt_cache_dirty_incr(session, page);
+        if (page->read_gen == WT_READGEN_WONT_NEED)
```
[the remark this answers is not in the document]
agorrod: The comment you added has a good level of information, but would be better if it didn't reference the name of the define. We try to use regular words in comments rather than variable and definition names - it forces the comment to be an abstraction rather than a statement of what the code is doing. You'll be able to find a ton of counter examples, but capturing this sort of thing is why we use the spell checker on comments in s_string.

PR #4954, opened 2019-11-04
WT-4987 Migrate Jenkins test-checkpoint job to Evergreen

raviprakashgiri29:
Added a task "test-checkpoint" & respective function "test checkpoint for filetype" in the evergreen.yml file to migrate "wiredtiger-test-checkpoint" job from jenkins to evergreen.

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -115,6 +115,17 @@ functions:
         rm -rf "wiredtiger"
         rm -rf "wiredtiger.tgz"
 
+  "test checkpoint for filetype":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/build_posix"
+      script: |
+        set -o errexit
+        set -o verbose
+        for fileType in m r c; do
+          ./test/checkpoint/t -t $fileType -n 1000000 -k 5000000 -C cache_size=100MB 2>&1
+        done
+
 pre:
   - func: "cleanup"
 post:
@@ -1384,6 +1395,15 @@ tasks:
               ./t -1 -C "direct_io=[data]" -c ../../../test/format/CONFIG.stress
             done
 
+  - name: test-checkpoint
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ CFLAGS="-ggdb"
+          posix_configure_flags: --enable-strict
+      - func: "test checkpoint for filetype"
+
 buildvariants:
 - name: ubuntu1804
   display_name: Ubuntu 18.04
@@ -1403,6 +1423,7 @@ buildvariants:
     - name: linux-directio
     - name: make-check-asan-test
     - name: configure-combinations
+    - name: test-checkpoint
 
 - name: ubuntu1804-python3
   display_name: Ubuntu 18.04 (Python3)
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -122,8 +122,7 @@ functions:
       script: |
         set -o errexit
         set -o verbose
-        ./t ${testParams} 2>&1
-
+        ./t ${checkpointArgs} 2>&1
 
 pre:
   - func: "cleanup"
@@ -1403,13 +1402,13 @@ tasks:
           posix_configure_flags: --enable-strict
       - func: "checkpoint test"
         vars:
-          testParams: "-t m -n 1000000 -k 5000000 -C cache_size=100MB"
+          checkpointArgs: -t m -n 1000000 -k 5000000 -C cache_size=100MB
       - func: "checkpoint test"
         vars:
-          testParams: "-t r -n 1000000 -k 5000000 -C cache_size=100MB"
+          checkpointArgs: -t r -n 1000000 -k 5000000 -C cache_size=100MB
       - func: "checkpoint test"
         vars:
-          testParams: "-t c -n 1000000 -k 5000000 -C cache_size=100MB"
+          checkpointArgs: -t c -n 1000000 -k 5000000 -C cache_size=100MB
 
 buildvariants:
 - name: ubuntu1804
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -115,16 +115,15 @@ functions:
         rm -rf "wiredtiger"
         rm -rf "wiredtiger.tgz"
 
-  "test checkpoint for filetype":
+  "test checkpoint":
     command: shell.exec
     params:
-      working_dir: "wiredtiger/build_posix"
+      working_dir: "wiredtiger/build_posix/test/checkpoint"
       script: |
         set -o errexit
         set -o verbose
-        for fileType in m r c; do
-          ./test/checkpoint/t -t $fileType -n 1000000 -k 5000000 -C cache_size=100MB 2>&1
-        done
+        ./t -t ${fileType} -n 1000000 -k 5000000 -C cache_size=100MB 2>&1
+
 
 pre:
   - func: "cleanup"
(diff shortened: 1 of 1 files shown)


wtbuild:

test/evergreen.yml
```diff
@@ -115,6 +115,17 @@ functions:
...
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/build_posix"
+      script: |
+        set -o errexit
+        set -o verbose
+        for fileType in m r c; do
+          ./test/checkpoint/t -t $fileType -n 1000000 -k 5000000 -C cache_size=100MB 2>&1
+        done
+
```
quchenhao: I think you don't need to extract this as a function as it is only used once.

test/evergreen.yml
```diff
@@ -1384,6 +1395,15 @@ tasks:
...
             done
 
+  - name: test-checkpoint
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ CFLAGS="-ggdb"
+          posix_configure_flags: --enable-strict
+      - func: "test checkpoint for filetype"
```
quchenhao: You can move the function to here

test/evergreen.yml
quchenhao: I see why you did this. I prefer it like this
./test/checkpoint/t ${checkout_args| some default value} 2>&1
and call this three times in the job

test/evergreen.yml
quchenhao: Ignore this.

test/evergreen.yml
```diff
@@ -115,6 +115,17 @@ functions:
         rm -rf "wiredtiger"
         rm -rf "wiredtiger.tgz"
 
+  "test checkpoint for filetype":
```
quchenhao: Just test checkpoint is better I think

test/evergreen.yml
quchenhao: we can set the working dir to wiredtiger/build_posix/test/checkpoint

raviprakashgiri29: Fixed. Calling the function with different arguments.

test/evergreen.yml
raviprakashgiri29: Fixed. Calling function with different arguments.

test/evergreen.yml
raviprakashgiri29: Fixed .  Renamed to "test checkpoint"

test/evergreen.yml
```diff
@@ -115,6 +115,16 @@ functions:
...
         rm -rf "wiredtiger.tgz"
 
+  "test checkpoint":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/build_posix/test/checkpoint"
+      script: |
+        set -o errexit
+        set -o verbose
+        ./t -t ${fileType} -n 1000000 -k 5000000 -C cache_size=100MB 2>&1
```
quchenhao: I think we should wrap this whole part in a variable
-t ${fileType} -n 1000000 -k 5000000 -C cache_size=100MB

test/evergreen.yml
```diff
@@ -1384,6 +1394,23 @@ tasks:
               ./t -1 -C "direct_io=[data]" -c ../../../test/format/CONFIG.stress
             done
 
+  - name: test-checkpoint
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ CFLAGS="-ggdb"
```
lukech: Is there any special consideration to use `clang` here? 
It would be good to add a comment to declare the reason why not enabling diagnostic - similar as what's being done in the current Jenkins shell.

test/evergreen.yml
```diff
@@ -1384,6 +1394,23 @@ tasks:
               ./t -1 -C "direct_io=[data]" -c ../../../test/format/CONFIG.stress
             done
 
+  - name: test-checkpoint
```
lukech: In current evergreen.yml there's another task called `checkpoint-test`. It would be nice to have a more specific name for this new task to avoid being confused with the other. I vote for `test-checkpoint-filetypes` :)

quchenhao: Would the convention be checkout-filetypes-test?

test/evergreen.yml
raviprakashgiri29: > Do you have a patch build with a run of the newly added `test-checkpoint` task?
> Alternatively, you can schedule the new task in the pull request triggered patch build.

https://evergreen.mongodb.com/version/5dc0c42f3627e02ee6bf7ca9 is the patch build for the newly added `test-checkpoint` task.

test/evergreen.yml
raviprakashgiri29: I used "test-checkpoint" task name to match it with the existing jenkins job ("wiredtiger-test-checkpoint"). Renaming this task to "test-checkpoint-filetypes" and respective function to "test checkpoint parameters". Please suggest for better name if any. . :)

lukech: Actually, `checkpoint-filetypes-test` sounds better to me.

test/evergreen.yml
lukech: Thanks.

test/evergreen.yml
raviprakashgiri29: Renaming the task to "checkpoint-filetypes-test" and respective function to "checkpoint filetypes test params"

test/evergreen.yml
raviprakashgiri29: The diagnostic is not enabled in the Jenkins job as well. Added the comment for not including diagnostic.

test/evergreen.yml
raviprakashgiri29: > I think we should wrap this whole part in a variable
> -t ${fileType} -n 1000000 -k 5000000 -C cache_size=100MB

Wrapped the whole parameters into a variable "$testParams"

test/evergreen.yml
```diff
@@ -115,6 +115,16 @@ functions:
...
+  "checkpoint test":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/build_posix/test/checkpoint"
+      script: |
+        set -o errexit
+        set -o verbose
+        ./t ${testParams} 2>&1
+
+
```
quchenhao: Remove extra empty line

test/evergreen.yml
```diff
@@ -1384,6 +1394,23 @@ tasks:
...
+  - name: checkpoint-filetypes-test
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          # Don't use diagnostic - this test looks for timing problems that are more likely to occur without it
+          posix_configure_flags: --enable-strict
+      - func: "checkpoint test"
+        vars:
+          testParams: "-t m -n 1000000 -k 5000000 -C cache_size=100MB"
```
quchenhao: You don't need the double quotes. Better to be consistent to other places.

test/evergreen.yml
```diff
@@ -115,6 +115,16 @@ functions:
...
         rm -rf "wiredtiger.tgz"
 
+  "checkpoint test":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/build_posix/test/checkpoint"
+      script: |
+        set -o errexit
+        set -o verbose
+        ./t ${testParams} 2>&1
```
quchenhao: testParams is too general. I'd prefer something like checkpoint_args

test/evergreen.yml
```diff
@@ -115,6 +115,15 @@ functions:
...
         rm -rf "wiredtiger.tgz"
 
+  "checkpoint test":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/build_posix/test/checkpoint"
+      script: |
+        set -o errexit
+        set -o verbose
+        ./t ${checkpointArgs} 2>&1
```
quchenhao: Can you change it to checkpoint_args just to be consistent.

PR #4957, opened 2019-11-04
WT-5189 Python2: handle errors that occur under the concurrent test option (-j)

ddanderson:


ddanderson:
@sueloverso, could you please review?

ddanderson:
I put some comments in the ticket about how I test this.

PR #4958, opened 2019-11-04
WT-4983 Migrating Jenkins wiredtiger job to evergreen

hbokhari:


hbokhari:
URL for evergreen patch test request for the "wiredtiger" test: https://evergreen.mongodb.com/version/5dc0bf141e2d176cdfca8463##0

test/evergreen.yml
```diff
@@ -1384,6 +1384,37 @@ tasks:
...
+            ../configure --enable-strict --enable-silent-rules --disable-shared
+            make
+            make clean
+
+            # We're going to run tests, include debugging
+            ../configure CFLAGS="-g -Werror" --enable-silent-rules --enable-diagnostic --disable-static
+            make all
+            [ -e /usr/local/lib/libeatmydata.so ] && export LD_PRELOAD=/usr/local/lib/libeatmydata.so
+            make check VERBOSE=1
+            make clean
```
lukech: We have 2 existing functions (`compile wiredtiger` and `make check all`) that could help simplying this section of code here.

test/evergreen.yml
```diff
@@ -1384,6 +1384,37 @@ tasks:
               ./t -1 -C "direct_io=[data]" -c ../../../test/format/CONFIG.stress
             done
 
+  - name: wiredtiger
```
lukech: The legacy name `wiredtiger` does not make much sense when migrating to Evergreen. I'd prefer a task name indicating the test purpose/area. In this case, the job runs `make check` to test a few different configurations.

test/evergreen.yml
quchenhao: Also try to utilise multi-thread to speed up things. Add ${smp_command} at each make statement

quchenhao:
Agreed, we can combine them if that's possible.

test/evergreen.yml
hbokhari: Let me have a look to see how we can reuse these two function with our parameters.

quchenhao: The convention is to have a build task and then have a test task depend on that build task, though the test task has to build the whole thing again, which is sad.

quchenhao: Leave the libeatmydata part for now, we have a separate ticket to handle that and you should not run make clean

hbokhari:
Modified the shell script for existing `config-combinations` Evergreen task instead of adding new task.

test/evergreen/configure_combinations.sh
```diff
@@ -64,8 +66,23 @@ for flag in $flags ; do
...
-               echo hello
        done
 done
 IFS=$saved_IFS
-exit $ecode
+if [ $ecode -eq 1 ]; then
+        exit 1
+fi
+
+# Handle special build combination for running all the diag tests.
```
quchenhao: diag means diagnostic?

test/evergreen/configure_combinations.sh
```diff
@@ -64,8 +66,23 @@ for flag in $flags ; do
...
+flag=`CFLAGS=\"-g -Werror\"`
```
quchenhao: Can't you merge this into build test?

test/evergreen/configure_combinations.sh
```diff
@@ -64,8 +66,23 @@ for flag in $flags ; do
...
+option=`--enable-silent-rules --enable-diagnostic --disable-static`
+echo "Building with: $flag $option"
+cd "$curdir"
+rm -rf ./build
+mkdir build
+cd ./build
+../configure $flag $option || { echo "*** ERROR: $flag, $option " ; exit 1; }
+make all "$@" || { echo "*** ERROR: $flag, $option " ; exit 1; }
+make check VERBOSE=1 || { echo "*** ERROR: $flag, $option " ; exit 1; }
+make clean
```
quchenhao: No need to clean here.

test/evergreen/configure_combinations.sh
```diff
@@ -7,15 +7,17 @@ sh build_posix/reconf
 curdir=`pwd`
 
 flags="CFLAGS=\"-Werror -Wall -Wextra -Waddress -Waggregate-return -Wbad-function-cast -Wcast-align -Wdeclaration-after-statement -Wformat-security -Wformat-nonliteral -Wformat=2 -Wmissing-declarations -Wmissing-field-initializers -Wmissing-prototypes -Wnested-externs -Wno-unused-parameter -Wpointer-arith -Wredundant-decls -Wshadow -Wundef -Wunused -Wwrite-strings -O -fno-strict-aliasing -Wuninitialized\"
-CC=clang CFLAGS=\"-Wall -Werror -Qunused-arguments -Wno-self-assign -Wno-parentheses-equality -Wno-array-bounds\""
+CC=clang CFLAGS=\"-Wall -Werror -Qunused-arguments -Wno-self-assign -Wno-parentheses-equality -Wno-array-bounds\"
+CFLAGS=\" \""
```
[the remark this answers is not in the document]
hbokhari: One of the combinations doesn't use CFLAGS options.

test/evergreen/configure_combinations.sh
hbokhari: I discussed it with @lukech. The main difference is `make` vs `make all`, and, `make -C examples/c check` vs `make check`. The special case test is more comprehensive compared to previous ones.

test/evergreen.yml
```diff
@@ -1550,3 +1550,4 @@ buildvariants:
   - name: generate-datafile-big-endian
   - name: verify-datafile-big-endian
   - name: verify-datafile-from-little-endian
+
```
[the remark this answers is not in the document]
quchenhao: oh, I see what's this. You should keep the change. Sorry.

test/evergreen/configure_combinations.sh
hbokhari: Yes

quchenhao:
Can you also resolve the conflicts? I think it's actually OK to reduce the test coverage of that test. Anyway make check has been covered in other tests as well.

test/evergreen/configure_combinations.sh
quchenhao: Just use diagnostic I think.

hbokhari:
In a short discussion I had a with @bvpvamsikrishna he told me that there is seperate task written up to optimise the evergreen jobs we are adding. We can probably do a pruning at that stage to see where are the overlaps and what sort of test coverage we are missing.

test/evergreen/configure_combinations.sh
lukech: What's the purpose of this line?

test/evergreen/configure_combinations.sh
```diff
@@ -7,15 +7,17 @@ sh build_posix/reconf
 curdir=`pwd`
 
 flags="CFLAGS=\"-Werror -Wall -Wextra -Waddress -Waggregate-return -Wbad-function-cast -Wcast-align -Wdeclaration-after-statement -Wformat-security -Wformat-nonliteral -Wformat=2 -Wmissing-declarations -Wmissing-field-initializers -Wmissing-prototypes -Wnested-externs -Wno-unused-parameter -Wpointer-arith -Wredundant-decls -Wshadow -Wundef -Wunused -Wwrite-strings -O -fno-strict-aliasing -Wuninitialized\"
-CC=clang CFLAGS=\"-Wall -Werror -Qunused-arguments -Wno-self-assign -Wno-parentheses-equality -Wno-array-bounds\""
+CC=clang CFLAGS=\"-Wall -Werror -Qunused-arguments -Wno-self-assign -Wno-parentheses-equality -Wno-array-bounds\"
```
lukech: It looks there's no coverage for `-Wno-unused` CFLAGS setting.

test/evergreen/configure_combinations.sh
```diff
@@ -64,8 +66,22 @@ for flag in $flags ; do
...
+option=`--enable-silent-rules --enable-diagnostic --disable-static`
+echo "Building with: $flag $option"
+cd "$curdir"
+rm -rf ./build
+mkdir build
+cd ./build
+../configure $flag $option || { echo "*** ERROR: $flag, $option " ; exit 1; }
+make all "$@" || { echo "*** ERROR: $flag, $option " ; exit 1; }
+make check VERBOSE=1 || { echo "*** ERROR: $flag, $option " ; exit 1; }
+exit 0
```
lukech: As this part is special in running full make check, and there's functions we can reuse to simplify the code, I'd prefer to move this part outside - into 2 separate tasks - say `compile-warnings-to-errors` and `make-check-warnings-to-errors`.

test/evergreen/configure_combinations.sh
hbokhari: This is to test a scenario where no explicit value is set for `CFLAGS`.

lukech: In that case, I'd rather leave without this line as we've got some coverage for that (e.g. `ubuntu1804` build variant). Also, adding one line here actually means adding tests for a list of configure options (listed below), which could extend the run time of this test by 1/2.

test/evergreen/configure_combinations.sh
hbokhari: I was unable to find reference to `-Wno-unused` flag in the GCC manual: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options. That's why I skipped this flag. Do you know history for this flag?

test/evergreen.yml
```diff
@@ -1435,6 +1435,15 @@ tasks:
...
             ./configure_combinations.sh ${smp_command|} 2>&1
+      # Handle special build combination for running all the diagnostic tests.
+      - func: "configure wiredtiger"
+        vars:
+          configure_env_vars: CFLAGS="-g -Werror"
+          posix_configure_flags: --enable-silent-rules --enable-diagnostic --disable-static
+      - func: "make wiredtiger"
+        vars:
+          make_command: make all
+      - func: "make check all"
```
lukech: Splitting the `make check all` part out is one solution - if the previous `configure_combinations.sh` test fails, there's risk the build_posix directory is to be overwritten by the later funcs added here, which means we lose environment to troubleshoot the test failure. That's actually why I prefer separate compile and make check tasks instead of separate funcs within the same task. 

The task is an unit in Evergreen that has its own artifacts archived.

test/evergreen/configure_combinations.sh
```diff
@@ -64,8 +65,8 @@ for flag in $flags ; do
                        echo "*** ERROR: $flag, $option"
                fi
                IFS="$cr_IFS"
-               echo hello
        done
 done
 IFS=$saved_IFS
 exit $ecode
+
```
lukech: An extra line that could be removed?

hbokhari: Yes I'll do it.

test/evergreen.yml
hbokhari: I just ran a mock test where the `configure_combination.sh` script returned 1. The task was aborted immediately and no further commands were executed from the task. Given this built-in safety mechanism, I believe it is safe to just add functions sequentially with a guarantee that no command/function will be executed if any of the previous command/function fails.

PR #4959, opened 2019-11-05
WT-5025 Migrate wiredtiger-test-coverage

quchenhao:


test/evergreen.yml
```diff
@@ -1521,6 +1584,7 @@ buildvariants:
     - name: make-check-asan-test
     - name: configure-combinations
     - name: checkpoint-filetypes-test
+    - name: coverage-test
```
lukech: Prefer to retain the original name `test-coverage`

test/evergreen.yml
```diff
@@ -1619,8 +1683,8 @@ buildvariants:
   run_on:
   - macos-1012
   expansions:
-    smp_command: -j $(sysctl -n hw.logicalcpu)
     configure_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb"
+    smp_command: -j $(sysctl -n hw.logicalcpu)
```
lukech: Any special reason to move this line?

test/evergreen.yml
```diff
@@ -187,23 +187,23 @@ tasks:
       - func: "get project"
       - func: "compile wiredtiger"
         vars:
-          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ CFLAGS="-fsanitize=address -fno-omit-frame-pointer -ggdb"
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-fsanitize=address -fno-omit-frame-pointer -ggdb"
```
lukech: I guess this change is to enable other toolchain binaries (e.g gcov) to be applied when executing the tests?

test/evergreen.yml
```diff
@@ -1501,6 +1501,69 @@ tasks:
...
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="--coverage -fPIC -ggdb" LDFLAGS=--coverage
+          posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-strict --enable-python --with-builtins=lz4,snappy,zlib
+      - func: "make check all"
+      - func: "unit test"
+        vars:
+          unit_test_args: -v 2 --long
+      - func: "test format"
+        vars:
+          test_format_args: -1 -q -c ../../../test/format/CONFIG.stress checkpoints=1 leak_memory=0 mmap=1 file_type=row compression=snappy logging=1 logging_compression=snappy logging_prealloc=1
```
lukech: There seem to be lots of duplicated format arg settings, and `-1 -q -c ../../../test/format/CONFIG.stress checkpoints=1 leak_memory=0 mmap=1` is pretty standard in this test. Can we split the `test_format_args` into multiple sub args and give them a reasonable default value, so that we can save some settings here?

test/evergreen.yml
```diff
@@ -1501,6 +1501,69 @@ tasks:
...
+      - func: "test format"
+        vars:
+          test_format_argas: -1 -q -c ../../../test/format/CONFIG.stress checkpoints=1 leak_memory=0 mmap=1 file_type=fix
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+            pip install gcovr --user
```
lukech: Add a `# FIX ME` comment mentioning this line should be removed once the BUILD ticket is resolved.

test/evergreen.yml
quchenhao: Well, I'm not sure about this. I want to minimize the number of arguments we introduce. I don't want to go into a path that we can potentially create argument for each command line options.

test/evergreen.yml
quchenhao: It can be confusing with the convention that we put test at the end. What about coverage-report

test/evergreen.yml
quchenhao: Just to make it consistent with other buildvariants.

test/evergreen.yml
quchenhao: Theoretically we should enforce all the tests using the consistent binaries to run even though they are not running the tests that are affected by mismatched binaries.

test/evergreen.yml
lukech: Basically, this task is to run some tests to generate a test coverage report. `coverage-report` sounds better than `coverage-test` or `test-coverage-test`.

test/evergreen.yml
lukech: Ah okay.

test/evergreen.yml
lukech: Sounds good. Hopefully, there won't be breakage to other tests (or we are ready to fix if any).

test/evergreen.yml
lukech: Probably too expensive for each option, but might be worthwhile for some frequently used ones e.g. `-c <cfg_file>`, `-1 (times to run)`, `file_type` etc. 
I'm not having a strong opinion though, just wondering what we can do to reduce duplication.

PR #4960, opened 2019-11-06
Wt 4990

raviprakashgiri29:
Added a new task "test-unit-ppc" to set the environment variable as needed and using the existing function "unit test"

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -231,7 +231,7 @@ tasks:
         vars:
           posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-strict --enable-python
 
-  # Compile for pcc unit test by enabling strict, python, diagnostic and snappy flags
+  # Compile pcc by enabling strict, python, diagnostic and snappy flags
   - name: compile-ppc
     commands:
       - func: "get project"
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -1508,13 +1508,9 @@ tasks:
         vars:
           configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ CPPFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib" CFLAGS="-g -fPIC"
           posix_configure_flags: --enable-strict --enable-python --enable-diagnostic --enable-snappy
-      - command: shell.exec
-        params:
-          working_dir: "wiredtiger/build_posix"
-          script: |
-            set -o errexit
-            set -o verbose
-            env LD_LIBRARY_PATH=.libs python ../test/suite/run.py -v 2 $TESTFLAGS 2>&1
+      - func: "unit test"
+        vars:
+          test_env_vars: env LD_LIBRARY_PATH=.libs
 
 buildvariants:
 - name: ubuntu1804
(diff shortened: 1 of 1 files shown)


wtbuild:

raviprakashgiri29:
Evergreen patch link
https://evergreen.mongodb.com/version/5dc34ce4d6d80a5bf151689a

test/evergreen.yml
```diff
@@ -1501,6 +1501,17 @@ tasks:
...
+  - name: test-unit-ppc
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ CPPFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib" CFLAGS="-g -fPIC"
+          posix_configure_flags: --enable-strict --enable-python --enable-diagnostic --enable-snappy
+      - func: "unit test"
+        vars:
+          test_env_vars: env LD_LIBRARY_PATH=.libs
```
quchenhao: No need to set this. The value is set at buildvariant level.

test/evergreen.yml
```diff
@@ -1501,6 +1501,17 @@ tasks:
         vars:
           checkpoint_args: -t c -n 1000000 -k 5000000 -C cache_size=100MB
 
+  - name: test-unit-ppc
+    commands:
+      - func: "get project"
```
quchenhao: The current convention is we have a compile task and the test tasks will depend on that compile task. As you can see, it is clunky as we still need to rebuild in the test tasks because of something silly in the autotools. There is a ticket to work on removing the rebuild in the future. But here I think we need to adhere to the convention to decouple the build task and the test.

test/evergreen.yml
```diff
@@ -1501,6 +1501,17 @@ tasks:
...
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ CPPFLAGS="-I/usr/local/include" LDFLAGS="-L/usr/local/lib" CFLAGS="-g -fPIC"
```
quchenhao: Better to change -g to -ggdb

quchenhao: Why we need this CPPFLAGS="-I/usr/local/include"?

test/evergreen.yml
raviprakashgiri29: Created a new compile task "compile-pcc" which depends on "test-unit-ppc".

test/evergreen.yml

agorrod:
ok to test

test/evergreen.yml
```diff
@@ -231,6 +231,16 @@ tasks:
         vars:
           posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-strict --enable-python
 
+  # Compile pcc by enabling strict, python, diagnostic and snappy flags
+  - name: compile-ppc
+    tags: ["pull_request"]
```
quchenhao: It's not a pull request

test/evergreen.yml
quchenhao: here we should use fetch artifacts

lukech:
I don't know why this PR got merged with a failed Jenkins PR test.

lukech:
And I have not got a chance reviewing the change.

raviprakashgiri29:
retest this please

raviprakashgiri29:
evergreen retry

PR #4961, opened 2019-11-07
WT-5174 truncation rollback locking is insufficient

sueloverso:


sueloverso:
@keithbostic hold up on this review. It needs more. I just checked my overnight run and it hit a problem accessing freed memory. So the assertion is going to need to be a coordination point instead.

keithbostic:
autobuild: retest this please

sueloverso:
With the current changes, I cannot reproduce the assertion failure. I am now consistently hitting WT-5169 instead. I'll either also add in the asserts regarding the usage of ref->page_del or merge this and open a new ticket for those assertions.

PR #4962, opened 2019-11-08
WT-5136 Fix reading freed memory due to birthmark after uncommitted updates freed 

quchenhao:
Obsolete birthmark cannot be freed if it follows a set of uncommitted
updates as the uncommitted updates can be aborted and freed causing the
entire update chain being removed in this case.

Birthmark can be discarded only if there is an update that is globally visible before it.

### DIFF
--- a/src/btree/row_modify.c
+++ b/src/btree/row_modify.c
@@ -309,11 +309,11 @@ __wt_update_obsolete_check(
      * Only updates with globally visible, self-contained data can terminate update chains.
      *
      * Birthmarks are a special case: once a birthmark becomes obsolete, it can be discarded if
-     * there is an update visible to all sessions before it and subsequent reads will see the
-     * on-page value (as expected). Inserting updates into the lookaside table relies on this
-     * behavior to avoid creating update chains with multiple birthmarks. We cannot discard the
-     * birthmark if it's the first globally visible update as the previous updates can be aborted
-     * and be freed as well causing the entire update chain being removed.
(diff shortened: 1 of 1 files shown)


quchenhao:
@agorrod Can you look at this change?

src/btree/row_modify.c
```diff
@@ -324,10 +329,14 @@ __wt_update_obsolete_check(
              */
             if (upd->start_ts != WT_TS_NONE && upd->start_ts >= oldest && upd->start_ts < stable)
                 ++upd_unstable;
-        } else if (first == NULL && upd->type == WT_UPDATE_BIRTHMARK)
-            first = prev;
-        else if (first == NULL && WT_UPDATE_DATA_VALUE(upd))
-            first = upd;
+        } else {
+            if (first == NULL && upd->type == WT_UPDATE_BIRTHMARK)
```
agorrod: The fix looks correct to me. It deserves a comment now, the logic in here is complex, so we should add a comment to describe why.

I'm also not a huge fan of ternary (?) statements, especially when they are nested within other conditional blocks. Could this be restructured? Maybe:

```
if (first == NULL) {
    if (upd_visible_all_seen && upd->type == WT_UPDATE_BIRTHMARK)
        first = prev;
    else if (upd->type == WT_UPDATE_BIRTHMARK || WT_UPDATE_DATA_VALUE(upd))
        first = upd;
}
```

It adds another level of nesting, but I think the flow is easier to follow.

PR #4964, opened 2019-11-08
WT-4990 Migrate Jenkins “wiredtiger-test-unit-ppc” job to Evergreen

raviprakashgiri29:
This is a new pull request after restore/revert.

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -231,7 +231,7 @@ tasks:
         vars:
           posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-strict --enable-python
 
-  # Compile for pcc unit test by enabling strict, python, diagnostic and snappy flags
+  # Compile pcc by enabling strict, python, diagnostic and snappy flags
   - name: compile-ppc
     commands:
       - func: "get project"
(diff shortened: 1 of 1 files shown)


test/evergreen.yml
```diff
@@ -231,6 +231,15 @@ tasks:
...
 
+  # Compile pcc by enabling strict, python, diagnostic and snappy flags
+  - name: compile-ppc
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb -fPIC"
+          posix_configure_flags: --enable-strict --enable-python --enable-diagnostic --enable-snappy
+
```
lukech: Instead of doing an apple-to-apple migration for this job, I'd prefer to understand the importance of the configure differences compared to existing `unit-test` task, see if we can come out a standard configure for unit tests (Python test suite). 
@hbokhari is doing similar investigation for his `wiredtiger-test-unit` job.

test/evergreen.yml
```diff
@@ -1521,6 +1543,8 @@ buildvariants:
     - name: make-check-asan-test
     - name: configure-combinations
     - name: checkpoint-filetypes-test
+    - name: compile-ppc
+    - name: test-unit-ppc
```
lukech: These new tasks should be run on `ubuntu1604-power8-build` or `ubuntu1604-power8-test` distros.

PR #4965, opened 2019-11-08
WT-5217 Write utility to generate operation tracking log from X-Ray trace

tetsuo-cpp:
Added an `xray_to_optrack` utility program to convert from XRay logs to the OpTrack format. I'm currently building with `../configure --enable-llvm --with-llvm-config=llvm-config-8`.

One issue that I'm seeing is that `xray_to_optrack` generates very big OpTrack logs. The `find-latency-spikes.py` seemed to handle it ok despite taking a while but `optrack_to_t2.py` never finished for me and I waited quite a while. This makes sense to me since I think OpTrack logs have historically been used to instrument a few functions of interest whereas by default, XRay will instrument anything that compiles down to more than 200 instructions. To make this more manageable, we might have to manually choose a set of functions to instrument when building with XRay. We can do it either with function attributes or with a special case file.

Screen Shot 2019-11-09 at 2 00 52 am

### DIFF
--- a/build_posix/aclocal/options.m4
+++ b/build_posix/aclocal/options.m4
@@ -275,4 +275,21 @@ no)	wt_cv_crc32_hardware=no
 	AC_MSG_RESULT(no);;
 esac
 
+AC_MSG_CHECKING(if --enable-llvm option specified)
+AC_ARG_ENABLE(llvm,
+	[AS_HELP_STRING([--enable-llvm],
+	    [Configure with LLVM.])], r=$enableval, r=no)
+case "$r" in
+no)	wt_cv_enable_llvm=no;;
+*)	wt_cv_enable_llvm=yes;;
+esac
+AC_MSG_RESULT($wt_cv_enable_llvm)
+AM_CONDITIONAL([LLVM], [test x$wt_cv_enable_llvm = xyes])
+
+AC_MSG_CHECKING(if --with-llvm-config option specified)
+AC_ARG_WITH(llvm-config,
+	[AS_HELP_STRING([--with-llvm-config=DIR],
+	    [Path to LLVM config.])])
+AC_MSG_RESULT($with_llvm_config)
+
 ])
--- a/build_posix/aclocal/options.m4
+++ b/build_posix/aclocal/options.m4
@@ -284,12 +284,14 @@ no)	wt_cv_enable_llvm=no;;
 *)	wt_cv_enable_llvm=yes;;
 esac
 AC_MSG_RESULT($wt_cv_enable_llvm)
+if test "$wt_cv_enable_llvm" = "yes"; then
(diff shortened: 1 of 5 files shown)


agorrod:
The idea looks good to me. I've subbed @keithbostic in for me as a reviewer here. He has a better eye for autoconf changes than me.

build_posix/aclocal/options.m4
```diff
@@ -275,4 +275,21 @@ no)	wt_cv_crc32_hardware=no
...
+esac
+AC_MSG_RESULT($wt_cv_enable_llvm)
+AM_CONDITIONAL([LLVM], [test x$wt_cv_enable_llvm = xyes])
+
+AC_MSG_CHECKING(if --with-llvm-config option specified)
+AC_ARG_WITH(llvm-config,
+	[AS_HELP_STRING([--with-llvm-config=DIR],
+	    [Path to LLVM config.])])
+AC_MSG_RESULT($with_llvm_config)
+
```
keithbostic: Can you discuss why you went in this direction, rather than `AC_CHECK_PROG` and/or `AC_CHECK_HEADER`?

It seems like we could build if we find `llvm-config` in `$PATH` and/or some of the required header files?

tools/xray_to_optrack/xray_to_optrack.cxx
```diff
@@ -0,0 +1,153 @@
+#include <llvm/ADT/DenseMap.h>
```
keithbostic: Needs a copyright notice if it's in the standard distribution.

keithbostic:
@tetsuo-cpp, I'm the wrong guy to review C++ code -- if you want a code review, I'd suggest adding someone else to the review team.

build_posix/aclocal/options.m4
keithbostic: If that's the case, then you probably will need to version the test, that is, only build if `llvm-config --version` reports a sufficiently up-to-date version?

keithbostic: As a simple approach, maybe it would be best to require `--enable-XXX` to do the build, but then use `$PATH` to find `llvm-config`? That way we don't build unless the user configures us to do so, but we use the compiler/binary/include-file configuration the user has set before running for the actual build.

tools/xray_to_optrack/xray_to_optrack.cxx
tetsuo-cpp: f88f593

build_posix/aclocal/options.m4
tetsuo-cpp: Understood. As discussed, I'll do it this way.

tetsuo-cpp: 491971a

tetsuo-cpp:
@fedorova Yep! I've created WT-5259 to track that work.
@keithbostic Could you please re-review the autotools changes?

PR #4966, opened 2019-11-08
WT-5142 Don't create huge root pages when rebalancing or bulk-loading objects with overflow keys

keithbostic:
@agorrod, I ended up drilling the hole that allows callers of the reconciliation split code to force a split regardless of the current disk image size. I found a path I don't think is too ugly, and so avoid the retry logic we discussed (and which neither of us particularly liked). I like this approach better because I think it's reasonable for future work to want to force a split in a similar way, and this moves reconciliation in a more generally useful direction than the retry implementation would.

@sulabhM, these changes are primarily in the split code you & I wrote awhile back. The idea is that a caller can force the underlying code to split regardless of the current size of the disk image.

One relatively subtle change is [at line 1255 in rec_write.c](https://github.com/wiredtiger/wiredtiger/blob/develop/src/reconcile/rec_write.c#L1255):
I removed the `!__wt_rec_need_split(r, 0)` clause because I couldn't think of any reason it could be needed. Please review that change carefully.

EDIT:
I have restored the `!__wt_rec_need_split(r, 0)` clause, removing it fundamentally changed eviction characteristics, which isn't the point of this ticket.

### DIFF
--- a/src/reconcile/rec_row.c
+++ b/src/reconcile/rec_row.c
@@ -478,6 +478,8 @@ __wt_rec_row_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
         r->cell_zero = false;
 
         /*
+         * Boundary: split or write the page.
+         *
          * We always instantiate row-store internal page keys in order to avoid special casing the
          * B+tree search code to handle keys that may not exist (and I/O in a search path). Because
          * root pages are pinned, overflow keys on the root page can cause follow-on cache effects
@@ -487,13 +489,10 @@ __wt_rec_row_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
          * which will impact search performance, but at some point, the application's configuration
          * is too stupid to survive.
          */
-        page_image += key->len + val->len;
         force = r->entries > 20 && page_image > btree->maxmempage_image;
-        if (force)
+        if (force || __wt_rec_need_split(r, key->len + val->len)) {
             page_image = 0;
 
-        /* Boundary: split or write the page. */
-        if (force || __wt_rec_need_split(r, key->len + val->len)) {
             /*
              * In one path above, we copied address blocks from the page rather than building the
              * actual key. In that case, we have to build the key now because we are about to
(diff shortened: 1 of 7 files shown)


src/reconcile/rec_row.c
```diff
@@ -467,25 +466,42 @@ __wt_rec_row_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
             key->cell_len = 0;
             key->len = key->buf.size;
             ovfl_key = true;
+            page_image += ikey->size;
```
sulabhM: `page_size` only account for the key size, that too only when overflow key is involved. Is that intentional. Maybe we can rename the variable to reflect that

src/reconcile/rec_row.c
```diff
@@ -467,25 +466,42 @@ __wt_rec_row_int(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_PAGE *page)
...
+         * We always instantiate row-store internal page keys in order to avoid special casing the
+         * B+tree search code to handle keys that may not exist (and I/O in a search path). Because
+         * root pages are pinned, overflow keys on the root page can cause follow-on cache effects
+         * where huge root pages tie down lots of cache space and eviction frantically attempts to
+         * evict objects that can't be evicted. If the in-memory image is too large, force a split.
+         * Potentially, limiting ourselves to 10 items per page is going to result in deep trees
+         * which will impact search performance, but at some point, the application's configuration
+         * is too stupid to survive.
+         */
+        force = r->entries > 20 && page_image > btree->maxmempage_image;
```
sulabhM: Is 10 per page a magic figure here? I do not have the knowledge to review how good or bad "10" would behave in this case.

src/reconcile/rec_row.c
keithbostic: Renamed to `key_overflow_size` in 08dab949c.

src/reconcile/rec_row.c
keithbostic: Me either. Basically, at this point we've created an internal page that will result in a too-large page in memory, so it's a trade-off between too few keys in the internal page and too large a page in memory.  I'm not sure what we can do other than bound the page to some combination of the above.  I don't think there are any good answers.

PR #4969, opened 2019-11-12
WT-4988 Migrate Jenkins unit-test job to Evergreen

hbokhari:


test/evergreen.yml
```diff
@@ -1583,10 +1592,11 @@ buildvariants:
   run_on:
   - ubuntu1804-test
   expansions:
-    test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
+    test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd) CFLAGS="-fPIC -ggdb"
     smp_command: -j $(grep -c ^processor /proc/cpuinfo)
     posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-python --enable-zlib --enable-snappy --enable-strict --enable-static --prefix=$(pwd)/LOCAL_INSTALL
     make_command: PATH=/opt/mongodbtoolchain/v3/bin:$PATH make
+    unit_test_args: -v 2
```
lukech: Is this line necessary if the setting here is the same as the default value?

raviprakashgiri29: You don't need to set the variable(unit_test_args) because by default the value will be (-v 2) in all the "unit test" functions if not set.

hbokhari: Added this to make testing more robust and make it future proof. There are few other runtime arguments from unit test python script that we are not using. For example, we can make `long` unit test as default option very easily.

raviprakashgiri29: Ignore this as it is already pointed out by @lukech .

test/evergreen.yml
```diff
@@ -1604,12 +1614,13 @@ buildvariants:
   run_on:
   - ubuntu1804-test
   expansions:
-    test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
+    test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd) CFLAGS="-fPIC -ggdb"
```
lukech: Should the addition of CFLAGS setting be put as part of `posix_configure_flags`? 
Same comment for other POSIX build variants below.

test/evergreen.yml
```diff
@@ -1583,10 +1592,11 @@ buildvariants:
...
```
lukech: Should the addition of CFLAGS setting be put as part of posix_configure_flags?
Same comment for other POSIX build variants below.

hbokhari: Just following the convention from other tasks in the file. I believe all existing tasks/function add CFLAGS under `configure_env_vars`.

hbokhari: Fixed the flag placement. Need to wait for RHEL and OSX job to confirm changes are working as expected.

test/evergreen.yml
lukech: It sounds like this is the preparation work for the `long` unit test, which is fine. I tend to do such restructure work in a more lazy way (at the time there's a real need), but all good if you choose to keep it.

PR #4970, opened 2019-11-12
WT-5126 Incorrect assertion regarding prepared update resolution

keithbostic:
@luke-pearson, I think I finally understand this one. I don't believe there's a problem in the code, I believe the asserts themselves are faulty. The commit comment in 39aef4d explains the problem.

I removed a statistic we were maintaining because I think it's redundant. That change is on a separate commit, 11e4cc9, feel free to discard the commit if you disagree.

There's lots of code shuffling and cleanup that I've kept, I think it's better than before. If you disagree with those changes, I can fall back to a simpler, more focused change.

One thing that's interesting about this is the requirement that we be able to "resolve" prepared updates more than once, that is, we can repeatedly resolve the same update during commit/rollback. I looked over the relevant code, and I think that's safe, but I think it's worth your careful review.

cc: @bvpvamsikrishna


luke-pearson:
It may be worth moving some of the detail in the commits to the jira ticket, up to you though.

keithbostic:
> It may be worth moving some of the detail in the commits to the jira ticket, up to you though.

Sure, done.

PR #4972, opened 2019-11-13
WT-4989 Migrating Jenkins unit-test-long job to Evergreen

hbokhari:


test/evergreen.yml
```diff
@@ -1027,6 +1027,16 @@ tasks:
       - func: "fetch artifacts"
       - func: "unit test"
 
+  - name: unit-test-long
+    tags: ["unit_test"]
```
lukech: Be cautious about setting this tag :) We only want this task to be executed under `ubuntu1804` build variant, for now.

test/evergreen.yml
```diff
@@ -1027,6 +1027,16 @@ tasks:
...
+    depends_on:
+    - name: compile
+    commands:
+      - func: "fetch artifacts"
+      - func: "unit test"
+        vars:
+          unit_test_args: ${unit_test_args} --long
```
lukech: Feel a bit over-engineered. `-v 2 --long` is fine to me.

test/evergreen.yml
hbokhari: Ah yes I get your point. Will make the change.

test/evergreen.yml
hbokhari: Noted.

hbokhari:
Here is the link for patch build `with uni-test-long' : https://evergreen.mongodb.com/version/5dcca118d1fe070d30cc182e

PR #4973, opened 2019-11-13
WT-5228 Fix conditions when on-disk record is initialised as an update

sulabhM:


keithbostic:
@sulabhM, 9e906f7 fixes `test_durable_ts03`, `test_las02` (and maybe `test_timestamp04`, I'm not sure about that one).

That said, I'm not convinced it's reasonable for there not to be any LAS records in this case, with your changes, can you please take a look? Feel free to revert without discussion if I'm going in the wrong direction.

src/btree/bt_read.c
```diff
@@ -255,89 +249,65 @@ __instantiate_lookaside(WT_SESSION_IMPL *session, WT_REF *ref)
...
-        if (las_btree_id != S2BT(session)->id)
-            break;
-
-        WT_ERR(__wt_compare(session, NULL, &las_key, &page_las->min_las_key, &cmp));
-        /* We got a key that is less than the minimum key of the page. */
-        if (cmp < 0)
-            break;
+    for (i = 0; i < page_las->birthmarks_cnt; ++i) {
+        if (page_las->birthmarks[i].txnid != WT_TXN_ABORTED) {
+            WT_ERR(__create_birthmark_upd(session, &page_las->birthmarks[i], &size, &upd));
```
sulabhM: please add a comment saying this is a birthmark

src/btree/bt_read.c
```diff
@@ -255,89 +249,65 @@ __instantiate_lookaside(WT_SESSION_IMPL *session, WT_REF *ref)
...
-                    /*
-                     * LAS key is greater than birthmark key, check next set of birthmark records.
-                     */
-                    continue;
-            }
+            /*
+             * Verify that our search_near has given us the right key.
+             *
+             * If it hasn't, this means that there are no longer any lookaside entries for that key.
+             * This is possible if the sweep server cleans some things up while the page was out of
```
sulabhM: sweep can't go past this record since this has not been checkpointed yet. We should assert/panic if we cant find this key

src/btree/bt_read.c
```diff
@@ -431,11 +401,8 @@ __instantiate_lookaside(WT_SESSION_IMPL *session, WT_REF *ref)
             las_prepare_cnt++;
         }
 
-        if (birthmark_record) {
-            birthmarkp->instantiated = true;
-            birthmark_record = false;
-        }
-
+        page_las->birthmarks[i].instantiated = true;
```
sulabhM: check if we need `instantiated` now. Also we do not need `__instantiate_birthmarks()` further down at line 414

tetsuo-cpp: 535a348

src/btree/bt_read.c
tetsuo-cpp: 535a348

src/btree/bt_read.c
tetsuo-cpp: 535a348

src/reconcile/rec_row.c
```diff
@@ -621,9 +621,6 @@ __rec_row_leaf_insert(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins)
             ret = __wt_illegal_value(session, upd->type);
             WT_ERR(ret);
         }
-        /* Free the update if it is external. */
```
tetsuo-cpp: @sulabhM We talked about this earlier but I don't think this is quite right. This is within a loop and I think only the last one should be kept around.

src/include/btmem.h
```diff
@@ -248,17 +248,14 @@ struct __wt_page_lookaside {
...
-    WT_ITEM max_las_key;           /* The maximum key in the LAS for the page */
-    WT_ITEM min_las_key;           /* The minimum key in the LAS for the page */
     struct __wt_birthmark_details {
         WT_ITEM key;
         uint64_t txnid;
         wt_timestamp_t durable_ts;
         wt_timestamp_t start_ts;
         uint8_t prepare_state;
-        bool instantiated;
     } * birthmarks;          /* Birthmark details for a record */
```
tetsuo-cpp: This no longer corresponds to an actual birthmark since we're storing these for every key written to las now. I'd like to rename this from `WT_BIRTHMARK_DETAILS` to `WT_KEY_MEMENTO` or `WT_LAS_KEY` or something along those lines.

src/reconcile/rec_visibility.c
```diff
@@ -646,8 +677,11 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
          * is the same as it was the previous reconciliation. Otherwise lookaside can end up with
          * two birthmark records in the same update chain.
          */
-        WT_ERR(__rec_append_orig_value(session, page, first_inmem_upd, vpack));
-        upd_select->upd = NULL;
+        WT_ERR(__rec_append_orig_value(session, page, ins, ripcip, first_inmem_upd, vpack));
```
tetsuo-cpp: @sulabhM Could you please double check what I've done here? I find this section of code a bit confusing since we're using `upd` and `upd_select->upd` and it's not 100% clear to me what we're expecting `upd` to be pointing at in this block.

src/reconcile/rec_visibility.c
```diff
@@ -646,8 +677,11 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        if (upd->ext != 0)
```
tetsuo-cpp: @sulabhM Could you please double check what I've done here? I find this section of code a bit confusing since we're using both `upd` and `upd_select->upd` and it's not 100% clear to me what we're expecting `upd` to be pointing at a given point in time.

src/reconcile/rec_visibility.c
```diff
@@ -231,6 +231,8 @@ __get_next_rec_upd(WT_SESSION_IMPL *session, WT_UPDATE **inmem_upd_pos, bool *la
         if (*inmem_upd_pos != NULL)
             *inmem_upd_pos = (*inmem_upd_pos)->next;
     }
+    if (*updp != NULL && WT_TXNID_LT(*max_txn, (*updp)->txnid))
```
tetsuo-cpp: Not 100% sure about this commit so you'll have to pay attention to this one. It seems to me that this is tracking the max transaction id of a page so we know if we can safely discard a page/tree. It seems to me that it shouldn't only apply to in-memory updates like it did before.

sulabhM:
@keithbostic can you please review this branch. Durable history work requires these changes merged to them sooner.

src/include/btmem.h
keithbostic: I support that.

PR #4976, opened 2019-11-14
WT-5181 Preload eatmydata for some tests

quchenhao:
patch build https://evergreen.mongodb.com/version/5dcc8e430ae6067e415e1c07


quchenhao:
Tested eatmydata lib and it works fine. I haven't done a complete check on which tests can benefit from this. Just to ask your opinions where should we put this? I have doubts adding that to the buildvariants but don't really have other good choices.

lukech:
I think the list of build variants currently being put in is a good starting point. 

There are 2 failure tasks in the patch build pasted above. Are they related to the eatmydata library loading change?

quchenhao:
Not sure. restarting the tests. I think the failure of direct io is related to this.

test/evergreen.yml
```diff
@@ -1667,8 +1667,7 @@ buildvariants:
   run_on:
   - rhel80-test
   expansions:
-    test_env_vars: PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
-    posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-python --enable-zlib --enable-strict --enable-static --prefix=$(pwd)/LOCAL_INSTALL CFLAGS="-fPIC -ggdb"
+    test_env_vars: LD_PRELOAD=/usr/local/lib/libeatmydata.so PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)
```
hbokhari: Can you please merge the existing value of `posix_configure_flags` to the new value of `test_env_vars`?

quchenhao: No, we shouldn't do that. We have discussed this and we decided to keep them separate.

hbokhari: But this change looks like if you have removed the `posix_configure_flags` all together?

quchenhao: That means it will use the default value.

quchenhao:
The tests are all passing now. I think the previous failures are not related.

PR #4977, opened 2019-11-14
WT-5247 Resizing modify operation log full values.

michaelcahill:
When modify operations resize part of a value, subsequent bytes are shuffled around and the operation is no longer idempotent.  Since WiredTiger's recovery can reapply operations that overlap with a checkpoint, it is not safe to log resizing modify operations.  Instead, log full update operations for that case.


agorrod:
@tetsuo-cpp and @keithbostic please review this change. It's intent is to fix an issue earlier versions of MongoDB run into, where they do modify operations on logged tables. It's not (very) relevant for current versions of MongoDB, since they don't generally do modify operations on logged tables.
It's not safe to do a modify, because the generated log message isn't idempotent at the moment (i.e: if it get's applied twice you get different results than if it's only applied once).

src/support/modify.c
```diff
@@ -38,6 +38,33 @@
...
+    size_t tmp;
+    const size_t *p;
+    int nentries;
+
+    /* Get the number of modify entries. */
+    p = modify;
+    memcpy(&tmp, p++, sizeof(size_t));
+    nentries = (int)tmp;
+
+    WT_MODIFY_FOREACH_BEGIN(mod, p, nentries, 0)
```
tetsuo-cpp: Nit: Not introduced by your change but could you please add `WT_MODIFY_FOREACH_BEGIN` to the `.clang-format` file as a loop macro? I must have missed this one.

src/support/modify.c
```diff
@@ -38,6 +38,33 @@
...
+    {
+        if (mod.size != mod.data.size)
```
tetsuo-cpp: I assume it's correct but even after looking at the struct definition, I don't what it means for the modify size to be different from its data size. A brief comment would be nice for us newbies.

src/support/modify.c
michaelcahill: Done.  Also sorted the list while I was in the area.

src/support/modify.c
michaelcahill: I added a comment: let me know if that helps.

tetsuo-cpp:
@michaelcahill Something went wrong with your last commit. I don't see the changes you described but I see the `clang-format` binary + some SWIG generated file.

michaelcahill:
Sigh, thanks @tetsuo-cpp, that was my first attempt to use Visual Code end-to-end.  I'll fix manually and push...

src/support/modify.c
tetsuo-cpp: I get it now. Thanks.

PR #4979, opened 2019-11-15
WT-5048 wiredtiger_open with a corrupted meta file should always return WT_TRY_SALVAGE

GWlodarek:


wtbuild:

src/conn/conn_api.c
```diff
@@ -1641,6 +1651,10 @@ __conn_single(WT_SESSION_IMPL *session, const char *cfg[])
         bytelock = false;
         ret = 0;
     }
+    if (!F_ISSET(conn, WT_CONN_READONLY) && ret == ENOENT) {
```
ddanderson: I think (given the `if` statement that precedes this one), that the check for `!F_ISSET(conn, WT_CONN_READONLY)` is redundant.  Secondly, this seems kind of subtle to me - how does the lock file not existing imply corruption?  Probably needs a comment?

bvpvamsikrishna:
@kommiharibabu, please review this PR. After your approval, please merge it to develop as Gregory will not have permission.

PR #4982, opened 2019-11-18
WT-4994 Migrate wiredtiger-test-spinlock job to Evergreen

hbokhari:


test/evergreen.yml
```diff
@@ -1586,6 +1586,26 @@ tasks:
...
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-python --with-spinlock=gcc --enable-strict
+      - func: "make check all"
+      - func: "test format"
+        vars:
+          times: 3
+      - func: "unit test"
+      - func: "compile wiredtiger"
```
quchenhao: Can we separate it into two tests so we can run them in parallel.

test/evergreen.yml
```diff
@@ -1586,6 +1586,26 @@ tasks:
           display_name: Coverage report
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
+  - name: spinlock-test
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
```
quchenhao: Also I think it's better to have a separate compile task so that we can run make check and python unit tests in parallel.

test/evergreen.yml
hbokhari: Good idea! I can break them into two tasks and have a common tag so we can still manage them by single call.

lukech: If breaking it out into 2 tasks I'd suggest task name `spinlock-gcc-test` and `spinlock-pthread_adaptive-test`.
And, a new tag feels a bit overkill to me :)

test/evergreen.yml
```diff
@@ -1593,6 +1593,65 @@ tasks:
           display_name: Coverage report
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
+  # creating bunch of spinlock tests so they can be scheduled in parallel
+  - name: test-spinlock-gcc-1
+    tags: ["test-spinlock"]
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
```
quchenhao: I'd prefer to have a separate compile task for this though I know we need to recompile this anyway. But I'm OK with this for now, up to @lukech

test/evergreen.yml
```diff
@@ -1593,6 +1593,65 @@ tasks:
           display_name: Coverage report
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
+  # creating bunch of spinlock tests so they can be scheduled in parallel
+  - name: test-spinlock-gcc-1
```
quchenhao: Can we name this spinlock-gcc-make-check-test? The convention is to put test in the end. Similar for the other tests.

test/evergreen.yml
hbokhari: I actually did have a seperate task but it simply looked unintuitive to compile code again after `fetch artifacts`. Also it saves the cost of uploading/downloading aritifacts between tasks (not sure if that actually helps with AWS billing :) ).

hbokhari:
@lukech I see your point here. Looking at the commit merge Evergreen tasks that are run right now, there are few long running tests that take between 1~2 hours to run. If I only break the test into two sub-tasks then the runtime should be of similar magnitude (checked this with an experimental config I wrote). 

Will fuse together the sub-tasks for gcc and pthread-adaptive.

test/evergreen.yml
```diff
@@ -1593,6 +1593,33 @@ tasks:
           display_name: Coverage report
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
+  # creating bunch of spinlock tests so they can be scheduled in parallel
```
lukech: Is this comment still valid?

test/evergreen.yml
```diff
@@ -1593,6 +1593,33 @@ tasks:
...
+  - name: spinlock-gcc-test
+    tags: ["spinlock-test"]
```
lukech: I don't see some new spinlock tests to be added in the foreseeable future, and think the use of tags is a bit over-kill.

test/evergreen.yml
hbokhari: I'll fix it

lukech:
Please wait for the PR testing to pass before merging. 
Actually, it's recommended to hold a review request until the PR testing is passed.

PR #4983, opened 2019-11-19
WT-5169 WT_REF_LIMBO pages cannot support fast (leaf-page only) searches

keithbostic:
@agorrod, @sulabhM, I'm uncertain on some of the possible `WT_REF_LIMBO` transitions, please consider if there's a better test we could use than turning off all fast searches of `WT_REF_LIMBO` pages.

I considered calling `__wt_las_page_skip()` from inside `__cursor_page_pinned()`, but that requires locking the page which might be too heavy-weight (and, as I said, I'm not completely sure that calling that function would be sufficient).

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -101,17 +101,15 @@ __cursor_page_pinned(WT_CURSOR_BTREE *cbt, bool search_operation)
         return (false);
 
     /*
-     * If we are doing an update, we need a page with history, release the page so we get it again
-     * with history if required. Eviction may be locking the page, wait until we see a "normal"
-     * state and then test against that state (eviction may have already locked the page again).
+     * We need a page with history: updates need complete update lists and a read might be based on
+     * a different timestamp than the one that brought the page into memory. Release the page and
+     * read it again with history if required. Eviction may be locking the page, wait until we see a
+     * "normal" state and then test against that state (eviction may have already locked the page
+     * again).
      */
-    if (F_ISSET(&session->txn, WT_TXN_UPDATE)) {
-        while ((current_state = cbt->ref->state) == WT_REF_LOCKED)
-            __wt_yield();
-        return (current_state == WT_REF_MEM);
-    }
-
-    return (true);
+    while ((current_state = cbt->ref->state) == WT_REF_LOCKED)
+        __wt_yield();
+    return (current_state == WT_REF_MEM);
 }
 
 /*
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -109,6 +109,7 @@ __cursor_page_pinned(WT_CURSOR_BTREE *cbt, bool search_operation)
      */
     while ((current_state = cbt->ref->state) == WT_REF_LOCKED)
         __wt_yield();
+    WT_ASSERT(session, current_state == WT_REF_LIMBO || current_state == WT_REF_MEM);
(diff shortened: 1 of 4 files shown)


src/btree/bt_cursor.c
```diff
@@ -101,17 +101,15 @@ __cursor_page_pinned(WT_CURSOR_BTREE *cbt, bool search_operation)
         return (false);
 
     /*
-     * If we are doing an update, we need a page with history, release the page so we get it again
```
agorrod: I know it's kind of a tangent, but can you add an assert that the cbt->ref->state is only in WT_REF_MEM or WT_REF_LIMBO here? It may be possible that it's a WT_REF_DELETED as well?

It took me a while to think through that the cursor has a hazard pointer, so the page should be in one of those states.

src/btree/bt_cursor.c
```diff
@@ -101,17 +101,15 @@ __cursor_page_pinned(WT_CURSOR_BTREE *cbt, bool search_operation)
...
+     * again).
      */
-    if (F_ISSET(&session->txn, WT_TXN_UPDATE)) {
-        while ((current_state = cbt->ref->state) == WT_REF_LOCKED)
-            __wt_yield();
-        return (current_state == WT_REF_MEM);
-    }
-
-    return (true);
+    while ((current_state = cbt->ref->state) == WT_REF_LOCKED)
```
agorrod: I'm mildly concerned that all cases could spin here now, but generally we are holding a hazard pointer so the page should only go into locked for long enough for us to decide that we can't get exclusive access.

i.e: I talked myself out of pushing back on this, but wanted to flag it.

src/btree/bt_cursor.c
keithbostic: > add an assert that the cbt->ref->state is only in WT_REF_MEM or WT_REF_LIMBO here? It may be possible that it's a WT_REF_DELETED as well?

That makes sense to me -- I don't think it can be `WT_REF_DELETED`, so it would be good to know if that's ever the case.

Pushed in 509bab7c8.

src/btree/bt_debug.c
```diff
@@ -697,6 +697,28 @@ __wt_debug_cursor_page(void *cursor_arg, const char *ofile)
     return (__wt_debug_page(cursor->session, cbt->btree, cbt->ref, ofile));
 }
 
+/*
+ * __wt_debug_cursor_las --
+ *     Dump the LAS tree given a user cursor.
+ */
+int
+__wt_debug_cursor_las(void *cursor_arg, const char *ofile)
```
sulabhM: I am curious if this will dump all of the las from the disk, or only the portion that is cached in-memory ?

keithbostic: It will dump everything, on-disk as well as in-memory.

This is actually @sueloverso's change I picked up. But I've hacked it in a dozen times myself, I agree with her it's generally useful. It would be good if we could dump just the records we care about from the LAS, which is another chunk of code I've hacked in at least a dozen times. :)

sulabhM: Thanks, this will be very useful. @tetsuo-cpp also made an attempt at writing a debug version for las, not sure if he ended up checking it in. I can merge develop to relevant history and get this functionality in there.

sueloverso: If you're running test/format, you might want the changes in the PR for WT-5211 also. Those are ready to be merged once they're reviewed.

PR #4984, opened 2019-11-19
WT-4995 Migrate Jenkins wiredtiger-test-wtperf job to Evergreen

raviprakashgiri29:
Added a new task "test-wtperf" under Ubuntu 18.04 buildvariant

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -1606,7 +1606,7 @@ tasks:
           script: |
             set -o errexit
             set -o verbose
-            # Python test suite generates WT_TEST directory
+
             dir=../bench/wtperf/stress
             for file in `ls $dir`
             do
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -1587,8 +1587,6 @@ tasks:
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
   - name: test-wtperf
-    depends_on:
-      - name: compile
     commands:
       - func: "get project"
       - func: "compile wiredtiger"
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -231,6 +231,13 @@ tasks:
         vars:
           posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-strict --enable-python
 
+  - name: compile-wtperf
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-strict --enable-diagnostic
+
   - name: make-check-test
     depends_on:
       - name: compile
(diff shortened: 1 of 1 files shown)


test/evergreen.yml
```diff
@@ -1586,6 +1586,27 @@ tasks:
...
+          script: |
+            set -o errexit
+            set -o verbose
+
+            mkdir WT_TEST
+            dir=../bench/wtperf/stress
+            for file in `ls $dir`
+            do
+              ./bench/wtperf/wtperf -O $dir/$file -o verbose=2
+              rm -rf WT_TEST/*
```
quchenhao: Can we change the code not overwriting WT_TEST?

test/evergreen.yml
```diff
@@ -1586,6 +1586,27 @@ tasks:
           display_name: Coverage report
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
+  - name: test-wtperf
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-strict --enable-diagnostic
```
quchenhao: I think it's better to separate the compile task but in this case we only have one test. It's up to you.

test/evergreen.yml
quchenhao: We should not remove the test output in evergreen.

raviprakashgiri29:
Evergreen patch build

https://evergreen.mongodb.com/version/5dd356532fbabe0ba751883c

test/evergreen.yml
lukech: I agree this should be separated, actually reusing the existing `compile` task is good enough (as . the configure options are a subset of that of `compile` task).

test/evergreen.yml
```diff
@@ -1586,6 +1586,27 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            mkdir WT_TEST
```
lukech: I know Python test suite will generate `WT_TEST` directory automatically, not sure about wtperf. We could save this line (though a comment would be helpful) if it's the case.

test/evergreen.yml
lukech: After rethinking this, I agree the test output could be useful for troubleshooting when the test fails. 
Even though Jenkins did not save the directory, we can simply do it by replacing this line with `cp WT_TEST WT_TEST_$file`.

test/evergreen.yml
```diff
@@ -1586,6 +1586,27 @@ tasks:
           display_name: Coverage report
           remote_file: wiredtiger/${build_variant}/${revision}/coverage_report/coverage_report_${build_id}.html
 
+  - name: test-wtperf
```
lukech: Can we call it `wtperf-test` (to be consistent with existing Evergreen tasks)?

test/evergreen.yml
raviprakashgiri29: Ok. Let me confirm and then remove the line if needed.

test/evergreen.yml
```diff
@@ -1586,6 +1593,27 @@ tasks:
...
+      - func: "fetch artifacts"
+        vars:
+          dependent_task: compile-wtperf
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+            # Python test suite generates WT_TEST directory
```
lukech: This comment is a bit confusing here as we're not running Python test suite in this task.

PR #4985, opened 2019-11-19
WT-5250 Stop relying on all entries for a given page being contiguous in the lookaside table

tetsuo-cpp:
This patch is adding functionality to store all keys for a page in memory when lookaside evicting and to use this information to instantiate when reading the page back into cache. On instantiation, we iterate over our stored keys and do a `search_near` to find the newest record for that key unless it is a birthmark which is indicated by an "aborted" transaction id.

This was originally going to be part of #4973 however we've been having trouble chasing down bugs in that branch and we know for a fact that this change works by itself.

### DIFF
--- a/src/include/btmem.h
+++ b/src/include/btmem.h
@@ -248,16 +248,12 @@ struct __wt_page_lookaside {
     wt_timestamp_t max_ondisk_ts;  /* Maximum timestamp on disk */
     wt_timestamp_t min_skipped_ts; /* Skipped in favor of disk version */
     bool has_prepares;             /* One or more updates are prepared */
-    // Delete max/min keys, not needed anymore
-    WT_ITEM max_las_key; /* The maximum key in the LAS for the page */
-    WT_ITEM min_las_key; /* The minimum key in the LAS for the page */
     struct __wt_birthmark_details {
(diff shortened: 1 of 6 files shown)


src/include/btmem.h
```diff
@@ -248,17 +248,14 @@ struct __wt_page_lookaside {
     wt_timestamp_t max_ondisk_ts;  /* Maximum timestamp on disk */
     wt_timestamp_t min_skipped_ts; /* Skipped in favor of disk version */
     bool has_prepares;             /* One or more updates are prepared */
-    WT_ITEM max_las_key;           /* The maximum key in the LAS for the page */
-    WT_ITEM min_las_key;           /* The minimum key in the LAS for the page */
     struct __wt_birthmark_details {
```
tetsuo-cpp: I'd like to rename this from `WT_BIRTHMARK_DETAILS` to `WT_KEY_MEMENTO` since this only indicates a birthmark when `txnid == WT_TXN_ABORTED`. What do people think of this?

tetsuo-cpp:
@keithbostic No need to review this one. @sulabhM was able to get #4973 in a reasonable state (with my changes) so we'll continue with that branch.

PR #4988, opened 2019-11-20
WT-5251 Increase frequency for Linux syscall test

lukech:
This is the change to increase test frequency for the Linux syscall test. It used to be triggered to run every week, and with this change it will run for each commit merge.


lukech:
@quchenhao could you help to take a look? Thanks.

PR #4989, opened 2019-11-21
WT-4565 Remove loop to wait in __tree_walk_internal() for an internal page

GWlodarek:
With the changes that you proposed @keithbostic, we will no longer be spinning on the __wt_page_swap() call in __tree_walk_internal() when it encounters a locked ref.


wtbuild:

lukech:
test this please

PR #4990, opened 2019-11-21
WT-5033 Migrate Jenkins wiredtiger-test-time-shift-sensitivity job to evergreen

raviprakashgiri29:
Created compile-enable-strict and time-shift-sensitivity-test task

### DIFF
--- a/test/csuite/time_shift_test.sh
+++ b/test/csuite/time_shift_test.sh
@@ -42,7 +42,7 @@ then
     exit $EXIT_FAILURE
 fi
 
-# Find test_rwlock file from the wiredtiger home directory.
+# Locate Wiredtiger home directory.
 WT_HOME_DIR=$(git rev-parse --show-toplevel)
 
 SEC1=`date +%s`
--- a/test/csuite/time_shift_test.sh
+++ b/test/csuite/time_shift_test.sh
@@ -43,12 +43,12 @@ then
 fi
 
 # Locate Wiredtiger home directory.
-WT_HOME_DIR=$(git rev-parse --show-toplevel)
+RW_LOCK_FILE=$(git rev-parse --show-toplevel)/build_posix/test/csuite/test_rwlock
 
 SEC1=`date +%s`
 if [ "$RUN_OS" = "Darwin" ]
 then
-    $WT_HOME_DIR/build_posix/test/csuite/test_rwlock
+    $RW_LOCK_FILE
 elif [ "$RUN_OS" = "Linux" ]
 then
     if [ -z $2 ]
@@ -57,7 +57,7 @@ then
     else
         CPU_SET=$2
     fi
-    taskset -c $CPU_SET $WT_HOME_DIR/build_posix/test/csuite/test_rwlock
+    taskset -c $CPU_SET $RW_LOCK_FILE
 else
     echo "not able to decide running OS, so exiting"
     exit $EXIT_FAILURE
@@ -71,9 +71,9 @@ if [ "$RUN_OS" = "Darwin" ]
 then
     export DYLD_FORCE_FLAT_NAMESPACE=y
     export DYLD_INSERT_LIBRARIES=$1
-    $WT_HOME_DIR/build_posix/test/csuite/test_rwlock &
+    $RW_LOCK_FILE &
 else
-    LD_PRELOAD=$1 taskset -c $CPU_SET $WT_HOME_DIR/build_posix/test/csuite/test_rwlock &
+    LD_PRELOAD=$1 taskset -c $CPU_SET $RW_LOCK_FILE &
 fi
 
 # get pid of test run in background
--- a/test/csuite/time_shift_test.sh
+++ b/test/csuite/time_shift_test.sh
@@ -42,9 +42,11 @@ then
     exit $EXIT_FAILURE
 fi
 
-# Find test_rwlock file from the wiredtiger directory(two level up ../../) level.
-rwlockfile=`find ../../ -type f -name "test_rwlock" -print -quit`
-if [ -z $rwlockfile ]
+# Find test_rwlock file from the wiredtiger home directory.
+WT_HOME_DIR=$(git rev-parse --show-toplevel)
+echo "======== top level Directory : $WT_HOME_DIR ======="
+RW_LOCK_FILE=`find $WT_HOME_DIR -type f -name "test_rwlock" -print -quit`
+if [ -z $RW_LOCK_FILE ]
 then
     echo "test_rwlock file not found, so exiting"
     exit $EXIT_FAILURE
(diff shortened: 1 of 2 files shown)


test/evergreen.yml
```diff
@@ -1614,6 +1621,25 @@ tasks:
...
+        params:
+          working_dir: "wiredtiger/build_posix/test/csuite"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            rundir=../../../test/csuite
+            cp $rundir/time_shift_test.sh ./
+            chmod a+x ./time_shift_test.sh
+            ./time_shift_test.sh /usr/local/lib/faketime/libfaketimeMT.so.1 0-1
```
quchenhao: PRE_LOAD /usr/local/lib/faketime/libfaketimeMT.so.1 in the test_env_vars. 
Can we get rid of these?
rundir=../../../test/csuite
cp $rundir/time_shift_test.sh ./
chmod a+x ./time_shift_test.sh

test/evergreen.yml
```diff
@@ -1704,6 +1730,8 @@ buildvariants:
     - name: make-check-test
     - name: unit-test
     - name: fops
+    - name: compile-enable-strict
+    - name: time-shift-sensitivity-test
```
lukech: The Jenkins job is currently configured to run once per day. If you want to move to a per-commit-merge testing frequency, you may want to trigger a talk with the team.

test/evergreen.yml
```diff
@@ -238,6 +238,13 @@ tasks:
         vars:
           posix_configure_flags: --enable-strict --enable-diagnostic
 
+  - name: compile-enable-strict
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-strict
+
```
lukech: There's nothing special for this configure flag setting compared to the default. I'd prefer to reuse the default `compile` task instead of creating a new one. I know it's safer to do an apple-to-apple translation and could understand if you prefer to keep this way.

test/evergreen.yml
lukech: The script copying code here makes me sick :) 
I know it's from the Jenkins job, and suggest to try and see if we can call the script directly without copying it over, or if the copying is necessary see if it's possible to improve the script to make it runnable under any directory.

raviprakashgiri29: I tried to run the script(time_shift_test.sh) directly without copying but it internally calls another script (test_rwlock) which is located at the copied to location. I will discuss with you if I can change the script.

test/evergreen.yml
raviprakashgiri29: I can either reuse "compile-enable-strict" task in "checkpoint-filetypes-test" which compiles only with --enable-strict or use default compile as you suggested.

test/evergreen.yml
raviprakashgiri29: Modified the script to run without copying to some location.

test/evergreen.yml
raviprakashgiri29: Using the default compile task.

test/evergreen.yml
raviprakashgiri29: As agreed with the team, the evergreen job will run at per-commit-merge frequency.

test/csuite/time_shift_test.sh
```diff
@@ -42,10 +42,18 @@ then
     exit $EXIT_FAILURE
 fi
 
+# Find test_rwlock file from the wiredtiger directory(two level up ../../) level.
+rwlockfile=`find ../../ -type f -name "test_rwlock" -print -quit`
```
quchenhao: I think it's better to use ```git rev-parse --show-toplevel``` to go to the wiredtiger home directory.

quchenhao: And then just use relative path to find the ```test_rwlock``` file instead of using find.

raviprakashgiri29: Updated the script to use relative path.

test/csuite/time_shift_test.sh
```diff
@@ -42,10 +42,19 @@ then
     exit $EXIT_FAILURE
 fi
 
+# Find test_rwlock file from the wiredtiger home directory.
+WT_HOME_DIR=$(git rev-parse --show-toplevel)
+RW_LOCK_FILE=`find $WT_HOME_DIR -type f -name "test_rwlock" -print -quit`
```
quchenhao: Why not just $WT_HOME_DIR/test/csuite/test_rwlock?

test/evergreen.yml
```diff
@@ -1614,6 +1614,24 @@ tasks:
...
+      - name: compile
+        vars:
+          posix_configure_flags: --enable-strict
+    commands:
+      - func: "fetch artifacts"
+        vars:
+          posix_configure_flags: --enable-strict
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
```
lukech: Why not setting to `wiredtiger/build_posix/test/csuite` directly?

test/evergreen.yml
```diff
@@ -1614,6 +1614,24 @@ tasks:
...
+          script: |
+            set -o errexit
+            set -o verbose
+
+            ../test/csuite/time_shift_test.sh /usr/local/lib/faketime/libfaketimeMT.so.1 0-1
```
lukech: Assume the library file exists on the `ubuntu1804-test` distro?

test/csuite/time_shift_test.sh
raviprakashgiri29: Yes, it can be hardcoded to call the binary (test_rwlock) as it was in the script. But I have added a validation check to exit if the file doesn't exist.

quchenhao: Then I think the validation check can also be removed.

test/evergreen.yml
raviprakashgiri29: faketime library has been installed in ubuntu18.04 and rhel80 as per ticket https://jira.mongodb.org/browse/BUILD-9213

test/evergreen.yml
raviprakashgiri29: It cannot be set to " wiredtiger/build_posix/test/csuite" as we are not copying the script file under build_posix level. But however, I will set the dir to "wiredtiger/test/csuite".

test/csuite/time_shift_test.sh
raviprakashgiri29: Removed

raviprakashgiri29:
> Could you trigger the new test in your Evergreen patch build? I did not find the test in the PR patch.

Evergreen patch build with the latest changes

https://evergreen.mongodb.com/version/5ddc8fd83e8e867c8abb45c0

PR #4991, opened 2019-11-22
WT-5232 Create a wrapper script to support format stress tests in Evergreen

keithbostic:
@lukech, here's my first cut at an Evergreen wrapper script for `format`, let me know what you think.


test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+		shift ; shift ;;
+	-F)
+		first_failure=1
+		shift ;;
+	-h)
+		home="$2"
+		shift ; shift ;;
+	-j)
+		jobs="$2"
+		[[ '^[1-9]+$' =~ "$jobs" ]] && {
```
quchenhao: I think the regular expression is not correct. 10 will be considered an illegal input.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+			echo "$0: -j option argument must be a non-zero integer"
+			exit 1
+		}
+		shift ; shift ;;
+	-n)
+		runs="$2"
+		[[ '^[1-9]+$' =~ "$runs" ]] && {
```
quchenhao: Same here.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+			echo "$0: -n option argument must be an non-zero integer"
+			exit 1
+		}
+		shift ; shift ;;
+	-S)
+		smoke_test=1
+		shift ;;
+	-t)
+		minutes="$2"
+		[[ '^[1-9]+$' =~ "$minutes" ]] && {
```
quchenhao: Same here.

keithbostic:
> I use parallel -j 20 -X ./t -1 -c CONFIG.5136 -h HOME-{} ::: 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 to run test format in parallel. It starts 20 jobs at the same time but it doesn't start a new job when a job finishes. Maybe we can use this as a base to simplify the code a bit.

I don't believe `parallel` is always installed on a vanilla system. I didn't use xargs for the same reason, I made the script standalone.

test/format/format.sh
keithbostic: Thanks, fixed in 4ef126e.

test/format/format.sh
keithbostic: Thanks, fixed in 4ef126e.

test/format/format.sh
keithbostic: Thanks, fixed in 4ef126e.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+		shift ; shift ;;
+	-F)
+		first_failure=1
+		shift ;;
+	-h)
+		home="$2"
+		shift ; shift ;;
+	-j)
+		jobs="$2"
+		[[ '^[0-9]+$' =~ "$jobs" ]] && {
```
quchenhao: Should 0 still be invalid? What about ```^[1-9][0-9]*$```

keithbostic: I flat-out don't understand bash regex parsing/quoting, but whatevs.  I pushed something in 802fa73 that appears to work on everything I threw at it.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+		shift ; shift ;;
+	-F)
+		first_failure=1
+		shift ;;
+	-h)
+		home="$2"
+		shift ; shift ;;
+	-j)
+		jobs="$2"
+		[[ "$jobs" =~ ^[1-9][0-9]?$ ]] || {
```
quchenhao: Are we also limiting the maximum number of jobs to 99? If so, better to indicate that in the error message as well.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+			echo "$0: -j option argument must be a non-zero integer"
+			exit 1
+		}
+		shift ; shift ;;
+	-n)
+		runs="$2"
+		[[ "$runs" =~ ^[1-9][0-9]?$ ]] || {
```
quchenhao: Same here.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+			echo "$0: -n option argument must be an non-zero integer"
+			exit 1
+		}
+		shift ; shift ;;
+	-S)
+		smoke_test=1
+		shift ;;
+	-t)
+		minutes="$2"
+		[[ "$minutes" =~ ^[1-9][0-9]?$ ]] || {
```
quchenhao: I think it's too restrictive to limit the running time to below 99 minutes.

keithbostic: Agreed, fixed in 60b7af1.

test/format/format.sh
keithbostic: Agreed, fixed in 60b7af1.

test/format/format.sh
keithbostic: Agreed, fixed in 60b7af1.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+	[[ $seconds -ne 0 ]] && {
+		now="$(date -u +%s)"
+		elapsed=$(($now - $start_time))
+
+		# If we've run out of time, terminate all running jobs.
+		[[ $elapsed -ge $seconds ]] && {
+			stop=1
+			forcequit=1
+		}
+	}
```
lukech: As this code block locates at the beginning of the outer while loop, the whole test could run longer compared to the configured value `-t <mins>`, as the inner while loop could take time (depending on the `-n <runs>` and `-j <jobs>` settings, and there's an additional 10 sec sleep at the end of the outer while loop.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+		break ;;
+	esac
+done
+format_args="$*"
+
+# Find the format and wt binaries (the latter is only required for abort/recovery testing).
+# Find the configuration file and run directory.
+build=$(dirname $0)
+format_binary="$build/t"
+wt_binary="$build/../../wt"
```
lukech: These directory settings feel assuming compile is under the top of wiredtiger tree. In Evergreen the standard way is to build under `build_posix` directory. I simulated the Evergreen compile way and ran into error locating the `t` binary.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+resolve()
+{
+	for dir in $home/RUNDIR.[0-9]*; do
+		# Skip log files and failures we've already reported.
+		[[ ! -d $dir ]] || [[ -f "$dir/reported" ]] && continue
+		log="$dir.log"
+
+		# Discard successful runs.
+		grep 'successful run completed' $log > /dev/null && {
+			rm -rf $dir $log
```
lukech: Do we write useful information e.g. format args, configuration in the log file for successful runs? If yes I'd keep them for ease of comparison with failure runs.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+				failure=$(($failure + 1))
+				echo "$0: job in $dir failed abort/recovery testing"
+			fi
+			continue
+		}
+
+		# Discard runs where the timer went off.
+		grep 'caught signal' $log > /dev/null && {
+			rm -rf $dir $log
+			[[ $verbose -ne 0 ]] && echo "$0: job in $dir aborted"
```
lukech: Considering making `[[ $verbose -ne 0 ]] && echo` a variable.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+			continue
+		}
+
+		# Report failures.
+		grep 'RUN PARAMETERS' $log > /dev/null && {
```
lukech: It's not super clear to me why `'RUN PARAMETERS'` indicates a failure. Some further comment would be helpful.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+    echo "$0 [-c $config] [-h $home] [-j $jobs] [-n $runs] [-t $minutes] $format_args"
+
+failure=0
+success=0
+# Resolve/cleanup completed jobs.
+resolve()
+{
+	for dir in $home/RUNDIR.[0-9]*; do
+		# Skip log files and failures we've already reported.
+		[[ ! -d $dir ]] || [[ -f "$dir/reported" ]] && continue
```
lukech: Not sure about the usefulness of the `[[ -f "$dir/reported" ]]` check. Would a dir be hit twice in the for loop?

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+		args=$format_args
+
+		# If abort/recovery testing is configured, do it 5% of the time.
+		[[ $abort_test -ne 0 ]] && [[ $(($count_runs % 20)) -eq 0 ]] && args="$args abort=1"
+
+		echo "$0: starting job in $dir"
+	fi
+	[[ $verbose -ne 0 ]] &&
+	    echo "$0: running $format_binary -c "$config" -h "$dir" -1 $args quiet=1"
+	$format_binary -c "$config" -h "$dir" -1 $args quiet=1 > $log 2>&1 &
```
lukech: Prefer to put the command into a variable, so that we don't need to remember updating 2 places next time there's a need to tweak the command.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+		grep 'RUN PARAMETERS' $log > /dev/null && {
+			echo "$0: failure status reported" > $dir/reported
+			failure=$(($failure + 1))
+			echo "$0: job in $dir failed"
+		}
+	done
+	return 0
+}
+
+# Start a single job.
```
lukech: Does it actually mean a run (set by `-n <runs>`), or a job (set by `j <jobs>`)?

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+
+seconds=$((minutes * 60))
+start_time="$(date -u +%s)"
+while :; do
+	# Check if our time has expired.
+	[[ $seconds -ne 0 ]] && {
+		now="$(date -u +%s)"
+		elapsed=$(($now - $start_time))
+
+		# If we've run out of time, terminate all running jobs.
```
lukech: Can we add a verbose logging here? It would be useful to see when the test actually runs out of time.

test/format/format.sh
```diff
@@ -0,0 +1,301 @@
...
+	    echo "$0: $success successful runs, $failure failed runs"
+
+	# Quit if we're done and there aren't any jobs left to wait for.
+	children=$(pgrep -P $$)
+	[[ $stop -ne 0 ]] && [[ -z "$children" ]] && break;
+
+	# Forcibly quit if there's a failure and first-failure configured.
+	[[ $failure -ne 0 ]] && [[ $first_failure -ne 0 ]] && {
+		forcequit=1
+		stop=1
```
lukech: There are multiple instances of setting both flags in the script. I guess the idea here is to make flag setting and handling unsynchronised?

sueloverso:
`I initially compiled under build_posix directory, and the script failed to locate the t binary.`

FYI, this is how I generally build my trees still also.

test/format/format.sh
keithbostic: Yes, you are correct. I didn't try to make format run for a specific number of seconds, it will quite likely take some number of seconds to exit.

Can you discuss why that's a problem?

test/format/format.sh
keithbostic: I don't think we want to do this, a successful run can use up a lot of disk space and multi-day runs will run out of disk space.

test/format/format.sh
keithbostic: Can you please give me an example of what you're suggesting? (This is probably my bash inexperience here.)

test/format/format.sh
keithbostic: The idea is that if `-F` (exit on first failure) is not set, we can accumulate failures over the run, and we only want to report each failure the first time we find it.

test/format/format.sh
keithbostic: Yes; there's a worker loop that handles killing running processes and waiting for them, let me know if there's a better way to do this.

If we're willing, for example, to guarantee that `GNU parallel` or `xargs` are installed everywhere we want to run this script, we might be able to simplify things.

test/format/format.sh
keithbostic: Fixed in 80667f7a0.

test/format/format.sh
keithbostic: I changed `format` to output an explicit failure message instead of checking for `RUN PARAMETERS`, fixed in 5dff87605.

test/format/format.sh
keithbostic: Fixed in 4261b9a93.

test/format/format.sh
keithbostic: The idea is a `job` is a single run of `format`, a maximum of `-n total-jobs` are executed, and no more than `-j parallel-jobs` are executed at once. I've reworked the script in 303cb3a62 to hopefully clarify it.

test/format/format.sh
keithbostic: Sure, done in deaa8eea4.

keithbostic:
> I initially compiled under build_posix directory, and the script failed to locate the t binary.

I believe this is fixed in 80667f7.

> I ran time test/format/format.sh -t 1 -v a few times, and the number of job runs being executed were different (15, 10, 11, etc), which I believe is expected as it depends on how fast each job run takes - the faster those job runs execute the more job runs would be fitted in.

Yes.

> The successful/failed runs summary was printed out multiple times, one from the outer while loop, and the others from inner while loop - feels a bit redundant though I could be wrong.

The successful/failed runs summary is intended to be printed out every time it's checked and it changes (so roughly every 10 seconds if it changes in that 10 seconds), and then once when the script completes. I'm happy to make that more/less verbose to meet your needs.

> A bit confusing about the term job referred to within the script, there's a -j <jobs> configured for the number of simultaneous runs, and another within the scope of format program I believe. Could we distinguish them in the script code/comments?

I've tried to do that in 303cb3a, let me know if there are additional changes that would help.

> It's not clear which line in the smoke_list is for the overflow testing.

Fixed in 73a6366dc.

> The output of the format run is saved in the log file. Would that make troubleshooting less efficient - in the sense people need to download & extract the artifacts to be able to see the detailed failure reason, compared to see the failure dumped out as the shell output?

Maybe?

When we see each failure, I'm happy to dump the output to stdout, I think this is your call to make.

test/format/format.sh
lukech: Fair enough

test/format/format.sh
lukech: Thanks for making it clear.

test/format/format.sh
lukech: Awesome, thanks.

test/format/format.sh
lukech: I don't think `GNU parallel` is preinstalled on all Evergreen distros, and it's better not to depend on that.

test/format/format.sh
lukech: I don't think it's a problem - it's not a big deal to have some additional sleep/exit time compared to the configured minutes to run. Just hope to line up expectation for the `-t <mins>` setting.

test/format/format.sh
lukech: Oh, I see.

test/format/format.sh
lukech: Looks better, thanks.

test/format/format.sh
lukech: Looks like using a function is more appropriate. I pushed 905513e.

lukech:
@keithbostic I prefer to dump the command output/error to stdout when we see a failure, to make test failure triage easier. 

This morning I run the command with `-h` hoping it gives me the help menu, and then realized the command argument `-h` has been used to set the home directory. 
`$ ../test/format/format.sh -h
format.sh: directory "" not found`
It would be great (and intuitive) if we can use `-h` for help menu print, maybe `-d` for the home directory setting instead?

sueloverso:
@lukech for WT specifically, `-h` is far more intuitive for home directory to me. It is used *everywhere* including other scripts. I thought that `-?` is most standard and intuitive for typical commands.

lukech:
@sueloverso thanks for letting me know. I was not aware of `-h` being widely used in other places. Happy to adapt to it.

test/format/format.sh
keithbostic: Looks good.

sueloverso:
@lukech it is not widely used in other places. But `-h` is consistently used in anything relating to WiredTiger.

keithbostic:
> I prefer to dump the command output/error to stdout when we see a failure, to make test failure triage easier.

Fixed in f91ffa24d.

keithbostic:
> This morning I run the command with -h hoping it gives me the help menu, and then realized the command argument -h has been used to set the home directory.

@sueloverso is correct that in WiredTiger sources generally use `-h` to mean `home directory`. I wasn't sure if it was widespread or not, so I looked: examples include `recover.sh`, `compatibility_test_for_mongodb_releases.sh`, `verify_wt_datafiles.sh`, `smoke.sh`,`wtperf_xray.sh` and `workgen_stat.sh`, as well as the `wt` and `wtperf` commands.

So, I guess I agree with @sueloverso, we should leave `-h` as `home`.

PR #4993, opened 2019-11-25
WT-5139 Incremental backup API

sueloverso:
Not ready for merge.


sueloverso:
@keithbostic and @kommiharibabu I think this branch is ready for review. This branch is the initial work for incremental backup. It implements the API and the incremental cursor, but the incremental cursor always returns full-file information only. There is some code that in preparation for the next step/ticket of maintaining block information but that code is not yet used.

If you'd prefer a cleaner change, I can remove and simplify here. There are 2 new tests and 1 new example program.

dist/api_data.py
```diff
@@ -1208,6 +1209,37 @@ def __ge__(self, other):
         characters are hexadecimal encoded.  These formats are compatible
         with the @ref util_dump and @ref util_load commands''',
         choices=['hex', 'json', 'print']),
+    Config('incremental', '', r'''
+        configure the cursor for block incremental backup usage. These formats
+        are only compatible with the backup data source; see @ref backup''',
+        type='category', subconfig=[
+        Config('enabled', 'false', r'''
+            whether to configure this backup as an incremental starting point''',
```
keithbostic: "an incremental starting point" -> "the starting point for a subsequent incremental backup" ?

dist/api_data.py
```diff
@@ -1208,6 +1209,37 @@ def __ge__(self, other):
...
+            type='boolean'),
+        Config('file', '', r'''
+            the file name of interest when opening a duplicate incremental backup
```
keithbostic: delete "of interest" ?

dist/api_data.py
```diff
@@ -1208,6 +1209,37 @@ def __ge__(self, other):
...
+            cursor. That duplicate cursor will walk the block modifications relevant
```
keithbostic: "will walk the" -> "will return the" ?

dist/api_data.py
```diff
@@ -1208,6 +1209,37 @@ def __ge__(self, other):
...
+            to the given file name'''),
+        Config('force_stop', 'false', r'''
+            causes all block incremental backup information to be released. This is
+            on an open_cursor call and the resources will be released when this
+            cursor is closed. No other operations should be done on this open cursor''',
+            type='boolean'),
+        Config('granularity', '16MB', r'''
+            this setting manages the granularity of how WiredTiger maintains modification
+            maps internally. The larger the granularity, the smaller amount of information
+            we need to maintain''',
```
keithbostic: "we need" -> "WiredTiger needs" ?

dist/api_data.py
```diff
@@ -1208,6 +1209,37 @@ def __ge__(self, other):
...
+            min='1MB', max='2GB'),
+        Config('src_id', '', r'''
```
keithbostic: "src_id" -> "source_id"
The contraction isn't buying us a lot?

dist/api_data.py
```diff
@@ -1208,6 +1209,37 @@ def __ge__(self, other):
...
+            a string that identifies a previous checkpoint backup source as the source
+            of this incremental backup. This identifier must have already been created
+            by use of the 'this_id' configuration in an earlier backup. A source id is
+            required to begin an incremental backup'''),
+        Config('this_id', '', r'''
+            a string that identifies the most recent checkpoint as a future backup source
+            for an incremental backup via 'src_id'. This identifier is required and an
+            error will be returned if one is not provided'''),
```
keithbostic: Doesn't this only need to be specified if we're doing incremental backups?

And, doesn't it identify a backup as the source, not a checkpoint?

src/conn/conn_handle.c
```diff
@@ -131,6 +132,9 @@ __wt_connection_destroy(WT_CONNECTION_IMPL *conn)
     __wt_cond_destroy(session, &conn->lsm_manager.work_cond);
 
     /* Free allocated memory. */
+    for (i = 0; i < WT_BLKINCR_MAX; ++i)
+        __wt_free(session, conn->incr_backups[i].id);
```
keithbostic: Don't these have to be persistent? That is, if we crash, do callers have to start with a full backup?

src/cursor/cur_backup.c
```diff
@@ -157,19 +202,23 @@ __wt_curbackup_open(WT_SESSION_IMPL *session, const char *uri, WT_CURSOR *other,
...
+    if (othercb != NULL)
+        WT_CURSOR_BACKUP_CHECK_STOP(othercb);
 
     /*
      * Start the backup and fill in the cursor's list. Acquire the schema lock, we need a consistent
      * view when creating a copy.
      */
     WT_WITH_CHECKPOINT_LOCK(
-      session, WT_WITH_SCHEMA_LOCK(session, ret = __backup_start(session, cb, other != NULL, cfg)));
+      session, WT_WITH_SCHEMA_LOCK(session, ret = __backup_start(session, cb, othercb, cfg)));
```
keithbostic: Should we split `__backup_start()` into two parts, duplicate and "not duplicate" cursor? They really don't share anything, do they?

src/cursor/cur_backup.c
```diff
@@ -180,6 +229,101 @@ __wt_curbackup_open(WT_SESSION_IMPL *session, const char *uri, WT_CURSOR *other,
...
+        }
+    }
+    /*
+     * We didn't find an entry. This should not happen.
+     */
+    ret = WT_NOTFOUND;
+err:
+    if (blk != NULL)
+        __wt_free(session, blk->id);
+    return (ret);
```
keithbostic: If you fall out of this loop, `blk` won't be NULL so bad stuff will happen.

I'd suggest a stronger message if we can't find a slot for the ID, and "not-found" doesn't seem like the right error.

This might be simpler as:
```
for (i = 0; i < WT_BLKINCR_MAX; ++i)
    if (!F_ISSET(blk, WT_BLKINCR_INUSE))
        break;

if (i == WT_BLKINCR_MAX)
     /* we're in trouble */

/* use the slot */
```

src/cursor/cur_backup.c
```diff
@@ -207,22 +351,204 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
...
+    WT_CONFIG targetconf;
+    WT_CONFIG_ITEM cval, k, v;
+    WT_CONNECTION_IMPL *conn;
+    WT_DECL_ITEM(tmp);
+    WT_DECL_RET;
+    const char *uri;
+    bool incremental_config, is_dup, log_config, target_list;
+
+    conn = S2C(session);
+    *foundp = *incr_only = *log_only = false;
```
keithbostic: Can we initialize returned side-effects separately, before local variables?

src/cursor/cur_backup.c
```diff
@@ -207,22 +351,204 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
...
+            WT_RET(__wt_config_gets(session, cfg, "incremental.granularity", &cval));
+            /* XXX may not need cb->incr_granularity */
+            if (conn->ckpt_incr_granularity != 0)
+                WT_RET_MSG(session, EINVAL, "Cannot change the incremental backup granularity");
+            conn->ckpt_incr_granularity = cb->incr_granularity = (uint64_t)cval.val;
+        }
+        /* XXX Granularity can only be set once at the beginning */
+        F_SET(conn, WT_CONN_INCR_BACKUP);
+        incremental_config = true;
+    }
```
keithbostic: Why can't we change the backup granularity? I don't see any point in it, but I don't see any reason it shouldn't work, either.

src/cursor/cur_backup.c
```diff
@@ -207,22 +351,204 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
     return (ret);
 }
 
+/*
+ * __backup_config --
```
keithbostic: This is another routine where it makes sense to me to separate the dup/non-dup cases, they really don't share anything, do they?

src/cursor/cur_backup_incr.c
```diff
@@ -0,0 +1,233 @@
...
+            a += 2;
+            --a_cnt;
+        } else if (b_cnt > 0) {
+            res[0] = b[0];
+            res[1] = b[1];
+            b += 2;
+            --b_cnt;
+        }
+    }
+    *res_cnt = total;
```
keithbostic: I think this all goes away -- let's talk through the plan for how we're getting/merging lists of block modifications (unless you've already decided what you want to do).

src/include/cursor.h
```diff
@@ -32,6 +32,21 @@
...
+ */
+#define WT_BLKINCR_MAX 2
+struct __wt_blkincr {
+    const char *id;        /* User's name for this backup. */
+    const char *ckpt_name; /* Requires WT-5115. All checkpoints must be this name */
+    void *data;
+/* AUTOMATIC FLAG VALUE GENERATION START */
+#define WT_BLKINCR_INUSE 0x1u /* This entry is active */
+#define WT_BLKINCR_VALID 0x2u /* This entry is valid */
+                              /* AUTOMATIC FLAG VALUE GENERATION STOP */
```
keithbostic: Weird indentation?

src/cursor/cur_backup.c
```diff
@@ -180,6 +229,101 @@ __wt_curbackup_open(WT_SESSION_IMPL *session, const char *uri, WT_CURSOR *other,
...
+
+    conn = S2C(session);
+    for (i = 0; i < WT_BLKINCR_MAX; ++i) {
+        blk = &conn->incr_backups[i];
+        /* If it isn't valid, skip it. */
+        if (!F_ISSET(blk, WT_BLKINCR_VALID))
+            continue;
+        if (WT_STRING_MATCH(blk->id, cval->str, cval->len)) {
+            if (F_ISSET(blk, WT_BLKINCR_INUSE))
+                WT_RET_MSG(session, EINVAL, "Incremental backup structure already in use.");
```
keithbostic: Trailing periods don't make sense because we build the finall error message from fragments.

src/cursor/cur_backup.c
```diff
@@ -207,22 +351,204 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
...
+        if (is_dup)
+            WT_RET_MSG(session, EINVAL,
+              "Incremental force stop can only be specified on a primary backup cursor");
+        F_SET(cb, WT_CURBACKUP_FORCE_STOP);
+        return (0);
+    }
+    WT_RET_NOTFOUND_OK(__wt_config_gets(session, cfg, "incremental.enabled", &cval));
+    if (cval.val) {
+
+        if (!F_ISSET(conn, WT_CONN_INCR_BACKUP)) {
```
keithbostic: Odd blank line.

dist/api_data.py
sueloverso: Fixed.

dist/api_data.py
sueloverso: Fixed.

dist/api_data.py
sueloverso: Fixed.

dist/api_data.py
sueloverso: Fixed.

dist/api_data.py
sueloverso: I'm going to push back. I don't think the expanded word buys us a lot either. Both seem clear to me. That was the API published in the tech design document that everyone signed off on and no one suggested this then nor felt it was unclear. So it seems a little gratuitous. But I don't feel super strongly either. @milkie do you have an opinion about which is clearer or easier for MongoDB to use?

dist/api_data.py
sueloverso: Yes to the first question. I'm unclear on what you mean in your second question, so some kind of rewording is probably in order. When a user opens a backup cursor with incremental and `this_id=xxx`, then we associate the current state of the system (as in most recently completed full checkpoint) as the stake in the ground for this identifier. What do you mean by "identify a backup as the source"? To me the backup is the destination not the source.

src/conn/conn_handle.c
sueloverso: Yes. That is a later/different ticket to make things persistent. I added a comment/reminder here to say that.

src/cursor/cur_backup.c
sueloverso: We could. I'll take a run at that. It could be deferred until a later ticket.

src/cursor/cur_backup.c
sueloverso: Good catch. I will fix.

src/cursor/cur_backup.c
sueloverso: Fixed. That was the only message with a trailing period.

src/cursor/cur_backup.c
sueloverso: Fixed.

src/cursor/cur_backup.c
sueloverso: I'll consider this and see how it turns out. I generally like having a single configuration parsing function rather than two because we do want to error so we'd have to check the set anyway. And then if the API changes, it is easier to remember and see we only need to change one location.

src/cursor/cur_backup.c
sueloverso: Fixed.

src/cursor/cur_backup.c
sueloverso: It is discussed in the design. I recall it had to do with allocating elements for the block lists and what we've already persisted, etc. For now it seems like it is not needed. If it is needed later, we can add a new feature ticket to do so.

src/cursor/cur_backup_incr.c
sueloverso: This function is from your original work. It is not being used at the moment because the incremental cursor is only returning "entire file" information.

src/include/cursor.h
sueloverso: Clang-format does bizarre indenting of that for some reason. @tetsuo-cpp do you know why? It is not limited to this one nor is it consistent among our header files. It appears to align the `STOP` with the flag comments. That is unfortunate, IMO. For instance, in `lsm.h` it is all over the place. Grepping for `AUTOMATIC` shows:
```
lsm.h:/* AUTOMATIC FLAG VALUE GENERATION START */
lsm.h:                                     /* AUTOMATIC FLAG VALUE GENERATION STOP */
lsm.h:/* AUTOMATIC FLAG VALUE GENERATION START */
lsm.h:    /* AUTOMATIC FLAG VALUE GENERATION STOP */
lsm.h:/* AUTOMATIC FLAG VALUE GENERATION START */
lsm.h:/* AUTOMATIC FLAG VALUE GENERATION STOP */
```

dist/api_data.py
keithbostic: fbm

dist/api_data.py
keithbostic: What I'm trying to say is that we've divorced checkpoints from incremental backups -- it's incidental in the API that we're using a checkpoint somewhere, right? So, we're identifying "the backup as a future base for an incremental".

src/cursor/cur_backup.c
keithbostic: I'm pretty sure there's no reason to disallow this, but I also don't see any reason anyone would change the setting, either.

src/cursor/cur_backup.c
keithbostic: In general, WiredTiger does not check for configurations that are not applicable to the current API, so by doing this, you're diverging a bit from the rest of the code. I get that it's more likely users will mistakenly use backup configurations with incremental backup configurations, but I'm not sure if that's a compelling argument.

It will be easier to understand/remember if there's a configuration function per API than if there's a configuration function that covers multiple distinct APIs.

src/cursor/cur_backup.c
```diff
@@ -65,6 +67,36 @@ __curbackup_reset(WT_CURSOR *cursor)
...
+    WT_UNUSED(cb);
+    WT_UNUSED(force);
+    conn = S2C(session);
+    /*
+     * Clear flags. Remove file. Release any memory information.
+     */
+    F_CLR(conn, WT_CONN_INCR_BACKUP);
+    for (i = 0; i < WT_BLKINCR_MAX; ++i) {
+        blk = &conn->incr_backups[i];
+        /* If it isn't valid, skip it. */
```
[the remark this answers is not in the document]
sueloverso: Good catch. Fixed.

src/cursor/cur_backup.c
```diff
@@ -180,6 +229,101 @@ __wt_curbackup_open(WT_SESSION_IMPL *session, const char *uri, WT_CURSOR *other,
...
+    WT_DECL_RET;
+    u_int i;
+
+    conn = S2C(session);
+    blk = NULL;
+    for (i = 0; i < WT_BLKINCR_MAX; ++i) {
+        blk = &conn->incr_backups[i];
+        /* If it isn't use we can use it. */
+        if (!F_ISSET(blk, WT_BLKINCR_INUSE)) {
+            if (blk->id != NULL)
```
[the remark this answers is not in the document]
sueloverso: That is very reasonable. I have renamed it `id_str`.

src/cursor/cur_backup.c
```diff
@@ -207,22 +351,204 @@ __backup_log_append(WT_SESSION_IMPL *session, WT_CURSOR_BACKUP *cb, bool active)
...
+    /*
+     * Compatibility checking.
+     *
+     * Log archive cannot mix with log-file based incremental backups (but if a duplicate cursor,
+     * archiving has been temporarily suspended).
+     *
+     * Duplicate backup cursors are only for log targets or block-based incremental backups. But log
+     * targets don't make sense with block-based incremental backup.
+     */
+    if (!is_dup && log_config && FLD_ISSET(conn->log_flags, WT_CONN_LOG_ARCHIVE))
```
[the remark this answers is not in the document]
sueloverso: There is no archive location. Log archiving means simply removal.

src/cursor/cur_backup_incr.c
```diff
@@ -0,0 +1,233 @@
...
+
+#include "wt_internal.h"
+
+/*
+ * __alloc_merge --
+ *     Merge two allocation lists.
+ */
+static void
+__alloc_merge(
+  uint64_t *a, uint64_t a_cnt, uint64_t *b, uint64_t b_cnt, uint64_t *res, uint64_t *res_cnt)
```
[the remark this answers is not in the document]
sueloverso: Since this is placeholder code for now, I'm going to defer these until we're really doing block allocation backups.

PR #4994, opened 2019-11-26
WT-4996 Migrate wiredtiger-test-check-long job to Evergreen

hbokhari:


hbokhari:
The default Evergreen patch testing was configured to include the new`check-test-long-1` and `check-test-long-2` tasks.

test/evergreen.yml
```diff
@@ -134,6 +134,30 @@ functions:
         for i in $(seq ${times|1}); do
           ./t -1 -c ${config|../../../test/format/CONFIG.stress} ${extra_args|}
         done
+  "many dbs test":
```
quchenhao: Why not just reuse ```test format```

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
               cp -rf WT_TEST WT_TEST_$file
             done
 
+  - name: check-test-long-1
```
quchenhao: Can we rename this to something meaningful

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+            ${test_env_vars|} $(pwd)/../test/csuite/random_abort/smoke.sh 2>&1
+            ${test_env_vars|} $(pwd)/../test/csuite/timestamp_abort/smoke.sh 2>&1
+            ${test_env_vars|} $(pwd)/test/csuite/test_truncated_log 2>&1
+
+  - name: check-test-long-2
```
quchenhao: Here as well

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix/bench/wtperf"
+          script: |
+            rundir=../../../bench/wtperf/runners
+            $rundir/wtperf_run.sh $rundir/mongodb-oplog.wtperf
+            num_truncates=`grep "^Truncate count:" ./wtperf.out | cut -d ' ' -f 3`
+            echo "YVALUE=$num_truncates" > ../../truncate_truncate_count.properties
+            $rundir/wtperf_track.sh -c truncate_count $num_truncates
+            ./wtperf -O $rundir/checkpoint-schema-race.wtperf
```
quchenhao: Do you know what this thing does? Can we simplify this by running everything under ```rundir```

test/evergreen.yml
hbokhari: Well it is a long test which is doing whole bunch of stuff and perhaps that's why they called it just test-check-long :) . If I am not mistaken this test is sort of a longer version of `make check all` test with more scenarios. I am open to suggestions on naming it differently.

test/evergreen.yml
hbokhari: The comment from Jenkins shell script for this part of job is "# We want to use the wtperf_track script here to give us a pass fail analysis." I will probably copy/paste the comments to make it more clear. 

As far as running it from `rundir` is concerned there shouldn't be any dramas. But maybe the intention of running the script from `wiredtiger/build_posix/bench/wtperf` was to keep the intermediate results in this particular directory rather than polluting `rundir`. Just a guess!

test/evergreen.yml
lukech: Maybe `long-test`?

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
               cp -rf WT_TEST WT_TEST_$file
             done
 
+  - name: check-test-long-1
+    tags: ["check-long-test"]
```
lukech: Why bothering with a tag setting here?

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+    commands: 
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: ac_cv_func_ftruncate=no
```
lukech: We have existing `compile wiredtiger no linux ftruncate` function. It may worth to confirm if this configure flag setting achieves the same effect as that of the function.

test/evergreen.yml
lukech: Now I see why you break the tests into 2 tasks. The tests included in this task is something about no ftruncate setting. Please note we have an existing task `linux-no-ftruncate` which is scheduled to run once per day.

test/evergreen.yml
lukech: `long-test`?

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+            ${test_env_vars|} $(pwd)/../test/csuite/random_abort/smoke.sh 2>&1
+            ${test_env_vars|} $(pwd)/../test/csuite/timestamp_abort/smoke.sh 2>&1
+            ${test_env_vars|} $(pwd)/test/csuite/test_truncated_log 2>&1
+
+  - name: check-test-long-2
+    tags: ["check-long-test"]
```
lukech: Same

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+  - name: check-test-long-1
+    tags: ["check-long-test"]
+    commands: 
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: ac_cv_func_ftruncate=no
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
```
lukech: Would it be better to see working_dir to `wiredtiger/test/csuite` considering that's where most of the tests invoke script runs?

test/evergreen.yml
hbokhari: Reminiscence of time when I actually started with bunch of tasks. Fused together tasks but didn't remember to remove the tags. Will fix it.

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+    tags: ["check-long-test"]
+    commands:
+      - func: "get project"
+      - func: "configure wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-g -Werror"
+          posix_configure_flags: --enable-silent-rules --enable-diagnostic --disable-static
+      - func: "make wiredtiger"
+        vars:
+          make_command: make all
```
lukech: Not necessary as `make` equal `make all` in wiredtiger. Also the `${smp_command|}` in the function will be missed if setting this way.

test/evergreen.yml
hbokhari: I have to check what is the difference between setting `ac_cv_func_ftruncate=no` and what is being done in `compile wiredtiger no linux ftruncate` to confirm if we are just doubling the effort.

test/evergreen.yml
hbokhari: I have to check what is the difference between setting ac_cv_func_ftruncate=no and what is being done in compile wiredtiger no linux ftruncate to confirm if we are just doubling the effort.

test/evergreen.yml
hbokhari: Sounds reasonable.

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+        vars:
+          test_env_vars: TESTUTIL_ENABLE_LONG_TESTS=1
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            WT3363_CHECKPOINT_OP_RACES=1 $(pwd)/test/csuite/./test_wt3363_checkpoint_op_races
```
lukech: `$(pwd)` feels no different than `.` here except it invokes an extra command execution.

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+      - func: "many dbs test"
+      - func: "many dbs test"
+        vars:
+          many_db_args: -I
+      - func: "many dbs test"
+        vars:
+          many_db_args: -D 40
+      - func: "many dbs test"
+        vars:
+          many_db_args: -I -D 40
```
lukech: Those comments in the Jenkins job are useful info that I'd hope to retain during migration.

test/evergreen.yml
```diff
@@ -134,6 +134,30 @@ functions:
...
         done
+  "many dbs test":
+    command: shell.exec
+    parms:
+      working_dir: "wiredtiger/build_posix/test/manydbs"
+      script: |
+        set -o errexit
+        set -o verbose
+        ${test_env_vars|} ./t ${many_db_args|}
+  "thread check long test":
```
lukech: Prefer `thread test`, though it's being called within the long test task :)

test/evergreen.yml
hbokhari: Yes I now realize it will be a good practice to bring in the original comments given some test are very tricky to understand.

test/evergreen.yml
hbokhari: Noted.

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+      - func: "test format"
+        vars:
+          extra_args: file_type=row
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix/bench/wtperf"
+          script: |
+            rundir=../../../bench/wtperf/runners
+            $rundir/wtperf_run.sh $rundir/mongodb-oplog.wtperf
+            num_truncates=`grep "^Truncate count:" ./wtperf.out | cut -d ' ' -f 3`
```
lukech: Those comments and echo lines from the Jenkins job are useful and should be retained.

test/evergreen.yml
hbokhari: I see - will change it to `make`. FYI, I believe changing the make command doesn't affect `${smp_command|}` parameter.

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+          extra_args: file_type=fix
+      - func: "test format"
+        vars:
+          extra_args: file_type=row
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix/bench/wtperf"
+          script: |
+            rundir=../../../bench/wtperf/runners
+            $rundir/wtperf_run.sh $rundir/mongodb-oplog.wtperf
```
lukech: There could be permission denied issue as I believe the file permission of the shell script is not properly set currently. Would you mind to fix it this time?

test/evergreen.yml
```diff
@@ -1638,6 +1662,105 @@ tasks:
...
+      - func: "test format"
+        vars:
+          extra_args: file_type=fix
+      - func: "test format"
+        vars:
+          extra_args: file_type=row
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix/bench/wtperf"
+          script: |
```
lukech: `set -o errexit` and `set -o verbose` are missed.

test/evergreen.yml
lukech: Oh, you are right.

test/evergreen.yml
```diff
@@ -1637,6 +1661,111 @@ tasks:
...
+      - func: "configure wiredtiger"
+        vars:
+          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/gcc CXX=/opt/mongodbtoolchain/v3/bin/g++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-g -Werror"
+          posix_configure_flags: --enable-silent-rules --enable-diagnostic --disable-static
+      - func: "make wiredtiger"
+      
+      # Run the long version of make check, that includes the full csuite tests
+      - func: "make check all"
+        vars:
+          test_env_vars: TESTUTIL_ENABLE_LONG_TESTS=1
```
quchenhao: Should it be `TESTUTIL_ENABLE_LONG_TESTS=1 LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libeatmydata.so PATH=/opt/mongodbtoolchain/v3/bin:$PATH LD_LIBRARY_PATH=$(pwd)/.libs top_srcdir=$(pwd)/.. top_builddir=$(pwd)`

hbokhari: or just `test_env_vars: ${test_env_vars} TESTUTIL_ENABLE_LONG_TESTS=1` ?

quchenhao: Does that work? If it works, it's great. But I think evergreen doesn't support recursive expansion.

hbokhari: I think it did work for me though I ended up not putting that in the file eventually. I will double check if it works or otherwise I have to put the full string.

hbokhari: It actually works: https://evergreen.mongodb.com/version/5ddf2e9ca4cf47413fe0a2f1

test/evergreen.yml
```diff
@@ -1637,6 +1661,111 @@ tasks:
...
+            set -o errexit
+            set -o verbose
+
+            WT3363_CHECKPOINT_OP_RACES=1 test/csuite/./test_wt3363_checkpoint_op_races 2>&1
+      
+      # Many dbs test - Run with:
+      # 1.  The defaults
+      # 2.  Set idle flag to turn off operations.
+      # 3.  More dbs.
+      #
```
lukech: I actually think the comment in front of each function call (to explain what each function and arg setting is doing) would be more useful.

test/evergreen.yml
```diff
@@ -1637,6 +1661,111 @@ tasks:
               cp -rf WT_TEST WT_TEST_$file
             done
 
+  - name: long-test-1
```
lukech: Though it was included as part of the long test job in Jenkins before, I think we should give it a more meaningful name to describe what type of tests to be executed e.g. `ftruncate-test`, as it's being split into a separate task.

test/evergreen.yml
```diff
@@ -1637,6 +1661,111 @@ tasks:
...
+      
+      # format test
+      - func: "test format"
+        vars:
+          extra_args: file_type=fix
+      - func: "test format"
+        vars:
+          extra_args: file_type=row
+      
+      #FIXME: Add wtperf testing from Jenkin "wiredtiger-test-check-long" after fixing WT-5270
```
lukech: Thanks for adding this FIXME reminder.

test/evergreen.yml
hbokhari: Fixed.

test/evergreen.yml
hbokhari: Good idea. I hav renamed the tests accordingly.

PR #4996, opened 2019-11-26
WT-5082 application threads are tasked with eviction even when pinning the oldest transaction ID

keithbostic:


keithbostic:
@bvpvamsikrishna, I think this is ready for review.

The change in 9d4f212 is the key change that fixes the problem, we weren't correctly setting operation timeout values. @sueloverso, you reviewed the original change, so I've added you as a reviewer here.

Vamsi, the change in 75398bd is a related change, that deserves careful review: I don't fully understand how the snapshot and transaction IDs are related, so I may have messed it up.

src/txn/txn.c
```diff
@@ -1589,82 +1587,62 @@ __wt_txn_global_shutdown(WT_SESSION_IMPL *session, const char *config, const cha
...
+        return (false);
 
     /*
-     * Check if the transaction is oldest one in the system. It's safe to ignore sessions allocating
-     * transaction IDs, since we already have an ID, they are guaranteed to be newer.
+     * Check if we hold the oldest pinned transaction ID in the system. It's safe to ignore sessions
+     * allocating transaction IDs, they are guaranteed to be newer as we already have an ID.
      */
-    for (i = 0, state = txn_global->states; i < session_cnt; i++, state++) {
+    WT_ORDERED_READ(session_cnt, conn->session_cnt);
```
agorrod: I think we could just check against the `txn_global->oldest_id` here, rather than walking and trying to figure out the pinned value.

keithbostic: @agorrod, I've pushed that change. Nice -- that avoids a ton of work.

PR #4997, opened 2019-11-26
WT-5253 Checkpoint history store file

luke-pearson:
So a few tests are failing as a result of changes made to the LAS file (not deleting it every time), it appears recovery is expecting something to happen in a different order to how it is happening. As this ticket is not a fix recovery ticket I think for now I'll leave those tests alone.


src/include/meta.h
```diff
@@ -27,8 +27,9 @@
 #define WT_METAFILE_SLVG "WiredTiger.wt.orig" /* Metadata copy */
 #define WT_METAFILE_URI "file:WiredTiger.wt"  /* Metadata table URI */
 
-#define WT_LAS_FILE "WiredTigerLAS.wt"     /* Lookaside table */
-#define WT_LAS_URI "file:WiredTigerLAS.wt" /* Lookaside table URI*/
+#define WT_LAS_FILE "WiredTigerLAS.wt"       /* Lookaside table */
+#define WT_LAS_URI "file:WiredTigerLAS.wt"   /* Lookaside table URI*/
+#define WT_LAS_URI_HASHED 0xf2f88f2e3d0971dc /* Lookaside table URI Hashed */
```
luke-pearson: I used the LAS hash to determine if we had a dhandle that was pointing to the LAS file. There wasn't a way I could determine that reliably told you if you had a LAS dhandle. You could get the btree from the dhandle and check if the btree LAS flag was set however this isn't set in all code paths, so there are situations where you don't have that flag. I think we could spend time making that certain. Another option was to add a dhandle flag but that has a variety of issues.

src/meta/meta_apply.c
```diff
@@ -10,55 +10,71 @@
...
-            continue;
+    return (ret);
+}
+/*
+ * __meta_btree_apply --
+ *     Walk all files listed in the metadata, apart from the metadata file and LAS file. Applying a
+ *     given function to each file. At the end apply this function to the LAS file.
+ */
+static inline int
+__meta_btree_walk_and_apply(WT_SESSION_IMPL *session, WT_CURSOR *cursor,
```
luke-pearson: __meta_btree_apply_all is used when we are required to open files for checkpointing, which is different to the usual applying checkpoint to only opened files. As such I split out the functionality here and moved the LAS file to the end.

src/conn/conn_dhandle.c
```diff
@@ -575,14 +575,22 @@ __wt_conn_btree_apply(WT_SESSION_IMPL *session, const char *uri,
...
+                continue;
+            }
 
             if (!F_ISSET(dhandle, WT_DHANDLE_OPEN) || F_ISSET(dhandle, WT_DHANDLE_DEAD) ||
               dhandle->type != WT_DHANDLE_TYPE_BTREE || dhandle->checkpoint != NULL ||
               WT_IS_METADATA(dhandle))
                 continue;
             WT_ERR(__conn_btree_apply_internal(session, dhandle, file_func, name_func, cfg));
         }
+        if (las_dhandle != NULL)
```
tetsuo-cpp: Comment might be good to explain why we're leaving this for last.

src/meta/meta_apply.c
```diff
@@ -10,55 +10,71 @@
...
+     * Handles that are busy are skipped without the whole operation failing. This deals among other
+     * cases with checkpoint encountering handles that are locked (e.g., for bulk loads or verify
+     * operations).
      */
-    while ((t_ret = cursor->next(cursor)) == 0) {
-        if ((t_ret = cursor->get_key(cursor, &uri)) != 0 || strcmp(uri, WT_METAFILE_URI) == 0) {
-            WT_TRET(t_ret);
-            continue;
-        }
+    if ((t_ret = __wt_session_get_dhandle(session, uri, NULL, NULL, 0)) != 0) {
```
tetsuo-cpp: Is it really fine to just proceed even if this fails? In this prev code, we were doing `continue` to stop ourselves.

src/txn/txn_ckpt.c
```diff
@@ -639,6 +643,31 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, bool *trackingp, const char *cfg[
...
+    /* Add the history store handle to the checkpoint handle queue. Currently recovery has
+     * an issue with this where the LAS file may not exist yet. As recovery runs checkpoint
+     * and it can run before the lookaside exists we ignore failures.
+     */
+    t_ret = ret;
+    if (session->ckpt_handle_next == 0 ||
+      !WT_IS_LAS(session->ckpt_handle[session->ckpt_handle_next - 1]))
+        WT_WITH_TABLE_READ_LOCK(session,
+          ret = __wt_schema_worker(session, WT_LAS_URI, __wt_checkpoint_get_handles, NULL, cfg, 0));
+    /* Ignore return value of schema worker call for time being. */
```
tetsuo-cpp: Are you doing this just to avoid a compiler warning for ignoring the return value? Afaik, you can just do `(void)__wt_schema_worker(...)`. I've seen us doing this elsewhere in the code.

src/txn/txn_ckpt.c
```diff
@@ -639,6 +643,31 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, bool *trackingp, const char *cfg[
...
+    ret = t_ret;
+
+#ifdef HAVE_DIAGNOSTIC
+    /*
+     * Walk the array of ckpt_handles validating that if the LAS file is in the list, it is at the
+     * end of the list.
+     */
+    for (i = 0; i < session->ckpt_handle_next; i++) {
```
tetsuo-cpp: Nit: Unnecessary braces on the `for` and `if`.

src/include/meta.h
```diff
@@ -37,6 +38,7 @@
  * Optimize comparisons against the metafile URI, flag handles that reference the metadata file.
  */
 #define WT_IS_METADATA(dh) F_ISSET((dh), WT_DHANDLE_IS_METADATA)
+#define WT_IS_LAS(dh) (dh->name_hash == WT_LAS_URI_HASHED)
```
quchenhao: What if there is a collision?

src/meta/meta_apply.c
luke-pearson: Good point, fixed.

src/txn/txn_ckpt.c
luke-pearson: Added WT_IGNORE_RET macro.

src/include/meta.h
luke-pearson: Definitely presents an issue, however low probability.

quchenhao: I don't think the probability is low enough to be ignored given mongodb is widely used. May be not in this PR, but we need to address that I think. Probably better to spend some time to use btree LAS flag instead.

luke-pearson: I think on btree open we could validate that the dhandle name is equal to the LAS uri using strcmp, I think that shouldn't present too much of a performance concern. At which point we set the btree LAS flag and can avoid comparing the hash. @bvpvamsikrishna what do you think?

tetsuo-cpp: Nit: Should be fine in this case but generally, macro arguments should be wrapped in brackets. So `dh => (dh)`.

src/meta/meta_apply.c
```diff
@@ -10,55 +10,72 @@
...
   int (*name_func)(WT_SESSION_IMPL *, const char *, bool *), const char *cfg[])
 {
     WT_DECL_RET;
     int t_ret;
-    const char *uri;
     bool skip;
 
+    skip = false;
+    if (name_func != NULL)
+        ret = name_func(session, uri, &skip);
```
tetsuo-cpp: Do you need to return on failure here too?

src/include/meta.h
luke-pearson: It seems like that issue I was referring to has gone away, either way I've updated the code in the diagnostic block to assert if the LAS file isn't where we expect it to be using strcmp. As such I've removed the hash comparison code and am only using the btree flag.

src/conn/conn_dhandle.c
```diff
@@ -575,14 +575,23 @@ __wt_conn_btree_apply(WT_SESSION_IMPL *session, const char *uri,
...
+            }
 
             if (!F_ISSET(dhandle, WT_DHANDLE_OPEN) || F_ISSET(dhandle, WT_DHANDLE_DEAD) ||
               dhandle->type != WT_DHANDLE_TYPE_BTREE || dhandle->checkpoint != NULL ||
               WT_IS_METADATA(dhandle))
                 continue;
             WT_ERR(__conn_btree_apply_internal(session, dhandle, file_func, name_func, cfg));
         }
+        /* Move the LAS dhandle to the end for checkpointing. */
+        if (las_dhandle != NULL)
```
bvpvamsikrishna: Don't we need to add this part (LAS to the end of dhandle list) incase of user-given checkpoint target list?

luke-pearson: I guess I was of the opinion that if the user specifies targets then we shouldn't add in extra things.

src/txn/txn_ckpt.c
```diff
@@ -847,8 +851,26 @@ __txn_checkpoint(WT_SESSION_IMPL *session, const char *cfg[])
...
+     */
+    WT_ERR(__wt_session_get_dhandle(session, WT_LAS_URI, NULL, NULL, 0));
+    las_dhandle = session->dhandle;
+
+    /*
+     * TODO: Is is possible that we don't have a LAS file in certain recovery scenarios. As such we
+     * could get a NULL dhandle.
+     */
+    if (las_dhandle != NULL) {
+        WT_WITH_DHANDLE(session, las_dhandle, ret = __wt_checkpoint(session, cfg));
```
tetsuo-cpp: What is the reason we do `__wt_checkpoint` instead of `__checkpoint_tree_helper` like every other dhandle.

luke-pearson: Checkpoint tree helper expects certain things like __checkpoint_lock_dirty_tree to have been run on the given dhandle. However we skip the LAS handle in __checkpoint_get_handles as it can set WT_BTREE_SKIP_CKPT on the history store file as the history store file may not have modifications. However as checkpoint runs it'll add modifications to the history store as such we'd not want to skip it. So because __checkpoint_lock_dirty_tree hasn't been called on the LAS handle we need to call it which is done in __wt_checkpoint.

src/txn/txn_ckpt.c
```diff
@@ -261,8 +261,11 @@ __wt_checkpoint_get_handles(WT_SESSION_IMPL *session, const char *cfg[])
 
     btree = S2BT(session);
 
-    /* Skip files that are never involved in a checkpoint. */
-    if (F_ISSET(btree, WT_BTREE_NO_CHECKPOINT))
+    /*
+     * Skip files that are never involved in a checkpoint. Skip the look aside file as it is,
+     * checkpointed manually later.
+     */
+    if (F_ISSET(btree, WT_BTREE_NO_CHECKPOINT) || F_ISSET(btree, WT_BTREE_LOOKASIDE))
```
keithbostic: I think it's odd we skip the metadata file in `__wt_conn_btree_apply()`, but skip the lookaside file here, it seems like handling them both in the same place makes more sense. I wouldn't fight it if that makes the change significantly more difficult.

luke-pearson: __wt_conn_btree_apply() is called in a few more scenarios than checkpoint, which I think we want to include the LAS file in, e.g. __txn_rollback_to_stable_btree. Additionally I think if I included it there then it would make sense to skip it in __wt_meta_apply_all as well, that change is actually quite ugly so I'm a fan of skipping it separately. As such I'll go ahead and merge.

PR #5000, opened 2019-11-28
WT-5254 Adding aggregate timestamp information to WT_REF

luke-pearson:


src/include/btmem.h
```diff
@@ -864,6 +864,14 @@ struct __wt_page_deleted {
 struct __wt_ref {
     WT_PAGE *page; /* Page */
 
+    /* Minimum start time pair. */
```
bvpvamsikrishna: Could we wrap a pair into a __wt_time_pair?

PR #5001, opened 2019-11-28
WT-5274 format.sh must handle core-dump signals and "gdb attach" build mode.

keithbostic:


test/format/format.sh
```diff
@@ -209,16 +226,34 @@ resolve()
 		log="$dir.log"
 
 		# Skip failures we've already reported.
-		[[ -f "$dir/reported" ]] && continue
+		[[ -f "$dir/$status" ]] && continue
+
+		# Get the proceess ID, ignore any jobs that aren't yet running.
+		pid=`grep -E 'process.*running' $log | awk '{print $3}'`
+		[[ "$pid" =~ ^[1-9][0-9]*$ ]] || continue
```
lukech: Does running into here mean the `RUNDIR.*` directory has been created, but the process has not started (or not yet logged the `process.*running` text)? Should the directory be cleaned up in this case?

test/format/format.sh
```diff
@@ -238,32 +273,60 @@ resolve()
...
+			signame="SIGSYS";;
+		esac
+		[[ ! -z $signame ]] && {
+			(echo
+			 echo "$name: job in $dir killed with signal $signame"
+			 echo "$name: there may be a core dump associated with this failure"
+			 echo) >> $log
+
+			echo "$name: job in $dir killed with signal $signame"
+			echo "$name: there may be a core dump associated with this failure"
```
lukech: There's a way to write standard output to both console and file in bash, e.g. `echo "output" | tee -a log.txt`

test/format/format.sh
keithbostic: It means the process hasn't yet logged the `process.*running` text (or maybe we found a `RUNDIR` directory that isn't ours!?), presumably we should keep going.

test/format/format.sh
keithbostic: *shrug*
I'll make the change if you really want it.

keithbostic:
@lukech, I've reworked `format.sh` to handle WiredTiger `gdb attach` builds. I'm just shoving those changes into this branch, I don't see any reason we need a new ticket, but it does mean a new review.

test/format/format.sh
```diff
@@ -195,11 +197,12 @@ config=$(find_file "$config")
 	exit 1
 }
 
-verbose "$name configuration: [-c $config] [-h $home]\
-[-j $parallel_jobs] [-n $total_jobs] [-t $minutes] $format_args"
+verbose "$name configuration: $format_binary [-c $config]\
+[-h $home] [-j $parallel_jobs] [-n $total_jobs] [-t $minutes] $format_args"
```
lukech: Is the purpose of introducing `--` new command argument to separate format arguments from other arguments? And a `--` is suggested to be put in front of `$format_args`  for clarity (though it's optional)?

test/format/format.sh
```diff
@@ -231,12 +238,24 @@ resolve()
...
 
-		# Get the proceess ID, ignore any jobs that aren't yet running.
+		# Get the process ID, ignore any jobs that aren't yet running.
 		pid=`grep -E 'process.*running' $log | awk '{print $3}'`
 		[[ "$pid" =~ ^[1-9][0-9]*$ ]] || continue
 
-		# If the job is still running, ignore it.
-		kill -s 0 $pid > /dev/null 2>&1 && continue
+		# Leave any process waiting for a gdb attach running, but report it as a failure.
+		grep -E 'waiting for debugger' $log > /dev/null && {
```
lukech: Is `waiting for debugger` the format program output that indicates a test is failed and has not been gdb-attached?

test/format/format.sh
```diff
@@ -371,41 +392,35 @@ while :; do
...
+		[[ $seconds -ne 0 ]] && [[ $(($seconds - $elapsed)) -lt 60 ]] && quit=1
 
-		[[ $stop -ne 0 ]] && break;
+		# Don't create more jobs if we're quitting for any reason.
+		[[ $force_quit -ne 0 ]] || [[ $quit -ne 0 ]] && break;
 
 		# Check if the maximum number of jobs in parallel has been reached.
-		n=`jobs -p | wc -l`
-		[[ $n -ge $parallel_jobs ]] && break
+		[[ $running -ge $parallel_jobs ]] && break
```
lukech: Could `$running` become obsolete here considering there could be a 10-sec-sleep after `resolve()` call, e.g. some previously running jobs conclude to either success or failure jobs? Or the strategy here is to use the value calculated last time to avoid the overhead of another resolve() call?

lukech: Those status counters will line up as part of the next resolve() call, probably not a big deal spawning some extra parallel jobs for a certain period of time.

test/format/format.sh
keithbostic: The only purpose is to separate `format.sh` arguments from `format` arguments, so, for example, if you want to pass `-l` to `format`, you have to be sure `format.sh` doesn't parse it, so you use `--`.

test/format/format.sh
keithbostic: `waiting for debugger` is the `format` output that means a process is waiting for `gdb` to attach to it.

test/format/format.sh
keithbostic: Running is the current count of running jobs, so I don't think it can be obsolete, it's how we know how many new jobs to start. Or am I misunderstanding you?

test/format/format.sh
lukech: I'm curious how a format process could be set to a state waiting for gdb attachment. Is it a `format` program argument?

test/format/format.sh
lukech: I mean the `resolve()` call will calculate how many jobs are running/success/failed, but if the 10-sec-sleep is invoked right after, and then we run into here again checking `$running` value that means we lost the past 10-sec job status update (some jobs running-->success, some running-->failure) without called another `resolve()`?

test/format/format.sh
keithbostic: It's a `format` build option, `--enable-attach`, which causes certain types of failures to print out a 
 message and then spin, waiting for `gdb` to attach.

test/format/format.sh
keithbostic: Yes; I don't think that's a problem, though, do you?

We'll loop until we have the right number of processes running (and we won't sleep unless the right number of processes, as calculated based on the results from `resolve()`, are running).

test/format/format.sh
```diff
@@ -178,12 +180,12 @@ format_binary=$(find_file "t")
...
-    [[ ! -x "$wt_binary" ]] && {
-	echo "$name: wt program \"$wt_binary\" not found"
-	exit 1
-    }
+[[ $abort_test -ne 0 ]] || [[ $smoke_test -ne 0 ]] && {
+	wt_binary=$(find_file "wt")
+	[[ ! -x "$wt_binary" ]] && {
+		echo "$name: wt program \"$wt_binary\" not found"
+		exit 1
+	}
```
sueloverso: Not sure why all this additional whitespace indenting was added here?

test/format/format.sh
sueloverso: `waiting for debugger` is actually WiredTiger output, not `format` output. It is a WiredTiger library build option, not specific to `format`.

test/format/format.sh
```diff
@@ -195,30 +197,85 @@ config=$(find_file "$config")
...
+			continue
+		}
+
+		# If the job is still running, ignore it unless we're forcibly quitting.
+		kill -s 0 $pid > /dev/null 2>&1 && {
+			[[ $force_quit -eq 0 ]] && {
+				running=$((running + 1))
+				continue
+			}
+			kill -s TERM $pid
```
sueloverso: If we're forcibly quitting why not use SIGKILL?

test/format/format.sh
```diff
@@ -238,32 +295,59 @@ resolve()
...
+			signame="SIGSYS";;
+		esac
+		[[ ! -z $signame ]] && {
+			(echo
+			 echo "$name: job in $dir killed with signal $signame"
+			 echo "$name: there may be a core dump associated with this failure"
+			 echo) >> $log
+
+			echo "$name: job in $dir killed with signal $signame"
+			echo "$name: there may be a core dump associated with this failure"
```
sueloverso: Could you look in `$dir` for `*core*` and tell the user the core dump file name?

test/format/format.sh
keithbostic: I reworked the whitespace indentation in this block to match the rest of the file. I don't think we have any shell script standard formatting, so not sure if it's "right" or not.

sueloverso: That's fine with me. Just wasn't sure if there was something more subtle going on that I missed.

test/format/format.sh
keithbostic: Thanks for clarifying that, @sueloverso.

test/format/format.sh
keithbostic: I guess I could use `SIGKILL`, but I don't see any advantage to doing so, and I think it makes `format.sh` a bit more complicated (there's another case to handle). Let me know if I'm missing something?

test/format/format.sh
keithbostic: No, because Linux let's you name the core dump anything you want.

PR #5004, opened 2019-11-30
WT-5263 format incorrectly attempting to repeat reads before transaction resolution

keithbostic:
@agorrod, @bvpvamsikrishna, I think this may be a test bug.

After a transaction has resolved, `format` won't repeat reads unless the read timestamp selected is at least as old as the oldest resolved commit timestamp in any thread. However, if the transaction has not yet been resolved, `format` will attempt to repeat reads where the read timestamp selected was simply the next available timestamp. I've been looking at the code, and I don't think that's safe, it seems to me it doesn't matter whether or not the transaction has been resolved, the read will or will not be repeatable based on the timestamp chosen. Is that correct?

### DIFF
--- a/test/format/ops.c
+++ b/test/format/ops.c
@@ -523,6 +523,8 @@ prepare_transaction(TINFO *tinfo)
     testutil_check(__wt_snprintf(buf, sizeof(buf), "prepare_timestamp=%" PRIx64, ts));
     ret = session->prepare_transaction(session, buf);
 
+    logop(session, "prepare ts=%" PRIu64, ts);
+
     testutil_check(pthread_rwlock_unlock(&g.ts_lock));
 
     return (ret);
--- a/test/format/snap.c
+++ b/test/format/snap.c
@@ -403,6 +403,17 @@ snap_repeat_txn(WT_CURSOR *cursor, TINFO *tinfo)
         if (current->opid != tinfo->opid)
             break;
 
+        /*
+         * First, reads may simply not be repeatable because the read timestamp chosen wasn't older
+         * than all concurrently running uncommitted updates.
+         */
+        if (!tinfo->repeatable_reads && current->op == READ)
+            continue;
+
+        /*
+         * Second, as we have not yet resolved the transaction, the rules are as if the transaction
+         * committed successfully.
+         */
         if (snap_repeat_ok_commit(tinfo, current))
             WT_RET(snap_verify(cursor, tinfo, current));
     }


bvpvamsikrishna:
@keithbostic, my understanding is same as yours. Whether reads are repeatable or not depends on the timestamp chosen, rather than on whether the transaction is resolved or not.

sueloverso:
@bvpvamsikrishna and @agorrod what are the semantics of snapshot isolation reads? Our documentation states that snapshot isolation allows for repeatable reads and that timestamp usage is in parallel with normal transaction semantics. So if I start a snapshot isolation transaction, and set read timestamp of T. Can I then, within that same transaction, do `read X; read X` and expect repeatable reads? T did not change. Our snapshot isolation txn id did not change. The only thing that might happen is that some other transaction in the system at some timestamp may have committed in between those two reads. But shouldn't we be isolated from the actions of other transactions with a snapshot isolation transaction?

bvpvamsikrishna:
> @bvpvamsikrishna and @agorrod what are the semantics of snapshot isolation reads? Our documentation states that snapshot isolation allows for repeatable reads and that timestamp usage is in parallel with normal transaction semantics. So if I start a snapshot isolation transaction, and set read timestamp of T. Can I then, within that same transaction, do `read X; read X` and expect repeatable reads? T did not change. Our snapshot isolation txn id did not change. The only thing that might happen is that some other transaction in the system at some timestamp may have committed in between those two reads. But shouldn't we be isolated from the actions of other transactions with a snapshot isolation transaction?

@sueloverso, In the above scenario, we should be isolated from the actions of other transactions. Let us say: 
1. "txn_id1" is the transaction started by us with the read timestamp (T10) 
2. "txn_id2" is the other transaction which will change 'X', with the commit timestamp (T5).
3. (X, Y) is a KV pair on which both the transactions will operate and with the commit timestamp as T1.

If we consider sequence of operations as follows:
1. txn_id1(read timestamp = T10) read X --> will get value Y.
2. txn_id2(commit timestamp = T5) updates (X, Y, T1) -> (X,Y,T5)
3. txn_id2 commits.
4. txn_id1(read timestamp = T10) read X.

In the above scenario, txn_id1 would have got its snapshot before or during step 1. So, as per that snapshot the txn_id2 is not committed. Hence in step 4 when txn_id1 does a read X, it will skp the update (X,Y,T5) as it is considered as uncommitted update and reads (X,Y, T1) as the committed updated. So, there should be no problem for the repeatable read performed by txn_id1.

keithbostic:
Concensus is there is a bug, it's not a problem in format, discarding this PR.

PR #5005, opened 2019-12-02
WT-5729 Disable failing tests in the preferred way

tetsuo-cpp:
As part of BUILD-9262, I've been investigating ways to keep CI green while developing a long-running feature branch.

For Python tests, we want to skip failing tests with the `@unitest.skip(...)` annotation. For `make check` tests, it's a bit less convenient. We need to comment out the appropriate line in the test's `Makefile.am` and then its related jobs in the `evergreen.yml`.

I've used these methods to skip known failures in the durable history branch so that we can work off a green build and detect regressions properly.


tetsuo-cpp:
@quchenhao Thanks! @sulabhM also had the same thought so I'm targeting `relevant_history_in_cache_migration` instead.

sulabhM:
Thanks @tetsuo-cpp , I took the liberty to merge these changes. It could be handy for the change I am working on.

PR #5006, opened 2019-12-02
WT-5165 Added debug mode for making a freeable copy of cursor data

ddanderson:


ddanderson:
retest this please

ddanderson:
@keithbostic, @sueloverso debug cursor is finally ready for review.  Generally, the cursor memory is allocated in getvalue, getkey (thanks for the tip, Keith), but needs to be freed at various places in the API.

src/conn/conn_api.c
```diff
@@ -1790,6 +1790,12 @@ __wt_debug_mode_config(WT_SESSION_IMPL *session, const char *cfg[])
     else
         WT_RET(__wt_calloc_def(session, conn->debug_ckpt_cnt, &conn->debug_ckpt));
 
+    WT_RET(__wt_config_gets(session, cfg, "debug_mode.cursor_copy", &cval));
+    if (cval.val)
+        F_SET(conn, WT_CONN_DEBUG_COPY_CURSOR_DATA);
+    else
+        F_CLR(conn, WT_CONN_DEBUG_COPY_CURSOR_DATA);
```
sueloverso: I have a preference for this name to be `WT_CONN_DEBUG_CURSOR_COPY` to keep the words in the same order. But won't fight you if you like this naming better.

src/include/cursor.i
```diff
@@ -17,6 +17,64 @@ __cursor_set_recno(WT_CURSOR_BTREE *cbt, uint64_t v)
...
+            __wt_explicit_overwrite(cursor->value.mem, cursor->value.memsize);
+            __wt_free(session, cursor->value.mem);
+            cursor->value.data = cursor->value.mem = copy_addr;
+        } else {
+            __wt_explicit_overwrite(cursor->value.mem, cursor->value.memsize);
+            __wt_buf_free(session, &cursor->value);
+        }
+    }
+    return (0);
+}
```
sueloverso: I took a while to eyeball these to make sure they were in fact identical. Could we refactor them into a single `cursor_free_item` and have `cursor_free_key` and `cursor_free_value` just be wrappers and pass in the WT_ITEM pointer for the key or value as needed?

src/btree/bt_curnext.c
```diff
@@ -606,7 +606,8 @@ __wt_btcur_next(WT_CURSOR_BTREE *cbt, bool truncating)
     if (truncating)
         LF_SET(WT_READ_TRUNCATE);
 
-    F_CLR(cursor, WT_CURSTD_KEY_SET | WT_CURSTD_VALUE_SET);
+    WT_ERR(__cursor_unset_key(session, cursor));
+    WT_ERR(__cursor_unset_value(session, cursor));
```
keithbostic: The `__wt_btcur_next_random()` function should probably be included in this change.

src/conn/conn_api.c
ddanderson: Sure, changed the naming.

src/include/cursor.i
ddanderson: Good idea, I just pushed the refactored version.

src/btree/bt_cursor.c
```diff
@@ -638,6 +640,7 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
      * they currently have pinned.)
      */
     WT_ERR(__cursor_localkey(cursor));
+    __cursor_free_value(session, cursor);
```
keithbostic: I would like to use different names than `__cursor_unset_xxx` and `__cursor_free_xxx`, because we're not only "unsetting" or "freeing" things, we're switching to a different chunk of memory, so I think it's confusing.

src/include/cursor.i
```diff
@@ -17,6 +17,55 @@ __cursor_set_recno(WT_CURSOR_BTREE *cbt, uint64_t v)
...
+            WT_RET(__wt_memdup(session, item->mem, item->memsize, &copy_addr));
+            __wt_explicit_overwrite(item->mem, item->memsize);
+            __wt_free(session, item->mem);
+            item->data = item->mem = copy_addr;
+        } else {
+            __wt_explicit_overwrite(item->mem, item->memsize);
+            __wt_buf_free(session, item);
+        }
+    }
+    return (0);
```
keithbostic: It's not pretty and it's slower, but I think we'll be safer if we use the buffer functions. For example, you're not handling the possibility that `item.data != item.mem` (which can happen when reading overflow records, for example), and nobody is going to want to debug that at some future date.

How much do you hate something like this?
```
{
    WT_DECL_item(tmp);
    WT_DECL_RET;
   
    if (!F_ISSET(S2C(session), WT_CONN_DEBUG_CURSOR_COPY))  
        return;

    if (WT_DATA_IN_ITEM(item)) {
        WT_ERR(__wt_scr_alloc(session, 0, &tmp));
        WT_ERR(__wt_buf_set(session, tmp, item->data, item->size));
        __wt_overwrite_and_free_len(session, item->mem, item->memsize);
        __wt_buf_free(session, item);
        WT_ERR(__wt_buf_set(session, item, tmp->data, tmp->size));
    } else {
        __wt_overwrite_and_free_len(session, item->mem, item->memsize);
        __wt_buf_free(session, item);
    }
err:
    __wt_scr_free(session, &tmp);
    return (ret);
} 
```

sueloverso:
Thanks! This looks ready to me now.

test/csuite/wt5165_cursor_copy_debug/main.c
```diff
@@ -0,0 +1,253 @@
...
+/*
+ * We expect that the memory was freed at some point by the WiredTiger API. However, that same
+ * memory may have been reused for another purpose. The best we can do is verify that either the
+ * memory is marked free by malloc, or if not, that it has been overwritten with some other value.
+ */
+#define ASSERT_FREE(mem, prev_mem)                       \
+    do {                                                 \
+        testutil_assert(strcmp(mem, prev_mem) != 0);     \
+        printf(" expect junk: %s\n", (const char *)mem); \
+    } while (0)
```
keithbostic: This is potentially going to drop core if the memory is no longer part of the application's address space.

sueloverso:
This still LGTM.

ddanderson:
@sueloverso and @keithbostic, I've addressed most of the issues here, but it's not quite ready to merge.  I tried turning the flag on globally and running the entire test suite, but I see a fault when calling truncate.  I haven't gotten the cycles to chase yet.  I don't expect the fix to be big, but I don't really know. So please hold off on review for the moment. Thanks!

src/include/cursor.i
```diff
@@ -17,6 +17,69 @@ __cursor_set_recno(WT_CURSOR_BTREE *cbt, uint64_t v)
...
+
+/*
+ * __cursor_copy_release_value --
+ *     Release memory used by the value in cursor copy debug mode.
+ */
+static inline int
+__cursor_copy_release_value(WT_SESSION_IMPL *session, WT_CURSOR *cursor)
+{
+    return (__cursor_copy_release_item(session, &cursor->value));
+}
```
keithbostic: These two functions could just take a `WT_CURSOR` and we can get the `WT_SESSION` from it (that's what most of the other cursor functions do, I think, and it saves you work in `__cursor_localkey()` and `__cursor_localvalue()`.

src/btree/bt_curnext.c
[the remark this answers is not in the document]
ddanderson: Yes, I moved the code from btree to `cur_file.c`.

ddanderson: Yes, included `next_random`.

src/include/cursor.i
ddanderson: I did go this route.  But I noticed that as soon as the memory used by the key/data is freed, it is used again.  I ended up adding yet another allocation so this doesn't happen (at least on this version of malloc or others that do some sort of LIFO allocation for similar sizes).  I added a comment to explain.

test/csuite/wt5165_cursor_copy_debug/main.c
ddanderson: Understood.  There's no great way to check if a memory address is valid before accessing.  Well, there's a clunky way that I used (use the address in a system call and see if `EACCESS` comes back).

src/include/cursor.i
ddanderson: Fixed.

dist/api_data.py
```diff
@@ -495,6 +495,13 @@ def __ge__(self, other):
             adjust log archiving to retain the log records of this number
             of checkpoints. Zero or one means perform normal archiving.''',
             min='0', max='1024'),
+        Config('cursor_copy', 'false', r'''
+            if true, use the system allocator to make a copy of any data
+            returned by a cursor operation and return the copy instead.
+            The copy is freed on the next cursor operation. This allows
+            memory sanitizers to detect inappropriate references to memory
+            owned by cursor cursors.''',
+            type='boolean'),
```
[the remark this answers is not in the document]
ddanderson: Fixed.

sueloverso:
@ddanderson I can understand that the test is difficult to get it work across all platforms. What about, at a minimum, adding a test of the API into `test/suite/test_debug_modeXX` to at least check turning the option on and off.

ddanderson:
@sueloverso, I like that idea and I added a test.

ddanderson:
@keithbostic, when you get a chance can you take another look at this?

src/cursor/cur_file.c
```diff
@@ -190,6 +202,7 @@ __curfile_search(WT_CURSOR *cursor)
     cbt = (WT_CURSOR_BTREE *)cursor;
     CURSOR_API_CALL(cursor, session, search, cbt->btree);
     WT_ERR(__cursor_checkkey(cursor));
+    WT_ERR(__cursor_copy_release_value(cursor));
```
keithbostic: I'm probably missing something, but can't the key be set here? A previous call that returned a key, followed by a search, wouldn't we get here and need to call `__cursor_copy_release_key()`?

If I'm correct, that change would apply to other places, of course.

src/cursor/cur_std.c
```diff
@@ -227,6 +227,49 @@ __wt_cursor_kv_not_set(WT_CURSOR *cursor, bool key) WT_GCC_FUNC_ATTRIBUTE((cold)
...
+     * If we own the memory for the item, make a copy and use that instead. That allows us to
+     * overwrite and free the original memory, potentially uncovering programming errors related to
+     * retaining pointers to key/value memory beyond API boundaries. If the item has memory not
+     * currently in use, free that as well.
+     */
+    if (WT_DATA_IN_ITEM(item)) {
+        /*
+         * We need an extra allocation and a free of it at the right time. Otherwise, with some
+         * allocators, we may get the exact same memory address back in the item. and we will never
+         * catch any accesses to the original memory. Ideally, We want the memory originally used by
```
keithbostic: "We" -> "we"

src/cursor/cur_std.c
```diff
@@ -227,6 +227,49 @@ __wt_cursor_kv_not_set(WT_CURSOR *cursor, bool key) WT_GCC_FUNC_ATTRIBUTE((cold)
...
+        WT_ERR(__wt_scr_alloc(session, 0, &tmp));
+        WT_ERR(__wt_buf_set(session, tmp, item->data, item->size));
+        __wt_explicit_overwrite(item->mem, item->memsize);
+        __wt_buf_free(session, item);
+        __wt_free(session, extra_mem);
+        WT_ERR(__wt_buf_set(session, item, tmp->data, tmp->size));
+    } else if (item->mem != NULL) {
+        __wt_explicit_overwrite(item->mem, item->memsize);
+        __wt_buf_free(session, item);
+    }
```
keithbostic: OK, that took a minute... 🥇 

I worry about gaming the system (for all we know, there's an allocator out there that does FIFO, and with 2 elements on the queue, we'll get the one we don't want).

Do the sanitizers actually not catch access with an intervening free/malloc pair? It would seem to me they **must**, otherwise they'd be useless? Did you actually see this in practice in a sanitizer run?

Anyway, much as it pains me to say it, I'd suggest creating a buffer swap function and explicitly swap the newly allocated `WT_ITEM` for the `WT_ITEM` we don't want.

src/cursor/cur_std.c
```diff
@@ -346,6 +389,10 @@ __wt_cursor_get_keyv(WT_CURSOR *cursor, uint32_t flags, va_list ap)
     if (!F_ISSET(cursor, WT_CURSTD_KEY_SET))
         WT_ERR(__wt_cursor_kv_not_set(cursor, true));
 
+    /* Force an allocated copy when using cursor copy debug. */
+    if (F_ISSET(S2C(session), WT_CONN_DEBUG_CURSOR_COPY))
+        WT_ERR(__wt_buf_grow(session, &cursor->key, cursor->key.size));
```
keithbostic: I don't understand why this forces the buffer to grow, unless we know at this point that `cursor` doesn't point to already-allocated memory.  Why is that known to be the case?

ddanderson: I'm not sure about your question.  We don't care about the buffer growing, only that the buffer `data` pointer puts to memory malloced by the buffer.  We call `__wt_buf_grow` because it has that side effect.  Put another way, if the `cursor` points to already-allocated memory, then that's okay (somebody `malloced` it in the past, so it will be tracked) and `__wt_buf_grow` will do nothing.  If `cursor` doesn't point to already-allocated memory, then `__wt_buf_grow` will make sure it does.

src/cursor/cur_file.c
ddanderson: You are correct.  I was thinking about some simple use cases and missed this.  I have fixed this.

src/cursor/cur_std.c
ddanderson: > Do the sanitizers actually not catch access with an intervening free/malloc pair? 
They do catch it.  I was thinking (hoping) to have the option have some good effect even when not using a sanitizer.  I'm realizing it leads to something that is too fragile.  So I think I'll take out the extra allocation, and make the code simpler.  And then it really has to be used with a sanitizer.

ddanderson: I'm looking into the buffer swap function.

ddanderson: I'm not really loving the buffer swap function idea.  If the buffer swap function does all this stuff, including freeing the old memory, then it will need to do overwrite the old memory (which needs to be done whether or not it is a diagnostic build).  That makes the buffer swap function kind of special case.  It's going to be either ugly here or ugly in `include/buf.i` .  I'll put it in if you still like it, but it's really only going to have a specific use.

ddanderson:
@keithbostic, this is ready for another look.  I'm not sure I resolved everything, let me know what you think.  At least it's converging :-) . Sorry I've taken so long on it.

src/cursor/cur_std.c
sueloverso: FYI, the sanitizer use case is the one MongoDB is looking for. So "it really has to be used with a sanitizer" is fine.

keithbostic: I agree with @sueloverso, this feature is intended for sanitizers, I'm happy to go with that focus.

src/cursor/cur_std.c
ddanderson: By the time `__wt_cursor_get_valuev` (or `__wt_cursor_get_keyv`) is called, the data for the value is already in the WT_CURSOR->value.  The `get_valuevv` function just assigns the caller's pointers into point into the block of memory that is the total set of values.  So any new memory needs to allocated and data copied into place before that's done.

src/cursor/cur_file.c
ddanderson: I did that, adding a pair to all the methods.  Since they are all pairs, I refactored, made them a single method.  In doing this, I found an error in the release method:  there's one case where the original `WT_ITEM` was being cleared (by a call to `__wt_buf_free`).  I went with the cleanest way to fix it was (I think), but it's potentially more wasteful.  If the item points to something external, it always gets a local copy made.  I don't think we're worried about waste here, but I'd appreciate it if you took another look at `__wt_cursor_copy_release_item`

ddanderson:
@keithbostic, I hope you don't mind taking another look at this.  In particular I changed `__wt_cursor_copy_release_item`: it was wrong before in the case of an external item, it cleared the item entirely after freeing the memory.  This function now treats internal and external items the same.  Thanks!

PR #5008, opened 2019-12-03
WT-5295 Disable remaining failing tests in durable history branch

tetsuo-cpp:
This PR is disabling `schema_abort` since we have a regression in the durable history feature branch.


tetsuo-cpp:
retest this please

PR #5009, opened 2019-12-03
WT-5042 Avoid calling config functions during checkpoint. (#4936)

quchenhao:
This is to backport WT-5042 and WT-5239.

When doing this, I found the syscall test on mongodb is broken and created a separate ticket WT-5297.

Since I cannot verify the change using syscall test, I manually verified that the change generates the expected change in stderr.txt outputted by syscall test. So I think the backport is good to go.


luke-pearson:
This LGTM but given the difference with `__wt_meta_block_metadata` I think @sueloverso should take a look before merging.

PR #5012, opened 2019-12-03
WT-5286 Enable data format to store time pair

kommiharibabu:
Change the default value of the WT build to store the time pair
data in the WT_CELL structure whenever a record is reconciled
into a disk image.


kommiharibabu:
@keithbostic, Yes we can remove the option entirely. The option can be completely removed once everything is working with data format changes.

kommiharibabu:
Created a new ticket WT-5308 to remove the entire configuration option and its code.

PR #5013, opened 2019-12-03
WT-5264 Add delete record for every insert into history store

kommiharibabu:
Every record that gets inserted into a history store must have a stop time pair.
To represent the stop time pair to automatically propagated to the cell structure,
add a delete record for every inserted record with a stop time pair as the commit time pair.

During reconciliation of a history page, these two records must be reconciled into
a single record with that have both start and stop time pairs.

History store records are special as they store the history of an actual record.
we need to retain the transaction and timestamp of those records as it was done by
the original transaction and timestamp, so make sure that the updates don't get
modified or validated for the history store transaction id during commit/rollback.


kommiharibabu:
This change leads to none of the lookaside pages gets evicted as the records in lookaside pages now have time pairs that lead to not to evict them. It can get fixed once the write the newest version is available. It shouldn't merge until writing the newest version change is available.

@agorrod and @bvpvamsikrishna, please review the changes.

src/cache/cache_las.c
```diff
@@ -793,13 +802,36 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
+            cache->las_txnid = upd->txnid;
+
             /*
              * Using update instead of insert so the page stays pinned and can be searched before
              * the tree.
              */
             WT_ERR(cursor->update(cursor));
             ++insert_cnt;
 
+            /* Append a delete record to represent stop time pair for the above insert record */
```
quchenhao: Hi Hari, the change is related to the writing newest committed version to disk I'm working on and I'm confused by your change. I don't think this is correct. The commit timestamp will become smaller and smaller because we iterate from the newest update to the oldest update and I think __wt_txn_set_commit_timestamp only allows commit timestamp to be set if it's larger or equal to the first_commit_timestamp by looking at the code. Also Though I read the PR description, I still don't understand why we have to add a delete pair. If we only write committed updates to the history store, we should have enough information to determine the stop time pair when we insert it. Probably you can explain this to me tomorrow.

quchenhao: Hi Hari, I now understand what you are doing after the discussion with Vamsi.

src/include/txn.i
```diff
@@ -375,7 +375,12 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
             op->type = WT_TXN_OP_BASIC_COL;
     }
     op->u.op_upd = upd;
-    upd->txnid = session->txn.id;
+
+    /* Use the original txnid for the las inserts */
+    if (S2C(session)->cache->las_fileid != 0 && op->btree->id == S2C(session)->cache->las_fileid)
+        upd->txnid = S2C(session)->cache->las_txnid;
```
agorrod: Given that the transaction code needs to special case inserts into the lookaside file regardless, I'm not convinced we should also use the trick of setting snapshot isolation in the lookaside cursor transaction.

OTOH: I don't see why we need to pull this transaction ID from somewhere special - maybe you can explain that to me face to face?

src/cache/cache_las.c
```diff
@@ -611,8 +611,17 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
     /* Wrap all the updates in a transaction. */
     WT_ERR(__wt_txn_begin(session, NULL));
     __las_set_isolation(session, &saved_isolation);
+
     local_txn = true;
 
+    /*
+     * Set the oldest timestamp as the first commit timestamp for the LAS transaction. None of the
+     * transactions with commit timestamp less than oldest timestamp will go into history store. AS
+     * they will be cleaned as part of the obsolete check.
```
bvpvamsikrishna: The assumption "None of the transactions with commit timestamp less than oldest timestamp will go into history store" is wrong. The version visible at oldest timestamp will go to history store, the version before that one can be discarded.

src/cache/cache_las.c
```diff
@@ -22,7 +22,7 @@ static void
 __las_set_isolation(WT_SESSION_IMPL *session, WT_TXN_ISOLATION *saved_isolationp)
 {
     *saved_isolationp = session->txn.isolation;
-    session->txn.isolation = WT_ISO_READ_UNCOMMITTED;
+    session->txn.isolation = WT_ISO_SNAPSHOT;
```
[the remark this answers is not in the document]
kommiharibabu: Reverted to original isolation level.  Cache variables are used to pass the information to the updates of lookaside with original timestamp and transaction id.

src/cache/cache_las.c
```diff
@@ -611,8 +611,17 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
     /* Wrap all the updates in a transaction. */
     WT_ERR(__wt_txn_begin(session, NULL));
     __las_set_isolation(session, &saved_isolation);
+
     local_txn = true;
 
+    /*
+     * Set the oldest timestamp as the first commit timestamp for the LAS transaction. None of the
```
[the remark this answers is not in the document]
kommiharibabu: Removed the code to set the commit timestamp.

src/cache/cache_las.c
kommiharibabu: The code was not removing all the versions older than the oldest and leaving one visible record. The comment was wrong. Removed it.

src/include/txn.i
```diff
@@ -375,7 +375,12 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
             op->type = WT_TXN_OP_BASIC_COL;
     }
     op->u.op_upd = upd;
-    upd->txnid = session->txn.id;
+
+    /* Use the original txnid for the las inserts */
```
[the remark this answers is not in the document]
kommiharibabu: fixed.

src/include/txn.i
```diff
@@ -375,7 +375,12 @@ __wt_txn_modify(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
+    if (S2C(session)->cache->las_fileid != 0 && op->btree->id == S2C(session)->cache->las_fileid)
```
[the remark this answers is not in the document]
kommiharibabu: Added a new macro to check whether it is a lookaside operation or not?

src/include/txn.i
kommiharibabu: As discussed, the transaction id is required to maintain the information of the original transaction. Removed the snapshot isolation change and kept the cache variable to pass the information.

src/cache/cache_las.c
```diff
@@ -572,6 +583,80 @@ __las_squash_modifies(WT_SESSION_IMPL *session, WT_CURSOR *cursor, WT_UPDATE **u
...
+        /*
+         * It is never OK to see a birthmark here - it would be referring to the wrong page image.
+         */
+        WT_RET(__wt_illegal_value(session, upd->type));
+    }
+
+    cursor->set_key(cursor, btree_id, key, upd->start_ts, upd->txnid, *stop_ts, *stop_txnid);
+
+    /*
+     * Set the current update start time pair as the commit time pair to the lookaside record.
```
agorrod: It's usual to keep the comment on a single line if possible - and I think you can here.

src/cache/cache_las.c
```diff
@@ -779,34 +844,9 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
 
-            /*
-             * If we encounter a modify with preceding updates that have the same timestamp and
-             * transaction id, they need to be squashed into a single lookaside entry to avoid
-             * displacing each other.
-             */
-            if (upd->type == WT_UPDATE_MODIFY && upd->next != NULL &&
-              upd->start_ts == upd->next->start_ts && upd->txnid == upd->next->txnid) {
-                WT_ERR(__las_squash_modifies(session, cursor, &upd));
-                ++insert_cnt;
```
agorrod: Is the insert counter going to be different in the new code layout?

src/include/cache.h
```diff
@@ -290,6 +294,11 @@ struct __wt_cache_pool {
     uint8_t flags;
 };
 
+/*
+ * Optimize comparisons against the lookaside URI, flag handles that reference the lookaside file.
+ */
+#define WT_IS_BTREE_LOOKASIDE(session) F_ISSET(S2BT((session)), WT_BTREE_LOOKASIDE)
```
agorrod: Sorry to be a pedant, but it's actually the DHANDLE not the BTREE. I'd call this `WT_DHANDLE_IS_LOOKASIDE`

src/cache/cache_las.c
```diff
@@ -35,6 +35,17 @@ __las_restore_isolation(WT_SESSION_IMPL *session, WT_TXN_ISOLATION saved_isolati
...
 }
 
+/*
+ * __las_store_time_pair --
+ *     Store the time pair to use for the lookaside inserts.
+ */
+static void
+__las_store_time_pair(WT_SESSION_IMPL *session, wt_timestamp_t timestamp, uint64_t txnid)
+{
+    S2C(session)->cache->org_timestamp_to_las = timestamp;
```
agorrod: The common abbreviation for `original` is `orig` - would you mind making that change?

src/cache/cache_las.c
kommiharibabu: I don't see a need to this counter along with other counters such as `las_remove_count` and their stats along with `txn_prepared_updates_lookaside_inserts`. Earlier these counters are used to identify how many las records are present, but now as we are moving into a different way of cleanup, there is no way to identify how many records are present. I prefer to remove them entirely. Comments?

src/cache/cache_las.c
kommiharibabu: corrected.

src/cache/cache_las.c

src/include/cache.h
kommiharibabu: Corrected.

src/cache/cache_las.c
```diff
@@ -35,6 +35,17 @@ __las_restore_isolation(WT_SESSION_IMPL *session, WT_TXN_ISOLATION saved_isolati
...
 }
 
+/*
+ * __las_store_time_pair --
+ *     Store the time pair to use for the lookaside inserts.
+ */
+static void
+__las_store_time_pair(WT_SESSION_IMPL *session, wt_timestamp_t timestamp, uint64_t txnid)
+{
+    S2C(session)->cache->orig_timestamp_to_las = timestamp;
```
bvpvamsikrishna: Why store the time pair info at the global level? Why not at session-level. What is the need of having time pair info at the global level and make it a critical section?

src/cache/cache_las.c
```diff
@@ -562,7 +573,7 @@ __las_squash_modifies(WT_SESSION_IMPL *session, WT_CURSOR *cursor, WT_UPDATE **u
     }
     cursor->set_value(
       cursor, start_upd->durable_ts, start_upd->prepare_state, WT_UPDATE_STANDARD, &las_value);
-    WT_ERR(cursor->insert(cursor));
+    WT_ERR(cursor->update(cursor));
```
bvpvamsikrishna: Please add a comment, as to why we need to use "update" instead of "insert".

src/include/cache.h
```diff
@@ -199,6 +199,10 @@ struct __wt_cache {
     uint64_t las_insert_count; /* Count of inserts to lookaside */
     uint64_t las_remove_count; /* Count of removes from lookaside */
 
+    /* Original transaction time pair to use for the lookaside inserts */
+    uint64_t orig_txnid_to_las;
```
bvpvamsikrishna: My preference to store this information is at session-level unless we have a compelling reason to make this global.

kommiharibabu: Moved to `WT_SESSION_IMPL` structure.

src/cache/cache_las.c
kommiharibabu: As the original las insert was done with update to let the page to pinned. I followed the same for the newly added code for squash modifies. I added a comment explaining the same in the code.

src/cache/cache_las.c
kommiharibabu: Moved to `WT_SESSION_IMPL` structure.

src/txn/txn.c
```diff
@@ -1019,10 +1019,8 @@ __wt_txn_commit(WT_SESSION_IMPL *session, const char *cfg[])
                 /*
                  * Writes to the lookaside file can be evicted as soon as they commit.
                  */
-                if (conn->cache->las_fileid != 0 && fileid == conn->cache->las_fileid) {
-                    upd->txnid = WT_TXN_NONE;
+                if (conn->cache->las_fileid != 0 && fileid == conn->cache->las_fileid)
```
bvpvamsikrishna: Enhance the comment, to say that we use different timestamps for different entries and we don't want to use __wt_txn_op_set_timestamp which will overwrite the timestamps of all the operations.

kommiharibabu: Updated comment to explain the reason.

src/cache/cache_las.c
agorrod: Removing them entirely makes sense if we don't need them any more. I'd rather do that than subtly change what they mean.

kommiharibabu: Currently, there is a usage for the `insert_cnt` variable in the las_sweep. Created a new ticket to take care of removing the `insert_cnt` once the las_sweep code is merged.

PR #5014, opened 2019-12-04
WT-5047 Always return WT_TRY_SALVAGE for corrupted log

hbokhari:


test/suite/test_txn19.py
```diff
@@ -304,8 +304,7 @@ def test_corrupt_log(self):
 
         if expect_fail:
             self.check_file_contains_one_of(errfile,
```
sulabhM: Maybe `self.check_file_contains_one_of(errfile, 'WT_TRY_SALVAGE: database corruption detected')` suits better

PR #5015, opened 2019-12-04
WT-5276 format needs to move the oldest timestamp forward during bulk load

keithbostic:


test/format/util.c
```diff
@@ -572,25 +572,45 @@ checkpoint(void *arg)
...
-
     testutil_check(__wt_snprintf(buf, sizeof(buf), "%s", "oldest_timestamp="));
 
+    /*
+     * Lock out transaction timestamp operations. The lock acts as a barrier ensuring we've checked
+     * if the workers have finished, we don't want that line reordered.
+     */
+    testutil_check(pthread_rwlock_wrlock(&g.ts_lock));
+
+    ret = conn->query_timestamp(conn, buf + strlen("oldest_timestamp="), "get=all_committed");
```
tetsuo-cpp: Should change `all_committed => all_durable`. They alias to the same timestamp but `all_committed` is deprecated and will be removed in the near future (whenever someone gets around to it).
Also it might be nice to give this string a name as I git a bit confused here. Maybe:
```
static const char *oldest_timestamp_str = "oldest_timestamp=";
```

test/format/util.c
```diff
@@ -572,25 +572,45 @@ checkpoint(void *arg)
...
+
+/*
+ * timestamp --
+ *     Periodically update the oldest timestamp.
+ */
+WT_THREAD_RET
+timestamp(void *arg)
+{
+    bool done;
+
```
tetsuo-cpp: Nit: `WT_UNUSED`.

test/format/util.c
keithbostic: Pushed in a792a4e38.

test/format/util.c
keithbostic: `format` doesn't use `WT_UNUSED` anywhere -- I can make a pass if you feel strongly.

tetsuo-cpp: No worries. It was just a nitpick. It's better if it's consistent with the rest of `format`.

PR #5016, opened 2019-12-04
WT-5222 Write the newest version when checkpointing

luke-pearson:
The idea behind this change is to get checkpoint to always write the newest committed version by running it without a read timestamp. Obviously this breaks most tests that checkpoint and then expect the version at the stable timestamp to be the one they see after restart.

As per the specification, use_timestamps=true is still used, we report the checkpoint timestamp as the stable timestamp when use_timestamps=true. In order to avoid checkpointing at the same stable timestamp being inconsistent I've included the change to skip checkpointing if we checkpoint at the same stable timestamp.

I think there may have been an attempt to skip checkpoints using the same logic, however it may have caused too many issues and was reverted? @keithbostic do you know anything about this?


keithbostic:
@luke-pearson 
> I think there may have been an attempt to skip checkpoints using the same logic, however it may have caused too many issues and was reverted? @keithbostic do you know anything about this?

No, I do not. @agorrod?

test/suite/test_backup08.py
```diff
@@ -125,6 +125,7 @@ def backup_and_recover(self, expected_rec_ts):
         self.pr("query recovery ts: " + q)
         self.assertTimestampsEqual(q, timestamp_str(expected_rec_ts))
 
+    @unittest.skip("Temporarily disabled")
```
keithbostic: These tests aren't going to be possible in the future, are they? Don't they have to eventually go away?

agorrod:
> I think there may have been an attempt to skip checkpoints using the same logic

I think you mean WT-4959 - or it's precursors to be more precise. I'm not sure if it's related - let me know if you'd like me to check.

test/suite/test_backup08.py
luke-pearson: The majority of them should be possible in the future, with rollback to stable implemented and perhaps with some small adjustments to the test themselves. So i'm hesitant to delete them entirely just yet.

src/txn/txn_ckpt.c
```diff
@@ -699,11 +684,15 @@ __txn_checkpoint_can_skip(
...
+     * If the checkpoint is using timestamps, and the stable timestamp hasn't been updated since the
+     * last checkpoint there is nothing more that could be written.
      */
-    if (use_timestamp)
+    if (use_timestamp && txn_global->has_stable_timestamp &&
+      txn_global->last_ckpt_timestamp != WT_TS_NONE &&
+      txn_global->last_ckpt_timestamp == txn_global->stable_timestamp) {
+        *can_skipp = true;
         return (0);
+    }
```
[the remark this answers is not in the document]
luke-pearson: It is if you open a new connection, don't set a stable and still call checkpoint with use_timestamps=true.

src/txn/txn_ckpt.c
```diff
@@ -699,11 +684,17 @@ __txn_checkpoint_can_skip(
...
     /*
-     * It isn't currently safe to skip timestamp checkpoints - see WT-4958. We should fix this so we
-     * can skip timestamp checkpoints if they don't have new content.
+     * If the checkpoint is using timestamps, and the stable timestamp hasn't been updated since the
+     * last checkpoint there is nothing more that could be written. Except when a non timestamped
+     * file has been modified, as such if the connection has been modified it is currently unsafe to
+     * skip checkpoints.
      */
-    if (use_timestamp)
+    if (!conn->modified && use_timestamp && txn_global->has_stable_timestamp &&
```
tetsuo-cpp: Isn't this entire check redundant? Immediately below you have:
```
    if (!conn->modified) {
        *can_skipp = true;
        return (0);
    }
```

src/txn/txn_ckpt.c
```diff
@@ -604,11 +604,9 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, bool *trackingp, const char *cfg[
          * because recovery doesn't set the recovery timestamp until its checkpoint is complete.
          */
         if (txn_global->has_stable_timestamp) {
-            txn->read_timestamp = txn_global->stable_timestamp;
```
tetsuo-cpp: Just confirming my understanding, is this the main change? Since we leave the read timestamp unset, we see the latest of each key?

luke-pearson: Yeah, this is the main change. We'll always see the latest committed update without a read timestamp.

src/txn/txn_ckpt.c
```diff
@@ -604,11 +604,9 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, bool *trackingp, const char *cfg[
...
-            txn_global->checkpoint_timestamp = txn->read_timestamp;
-            F_SET(txn, WT_TXN_HAS_TS_READ);
+            txn_global->checkpoint_timestamp = txn_global->stable_timestamp;
             if (!F_ISSET(conn, WT_CONN_RECOVERING))
-                txn_global->meta_ckpt_timestamp = txn->read_timestamp;
+                txn_global->meta_ckpt_timestamp = txn_global->stable_timestamp;
```
bvpvamsikrishna: could we use txn_global->checkpoint_timestamp instead of txn_global->stable_timestamp ?

src/txn/txn_ckpt.c
luke-pearson: Thanks, fixed.

src/txn/txn_ckpt.c
```diff
@@ -619,19 +617,6 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, bool *trackingp, const char *cfg[
 
     __wt_writeunlock(session, &txn_global->rwlock);
 
-    if (F_ISSET(txn, WT_TXN_HAS_TS_READ)) {
-        __wt_verbose_timestamp(
```
bvpvamsikrishna: We still need to have this verbose msg.

src/txn/txn_ckpt.c
luke-pearson: It won't change anything functionally as we're holding the txn_global->rwlock but I've updated it.

src/txn/txn_ckpt.c
luke-pearson: Added it back, it will still log if the stable is 0 which to me seems like correct behaviour.

src/txn/txn_ckpt.c
```diff
@@ -604,13 +604,14 @@ __checkpoint_prepare(WT_SESSION_IMPL *session, bool *trackingp, const char *cfg[
...
-            txn_global->checkpoint_timestamp = txn->read_timestamp;
-            F_SET(txn, WT_TXN_HAS_TS_READ);
+            txn_global->checkpoint_timestamp = txn_global->stable_timestamp;
             if (!F_ISSET(conn, WT_CONN_RECOVERING))
-                txn_global->meta_ckpt_timestamp = txn->read_timestamp;
+                txn_global->meta_ckpt_timestamp = txn_global->checkpoint_timestamp;
         } else if (!F_ISSET(conn, WT_CONN_RECOVERING))
             txn_global->meta_ckpt_timestamp = txn_global->recovery_timestamp;
+
+        __wt_verbose_timestamp(
```
bvpvamsikrishna: verbose statements better be outside critical section i.e. after "unlock". I think you could use checkpoint_timestamp instead of stable_timestamp.

PR #5018, opened 2019-12-05
WT-5277 Cursor key out-of-order detected in the lookaside file

keithbostic:
@sueloverso, @tetsuo-cpp, I misread the code when I made the changes for WT-5120.

The row-search code doesn't abandon the leaf-page search in cases where it can't be sure the search is safe, it continues with the search and relies on its caller to use the `WT_CURSOR_BTREE.compare` value to decide if a match was found. That's fine for searches, but inserts quite likely don't have an exact match, and so we have to ignore any leaf page results where we don't get an exact match and we're at the edge of the page, as the item might belong on a different page entirely.

I have not re-run the WT-5120 tests to confirm this fix still works, but it seems like it would have minimal impact.

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -647,14 +647,15 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
      * WT_CURSOR.search, ignore pinned pages in the case of column-store, search-near isn't an
      * interesting enough case for column-store to add the complexity needed to avoid the tree
      * search.
-     *
-     * Set the "insert" flag for the btree row-store search; we may intend to position the cursor at
-     * the end of the tree, rather than match an existing record.
      */
     valid = false;
     if (btree->type == BTREE_ROW && __cursor_page_pinned(cbt, true)) {
         __wt_txn_cursor_op(session);
 
+        /*
+         * Set the "insert" flag for row-store search; we may intend to position the cursor at the
+         * the end of the tree, rather than match an existing record. (LSM requires this semantic.)
+         */
         WT_ERR(__cursor_row_search(cbt, true, cbt->ref, &leaf_found));
 
         /*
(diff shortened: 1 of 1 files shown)


sueloverso:
I am assuming that since this is locked and hasn't passed PR testing that it isn't ready for review yet. If that is wrong, let me know.

keithbostic:
@sueloverso, @tetsuo-cpp: I've updated the WT-5277 PR with 5f1d3de, which fixes a bug where we don't start a necessary implicit transaction in `WT_SESSION.truncate` in the case of start/stop range cursors that successfully searched pinned pages, and makes sure we don't introduce the same bug in `WT_SESSION.update` should we ever use pinned page searches in updates for anything other lookaside file updates.

To summarize:

* the change to add a test for `(cbt->slot != 0 && cbt->slot != cbt->ref->page->entries - 1)` to `__btcur_update()` is the fix for WT-5277,
* the change to add a test for `cbt->compare == 0` to `__wt_btcur_search_near()` and `__btcur_update()` is an EXTENSION to the existing semantics of pinned page searches, relaxing the requirement that we never use the first/last slots of the page,
* the change to add a `__wt_txn_autocommit_check()` call to `__wt_btcur_range_truncate()` fixes the truncate bug described above,
* the change to add a `__wt_txn_autocommit_check()` call to `__btcur_update()` is the same fix for updates.

src/btree/bt_cursor.c
```diff
@@ -1231,10 +1232,19 @@ __btcur_update(WT_CURSOR_BTREE *cbt, WT_ITEM *value, u_int modify_type)
...
         __wt_txn_cursor_op(session);
+        WT_ERR(__wt_txn_autocommit_check(session));
 
         WT_ERR(btree->type == BTREE_ROW ? __cursor_row_search(cbt, true, cbt->ref, &leaf_found) :
                                           __cursor_col_search(cbt, cbt->ref, &leaf_found));
-        if (leaf_found)
+        /*
+         * If searching an already pinned page doesn't find an exact match and returns the first or
+         * last page slots, discard the results and search the full tree, a non-existent record
+         * might belong on an entirely different page. This test is simplistic as we're ignoring
```
sueloverso: I find both of these (identical) comments confusing. It is essentially opposite of what the code directly below it is actually doing. They talk about not finding an exact match and discarding the results, searching the full tree, etc, but the code below is actually testing for an exact match and returning those results.

keithbostic: Reworked in a21f9e0e7.

PR #5022, opened 2019-12-06
WT-5302 Begin writing validity window for new data format

tetsuo-cpp:
After we begin writing the new data format, we need to begin populating the on-disk start and stop time pairs. There was some basic code to write the start time pair so I've added some extra logic to handle tombstone updates (which means we should choose the previous value for reconciliation and set the stop time pair).

I don't expect this change to work yet as it requires the latest versions to be written in eviction and checkpoint. But I'd like to get it reviewed ahead of time to make sure this change is on the right track.


src/reconcile/rec_visibility.c
```diff
@@ -586,35 +586,59 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+             */
+            upd_select->durable_ts = upd_select->start_ts = vpack->start_ts;
+            upd_select->start_txn = vpack->start_txn;
+        }
 
         /*
          * Finalize the timestamps and transactions, checking if the update is globally visible and
          * nothing needs to be written.
          */
-        if ((upd_select->stop_ts == WT_TS_MAX && upd_select->stop_txn == WT_TXN_MAX) &&
```
tetsuo-cpp: This part of the check looked redundant to me so I just removed it.

src/reconcile/rec_visibility.c
```diff
@@ -586,35 +587,70 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+             * we'll have an empty update list. And applying a delete on top of that will result in
+             * ONLY a tombstone in the update list.
+             *
+             * In this case, we should leave the selected update unset to indicate that we want to
+             * keep the same on-disk value but set the stop time pair to indicate that the validity
+             * window ends when this tombstone started.
+             */
+            upd_select->durable_ts = upd_select->start_ts = vpack->start_ts;
+            upd_select->start_txn = vpack->start_txn;
+            /*
```
tetsuo-cpp: I considered making this change at a lower level where we're writing to the cell however, it doesn't seem that easy to just skip writing the value and edit only the timestamps. We can think about doing that as an optimization at a later point however, I don't think it matters much because this is not the regular case.

src/reconcile/rec_visibility.c
```diff
@@ -586,35 +587,70 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        WT_ASSERT(session, upd == NULL || upd_select->upd->type != WT_UPDATE_TOMBSTONE);
 
         /*
          * Finalize the timestamps and transactions, checking if the update is globally visible and
          * nothing needs to be written.
          */
-        if ((upd_select->stop_ts == WT_TS_MAX && upd_select->stop_txn == WT_TXN_MAX) &&
-          ((upd_select->start_ts == WT_TS_NONE && upd_select->start_txn == WT_TXN_NONE) ||
-              __wt_txn_visible_all(session, upd_select->start_txn, upd_select->start_ts))) {
+        if (__wt_txn_visible_all(session, upd_select->start_txn, upd_select->start_ts)) {
```
kommiharibabu: I feel this part of checking and assigning can be completely removed. Already at the beginning the start and stop time pair is assigned properly. Irrespective of the visibility keeping start and stop time pair is better. Otherwise, already set the stop time pair may be reset here.

src/reconcile/rec_visibility.c
kommiharibabu: Looks fine to me.

src/reconcile/rec_visibility.c
```diff
@@ -586,35 +587,70 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+            /*
+             * Leaving the update unset means that we can skip reconciling. If we've set the stop
+             * time pair because of a tombstone after the on-disk value, we still have work to do so
+             * that is NOT ok. Let's allocate an update equivalent to the on-disk value and continue
+             * on our way!
+             */
+            WT_RET(__wt_scr_alloc(session, 0, &tmp));
+            WT_ERR(__wt_page_cell_data_ref(session, page, vpack, tmp));
+            WT_ERR(__wt_update_alloc(session, tmp, &upd_select->upd, &size, WT_UPDATE_STANDARD));
+            upd_select->upd->ext = 1;
```
kommiharibabu: Do all the WT_UPDATE are allocated with external memory contents? Do we need this special flag?

src/reconcile/rec_visibility.c
tetsuo-cpp: 62b594a

I agree with what you're saying. There is no value in fudging the start and stop time pair.

src/reconcile/rec_visibility.c
tetsuo-cpp: The `ext` flag is to signal to callers of `__wt_rec_upd_select` that they are responsible for freeing the update. At the moment, this can happen if reconciliation chooses a lookaside update (rather than one from the in-memory update list which you would NOT want to free). In this case, we're allocating an update specifically for reconciliation purposes so we'll need to free it when we're done with it.

src/reconcile/rec_visibility.c
```diff
@@ -586,40 +587,64 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        WT_ASSERT(session, upd_select->upd != NULL);
+        if (upd_select->upd->type == WT_UPDATE_TOMBSTONE) {
+            upd_select->stop_ts = upd_select->upd->start_ts;
+            upd_select->stop_txn = upd_select->upd->txnid;
+            upd_select->upd = upd_select->upd->next;
+        }
+        if (upd_select->upd != NULL) {
+            /* The beginning of the validity window is the selected update's time pair. */
+            upd_select->durable_ts = upd_select->start_ts = upd_select->upd->start_ts;
+            upd_select->start_txn = upd_select->upd->txnid;
```
keithbostic: In these two clauses, can't we use `upd` instead of `upd_select->upd`? (If you're doing it deliberately, I get it, no objection.)

src/reconcile/rec_visibility.c
```diff
@@ -586,40 +587,64 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        } else {
+            /*
+             * The only time we don't get the unpacked value is when we insert a key in which case
+             * we can't have a tombstone.
+             */
+            WT_ASSERT(session, vpack != NULL);
```
keithbostic: I don't think that comment is true (see `__wt_rec_col_fix` for a counter-example), although I agree with the intent. Can you rephrase?

src/reconcile/rec_visibility.c
```diff
@@ -586,40 +587,64 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+            /*
+             * Leaving the update unset means that we can skip reconciling. If we've set the stop
+             * time pair because of a tombstone after the on-disk value, we still have work to do so
+             * that is NOT ok. Let's allocate an update equivalent to the on-disk value and continue
+             * on our way!
+             */
+            WT_RET(__wt_scr_alloc(session, 0, &tmp));
+            WT_ERR(__wt_page_cell_data_ref(session, page, vpack, tmp));
+            WT_ERR(__wt_update_alloc(session, tmp, &upd_select->upd, &size, WT_UPDATE_STANDARD));
+            upd_select->upd->ext = 1;
```
keithbostic: We're allocating 2 temporary buffers now (`key` and `tmp`), and I don't think the usage overlaps, perhaps we don't need both?

I'm also pretty sure `las_key` and `las_value` are misused -- there's an assumption that calling the LAS cursor for a key/value pair won't allocate space in the WT_ITEM, and I don't see any reason to believe that's guaranteed.

src/reconcile/rec_visibility.c
```diff
@@ -586,40 +587,64 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+             */
+            upd_select->durable_ts = upd_select->start_ts = vpack->start_ts;
+            upd_select->start_txn = vpack->start_txn;
+            /*
+             * Leaving the update unset means that we can skip reconciling. If we've set the stop
+             * time pair because of a tombstone after the on-disk value, we still have work to do so
+             * that is NOT ok. Let's allocate an update equivalent to the on-disk value and continue
+             * on our way!
+             */
+            WT_RET(__wt_scr_alloc(session, 0, &tmp));
```
keithbostic: This can't be `WT_RET`, we've already allocated `key`.

Come to think of it, there are `WT_ERR` calls before this line in this function, the dist scripts should have caught this case.

src/reconcile/rec_visibility.c
keithbostic: We need to get rid of `WT_UPDATE.ext`, it's not a good thing. I think @sulabhM has a plan...

src/reconcile/rec_visibility.c
keithbostic: I went through the same thought process when I first pushed the data format changes in -- my first attempt did involve doing this work in the cell code, and I couldn't make it worked.

src/reconcile/rec_visibility.c
tetsuo-cpp: 3ebd6ac

src/reconcile/rec_visibility.c
```diff
@@ -586,40 +587,64 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+         * indicate that the value is visible to any timestamp/transaction id ahead of it.
          */
-        if ((upd_select->stop_ts == WT_TS_MAX && upd_select->stop_txn == WT_TXN_MAX) &&
-          ((upd_select->start_ts == WT_TS_NONE && upd_select->start_txn == WT_TXN_NONE) ||
-              __wt_txn_visible_all(session, upd_select->start_txn, upd_select->start_ts))) {
-            upd_select->start_ts = WT_TS_NONE;
-            upd_select->start_txn = WT_TXN_NONE;
-            upd_select->stop_ts = WT_TS_MAX;
-            upd_select->stop_txn = WT_TXN_MAX;
+        WT_ASSERT(session, upd_select->upd != NULL);
```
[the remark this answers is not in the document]
tetsuo-cpp: 3ebd6ac

src/reconcile/rec_visibility.c
tetsuo-cpp: 3ebd6ac

src/reconcile/rec_visibility.c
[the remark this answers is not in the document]
tetsuo-cpp: 3ebd6ac

src/reconcile/rec_visibility.c
tetsuo-cpp: 3ebd6ac

tetsuo-cpp: 3ebd6ac

src/reconcile/rec_visibility.c
tetsuo-cpp: Yeah, I am using `ext` because it fits in the current view of the world. But it does need to be addressed at some point.

src/reconcile/rec_visibility.c
tetsuo-cpp: Ok, we can use the same buffer.

What do you mean about misusing `las_key` and `las_value`? Are you saying that there is a missing free in the existing code?

src/reconcile/rec_visibility.c
tetsuo-cpp: @quchenhao This is the part where I'm relying on `ext`.

src/reconcile/rec_visibility.c
keithbostic: > Are you saying that there is a missing free in the existing code?

Yes. We clear a structure, then do `las_cursor->get_key/get_value` and there's no guarantee that `WT_CURSOR.get` not allocate memory in the `WT_ITEM`.

On second thought, I think that's true for `las_key` -- I think `las_value` is worse, there are calls to `__wt_buf_set` which, unless I'm missing something, guarantees we'll leak memory.

src/reconcile/rec_visibility.c
keithbostic: Do we know why the dist script didn't catch this?

tetsuo-cpp: It did catch it. I thought it was incorrectly flagging a different line and "fixed" the `dist` script (which I've reverted in the commit).

src/reconcile/rec_visibility.c
tetsuo-cpp: 79b0881

src/reconcile/rec_visibility.c
tetsuo-cpp: Hi @keithbostic. I've had more of a think about this and I'd rather keep it as is.

My reasoning is that @quchenhao is currently working on a change to write the newest versions to disk. If this is the case, we're not going to have lookaside searching in this function which will remove the need for a `key` buffer. Also, the `las_key` and `las_value` will no longer exist.

So none of these issues will exist in the future but fixing them temporarily will make my conflicts harder to resolve because it will further tangle my patch with this version of the function.

Let me know if you don't agree with this or want to talk about it further.

src/reconcile/rec_visibility.c
```diff
@@ -586,40 +587,61 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
-            upd_select->stop_ts = WT_TS_MAX;
-            upd_select->stop_txn = WT_TXN_MAX;
+        if (upd->type == WT_UPDATE_TOMBSTONE) {
+            upd_select->stop_ts = upd->start_ts;
+            upd_select->stop_txn = upd->txnid;
+            upd_select->upd = upd = upd->next;
         }
+        if (upd != NULL) {
+            /* The beginning of the validity window is the selected update's time pair. */
+            upd_select->durable_ts = upd_select->start_ts = upd->start_ts;
```
bvpvamsikrishna: we should use the upd->durable_ts for upd_select->durable_ts, else we will loose that information.
Prepared updates after committed could have a different durable timestamp compared to commit timestamp.

bvpvamsikrishna: As currently we don't have durable timestamp as part of Cell, ignore this comment. We will handle this part later.

tetsuo-cpp:
Thanks for the reviews! I'm going to leave this PR locked until we have merged the work to write the newest versions in eviction and checkpoint.

tetsuo-cpp:
retest this please

tetsuo-cpp:
I will merge this now.

I've had to disable any test that involves reopening a connection or opening a connection on an existing database. The reason is that when we open a connection, any new txn ids begin from 1 irrespective of what is on the disk. So on-disk values that were written in a previous connection may have later timestamps/txn ids than newer updates.

I've added some rudimentary guards to ensure that we don't write invalid cells (stop time pair earlier than start time pair), but we need a solution to this. I've added a `FIXME` to ensure that we address this when we do the recovery portion of this project.

PR #5023, opened 2019-12-06
WT-5005 Migrate Jenkins Wiredtiger test-recovery-stress test

hbokhari:


lukech:
@hbokhari please note the Jenkins job is currently configured to run continuously (triggering itself as a post-action).

lukech:
Another thing is the current Jenkins job runs on both Linux and zSeries. Please make sure zSeries is not overlooked in this PR.

test/evergreen.yml
```diff
@@ -1784,6 +1814,48 @@ tasks:
...
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-strict --enable-diagnostic --with-builtins=snappy,zlib
+      - func: "recovery stress test script"
+        vars:
+          # Repeat this script enough times to make this task run for 24 hours
+          # At the time of writing this script, one cal to underlying scripts takes about 8 mins to finish.
+          times: 180
+
+  - name: split-stress-test
```
hbokhari: This is from merging. Not my changes.

test/evergreen.yml
```diff
@@ -132,32 +132,76 @@ functions:
         set -o errexit
         set -o verbose
         for i in $(seq ${times|1}); do
-          ./t -1 -c ${config|../../../test/format/CONFIG.stress} ${extra_args|}
+          ./t -1 -c ${config|../../../test/format/CONFIG.stress} ${extra_args|} 2>&1
         done
   "many dbs test":
     command: shell.exec
-    parms:
+    params:
```
hbokhari: There are few typos for `parms` -> `params` that I am fixing using this ticket.

hbokhari:
@lukech and @raviprakashgiri29. I am requesting review to get an early feed before the test results are finally in. The testing should take around 24 hours to finish.

test/evergreen.yml
lukech: Good catch. I'm surprised `evergreen validate` did not pick it out.

test/evergreen.yml
```diff
@@ -283,7 +327,7 @@ tasks:
           posix_configure_flags: --enable-silent-rules --enable-strict --enable-diagnostic --disable-static
       - func: "make check all"
         vars:
-          test_env_vars: MSAN_OPTIONS=abort_on_error=1:disable_coredump=0 MSAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
+          test_env_vars: MSAN_OPTIONS=abort_on_error=1:disable_coredump=0 MSAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer TESTUTIL_SLOW_MACHINE=1
```
lukech: Could you share the reason this change is made as part of this PR?

test/evergreen.yml
```diff
@@ -1784,6 +1814,48 @@ tasks:
...
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: --enable-strict --enable-diagnostic --with-builtins=lz4,snappy,zlib
+      - func: "recovery stress test script"
+        vars:
+          # Repeat this script enough times to make this task run for 24 hours
+          # At the time of writing this script, one cal to underlying scripts takes about 8 mins to finish.
```
lukech: `cal` typo?
Is 8 mins the average runtime in Jenkins or Evergreen?

test/evergreen.yml
```diff
@@ -1784,6 +1814,48 @@ tasks:
...
+          times: 180
```
lukech: Should we leave some space in case some runs take longer than expected?

test/evergreen.yml
lukech: It's weird the merged change shows up as part of the PR. Did you use merge or rebase?

test/evergreen.yml
```diff
@@ -1965,3 +2039,5 @@ buildvariants:
   tasks:
   - name: compile
   - name: unit-test
+  - name: recovery-stress-test-without-lz4 # Evergreen Z-series distro doesn't support lz4
```
lukech: Will you be creating another follow-on ticket to replace this task once the zSeries Evergreen distro get lz4 installed?

test/evergreen.yml
```diff
@@ -132,32 +132,76 @@ functions:
...
+
+          # V1 log compatibility mode with write-no-sync txns.
+          ${test_env_vars|} ./test_random_abort -C 2>&1
+          ${test_env_vars|} ./test_timestamp_abort -C 2>&1
+
+          # V1 log compatibility mode with memory-based txns.
+          ${test_env_vars|} ./test_random_abort -C -m 2>&1
+          ${test_env_vars|} ./test_timestamp_abort -C -m 2>&1
+
+          ${test_env_vars|} ./test_truncated_log ${truncated_log_args|} 2>&1
```
lukech: I guess any line of test fails would cause the Evergreen task fails and stops?

test/evergreen.yml
```diff
@@ -132,32 +132,76 @@ functions:
...
+        done
```
lukech: How about adding a sleep before starting another round of testing?

test/evergreen.yml
hbokhari: I used merge but somehow this weird merge issues are showing up. I will probably revert to pre-merge state and try to redo the merge.

test/evergreen.yml
hbokhari: Yes I was surprised as well and I only picked it once I realized that some of sub-tasks were returning within few microseconds.

test/evergreen.yml
hbokhari: Ah again some weird merge issue. Will fix all of them together.

test/evergreen.yml
hbokhari: Will fix the typo `cal` -> `call`.

The average runtime was from Evergreen patch build. I am running a full 24 hours test right now and that will give me better estimate. Will update this number once I have more data.

test/evergreen.yml
hbokhari: I am running a full test that I am expecting to take around <= 24 hours. I will adjust this number once that job finished to make room for any variance in execution time.

test/evergreen.yml
hbokhari: Good idea. Will add a short sleep after all script calls.

test/evergreen.yml
hbokhari: I am not sure on this one. Should we wait for zSeries distro fix or should we go ahead like this and create a Jira ticket in to remove this special function once distro fix is in place?

test/evergreen.yml
hbokhari: With `set -o errexit` statement in script, the task should fail if any of the commands shall fail.

lukech: That means the new task won't be running through 24hrs if any line of test fails in any round?

test/evergreen.yml
lukech: Will leave to you to handle it

test/evergreen.yml
hbokhari: That's correct. The only way out would be to use something like `format.sh` script that can manage spawning job that can potentially fail.

lukech: Got it

PR #5024, opened 2019-12-06
WT-5037 Remove wrong assert of las transaction rollback

kommiharibabu:
Remove the assert that checks that no lookaside operations
can rollback. It is possible that las transaction can rollback.
it is a rare scenario, but better to avoid it.


tetsuo-cpp:
Why is this possible? In `develop`, lookaside writes are never timestamped so I'd have thought that they are always "stable" by definition.

kommiharibabu:
All the las records are inserted in a single transaction. In case of failure that happened during the insertion led to a transaction rollback which hit the removed assert. The assert is expecting that as none of the lookaside transactions rollbacks. Yes, it is a rare scenario of lookaside transaction fails, but if it fails it shouldn't crash on an assert.

PR #5025, opened 2019-12-06
WT-5283 Write newest version to disk for eviction and checkpoint

quchenhao:
As discussed with Vamsi, we plan to write new functions for the eviction module and then replace the old module with the new module when we are ready for testing.

This is the function in the new eviction module for selecting the version to write for both eviction and checkpoint. Since we are overhauling the whole module, I think it's better to simply this function and move as much logic as possible outside this function. I'm not sure if it's still necessary for us to maintain the list of pointers and counters in this function. If we can remove some of them, it would be great. Also I keep the time pair logic and save update logic in this function for now but we may also move them outside the function later.

### DIFF
--- a/src/reconcile/rec_visibility.c
+++ b/src/reconcile/rec_visibility.c
@@ -843,16 +843,6 @@ __wt_rec_upd_select_new(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *in
             if (F_ISSET(r, WT_REC_EVICT))
                 ++r->updates_unstable;
 
-            /*
-             * Rare case: when applications run at low isolation levels, update/restore eviction may
-             * see a stable update followed by an uncommitted update. Give up in that case: we need
-             * to discard updates from the stable update and older for correctness and we can't
-             * discard an uncommitted update.
-             */
-            if (F_ISSET(r, WT_REC_UPDATE_RESTORE) && upd_select->upd != NULL &&
-              (list_prepared || list_uncommitted))
-                return (__wt_set_return(session, EBUSY));
-
             continue;
         }
 
--- a/src/reconcile/rec_visibility.c
+++ b/src/reconcile/rec_visibility.c
@@ -758,3 +758,253 @@ err:
 
     return (ret);
 }
+
+/*
+ * __wt_rec_upd_select_new --
+ *     Return the update in a list that should be written for both eviction and checkpoint (or NULL
+ *     if none can be written).
+ */
+int
+__wt_rec_upd_select_new(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, void *ripcip,
+  WT_CELL_UNPACK *vpack, WT_UPDATE_SELECT *upd_select)
+{
+    WT_PAGE *page;
+    WT_UPDATE *first_upd, *upd;
+    wt_timestamp_t max_ts;
+    size_t upd_memsize;
+    uint64_t max_txn, txnid;
+    bool all_stable, list_prepared, list_uncommitted;
+
+    /*
+     * The "saved updates" return value is used independently of returning an update we can write,
+     * both must be initialized.
+     */
+    upd_select->upd = NULL;
+    upd_select->upd_saved = false;
+
+    page = r->page;
+    upd = NULL;
+    upd_memsize = 0;
+    max_ts = WT_TS_NONE;
+    max_txn = WT_TXN_NONE;
+    list_prepared = list_uncommitted = false;
+
+    /*
+     * If called with a WT_INSERT item, use its WT_UPDATE list (which must
+     * exist), otherwise check for an on-page row-store WT_UPDATE list
+     * (which may not exist). Return immediately if the item has no updates.
+     */
+    if (ins != NULL)
+        first_upd = ins->upd;
+    else if ((first_upd = WT_ROW_UPDATE(page, ripcip)) == NULL)
+        return (0);
+
+    for (upd = first_upd; upd != NULL; upd = upd->next) {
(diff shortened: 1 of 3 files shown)


src/reconcile/rec_visibility.c
```diff
@@ -758,3 +758,253 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+             * discard an uncommitted update.
+             */
+            if (F_ISSET(r, WT_REC_UPDATE_RESTORE) && upd_select->upd != NULL &&
+              (list_prepared || list_uncommitted))
+                return (__wt_set_return(session, EBUSY));
+
+            continue;
+        }
+
+        if (!F_ISSET(r, WT_REC_EVICT))
```
quchenhao: If we don't need to track all the counters, we can remove this check as well and just break.

src/reconcile/rec_visibility.c
```diff
@@ -758,3 +758,253 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+    WT_ASSERT(session, !WT_IS_METADATA(session->dhandle) || upd == NULL ||
+        upd->txnid == WT_TXN_NONE || upd->txnid != S2C(session)->txn_global.checkpoint_state.id ||
+        WT_SESSION_IS_CHECKPOINT(session));
+
+    /* If all of the updates were aborted, quit. */
+    if (upd == NULL) {
+        return (0);
+    }
+
+    if (upd != NULL && upd->durable_ts > r->max_ondisk_ts)
```
quchenhao: Not sure if we still need this.

src/reconcile/rec_visibility.c
```diff
@@ -758,3 +758,253 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        r->max_ondisk_ts = upd->durable_ts;
+
+    /*
+     * TIMESTAMP-FIXME The start timestamp is determined by the commit timestamp when the key is
+     * first inserted (or last updated). The end timestamp is set when a key/value pair becomes
+     * invalid, either because of a remove or a modify/update operation on the same key.
+     */
+    if (upd != NULL) {
```
quchenhao: Probably better to move this part outside this function. I think it is enough for this function just to return the selected update. Thoughts? The caller of the function should figure out the start and stop time pairs.

src/reconcile/rec_visibility.c
```diff
@@ -758,3 +758,253 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+        r->max_txn = max_txn;
+
+    /* Update the maximum timestamp. */
+    if (max_ts > r->max_ts)
+        r->max_ts = max_ts;
+
+    /*
+     * If the update we chose was a birthmark.
+     */
+    if (upd != NULL && upd->type == WT_UPDATE_BIRTHMARK) {
```
quchenhao: Is birthmark still relevant? I think we can remove this or remove this later?

src/reconcile/rec_visibility.c
```diff
@@ -758,3 +758,253 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+     *
+     * Both paths are configured outside of reconciliation: the save/restore path is the
+     * WT_REC_UPDATE_RESTORE flag, the lookaside table path is the WT_REC_LOOKASIDE flag.
+     */
+    if (!F_ISSET(r, WT_REC_LOOKASIDE | WT_REC_UPDATE_RESTORE))
+        return (__wt_set_return(session, EBUSY));
+    if (list_uncommitted && !F_ISSET(r, WT_REC_UPDATE_RESTORE))
+        return (__wt_set_return(session, EBUSY));
+
+check_original_value:
```
quchenhao: May be a better name. Here we should save updates for both eviction and checkpoint

src/reconcile/rec_visibility.c
```diff
@@ -758,3 +758,253 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+            if (F_ISSET(r, WT_REC_EVICT))
+                ++r->updates_unstable;
+
+            /*
+             * Rare case: when applications run at low isolation levels, update/restore eviction may
+             * see a stable update followed by an uncommitted update. Give up in that case: we need
+             * to discard updates from the stable update and older for correctness and we can't
+             * discard an uncommitted update.
+             */
+            if (F_ISSET(r, WT_REC_UPDATE_RESTORE) && upd_select->upd != NULL &&
```
quchenhao: According to Vamsi, this can happen when the same transaction updates the same key multiple times and we try to evict this page in the middle of transaction commit.

No need to worry about this case in durable history I think as we have selected the newer version to write to disk, which should be the correct value visible after the transaction commit. The uncommitted value will be squashed and discarded when we insert the following updates to history store.

I'm not sure why this only happens in update/restore eviction according to the comment and is there any other edge case that can result to this situation.

quchenhao:
I fixed the memory corruption issue so may be we don't need write a new function for this change.

src/reconcile/rec_visibility.c
tetsuo-cpp: I'd prefer it if you hold off from refactoring this part. My change to write to the on-disk validity window touches this part of `__wt_rec_upd_select` so any further changes here are going to make it difficult for me.

quchenhao: Understood. I'll keep it as it is for now.

PR #5026, opened 2019-12-06
WT-5289 Remove LAS garbage collection sweep code

ddanderson:
As far as I can see the sweep code for LAS only performs one function, removal of out-of-date entries.  Consequently, this change removes the entire LAS sweep.


src/include/cache.h
```diff
@@ -199,18 +197,6 @@ struct __wt_cache {
...
-    WT_ITEM las_sweep_key;            /* Track sweep position. */
-    uint32_t las_sweep_dropmin;       /* Minimum btree ID in current set. */
-    uint8_t *las_sweep_dropmap;       /* Bitmap of dropped btree IDs. */
-    uint32_t las_sweep_dropmax;       /* Maximum btree ID in current set. */
-    uint32_t las_max_btree_id;        /* Maximum btree ID for sweep. */
-    WT_ITEM las_max_key;              /* Maximum key for sweep. */
-    wt_timestamp_t las_max_timestamp; /* Maximum timestamp for sweep. */
-    uint64_t las_max_txnid;           /* Maximum txn ID for sweep. */
-
     uint32_t *las_dropped;    /* List of dropped btree IDs. */
```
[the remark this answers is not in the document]
ddanderson: Thanks, @kommiharibabu, I've removed references to the dropped btrees.  Please take another look.  One question, I just want to confirm that I would be merging into pm-1521-durable-history when this is approved?

src/include/cache.h
```diff
@@ -197,10 +197,6 @@ struct __wt_cache {
     uint64_t las_insert_count; /* Count of inserts to lookaside */
     uint64_t las_remove_count; /* Count of removes from lookaside */
```
kommiharibabu: Apologies for not mentioning it in the earlier comment. I think `las_remove_count` is not used anywhere now. it may be better to remove it?

src/include/cache.h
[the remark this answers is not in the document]
kommiharibabu: Thanks for the changes @ddanderson. Yes, this change will go into `pm-1521-durable-history` branch. That is the feature branch for `durable history` project.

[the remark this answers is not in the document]
sueloverso: Just curious why that branch is named `pm-1521-...`? That is the first and only branch ever to be named for a PM ticket and the merge will go into the Product Management ticket instead of a WiredTiger ticket. That means later searching in JIRA of the WiredTiger ticket space will not find it. Is this a new process I'm not familiar with?

src/btree/bt_handle.c
```diff
@@ -251,12 +245,11 @@ __wt_btree_close(WT_SESSION_IMPL *session)
     F_SET(btree, WT_BTREE_CLOSED);
 
     /*
-     * If closing a tree let sweep drop lookaside entries for it.
+     * Verify the lookaside state.
```
sueloverso: A bit more detail about what a valid lookaside state means would be helpful here. The conditional below is fairly complicated.

src/conn/conn_sweep.c
```diff
@@ -266,14 +266,13 @@ __sweep_server(void *arg)
     WT_DECL_RET;
     WT_SESSION_IMPL *session;
     uint64_t last, now;
-    uint64_t last_las_sweep_id, min_sleep, oldest_id, sweep_interval;
+    uint64_t min_sleep, sweep_interval;
```
sueloverso: We don't need `min_sleep` anymore. We can just use `conn->sweep_interval` in the code.

src/include/cache.h
sueloverso: I think it is. It currently is used in `wt_las_remove_block` which is not part of this change AFAICT.

src/include/cache.h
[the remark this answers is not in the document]
kommiharibabu: `pm-1521-durable-history` is the feature branch of durable history. All the code changes that are done for the durable history, `pm-1521-...` is the final branch. Once all the tests pass, this branch will get merged into `develop`. It is just the branch name that is named with a PM ticket. All the work done with WT ticket changes still represents inside their tickets itself. There is no change in the process, it is just the final branch that the code gets merged is different.

src/include/cache.h
kommiharibabu: The function `wt_las_remove_block` is already removed as part of the code changes from `relevant_history` project. Now whenever a page gets instantiated, the lookaside contents are not brought back into memory and they stay in the lookaside.

src/include/cache.h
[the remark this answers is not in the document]
sueloverso: It is not just a branch name. There are github hooks that push all changes/commits into the ticket that is the prefix of the branch. So every change that has been merged into that branch, including all sub-branches, will be sent to the Jira ticket PM-1521. That is a change in process because no other PM ticket has all project commits pushed to it. Maybe that's okay and the direction that Product management, the owner of the PM ticket space, wants to go.

You are correct that the individual sub-branch merges should go to their tickets. But it is not "just" a branch name. It will have an effect when merged with develop.

src/include/cache.h
sueloverso: Okay, then it should be removed. I'm wondering why the scripts that check for unused fields didn't flag it. Maybe those check for other unused things but not fields.

ddanderson: Yes, I think they check for unused functions, and maybe unused defines.

src/btree/bt_handle.c
ddanderson: Thanks, I added more detail to this comment.

src/include/cache.h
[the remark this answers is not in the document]
ddanderson: Note - the code changes for this discussion are resolved.  But I'm leaving this open for now, I think the discussion about branch naming has not been resolved.

ddanderson:
@kommiharibabu, the failing PR tests are in `test_txn07.py` in the test suite.  These tests are failing in the `pm-1521-durable-history` branch on which this is based.  Is this known?  Should I go ahead and merge?  All the changes from the review are in this branch.

ddanderson:
Here's one scenario from test_txn07, run on the pm-1521-durable-history branch, synced today.  I just inserted some debugging showing that the stats for `log_compress_len` and `log_compress_mem` are both 0.  That's the same reason for failure in my branch.
```
/opt/mongodbtoolchain/v3/bin/python2.7 ../test/suite/run.py test_txn07.test_txn07.test_ops -s 4
compress=snappy, clen=0, cmem=0
cwrites=0, cfails=0, csmall=2
FERROR in test_txn07.test_txn07.test_ops -s 4 (row.trunc-start.t1c.snappy)

======================================================================
FAIL: test_txn07.test_txn07.test_ops -s 4 (row.trunc-start.t1c.snappy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/mnt/data0/dda/work/wt/git/newx/test/suite/test_txn07.py", line 230, in test_ops
    self.assertEqual(clen < cmem, True)
AssertionError: False != True

----------------------------------------------------------------------
Ran 1 test in 2.804s

FAILED (failures=1)
```

sueloverso:
@ddanderson is that branch actually logging anything? The test purposefully chooses a very large value so that logging will compress it.
```            
           # Choose large compressible values for the string cases.
            value = 'abc' * 1000000
```
The fact you're seeing both as 0 seems to imply that logging isn't happening.

sueloverso:
@ddanderson turns out WT-5314 is the same thing too. You might look at that and/or chat with chen hao.

kommiharibabu:
@ddanderson, Once the conflicts are resolved, you can merge this ticket.

ddanderson:
@sueloverso, I don't think the error I'm seeing is not related to WT-5314.  That one gives a ERROR message in the specific case that compression is off.  This one fails when compression is on.  But `pm-1521-durable-history` is failing in the same way, so I think we need to address that in that branch.

PR #5031, opened 2019-12-08
WT-5309 update format script add prefix command

raviprakashgiri29:
Added a new argument(-b) to the format wrapper script to pass the fomat prefix with the binary 

For example : 
a. In some cases the format binary is run with nice command
    nice ./format.sh <other_arguments>
b. In some cases the format binary is run with nice & checksegv command
    nice checksegv ./format.sh <other_arguments>

So, with the new argument (-b) user can pass prefix with the binary.

For ex : 
format.sh -b "nice ./t"
format.sh -b "checksegv ./t"
format.sh -b "nice checksegv ./t"
format.sh 

NOTE - By default the format binary it is set to "./t" if -b is not specified

### DIFF
--- a/test/format/format.sh
+++ b/test/format/format.sh
@@ -18,7 +18,7 @@ trap 'onintr' 2
 
 usage() {
 	echo "usage: $0 [-aFSv] [-c config] "
-	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [-p prefix-command][format-configuration]"
+	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-p prefix-command] [-t minutes] [format-configuration]"
 	echo
 	echo "    -a           abort/recovery testing (defaults to off)"
 	echo "    -c config    format configuration file (defaults to CONFIG.stress)"
--- a/test/format/format.sh
+++ b/test/format/format.sh
@@ -21,8 +21,8 @@ usage() {
 	echo "    [-b format-binary-cmd] [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [format-configuration]"
 	echo
 	echo "    -a           abort/recovery testing (defaults to off)"
-	echo "    -c config    format configuration file (defaults to CONFIG.stress)"
 	echo "    -b binary    format binary command (defaults to "./t")"
+	echo "    -c config    format configuration file (defaults to CONFIG.stress)"
 	echo "    -F           quit on first failure (defaults to off)"
 	echo "    -h home      run directory (defaults to .)"
 	echo "    -j parallel  jobs to execute in parallel (defaults to 8)"
@@ -82,11 +82,11 @@ while :; do
 	-a)
 		abort_test=1
 		shift ;;
-	-c)
-		config="$2"
-		shift ; shift ;;
 	-b)
 		format_binary_cmd="$2"
+		shift ; shift ;;
+	-c)
+		config="$2"
 		shift ; shift ;;
 	-F)
 		first_failure=1
--- a/test/format/format.sh
+++ b/test/format/format.sh
@@ -18,7 +18,7 @@ trap 'onintr' 2
 
 usage() {
 	echo "usage: $0 [-aFSv] [-c config] "
-	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [format-configuration]"
+	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [-p prefix-command][format-configuration]"
 	echo
 	echo "    -a           abort/recovery testing (defaults to off)"
 	echo "    -c config    format configuration file (defaults to CONFIG.stress)"
@@ -29,6 +29,7 @@ usage() {
 	echo "    -S           run smoke-test configurations (defaults to off)"
 	echo "    -t minutes   minutes to run (defaults to no limit)"
 	echo "    -v           verbose output (defaults to off)"
+	echo "    -p           prefix command to prepend before executable (defaults to empty)"
 	echo "    --           separates $name arguments from format arguments"
 
 	exit 1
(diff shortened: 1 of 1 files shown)


raviprakashgiri29:
> Curiousity, what's the reason for this?

Hi Keith,

This change is to pass the pre-settings to prepend before binary.

For example :
a. In some cases the format binary is run with nice command
nice ./format.sh <other_arguments>
b. In some cases the format binary is run with nice & checksegv command
nice checksegv ./format.sh <other_arguments>

So, with the new argument (-p) user can pass the pre-settings to prepend before the binary.

Thanks.

test/format/format.sh
```diff
@@ -18,14 +18,15 @@ trap 'onintr' 2
...
-	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [format-configuration]"
+	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-p pre-setting] [-t minutes] [format-configuration]"
 	echo
 	echo "    -a           abort/recovery testing (defaults to off)"
 	echo "    -c config    format configuration file (defaults to CONFIG.stress)"
 	echo "    -F           quit on first failure (defaults to off)"
 	echo "    -h home      run directory (defaults to .)"
 	echo "    -j parallel  jobs to execute in parallel (defaults to 8)"
 	echo "    -n total     total jobs to execute (defaults to no limit)"
+	echo "    -p           pre settings to prepend before binary (defaults to empty)"
```
keithbostic: "pre setting" doesn't really mean anything to me, I'd suggest something like "format-prefix"?

To be clear, I don't feel strongly, if you & @lukech prefer this, that's fbm.

test/format/format.sh
```diff
@@ -368,7 +373,7 @@ format()
 		echo "$name: starting job in $dir"
 	fi
 
-	cmd="$format_binary -c "$config" -h "$dir" -1 $args quiet=1"
+	cmd="$pre_setting $format_binary -c "$config" -h "$dir" -1 $args quiet=1"
```
keithbostic: An alternative approach might be "-b format-binary", which defaults to "./t" and can be anything you want, so:
```
format.sh -b "nice checksegv ./t"
```
would replace the format command with something else.

lukech: I prefer this option.

test/format/format.sh
raviprakashgiri29: I agree. Updated the script to use "format-prefix"

test/format/format.sh
raviprakashgiri29: I also prefer this option. But I realised that "format_binary" is considered as an executable file in the script and using -b option to prefix something before binary will impact other exisiting validation checks. 

For example : 
If format_binary = "nice ./t", then the existing logic of checking of binary [-x $format_binary] would fail and changing directory to "$build_posix_directory" needs to changed.

I can still change the existing validation checks & to use -b option if @keithbostic & @lukech suggests. :-)

keithbostic: I honestly don't feel strongly -- this makes it easier to replace the name of the format command and it occurred to me @lukech might find that useful, but it's at the cost of finding the last element of whatever is specified to the option.

lukech: @raviprakashgiri29 in that case just choose whatever option that's easy to implement e.g. `format-prefix`.

test/format/format.sh
```diff
@@ -18,10 +18,11 @@ trap 'onintr' 2
 
 usage() {
 	echo "usage: $0 [-aFSv] [-c config] "
-	echo "    [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [format-configuration]"
+	echo "    [-b format-binary-cmd] [-h home] [-j parallel-jobs] [-n total-jobs] [-t minutes] [format-configuration]"
 	echo
 	echo "    -a           abort/recovery testing (defaults to off)"
 	echo "    -c config    format configuration file (defaults to CONFIG.stress)"
+	echo "    -b binary    format binary command (defaults to "./t")"
```
lukech: This line becomes obsolete.

lukech: Better to put it one line up to make it alphabetical-order.

test/format/format.sh
```diff
@@ -83,6 +85,9 @@ while :; do
 	-c)
 		config="$2"
 		shift ; shift ;;
+	-b)
+		format_binary_cmd="$2"
+		shift ; shift ;;
```
lukech: Moving it up to make it alphabetic-order.

raviprakashgiri29:
@hbokhari @tammybailey : Please have a look at the format script changes to pass format binary as an option.

test/format/format.sh
```diff
@@ -155,11 +160,13 @@ cd $(dirname $0) || exit 1
...
-# build_posix tree, move there, we have to run in the directory where the format binary lives
-# because the format binary "knows" the wt utility is two directory levels above it.
-format_binary="./t"
+# Find the last part of format_binary_cmd, which is format binary. Builds are normally in the
+# WiredTiger source tree, in which case it's in the same directory as format.sh, else it's in
+# the build_posix tree. If the build is in the build_posix tree, move there, we have to run in
+# the directory where the format binary lives because the format binary "knows" the wt utility
+# is two directory levels above it.
+format_binary=${format_binary_cmd##* }
+
```
keithbostic: Would `${xformat_binary_cmd##*[:isspace:]}` be safer?

test/format/format.sh
```diff
@@ -74,12 +75,16 @@ parallel_jobs=8
 smoke_test=0
 total_jobs=0
 verbose=0
+format_binary_cmd="./t"
```
keithbostic: Don't really need `format_binary_cmd`, just use `format_binary`?

raviprakashgiri29: Yes, I have removed format_binary_cmd and using just format_binary instead. Thanks.

PR #5032, opened 2019-12-09
WT-5000 Adding new test format-stress-sanitizer-test to Evergreen

hbokhari:


test/evergreen.yml
```diff
@@ -1783,6 +1792,31 @@ tasks:
...
+            if [ -f /usr/bin/llvm-symbolizer ]; then
+              ASAN_SYMBOLIZER_PATH="/usr/bin/llvm-symbolizer"
+            else
+              ASAN_SYMBOLIZER_PATH=`ls /usr/bin/llvm-symbolizer* | tail -1`
+            fi
+            export ASAN_SYMBOLIZER_PATH
+      - func: "format test script"
+        vars:
+          # run for 24 hours ( 24 * 60 = 1440 minutes), stop at first error, use default config
+          format_test_script_args: "-F -t 2"
```
hbokhari: Ignore the 2 min argument. This is to quickly test the PR. Final commit will have this set as 1440.

lukech:
@tammybailey This is the 1st PR about the stress job migration. Added you as a reviewer so you are aware of the changes needed.

test/evergreen.yml
```diff
@@ -190,6 +190,15 @@ functions:
...
         ./t ${checkpoint_args} 2>&1
+  
+  "format test script":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/test/format"
+      script: |
+        set -o errexit
+        set -o verbose
+        ${test_env_vars|} ./format.sh ${format_test_script_args|} ${smp_command|} 2>&1
```
lukech: Can this section being moved up next to the existing `"test format"` function? And to keep the naming consistent the older function is better to be called `"format test"` instead :)

test/evergreen.yml
```diff
@@ -1783,6 +1792,31 @@ tasks:
             set -o verbose
 
             ./time_shift_test.sh /usr/local/lib/faketime/libfaketimeMT.so.1 0-1 2>&1
+  
+  - name: format-stress-sanitizer-test
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          posix_configure_flags: CC="clang -fsanitize=address" CFLAGS="-g -fno-omit-frame-pointer" --enable-diagnostic --with-builtins=lz4,snappy,zlib
```
lukech: By convention, only those configure flags e.g. `--enable-diagnostic ----with-builtins=lz4,snappy,zlib` should be set in `posix_configure_flags` variable. Those environment variable settings should go to `configure_env_vars` variable.
Also, consider creating a new function (`compile-builtin`?) as I believe most stress jobs use the same builtin configure setting.

test/evergreen.yml
```diff
@@ -1783,6 +1792,31 @@ tasks:
...
+            set -o errexit
+            set -o verbose
+
+            export ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0"
+            if [ -f /usr/bin/llvm-symbolizer ]; then
+              ASAN_SYMBOLIZER_PATH="/usr/bin/llvm-symbolizer"
+            else
+              ASAN_SYMBOLIZER_PATH=`ls /usr/bin/llvm-symbolizer* | tail -1`
+            fi
+            export ASAN_SYMBOLIZER_PATH
```
lukech: Consider setting those ASAN related variables within `test_env_vars`. There are examples in test/evergreen.yml you may want to have a check.

test/evergreen.yml
lukech: Even for a shorter-interval test, 2 mins is probably too short, because the default timer setting in `test/format/CONFIG.stress` is 4 mins, which means each format job will run ~4 mins (could be more depending on how quick the format test process responds to the stop signal). From your patch build you can see all 4 format jobs got killed prematurely due to the 2 mins format.sh `-t` setting.

lukech: Suggest to create a patch build to run the 24hr-testing sooner than later as the test takes time to finish :)

hbokhari: Good point. I am addressing the comments and will start a evergreen build hopefully before EOD.

test/evergreen.yml
hbokhari: Good idea. I will move and rename functions.

test/evergreen.yml
hbokhari: Yes CC and CFLAGS should go to `configure_env_vars`. Not sure yet on creating a new task yet. Will followup on this idea once we start working on other stress tickets.

test/evergreen.yml
hbokhari: Yes `test_env_vars` should do the trick for the first export. Not sure how to manage `ASAN_SYMBOLIZER_PATH` variable.

test/evergreen.yml
```diff
@@ -1783,6 +1792,31 @@ tasks:
             set -o verbose
 
             ./time_shift_test.sh /usr/local/lib/faketime/libfaketimeMT.so.1 0-1 2>&1
+  
+  - name: format-stress-sanitizer-test
+    commands:
+      - func: "get project"
```
raviprakashgiri29: Can you break the get and compile stuff into a different task ?

hbokhari: Right now there is no other task using this particular compilation combination. Maybe if there are more than one task in future that use this compilation step then we can make a separate task and make this task depend on that.

raviprakashgiri29:
@hbokhari : Can you make the overflow tests into a seperate task as the overflow tests are being done by other stress tests as well. Thanks.

hbokhari:
@raviprakashgiri29 @lukech I am creating a seperate task for overflow testing

test/evergreen.yml
```diff
@@ -97,6 +97,19 @@ functions:
           set -o verbose
           echo '#undef HAVE_FTRUNCATE' >> wiredtiger_config.h
     - *make_wiredtiger
+  "compile wiredtiger sanitizer":
```
raviprakashgiri29: It looks like doing the similar stuff which "compile wiredtiger" does. 
How about : 
Create a task say "compile-wiredtiger-sanitizer" and call the functions 
"get project"  
"compile wiredtiger"
 set your own configure_env_vars and posix_configure_flags

hbokhari: We need to compile with this particular `configure_env_vars` and `posix_configure_flags` combination for other sanitize tests in future as well. That's why I made a special function that can be called by all those tasks.

hbokhari:
I have started a patch build with 24 hour time limit: https://evergreen.mongodb.com/version/5def24acd1fe07065bbd6220

test/evergreen.yml
raviprakashgiri29: I understand, but I prefer to use the reconf and configure options of existing functions instead of writing new. 

Something like the below task "compile-wiredtiger-sanitizer" will do the job and also it can be used in other sanitize tests.

  - name: compile-wiredtiger-sanitizer
    commands:
      - func: "get project"
      - func: "compile wiredtiger"
        vars:
          configure_env_vars: CC="clang -fsanitize=address" CFLAGS="-g -fno-omit-frame-pointer"
          posix_configure_flags: --enable-diagnostic --with-builtins=lz4,snappy,zlib

hbokhari: I actually wrote a special compile task for this and made the sanitizer test to depend on that test. However, the `format.sh` script failed as we have seen before that some tests are unable to use the precompiled directories.

I wrote this special compile function as a workaround.

test/evergreen.yml
```diff
@@ -1783,6 +1804,34 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/test/format"
+          shell: bash
+          script: |
+            set -o errexit
+            set -o verbose
+            # to emulate the original Jenkins job's test coverage, we are running the smoke test 16 times
+            # run smoke tests, stop at first error, use default config (-F -S)
+            for i in {1..2}; do
```
hbokhari: Ignore this argument now. I am running a seperate patch build to verify the actual arguments.

hbokhari:
Latest patch build for testing full arguments: https://evergreen.mongodb.com/version/5df0437b2a60ed5b1656e38e

test/evergreen.yml
```diff
@@ -97,6 +97,19 @@ functions:
...
+  "compile wiredtiger sanitizer":
+    - command: shell.exec
+      params:
+        working_dir: "wiredtiger/build_posix"
+        shell: bash
+        script: |
+          set -o errexit
+          set -o verbose
+          sh reconf
+          CC="clang -fsanitize=address" CFLAGS="-g -fno-omit-frame-pointer" \
```
raviprakashgiri29: Default CFLAGS values are "-ggdb -fPIC". So, I think the CFLAGS should be set to "-ggdb -fPIC -fno-omit-frame-pointer" here and CC, CXX, PATH environment variable should point to mongodbtoolchains absolute path.

hbokhari: The compilation is unique for this one: http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer/configure . These setting are invoking the address sanitizer and for that we are using clang rather than gcc. 

I will fix the `PATH` settings.

hbokhari: You are actually correct about CFLAGS - will fix that as well.

hbokhari:
Long running Evergreen build with long testing: https://evergreen.mongodb.com/version/5df179c2e3c33115f7247628

test/evergreen.yml
```diff
@@ -97,6 +97,19 @@ functions:
...
+  "compile wiredtiger sanitizer":
+    - command: shell.exec
+      params:
+        working_dir: "wiredtiger/build_posix"
+        shell: bash
+        script: |
+          set -o errexit
+          set -o verbose
+          sh reconf
+          CC="/opt/mongodbtoolchain/v3/bin/clang -fsanitize=address" CXX=/opt/mongodbtoolchain/v3/bin/g++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb -fPIC -fno-omit-frame-pointer" \
```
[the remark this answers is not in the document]
hbokhari: fixed

test/evergreen.yml
```diff
@@ -1783,6 +1811,41 @@ tasks:
...
+    depends_on:
+      - name: compile-sanitizer
+    commands:
+      - func: "fetch artifacts"
+        vars: 
+          dependent_task: compile-sanitizer
+      - func: "compile wiredtiger sanitizer"
+      - func: "format test script"
+        vars:
+          test_env_vars: ${test_env_vars|} ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
```
raviprakashgiri29: Remove assigning "test_env_vars" to itself ?

test/evergreen.yml
```diff
@@ -1783,6 +1811,41 @@ tasks:
...
+    commands:
+      - func: "fetch artifacts"
+        vars:
+          dependent_task: compile-sanitizer
+      - func: "compile wiredtiger sanitizer"
+      - func: "format test script"
+        # to emulate the original Jenkins job's test coverage, we are running the smoke test 16 times
+        # run smoke tests, stop at first error, use default config (-F -S)
+        vars:
+          test_env_vars: ${test_env_vars|} ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
```
raviprakashgiri29: Remove assigning "test_env_vars" to itself ?

test/evergreen.yml
hbokhari: Fixed

test/evergreen.yml
hbokhari: Fixed

test/evergreen.yml
```diff
@@ -134,6 +147,16 @@ functions:
...
         done
+  "format test script":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/test/format"
+      script: |
+        set -o errexit
+        set -o verbose
+        for i in $(seq ${times|1}); do
+          ${test_env_vars|} ./format.sh ${format_test_script_args|} ${smp_command|} 2>&1
```
lukech: `format.sh` has `[format-configuration]` as it's last command argument, which is used to pass format program arguments e.g. `timer=5 run=4`. Is `format_test_script_args` going to cover `[format-configuration] argument`? If yes the `${smp_command|}` part needs to be moved forward.

test/evergreen.yml
```diff
@@ -255,6 +278,11 @@ tasks:
         vars:
           posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-strict --enable-python
 
+  - name: compile-sanitizer
```
lukech: It looks there's an existing task `compile-asan` having the same configure flags, which might be reused with an overload of `posix_configure_flags` to meet the need here.

test/evergreen.yml
```diff
@@ -1337,7 +1365,7 @@ tasks:
     commands:
       - func: "fetch artifacts"
       - func: "compile wiredtiger"
-      - func: "test format"
+      - func: "format test"
```
lukech: Thanks for making this naming refactoring change.

test/evergreen.yml
hbokhari: Good catch. Will move the `${smp_command|}` option

test/evergreen.yml
hbokhari: Ah we went circles and Ravi and I have decided not to use task for compilation. We will follow the same style for rest of the tickets as well.

test/evergreen.yml
lukech: To avoid confusion with other sanitizers e.g. memory, undefined behavior, suggest adding `address` into the function name.

test/evergreen.yml
```diff
@@ -97,6 +97,19 @@ functions:
...
+      params:
+        working_dir: "wiredtiger/build_posix"
+        shell: bash
+        script: |
+          set -o errexit
+          set -o verbose
+          sh reconf
+          CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb -fPIC -fno-omit-frame-pointer -fsanitize=address" \
+            ../configure ${configure_python_setting|} \
+            --enable-diagnostic --with-builtins=lz4,snappy,zlib
```
lukech: Is it possible to call `*configure_wiredtiger` and overload `configure_env_vars` and `posix_configure_flags` here? It would be neater, and `*configure_wiredtiger` has taken Windows into consideration, which may benefit future work (e.g. add santizer test coverage for Windows).

tammybailey: I was about to ask the same. From what I understand, the initial motivation was to make this function apply to all sanitizer tasks, but from what Luke is saying it seems as if they will require different flags.

lukech: `compile-asan` task could be a good reference.

test/evergreen.yml
```diff
@@ -1783,6 +1806,33 @@ tasks:
...
+    #set a 25 hours timeout
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger sanitizer"
+      - func: "format test script"
+        vars:
+          test_env_vars: ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
+          # run for 24 hours ( 24 * 60 = 1440 minutes), stop at first error, use default config
+          format_test_script_args: -F -t 60
```
lukech: Assuming the `-t` setting will be adjusted to 24hrs. 
We may not want the task stops upon the first failure, instead tolerate the failures and run through 24hrs.

test/evergreen.yml
hbokhari: Will fix it.

test/evergreen.yml
hbokhari: So essentially remove the -F tag and let run the script running for full cycle?

test/evergreen.yml
```diff
@@ -1783,6 +1806,33 @@ tasks:
...
+      - func: "get project"
+      - func: "compile wiredtiger sanitizer"
+      - func: "format test script"
+        # to emulate the original Jenkins job's test coverage, we are running the smoke test 16 times
+        # run smoke tests, stop at first error, use default config (-F -S)
+        vars:
+          test_env_vars: ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
+          # run for 24 hours ( 24 * 60 = 1440 minutes), stop at first error, use default config
+          format_test_script_args: -F -S
+          times: 2
```
lukech: Assuming you are going to change it to 16 to match the comment?

hbokhari: Yup. Need this will be part of final version for approval

test/evergreen.yml
lukech: Yup, running full cycle of 24hrs is to simulate continuous testing we currently have in Jenkins.

test/evergreen.yml
lukech: @tammybailey `--with-builtins=lz4,snappy,zlib` is special for stress testing, compared to non-stress asan tests.

hbokhari: @lukech @tammybailey - So I looked into calling `*configure_wiredtiger` with expansion variables and it doesn't work. I looked into documents and by the looks if it we can't call another function from a function as we can call it from a task definition. 

I can copy the logic for windows/linux configuration. I know this won't be super neat either but we are sort of tied by the limitations of Evergreen functions semantics and the fact that `fetch arctifacts` is sometime useless and we still need do compile afterwards (and manually set the function parameters in calling task).

tammybailey: Oh sorry I meant at the task level. It looks like e.g. the `configure-combinations` calls the "configure wiredtiger" function to set specific vars and flags, followed by a call to the "make wiredtiger" function.

hbokhari: @tammybailey Yes can call it at task context with expansions. But we can't call the task within a task :) . What we can do is make our main task depend on this new special compile task and this is where we hit another problem. Apparently we can't use the output from the special task in our main task because of some issue with embedded static paths in binaries that change from task to task. Have a look at this file for examples where it is done this way and then there is another call for `compile wiredtiger` after fetching the results from compile task! 

No easy option to pick :(

tammybailey: I'm not sure what you mean by task within a task? I was thinking something like this:
```
  - name: format-stress-sanitizer-test
    #set a 25 hours timeout
    exec_timeout_secs: 90000
    commands:
      - func: "get project"
      - func: "configure wiredtiger"
        vars:
          configure_env_vars: CC=/opt/mongodbtoolchain/v3/bin/clang CXX=/opt/mongodbtoolchain/v3/bin/clang++ PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb -fPIC -fno-omit-frame-pointer -fsanitize=address"
          posix_configure_flags: ${configure_python_setting|} --enable-diagnostic --with-builtins=lz4,snappy,zlib
      - func: "make wiredtiger"
      - func: "format test script"
        vars:
          test_env_vars: ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
          # run for 24 hours ( 24 * 60 = 1440 minutes), stop at first error, use default config
          format_test_script_args: -F -t 60
```

hbokhari: Yes that is another option and you can use this as well. But we have multiple tasks that are compiling with these configure options and again the idea is to have a single compile functions for all stress tests that need address sanitization. So it's either between writing a special compile function and call it without any arguments in all relevant task (there are few more such tasks) or calling the generic `compile wiredtiger` with extra arguments in each relevant task.

hbokhari:
There are two failure in Evergreen testing which are not related to changes introduced in this commit. One of the failures is a known issue (WT-5219) whereas I have created a new Jira ticket to investigate the other issue (WT-5322).

test/evergreen.yml
```diff
@@ -1783,6 +1808,32 @@ tasks:
...
+    #set a 25 hours (25*60*60 = 90000 seconds) timeout
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger address sanitizer"
+      - func: "format test script"
+        vars:
+          test_env_vars: ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
+          # run for 24 hours ( 24 * 60 = 1440 minutes), don't stop at failed tests, use default config
+          format_test_script_args: -t 1440
```
lukech: Is `-n 10` missed in the args setting?

test/evergreen.yml
```diff
@@ -1783,6 +1808,32 @@ tasks:
...
+  
+  - name: format-stress-sanitizer-smoke-test
+    #set a 25 hours timeout
+    exec_timeout_secs: 90000
```
lukech: This setting feels a bit over the needed. Will the default setting (6 hrs) be enough for this task?

test/format/format.sh
```diff
@@ -48,9 +48,6 @@ smoke_list=(
 	"$smoke_base_1 file_type=row huffman_key=1 huffman_value=1"
 	"$smoke_base_1 file_type=var huffman_key=1 huffman_value=1"
 
-	# Abort/recovery test.
-	"$smoke_base_1 file_type=row abort=1"
-
```
lukech: Any reason to remove the abort/recovery test?

test/evergreen.yml
hbokhari: Not sure why `-n 10` would be needed here?

test/format/format.sh
hbokhari: Refer to discussion in WT-5322. Keith suggested removing abort/recovery as it will use the some empty page trick that doesn't work well with AddressSanitizer.

lukech:
@agorrod This PR is likely to be merged soon as the 1st stress test being migrated. The runtime of one of the new tasks is set to 24 hrs (as we lined up before).

@hbokhari had a [patch build](https://evergreen.mongodb.com/task/wiredtiger_ubuntu1804_format_stress_sanitizer_test_patch_c4be1b8dd55237f9f940761884dce9e8fe5acb7b_5df3191657e85a7afac46c28_19_12_13_04_52_39) ran through 24 hrs, and the task was marked as failed (capturing a known heap-use-after-free issue).

Just want to set expectations those stress tasks could fail and show red in the waterfall builds, considering their long-running nature which raises the chance of catching issues (known or unknown). We can always tweak it later if it turns out too many stress tasks fail in the waterfall builds.

agorrod:
Thanks for the heads up @lukech, and I'm excited this is going in.

I understand the risk of continual failure - we'll just have to monitor and see if it's a problem we need to address.

test/evergreen.yml
lukech: The Jenkins job runs the test in a for loop for 10 times, and I believe the `-n 10` will help to achieve that. Without `-n` setting it will run as many times as possible (i.e. no limit).

test/format/format.sh
lukech: Oh thanks. Looks I missed the discussion on that ticket.

test/evergreen.yml
hbokhari: We are using -t option to limit the runtime of this task to 24 hours. The `format.sh` script will keep running for 1440 mins (24 hours), keep spawning new job when old ones complete, and then exit gracefully. The `-n` options dictates `format.sh` to run for 10 times and exit which we don't want. The loop in original Jenkins job runs for 10 times and then the job itself get re-triggered. We are using `-t` options of `format.sh` to simulate the for loop in original Jenkins job and the self trigger mechanism.

test/evergreen.yml
hbokhari: This one takes about 5h30mins to finish so 6hr limit may or may not be an issue. Perhaps 7hr might be good number.

lukech: 7 hrs sounds good to me.

test/evergreen.yml
lukech: > The -n options dictates format.sh to run for 10 times and exit which we don't want.
I think `-n` only controls the number of run times in each round. How many rounds of testing to be executed is controlled by the `-t` setting (or the default timer value is `-t` is not specified). 

Anyway, from realizing continuous testing perspective it's not that critical to have `-n` setting. The only benefit of setting `-n 10` is to make each round of testing to execute the same number of runs as what Jenkins does today.  I'm okay with the current implementation.

hbokhari: I believe if both `-t` and `-n` options are given to `format.sh` as input arguments, the script either spawns `n` number of jobs and exit when these job finish, or, runs until `t` minutes have elapsed and spawning as many job as required given whichever condition get met first.

hbokhari:
@lukech Thanks for reviewing this PR. The failures are known WT issues WT-5219 and WT-5312

PR #5033, opened 2019-12-09
WT-5298 checkpoint to check validity window of history store pages

kommiharibabu:
Checkpoint has to check for the obsolete pages in history
store while checkpointing the history store by checking the
clean pages whether the stop time pair of the page makes it
all visible in the page or not? If all are visible on the page
means that the page is obsolete. Marke the obsolete page as
empty and their parent as dirty. During the parent page
processing, this page reference is removed in the parent thus
this page is no more referenced in the new checkpoint.

Whenever the old checkpoint that refers this page deleted
automatically these pages are deleted.


src/btree/bt_sync.c
```diff
@@ -248,6 +250,29 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                    if (mod == NULL)
+                        WT_ERR(__wt_page_modify_init(session, walk->page));
+
+                    /*
+                     * Mark the page as empty and also set the parent page as dirty. This is to
+                     * ensure when the parent page is checkpointing, the empty child page will be
+                     * cleaned.
+                     */
+                    if (!WT_PAGE_IS_INTERNAL(walk->page)) {
+                        walk->page->modify->rec_result = WT_PM_REC_EMPTY;
```
kommiharibabu: Is setting the page as empty directly without reconciling code is fine?

kommiharibabu:
test this please

src/btree/bt_sync.c
agorrod: @keithbostic could you take a look at Hari's question?

keithbostic: I don't think it's dangerous, but I don't think it's a useful optimization to do in the checkpoint code.

Why not have reconciliation do the same check? (We'd want to fast-track LAS file eviction too, I think.)

So, I wouldn't try and fast-path it here.

kommiharibabu: @keithbostic, Sorry I didn't understand your point clearly. Currently, only the clean pages are checked for obsolete in the checkpoint as they are skipped to pass to reconciliation. So moving this part of the code to reconciliation by changing to ensure all the pages go to reconciliation even if it is clean and handle this obsolete check there?

keithbostic: @kommiharibabu, of course -- thank you for pointing that out!

I would still not do this work here. It's not saving us any work in this checkpoint, and the "normal" work of handling empty pages wouldn't get done (for example, reverse splits up the tree).

Please let me know if I'm missing the idea, that there is a win here by getting the checkpoint done more quickly, or more simply.

Maybe it would be reasonable for eviction to take into account the visibility of the LAS pages when considering the pages to choose for eviction? And, of course, reconciliation could fast-track and not do any work.

keithbostic: @kommiharibabu, I hadn't read WRITING-3405, so I didn't understand what you're trying to do here.

I think I understand better now.

I think the way this has to be work is by setting `WT_REF.state = WT_REF_DELETED` for the leaf page. Ideally, we can use the `WT_REF_DELETED` state and not invent a new one. The tradeoff is that much of the reconciliation code is already written to handle the `WT_REF_DELETED` state, but obviously there may not be enough overlap between fast truncate and discarding history file pages for this to work.

It seems to me that checkpoint (`bt_sync.c`) is going to switch the leaf page `WT_REF.state` field (which may require locking and/or atomic operations), then reconciliation is going to handle the rest. Specifically, `__rec_child_modify()` on the internal parent page is going to notice the `WT_REF.state` (https://github.com/wiredtiger/wiredtiger/blob/develop/src/reconcile/rec_child.c#L141-L154), call `__rec_child_deleted()`, which is going to actually free the underlying block in the checkpoint (https://github.com/wiredtiger/wiredtiger/blob/develop/src/reconcile/rec_child.c#L72).

We could duplicate the code in reconciliation to handle a new state, but I think that's going to be more work overall than teasing apart `WT_REF.state = WT_REF_DELETED` with and without an associated `WT_REF.page_del` value.

kommiharibabu: @keithbostic, Thanks for your input. I implemented according to your suggestion for a leaf page that has `WT_ADDR` currently. Created an another ticket to handle the scenario of the page is in-memory and has multiple blocks that are obsolete.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
                     btree->rec_max_txn = mod->rec_max_txn;
                 if (mod != NULL && btree->rec_max_timestamp < mod->rec_max_timestamp)
                     btree->rec_max_timestamp = mod->rec_max_timestamp;
+
+                /*
+                 * Check whether the tree is lookaside btree and verify whether the page contents
+                 * are obsolete or not by checking the visibility of the page stop time pair.
+                 */
+
```
sueloverso: Remove this blank line.

src/btree/bt_sync.c
```diff
@@ -235,6 +252,7 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
 
         for (;;) {
             WT_ERR(__sync_dup_walk(session, walk, flags, &prev));
+        skip_dup_walk:
```
sueloverso: Clang format is supposed to leave goto labels at the margin. I think there's a problem with ones that have underscores in them. And having it at the margin is important. Can you try renaming it to `skipwalk` and moving it to the margin and see if it stays there?

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                /*
+                 * Check whether the tree is lookaside btree and verify whether the page contents
+                 * are obsolete or not by checking the visibility of the page stop time pair.
+                 */
+
+                if (is_las) {
+                    obsolete = false;
+
+                    __wt_ref_is_obsolete(session, walk, &obsolete);
+
```
sueloverso: Remove both of these blank lines. Also, the first thing you do in `wt_ref_is_obsolete` is initialize `obsoletep` so you don't need to do it here also. This one should be removed.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                    if (obsolete) {
+                        /*
+                         * Get the duplicate tree walk point before marking the current tree page as
```
sueloverso: I think `point` should be `pointer`?

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                         * deleted, because the duplicate tree walk expects the ref state should be
+                         * in memory.
```
sueloverso: This comment is hard to follow. I suggest reordering the sentence maybe like this:
```
/*
 * The duplicate tree walk code expects the ref state to be in memory. We need to get the duplicate tree walk
 * pointer before marking the current tree page as deleted.
 */
```

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                         */
+                        WT_ERR(__sync_dup_walk(session, walk, flags, &prev));
+
+                        /*
+                         * Mark the page as empty and also set the parent page as dirty. This is to
```
sueloverso: I suggest: `Mark the page as deleted...` since that is the state set below.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                         * ensure when the parent page is checkpointing, the empty child page will
```
sueloverso: Maybe replace the word `empty` with `unneeded`.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                            if (!WT_REF_CAS_STATE(session, walk, previous_state, WT_REF_LOCKED))
+                                continue;
+
+                            /* Mark the page as deleted */
+                            WT_REF_SET_STATE(walk, WT_REF_DELETED);
+                            WT_ERR(__wt_page_parent_modify_set(session, walk, true));
+                        } else {
+                            /* Mark all the child pages as obsolete, including the current page */
+                        }
+
```
sueloverso: Remove blank line.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                        goto skip_dup_walk;
+                    }
+                }
+
```
sueloverso: Remove blank line.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                        previous_state = walk->state;
+                        if (!WT_PAGE_IS_INTERNAL(walk->page)) {
+                            if (!WT_REF_CAS_STATE(session, walk, previous_state, WT_REF_LOCKED))
+                                continue;
+
+                            /* Mark the page as deleted */
+                            WT_REF_SET_STATE(walk, WT_REF_DELETED);
+                            WT_ERR(__wt_page_parent_modify_set(session, walk, true));
+                        } else {
+                            /* Mark all the child pages as obsolete, including the current page */
```
sueloverso: Is this a place holder for future work? Is that part of this ticket? Otherwise, the (non-existent) code and comment don't match here.

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+
                 continue;
             }
 
+            /*
+             * Scan the internal page of the history store to find out the obsolete leaf pages and
+             * discard them.
+             */
+            if (is_las && WT_PAGE_IS_INTERNAL(walk->page)) {
+            }
```
sueloverso: Do we need this placeholder in here? Or can you keep it aside and merge in clean code?

src/btree/bt_sync.c
```diff
@@ -248,9 +266,56 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                                continue;
+
+                            /* Mark the page as deleted */
+                            WT_REF_SET_STATE(walk, WT_REF_DELETED);
+                            WT_ERR(__wt_page_parent_modify_set(session, walk, true));
+                        } else {
+                            /* Mark all the child pages as obsolete, including the current page */
+                        }
+
+                        goto skip_dup_walk;
```
sueloverso: I think the unconditional go to needs a comment about why we're doing it again here.

src/btree/bt_sync.c
kommiharibabu: Fixed

src/btree/bt_sync.c
kommiharibabu: corrected as per your suggestion and it works with clang-format.

src/btree/bt_sync.c

src/btree/bt_sync.c

src/btree/bt_sync.c
kommiharibabu: Corrected as per your suggestion.

src/btree/bt_sync.c

src/btree/bt_sync.c
kommiharibabu: I went with your above suggestion of deleted.

src/btree/bt_sync.c

src/btree/bt_sync.c

src/btree/bt_sync.c
kommiharibabu: I removed the future ticket work code.

src/btree/bt_sync.c
kommiharibabu: I removed the future ticket code.

src/btree/bt_sync.c
kommiharibabu: Added a comment to explain the reason.

src/btree/bt_sync.c
```diff
@@ -248,6 +266,44 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+
+                            /* Mark the page as deleted */
+                            WT_REF_SET_STATE(walk, WT_REF_DELETED);
+                            WT_ERR(__wt_page_parent_modify_set(session, walk, true));
+                        }
+
+                        /*
+                         * The Duplicate tree walk pointer is already obtained whenever an obsolete
+                         * page is identified before mark the page as deleted, continue with the
+                         * rest of the tree walk without duplicating walk pointer.
```
sueloverso: Thanks! I suggest the following rewording, based on my understanding here:
```
/*
 * The duplicate tree walk pointer, prev, was obtained before we marked the page as deleted.
 * Continue the rest of the tree walk using that pointer.
 */
```

src/btree/bt_sync.c
```diff
@@ -100,6 +100,22 @@ __sync_dup_walk(WT_SESSION_IMPL *session, WT_REF *walk, uint32_t flags, WT_REF *
...
+{
+    WT_ADDR *addr;
+
+    *obsoletep = false;
+
+    addr = ref->addr;
+    if (addr && __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts))
+        *obsoletep = true;
+}
+
```
keithbostic: This function is static to the file and has no error return value, so I'd suggest:
```
/*
 * __sync_ref_is_obsolete --
 *     Return whether the ref is obsolete according to the newest stop time pair.
 */
static bool
__sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref) 
{
    WT_ADDR *addr;

    addr = ref->addr;

    return (addr != NULL && 
        __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts));
}
```

Regardless of what you choose, `addr` isn't a boolean, so needs an explicit comparison against `NULL`.

src/btree/bt_sync.c
```diff
@@ -248,6 +266,43 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+                            WT_ERR(__wt_page_parent_modify_set(session, walk, true));
+                        }
+
+                        /*
+                         * The duplicate tree walk pointer, prev, was obtained before we marked the
+                         * page as deleted. Continue the rest of the tree walk using that pointer.
+                         */
+                        goto skipwalk;
+                    }
+                }
```
keithbostic: It looks to me like we can jump to `skipwalk` if we fail to set the `WT_REF` to `WT_REF_LOCKED`, which is a little simpler.

We also don't use `previous_state` for anything I don't see any reason to have it

Something like:
```
/*  
 * Check whether the tree is lookaside btree and verify whether the page contents
 * are obsolete or not by checking the visibility of the page stop time pair.
 */  
if (is_las) {
    __wt_ref_is_obsolete(session, walk, &obsolete);
    if (obsolete) {
        /*  
         * The duplicate tree walk code expects the ref state to be in memory. We
         * need to get the duplicate tree walk pointer before marking the current
         * tree page as deleted.
         */  
        WT_ERR(__sync_dup_walk(session, walk, flags, &prev));
        
        /*  
         * Mark the page as deleted and also set the parent page as dirty. This is
         * to ensure when the parent page is checkpointing, the empty child page
         * will be cleaned.
         */  
        if (WT_PAGE_IS_INTERNAL(walk->page) &&
            WT_REF_CAS_STATE(session, walk, walk->state, WT_REF_LOCKED)) {
            WT_REF_SET_STATE(walk, WT_REF_DELETED);
            WT_ERR(__wt_page_parent_modify_set(session, walk, true));
        }   

        /*  
         * The duplicate tree walk pointer, prev, was obtained before we marked the 
         * page as deleted. Continue the rest of the tree walk using that pointer.
         */  
        goto skipwalk;
    }   
}   
continue;
```

src/include/btmem.h
```diff
@@ -858,7 +858,8 @@ struct __wt_page_deleted {
     WT_UPDATE **update_list; /* List of updates for abort */
 };
 
-/* WT_TIME_PAIR --
+/*
+ * WT_TIME_PAIR --
```
keithbostic: We no longer use `WT_TIEM_PAIR` anywhere, it can be removed.

src/btree/bt_sync.c
kommiharibabu: I changed the function as per your suggestion. Later this function gets modified to include the pages that are in-memory and obsolete.  These will be handled in another ticket. Currently, the in-memory pages are ignored and they never are deleted until they get evicted. Whenever an in-memory page is found obsolete, planning to add them to urgent eviction queue and later when the checkpoint is happening on history store, those pages get removed while traversing the internal page of history store.

src/btree/bt_sync.c
kommiharibabu: I further optimized the code into a function as this function is further used when traversing internal page cleanup stage.

src/include/btmem.h
kommiharibabu: `WT_TIME_PAIR` is currently used in many functions in `pm-1521-...` branch, so I left it as it is.

src/btree/bt_sync.c
```diff
@@ -100,6 +100,48 @@ __sync_dup_walk(WT_SESSION_IMPL *session, WT_REF *walk, uint32_t flags, WT_REF *
...
+static int
+__sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
+{
+    /* Ignore root pages as they are never be deleted */
+    if (__wt_ref_is_root(ref))
+        return (0);
+
+    /* Guard against marking a page to be deleted when its reconciliation state is not empty. */
+    if (ref->page && ref->page->modify && ref->page->modify->rec_result != WT_PM_REC_EMPTY)
+        return (0);
```
keithbostic: Both `ref.page` and `ref.page.modify` are pointers, and so should be compared to NULL, not treated like booleans.

keithbostic: I'm a little confused why this test is correct: generally, I would expect pages that reconcile to `WT_PM_REC_EMPTY` to be reverse split and then discarded before we get here (see the `WT_REF_EMPTY` case in `__evict_page_dirty_update()`), but I'm willing to believe it's right based on other changes in this branch.

kommiharibabu: `__evict_page_dirty_update()` is called only during the page evict. After reconciliation, if the page is empty, it is removed during the eviction, but not in the checkpoint, because of this reason, it is possible.

kommiharibabu:
retest this please

kommiharibabu:
evergreen retry

PR #5034, opened 2019-12-10
WT-5221 Bypass test_wt2853_perf in Evergreen make-check-msan-test

tammybailey:
I added `TESTUTIL_SLOW_MACHINE=1` to the environment variable list for `make-check-msan-test` to bypass the `test_wt2853_perf` test. The existing code already checks for this environment variable, so the only change required was to the Evergreen configuration.

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -283,7 +283,7 @@ tasks:
           posix_configure_flags: --enable-silent-rules --enable-strict --enable-diagnostic --disable-static
       - func: "make check all"
         vars:
-          test_env_vars: MSAN_OPTIONS=abort_on_error=1:disable_coredump=0 MSAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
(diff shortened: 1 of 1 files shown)


hbokhari:
@lukech Just wondering what are the implication of skipping `test_wt2853_perf` test? If this test isn't part of any other testing we will be missing on test coverage.

lukech:
@hbokhari good question. `test_wt2853_perf` is part of csuite tests, and will be triggered by `make check`. There are some Evergreen tasks (e.g. `make-check-test`, `csuite-wt2853-perf-test`) have the test covered already. This specific test has some history of running super slow on slow machines (e.g. zSeries) and in valgrind job, which is why the `TESTUTIL_SLOW_MACHINE` and `TESTUTIL_BYPASS_VALGRIND` flag checkings were added to bypass the test. I believe memory sanitizer is yet another such case.

tammybailey:
@lukech Yes, I should have mentioned that I re-ran the patch build at least six times thus far and it has passed every time.

hbokhari:
LGTM in that case then.

raviprakashgiri29:
LGTM

PR #5036, opened 2019-12-11
WT-4969 remove lsm_merge option from auto-generated config definitions

jeremytay90:


wtbuild:

lukech:
test this please

tetsuo-cpp:
@jeremytay90 PR testing failed because the `dist/s_all` script is complaining that the `__wt_config_upgrade` function always returns `0` so it could have `void` return type instead. Imo, this function should just go in the exceptions list.

dist/s_void
```diff
@@ -124,6 +124,7 @@ func_ok()
 	    -e '/int wiredtiger_extension_init$/d' \
 	    -e '/int wiredtiger_extension_terminate$/d' \
 	    -e '/int wiredtiger_pack_close$/d' \
+		-e '/int __wt_config_upgrade$/d' \
```
tetsuo-cpp: Nit: Indentation isn't consistent here.

src/config/config_upgrade.c
```diff
@@ -15,17 +15,8 @@
...
-    config = buf->data;
-
-    /*
-     * wiredtiger_open:
-     *	lsm_merge=boolean -> lsm_manager=(merge=boolean)
-     */
-    if (__wt_config_getones(session, config, "lsm_merge", &v) != WT_NOTFOUND)
-        WT_RET(__wt_buf_catfmt(session, buf, ",lsm_manager=(merge=%s)", v.val ? "true" : "false"));
-
+    // Currently not used and included in the func_ok list of s_void
```
tetsuo-cpp: In WiredTiger, we use `/* blah */` instead of `// blah`.

src/config/config_upgrade.c
```diff
@@ -15,17 +15,8 @@
 int
 __wt_config_upgrade(WT_SESSION_IMPL *session, WT_ITEM *buf)
```
tetsuo-cpp: We had a chat with @agorrod and his suggestion is to remove this function entirely. It's unlikely we'll need to symlink API options to each other in the future.

jeremytay90: Removed this and references to this

src/config/config_upgrade.c
jeremytay90: Fixed and noted for future reference

dist/s_void
jeremytay90: resolved

PR #5039, opened 2019-12-11
WT-5263 prepared updates written to the lookaside file are not always read as needed

keithbostic:
@bvpvamsikrishna, this change fixes the failure, but I'm not familiar enough with the lookaside visibility rules to be confident it's the correct fix.

Additionally, I would suggest reviewing the update walk loop in the `__wt_rec_upd_select()` function for any other information not being collected because prepared updates are detected and ignored too early in the loop. I reviewed it and don't see anything, but I think it's worth further review.


bvpvamsikrishna:
@keithbostic, seems fine on the first look, I would like to stare it for more time tomorrow.

PR #5046, opened 2019-12-13
WT-5257 Coverity dereference after null check

jeremytay90:
I've added an assert to check that the previous reconciliation buffer (r->prev_ptr) is not null. This is because this value should always be initialized when a write is triggered either because it was 'forced' or part of a 'bulk-load'. 

The hope is that this silences Coverity, otherwise I will manually mark the report as a false positive. The assertion remains valid in order to find any actual exceptions.


wtbuild:

lukech:
add to whitelist

lukech:
ok to test

PR #5048, opened 2019-12-13
WT-5319 Avoid clearing the saved last-key when no instantiated key

kommiharibabu:
Reconciliation has some fast-path code to work on prefix-compressed
to avoid figuring out each key from scratch which would require
going back some number of K/V pairs to key that's not
prefix-compressed and then rolling forward), we use the complete
key we just built for the last K/V pair's reconciliation, and
apply the current key's prefix-compression change to it, saving
us quite a bit of work. Clearing the last-key information can
leads to a performance drop.

### DIFF
--- a/src/include/btree.i
+++ b/src/include/btree.i
@@ -838,10 +838,6 @@ __wt_row_leaf_key_info(
             *ikeyp = NULL;
         if (cellp != NULL)
             *cellp = WT_PAGE_REF_OFFSET(page, WT_CELL_DECODE_OFFSET(v));
-        if (datap != NULL) {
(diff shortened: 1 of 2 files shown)


kommiharibabu:
@keithbostic, The earlier fix was producing some warnings. This fix has included the warnings fixes also.

kommiharibabu:
retest this please

keithbostic:
@kommiharibabu, I hijacked your PR to add a comment into `format`. No code change or anything, just a comment. I forgot to add it in another ticket, and it was hard enough to figure out that I really wanted a comment there. Hope that's OK!

PR #5049, opened 2019-12-13
WT-5314 For python tests, if the name of an extension is blank, don't try to load it

ddanderson:
Extensions that have a name of None, or "" or "none" are treated the same, they are not loaded.


ddanderson:
@quchenhao, can you review this please?  It eliminates the ignored `"ERROR"` that you see, in fact it runs those tests, which it didn't before.

quchenhao:
lgtm

PR #5050, opened 2019-12-14
WT-4997 Migrate Jenkins wiredtiger-test-format-stress job to evergreen

raviprakashgiri29:


raviprakashgiri29:
retest this please

raviprakashgiri29:
Ignore "format test script" function implementation. Added this just to run all the evergreen jobs. I will be reusing the function implementation from "WT-5000" PR.

test/evergreen.yml
```diff
@@ -1798,6 +1823,30 @@ tasks:
...
+    # Set 25 hours timeout
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger with builtins"
+      - func: "format test script"
+        vars:
+          # to emulate the original Jenkins job's test coverage, we are running the smoke test 16 times
+          # run smoke tests, use default config (-S)
+          format_test_script_args: -S
```
hbokhari: Command prefixes `SEGFAULT_SIGNALS=all catchsegv` are missing.

test/evergreen.yml
```diff
@@ -97,6 +97,21 @@ functions:
           set -o verbose
           echo '#undef HAVE_FTRUNCATE' >> wiredtiger_config.h
     - *make_wiredtiger
+  "compile wiredtiger with builtins":
```
hbokhari: Just curious - do we use this compilation flag combination in other tests as well?

raviprakashgiri29: Yes, other format-stress jobs  ppc, zseries can reuse this compile function.

test/evergreen.yml

raviprakashgiri29:
> Personally, I believe it is cleaner to override the environment variables and flags at the task level rather than adding several variations of "compile wiredtiger, with the only difference between them being their environment variables and flags. I also don't have enough insight into the reusability of these functions - I assume that more than just the two tasks here will be able to use the "compile wiredtiger with builtins" function?

I agree, using "compile wiredtiger" and overriding flags would be an easy way. But, if the combination of compilation flags are to be used in more than 2 tasks then it would be better to make it as a function and reuse. In this case, "compile wiredtiger with builtins"  would be used in ppc, zseries and the existing format stress jobs.

hbokhari:
@raviprakashgiri29 Can you please run `format-stress-test` and `format-stress-smoke-test` tasks to make sure scripts run through intended runtime?

raviprakashgiri29:
> @raviprakashgiri29 Can you please run `format-stress-test` and `format-stress-smoke-test` tasks to make sure scripts run through intended runtime?

Evergreen patch build

https://evergreen.mongodb.com/version/5df301ff32f417723e9f155c

raviprakashgiri29:
> > @raviprakashgiri29 Can you please run `format-stress-test` and `format-stress-smoke-test` tasks to make sure scripts run through intended runtime?
> 
> Evergreen patch build which ran for 24 hours
> 
> https://evergreen.mongodb.com/version/5df4cb9fc9ec446209871d0c

test/evergreen.yml
```diff
@@ -134,6 +149,16 @@ functions:
...
+  "format test script":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/test/format"
+      script: |
+        set -o errexit
+        set -o verbose
+        for i in $(seq ${times|1}); do
+          ${test_env_vars|} ./format.sh ${smp_command|} ${format_test_script_args|} 2>&1
+        done
```
lukech: This new function seems to be introduced in the PR of WT-5000 as well. Depending on which ticket is merged first, the other ticket may want to drop this new function definition. @raviprakashgiri29 @hbokhari can I leave it to you to coordinate the 2 PR merges.

test/evergreen.yml
```diff
@@ -1798,6 +1823,30 @@ tasks:
...
+      - func: "get project"
+      - func: "compile wiredtiger with builtins"
+      - func: "format test script"
+        vars:
+          #run for 24 hours ( 24 * 60 = 1440 minutes), use default config
+          format_test_script_args: -b "SEGFAULT_SIGNALS=all catchsegv ./t" -t 1440
+
+  - name: format-stress-smoke-test
+    # Set 25 hours timeout
+    exec_timeout_secs: 90000
```
lukech: I believe this smoke test won't be running up to 24 hrs, so the timeout setting here feels too much. Please refer to #5032 for a similar change.

hbokhari: @lukech I have already merges my changes. @raviprakashgiri29 Please pull latest changes from `develop` and merge the changes.

test/evergreen.yml
raviprakashgiri29: Yes, I used this function just for my testing to progress. Merged the latest changes by @hbokhari.

raviprakashgiri29:
> Changes LGTM. Assume you had some patch build ran through the 24hrs time, what about the test result?

https://evergreen.mongodb.com/version/5df4cb9fc9ec446209871d0c

I ran the above patch build for 24 hours with the old changes. However, I have scheduled a new patch build with the latest changes and will monitor it before merging to develop. Thanks.

https://evergreen.mongodb.com/version/5dfb0a372a60ed552550e57f

lukech:
Oh great, the 24hrs run was successful!

PR #5051, opened 2019-12-15
WT-5125 eviction target strategy stats (#4948) [Backport to 4.0]

tammybailey:
* WT-5125 Additional stats for eviction server

New stats are added to eviction server to find out the strategy
of the eviction server while finding out the pages to be queued for
eviction. Based on the eviction strategy, the eviction server
checks the minimum number of pages before it gives up.

Clean strategy and both clean and dirty strategy  - 10 times target pages
Dirty strategy - 100 times target pages

Along with with them, a new stat that tracks the internal pages and also
the pages that are already present in the queue.

Additional stats related to finding out the reason behind why the page is not able
to be evicted even if it is added to the queue are added.

* Revert the number of records

* Fix the uninitialized variable usage

* Better naming for the page with overflow keys

Eviction cannot evict a page that is dirty that have
overflow keys

* Fix the stat to refer the parent instead of the current

* Fix the merge conflicts

(cherry picked from commit 700d3d96dbd58d9ce653b452cbb06d126ae5477a)


tammybailey:
retest this please

PR #5057, opened 2019-12-16
WT-4921 Add debug mode which slows down checkpoints

jeremytay90:


luke-pearson:
@agorrod you originally requested this change, could you please have a quick look at this PR? I think we could make the creation of a relevant wtperf workload a separate follow on ticket outside of the implementation of the debug mode. What do you think?

src/btree/bt_sync.c
```diff
@@ -271,6 +271,10 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
             if (WT_PAGE_IS_INTERNAL(page)) {
                 internal_bytes += page->memory_footprint;
                 ++internal_pages;
+                /* slow down checkpoints */
```
agorrod: This should be capitalized and have a period at the end: "Slow down checkpoints."

dist/api_data.py
```diff
@@ -512,6 +512,11 @@ def __ge__(self, other):
             versions of WiredTiger. These operations are informational and
             skipped in recovery.''',
             type='boolean'),
+        Config('slow_checkpoint', 'false', r'''
+            if true, each time checkpoint visits an internal page it will
+            sleep for 10000 ms.
```
agorrod: I'd be inclined to not mention a specific time in here, rather make it a bit more generic. Something like: "if true, slow down checkpoint creation by slowing down internal page processing"

src/btree/bt_sync.c
```diff
@@ -271,6 +271,10 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
             if (WT_PAGE_IS_INTERNAL(page)) {
                 internal_bytes += page->memory_footprint;
                 ++internal_pages;
+                /* Slow down checkpoints. */
+                if (conn->debug_slow_ckpt)
+                    __wt_sleep(0, 10000);
+
```
sueloverso: Remove extra blank line.

dist/api_data.py
```diff
@@ -512,6 +512,11 @@ def __ge__(self, other):
             versions of WiredTiger. These operations are informational and
             skipped in recovery.''',
             type='boolean'),
+        Config('slow_checkpoint', 'false', r'''
+            if true, slow down checkpoint creation by slowing down internal 
+            page processing.
+            ''',
+            type='boolean')
```
sueloverso: This needs to move to be listed alphabetically in the set of options, between `rollback_error` and `table_logging`.

src/conn/conn_api.c
```diff
@@ -1805,6 +1805,9 @@ __wt_debug_mode_config(WT_SESSION_IMPL *session, const char *cfg[])
     else
         FLD_CLR(conn->log_flags, WT_CONN_LOG_DEBUG_MODE);
 
+    WT_RET(__wt_config_gets(session, cfg, "debug_mode.slow_checkpoint", &cval));
+    conn->debug_slow_ckpt = cval.val ? true : false;
```
sueloverso: Isn't this the same as `conn->debug_slow_ckpt = cval.val`?

jeremytay90: `cval.val` is a `int64_t`/`signed long int` so I wanted to turn that into a `bool`

jeremytay90: I just tested changing it to what you said and it still works. I suppose our compilation settings allow for this?

luke-pearson:
retest this please

test/suite/test_debug_mode06.py
```diff
@@ -0,0 +1,48 @@
...
+# test_debug_mode06.py
+#   Test the debug mode settings. Test slow_checkpoint use (from WT-4981).
+#   Note: testing timing will make results unreliable so we won't do that.
+class test_debug_mode06(wttest.WiredTigerTestCase):
+    conn_config = 'log=(enabled=true),debug_mode=(slow_checkpoint=true)'
+    uri = 'file:test_debug_mode06'
+
+    # Test (re)configuring the flag on and off.
+    def test_slow_checkpoints(self):
+        self.session.create(self.uri, 'key_format=s,value_format=s')
```
sueloverso: While this tests that the API configuration string can be parsed, is there anything the test can do to verify it works as expected? For instance, instead of having two separate tests, have one test that might do something like this that tests both the API on/off and that it does what it is supposed to do:
* set the `conn_config` as you do above.
* create the table.
* insert a bunch of data into the table.
* perform a checkpoint and record the time it takes to do so.
* reconfigure the setting to false.
* insert the same amount of data into the table.
* perform a checkpoint and record the time it takes.
* assert time1 > time2.

dist/api_data.py
```diff
@@ -505,13 +505,18 @@ def __ge__(self, other):
...
+            ''',
+            type='boolean'),
         Config('table_logging', 'false', r'''
             if true, write transaction related information to the log for all
             operations, even operations for tables with logging turned off.
             This setting introduces a log format change that may break older
             versions of WiredTiger. These operations are informational and
             skipped in recovery.''',
-            type='boolean'),
+            type='boolean')
```
sueloverso: All of the categories have the final entries with trailing comma. I suggest leaving this alone.

src/include/connection.h
```diff
@@ -213,6 +213,7 @@ struct __wt_connection_impl {
 
     WT_LSN *debug_ckpt;      /* Debug mode checkpoint LSNs. */
     uint32_t debug_ckpt_cnt; /* Checkpoint retention number */
+    bool debug_slow_ckpt;    /* Debug mode slow checkpoints */
```
sueloverso: It occurs to me now that there really isn't a need to expand the size of the connection structure for this work. Can you, instead, define and use a bit flag in the connection instead? I suggest calling it `WT_CONN_DEBUG_SLOW_CKPT`. @luke-pearson should be able to help with how to do so using the automatic flag generation/renumbering. See `WT_CONN_LEAK_MEMORY` for similar boolean style usage.

test/suite/test_debug_mode06.py
agorrod: We did have some conversation about testing the actual functionality in Python and I recommended against it:
https://jira.mongodb.org/browse/WT-4921?focusedCommentId=2635468&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2635468
We've had a lot of difficulties getting timing related Python testing reliable.

jeremytay90: Perhaps I can add a few lines which just executes a checkpoint. That way it just ensures that having the flag set doesn't break existing functionality.

sueloverso: Ok, I can agree with relative timings. How about this as a sort of middle ground that does not do comparison between checkpoints:
* insert a few items into the table (so checkpoint should actually be fast)
* perform a checkpoint with the setting on.
* get the last checkpoint time.
* assert it took longer than 10000 usecs or whatever the sleep unit it.

sueloverso: The statistic I have in mind is `txn_checkpoint_time_recent`. So something like:
`ckpt_time = stat__cursor[stat.conn.txn_checkpoint_time_recent][2]`
`self.assertTrue(ckpt_time >= 10)`

(The `wt_sleep` unit is usecs and the stat unit is msecs.)

jeremytay90: I've added a simple insert of data + checkpoint just to make sure the flag doesn't affect functionality.

test/suite/test_debug_mode06.py
```diff
@@ -0,0 +1,58 @@
...
+    uri = 'file:test_debug_mode06'
+
+    # Insert some data to ensure setting/unsetting the flag does not
+    # break existing functionality
+    def insert_data(self):
+        self.session.create(self.uri, 'key_format=s,value_format=s')
+        cursor = self.session.open_cursor(self.uri, None)
+        cursor['key'] = 'value'
+        cursor.close()
+        self.session.checkpoint()
```
sueloverso: While I would prefer the minor statistic check, I won't hold up the merge for it either.

dist/api_data.py
```diff
@@ -505,6 +505,11 @@ def __ge__(self, other):
             return a WT_ROLLBACK error from a transaction operation about
             every Nth operation to simulate a collision''',
             min='0', max='10M'),
+        Config('slow_checkpoint', 'false', r'''
+            if true, slow down checkpoint creation by slowing down internal
+            page processing.
+            ''',
```
sueloverso: Minor nit: move the ''' to the end of the previous line.

PR #5060, opened 2019-12-18
WT-5324 review of error handling in the WiredTiger API

keithbostic:


agorrod:
I'd like to see this one passing tests here and in MongoDB evergreen before we decide if it's the best path - is that OK @keithbostic?

keithbostic:
@agorrod 
> I'd like to see this one passing tests here and in MongoDB evergreen before we decide if it's the best path

Oh, yes. Definitely.

bvpvamsikrishna:
@keithbostic, @agorrod I have created a patch build for this branch and the url is : https://evergreen.mongodb.com/version/5df9b5c50305b97be9fba14c

src/include/api.h
```diff
@@ -86,6 +86,7 @@
     do {                                                                     \
         bool __autotxn = false, __update = false;                            \
         API_CALL(s, h, n, bt, config, cfg);                                  \
+        WT_ERR(__wt_txn_err_chk(s));                                         \
```
sueloverso: To be clear in my understanding, this piece, among all these other changes, really is the core of the change. This function and the two calls here.

src/include/api.h
```diff
@@ -167,24 +169,24 @@
...
-    WT_ERR(__wt_txn_context_prepare_check((s)))
+#define SESSION_TXN_API_CALL(s, n, config, cfg)                    \
+    WT_RET(__wt_txn_context_prepare_check((s), "WT_SESSION." #n)); \
+    TXN_API_CALL(s, WT_SESSION, n, NULL, config, cfg)
 
 #define CURSOR_API_CALL(cur, s, n, bt)                                                     \
     (s) = (WT_SESSION_IMPL *)(cur)->session;                                               \
+    WT_RET(__wt_txn_context_prepare_check((s), "WT_CURSOR." #n));                          \
     API_CALL_NOCONF(s, WT_CURSOR, n, ((bt) == NULL) ? NULL : ((WT_BTREE *)(bt))->dhandle); \
-    WT_ERR(__wt_txn_context_prepare_check((s)));                                           \
```
sueloverso: So has this been wrong the entire time? I.e. we used to check after the `API_CALL` macro? Or is this necessary based on your changes?

test/suite/test_timestamp09.py
```diff
@@ -62,7 +62,6 @@ def test_timestamp_api(self):
             lambda: self.session.timestamp_transaction(
                 'commit_timestamp=' + timestamp_str(2)),
                 '/older than the first commit timestamp/')
-        c[2] = 2
```
sueloverso: Why are these changes necessary? What was the test doing and why was it wrong?

dist/s_define.list
```diff
@@ -2,6 +2,7 @@
 API_CALL
 API_CALL_NOCONF
 API_SESSION_INIT
+API_END
```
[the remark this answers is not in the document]
keithbostic: Fixed in 429c3da90.

src/include/api.h
keithbostic: This is the core of the change originally considered in WT-5324, to try and catch the failure sooner. In looking at our WT_TXN_ERROR handling, I came across some problems, and they were intertwined enough that I didn't think creating new tickets would help.

src/include/api.h
keithbostic: Yes; wrong the entire time.

test/suite/test_timestamp09.py
keithbostic: Setting the timestamp is a "hard" failure, that is, it sets WT_TXN_ERROR. Updating a key/value after that failure hits the assert I added in DIAGNOSTIC mode, that we don't try operations after a transaction has failed. So, this change is required by my other changes.

keithbostic:
I had to run out last night, I apologize for leaving a trail of breadcrumbs for you, @sueloverso.

Here are the reviewer's notes:

1. The 57a9d0f commit is a standalone, minor fix, we were just making a check that wasn't useful.
1. The f520f2f commit is the first interesting bug fix: if the transaction has failed, maede updates and then the application commits, we neither commit nor roll back, we jump out of the function. The 0655cf8 commit is a follow-on, and removes the assert you and I discussed, because I don't believe it's possible any more.
1. The 137eefe commit is a combination of the check of `WT_TXN_ERROR` before doing any operation, (that you and I discussed), and the transaction prepare fix that I discussed separately with @agorrod and @bvpvamsikrishna. In summary, the latter change is for a bug where a prepare, followed by an operation that's not supposed to be done after a prepare, is a hard failure, setting `WT_TXN_ERROR` and forcing an eventual rollback, that is, the operation can no longer be committed. The change is conceptually simple, but requires a lot of code shuffling.

The places of interest for #3 are:

* After an error, the cursor table `APPLY_CG` macro no longer continues to attempt the operation on every related table in the group, unless the error is the `WT_NOTFOUND` error. This is a real change and deserves careful review. Exempting `WT_NOTFOUND` seems correct to me, we might be removing an entry that never made it into all of the tables in the group, for example.
* There was a bunch of code shuffling around the `set_key` and `set_value` functions so that they could return an error using the standard `api.h` macros and ignore it. The obvious alternative was a set of new macros that ignored the return from `__wt_txn_context_prepare_check()` in the macro, but I thought `api.h` was getting complex enough, and we already used the pattern of two functions, one of which returns an error and one which is just a wrapper of the first function and ignores the returned error, in other places for similar reasons.
* Because we call the `__wt_txn_context_prepare_check()` function before setting up the `WT_SESION_IMPL.name` information, I had to do it manually and pass it to `__wt_txn_context_prepare_check()`. That's pretty ugly and messes up the standard error function handling. I'd like a better solution.
* Finally, the addition of the `__wt_txn_err_chk()` function, and the checks of it when we first enter the API.

test/suite/test_timestamp09.py
sueloverso: Instead of removing, would it be useful to change the failure from an assertion to an error and test getting the error here? I realize that an error is sometimes less useful than an error. I don't know if there is an easy way in the python test to know if we're diagnostic or not and only do the test if not diagnostic.

src/include/api.h
sueloverso: I need more explanation. What is wrong with it the way it was? Looking at `API_CALL_NOCONF` it just calls `API_SESSION_INIT` and that pretty much just sets up the session name and does a few random checks. Why is any of that bad and why should the prepare check be done before, say, checking PANIC? In fact, it seems like this new `WT_TXN_ERROR` code could be very similar to `WT_SESSION|CONN_CHECK_PANIC`. You could add a `WT_SESSION_CHECK_TXN` macro at the same place.

keithbostic: The problem is that once we enter the standard API call macros, any error jumps to the `err` label, that calls `API_END`, which, if `ret != 0`, sets `WT_TXN_ERROR`. If a transaction is prepared, we aren't allowed to roll it back if the application chooses to commit it, and setting `WT_TXN_ERROR` will change the behavior of commit to roll back the function instead of commit.

test/suite/test_timestamp09.py
keithbostic: I don't think this tests anything particularly interesting. We could certainly add tests that we catch attempts to perform additional operations after a transactional failure, for a variety of methods. I thought about that, but didn't because I didn't think it would add a lot of value and it would be a chunk of work.

sueloverso: Then that is fine with me. If it was easy, that would be good. If it is a chunk of work then removal of the line is good with me.

src/include/txn.i
```diff
@@ -34,6 +34,50 @@ __wt_ref_cas_state_int(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t old_state
...
+ */
+static inline int
+__wt_txn_context_check(WT_SESSION_IMPL *session, bool requires_txn)
+{
+    if (requires_txn && !F_ISSET(&session->txn, WT_TXN_RUNNING))
+        WT_RET_MSG(session, EINVAL, "%s: only permitted in a running transaction", session->name);
+    if (!requires_txn && F_ISSET(&session->txn, WT_TXN_RUNNING))
+        WT_RET_MSG(session, EINVAL, "%s: not permitted in a running transaction", session->name);
+    return (0);
+}
```
sueloverso: I assume the two functions above were just code movements. If you made content changes to them, please let me know so that I can closely look.

keithbostic: There was a minor change to one function (you don't have to print out `session.name` explicitly, the error functions will do it for you), but otherwise it was just a movement.

agorrod:
Sorry I didn't get to this review yet @keithbostic - can it wait until next week?

keithbostic:
@agorrod 
> can it wait until next week?

I don't think there's anything urgent here. The problems we're fixing are triggered by unexpected use of the WiredTiger API (commit when rollback should have been called, or calling unsupported operations after a prepare).

src/include/api.h
```diff
@@ -32,24 +32,31 @@
...
+    WT_TRACK_OP_DECL;                                              \
+    API_SESSION_PUSH(s, h, n, dh);                                 \
+    /*                                                             \
+     * No code before this line, otherwise error handling won't be \
+     * correct.                                                    \
+     */                                                            \
+    WT_ERR(WT_SESSION_CHECK_PANIC(s));                             \
+    WT_SINGLE_THREAD_CHECK_START(s);                               \
+    WT_ERR(__wt_txn_err_chk(s));                                   \
+    WT_TRACK_OP_INIT(s);                                           \
```
agorrod: Did you deliberately move this one further down? It used to be before the single thread check.

keithbostic: @agorrod 
> Did you deliberately move this one further down? It used to be before the single thread check.

Yes -- I put all the error/panic checks first, and then initialization afterward, they were randomly mixed up before. There's no correctness issue, it just seemed cleaner to me.

PR #5061, opened 2019-12-18
WT-4919 Add exact memory reallocation debug configuration

jeremytay90:


luke-pearson:
This is going to conflict with the other debug mode change but we can still review the functional part, once thats complete and the other change is merged. Rebase this one and re-request review.

src/include/connection.h
```diff
@@ -214,6 +214,8 @@ struct __wt_connection_impl {
     WT_LSN *debug_ckpt;      /* Debug mode checkpoint LSNs. */
     uint32_t debug_ckpt_cnt; /* Checkpoint retention number */
 
+    bool debug_realloc_exact; /* Reallocate exact size */
+
```
sueloverso: Similar to the other one, no need to expand the connection structure for this. You can add a bit in the flags `WT_CONN_DEBUG_REALLOC_EXACT`.

dist/api_data.py
```diff
@@ -501,6 +501,11 @@ def __ge__(self, other):
             is not limited to not skewing newest, not favoring leaf pages,
             and modifying the eviction score mechanism.''',
             type='boolean'),
+        Config('realloc_exact', 'false', r'''
+            if true, reallocation of memory will only provide the exact
+            amount requested. This will help with spotting memory allocation
+            issues more easily.''',
+            type='boolean'),
```
sueloverso: (Not related to this exact location) There should be a python test added for this API. While there isn't actually anything you can measure in the test, it should do something similar to your first test: turn it on, reconfigure it off and do some operations that should cause a reallocation (maybe add in a printf during debugging that does not get committed to know the realloc call was made).

src/include/connection.h
jeremytay90: I'll add this after WT-4921 is merged in case the automatic flag value generation clashes

test/suite/test_debug_mode07.py
```diff
@@ -0,0 +1,57 @@
...
+
+import wttest, wiredtiger
+
+# test_debug_mode06.py
+#   Test the debug mode settings. Test realloc_exact use (from WT-4919).
+class test_debug_mode07(wttest.WiredTigerTestCase):
+    conn_config = 'log=(enabled=true),debug_mode=(realloc_exact=true),statistics=(all)'
+    uri = 'file:test_debug_mode07'
+
+    # Insert some data to ensure setting/unsetting the flag does not
```
luke-pearson: Maybe include in this comment something like: "by calling checkpoint we force the relevant reallocation function to be invoked multiple times"

luke-pearson:
@sueloverso, if this change looks good to you could you please approve and merge? If not it'll have to wait to the new year. Thanks :)

PR #5063, opened 2019-12-18
WT-5330 Read-only transactions are being rolled back in MongoDB

keithbostic:


keithbostic:
>  I would like some test/format testing on this before it's merged. I agree that it's unlikely to break format due to configuring for read timeouts, but I think it's possible.

I've run some minor, local testing of a few CONFIGs, but nothing formal. What would you like to see?

sueloverso:
@keithbostic perhaps adding something to `test_debug_mode03.py` or `test_debug_mode05.py` that test `table_logging`. Neither test, currently, tests debug_mode and reading. But it should be easy to add, and `test_debug_mode03.py` does the necessary verification of the log records.

agorrod:
I've switched http://build.wiredtiger.com:8080/job/wiredtiger-test-format-stress-sanitizer over here. If it runs clean for a day I'm happy for this to be merged in (and for testing to be switched back to develop)

keithbostic:
@agorrod, I'm not seeing any usual behavior (the usual couple of "clean cache" timeouts), so I'm merging WT-5330 and switching the Jenkins tester back to develop.

PR #5066, opened 2019-12-19
WT-5280 Checkpoint writes to lookaside

luke-pearson:
So this change was mostly a lot of fiddling with flags sent into reconciliation from bt_sync.c when we're taking a checkpoint, and then dealing with odd fallout mainly via modifying flow through __rec_upd_select. I'm not a huge fan of how fiddly the if statement is there but it seems to work now in all cases we have in our python test suite.

Additionally I added flags to the update and merged the ext with those flags.

### DIFF
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -304,9 +304,9 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
             }
             tried_eviction = false;
 
-            /* Add in lookaside reconcilliation for standard files.*/
+            /* Add in lookaside reconciliation for standard files.*/
             if (!F_ISSET(btree, WT_BTREE_LOOKASIDE) && !WT_IS_METADATA(btree->dhandle))
-                rec_flags |= WT_REC_LOOKASIDE | WT_REC_EVICT;
+                rec_flags |= WT_REC_LOOKASIDE;
 
             WT_ERR(__wt_reconcile(session, walk, NULL, rec_flags, NULL));
 
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -226,8 +226,6 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
     /* Only visit pages in cache and don't bump page read generations. */
     flags = WT_READ_CACHE | WT_READ_NO_GEN;
 
-    rec_flags = WT_REC_CHECKPOINT;
-
     /*
      * Skip all deleted pages. For a page to be marked deleted, it must have been evicted from cache
      * and marked clean. Checkpoint should never instantiate deleted pages: if a truncate is not
@@ -327,8 +325,14 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
         btree->syncing = WT_BTREE_SYNC_RUNNING;
         is_las = F_ISSET(btree, WT_BTREE_LOOKASIDE);
 
-        /* Add in lookaside reconciliation for standard files. */
-        if (!F_ISSET(btree, WT_BTREE_LOOKASIDE) && !WT_IS_METADATA(btree->dhandle))
+        /*
+         * Add in lookaside reconciliation for standard files.
+         *
+         * FIXME-PM-1521: Remove the WT_BTREE_LOOKASIDE check, and assert that no updates from
+         * lookaside are copied to lookaside recursively.
+         */
+        rec_flags = WT_REC_CHECKPOINT;
+        if (!WT_IS_LOOKASIDE(btree) && !WT_IS_METADATA(btree->dhandle))
             rec_flags |= WT_REC_LOOKASIDE;
 
         /* Write all dirty in-cache pages. */
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -323,16 +323,16 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
         btree->syncing = WT_BTREE_SYNC_WAIT;
         __wt_gen_next_drain(session, WT_GEN_EVICT);
         btree->syncing = WT_BTREE_SYNC_RUNNING;
-        is_las = F_ISSET(btree, WT_BTREE_LOOKASIDE);
+        is_las = WT_IS_LAS(btree);
 
         /*
          * Add in lookaside reconciliation for standard files.
          *
-         * FIXME-PM-1521: Remove the WT_BTREE_LOOKASIDE check, and assert that no updates from
-         * lookaside are copied to lookaside recursively.
+         * FIXME-PM-1521: Remove the is_las check, and assert that no updates from lookaside are
+         * copied to lookaside recursively.
          */
         rec_flags = WT_REC_CHECKPOINT;
-        if (!WT_IS_LOOKASIDE(btree) && !WT_IS_METADATA(btree->dhandle))
+        if (!is_las && !WT_IS_METADATA(btree->dhandle))
             rec_flags |= WT_REC_LOOKASIDE;
 
         /* Write all dirty in-cache pages. */
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -229,6 +229,10 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
         __wt_gen_next_drain(session, WT_GEN_EVICT);
         btree->syncing = WT_BTREE_SYNC_RUNNING;
 
+        /* Add in lookaside reconciliation for standard files.*/
+        if (!F_ISSET(btree, WT_BTREE_LOOKASIDE) && !WT_IS_METADATA(btree->dhandle))
+            rec_flags |= WT_REC_LOOKASIDE;
+
         /* Write all dirty in-cache pages. */
         LF_SET(WT_READ_NO_EVICT);
 
@@ -303,10 +307,6 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
                 continue;
             }
             tried_eviction = false;
-
-            /* Add in lookaside reconciliation for standard files.*/
-            if (!F_ISSET(btree, WT_BTREE_LOOKASIDE) && !WT_IS_METADATA(btree->dhandle))
-                rec_flags |= WT_REC_LOOKASIDE;
 
             WT_ERR(__wt_reconcile(session, walk, NULL, rec_flags, NULL));
 
--- a/src/btree/bt_sync.c
+++ b/src/btree/bt_sync.c
@@ -101,6 +101,103 @@ __sync_dup_walk(WT_SESSION_IMPL *session, WT_REF *walk, uint32_t flags, WT_REF *
 }
 
 /*
+ * __sync_ref_is_obsolete --
+ *     Return whether the ref is obsolete according to the newest stop time pair.
+ */
+static bool
+__sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
+{
+    WT_ADDR *addr;
+    WT_CELL_UNPACK vpack;
+    WT_MULTI *multi;
+    WT_PAGE_MODIFY *mod;
+    wt_timestamp_t multi_newest_stop_ts;
+    uint64_t multi_newest_stop_txn;
+    uint32_t i;
+
+    /* Ignore root pages as they can never be deleted. */
+    if (__wt_ref_is_root(ref))
+        return (false);
+
+    /* Ignore internal pages, these are taken care of during reconciliation. */
+    if (!__wt_ref_is_leaf(session, ref))
+        return (false);
+
+    /* Check for the page obsolete, if the page is modified and reconciled. */
+    mod = (ref->page != NULL) ? (ref->page->modify) : NULL;
+    if (mod != NULL && mod->rec_result == WT_PM_REC_REPLACE)
+        return (__wt_txn_visible_all(
+          session, mod->mod_replace.newest_stop_txn, mod->mod_replace.newest_stop_ts));
+
+    multi_newest_stop_ts = WT_TS_NONE;
+    multi_newest_stop_txn = WT_TXN_NONE;
+    if (mod != NULL && mod->rec_result == WT_PM_REC_MULTIBLOCK) {
+        /* Calculate the max stop time pair by traversing all multi addresses. */
+        for (multi = mod->mod_multi, i = 0; i < mod->mod_multi_entries; ++multi, ++i) {
+            multi_newest_stop_ts = WT_MAX(multi_newest_stop_ts, multi->addr.newest_stop_ts);
+            multi_newest_stop_txn = WT_MAX(multi_newest_stop_txn, multi->addr.newest_stop_txn);
+        }
+        return (__wt_txn_visible_all(session, multi_newest_stop_txn, multi_newest_stop_ts));
+    }
+
+    /* Check if the page is obsolete using the page disk address. */
+    addr = ref->addr;
+    if (addr != NULL) {
+        if (!__wt_off_page(ref->home, addr)) {
+            __wt_cell_unpack(session, ref->home, (WT_CELL *)addr, &vpack);
+            return (__wt_txn_visible_all(session, vpack.newest_stop_txn, vpack.newest_stop_ts));
+        }
+        return (__wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts));
+    }
+
+    return (false);
+}
+
+/*
+ * __sync_ref_evict_or_mark_deleted --
+ *     Evict the in memory page or delete the on disk page.
+ */
+static int
+__sync_ref_evict_or_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
+{
+    /*
+     * Mark the page as deleted and also set the parent page as dirty. This is to ensure the parent
+     * page must be written during checkpoint and the child page discarded.
+     */
+    if (WT_REF_CAS_STATE(session, ref, WT_REF_DISK, WT_REF_LOCKED)) {
+        WT_REF_SET_STATE(ref, WT_REF_DELETED);
+        return (__wt_page_parent_modify_set(session, ref, true));
+    }
+
+    /* Evict the in-memory obsolete page */
+    WT_IGNORE_RET_BOOL(__wt_page_evict_urgent(session, ref));
+    return (0);
+}
+
+/*
+ * __sync_ref_int_obsolete_cleanup --
+ *     Traverse the internal page and identify the leaf pages that are obsolete and mark them as
+ *     deleted.
+ */
+static int
+__sync_ref_int_obsolete_cleanup(WT_SESSION_IMPL *session, WT_REF *parent)
+{
+    WT_PAGE_INDEX *pindex;
+    WT_REF *ref;
+    uint32_t slot;
+
+    WT_INTL_INDEX_GET(session, parent->page, pindex);
+    for (slot = 0; slot < pindex->entries; slot++) {
+        ref = pindex->index[slot];
+
+        if (__sync_ref_is_obsolete(session, ref))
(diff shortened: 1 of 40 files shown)


src/btree/bt_sync.c
```diff
@@ -227,6 +229,10 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
         __wt_gen_next_drain(session, WT_GEN_EVICT);
         btree->syncing = WT_BTREE_SYNC_RUNNING;
 
+        /* Add in lookaside reconciliation for standard files. */
+        if (!F_ISSET(btree, WT_BTREE_LOOKASIDE) && !WT_IS_METADATA(btree->dhandle))
+            rec_flags |= WT_REC_LOOKASIDE;
```
agorrod: This flag is actually saying "is this the lookaside or metadata" - could we either skip using a flag, or call the flag something more in line with that?

src/reconcile/rec_visibility.c
```diff
@@ -408,11 +408,18 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+     * we're doing checkpoint reconciliation, and the update doesn't have any further updates that
+     * need to be written to the history store, skip saving the update as saving the update will
+     * cause reconciliation to think there is work that needs to be done when there might not be.
      */
-    if (!F_ISSET(r, WT_REC_EVICT))
+    if ((upd_select->upd != NULL && F_ISSET(upd_select->upd, WT_UPDATE_RESTORED_FROM_DISK)) ||
+      (F_ISSET(r, WT_REC_CHECKPOINT) &&
+          (!F_ISSET(r, WT_REC_LOOKASIDE) ||
+            (upd_select->upd == NULL || upd_select->upd->next == NULL))))
         goto check_original_value;
```
agorrod: It'd be nice to get rid of this goto. I wonder if we could change it to a flag instead? Or refactor this function so that we don't need a goto?

src/reconcile/rec_visibility.c
```diff
@@ -408,11 +408,18 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+    /*
+     * If the update we've selected was an on disk update restored from lookaside or the data store
+     * skip saving an update as it should always be the update at the end of the update chain, and
+     * therefore not have any updates that need to be written to the history store after it. If
+     * we're doing checkpoint reconciliation, and the update doesn't have any further updates that
+     * need to be written to the history store, skip saving the update as saving the update will
+     * cause reconciliation to think there is work that needs to be done when there might not be.
      */
-    if (!F_ISSET(r, WT_REC_EVICT))
+    if ((upd_select->upd != NULL && F_ISSET(upd_select->upd, WT_UPDATE_RESTORED_FROM_DISK)) ||
```
quchenhao: In upd_select WT_UPDATE_RESTORED_FROM_DISK is always not set. So we can ignore this case.

src/reconcile/rec_visibility.c
```diff
@@ -429,13 +436,24 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
      */
     if (!F_ISSET(r, WT_REC_LOOKASIDE | WT_REC_UPDATE_RESTORE))
         WT_ERR(__wt_set_return(session, EBUSY));
-    if (list_uncommitted && !F_ISSET(r, WT_REC_UPDATE_RESTORE))
+
+    /*
+     * If we have a list with an uncommitted update on it we still want to write out to the
+     * lookaside file in checkpoint.
+     */
+    if (list_uncommitted && !F_ISSET(r, WT_REC_UPDATE_RESTORE) && !F_ISSET(r, WT_REC_CHECKPOINT))
```
quchenhao: Why we don't do that for prepared updates?

quchenhao: OK, I think I missed this in my previous change. We should check prepared updates here as well.

src/reconcile/rec_visibility.c
```diff
@@ -429,13 +436,24 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
         WT_ERR(__wt_set_return(session, EBUSY));
 
     WT_ASSERT(session, r->max_txn != WT_TXN_NONE);
 
     WT_ERR(__rec_update_save(session, r, ins, ripcip, upd_select->upd, upd_memsize));
-    upd_select->upd_saved = true;
+
+    /*
```
quchenhao: Can you elaborate this case in the comment?

test/suite/test_checkpoint03.py
```diff
@@ -0,0 +1,106 @@
...
+        self.session.commit_transaction('commit_timestamp=4')
+
+        # Call checkpoint.
+        self.session.checkpoint()
+
+        # Validate that we wrote to lookaside, note that the lookaside statistic is not counting
+        # how many writes we did, just that we did write. Hence for multiple writes it may only
+        # increment a single time.
+        las_writes = self.get_stat(stat.conn.cache_write_lookaside)
+        self.assertEqual(1, las_writes)
```
quchenhao: Can this check pass deterministically? What if eviction also writes to lookaside?

test/suite/test_checkpoint03.py
```diff
@@ -0,0 +1,106 @@
...
+def timestamp_str(t):
+    return '%x' % t
+
+class test_checkpoint03(wttest.WiredTigerTestCase, suite_subprocess):
+    tablename = 'test_checkpoint03'
+    conn_config = 'statistics=(all)'
+    uri = 'table:' + tablename
+    session_config = 'isolation=snapshot, '
+
+    def get_stat(self, stat):
```
quchenhao: Can we verify the content in las? In WT-5284, I can open a cursor on las in a checkpoint and scan the table.

src/reconcile/rec_visibility.c
luke-pearson: Thats not entirely true, test_bug004 is an example where WT_UPDATE_RESTORED_FROM_DISK will be set in __rec_upd_select, however this check is still not needed as its actually covered by the next-> == NULL check.

quchenhao: Then that's weird. We shouldn't choose any version from las. I'll have a look at it later.

quchenhao: What I'm suggesting is that we can simply append the ext upd Alex C constructed to the end of the tombstone. Anyway, it's a very cheap operation. If the page get kicked out of the memory, both the tombstone and the appended upd get discarded. If we keep the page in memory, that is also OK. In this way, we can get rid of your special check.

src/reconcile/rec_visibility.c
luke-pearson: Can we merge list_prepared and list_uncommitted, in the scenario we add list_prepared to this list its used identically to list_uncommitted.

quchenhao: According to Vamsi, we probably need to insert prepared updates into las in the future. I think it's better to keep them separate?

src/reconcile/rec_visibility.c
luke-pearson: I removed the goto, I considered pulling the decision of saving the update into its own function but currently I feel like too many local variables are involved in the decision to make it worthwhile.

src/reconcile/rec_visibility.c
luke-pearson: If you append that update to the end of the tombstone you end up selecting the tombstone as the value to get written to the disk but it won't get squashed into the original update as the code that squashes it is the code that alex wrote. So I'm reasonably confused by your suggestion here. I did however remove the check as it wasn't required as it was being covered by the other check.

test/suite/test_checkpoint03.py
luke-pearson: Changed to a greater than or equal to check

test/suite/test_checkpoint03.py
luke-pearson: For now I'd rather not as thats exactly what the test you're writing is doing. This test is merely checking that we write to the expected locations.

src/reconcile/rec_visibility.c
luke-pearson: Unfortunately not without an extensive amount of time spent debugging test_las06, theres a case where eviction seems to have an insert list and still need to write content to lookaside. I attempted to elaborate a bit in the comment.

src/reconcile/rec_visibility.c
quchenhao: No, Alex's change ensures we select the newly appended upd instead of the tombstone. Actually, we can conveniently append a birthmark to the tombstone, i think.

src/btree/bt_sync.c
luke-pearson: I would prefer to keep the flag as is.

src/reconcile/rec_visibility.c
luke-pearson: Spoke offline, agreed to merge the two local values but keep the values in the reconciliation structure.

src/include/btmem.h
```diff
@@ -1114,8 +1114,11 @@ struct __wt_update {
      */
     volatile uint8_t prepare_state; /* prepare state */
 
-    /* Temp: Remove after integrating las cursors with btree cursors */
-    uint8_t ext;
+/* AUTOMATIC FLAG VALUE GENERATION START */
+#define WT_UPDATE_HISTORY_STORE 0x1u      /* Update has been written to history store. */
+#define WT_UPDATE_RESTORED_FROM_DISK 0x2u /* Update is temporary retrieved from LAS. */
+                                          /* AUTOMATIC FLAG VALUE GENERATION STOP */
+    uint32_t flags;
```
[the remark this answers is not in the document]
luke-pearson: I was following the convention of using a 32 bit variable for flags, but I agree as it currently is we can fit it into an 8 bit variable. As such I'll update it.

src/reconcile/rec_visibility.c
quchenhao: Created WT-5338 for this.

luke-pearson:
retest this please

src/btree/bt_sync.c
bvpvamsikrishna: I would like to update the comment with a TBD as " Remove the WT_BTREE_LOOKASIDE flag, and assert that no updates from lookaside are copied to lookaside recursively". A way to verify of design idea and to check the fallout if any.

bvpvamsikrishna:
@luke-pearson, I like the way this PR turned out into with minimal invasive changes.

luke-pearson:
retest this please

src/btree/bt_sync.c
keithbostic: I'd suggest initializing `rec_flags` in one spot -- it never changes, so I'd do it all here.

keithbostic: Can we switch to one model of "is this a special file"? I'd suggest `WT_IS_LOOKASIDE`.

src/btree/row_modify.c
```diff
@@ -273,6 +273,7 @@ __wt_update_alloc(WT_SESSION_IMPL *session, const WT_ITEM *value, WT_UPDATE **up
         memcpy(upd->data, value->data, value->size);
     }
     upd->type = (uint8_t)modify_type;
+    upd->flags = 0;
```
keithbostic: We just did a `calloc`, this isn't necessary?

src/include/btmem.h
```diff
@@ -1114,8 +1114,12 @@ struct __wt_update {
...
     volatile uint8_t prepare_state; /* prepare state */
 
-    /* Temp: Remove after integrating las cursors with btree cursors */
-    uint8_t ext;
+/* AUTOMATIC FLAG VALUE GENERATION START */
+#define WT_UPDATE_HISTORY_STORE 0x1u      /* Update has been written to history store. */
+#define WT_UPDATE_RESTORED_FROM_DISK 0x2u /* Update is temporary retrieved from LAS. */
+                                          /* AUTOMATIC FLAG VALUE GENERATION STOP */
+    uint8_t flags;
+    uint8_t unused; /* Unused value for structure padding. */
```
keithbostic: Why do we need structure padding?

src/reconcile/rec_col.c
```diff
@@ -1071,8 +1073,8 @@ __wt_rec_col_var(
                       (last.start_ts == start_ts && last.start_txn == start_txn &&
                         last.stop_ts == stop_ts && last.stop_txn == stop_txn)) &&
                   ((deleted && last.deleted) ||
-                      (!deleted && !last.deleted && last.value->size == size &&
-                        memcmp(last.value->data, data, size) == 0))) {
+                      (!deleted && !last.deleted && last.value->size != 0 &&
+                        last.value->size == size && memcmp(last.value->data, data, size) == 0))) {
```
keithbostic: I don't understand this change, and there's an almost identical bit of code 100 lines earlier in the function that didn't have to change, which concerns me. The commit was `clang sanitiser issuer`, which doesn't help much. Can you please explain what's going on?

src/reconcile/rec_visibility.c
```diff
@@ -397,47 +396,65 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+             * We are attempting eviction with changes that are not yet stable (i.e. globally
+             * visible). There are two ways to continue, the save/restore eviction path or the
+             * lookaside table eviction path. Both cannot be configured because the paths track
+             * different information. The update/restore path can handle uncommitted changes, by
+             * evicting most of the page and then creating a new, smaller page to which we re-attach
+             * those changes. Lookaside eviction writes changes into the lookaside table and
+             * restores them on demand if and when the page is read back into memory.
+             *
+             * Both paths are configured outside of reconciliation: the save/restore path is the
+             * WT_REC_UPDATE_RESTORE flag, the lookaside table path is the WT_REC_LOOKASIDE flag.
```
keithbostic: This comment needs thought, it's not necessarily eviction at this point.

src/cache/cache_las.c
```diff
@@ -796,6 +796,9 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
             WT_ERR(__las_insert_record(session, cursor, btree_id, key, upd, &stop_ts, &stop_txnid));
             ++insert_cnt;
 
+            /* Flag the update as now in the lookaside file. */
+            F_SET(upd, WT_UPDATE_HISTORY_STORE);
```
quchenhao: I think we can't set the flag here as the transaction hasn't been committed. If we roll back, we will lose the updates.

quchenhao: And I think we only need to mark the newest version in lookaside with that flag.

src/reconcile/rec_col.c
luke-pearson: So there was a clang sanitizer issue flagged with this specific if statement as a result of the changes on this branch. There is a follow up ticket to address this and determine what exactly is triggering the sanitizer.

src/btree/row_modify.c
luke-pearson: Oh good point, thanks.

src/btree/bt_sync.c
luke-pearson: Added WT_IS_LOOKASIDE macro.

src/include/btmem.h
luke-pearson: @hbokhari suggested this change as it pads out the structure to be a multiple of 4 bytes. Could you comment on this @hbokhari?

luke-pearson: @kommiharibabu suggested this change as it pads the structure out to be a multiple of 4 bytes. Could you answer Keith's question, @kommiharibabu?

kommiharibabu: The change was suggested to use a 1-byte flag and a 1-byte used variable to make sure that `upd->data` variable gets aligned to the multiple of 4 bytes.

keithbostic: @kommiharibabu, why do you want `upd.data` to be aligned to a multiple of 4 bytes?

src/reconcile/rec_col.c
keithbostic: If you have reviewed the code and confirmed that checking `last.value.size != 0` does not change the RLE matching functionality of column-store, it's fbm.

src/include/btmem.h
kommiharibabu: During the discussion of changing the `flags` variable from `uint32_t` to `uint8_t`, the padding discussion arose as `upd.data` is not aligned while discussing it with Vamsi, so the suggestion of adding `unused` variable has popped up. 

In the code, `upd.data` is mostly used to assign the address. Accessing the `upd.data` needs more instructions if it is not aligned?

src/reconcile/rec_col.c
luke-pearson: Going to review as part of WT-5341.

src/include/btmem.h
keithbostic: Is it any slower to access a single byte if the address isn't aligned? Since we're not doing multi-byte instructions, does it matter?

src/btree/bt_sync.c
```diff
@@ -325,6 +325,16 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
 
+        /*
+         * Add in lookaside reconciliation for standard files.
+         *
+         * FIXME-PM-1521: Remove the WT_BTREE_LOOKASIDE check, and assert that no updates from
+         * lookaside are copied to lookaside recursively.
+         */
+        rec_flags = WT_REC_CHECKPOINT;
+        if (!WT_IS_LOOKASIDE(btree) && !WT_IS_METADATA(btree->dhandle))
+            rec_flags |= WT_REC_LOOKASIDE;
```
keithbostic: The `is_las` assignment could use `WT_IS_LOOKASIDE`.

Additionally, can you review & cleanup explicit checks of `WT_BTREE_LOOKASIDE` in other parts of the tree?

src/cache/cache_las.c
```diff
@@ -683,7 +684,7 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
          * and start timestamp WT_TS_NONE, it is simply ignored.
          */
         for (; upd != NULL; upd = upd->next) {
-            if (upd->txnid == WT_TXN_ABORTED)
+            if (upd->txnid == WT_TXN_ABORTED || F_ISSET(upd, WT_UPDATE_HISTORY_STORE))
```
quchenhao: Can you remove this line of change for now? I'll handle it in a later change. It is not correct here.

src/cache/cache_las.c
```diff
@@ -755,6 +753,9 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
                 prev_full_value->size = prev_upd->size;
             }
 
+            /* Flag the update as now in the lookaside file. */
+            F_SET(upd, WT_UPDATE_HISTORY_STORE);
```
quchenhao: We should set the flag after we insert it to las.

src/include/btmem.h
kommiharibabu: @keithbostic, Thanks for the explanation.
Most of the places where the `upd.data` is passed between the functions, it is passed via `void *` pointer, so there is no access to the data. And cases where there is a need for data access, the data is `memcpy` into the aligned address variable before it is accessed.

>    
    /* Get the number of modify entries. */
    p = modify;
    memcpy(&tmp, p++, sizeof(size_t));
    nentries = (int)tmp;

The above code in `__wt_modify_idempotent()` function, where the `upd.data` address accessed by copying into a alinged variable.

Already the existing code in the `WiredTiger` already taken care of accessing the `upd.data` as an unaligned address whenever there is a need. So there is no need of doing structure padding that increases the memory size.

src/btree/bt_sync.c
luke-pearson: Did it for all checks except one as it was more readable to leave it for that one check.

luke-pearson:
retest this please

luke-pearson:
retest this please

PR #5067, opened 2019-12-20
WT-5004 Migrate Jenkins “wiredtiger-test-checkpoint-stress” job to Evergreen

raviprakashgiri29:
Migrate Jenkins “wiredtiger-test-checkpoint-stress” job to Evergreen

Added task level and command level timeout configuration to run for 24 hours max.


raviprakashgiri29:
Evergreen patch build to run 10 processes twice - Took ~ 19 hours

https://evergreen.mongodb.com/task/wiredtiger_ubuntu1804_checkpoint_stress_test_patch_2d2e2854f174d2e04e7281150ceda5f60ecbabdb_5dfb0cc41e2d175e1b323443_19_12_19_05_38_36#/log/T

Evergreen patch build to run 10 processes once - Took ~ 10 hours

https://evergreen.mongodb.com/task/wiredtiger_ubuntu1804_checkpoint_stress_test_patch_2d2e2854f174d2e04e7281150ceda5f60ecbabdb_5dfb100c3627e0398e9c765b_19_12_19_05_52_26

lukech:
Thanks Ravi. 19 hrs sounds good enough to me.

test/evergreen.yml
```diff
@@ -278,6 +278,27 @@ functions:
...
+      script: |
+        set -o errexit
+        set -o verbose
+
+        export WIREDTIGER_CONFIG='checkpoint_sync=0,transaction_sync=(method=none)'
+        CMD='./t -h WT_TEST.$i.$t -t r -r 2 -W 3 -n 1000000 -k 1000000 -C "cache_size=100MB"'
+
+        for i in $(seq ${times|1}); do
+          for t in $(seq ${no_of_procs|1}}); do
+            eval nohup $CMD > nohup.out.$i.$t 2>&1 &
```
lukech: I like this line that prevents log file overwriting across test runs :)

test/evergreen.yml
```diff
@@ -278,6 +278,27 @@ functions:
...
+          done
+
+          for t in $(seq ${no_of_procs|1}); do
+            wait -n || exit $?
```
lukech: I had a vague memory this line needs bash to run properly?

test/evergreen.yml
```diff
@@ -1952,6 +1973,20 @@ tasks:
           format_test_script_args: -b "SEGFAULT_SIGNALS=all catchsegv ./t" -S
           times: 16
 
+  - name: checkpoint-stress-test
+    commands:
+      - command: timeout.update
+        params:
+          # set 24 hour timeout for the task and command
+          exec_timeout_secs: 86400
+          timeout_secs: 86400
```
lukech: Thanks for figuring out the way to customize both timeout settings. 
It would be nice to remind Evergreen team to add this command into the wiki page.

test/evergreen.yml
```diff
@@ -1952,6 +1973,20 @@ tasks:
...
+      - func: "get project"
+      - func: "compile wiredtiger with builtins"
+      - func: "checkpoint stress test"
+        vars:
+          times: 2
```
lukech: Could a comment be added to explain the rationale of the value setting here?

test/evergreen.yml
raviprakashgiri29: This is actually to wait for the processes which was triggered with nohup command in the background.

test/evergreen.yml

test/evergreen.yml
```diff
@@ -278,6 +278,27 @@ functions:
...
 
+  "checkpoint stress test":
+    command: shell.exec
+    params:
+      working_dir: "wiredtiger/test/format"
+      script: |
+        set -o errexit
+        set -o verbose
+
+        export WIREDTIGER_CONFIG='checkpoint_sync=0,transaction_sync=(method=none)'
```
hbokhari: Not sure but can we avoid this export and run set something like: `CMD='WIREDTIGER_CONFIG='checkpoint_sync=0,transaction_sync=(method=none)' ./t -h WT_TEST.$i.$t -t r -r 2 -W 3 -n 1000000 -k 1000000 -C "cache_size=100MB"'`

It is just style thing. We have avoiding export for most of the other tasks as well.

raviprakashgiri29: I tried to run the command by including WIREDTIGER_CONFIG in the CMD variable but it fails for some reason (brackets in the config value may be). Below is the failed patch build. 

https://evergreen.mongodb.com/task/wiredtiger_ubuntu1804_checkpoint_stress_test_patch_9a70a0cad47875afe7631318fc06182f3add2b47_5dfc4c19a4cf473bb06368d7_19_12_20_04_21_03

hbokhari: Never mind. I guess something in the string need to be escaped but I am okay with the existing config.

test/evergreen.yml
raviprakashgiri29: > Thanks for figuring out the way to customize both timeout settings.
> It would be nice to remind Evergreen team to add this command into the wiki page.

Yeah, sure. I will do that.

test/evergreen.yml
lukech: I actually meant to explain why 2 was selected e.g. the measured runtime of 2 runs is close to 24hrs to simulate continuously running job in Jenkins

PR #5068, opened 2019-12-20
WT-5300 history store internal page cleanup

kommiharibabu:
Whenever checkpoint is flushing the internal page, it scans the
entire page to find out the obsolete child pages and mark them
as obsolete, during the reconciliation of the internal page, all
the child pages are removed. If the internal page is a clean page
and empty, it will marked as deleted.
    
Apart from the above, the tree walk reads the internal pages into
memory if they don't exist in memory and they are from history
store for the obsolete page removal.


kommiharibabu:
test this please

src/btree/bt_sync.c
```diff
@@ -100,6 +100,88 @@ __sync_dup_walk(WT_SESSION_IMPL *session, WT_REF *walk, uint32_t flags, WT_REF *
...
+    WT_ADDR *addr;
+    bool obsolete;
+
+    obsolete = false;
+
+    addr = ref->addr;
+    if (addr && __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts))
+        obsolete = true;
+
+    return obsolete;
```
ddanderson: I find this code clearer without the `obsolete` boolean flag.  E.g.
```
  return (addr && __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts));
```

src/btree/bt_sync.c
```diff
@@ -100,6 +100,88 @@ __sync_dup_walk(WT_SESSION_IMPL *session, WT_REF *walk, uint32_t flags, WT_REF *
...
+/*
+ * __wt_ref_mark_deleted --
+ *     Lock the ref and change state as deleted. Also mark the parent page as dirty.
+ */
+static int
+__wt_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
+{
+    uint32_t previous_state;
+
+    /* Ignore root pages as they never be deleted */
```
ddanderson: "...as they are never..."

test/suite/test_gc01.py
```diff
@@ -0,0 +1,185 @@
...
+        # Pin oldest and stable to timestamp 1.
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(1) +
+            ',stable_timestamp=' + timestamp_str(1))
+
+        bigvalue = "aaaaa" * 100
+        bigvalue2 = "ddddd" * 100
+        self.large_updates(uri, bigvalue, ds, nrows, 10)
+
+        # Check that all updates are seen
+        #self.check(bigvalue, uri, nrows, 20)
```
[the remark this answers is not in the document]
kommiharibabu: Thanks for the suggestion, I added them to the tests.

test/suite/test_gc01.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Load a slight modification with a later timestamp.
+        cursor = self.session.open_cursor(uri)
+        self.session.begin_transaction()
+        for i in range(1, nrows):
+            cursor.set_key(i)
+            mods = [wiredtiger.Modify('A', 10, 1)]
+            self.assertEqual(cursor.modify(mods), 0)
+        self.session.commit_transaction('commit_timestamp=' + timestamp_str(210))
+
```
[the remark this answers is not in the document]
kommiharibabu: I reorganized the tests. I will do class level reorganization during the full test ticket handling.

test/suite/test_gc01.py
```diff
@@ -0,0 +1,185 @@
...
+import time
+from helper import copy_wiredtiger_home
+import unittest, wiredtiger, wttest
+from wtdataset import SimpleDataSet
+
+def timestamp_str(t):
+    return '%x' % t
+
+# test_gc01.py
+# Test that checkpint cleans the obsolete lookaside pages.
```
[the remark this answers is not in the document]
kommiharibabu: Fixed.

src/btree/bt_sync.c
kommiharibabu: This function adds some more additional conditions later to handle in-memory pages, so I left it as it is currently.

src/btree/bt_sync.c

kommiharibabu:
@keithbostic, @sueloverso, Please review the next set of checkpoint garbage collection changes that does the internal page clean up.

src/btree/bt_sync.c
```diff
@@ -143,6 +143,42 @@ __sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
...
+static int
+__sync_ref_int_obsolete_cleanup(WT_SESSION_IMPL *session, WT_REF *intref)
+{
+    WT_PAGE_INDEX *pindex;
+    WT_REF *ref;
+    uint32_t slot;
+
+    pindex = NULL;
+
+    pindex = WT_INTL_INDEX_GET_SAFE(intref->page);
```
keithbostic: We're initializing `pindex` twice.

keithbostic: I'm confused why we're using `WT_INTL_INDEX_GET_SAFE` instead of `WT_INTL_INDEX_GET`, what makes this call safe?

src/btree/bt_sync.c
```diff
@@ -143,6 +143,42 @@ __sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    for (slot = 0; slot < pindex->entries; slot++) {
+        /*
+         * Get a reference, setting the reference hint if it's wrong (used when we continue the
+         * walk). We don't always update the hints when splitting, it's expected for them to be
+         * incorrect in some workloads.
+         */
+        ref = pindex->index[slot];
+        if (ref->pindex_hint != slot)
+            ref->pindex_hint = slot;
```
keithbostic: I don't understand why we're resetting the reference hints in this loop? I mean, it's not wrong, but I'm not sure why we want to do that work here?

src/btree/bt_sync.c
```diff
@@ -143,6 +143,42 @@ __sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
...
+
+        /* Skip the deleted child pages. */
+        if (ref->state == WT_REF_DELETED)
+            continue;
+
+        if (__sync_ref_is_obsolete(session, ref))
+            WT_RET(__sync_ref_mark_deleted(session, ref));
```
keithbostic: Doesn't this need to go recursive in order to handle internal pages that are the parents of a set of internal pages? That is, why can we assume we'll visit every internal page in the system that is a parent of a leaf page if we only visit the internal pages we encounter in a specific in-cache walk?

src/btree/bt_read.c
```diff
@@ -700,6 +700,9 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
             }
             goto read;
         case WT_REF_DISK:
+            /* Read internal pages if requested */
+            if (LF_ISSET(WT_READ_INT_PAGE) && ((WT_ADDR *)ref->addr)->type == WT_ADDR_INT)
+                goto read;
```
keithbostic: Why can't we be here with `WT_REF.addr` pointing to an on-page cell?

src/btree/bt_sync.c

kommiharibabu: Thanks, now I understood. I used the `WT_WITH_PAGE_INDEX` to protect it.

src/btree/bt_sync.c
kommiharibabu: Already the tree walk function has taken care of resetting the slot that are not matching, this code is not required. Removed.

src/btree/bt_sync.c
kommiharibabu: The tree walk function returns all the pages that are in memory. By reading all the internal pages that are part of the history store btree even if they are not in memory will get all the internal pages one by one and in the outer function, we traverse each page for finding out the obsolete pages. This way we ensure that all the internal pages are traversed.

src/btree/bt_read.c
kommiharibabu: I didn't clearly understand the comment. Maybe I can explain what I am thinking, let me know if any problem.

The above code is to ensure to read the internal pages that are not in memory for the history store.
This is done when the parent page is getting processed, all the `pindex` entries are processed one by one. Only the pages that are not in memory will come to this case to let them load only when they are internal pages.

src/btree/bt_sync.c
keithbostic: I get it now, I missed what you were trying to do when I first looked at it.

I'd suggest creating a function like `__ref_is_leaf()` that returns if the page is an internal page.

keithbostic: I'd also suggest `WT_READ_CACHE_LEAF` which is a subset of `WT_READ_CACHE`?

src/btree/bt_read.c
keithbostic: Got it, this makes sense now. Again, I'd suggest creating a function like `__ref_is_leaf()` that returns if the page is an internal page.

src/btree/bt_sync.c
kommiharibabu: Used `WT_READ_CACHE_LEAF`.

src/btree/bt_read.c
kommiharibabu: Thanks, used `__ref_is_leaf()` function to find out the page is leaf or internal.

kommiharibabu:
@keithbostic, I fixed all the comments raised, can you check the changes once again.

src/btree/bt_sync.c
```diff
@@ -143,6 +143,33 @@ __sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    WT_REF *ref;
+    uint32_t slot;
+
+    WT_INTL_INDEX_GET(session, intref->page, pindex);
+    for (slot = 0; slot < pindex->entries; slot++) {
+        ref = pindex->index[slot];
+
+        /* Skip the deleted child pages. */
+        if (ref->state == WT_REF_DELETED)
+            continue;
```
keithbostic: I think it's odd that we test the case `WT_REF.state == WT_REF_DELETED` here, but other tests like `ref->page->modify->rec_result != WT_PM_REC_EMPTY` appear in `__sync_ref_is_obsolete()`. Why is that? Would it make more sense to put all the tests for "don't mark this page deleted" in one place?

src/btree/bt_sync.c
```diff
@@ -382,6 +419,18 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
+            if (is_las && page->modify != NULL && page->modify->rec_result == WT_PM_REC_EMPTY) {
+                /*
+                 * The duplicate tree walk code expects the ref state to be in memory. We need to
+                 * get the duplicate tree walk pointer before marking the current tree page as
+                 * deleted.
+                 */
+                WT_ERR(__sync_dup_walk(session, walk, flags, &prev));
+                WT_ERR(__sync_ref_mark_deleted(session, walk));
+                goto skipwalk;
+            }
```
keithbostic: If we reconcile a history page and there are no records to write, why does that mean we should mark it for deletion? Imagine a page with 5 records, none of which need to be written, so we get a `WT_PM_REC_EMPTY` result, but after the reconciliation completes, additional records are inserted. Is there something about the insert pattern of the history file that makes that impossible?

src/btree/bt_sync.c
```diff
@@ -143,6 +143,33 @@ __sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
...
+
+        /* Skip the deleted child pages. */
+        if (ref->state == WT_REF_DELETED)
+            continue;
+
+        if (__sync_ref_is_obsolete(session, ref))
+            WT_RET(__sync_ref_mark_deleted(session, ref));
+    }
+
+    return (0);
```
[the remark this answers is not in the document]
kommiharibabu: Changed the condition in `__sync_ref_mark_deleted` function to not let the in-memory pages to be marked as deleted. Later in another ticket, these pages will be added to `urgent eviction queue`. So all the internal pages are in memory, so we don't need any special handling for them.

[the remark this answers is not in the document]

src/btree/bt_sync.c
kommiharibabu: I understood the race condition that is possible in marking deleted of an in-memory page when it is used for insertion. So marking to be deleted after reconciliation is removed. Take care of these in-memory pages when we traversing the parent page.

src/btree/bt_sync.c
```diff
@@ -122,13 +122,12 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
-    /* Ignore root pages as they are never be deleted */
+    /* Ignore root pages as they can never be deleted. */
     if (__wt_ref_is_root(ref))
         return (0);
 
-    /* Guard against marking a page to be deleted when its reconciliation state is not empty. */
-    if (ref->page != NULL && ref->page->modify != NULL &&
-      ref->page->modify->rec_result != WT_PM_REC_EMPTY)
+    /* Guard against marking an in-memory page to be deleted. */
+    if (ref->page != NULL)
```
keithbostic: I'd go with `WT_REF.state == WT_REF_DISK`. You want something like this:

```
if (ref->state != WT_REF_DISK || !WT_REF_CAS_STATE(session, ref, WT_REF_DISK, WT_REF_LOCKED))
        return (0);
```

Otherwise, you can do the test, find `WT_REF.state == WT_REF_DISK`, then swap it to `WT_REF_LOCKED`, but the page went to `WT_REF_MEM` between the test and the lock.

src/btree/bt_sync.c
kommiharibabu: I removed this case entirely while processing the internal pages as this case never happens now because we never mark the empty pages as deleted after reconciliation.

src/btree/bt_sync.c
```diff
@@ -143,6 +142,29 @@ __sync_ref_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    WT_PAGE_INDEX *pindex;
+    WT_REF *ref;
+    uint32_t slot;
+
+    WT_INTL_INDEX_GET(session, intref->page, pindex);
+    for (slot = 0; slot < pindex->entries; slot++) {
+        ref = pindex->index[slot];
+
+        if (__sync_ref_is_obsolete(session, ref))
+            WT_RET(__sync_ref_mark_deleted(session, ref));
```
keithbostic: `__sync_ref_is_obsolete()` should ignore internal pages.

I'd move the rest of the tests there, too. It makes more sense that a function named "is obsolete" checks and returns false than a function named "mark deleted".

src/btree/bt_sync.c
kommiharibabu: Thanks, Fixed accordingly.

src/btree/bt_sync.c
kommiharibabu: Done.

src/btree/bt_sync.c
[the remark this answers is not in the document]
kommiharibabu: As we discussed, there is no need to get the on-page cell.

src/btree/bt_read.c
```diff
@@ -700,6 +700,9 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
             }
             goto read;
         case WT_REF_DISK:
+            /* Read internal pages if requested */
+            if (LF_ISSET(WT_READ_CACHE_LEAF) && !__wt_ref_is_leaf(session, ref))
+                goto read;
```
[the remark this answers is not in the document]
kommiharibabu: Changed as per the above suggestion.

src/btree/bt_sync.c
```diff
@@ -109,8 +109,19 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
-    addr = ref->addr;
+    /* Ignore root pages as they can never be deleted. */
+    if (__wt_ref_is_root(ref))
+        return false;
+
+    /* Guard against marking an in-memory page to be deleted. */
+    if (ref->state != WT_REF_DISK || !WT_REF_CAS_STATE(session, ref, WT_REF_DISK, WT_REF_LOCKED))
+        return false;
+
+    /* Ignore internal pages, these are taken care during reconciliation. */
```
keithbostic: Comment is missing a word, "these are taken care of".

src/btree/bt_sync.c
```diff
@@ -285,6 +316,12 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
             if (walk == NULL)
                 break;
 
+            /* Traverse through the internal page for obsolete child pages. */
+            if (is_las && WT_PAGE_IS_INTERNAL(walk->page)) {
+                WT_WITH_PAGE_INDEX(session, ret = __sync_ref_int_obsolete_cleanup(session, walk));
+                WT_ERR(ret);
+            }
+
```
keithbostic: The chunk of code farther down:
```
/*  
 * Check whether the tree is lookaside btree and verify whether the page contents
 * are obsolete or not by checking the visibility of the page stop time pair.
 */  
if (is_las && __sync_ref_is_obsolete(session, walk)) {
    /*  
     * The duplicate tree walk code expects the ref state to be in memory. We need
     * to get the duplicate tree walk pointer before marking the current tree page
     * as deleted.
     */  
    WT_ERR(__sync_dup_walk(session, walk, flags, &prev));

    /* Mark the page as deleted */
    WT_ERR(__sync_ref_mark_deleted(session, walk));
                                                
    /*  
     * The duplicate tree walk pointer, prev, was obtained before we marked the page
     * as deleted. Continue the rest of the tree walk using that pointer.
     */  
    goto skipwalk;
}   
continue;
```
Can this still happen?

If it's an internal page, we can't mark it deleted, if it's a leaf page, by definition it has to be in memory, otherwise walk wouldn't have returned it to us?

src/btree/bt_sync.c
```diff
@@ -109,8 +109,19 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    if (__wt_ref_is_root(ref))
+        return false;
+
+    /* Guard against marking an in-memory page to be deleted. */
+    if (ref->state != WT_REF_DISK || !WT_REF_CAS_STATE(session, ref, WT_REF_DISK, WT_REF_LOCKED))
+        return false;
+
+    /* Ignore internal pages, these are taken care during reconciliation. */
+    if (!__wt_ref_is_leaf(session, ref))
+        return false;
```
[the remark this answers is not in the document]
kommiharibabu: Thanks. Simplified as per the above. I don't clearly understand `UNPACK CELL` requirement here to refer the stop time pair.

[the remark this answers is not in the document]
kommiharibabu: Corrected.

src/btree/bt_sync.c
kommiharibabu: Corrected.

src/btree/bt_sync.c
kommiharibabu: No, this code is not required now. But this code comes back again when we are handling the in-memory pages that are needed to be added to urgent eviction queue.

src/btree/bt_sync.c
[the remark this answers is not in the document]
kommiharibabu: fixed.

src/btree/bt_sync.c
```diff
@@ -109,8 +109,21 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    if (ref->state != WT_REF_DISK)
+        return (false);
+
+    /* Ignore internal pages, these are taken care of during reconciliation. */
+    if (!__wt_ref_is_leaf(session, ref))
+        return (false);
+
+    /* DON'T WE HAVE TO POTENTIALLY UNPACK THE CELL? */
+
+    addr = ref->addr;
```
keithbostic: I think this looks something like:
```
addr = ref->addr;
if (__wt_off_page(parent->page, addr))
    return (false);
__wt_cell_unpack(session, parent->page, addr, vpack);
return (__wt_txn_visible_all(session, vpack->newest_stop_txn, vpack->newest_stop_ts));
```
In short, we get the `WT_REF.addr` and check to make sure it points into the page (if it doesn't, we're done, this isn't a disk reference). Then, we unpack the on-page cell and use its values.

This is a little tricky because we're not holding any kind of lock on the `WT_REF` structure, so it could be changed out from under us while we're looking at it.

src/btree/bt_sync.c
```diff
@@ -122,27 +135,41 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+
+    WT_INTL_INDEX_GET(session, intref->page, pindex);
+    for (slot = 0; slot < pindex->entries; slot++) {
+        ref = pindex->index[slot];
+
+        if (__sync_ref_is_obsolete(session, ref))
+            WT_RET(__sync_ref_mark_deleted(session, ref));
+    }
+
+    return (0);
```
keithbostic: I'd suggest `parent` instead of `intref`, that's usually how we refer to the relationship.

src/btree/bt_sync.c
kommiharibabu: Thanks for the details. As per my understanding, I updated the code to unpack the cell and find out the details. Please review my understanding

src/btree/bt_sync.c

src/btree/bt_sync.c
```diff
@@ -108,11 +108,26 @@ static bool
...
+        return (false);
 
-    return (
-      addr != NULL && __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts));
+    WT_ORDERED_READ(addr, ref->addr);
+    if (!__wt_off_page(ref->home, addr)) {
+        __wt_cell_unpack(session, ref->home, (WT_CELL *)addr, &vpack);
+        return (__wt_txn_visible_all(session, vpack.newest_stop_txn, vpack.newest_stop_ts));
+    } else
+        return (__wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts));
```
keithbostic: There's no need for an `else` after a `return`.

src/btree/bt_sync.c
```diff
@@ -122,27 +137,41 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
-    if (ref->page != NULL && ref->page->modify != NULL &&
-      ref->page->modify->rec_result != WT_PM_REC_EMPTY)
-        return (0);
-
     /*
      * Mark the page as deleted and also set the parent page as dirty. This is to ensure when the
      * parent page is checkpointing, the empty child page will be cleaned.
      */
-    if (WT_REF_CAS_STATE(session, ref, ref->state, WT_REF_LOCKED)) {
+    if (WT_REF_CAS_STATE(session, ref, WT_REF_DISK, WT_REF_LOCKED)) {
```
keithbostic: I think the sentence `This is to ensure when the parent page is checkpointing, the empty child page will be cleaned.` is confusing, what do you think of `This is to ensure the parent page must be written during checkpoint and the child page discarded.`

kommiharibabu: Used the provided sentence.

src/btree/bt_sync.c
kommiharibabu: fixed.

src/btree/bt_sync.c
```diff
@@ -325,6 +333,17 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
              */
             page = walk->page;
 
+            /*
+             * Skip clean pages, but need to make sure maximum transaction ID is always updated.
+             */
+            if (!__wt_page_is_modified(page)) {
+                if (((mod = page->modify) != NULL) && mod->rec_max_txn > btree->rec_max_txn)
+                    btree->rec_max_txn = mod->rec_max_txn;
+                if (mod != NULL && btree->rec_max_timestamp < mod->rec_max_timestamp)
```
sueloverso: Minor nit: Is there a reason in the first `if` statement you did `mod->rec_max_txn > btree->rec_max_txn` and in the 2nd `if` statement you did `btree->rec_max_timestamp < mod->rec_max_timestamp`? It requires thought about why they're different directions? I suggest making them consistent with each other one way or the other.

PR #5069, opened 2019-12-20
WT-5333 Simplify txn read and remove invisible birthmark optimisation

quchenhao:
Now the algorithm can be simplified to first walk the update chain,
then the history store, and return the first visibile update.

### DIFF
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -716,20 +716,23 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT
     *updp = NULL;
     for (; upd != NULL; upd = upd->next) {
         /* Skip reserved place-holders, they're never visible. */
-        if (upd->type != WT_UPDATE_RESERVE) {
-            upd_visible = __wt_txn_upd_visible_type(session, upd);
-            if (upd_visible == WT_VISIBLE_TRUE)
-                break;
-            if (upd_visible == WT_VISIBLE_PREPARE)
-                return (WT_PREPARE_CONFLICT);
+        if (upd->type == WT_UPDATE_RESERVE)
+            continue;
+        upd_visible = __wt_txn_upd_visible_type(session, upd);
+        if (upd_visible == WT_VISIBLE_TRUE) {
+            if (upd->type != WT_UPDATE_BIRTHMARK)
+                *updp = upd;
+            return (0);
         }
+        if (upd_visible == WT_VISIBLE_PREPARE)
+            return (WT_PREPARE_CONFLICT);
     }
 
-    /* Cannot find the update in memory. Try lookaside. */
-    if (upd == NULL && __wt_page_las_active(session, cbt->ref))
+    /* If there's no visible update in memory, check the history file. */
+    if (__wt_page_las_active(session, cbt->ref))
         WT_RET_NOTFOUND_OK(__wt_find_lookaside_upd(session, cbt, &upd, false));
 
-    *updp = upd == NULL || upd->type == WT_UPDATE_BIRTHMARK ? NULL : upd;
+    *updp = upd;
     return (0);
 }
 
--- a/src/include/txn.i
+++ b/src/include/txn.i
@@ -728,7 +728,7 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT
             return (WT_PREPARE_CONFLICT);
     }
 
-    /* If there's no visible update in memory, check the history file. */
+    /* If there's no visible update in the update chain, check the lookaside file. */
     if (__wt_page_las_active(session, cbt->ref))
         WT_RET_NOTFOUND_OK(__wt_find_lookaside_upd(session, cbt, &upd, false));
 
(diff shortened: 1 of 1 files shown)


src/include/txn.i
```diff
@@ -711,14 +711,14 @@ __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
 static inline int
 __wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT_UPDATE **updp)
 {
-    static WT_UPDATE tombstone = {.txnid = WT_TXN_NONE, .type = WT_UPDATE_TOMBSTONE};
-    WT_DECL_RET;
-    WT_UPDATE *tmp_upd;
+    /* FIXME-PM1521: To be Removed by WT-5282 */
```
luke-pearson: Is it not better if you just remove the code in this ticket anyway?

quchenhao: I'd like to. Vamsi has different opinion that.

luke-pearson: @bvpvamsikrishna please confirm.

bvpvamsikrishna: Yes, I asked Chenhao to comment out the code, for birthmarks instead of deleting it.

bvpvamsikrishna: Maybe I am thinking too much, @quchenhao  delete the code instead of commenting out.

quchenhao: deleted

src/include/txn.i
```diff
@@ -727,29 +723,11 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT
...
-                __wt_free_update_list(session, &tmp_upd);
-            }
-        }
     }
 
-    if (upd == NULL && skipped_birthmark)
-        upd = &tombstone;
+    /* Cannot find the update in memory. Try lookaside. */
+    if (upd == NULL && __wt_page_las_active(session, cbt->ref))
+        WT_RET_NOTFOUND_OK(__wt_find_lookaside_upd(session, cbt, &upd, false));
```
keithbostic: I think the code flow is more confusing (and the `Cannot find the update in memory` comment is now incorrect, as is the function description).

What would you think of something like:
```
/*
 * __wt_txn_read --
 *     Get the first visible update in a list or the history file (or NULL if none are visible).
 */ 
static inline int
__wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT_UPDATE **updp)
{   
    WT_VISIBLE_TYPE upd_visible;
        
    *updp = NULL;

        /* Skip reserved place-holders, they're never visible. */
        if (upd->type == WT_UPDATE_RESERVE)   
                continue;
        upd_visible = __wt_txn_upd_visible_type(session, upd);
        if (upd_visible == WT_VISIBLE_TRUE) {
            if (upd->type == WT_UPDATE_BIRTHMARK)
                *updp = upd;
            return (0);
        }
        if (upd_visible == WT_VISIBLE_PREPARE)
            return (WT_PREPARE_CONFLICT);
    }
        
    /* If there's no visible update in memory, check the history file. */
    if (__wt_page_las_active(session, cbt->ref))
        WT_RET_NOTFOUND_OK(__wt_find_lookaside_upd(session, cbt, &upd, false));
    
    *updp = upd == NULL || upd->type == WT_UPDATE_BIRTHMARK ? NULL : upd;
    return (0);
}   
```

quchenhao: Changed as suggested. But I think the line *updp = upd changes the behaviour. It should be upd->type != WT_UPDATE_BIRTHMARK

src/include/txn.i
```diff
@@ -716,20 +716,23 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT
...
+                *updp = upd;
+            return (0);
         }
+        if (upd_visible == WT_VISIBLE_PREPARE)
+            return (WT_PREPARE_CONFLICT);
     }
 
-    /* Cannot find the update in memory. Try lookaside. */
-    if (upd == NULL && __wt_page_las_active(session, cbt->ref))
+    /* If there's no visible update in memory, check the history file. */
```
bvpvamsikrishna: Better to use lookaside file instead of history file as of now. We will change the comment also when we change the file name.
Update the comment from "If there's no visible update in memory" to "If there's no visible update in data store".

src/include/txn.i
```diff
@@ -711,47 +711,28 @@ __wt_txn_upd_visible(WT_SESSION_IMPL *session, WT_UPDATE *upd)
...
     }
 
-    if (upd == NULL && skipped_birthmark)
-        upd = &tombstone;
+    /* If there's no visible update in memory, check the history file. */
+    if (__wt_page_las_active(session, cbt->ref))
+        WT_RET_NOTFOUND_OK(__wt_find_lookaside_upd(session, cbt, &upd, false));
 
-    *updp = upd == NULL || upd->type == WT_UPDATE_BIRTHMARK ? NULL : upd;
+    *updp = upd;
```
bvpvamsikrishna: Please add an assert that upd is not a birthmark.

quchenhao: Added as suggested

src/include/txn.i
keithbostic: Yes, I agree -- in general, if I don't claim some code suggestion is tested, it almost certainly wasn't. :)

quchenhao: Just to confirm :)

src/include/txn.i
quchenhao: I think data store is inaccurate here as we are reading the update chain. Changed to update chain.

PR #5070, opened 2019-12-20
WT-5335 Read the start and stop time pairs back from cell

quchenhao:
This change read the time pairs we write to the cell back to memory when we append the on page value to the update chain. If the cell has a stop time pair, we need to append an extra tombstone before the on page value.

### DIFF
--- a/src/btree/bt_ret.c
+++ b/src/btree/bt_ret.c
@@ -94,12 +94,18 @@ __wt_value_return_buf(WT_CURSOR_BTREE *cbt, WT_REF *ref, WT_ITEM *buf)
     if (page->type == WT_PAGE_ROW_LEAF) {
         rip = &page->pg_row[cbt->slot];
 
-        /* Simple values have their location encoded in the WT_ROW. */
+        /*
+         * Simple values have their location encoded in the WT_ROW.
+         *
+         * Cannot check visibility here so there is chance some visibility issues not caught.
+         */
         if (__wt_row_leaf_value(page, rip, buf))
             return (0);
 
         /* Take the value from the original page cell. */
         __wt_row_leaf_value_cell(session, page, rip, NULL, &unpack);
+
+        WT_ASSERT(session, __wt_txn_visible(session, unpack.start_txn, unpack.start_ts));
         return (__wt_page_cell_data_ref(session, page, &unpack, buf));
     }
 
@@ -107,6 +113,8 @@ __wt_value_return_buf(WT_CURSOR_BTREE *cbt, WT_REF *ref, WT_ITEM *buf)
         /* Take the value from the original page cell. */
         cell = WT_COL_PTR(page, &page->pg_var[cbt->slot]);
         __wt_cell_unpack(session, page, cell, &unpack);
+
+        WT_ASSERT(session, __wt_txn_visible(session, unpack.start_txn, unpack.start_ts));
         return (__wt_page_cell_data_ref(session, page, &unpack, buf));
     }
 
--- a/src/btree/bt_ret.c
+++ b/src/btree/bt_ret.c
@@ -105,7 +105,8 @@ __wt_value_return_buf(WT_CURSOR_BTREE *cbt, WT_REF *ref, WT_ITEM *buf)
         /* Take the value from the original page cell. */
         __wt_row_leaf_value_cell(session, page, rip, NULL, &unpack);
 
-        WT_ASSERT(session, __wt_txn_visible(session, unpack.start_txn, unpack.start_ts));
+        /* FIXME-PM-1521: Temporarily disabled due to large number of failed tests */
+        // WT_ASSERT(session, __wt_txn_visible(session, unpack.start_txn, unpack.start_ts));
         return (__wt_page_cell_data_ref(session, page, &unpack, buf));
     }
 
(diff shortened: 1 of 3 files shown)


quchenhao:
retest this please

src/reconcile/rec_visibility.c
```diff
@@ -100,14 +86,23 @@ __rec_append_orig_value(WT_SESSION_IMPL *session, WT_PAGE *page, WT_INSERT *ins,
...
-    size = 0;      /* -Wconditional-uninitialized */
+    append = tombstone = NULL; /* -Wconditional-uninitialized */
+    size = 0;                  /* -Wconditional-uninitialized */
     if (unpack == NULL || unpack->type == WT_CELL_DEL)
         WT_RET(__wt_update_alloc(session, NULL, &append, &size, WT_UPDATE_TOMBSTONE));
     else {
         WT_RET(__wt_scr_alloc(session, 0, &tmp));
         WT_ERR(__wt_page_cell_data_ref(session, page, unpack, tmp));
         WT_ERR(__wt_update_alloc(session, tmp, &append, &size, WT_UPDATE_STANDARD));
+        append->start_ts = append->durable_ts = unpack->start_ts;
```
kommiharibabu: I don't think there is a need for a `tombstone` record for every append record. The `tombstone` is required only for history store records to represents the stop time pair of the update. Whereas for the data updates, the next update in the list represents in the stop time pair. The added code can be removed.

quchenhao: Added a comment

quchenhao:
retest this please

src/btree/bt_ret.c
```diff
@@ -94,19 +94,29 @@ __wt_value_return_buf(WT_CURSOR_BTREE *cbt, WT_REF *ref, WT_ITEM *buf)
...
+         * Simple values have their location encoded in the WT_ROW.
+         *
+         * Cannot check visibility here so there is chance some visibility issues not caught.
+         */
         if (__wt_row_leaf_value(page, rip, buf))
             return (0);
 
         /* Take the value from the original page cell. */
         __wt_row_leaf_value_cell(session, page, rip, NULL, &unpack);
+
```
quchenhao: Not sure if we need to check and return not found or onpage value here. From my understanding of the current search flow, we don't need to ATM, but the flow itself may need to be changed.

src/reconcile/rec_visibility.c
```diff
@@ -49,48 +49,34 @@ __rec_update_save(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, voi
...
     WT_DECL_RET;
-    WT_PAGE_MODIFY *mod;
-    WT_UPDATE *append, **upd_entry;
+    WT_UPDATE *append, *tombstone;
     size_t size;
 
-    mod = page->modify;
-    upd_entry = NULL;
+    WT_ASSERT(session, upd != NULL);
+    /* Done if at least one self-contained update is globally visible. */
```
keithbostic: There's no need for the `WT_ASSERT` -- we're about to use `upd`, and we'll drop core if it's NULL.

src/reconcile/rec_visibility.c
```diff
@@ -100,14 +86,32 @@ __rec_append_orig_value(WT_SESSION_IMPL *session, WT_PAGE *page, WT_INSERT *ins,
...
     else {
         WT_RET(__wt_scr_alloc(session, 0, &tmp));
         WT_ERR(__wt_page_cell_data_ref(session, page, unpack, tmp));
         WT_ERR(__wt_update_alloc(session, tmp, &append, &size, WT_UPDATE_STANDARD));
+        append->start_ts = append->durable_ts = unpack->start_ts;
+        append->txnid = unpack->start_txn;
+
+        /*
+         * We need to append a TOMBSTONE before the on page value if the onpage value has a valid
+         * stop pair.
```
keithbostic: "on page" vs. "onpage"? Don't care which, myself.

src/reconcile/rec_visibility.c
```diff
@@ -117,18 +121,16 @@ __rec_append_orig_value(WT_SESSION_IMPL *session, WT_PAGE *page, WT_INSERT *ins,
...
+        append = tombstone;
 
     /* Append the new entry into the update list. */
-    if (upd_entry != NULL)
-        WT_PUBLISH(*upd_entry, append);
-    else
-        WT_PUBLISH(upd->next, append);
+    WT_PUBLISH(upd->next, append);
+    /* Timestamp should always be in descending order */
+    WT_ASSERT(session, upd->start_ts >= append->start_ts);
```
keithbostic: What (I think), we're really asserting here is that `upd->start_ts >= unpack->start_ts`, and it makes sense to me, but I'd put it above, immediately before we allocate space for append/tombstone.

src/reconcile/rec_visibility.c
```diff
@@ -117,18 +121,16 @@ __rec_append_orig_value(WT_SESSION_IMPL *session, WT_PAGE *page, WT_INSERT *ins,
...
     __wt_cache_page_inmem_incr(session, page, size);
```
keithbostic: I'd suggest moving this line below the code to reset the `upd` type. It doesn't need to be in the middle of stuff, it's kind of an after-thought.

quchenhao: Thanks, I didn't realized I need to take care of the size and I mishandled it completely. Fixed now.

src/reconcile/rec_visibility.c
quchenhao: Thanks, fixed.

src/reconcile/rec_visibility.c
quchenhao: Thanks, fixed.

src/reconcile/rec_visibility.c
quchenhao: Thanks, removed.

src/btree/bt_ret.c
quchenhao: @keithbostic Can you elaborate on that?

PR #5072, opened 2019-12-20
WT-5081 Added --asan option to run WT built with the address sanitizer.

ddanderson:


ddanderson:
@quchenhao, I don't know what your Python skills are like.  Can you review if appropriate?  Thanks!

test/suite/run.py
```diff
@@ -369,6 +396,68 @@ def testsFromArg(tests, loader, arg, scenario):
...
+        #
+        # Our solution is to set the variables as appropriate, and then restart
+        # Python with the same argument list. The shared library loader will
+        # have everything it needs on the second go round.
+
+        ASAN_ENV = "__WT_TEST_SUITE_ASAN"    # if set, we've been here before
+        ASAN_SYMBOLIZER_PROG = "llvm-symbolizer"
+        ASAN_SYMBOLIZER_ENV = "ASAN_SYMBOLIZER_PATH"
+        LD_PRELOAD_ENV = "LD_PRELOAD"
+        SO_FILE_NAME = "libclang_rt.asan-x86_64.so"
```
quchenhao: Don, are you sure the file name is the same for all the clang versions we support?

test/suite/run.py
```diff
@@ -369,6 +396,68 @@ def testsFromArg(tests, loader, arg, scenario):
...
+            if not os.environ.get(ASAN_SYMBOLIZER_ENV):
+                os.environ[ASAN_SYMBOLIZER_ENV] = which(ASAN_SYMBOLIZER_PROG)
+            if not os.environ.get(ASAN_SYMBOLIZER_ENV):
+                print('*** ERROR: %s: symbolizer program not found in PATH'
+                      % ASAN_SYMBOLIZER_ENV)
+            show_env(verbose, ASAN_SYMBOLIZER_ENV)
+            if not os.environ.get(LD_PRELOAD_ENV):
+                symbolizer = follow_symlinks(os.environ[ASAN_SYMBOLIZER_ENV])
+                bindir = os.path.dirname(symbolizer)
+                sofiles = []
```
quchenhao: Using a list here seems a bit overkill. But if the find api only supports list, then that's OK.

test/suite/run.py
```diff
@@ -133,6 +132,32 @@ def usage():
...
+        if os.path.exists(fname) and os.access(fname, os.X_OK):
+            return fname
+    return None
+
+# Follow a symbolic link, returning the target
+def follow_symlinks(pathname):
+    return os.path.realpath(pathname)
+
+# Find all instances of a filename under a directory
+def find(topdir, filename):
```
quchenhao: Sorry, I'm not familiar with os.walk. Now I understand what you are trying to do.

test/suite/run.py
```diff
@@ -369,6 +396,68 @@ def testsFromArg(tests, loader, arg, scenario):
...
+                      % ASAN_SYMBOLIZER_ENV)
+            show_env(verbose, ASAN_SYMBOLIZER_ENV)
+            if not os.environ.get(LD_PRELOAD_ENV):
+                symbolizer = follow_symlinks(os.environ[ASAN_SYMBOLIZER_ENV])
+                bindir = os.path.dirname(symbolizer)
+                sofiles = []
+                if os.path.basename(bindir) == 'bin':
+                    libdir = os.path.join(os.path.dirname(bindir), 'lib')
+                    sofiles = find(libdir, SO_FILE_NAME)
+                if len(sofiles) != 1:
```
quchenhao: We expect to only find one but we may find many. In this case, the error message is wrong.

test/suite/run.py
quchenhao: I did a global search on my dev host and the filename is the same for all the clang versions we have.

quchenhao:
Tried again with /usr/bin/clang and /usr/bin/clang++, still have the same problem
```
Enabling ASAN environment and rerunning python
LD_LIBRARY_PATH=/mnt/data0/chenhaoqu/work/wiredtiger/build_posix/.libs
ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer
LD_PRELOAD=/usr/lib/llvm-8/lib/clang/8.0.1/lib/linux/libclang_rt.asan-x86_64.so
Python restarted for ASAN
==23501==Shadow memory range interleaves with an existing memory mapping. ASan cannot proceed correctly. ABORTING.
==23501==ASan shadow was supposed to be located in the [0x00007fff7000-0x10007fff7fff] range.
==23501==This might be related to ELF_ET_DYN_BASE change in Linux 4.12.
==23501==See https://github.com/google/sanitizers/issues/856 for possible workarounds.
==23501==Process memory map follows:
	0x00007fff7000-0x00008fff7000
	0x00008fff7000-0x02008fff7000
	0x02008fff7000-0x10007fff8000
	0x55f0c46ff000-0x55f0c4700000	/home/chenhaoqu/miniconda3/envs/mongo-build-python2/bin/python2.7
	0x55f0c4700000-0x55f0c4701000	/home/chenhaoqu/miniconda3/envs/mongo-build-python2/bin/python2.7
	0x55f0c4701000-0x55f0c4702000	/home/chenhaoqu/miniconda3/envs/mongo-build-python2/bin/python2.7
	0x55f0c4702000-0x55f0c4703000	/home/chenhaoqu/miniconda3/envs/mongo-build-python2/bin/python2.7
	0x55f0c4703000-0x55f0c4704000	/home/chenhaoqu/miniconda3/envs/mongo-build-python2/bin/python2.7
```

quchenhao:
Try to workaround this problem by building with LDFLAGS="-no-pie" as suggested in https://github.com/google/sanitizers/issues/856, the link command failed with (.text+0x20): undefined reference to `main' .

quchenhao:
BTW, I'm on `Linux 4.15.0-72-generic`

ddanderson:
@quchenhao, wow a big can of worms there.  Maybe you can try building with: these options
`-mllvm -asan-force-dynamic-shadow=1` .  (I saw this mentioned here: https://stackoverflow.com/questions/45757919/how-to-get-asan-working-with-a-4-12-3-linux-kernel ) .  In theory that means the shadowing addresses used by ASAN will not be at a fixed address, so you shouldn't get these conflicts.

ddanderson:
Another possible approach is to do `$ ld.so .../python` when we are rerunning Python.  This is seen in https://github.com/google/sanitizers/issues/837 .  I'm not sure it would make any difference for us.

The conflicting address in the mapping error message appears to be:
`0x02008fff7000-0x10007fff8000`
It would be nice to know what that related to (was it mmap'ed?  .so file? python program executable?) . Then we might stand a chance at understanding how to get it out of the way.

quchenhao:
@ddanderson None of the suggestion in https://stackoverflow.com/questions/45757919/how-to-get-asan-working-with-a-4-12-3-linux-kernel worked. clang doesn't recognise `-no-pie` so I have to use `-f-no-pie` but that doesn't fix the problem either.

Some of the code still building with gcc even CC is set to `clang` so setting `-mllvm -asan-force-dynamic-shadow=1` breaks the build. See the below error. I think this is another problem we need to address anyway.

```
gcc -pthread -shared -B /home/chenhaoqu/miniconda3/envs/mongo-build-python2/compiler_compat -L/home/chenhaoqu/miniconda3/envs/mongo-build-python2/lib -Wl,-rpath=/home/chenhaoqu/miniconda3/envs/mongo-build-python2/lib -Wl,--no-as-needed -Wl,--sysroot=/ -fsanitize=address -fno-omit-frame-pointer -mllvm -asan-force-dynamic-shadow=1 -ggdb -O0 /home/chenhaoqu/work/wiredtiger/build_posix/lang/python/wiredtiger_wrap.o -L/home/chenhaoqu/work/wiredtiger/build_posix/.libs -L/home/chenhaoqu/miniconda3/envs/mongo-build-python2/lib -lwiredtiger -lpython2.7 -o /home/chenhaoqu/work/wiredtiger/build_posix/lang/python/_wiredtiger.so
gcc: error: unrecognized command line option ‘-mllvm’
gcc: error: unrecognized command line option ‘-asan-force-dynamic-shadow=1’
error: command 'gcc' failed with exit status 1
```

quchenhao:
`-mllvm -asan-force-dynamic-shadow=1` can work around the problem but you need to build twice as oddly the first attempt will fail with:
```
gcc -pthread -shared -B /home/chenhaoqu/miniconda3/envs/mongo-build-python2/compiler_compat -L/home/chenhaoqu/miniconda3/envs/mongo-build-python2/lib -Wl,-rpath=/home/chenhaoqu/miniconda3/envs/mongo-build-python2/lib -Wl,--no-as-needed -Wl,--sysroot=/ -fsanitize=address -fno-omit-frame-pointer -mllvm -asan-force-dynamic-shadow=1 -ggdb -O0 /home/chenhaoqu/work/wiredtiger/build_posix/lang/python/wiredtiger_wrap.o -L/home/chenhaoqu/work/wiredtiger/build_posix/.libs -L/home/chenhaoqu/miniconda3/envs/mongo-build-python2/lib -lwiredtiger -lpython2.7 -o /home/chenhaoqu/work/wiredtiger/build_posix/lang/python/_wiredtiger.so
gcc: error: unrecognized command line option ‘-mllvm’
gcc: error: unrecognized command line option ‘-asan-force-dynamic-shadow=1’
error: command 'gcc' failed with exit status 1
```

ddanderson:
> -mllvm -asan-force-dynamic-shadow=1 can work around the problem but you need to build twice as oddly the first attempt will fail with:
> 
> gcc -pthread -shared ...
> gcc: error: unrecognized command line option ‘-mllvm’
> gcc: error: unrecognized command line option ‘-asan-force-dynamic-shadow=1’
> error: command 'gcc' failed with exit status 1

I think these options are only available with `clang`.  Perhaps your second build used `clang`?  If you found a way to build and run ASAN, are there any changes needed in the `run.py` to support the running?  I'll request you to re-review.

quchenhao:
@ddanderson I was building with clang both times but in the first attempt, gcc was used for a file in the python module.

I'm ready to re-review it. I think there is nothing we can do in `run.py` to fix this problem. May be adding a comment to indicate building with `-mllvm -asan-force-dynamic-shadow=1` if the user encounters the problem.

test/suite/run.py
quchenhao: Can we output a separate error message if we find multiple so files?

PR #5073, opened 2019-12-22
WT-4998 Migrate jenkins wiredtiger-test-format-stress-ppc job to evergreen

raviprakashgiri29:


test/evergreen.yml
```diff
@@ -1990,6 +1990,47 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            cat > split_heavy.wtperf << EOF
+            conn_config="cache_size=100MB,log=(enabled=false),statistics=[fast],statistics_log=(wait=5,json=false)"
+            table_config="leaf_page_max=8k,internal_page_max=8k,leaf_item_max=1433,internal_item_max=3100,type=file,memory_page_max=1MB,split_deepen_min_child=100"
```
quchenhao: Why the line is not aligned?

test/evergreen.yml
```diff
@@ -1990,6 +1990,47 @@ tasks:
...
 
+  - name: format-stress-ppc-zseries-test
+    # Set 25 hours timeout
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger with builtins"
+      - func: "format test script"
+        vars:
+          test_env_vars: ulimit -c unlimited
```
quchenhao: Does this work? This doesn't look like a env variable to me. If it works, can we use another variable instead of test_env_vars?

test/evergreen.yml

test/evergreen.yml
raviprakashgiri29: Added a new variable "format_test_setting" .

test/evergreen.yml
```diff
@@ -1990,6 +1990,48 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          script: |
+            set -o errexit
+            set -o verbose
+
+            cat > split_heavy.wtperf << EOF
+                  conn_config="cache_size=100MB,log=(enabled=false),statistics=[fast],statistics_log=(wait=5,json=false)"
+                  table_config="leaf_page_max=8k,internal_page_max=8k,leaf_item_max=1433,internal_item_max=3100,type=file,memory_page_max=1MB,split_deepen_min_child=100"
```
quchenhao: Still not aligned here. The rest lgtm.

test/evergreen.yml
```diff
@@ -2165,6 +2207,9 @@ buildvariants:
   tasks:
   - name: compile
   - name: unit-test
+  - name: format-stress-ppc-zseries-test
```
tammybailey: Should the `format-stress-ppc-zseries-test` also be included under the `ubuntu1804-zseries-build` section?

test/evergreen.yml
```diff
@@ -1990,6 +1990,48 @@ tasks:
           # No of processes to run in the background
           no_of_procs: 10
 
+  # The task name is ppc-zseries because this task will be used in both buildVariants
```
hbokhari: I just had a look at the `wiredtiger-test-format-stress-ppc` and `wiredtiger-test-format-stress-zseries` in Jenkins and I can't see there is a similarity between two tests. Can you double check if one task will work in both build variants or do we need seperate tasks?

test/evergreen.yml
raviprakashgiri29: No, not in this ticket. There is seperate ticket(WT-4999) to migrate format-stress-zseries job in which I would include this task under zseries buildvariant.

tammybailey: Ah okay, thanks!

test/evergreen.yml
```diff
@@ -1990,6 +1990,48 @@ tasks:
...
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger with builtins"
+      - func: "format test script"
+        vars:
+          format_test_setting: ulimit -c unlimited
+          #run for 24 hours ( 24 * 60 = 1440 minutes), use default config
+          format_test_script_args: -b "SEGFAULT_SIGNALS=all catchsegv ./t" -t 1440
+
+  - name: format-wtperf-test
```
hbokhari: One thing to note about all stress tests is that they were supposed to be self triggering, i.e., all steps of original Jenkins tests are repeated few times.

I wonder if we need to repeat the steps in `format-wtperf-test` few times to emulate stress testing? You are doing similar sort of repeated testing in `format-stress-smoke-test` as well.

hbokhari: Just confirm if it would be beneficial to repeat the `wtperf` test with similar configs. I am not sure if `wtperf` randomly selects missing configs.

raviprakashgiri29: I just confirmed with @kommiharibabu that wtperf doesn't randomly sets missing configs before executing, unlike format-test. So, it doesn't make sense to run it multiple times with similar configuration.

hbokhari: Thanks for checking this.

raviprakashgiri29:
> LGTM.
> 
> I will still suggest discussing the runtime of the tasks included here with @lukech to make sure we are not clogging the queue for PPC platform.

I have updated the ppc task to 12 hour. Will wait for @lukech comment before merging.
FYI - PPC distro ubuntu1804-power8-test has 11 hosts unlike zseries (18 hosts).

test/evergreen.yml
```diff
@@ -1990,6 +1990,48 @@ tasks:
...
+  # The task name is ppc-zseries because this task will be used in both buildVariants
+  - name: format-stress-ppc-zseries-test
+    # Set 2.5 hours timeout (60 * 60 * 2.5)
+    exec_timeout_secs: 9000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger with builtins"
+      - func: "format test script"
+        vars:
+          format_test_setting: ulimit -c unlimited
```
lukech: Retain the comment explaining the purpose of this setting would be useful.

test/evergreen.yml
```diff
@@ -1990,6 +1990,48 @@ tasks:
...
+          #run for 2 hours ( 2 * 60 = 120 minutes), use default config
+          format_test_script_args: -b "SEGFAULT_SIGNALS=all catchsegv ./t" -t 120
```
lukech: I agree it's a good start to run the test for 2 hrs on PPC and zSeries, as these 2 platforms have limited host resources in the relevant Evergreen distro pools.

test/evergreen.yml
```diff
@@ -1990,6 +1990,48 @@ tasks:
...
+                checkpoint_threads=0
+                report_interval=5
+                run_time=10
+                threads=((count=20,inserts=1))
+                value_sz=200
+                key_sz=64
+                reopen_connection=false
+            EOF
+
+            ./bench/wtperf/wtperf -O ./split_heavy.wtperf -o verbose=2
```
lukech: It would be neater to put this wtperf configuration into a file instead of crafting it every time running the test. Could the working_dir be adjusted to `wiredtiger/build_posix/bench/wtperf`?

test/evergreen.yml

test/evergreen.yml
raviprakashgiri29: Yes, Thanks for pointing it out. Fixed.

PR #5074, opened 2019-12-23
WT-4978 Migrate pull-request-compilers to evergreen

quchenhao:
I tried to move all the ubuntu test to the new ubuntu1804-wt-build distro. However, there are only limited number of instances running in evergreen. I ended up creating a new build variant for this task and keeping running everything else using ubuntu18.04-test.


hbokhari:
@quchenhao I have an overall comment on the way you have created compilation tasks. Do we really need to have a seperate task for each gcc and clang version? I am trying to understand how efficient (or inefficient) it is to spawn 13 new AWS instances whereas each compilation task should hardly take few minutes to run. Can we reduce the number of tasks perhaps by combining gcc compilation steps in one task and clang compilation in another?

raviprakashgiri29:
I wonder if CXX and CFLAGS need to be set in the configure_env_vars variable under all compiler tasks ?

quchenhao:
True, we should set CXX and CFLAGS, will fix that.

quchenhao:
@hbokhari  I think it depends on whether that task becomes the bottleneck of pull request. Merging gcc and clang builds generally makes sense.  But the downside is that we only get the artefact of the last build for gcc and clang. I would rather still keep them separate for now.

raviprakashgiri29:
> True, we should set CXX and CFLAGS, will fix that.

Thank you for making the changes @Chenhao-Huang .

hbokhari:
@hbokhari For most of the tasks we only use the compilation artefacts from the default compiler version. Unless we are planning to include more tasks in future that want to test different compiled binaries, I will suggest keeping number of tasks low. 

Compilation normally takes around ~4 min, and with 7 versions for gcc, the total runtime of one such task would be around ~30 mins. Any longer and I will definitely break down them into seperate smaller tasks (only if there no other pull request task is taking longer than fused task).

quchenhao:
@hbokhari I understand your concern. Let's wait @lukech's opinion.

lukech:
A typical testing model in Evergreen is to separate tests into multiple tasks so that they can run in parallel to reduce the build variant "Makespan" (or runtime), which is critical for time-sensitive cases like pull request testing. 

This case is a bit special as all "tests" are actually compilations, which means none of the artifacts generated by one task will be reused by other tasks. If we can control the task "Makespan" within 20 mins (not becoming the bottleneck for pull request patch builds) I'd prefer combining them to save some building resources. Maybe one task for gcc (covering all gcc versions) and another for clang (covering all clang version)?

With regard to the concern of artifacts being overwritten across multiple compilations, I think only artifacts of those failed compilations deserve retaining (for troubleshooting), and I assume the task will stop (and being marked as failed) once a given gcc/clang version compilation failure is encountered. Thoughts?

quchenhao:
Since both you guys prefer to combine them, I'll do that. One job for gcc and one job for clang.

quchenhao:
@hbokhari  @lukech Please check this.

PR #5075, opened 2019-12-23
WT-5360 Add template for a glossary in developer docs

agorrod:
There is no real content yet, but I thought I'd get it started.

### DIFF
--- a/src/docs/devdoc-glossary.dox
+++ b/src/docs/devdoc-glossary.dox
@@ -4,7 +4,7 @@ WiredTiger has a lot of domain specific nomenclature - this page attempts
 to decode it. This is intended for those navigating the WiredTiger source
 tree - it describes terms internal to the storage engine.
 
-@section devdoc-glossary-general General terms Used in WiredTiger
+@section devdoc-glossary-general General Terms Used in WiredTiger
 
 <table>
 <caption id="general_terms">General Term Table</caption>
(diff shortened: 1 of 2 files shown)


agorrod:
@tammybailey could you review this for me please? I expect to create a follow on ticket to populate the content.

src/docs/devdoc-index.dox
```diff
@@ -1,5 +1,10 @@
 /*! @page devdoc-index Developer Documentation
 
+@subpage devdoc-glossary
+
+WiredTiger assigns specific meanings to some words, the following page decodes
```
tammybailey: Technically, there is no following page as the link is before the text. Maybe something like _"WiredTiger assigns meanings to specific meanings to certain words. Here we decode them."_  Or similar?

src/docs/devdoc-glossary.dox
```diff
@@ -0,0 +1,30 @@
+/*! @page devdoc-glossary Developer Documentation Glossary of Terms
+
+WiredTiger has a lot of domain specific nomenclature - this page attempts
+to decode it. This is intended for those navigating the WiredTiger source
+tree - it describes terms internal to the storage engine.
+
+@section devdoc-glossary-general General terms Used in WiredTiger
```
tammybailey: terms -> Terms

agorrod:
Thanks for the quick code review turnaround @tammybailey

PR #5079, opened 2019-12-25
WT-5366 read-committed and read-uncommitted transactions can stall eviction

keithbostic:
This change works, but I'm unsure if it's correct. Specific questions:

* Should the check be wrapped in `!F_ISSET(txn, WT_TXN_HAS_SNAPSHOT)` or `txn->isolation == WT_ISO_READ_UNCOMMITTED || txn->isolation == WT_ISO_READ_COMMITTED` or something else?

* I don't understand why `txn_state->metadata_pinned` needs to be checked; my reading of the code indicates it shouldn't be blocking the oldest ID from moving forward permanently, but it does (and it does not move in lockstep with `txn_state->pinned_id`, they can have different values).

EDIT:
In commit a7e0242, I made the checks of `WT_TXN_STATE.pinned_id` and `WT_TXN_STATE.metadata_pinned` independent. As far as I can tell, they have no consistent relationship and either one can independently stall the system. Again, I need some guidance here to get to the correct fix.

### DIFF
--- a/src/txn/txn.c
+++ b/src/txn/txn.c
@@ -1595,14 +1595,13 @@ __wt_txn_global_shutdown(WT_SESSION_IMPL *session, const char *config, const cha
 int
 __wt_txn_is_blocking(WT_SESSION_IMPL *session)
 {
-    WT_CONNECTION_IMPL *conn;
     WT_TXN *txn;
     WT_TXN_STATE *txn_state;
-    uint64_t txn_oldest;
+    uint64_t global_oldest, txn_oldest;
 
-    conn = S2C(session);
     txn = &session->txn;
     txn_state = WT_SESSION_TXN_STATE(session);
+    global_oldest = S2C(session)->txn_global.oldest_id;
 
     /* We can't roll back prepared transactions. */
     if (F_ISSET(txn, WT_TXN_PREPARE))
@@ -1616,24 +1615,19 @@ __wt_txn_is_blocking(WT_SESSION_IMPL *session)
     if (txn->mod_count == 0 && !__wt_op_timer_fired(session))
         return (false);
 
-    /* Check the oldest transaction ID of either the current transaction ID or the snapshot. */
-    txn_oldest = txn->id;
-    if (F_ISSET(txn, WT_TXN_HAS_SNAPSHOT) && txn->snap_min != WT_TXN_NONE &&
-      (txn_oldest == WT_TXN_NONE || WT_TXNID_LT(txn->snap_min, txn_oldest)))
-        txn_oldest = txn->snap_min;
-
     /*
+     * Check the oldest transaction ID of either the current transaction ID or the snapshot.
+     *
      * Read-committed and read-uncommitted transactions enter transaction IDs into the global table
      * to stop updates they're reading from being freed, and that can prevent the oldest ID from
      * moving forward.
      */
-    if (txn_state->pinned_id != WT_TXN_NONE && WT_TXNID_LT(txn_state->pinned_id, txn_oldest))
-        txn_oldest = txn_state->pinned_id;
-    if (txn_state->metadata_pinned != WT_TXN_NONE &&
-      WT_TXNID_LT(txn_state->metadata_pinned, txn_oldest))
-        txn_oldest = txn_state->metadata_pinned;
-
-    return (txn_oldest == conn->txn_global.oldest_id ?
+    txn_oldest = txn->id;
+    if (F_ISSET(txn, WT_TXN_HAS_SNAPSHOT) && txn->snap_min != WT_TXN_NONE &&
+      (txn_oldest == WT_TXN_NONE || WT_TXNID_LT(txn->snap_min, txn_oldest)))
+        txn_oldest = txn->snap_min;
+    return (txn_oldest == global_oldest || txn_state->pinned_id == global_oldest ||
+          txn_state->metadata_pinned == global_oldest ?
         __wt_txn_rollback_required(
           session, "oldest pinned transaction ID rolled back for eviction") :
         0);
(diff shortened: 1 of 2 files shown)


src/txn/txn.c
```diff
@@ -1614,12 +1615,19 @@ __wt_txn_is_blocking(WT_SESSION_IMPL *session)
...
+     * to stop updates they're reading from being freed, and that can prevent the oldest ID from
+     * moving forward.
+     */
     txn_oldest = txn->id;
     if (F_ISSET(txn, WT_TXN_HAS_SNAPSHOT) && txn->snap_min != WT_TXN_NONE &&
       (txn_oldest == WT_TXN_NONE || WT_TXNID_LT(txn->snap_min, txn_oldest)))
         txn_oldest = txn->snap_min;
-    return (txn_oldest == conn->txn_global.oldest_id ?
+    return (txn_oldest == global_oldest || txn_state->pinned_id == global_oldest ||
+          txn_state->metadata_pinned == global_oldest ?
```
agorrod: This seems reasonable to me, though I find the code structure a bit surprising. You are choosing between the earliest of `txn->id` and `txn->snap_min`, then comparing that, `pinned_id` and `metadata_pinned` directly to `global_oldest`.

keithbostic: Isn't that comparing the oldest of `txn.id` and `txn.snap_min`?

The idea is to take `snap_min` if `id` is `none` or `snap_min` is less than `id`, that is, older than `id`?

And then, of course, if that value or either of `pinned_id` and `metadata_pinned` match the oldest id, we flag the thread.

All that said, I was hoping for some direction on `pinned_id` and `metadata_pinned`. I don't know how they are supposed to work, this was just a guess on my part.

michaelcahill: @keithbostic, I've kicked off testing with this change (I think it's basically equivalent to your except it doesn't check `metadata_pinned`):

```c
    /*
     * Check the oldest transaction ID of either the current transaction ID or its pinned ID.
     */
    txn_oldest = txn_state->pinned_id != WT_TXN_NONE && WT_TXNID_LT(txn_state->pinned_id, txn->id) ?
        txn_state->pinned_id : txn->id;
    return (global_oldest == txn_oldest ?
        __wt_txn_rollback_required(
          session, "oldest pinned transaction ID rolled back for eviction") :
        0);
```

So far it's running for me without getting stuck: was this hitting quickly or did it take a long time?

The `metadata_pinned` code is deliberately separating checkpoint transaction IDs from the rest, since checkpoints are long-running but only write to metadata.  My intuition is that it shouldn't matter much if a transaction happens to have the oldest `metadata_pinned` in the system: that will keep metadata pages from being evicted, but if the cache is full of dirty metadata we're already in a bad state.

I was hoping to reproduce the problem and poke around: I'll leave this version running and see if it hits overnight.

keithbostic: @michaelcahill, it's been a few weeks, I don't remember the details any more, but I think test test for `WT_TXN.pinned_id` hit relatively quickly, and the test for `WT_TXN. metadata_pinned` was much less likely. IIRC, I only saw it once, and it's quite possible I misread the core dump.

The code snippet you posted sets `txn_oldest`, does that mean you're no longer checking the transaction ID or the snapshot?

Various versions of the transaction ID and the snapshot test have been included in the code for awhile, the current version was as the result of review by @agorrod, in WT-5082, [this commit](https://github.com/wiredtiger/wiredtiger/commit/279808ccdec6ffdb1b379abdd9b06217f0073f90). WT-5082 was a tracking ticket for lots of cache failures, [this comment](https://jira.mongodb.org/browse/WT-5082?focusedCommentId=2567208&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2567208) is probably the most interesting one.

As best I remember, we started off checking `WT_TXN.id` against `WT_TXN_GLOBAL.oldest_id`, then we added a check for `WT_TXN.snap_min` on transactions with `WT_TXN_HAS_SNAPSHOT` set (to catch read-only, read-committed transactions that didn't have `WT_TXN.id` set), and in this latest round I first added a check for `WT_TXN.pinned_id` and then saw a failure where it appeared `WT_TXN.metadata_pinned` was earlier than `WT_TXN.pinned_id` and it was the oldest ID in the system.

All that said, I don't understand how this code should work, so I'm happy to have you commit what you think is right, and I'll watch for failures, and bring them to your attention if they happen.

michaelcahill: > All that said, I don't understand how this code should work...

What can we do about this?  Would more / longer comments help, more internals documentation, I could talk it through, or ...?

First up, `txn->id` and `txn_state->id` are mostly aliases for each other.  They are equal practically everywhere.

Transactions get a snapshot for both snapshot isolation and read-committed.  In that case, `txn->snap_min` is the minimum ID of concurrent transactions.  That value is usually published as `txn_state->pinned_id`.  But if one of the concurrent transactions happens to be a checkpoint, we exclude it from the calculation of `txn_state->pinned_id` (which will be the minimum of all *other* concurrent transaction IDs) and instead track it in `txn_state->metadata_pinned`.

For read-uncommitted, `txn->snap_min` isn't meaningful but we still set `txn_state->pinned_id` during operations so an update isn't freed underneath a reader.

Across all sessions, the minimum of `txn_state->pinned_id` and `txn_state->id` is tracked in `txn_global->oldest_id`.  And the minimum of all the `txn_state->metadata_pinned` is tracked in `txn_global->metadata_pinned`.

So comparing `txn_state->metadata_pinned` to `txn_global->oldest_id` doesn't make much sense to me: we deliberately allow the `oldest_id` calculation to go past checkpoint transactions, which are tracked separately in `metadata_pinned`.

In short, to include read-uncommitted updates blocking the cache, I think the code here should mirror the `oldest_id` calculation in `__txn_oldest_scan` -- if a session's `txn_state->id` or `txn_state->pinned_id` is equal to `txn_global->oldest_id`, it is preventing the global oldest from moving forward.

keithbostic: @michaelcahill, thank you for the explanation. If you'll push that change into the tree, we can merge it and now I'll know what to check for if there are any subsequent failures.

PR #5080, opened 2019-12-25
WT-5367 Log a message when rolling a thread back because it holds the oldest timestamp

keithbostic:


tetsuo-cpp:
@keithbostic Is this change because the logging is too verbose? In the original ticket description of WT-5090, there was a suggestion to possibly put it behind a `verbose=rollback_messages` config string if the original change caused too much testing noise.

keithbostic:
@tetsuo-cpp 
> Is this change because the logging is too verbose? In the original ticket description of WT-5090, there was a suggestion to possibly put it behind a verbose=rollback_messages config string if the original change caused too much testing noise.

Yes. Based on @agorrod's comment in WT-5090 ("I do want to keep the notion of logging messages when rolling back transactions due to cache pressure."), my understanding was Alex wanted logging in the field, not just if a particular verbose option was configured.

agorrod:
@keithbostic something time based would be nice. Could we do something similar to `__wt_checkpoint_progress` here? In fact, maybe we could generalize out that functionality?

You are right that I want this in the field, MongoDB does configure verbose options for some logging, so there is precedent for putting this behind a verbose option (checkpoint_progress and recovery_progress are two examples of that).

keithbostic:
Closing and discarding the PR, we're not going to make this change.

PR #5081, opened 2019-12-27
WT-5328 in memory obsolete page cleanup

kommiharibabu:
Checkpoint garbage cleanup cannot delete the obsolete in-memory pages.
As these pages are obsolete, no session can refer these pages in future.
Lets add them to the urgent eviction queue, once they get evicted, the
next checkpoint of history store takes care of removing those obsolete
pages.


src/btree/bt_sync.c
```diff
@@ -109,44 +109,53 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
 
-    /* Guard against marking an in-memory page to be deleted. */
-    if (ref->state != WT_REF_DISK)
-        return (false);
-
     /* Ignore internal pages, these are taken care of during reconciliation. */
     if (!__wt_ref_is_leaf(session, ref))
         return (false);
 
+    /* In-memory ref address must be from an off page address. */
```
kommiharibabu: @keithbostic Is my understanding correct that an in-memory page that `rec_result` is not empty must have an address from off-page?

keithbostic: > Is my understanding correct that an in-memory page that rec_result is not empty must have an address from off-page?

No; we create pages  in memory in some cases (as one example, an insert split will create a new page in memory that was never read from disk). Those pages are reconciled normally, and so `rec_result` can have a value.

In short, `rec_result` tells you the result of the last reconciliation, but nothing else.

src/btree/bt_sync.c
```diff
@@ -109,44 +109,53 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    mod = (ref->page != NULL) ? (ref->page->modify) : NULL;
+    if (mod != NULL && mod->rec_result == WT_PM_REC_REPLACE)
+        return (__wt_txn_visible_all(
+          session, mod->mod_replace.newest_stop_txn, mod->mod_replace.newest_stop_ts));
+
+    if (mod != NULL && mod->rec_result == WT_PM_REC_MULTIBLOCK)
+        return (__wt_txn_visible_all(
+          session, mod->mod_multi_newest_stop_txn, mod->mod_multi_newest_stop_ts));
+
```
keithbostic: I don't strongly object to this, but I don't see much point in it either. Normal eviction mechanisms should find and evict this page, so I think it's an unnecessary complication to try and push in-memory pages out during the checkpoint. Maybe it's necessary, but we don't yet have any data (of which I'm aware), that shows that to be true.

src/btree/bt_sync.c
```diff
@@ -109,44 +109,53 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
     WT_ORDERED_READ(addr, ref->addr);
+    WT_ASSERT(session, addr != NULL);
```
keithbostic: I don't believe you need an ordered read here, can you please write down what made you think one was necessary?

There's not much point in asserting a value isn't NULL, is there? We're about to drop core if it's NULL anyway.

src/btree/bt_sync.c
```diff
@@ -341,6 +350,15 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
...
                     btree->rec_max_timestamp = mod->rec_max_timestamp;
+
+                /*
+                 * Check whether the tree is lookaside btree and verify whether the page contents
+                 * are obsolete or not by checking the visibility of the page stop time pair.
+                 */
+                if (is_las && __sync_ref_is_obsolete(session, walk))
+                    /* Must be an in-memory leaf page. Evict it. */
+                    WT_ERR(__sync_ref_evict_or_mark_deleted(session, walk));
+
```
keithbostic: This isn't necessarily an in-memory leaf page, it's either on-disk or in-memory, isn't it?

src/btree/bt_sync.c
keithbostic: It seems confusing to me we use the `mod_replace` structure directly in this case, but we've aggregated the information for the `mod_multi` structure. Why are we handling them differently?

src/btree/bt_sync.c
kommiharibabu: Thanks. I updated the comments.

src/btree/bt_sync.c
kommiharibabu: I don't find a scenario, so removed it. Added a condition `addr != NULL` to find out scenarios where the page is not reconciled yet and not disk address.

src/btree/bt_sync.c
kommiharibabu: Yes, Updated the comment to just reflect that it should be a leaf page.

src/btree/bt_sync.c
kommiharibabu: I feel checkpoint adding the obsolete pages to the eviction queue will improve the cache usage than eviction server picking the pages. Eviction server works on picking random pages, so it is not guaranteed that the obsolete pages are cleaned as it is done by the checkpoint.

Lets wait to here others opinions and then we can decide about it.

kommiharibabu: When the page is reconciled into multiple blocks, Instead of checking each individual blocks to decide the obsoleteness, we aggregated them, but whereas for `mod_replace` scenario, it is just one block, I feel directly referring it may be enough.

keithbostic: My approach would be to aggregate when doing the work in checkpoint, rather than modifying reconciliation and adding new fields to the `WT_PAGE_MODIFY` structure.

test/suite/test_gc03.py
```diff
@@ -0,0 +1,156 @@
...
+            ',stable_timestamp=' + timestamp_str(300))
+
+        # Checkpoint to ensure that the history store is gets populated and cleaned
+        self.session.checkpoint()
+
+        # Check that the new updates are only seen after the update timestamp
+        #self.check(bigvalue, uri, nrows, 300)
+
+        # When this limitation is fixed we'll need to uncomment the calls to self.check
+        self.KNOWN_LIMITATION('values stored by this test are not yet validated and checkpoint has to write to history store')
```
ddanderson: Okay, this is good that there is a `KNOWN_LIMITATION`, but I still fear that this will get ignored and we'll forget to fix this down the road.  Can you please create a ticket to address the `KNOWN_LIMITATION` in this file and any others that you've added?

src/btree/bt_sync.c
sueloverso: I would prefer simplicity and let eviction do eviction and keep this as something we can do if necessary. Simpler code is better to start.
If we keep this, the comment needs to be clearer. The code is subtle in the fields it is using that are only different in the middle of a long name and dereference.

src/btree/bt_sync.c
```diff
@@ -109,44 +109,57 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    mod = (ref->page != NULL) ? (ref->page->modify) : NULL;
+    if (mod != NULL && mod->rec_result == WT_PM_REC_REPLACE)
+        return (__wt_txn_visible_all(
+          session, mod->mod_replace.newest_stop_txn, mod->mod_replace.newest_stop_ts));
+
+    if (mod != NULL && mod->rec_result == WT_PM_REC_MULTIBLOCK)
+        return (__wt_txn_visible_all(
+          session, mod->mod_multi_newest_stop_txn, mod->mod_multi_newest_stop_ts));
+
+    /* Check for the obsolete from page disk address. */
```
sueloverso: The comment is unclear. Does it mean:
```Check if the page is obsolete using the page disk address.``` ?

File: src/btree/bt_sync.c [hunk 7]
```diff
@@ -341,6 +354,17 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
                     btree->rec_max_txn = mod->rec_max_txn;
                 if (mod != NULL && btree->rec_max_timestamp < mod->rec_max_timestamp)
                     btree->rec_max_timestamp = mod->rec_max_timestamp;
+
+                /*
+                 * Check whether the tree is lookaside btree and verify whether the page contents
+                 * are obsolete or not by checking the visibility of the page stop time pair.
```
sueloverso: I would end the sentence after `are obsolete`. The detail of the page stop time pair is not in this code here and could become outdated and wrong.

test/suite/test_gc03.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # Check that the new updates are only seen after the update timestamp
+        #self.check(bigvalue, uri, nrows, 300)
+
+        # Pin oldest and stable to timestamp 400.
+        self.conn.set_timestamp('oldest_timestamp=' + timestamp_str(300) +
+            ',stable_timestamp=' + timestamp_str(300))
+
+        # Checkpoint to ensure that the history store is gets populated and cleaned
+        self.session.checkpoint()
```
sueloverso: Are there, or will there be, any statistics we can check here to make sure that this happened?

src/btree/bt_sync.c

src/btree/bt_sync.c
kommiharibabu: Yes, Use that sentence.

src/btree/bt_sync.c

test/suite/test_gc03.py
kommiharibabu: The stats ticket is not yet implemented, Once all the implementation tickets are finished, we have a test ticket that should take care of all the pending test changes.

test/suite/test_gc03.py
kommiharibabu: There is a pending test ticket in the project, I will update that ticket to take care of all the remaining test changes.

src/btree/bt_sync.c
kommiharibabu: As discussed, Currently I am going ahead with checkpoint adding the pages to urgent evict queue. Later we can revisit this decision if any problem arises.

PR #5083, opened 2019-12-28
WT-5327 Backup cursor API can fail to restore WT_SESSION name information

keithbostic:


keithbostic:
autobuild retest this please

keithbostic:
Moving the macros isn't the fix, changing the `WT_RET` to `WT_ERR` in the `WT_CURSOR_BACKUP_CHECK_STOP` macro is the fix.

I moved the macro mostly because I thought it maybe (might? possibly?), help us not to reintroduce the bug some time in the future.

sueloverso:
Thanks for pointing that out. I missed that subtlety and now makes sense.

PR #5084, opened 2019-12-29
WT-5337 Example to use traits pattern to simplify code

quchenhao:
This is an example to explore whether it's feasible to use traits pattern to get rid of some of the switch and if statements in the code.

I managed to get it working and successfully removed some of the switch statements. Of course there are more I haven't tried to tackle yet. However, the way I chose to implement consumes a little bit more memory than before as I store the traits reference in the in-memory data structures.

I implemented the traits pattern in hierarchy. In each module, I have a traits file, which defines the traits structs for that module. If a module depends on another module, its traits will include the traits of the modules it depends on. In my example, I demonstrate it with the rec_traits, which contains cache_las_traits. The idea here is to encapsulate the implementation in each module instead of having a monolithic traits file for everything, so that we can potentially build each module separately in the future.

Please put your thoughts about this in the Jira ticket (WT-5337) :).


quchenhao:
@keithbostic Thanks for the comment. I totally agree it's not ideal to put code in the traits file. We should abstract things in header files for each .c file in the modules, include the header files in the traits file, and keep the implementation in other .c files. However, it' not easy to achieve that ATM with the current build framework.

src/btree/bt_traits.c
```diff
@@ -0,0 +1,152 @@
...
+    if (cbt->ins != NULL && !F_ISSET(cbt, WT_CBT_VAR_ONPAGE_MATCH))
+        return (0);
+
+    /*
+     * Although updates would have appeared as an "insert" objects, variable-length column store
+     * deletes are written into the backing store; check the cell for a record already deleted when
+     * read.
+     */
+    if (__wt_cell_type(WT_COL_PTR(page, &page->pg_var[cbt->slot])) == WT_CELL_DEL)
+        return (0);
```
ddanderson: The original code used temporary variables `cip` and `cell`.  Even though the new code is more compact, I'd keep the old variables, as they were inserted to clarify the code.

ddanderson:
@quchenhao, I would agree with @keithbostic that it would be good to try to find a way to move the traits functions near the related functions.  What prevents us from doing that?

src/btree/bt_huffman.c
```diff
@@ -197,12 +197,14 @@ __wt_btree_huffman_open(WT_SESSION_IMPL *session)
 {
     struct __wt_huffman_table *table;
     WT_BTREE *btree;
+    const WT_BT_TRAITS *bt_traits;
```
sueloverso: This should alphabetize before `WT_BTREE`.

src/btree/bt_traits.c
```diff
@@ -0,0 +1,152 @@
...
+static int __bt_row_cursor_valid(WT_CURSOR_BTREE *cbt, WT_UPDATE **updp, bool *valid);
+
+/*
+ * __bt_col_fix_huffman --
+ *     col fix doesn't support huffman.
+ */
+static int
+__bt_col_fix_huffman(WT_SESSION_IMPL *session, size_t len)
+{
+    (void)len; /* Unused */
```
sueloverso: These should all be `WT_UNUSED(...)` I think.

src/include/wt_internal.h
```diff
@@ -379,6 +385,9 @@ typedef uint64_t wt_timestamp_t;
 #include "block.h"
 #include "bloom.h"
 #include "btmem.h"
+#include "bt_traits.h"
+#include "cache_las_traits.h"
+#include "rec_traits.h"
```
sueloverso: Although I agree with the comments of others on the file naming and moving things out of the traits files, if this sticks, these need to be alphabetized.

quchenhao:
@ddanderson  I think we can move the implementation outside of the traits file. What we need to do is to create a header file which declares these functions. Then we have the traits file include this header file and move the implementation to the .c file. The thing I don't like of this path is that we need to put all the header files in the include folder ATM. It feels odd to expose something private to the module there. If we can build each module separately, it would be much nicer and we can put the header files in each module. I will try this out when I have time.

quchenhao:
Close until I have time to look at this again.

src/btree/bt_cursor.c
```diff
@@ -288,77 +284,7 @@ __wt_cursor_valid(WT_CURSOR_BTREE *cbt, WT_UPDATE **updp, bool *valid)
...
-                if (upd->type == WT_UPDATE_TOMBSTONE)
-                    return (0);
-                if (updp != NULL)
-                    *updp = upd;
-            }
-        }
-        break;
-    }
-    *valid = true;
+    WT_RET(bt_traits->cursor_valid(cbt, updp, valid));
```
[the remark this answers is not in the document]
quchenhao: I don't like that either. But `__wt_cursor_valid` also sets `*valid`. Cannot reset that in the traits function and making them owning it.

quchenhao:
@keithbostic @sueloverso I moved implementations in the traits files to separate files. I can break them into smaller files if needed. However, it is a little bit annoying that I have to make the functions `extern` and add the following exceptions to `s_export`:
```
egrep -v '^__bt' |
egrep -v '^__las' |
egrep -v '^__rec' |
```

quchenhao:
retest this please

PR #5085, opened 2019-12-30
WT-5290 checkpoint garbage collection verbose output

kommiharibabu:
Added a new verbose macro to enable to get the verbose logs of checkpoint
doing the garbage cleanup of history store.


src/btree/bt_sync.c
```diff
@@ -109,44 +109,69 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
     WT_CELL_UNPACK vpack;
+    WT_PAGE_MODIFY *mod;
+    bool obsolete;
 
     /* Ignore root pages as they can never be deleted. */
-    if (__wt_ref_is_root(ref))
-        return (false);
-
-    /* Guard against marking an in-memory page to be deleted. */
-    if (ref->state != WT_REF_DISK)
```
ddanderson: These lines appear to have been lost in the change.
I'm assuming this PR and ticket are only to add verbose output and no other functional changes.

src/btree/bt_sync.c
```diff
@@ -109,44 +109,69 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    else if (mod != NULL && mod->rec_result == WT_PM_REC_MULTIBLOCK)
+        obsolete = __wt_txn_visible_all(
+          session, mod->mod_multi_newest_stop_txn, mod->mod_multi_newest_stop_ts);
+    /* Check for the obsolete from page disk address. */
+    else if (addr != NULL) {
+        if (!__wt_off_page(ref->home, addr)) {
+            __wt_cell_unpack(session, ref->home, (WT_CELL *)addr, &vpack);
+            obsolete = __wt_txn_visible_all(session, vpack.newest_stop_txn, vpack.newest_stop_ts);
+        } else
+            obsolete = __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts);
```
ddanderson: This is a lot of new logic that seems to have been added.  I wonder if a merge didn't work correctly?

dist/api_data.py
```diff
@@ -682,6 +682,7 @@ def __ge__(self, other):
             'api',
             'block',
             'checkpoint',
+            'checkpoint_garbage_collection',
```
sueloverso: I don't feel strongly, but given that the tests are all `test_gc##` and such, I think `checkpoint_gc` is equally as clear and shorter.

src/btree/bt_sync.c
sueloverso: I agree with @ddanderson here. There are a lot of non-verbose related changes that are concerning to me. Please address that and then I'll do another review.

PR #5086, opened 2019-12-30
WT-5235 Benchmark for lookaside stress testing

hbokhari:
This is an example Workgen benchmark to stress lookaside/LAS. This is achieved through: 

- Long running transactions consisting of read and update operations.
- Low cache size (~20%) for a reasonably sized WT table with large documents.
- Pareto distribution for operations in long running transactions. This will cause skewed access for a selective set of keys in WT table.
- Relatively large number of read operation threads to stress cache.

### DIFF
--- a/bench/workgen/runner/evict-btree-lookaside.py
+++ b/bench/workgen/runner/evict-btree-lookaside.py
@@ -34,6 +34,41 @@
 #     skewed access for a selective set of keys in WT table.
 #   - Relatively large number of read operation threads to stress cache.
 
+# This benchmark is based on a wtperf config: wiredtiger/bench/wtperf/runners/evict-btree-scan.wtperf
+# There are number of changes made to original configs such as:
+#   - Reduced the number of documents inserts during initial warm-up phase.
+#   - Increased the sizes of key and value.
+#   - Reduced the WT cache size.
+#   - Added transaction based operations and repurposed some of the read threads as long-running 
+#     transcation threads.
+
+'''
+# wtperf options file: evict btree configuration
+conn_config="cache_size=40G,checkpoint=(wait=60,log_size=2GB),eviction=(threads_min=12,threads_max=12),log=(enabled=true),session_max=600,eviction_target=60,statistics=(fast),statistics_log=(wait=1,json)"
+# 1B records * (key=12 + value=138) is about 150G total data size
+key_sz=12
+value_sz=138
+log_like_table=true
+table_config="type=file"
+icount=1000000000
+report_interval=5
+run_time=3600
+# Scans every 10 minutes for all the scan specific tables.
+# .4B records * (key=12 + value=138) is about 60G total data size for scan
+# Running on a machine with 64G physical memory, this exhausts both the
+# WT cache and the system cache.
+scan_interval=600
+scan_pct=100
+scan_table_count=20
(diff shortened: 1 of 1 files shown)


bench/workgen/runner/evict-btree-lookaside.py
```diff
@@ -0,0 +1,123 @@
...
+ops = op_multi_table(ops, tables, False)
+ops = op_log_like(ops, log_table, 0)
+thread2 = Thread(ops)
+# These operations include log_like operations, which will increase the number
+# of insert/update operations by a factor of 2.0. This may cause the
+# actual operations performed to be above the throttle.
+thread2.options.throttle=500
+thread2.options.throttle_burst=1.0
+
+# Long running transcations
```
ddanderson: spelling: transcations

ddanderson: and perhaps a longer comment here to mention that each transaction will take at least 16 minutes (there will be 10000 sleeps of 0.1 second embedded in the other activity).

bench/workgen/runner/evict-btree-lookaside.py
```diff
@@ -0,0 +1,123 @@
...
+# OTHER DEALINGS IN THE SOFTWARE.
+
+# This benchmark is designed to stress disk access to the LAS/History Store file.
+# This is achieved through:
+#   - Long running transactions consisting of read and update operations.
+#   - Low cache size (~20%) for a reasonably sized WT table with large documents.
+#   - Pareto distribution for operations in long running transactions. This will cause
+#     skewed access for a selective set of keys in WT table.
+#   - Relatively large number of read operation threads to stress cache.
+
```
ddanderson: I'd like to see the original wtperf file you used embedded here as a comment, and clearly mark the sections of the generated code that you changed or inserted.  I haven't always done this in the past, but I think it would be a good policy going forward.  The benefit is that this can allow any reader familiar with wtperf options (which are nicely concise) to tell at a glance at what this program does without needing to understand all the extra boiler-plate, like log-like files, etc.

ddanderson:
@hbokhari, I have a general question about the results you are seeing.  Are you seeing much variation if you change any parameters, like number of threads running for each activity?  Or number of files?  I don't have anything to compare your results to, and was just wondering if there's anything you try that gives more pronounced cache overflow activity.  The more we get the better for this benchmark, I think?

bench/workgen/runner/evict-btree-lookaside.py
hbokhari: Good catch on typo. Will also add the comment on duration of these transactions.

bench/workgen/runner/evict-btree-lookaside.py
hbokhari: Good idea. I had that before committing the changes and I wasn't sure if I should add the original config file. I will add the original config and point out the major changes I have done.

hbokhari:
@ddanderson, Yes I had to try few config before reaching a point where I can confidently generate LAS activity. I do know that having large documents (relatively larger key-value size) and limiting the key distribution(pareto) for operation in transactions helped. I haven't tried more configs since then but I will spend some time today to understand which parameters can control the intended LAS stress. I will update my findings on Jira ticket.

PR #5087, opened 2019-12-31
WT-5290 checkpoint garbage collection verbose output

kommiharibabu:
Added a new verbose macro to enable to get the verbose logs of checkpoint
doing the garbage cleanup of history store.


src/btree/bt_sync.c
```diff
@@ -114,43 +114,50 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
-        return (__wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts));
+            obsolete = __wt_txn_visible_all(session, vpack.newest_stop_txn, vpack.newest_stop_ts);
+        } else
+            obsolete = __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts);
     }
 
-    return (false);
+    if (obsolete)
+        __wt_verbose(session, WT_VERB_CHECKPOINT_GC, "%p: page is found as obsolete", (void *)ref);
+    return (obsolete);
```
keithbostic: The `__wt_off_page()` call will fail if `addr == NULL`, so I don't think you need both that call and a test for `addr == NULL`.

kommiharibabu: Removed the unnecessary NULL check.

src/btree/bt_sync.c
```diff
@@ -114,43 +114,47 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+        obsolete = __wt_txn_visible_all(session, multi_newest_stop_txn, multi_newest_stop_ts);
+    } else if (!__wt_off_page(ref->home, addr)) {
+        /* Check if the page is obsolete using the page disk address. */
+        __wt_cell_unpack(session, ref->home, (WT_CELL *)addr, &vpack);
+        obsolete = __wt_txn_visible_all(session, vpack.newest_stop_txn, vpack.newest_stop_ts);
+    } else
+        obsolete = __wt_txn_visible_all(session, addr->newest_stop_txn, addr->newest_stop_ts);
+
+    if (obsolete)
+        __wt_verbose(session, WT_VERB_CHECKPOINT_GC, "%p: page is found as obsolete", (void *)ref);
```
sueloverso: Users of verbose are generally debugging or trying to understand what's going on. I suggest adding a reason string for each possible path and printing that out, so that the user knows exactly which of these many paths we took to determine the page is obsolete. It might also be useful to record and print out the values of `stop_txn` and `stop_ts` from whatever source it is.

kommiharibabu: Added additional logs in the obsolete check paths including the stop time pair.

PR #5088, opened 2019-12-31
WT-5369 Don't use transaction ids from a page with a previous startup generation

tetsuo-cpp:
When we started writing validity windows to disk in #5022, we began hitting assertions telling us that some cells had a stop time pair earlier than the start time pair. This can happen when you call `wiredtiger_open` on an existing database since your transaction id allocator resets to 0 and already existing cells will have their own transaction ids. The solution in this PR is to ignore transaction ids from a previous startup generation.


src/btree/bt_handle.c
```diff
@@ -549,8 +549,12 @@ __btree_conf(WT_SESSION_IMPL *session, WT_CKPT *ckpt)
     btree->syncing = WT_BTREE_SYNC_OFF; /* Not syncing */
                                         /* Checkpoint generation */
     btree->checkpoint_gen = __wt_gen(session, WT_GEN_CHECKPOINT);
+
+    /* Keep track of the latest write gen for this btree from the previous run. */
+    btree->base_write_gen = ckpt->write_gen;
```
tetsuo-cpp: @keithbostic 
The crux of the change is here. Since the write gen is incremented per btree, I am maintaining a base write gen for each btree whereas the base write gen was previously the write gen of the metadata file (which is the highest write gen seen on any btree at the time of checkpoint).

If you imagine a case where one btree has gone through many write gens for whatever reason (more pages, more eviction, etc) and another btree which has gone through only a few write gens, the `base_write_gen` would be a very high number. And any pages written on the less active btree will have their transaction ids ignored with our previous strategy.

src/btree/bt_page.c
```diff
@@ -317,14 +317,11 @@ __inmem_col_var_repeats(WT_SESSION_IMPL *session, WT_PAGE *page, uint32_t *np)
...
-__unstable_skip(WT_SESSION_IMPL *session, const WT_PAGE_HEADER *dsk, WT_CELL_UNPACK *unpack)
+__unstable_skip(WT_CELL_UNPACK *unpack)
 {
-    /*
-     * Skip unstable entries after downgrade to releases without validity windows and from previous
-     * wiredtiger_open connections.
-     */
+    /* Skip unstable entries after downgrade to releases without validity windows. */
     return ((unpack->stop_ts != WT_TS_MAX || unpack->stop_txn != WT_TXN_MAX) &&
-      (S2C(session)->base_write_gen > dsk->write_gen || !__wt_process.page_version_ts));
```
tetsuo-cpp: As discussed, we don't want to skip this cell entirely.

src/btree/bt_handle.c
keithbostic: I'm still confused here (yeah, you probably guessed that). Let me try to describe the two scenarios I see:
```
Generation 1
    Btree A: 1-50
    Btree B: 1-30
Generation 2-10
    Btree A: +50 more pages each generation

Btree A has generations 1-500 on disk, Btree B has generations 1-30 on disk.

The database restarts, and we start modifying both tables and write a few pages.
```
In model #1, Btree A has generations 1-510 on disk and Btree B has generations 1-30 and 501-510 on disk, and the test for Btree B is "generation less than 501 (the maximum base generation as of this restart)".

In model #2, Btree A has generations 1-510 on disk and Btree B has generations 1-40 on disk, and the test for Btree B is "generation less than 31 (the Btree generation for Btree B as of this restart)".

How are those two questions different?

tetsuo-cpp: Ah, you're correct. I forgot the part where we start the write gen of all Btrees to the maximum at the time of restart:
```
btree->write_gen = WT_MAX(ckpt->write_gen, conn->base_write_gen);
```

src/meta/meta_ckpt.c
```diff
@@ -590,6 +590,10 @@ __wt_meta_ckptlist_set(
     WT_DECL_RET;
     bool has_lsn;
 
+    /* Review the checkpoint's write generation. */
```
tetsuo-cpp: @keithbostic 
Ok, the culprit is here. The `__ckptlist_review_write_gen` function is where we intercept if we're looking at the metadata checkpoint and set it to the current `WT_CONNECTION::max_write_gen` value. This needs to be done BEFORE updating the metadata or otherwise when we restart, we will read the previous checkpoint's max write gen.

src/meta/meta_ckpt.c
```diff
@@ -590,6 +590,10 @@ __wt_meta_ckptlist_set(
...
+    WT_CKPT_FOREACH (ckptbase, ckpt)
+        __ckptlist_review_write_gen(session, ckpt);
+
```
keithbostic: I believe the current code (before your change), and this version, both update the write generation in all checkpoints, not just the new one. I don't think the write generation is used for anything at the moment and it isn't a problem, but it seems wrong to rewrite historic checkpoints with new write generations.

If you agree, maybe we should move the call to `__ckptlist_review_write_gen()` into `__wt_meta_ckptlist_to_meta()`, and only call it if `WT_CKPT_ADD` is set?

src/btree/bt_vrfy_dsk.c
```diff
@@ -238,8 +238,16 @@ __verify_dsk_ts_addr_cmp(WT_SESSION_IMPL *session, uint32_t cell_num, const char
...
+  uint64_t txn1, const char *txn2_name, uint64_t txn2, bool gt, const char *tag,
+  const WT_PAGE_HEADER *dsk)
 {
+    /*
+     * If we unpack a value that was written as part of a previous startup generation, we set start
+     * id to "none" and the stop id to "max" so we need an exception here.
+     */
+    if ((txn1 == WT_TXN_NONE || txn1 == WT_TXN_MAX) &&
+      dsk->write_gen <= S2C(session)->base_write_gen)
+        return (0);
```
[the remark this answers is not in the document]
tetsuo-cpp: 6acdbb9

src/btree/bt_vrfy.c
```diff
@@ -739,8 +739,15 @@ __verify_ts_addr_cmp(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t cell_num, c
 static int
 __verify_txn_addr_cmp(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t cell_num,
   const char *txn1_name, uint64_t txn1, const char *txn2_name, uint64_t txn2, bool gt,
-  WT_VSTUFF *vs)
+  WT_VSTUFF *vs, const WT_PAGE_HEADER *dsk)
```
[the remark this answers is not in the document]
tetsuo-cpp: 6acdbb9

src/btree/bt_vrfy.c
```diff
@@ -739,8 +739,15 @@ __verify_ts_addr_cmp(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t cell_num, c
...
 {
+    /*
+     * If we unpack a value that was written as part of a previous startup generation, we set start
+     * id to "none" and the stop id to "max" so we need an exception here.
+     */
+    if ((txn1 == WT_TXN_NONE || txn1 == WT_TXN_MAX) &&
+      dsk->write_gen <= S2C(session)->base_write_gen)
+        return (0);
```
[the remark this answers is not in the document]
tetsuo-cpp: 6acdbb9

src/meta/meta_ckpt.c
tetsuo-cpp: 0f91e4c

src/btree/bt_vrfy_dsk.c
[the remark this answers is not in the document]
tetsuo-cpp: 6acdbb9

bvpvamsikrishna:
I would like to understand how the below case is handled.
1. Assume A page startup generation number is 1. It has (k2, v2, t2) and (k4, v4, t4).
2. After the restart, now the startup generation number is 2.
3. We read the page, as part of unpack we update the transaction id t2 -> 0, and t4 -> 0.
4. New key/value pair ( k6, v6, t6) is added to page.
5. Now when the page is flushed to disk, during checkpoint, what will be txn id's in the cell on the disk image. will they be (t2 -> 0, t4 -> 0, t6) or (t2, t4, t6) ?

bvpvamsikrishna:
> I would like to understand how the below case is handled.
> 
> 1. Assume A page startup generation number is 1. It has (k2, v2, t2) and (k4, v4, t4).
> 2. After the restart, now the startup generation number is 2.
> 3. We read the page, as part of unpack we update the transaction id t2 -> 0, and t4 -> 0.
> 4. New key/value pair ( k6, v6, t6) is added to page.
> 5. Now when the page is flushed to disk, during checkpoint, what will be txn id's in the cell on the disk image. will they be (t2 -> 0, t4 -> 0, t6) or (t2, t4, t6) ?

The cells will be rebuild, so  (t2 -> 0, t4 -> 0, t6) is correct.
Thank you for updating @tetsuo-cpp

tetsuo-cpp:
retest this please

src/include/cell.i
```diff
@@ -811,6 +795,31 @@ restart:
     if (cell->__chunk[0] & WT_CELL_64V) /* skip value */
         WT_RET(__wt_vunpack_uint(&p, end == NULL ? 0 : WT_PTRDIFF(end, p), &unpack->v));
 
+    /*
```
tetsuo-cpp: @bvpvamsikrishna This is the change that we discussed.

tetsuo-cpp:
@bvpvamsikrishna This change is causing too many test failures to be merged. I think that search has to be fixed first for this to work nicely.

keithbostic:
@tetsuo-cpp 
> I think that if a delete was made from the previous run, the transaction id should be set to WT_TXN_NONE and IF there is no stop timestamp in the cell, the stop timestamp should be WT_TS_NONE.

This doesn't make sense to me, but I don't know enough about the current state of the branch to have a strong opinion. If you aren't confident & want another opinion, let's talk it over.

src/btree/bt_vrfy.c
```diff
@@ -311,23 +311,13 @@ __verify_addr_ts(WT_SESSION_IMPL *session, WT_REF *ref, WT_CELL_UNPACK *unpack,
 {
     char ts_string[2][WT_TS_INT_STRING_SIZE];
 
-    if (unpack->newest_stop_ts == WT_TS_NONE)
```
tetsuo-cpp: These have to go. I thought I may be able to keep them by moving the logic to a higher level, however, if we reconcile the page then these "none" transaction ids and timestamps will actually get written to the disk. So I don't think we can assert on anything useful here.

src/include/txn.i
```diff
@@ -765,22 +765,13 @@ __upd_alloc_tombstone(
 }
 
 /*
- * __wt_txn_read --
+ * __wt_txn_read_upd_list --
```
tetsuo-cpp: This change is not strictly related. But my change exposed another visibility issue so I figured I should fix it and keep the build green since it isn't hard to do.

Essentially, when we call `__wt_txn_read` to search an insert list, we're not intending to check the onpage value or the history store as there shouldn't be any content there anyway. So I've extracted out the iteration of the update list into a separate function and used that where appropriate.

src/reconcile/rec_visibility.c
```diff
@@ -350,11 +350,18 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
             upd_select->upd = upd = upd->next;
         }
         if (upd != NULL) {
-            /* The beginning of the validity window is the selected update's time pair. */
-            if (upd->start_ts < upd_select->stop_ts)
+            /*
+             * The beginning of the validity window is the selected update's time pair.
+             *
+             * FIXME-PM-1521: We shouldn't need to check for this. We're actually allowed to commit
```
tetsuo-cpp: Updates can be committed out of timestamp order. So we can have updates with an earlier stop time pair than their start time pair. This breaks a lot of assumptions we've made so there is going to be a discussion with the execution team about whether we need to continue supporting this. For now, I'm just guarding this check to prevent failures.

src/include/cell.i
```diff
@@ -924,6 +908,43 @@ __wt_cell_unpack_dsk(
     }
 
     WT_IGNORE_RET(__wt_cell_unpack_safe(session, dsk, cell, unpack, NULL));
+
+    /*
```
tetsuo-cpp: This is the main guts of my change. This logic corresponds to the table that I showed in the knowledge sharing session this week.

src/reconcile/rec_visibility.c
```diff
@@ -350,11 +350,18 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+             * updates to the same key out of timestamp order. So we can have stop time pairs
+             * earlier than their respective start time pair. We need to figure out what to do in
+             * WT-5469.
```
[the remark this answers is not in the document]
tetsuo-cpp: We have tests exercising this so we can't assert here. If we find that we're allowed to enforce this then we can remove the tests and assert here.

src/include/cell.i
tetsuo-cpp: Done.

src/reconcile/rec_visibility.c
```diff
@@ -377,10 +384,10 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
              * but this logic still isn't correct and should be handled properly when we begin the
              * recovery work.
```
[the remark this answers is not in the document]
tetsuo-cpp: That `FIXME` doesn't apply anymore so I've removed it. How is this related to rollback to stable?

src/include/txn.i
keithbostic: Maybe this should be `__wt_txn_read_ins_list` then? There are update lists that aren't insert lists?

src/btree/bt_vrfy.c
michaelcahill: I think the case to consider is a table where timestamps aren't being used.  We might still have a range of transaction IDs that need to be retained, so when we evict there could be a tombstone that isn't obsolete (it has an active `stop_txn` but `stop_ts == WT_TS_NONE`).  That value shouldn't be visible after a restart but could be reconciled then read back into the same process.

src/include/txn.i
tetsuo-cpp: I use this in `__wt_txn_read` to read the update list on the modify structure (not insert list). So I don't think I can rename this to something more specific.

src/include/txn.i
```diff
@@ -796,7 +787,33 @@ __wt_txn_read(WT_SESSION_IMPL *session, WT_CURSOR_BTREE *cbt, WT_UPDATE *upd, WT
...
+{
+    WT_ITEM buf;
+    WT_TIME_PAIR start, stop;
+    size_t size;
+
+    WT_CLEAR(buf);
+    start.txnid = WT_TXN_NONE;
+    start.timestamp = WT_TS_NONE;
+    stop.txnid = WT_TXN_MAX;
+    stop.timestamp = WT_TS_MAX;
```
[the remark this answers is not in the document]
tetsuo-cpp: c605577

src/btree/bt_vrfy_dsk.c
```diff
@@ -227,12 +227,19 @@ __verify_dsk_ts_addr_cmp(WT_SESSION_IMPL *session, uint32_t cell_num, const char
...
     if (gt && txn1 >= txn2)
         return (0);
     if (!gt && txn1 <= txn2)
         return (0);
+    /*
+     * If we unpack a value that was written as part of a previous startup generation, we set start
+     * id to "none" and the stop id to "max" so we need an exception here.
+     */
+    if (dsk->write_gen <= S2C(session)->base_write_gen)
+        return (0);
```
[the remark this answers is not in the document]
tetsuo-cpp: 08be381

Ok, I've adjusted the comment. It happens regardless of the correction. In this case, we haven't done the correction yet but that means we could have very high transaction ids from the previous run which still causes this problem.

src/btree/bt_vrfy.c
tetsuo-cpp: > As a possible change, did you consider using some other value than `none`, creating a valid window that would be ignored in all cases, like (0,1)?

ba76501

Change to use timestamp 1 instead. We can't do the same for transaction ids but it's nice that we can at least keep some of these checks.

> I'm confused how these values get written to disk -- shouldn't reconciliation simply discard these pages, doesn't it mean there's nothing useful on them?

If a page with a previous startup generation gets read in, some keys are modified and it gets reconciled again, any cells for values that did NOT get modified could have `stop_txn == WT_TXN_NONE` and these time pairs will actually get written to the disk. So the next time it gets read in, your write gen says it is from the current run however you'll see these corrected time pairs.

tetsuo-cpp:
The test failure is due to `hs09` which currently isn't passing in the durable history branch. Merging now.

PR #5089, opened 2019-12-31
WT-5374 Commit to las for each key and free updates for eviction

quchenhao:


quchenhao:
@bvpvamsikrishna This change is not ready for review yet. Had a discussion with @kommiharibabu and it seems the insert takes effect immediately and the transaction should never be rolled back (transaction here is just to allocate some resources). So this change may not be required. However, in another branch, I observe that when we are trying to evict lookaside table, it has uncommitted changes, which contradicts to @kommiharibabu's theory. Need to check with him.

bvpvamsikrishna:
@quchenhao, I have locked the PR, remove the locked label when you think, it is ready for review.

PR #5090, opened 2020-01-01
WT-5377 variable-length column store insert locking in format test program can stall eviction

keithbostic:


test/format/ops.c
```diff
@@ -1706,6 +1613,63 @@ row_insert(TINFO *tinfo, WT_CURSOR *cursor, bool positioned)
...
+     * total rows. This is wasteful, but we want to give other threads immediate access to the row,
+     * ideally they'll collide with our insert before we resolve.
+     *
+     * Process the existing records and advance the last row count until we can't go further.
+     */
+    do {
+        WT_ORDERED_READ(v, g.rows);
+        for (i = 0, p = tinfo->insert_list; i < WT_ELEMENTS(tinfo->insert_list); ++i) {
+            if (*p == v + 1) {
+                testutil_assert(__wt_atomic_casv64(&g.rows, v, v + 1));
```
sueloverso: Just to make sure I understand: At first I was surprised that we're dying if the atomic operation fails. But this thread should be the only one with an insert that == v + 1, correct? So the atomic CAS should actually never fail for this thread.

keithbostic: Yes, that's correct.

We don't really need the atomic update because we're the only writer as you pointed out, but we do want a barrier to ensure the write gets done immediately, so I used the atomic.

Let me know if you see a simpler solution.

PR #5093, opened 2020-01-02
WT-5321 Stats update according to durable history

kommiharibabu:
As part of the durable history project, the cache overflow
functionality is replaced with the new functionality of the history
store. As part of this change, some of the existing
stats related to lookaside are invalid and those gets removed.
But these are stats are quite important to find out the performance
problems in the customer environments, because of that reason,
we will maintain some stats that are possible to provide with
history store useful for finding out performance issues.


dist/stat_data.py
```diff
@@ -283,24 +283,22 @@ def __init__(self, name, desc, flags=''):
     CacheStat('cache_hazard_checks', 'hazard pointer check calls'),
     CacheStat('cache_hazard_max', 'hazard pointer maximum array length', 'max_aggregate,no_scale'),
     CacheStat('cache_hazard_walks', 'hazard pointer check entries walked'),
+    CacheStat('cache_hs_cursor_wait_application', 'history store cursor application thread wait time (usecs)'),
+    CacheStat('cache_hs_cursor_wait_internal', 'history store cursor internal thread wait time (usecs)'),
+    CacheStat('cache_hs_insert', 'history store table insert calls'),
+    CacheStat('cache_hs_ondisk', 'history store table on-disk size', 'no_clear,no_scale,size'),
+    CacheStat('cache_hs_ondisk_max', 'history store table max on-disk size', 'no_clear,no_scale,size'),
+    CacheStat('cache_hs_read', 'history store table update reads'),
```
bvpvamsikrishna: I prefer to get away with "update" in the description, same for below one's as well.

dist/stat_data.py
```diff
@@ -703,18 +701,18 @@ def __init__(self, name, desc, flags=''):
...
     CacheStat('cache_eviction_walks_gave_up_ratio', 'eviction walks gave up because they saw too many pages and found too few candidates'),
     CacheStat('cache_eviction_walks_stopped', 'eviction walks gave up because they restarted their walk twice'),
+    CacheStat('cache_hs_read', 'history store table update reads'),
     CacheStat('cache_inmem_split', 'in-memory page splits'),
     CacheStat('cache_inmem_splittable', 'in-memory page passed criteria to be split'),
-    CacheStat('cache_lookaside_read', 'cache overflow table update reads'),
-    CacheStat('cache_page_instantiate_read_lookaside', 'pages read into cache requiring cache overflow entries.'),
-    CacheStat('cache_page_instantiate_read_lookaside_checkpoint', 'pages read into cache by checkpoint requiring cache overflow entries.'),
+    CacheStat('cache_page_instantiate_read_hs', 'pages read into cache requiring history store entries.'),
+    CacheStat('cache_page_instantiate_read_hs_checkpoint', 'pages read into cache by checkpoint requiring history store entries.'),
```
bvpvamsikrishna: Does this stat still makes sense ? Even if we read a page which has history store entries, how that will impact ?

dist/stat_data.py
kommiharibabu: Fixed.

dist/stat_data.py
kommiharibabu: No. I removed the stats and its usage. The function that uses these stats is not removed, as that cleanup is taken care in another ticket.

PR #5096, opened 2020-01-03
WT-5334 Add testing coverage for static wt build

raviprakashgiri29:


raviprakashgiri29:
Evergreen patch build link

https://evergreen.mongodb.com/task/wiredtiger_ubuntu1804_static_wt_build_test_patch_88b55001df864e762b96303fa2a6bfd621e19469_5e0eb71c7742ae0b839b8fc9_20_01_03_03_38_23

test/evergreen.yml
```diff
@@ -2067,6 +2067,35 @@ tasks:
             cp ../../../bench/wtperf/split_heavy.wtperf .
             ./wtperf -O ./split_heavy.wtperf -o verbose=2
 
+  - name: static-wt-build-test
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
```
lukech: I believe the `CC='eval "g++ -static"'` setting is essential for this test. A `configure_env_vars` var overloading will do?

test/evergreen.yml
```diff
@@ -2067,6 +2067,35 @@ tasks:
...
+          posix_configure_flags: --enable-silent-rules --enable-diagnostic --enable-static
```
lukech: @agorrod could you please confirm if the `--disable-shared --with-builtins=snappy,zlib` configure settings pasted in Jira are essential (or optional) for this test?

agorrod: They are necessary. Without those options the compile will build a shared library, when this test is explicitly about a static library.

raviprakashgiri29: @lukech & @agorrod : I think "--with-builtins=snappy,zlib" is not necessary here because "wt" utility in this test is not used for any compression, it is just used to print the wiredtiger library version and validates if "wt" utility build is static. 
But I am open for suggestions :)

agorrod: I agree that it's not confirmed to work by the test, but it is valuable to be doing the static build with that option. The reason we have this is to ensure we can compile a `wt` binary that can be sent as a single file and used to investigate issues with MongoDB deployments.

test/evergreen.yml
raviprakashgiri29: > I believe the `CC='eval "g++ -static"'` setting is essential for this test. A `configure_env_vars` var overloading will do?

`CC='eval "g++ -static"'` will be set to rebuild static "wt" util.

test/evergreen.yml
raviprakashgiri29: Thank you @agorrod , updated to configure with snappy & zlib options.

raviprakashgiri29:
Evergreen patch build

https://evergreen.mongodb.com/task/wiredtiger_ubuntu1804_static_wt_build_test_patch_88b55001df864e762b96303fa2a6bfd621e19469_5e12badf56234354cb4ece68_20_01_06_04_43_31

test/evergreen.yml
```diff
@@ -2067,6 +2067,35 @@ tasks:
...
+
+            # -V option displays Wiredtiger library version
+            ./wt -V
+
+            if [ $? -ne 0 ]; then
+              echo "Error, WT util was not generated or is not functioning"
+              exit 1
+            fi
+
+            ldd wt || FAIL=1
```
lukech: I was a bit confused while reading the FAIL setting and the later checking to determine errors. `FAIL=1` here actually means failing to run `ldd wt` i.e. wt was statically built. Suggest either changing the boolean variable name, or adding a descriptive comment.

test/evergreen.yml
```diff
@@ -2067,6 +2067,39 @@ tasks:
...
+      - command: shell.exec
+        params:
+          working_dir: "wiredtiger/build_posix"
+          shell: bash
+          script: |
+            set -o errexit
+            set -o verbose
+
+            # Delete wt util
+            rm -f wt
```
lukech: Not sure this is necessary.

test/evergreen.yml
```diff
@@ -2067,6 +2067,39 @@ tasks:
...
+            # Build static wt util
+            make CC='eval "g++ -static"'
```
lukech: Can the make part being covered by the above `compile wiredtiger` function?

test/evergreen.yml

test/evergreen.yml
raviprakashgiri29: I think it is safe and prefer to delete wt util then rebuild/remake with "-static" option.

test/evergreen.yml
raviprakashgiri29: The idea here is to
1. Build wiredtiger with "--disable-shared --with-builtins=snappy,zlib" configure option and default compiler(CC=mongodbtoolchain-gcc, CXX=mongodbtoolchain-g++). 
2. Delete wt util
3. Rebuild/remake wiredtiger with -static option.
4. Check if "wt" util is statically linked.

lukech: Thanks for the clarification. Now it makes sense to me. 
Would it be cleaner to call `make wiredtiger` function here, or at least adding the `${smp_command}` in the make command?

raviprakashgiri29: Thanks for pointing it out. I cannot call "make wiredtiger" function here because CC variable cannot be modified under this function to be set to "g++ -static". 
Added "${smp_command}" instead in the make command.

lukech: Sounds good, thanks.

test/evergreen.yml
agorrod: It is relied on by the first check below. If we don't remove `wt` here, it'll still be there after a failed static compile.

PR #5097, opened 2020-01-03
WT-5343 wt list show history

jeremytay90:


src/utilities/util_list.c
```diff
@@ -143,7 +143,7 @@ list_print(WT_SESSION *session, const char *uri, bool cflag, bool vflag)
          */
         if (!vflag && WT_PREFIX_MATCH(key, WT_SYSTEM_PREFIX))
             continue;
-        if (cflag || vflag || (strcmp(key, WT_METADATA_URI) != 0 && strcmp(key, WT_LAS_URI) != 0))
+        if (cflag || vflag || strcmp(key, WT_METADATA_URI) != 0)
```
sulabhM: The comment needs to be updated around "We don't normally say anything about the WiredTiger metadata and lookaside tables".

PR #5098, opened 2020-01-03
WT-5291 Added new statistics for garbage collection of history store

ddanderson:
The new statistics didn't seem to fit in an existing statistics category, so I made a new one with the idea that the existing history stats can be migrated.


src/btree/bt_sync.c
```diff
@@ -384,9 +385,13 @@ __wt_sync_file(WT_SESSION_IMPL *session, WT_CACHE_OP syncop)
                 break;
 
             /* Traverse through the internal page for obsolete child pages. */
-            if (is_las && WT_PAGE_IS_INTERNAL(walk->page)) {
-                WT_WITH_PAGE_INDEX(session, ret = __sync_ref_int_obsolete_cleanup(session, walk));
-                WT_ERR(ret);
+            if (is_las) {
```
kommiharibabu: Adding the refs visited stat here doesn't provide a count of all refs that are visited, during the internal page processing, we will go through each child page of that internal page. So, I feel it is better to place the counter in the `__sync_ref_is_obsolete()` function would be good.

src/btree/bt_sync.c
```diff
@@ -194,6 +194,7 @@ __sync_ref_evict_or_mark_deleted(WT_SESSION_IMPL *session, WT_REF *ref)
 
     /* Evict the in-memory obsolete page */
     WT_IGNORE_RET_BOOL(__wt_page_evict_urgent(session, ref));
+    WT_STAT_CONN_INCR(session, hs_gc_pages_removed);
```
kommiharibabu: The `removed` stat here doesn't provide the actual pages that are removed.  The removed case is in the above if check. Move the stat there.

dist/stat_data.py
```diff
@@ -384,6 +388,12 @@ def __init__(self, name, desc, flags=''):
     DhandleStat('dh_sweep_tod', 'connection sweep time-of-death sets'),
     DhandleStat('dh_sweeps', 'connection sweeps'),
 
+    ##########################################
+    # History statistics
+    ##########################################
+    HistoryStat('hs_gc_pages_visited', 'history pages visited for garbage collection'),
+    HistoryStat('hs_gc_pages_removed', 'history pages removed for garbage collection'),
+
```
kommiharibabu: How about adding one more stat to indicate `how many pages are added for eviction`? Having this information also will help to identify easily how the GC is doing.

dist/stat_data.py
```diff
@@ -384,6 +388,12 @@ def __init__(self, name, desc, flags=''):
     DhandleStat('dh_sweep_tod', 'connection sweep time-of-death sets'),
     DhandleStat('dh_sweeps', 'connection sweeps'),
 
+    ##########################################
+    # History statistics
+    ##########################################
+    HistoryStat('hs_gc_pages_visited', 'history pages visited for garbage collection'),
+    HistoryStat('hs_gc_pages_removed', 'history pages removed for garbage collection'),
```
sueloverso: Minor nit: alphabetize them.

dist/stat_data.py
sueloverso: @kommiharibabu is this new stat you're requesting really part of the changes that are under review already?

kommiharibabu: @sue, No. The requested stat code already exists.  Currently the code either `delete` the page or `add it to urgent eviction` queue.

sueloverso: FYI my username is not @ sue. Someone else got that notification!

dist/stat_data.py
ddanderson: fixed.

dist/stat_data.py
ddanderson: @kommiharibabu I've added the new stat and pushed, please check it.

src/btree/bt_sync.c
ddanderson: Thanks!  Done.

src/btree/bt_sync.c
ddanderson: Fixed.

PR #5099, opened 2020-01-05
WT-5387 prepared transaction resolution can stall eviction on active pages

keithbostic:
@agorrod, @bvpvamsikrishna: I'm not sure who should do the review of this one. The 1798c9b commit is an eviction change (57f7c60, c8d9b24 and 8006ddd are debugging fixes to that change). The 98c8e3c commit is the prepared-transaction change (and ea769ec is a debugging fix to that change). I've asked you two for reviews, but don't hesitate to turf your review slot to someone else.

I think 1798c9b is the more contentious change: basically, it seems wrong to me to involve any thread that's resolving a transaction in forced eviction, and that was slowing the workload that I was working with. In summary, if a page is large, we're already having trouble evicting it and a thread resolving a transaction is going to make it more likely we can evict the page. in other words, if transaction commit/prepare/rollback wants a page, by definition that page has an unresolved transaction and can't be evicted.


keithbostic:
Three questions from con call:
* Consider using `qsort` as the sort function.
* Use flags instead of doing `WT_SESSION.name` string comparisons.
* Sort by key as well as by file ID to minimize pages brought into memory.

keithbostic:
@bvpvamsikrishna, @agorrod, all of the changes discussed in today's call have been made.

src/btree/bt_cursor.c
```diff
@@ -473,17 +473,21 @@ __wt_btcur_reset(WT_CURSOR_BTREE *cbt)
...
     WT_UPDATE *upd;
 
+    *updp = NULL;
+
+    cbt = (WT_CURSOR_BTREE *)cursor;
     btree = cbt->btree;
-    cursor = &cbt->iface;
     session = (WT_SESSION_IMPL *)cursor->session;
-    *updp = upd = NULL; /* -Wuninitialized */
+    upd = NULL; /* -Wuninitialized */
```
agorrod: Why take out the initialization of the output parameter pointer?

src/btree/bt_read.c
```diff
@@ -715,8 +722,7 @@ __wt_page_in_func(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t flags
             /*
              * Check if the page requires forced eviction.
              */
-            if (evict_skip || LF_ISSET(WT_READ_NO_SPLIT) || btree->evict_disabled > 0 ||
-              btree->lsm_primary)
```
agorrod: Is it OK to move these flags (and the new check) up to the initialization of evict_skip? That local variable gets overwritten at other places in the loop here, so it seems possible that moving the flag checking to the initialization will change the behavior.

src/session/session_api.c
```diff
@@ -1651,12 +1651,14 @@ __session_commit_transaction(WT_SESSION *wt_session, const char *config)
         ret = EINVAL;
     }
 
+    F_SET(session, WT_SESSION_RESOLVE);
```
agorrod: I'd prefer `WT_SESSION_RESOLVING_TXN` to `WT_SESSION_RESOLVE`

src/txn/txn.c
```diff
@@ -866,6 +871,9 @@ __txn_commit_timestamps_assert(WT_SESSION_IMPL *session)
     return (0);
 }
 
+/* File object comparison macro for the insertion sort of transaction operations. */
+#define WT_TXN_OP_CMP_LT(entry1, entry2) ((entry1).btree->id < ((entry2).btree->id))
```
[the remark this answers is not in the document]
keithbostic: The current version sorts by key as well as by btree ID, pushed in 47934d4.

There's a limitation, we can't sort row-store keys if a collator has been configured, and I don't see any way around that limitation.

src/btree/bt_cursor.c
keithbostic: I moved the initialization of the output parameter to the top of the function.

I like to initialize side-effects early on, as a policy to ensure callers are never surprised by an early error in the function.

src/btree/bt_read.c
keithbostic: I don't see any way we can get into trouble, the things we're checking here don't change in the function (and the only other action on `evict_skip` in the function is to set it to `true`).

That said, I originally moved the code because the test was complex (with all the `WT_SESSION.name` strcmp calls. Now that it's simpler, I'll move it back to where it was just for simplicity of the change.

Pushed in acc79033f.

src/session/session_api.c
keithbostic: I agree, that's a better name.

Pushed in e23e16482.

src/btree/bt_cursor.c
agorrod: Sorry, I didn't see that.

PR #5101, opened 2020-01-05
WT-5388 Fix las write squash statistic

quchenhao:


src/cache/cache_las.c
```diff
@@ -686,6 +686,7 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
         }
 
         upd = prev_upd = NULL;
+        squashed = false;
```
keithbostic: I'd suggest moving the initialization of `squashed` and `prev_upd` to the loop where they're used, but I wouldn't argue for it.

src/cache/cache_las.c
```diff
@@ -766,8 +767,13 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
                 F_SET(upd, WT_UPDATE_HISTORY_STORE);
                 ++insert_cnt;
+
+                if (squashed) {
+                    WT_STAT_CONN_INCR(session, cache_hs_write_squash);
+                    squashed = false;
+                }
             } else
-                WT_STAT_CONN_INCR(session, cache_hs_write_squash);
+                squashed = true;
```
keithbostic: Correct me if I'm wrong, but I recall the original code updated the statistic outside the loop, and so it could only be updated once no matter what. This looks like that's not true -- squashed could toggle between true & false and get updated once per toggle. Maybe that's simply not possible?

Regardless, I prefer the previous structure, where it was obvious it was only updated a single time.

Again I won't argue against what you've got, it's your call.

quchenhao: Yeah, I found the previous code also changes the behaviour. It doesn't handle the situation where we squash multiple times on the same chain, like:

5 - 5 - 5 - 4 - 3 - 3 - 3

It should be counted as two squashes.

PR #5105, opened 2020-01-06
WT-5393 prepared transaction rollback and API error handling fixes

keithbostic:
@agorrod, @sueloverso, I got the name check wrong in WT-5324.

Fixing it reveals that we need to allow at least two more calls in the context of transactional errors that have already occurred.

I'm pushing forward down this path, but there's certainly an argument to be made that it's the wrong path to take, as it gains complexity.

We should definitely make sure Evergreen runs cleanly with this change before merging.


src/include/txn.i
```diff
@@ -68,8 +68,9 @@ static inline int
 __wt_txn_err_chk(WT_SESSION_IMPL *session)
 {
     /* Allow transaction rollback, but nothing else. */
-    if (!F_ISSET(&(session->txn), WT_TXN_ERROR) ||
-      strcmp(session->name, "rollback_transaction") != 0)
+    if (!F_ISSET(&(session->txn), WT_TXN_ERROR) || strcmp(session->name, "WT_CURSOR.close") == 0 ||
+      strcmp(session->name, "WT_CURSOR.reset") == 0 ||
+      strcmp(session->name, "WT_SESSION.rollback_transaction") == 0)
```
sueloverso: I have two comments on this. While I'm fine with this if it is the correct change, I think it deserves two changes:
1. The comment is out of date with the exceptions. And I think the exceptions deserve a sentence of explanation about why they're exceptions.
2. Now that we're up to 3 exceptions, maybe add a macro for easy adding/removal of them.

keithbostic: I totally agree with your #1, fixed in 3b6108174.

On #2, I'm not sure why a macro makes it easier to add? Can you give me an example?

keithbostic:
@agorrod, @bvpvamsikrishna, here's a PR that cleans up prepared transaction handling, for review and discussion. As I mentioned in our earlier conversation, this PR crashes the system if we can't commit a prepared transaction for any reason.

keithbostic:
@bvpvamsikrishna, when you review this change (or for whoever you ask to review in your place), can you or they check the test changes and confirm the tests I commented out are tests that we should no longer allow? -- Thanks.

src/include/txn.i
```diff
@@ -60,16 +60,51 @@ __wt_txn_context_check(WT_SESSION_IMPL *session, bool requires_txn)
...
-    /* Allow transaction rollback, but nothing else. */
-    if (!F_ISSET(&(session->txn), WT_TXN_ERROR) ||
-      strcmp(session->name, "rollback_transaction") != 0)
+    /*
+     * Allow only a few select operations after an error: WT_CURSOR.{close,reset} are called
+     * internally and WT_SESSION.rollback_transaction is the expected application resolution.
+     */
+    if (!F_ISSET(&(session->txn), WT_TXN_ERROR) || strcmp(session->name, "WT_CURSOR.close") == 0 ||
+      strcmp(session->name, "WT_CURSOR.reset") == 0 ||
+      strcmp(session->name, "WT_SESSION.rollback_transaction") == 0)
```
agorrod: Doing string comparison against `session->name` doesn't feel right to me - we usually use that string for diagnostics, not for making decisions. Can you think of another reasonable solution?

keithbostic: This is a little more complicated than [the similar change in the WT-5387 PR](https://github.com/wiredtiger/wiredtiger/pull/5099/commits/d63513c6f2d09911821be7bfb5a8cd62e9c004c3) because it involves not just a `WT_SESSION` method, but also `WT_CURSOR` methods, there's not just a single flag.

I don't think it's a performance problem because we only check if `WT_TXN_ERROR` is already set, which should be rare.

If you prefer adding flags to both structures, or a structure boolean or some other solution, let me know and I'll make the change.

agorrod: You are right - I don't have an obvious better way. The way we've done this before is by splitting macros and using different variants depending on need, but that's pretty ugly and this code is moving away from those API level macros (which I think is generally good).
So I'm happy with it as-is, but will keep looking for ways to avoid using strcmp against the interface names saved in the session handle.

PR #5108, opened 2020-01-06
WT-5391 Merge develop into durable history and resolve conflicts

luke-pearson:
The only conflict I saw that was tricky was in rec_visibility.c where @keithbostic had added:
```         
            /*
             * Track the oldest update not on the page, used to decide whether reads can use the
             * page image, hence using the start rather than the durable timestamp.
             */
            if (upd->start_ts < r->min_skipped_ts)
                r->min_skipped_ts = upd->start_ts;
```


keithbostic:
Taking myself off the review list, the branch has been closed by @luke-pearson.

PR #5109, opened 2020-01-06
WT-5392 Adding history store page back to urgent queue if eviction fails

kommiharibabu:
The obsolete history store pages that are added to evition is not able to evict
because of evicton threads try to evict the page before the checkpoint release
the hazard pointer on them. Try to add them back to the urgent queue until they
get evicted.

Added additional checks to avoid edge cases where the child page is deleted, but
not yet removed from internal page.

New statistic is added to find out how the eviction is performing with adding back
the obsolete page to the queue again.

Reorganize the tests to reduce the duplicate code


src/evict/evict_page.c
```diff
@@ -213,6 +213,21 @@ __wt_evict(WT_SESSION_IMPL *session, WT_REF *ref, uint32_t previous_state, uint3
...
+         * them back to the queue and try to evict again.
+         *
+         * It is possible that an eviction failure occurs to a page that is not added by a
+         * checkpoint thread, but we consider those scenarios are limited.
+         */
+        if (WT_IS_LAS(S2BT(session))) {
+            WT_IGNORE_RET_BOOL(__wt_page_evict_urgent(session, ref));
+            WT_STAT_CONN_INCR(session, cache_eviction_history_store_pages_queued_urgent_again);
+            __wt_yield();
+        }
```
keithbostic: We can't do this in all cases, I suspect the only case where this is appropriate is when `ret == EBUSY`, and probably only if `EBUSY` is returned from `__evict_exclusive()`, but I'm not sure of that.

Further, why are we doing this to every LAS page that fails to evict? Why not only to ones that were on the urgent queue?

keithbostic: Why are we calling `__wt_yield()`?

keithbostic: Why did you add this functionality to `__wt_evict()`? That seems like the wrong place, it seems like it would be a change to the code that is handling eviction from the urgent queue, perhaps `__evict_get_ref()`?

keithbostic: My opinion is we should ignore these failures entirely.

As far as I know, we don't have any data that indicates this work is necessary, and eviction will get around to these pages. I don't yet have any reason to believe it's going to be urgent for us to handle these pages, so I would prefer to not complicate the code absent evidence it's necessary.

src/btree/bt_sync.c
```diff
@@ -122,8 +122,15 @@ __sync_ref_is_obsolete(WT_SESSION_IMPL *session, WT_REF *ref)
...
+    /* Ignore deleted pages. */
+    if (ref->state == WT_REF_DELETED) {
+        __wt_verbose(session, WT_VERB_CHECKPOINT_GC, "%p: skipping deleted page", (void *)ref);
+        return (false);
+    }
+
     /* Ignore internal pages, these are taken care of during reconciliation. */
-    if (!__wt_ref_is_leaf(session, ref)) {
+    if ((ref->addr != NULL && !__wt_ref_is_leaf(session, ref)) ||
+      (ref->page != NULL && WT_PAGE_IS_INTERNAL(ref->page))) {
```
keithbostic: This doesn't look safe to me.

If this code is called from `__sync_ref_int_obsolete_cleanup()`, what prevents the `WT_REF` from being evicted out from under us? That is, can't `WT_ref.page` be switched to NULL after the test and before the call to `WT_PAGE_IS_INTERNAL`?

Maybe based on where we are in the checkpoint walk at that point it can't happen, is there a guarantee I'm missing?

sueloverso:
@keithbostic asks a lot of good fundamental questions that I am interested in the answers to also. Those answers could radically change the code changes here, or make this change unneeded. So I will hold off my review until then.

test/suite/test_gc01.py
```diff
@@ -35,12 +35,9 @@ def timestamp_str(t):
...
 # test_gc01.py
-# Test that checkpoint cleans the obsolete lookaside pages.
-class test_gc01(wttest.WiredTigerTestCase):
-    # Force a small cache.
-    conn_config = 'cache_size=50MB,log=(enabled)'
-    session_config = 'isolation=snapshot'
 
+# Shared base class used by gc tests.
+class test_gc_base(wttest.WiredTigerTestCase):
     def large_updates(self, uri, value, ds, nrows, commit_ts):
```
ddanderson: I like that this test code is getting refactored.  Does any of the refactoring have to do with this ticket?  If not, it might be nice to put it in a separate ticket/commit.

ddanderson: On a related note, I wonder if any of these test changes expose a problem with eviction that is motivating this ticket?

src/btree/bt_sync.c
kommiharibabu: Yes, you are correct. I don't think this check is required. Anyway, when the leaf page is clean, it is handled in the different code path. I will remove this code.

src/evict/evict_page.c
kommiharibabu: I thought `__wt_yield()` can give a chance to the other threads to process so that it can improve the chance of evicting the page.

With `__wt_yield()` the eviction is able to evict the page after 546 times retry.
Without `__wt_yield()` the eviction is able to evict the page after 836 times retry.

kommiharibabu: Yes, I will fix it to use`EBUSY` cases. During the test of `test_gc03` after adding obsolete pages to the eviction, the pages didn't get evicted from either the hazard pointer availability in `__evict_exclusive()` and `__split_internal_lock()` function.

I don't find a way to know whether the page is from the urgent queue or not? And also I thought the failure from LRU pages is rare (maybe I am wrong).

kommiharibabu: The eviction server chooses the trees randomly to evict the pages under them.  Due to randomness, eviction may not evict the obsolete history pages and evict the pages that are may actually require. This may lead to cache use by pages that are not required.

If I am the only one concern about this scenario, I am fine to ignore this problem.
Another approach of how about choosing `LAS` as the first dhandle by the eviction walk when considering the pages to be evicted?

@agorrod, @bvpvamsikrishna Any suggestions?

test/suite/test_gc01.py
kommiharibabu: I got this problem when I was testing the `test_gc03.py` test. As the testing ticket has a dependency on other tickets to let the `search` work properly. I thought of pushing the test code reconsiderations here instead of waiting. I am fine to move them to test ticket.

agorrod:
@sueloverso and @keithbostic should we have a call to discuss the merits of this change? Or is there some information we could provide that would help understand whether this change is worthwhile?

keithbostic:
@agorrod, I don't object to the change, but I don't see any evidence it's worth doing.

I think the change is more difficult than it appears on the surface (it should be moved into eviction, at the least), I'd wait until there's evidence we need it.

Happy to go with your decision without further discussion.

kommiharibabu:
closing this pull request as the approach to handle the obsolete in-memory page is handled in another approach.

PR #5111, opened 2020-01-06
WT-5003 Migrate wiredtiger-test-race-condition-stress-sanitizer from Jenkins to Evergreen

tammybailey:
I updated the `format.sh` script to add a command line argument to run the timing stress split testing.

* The Jenkins test is [here](http://build.wiredtiger.com:8080/job/wiredtiger-test-race-condition-stress-sanitizer/)
* A successful patch build run over one hour is [here](https://evergreen.mongodb.com/version/5e12cb15a4cf47798eb0f30a)
* The 24-hour patch build is [here](https://evergreen.mongodb.com/build/wiredtiger_ubuntu1804_patch_d6ab08870356f990972e588c7be39fd9f396df8c_5e12ca6d0305b91a4325fe41_20_01_06_05_53_17)

### DIFF
--- a/test/evergreen.yml
+++ b/test/evergreen.yml
@@ -1954,6 +1954,21 @@ tasks:
           format_test_script_args: -S
           times: 16
 
+  - name: race-condition-stress-sanitizer-test
+    # set an evergreen timeout of 25 hours
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC="/opt/mongodbtoolchain/v3/bin/clang -fsanitize=address" PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb -fPIC -fno-omit-frame-pointer"
+          posix_configure_flags: --enable-diagnostic --with-builtins=lz4,snappy,zlib
+      - func: "format test script"
+        vars:
+          # set a format.sh timeout of 24 hours
+          format_test_script_args: -R -t 1440
+          test_env_vars: ASAN_OPTIONS="detect_leaks=1:abort_on_error=1:disable_coredump=0" ASAN_SYMBOLIZER_PATH=/opt/mongodbtoolchain/v3/bin/llvm-symbolizer
+
   - name: recovery-stress-test
     #set a 25 hours timeout
     exec_timeout_secs: 90000
@@ -2102,6 +2117,7 @@ buildvariants:
     - name: split-stress-test
     - name: format-stress-test
     - name: format-stress-smoke-test
+    - name: race-condition-stress-sanitizer-test
     - name: checkpoint-stress-test
 
 - name: ubuntu1804-compilers
--- a/test/format/format.sh
+++ b/test/format/format.sh
@@ -372,6 +372,7 @@ format()
 		for k in {1..7}; do
 			args+=" timing_stress_split_$k=$(($RANDOM%2))"
 		done
+		echo "$name: starting timing-stress-split job in $dir ($(date))"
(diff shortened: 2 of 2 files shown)


test/evergreen.yml
```diff
@@ -1954,6 +1954,21 @@ tasks:
...
           times: 16
 
+  - name: race-condition-stress-sanitizer-test
+    # set an evergreen timeout of 25 hours
+    exec_timeout_secs: 90000
+    commands:
+      - func: "get project"
+      - func: "compile wiredtiger"
+        vars:
+          configure_env_vars: CC="/opt/mongodbtoolchain/v3/bin/clang -fsanitize=address" PATH=/opt/mongodbtoolchain/v3/bin:$PATH CFLAGS="-ggdb -fPIC -fno-omit-frame-pointer"
```
raviprakashgiri29: I think "-fsanitize=address" option should be part of CFLAGS. Just curious to know the reason for not setting CXX variable.

test/format/format.sh
```diff
@@ -362,6 +367,11 @@ format()
 		args=${smoke_list[$smoke_next]}
 		smoke_next=$(($smoke_next + 1))
 		echo "$name: starting smoke-test job in $dir ($(date))"
+	elif [[ $timing_stress_split_test -ne 0 ]]; then
+		args=$format_args
+		for k in {1..7}; do
+			args+=" timing_stress_split_$k=$(($RANDOM%2))"
+		done
```
lukech: It would be helpful to add an echo line indicating the starting of the job, just as what `if` and `else` branches do. 
Also, the Jenkins job prints out the timeing_stress_split_x values being set for the format run, which would be nice to retain.

keithbostic: There are currently 8 timing stress tests -- no clue why they weren't in the Jenkins job? That may be worth figuring out.

lukech:
@keithbostic all the comments in the Jenkins job configuration still valid?
>     # Don't choose the checkpoint_timing stress option more than
>     # test/format normally does - it serves a different purpose,
>     # and enabling would make this job get through generally less ops.
Setting a random value provides 50% chance (compared to default 2%) a given timeing_stress_split_x being set to 1. Is it the intention of this Jenkins job in the first place?

keithbostic:
@lukech, @tammybailey: Luke's point seems valid to me, question answered. Might be worth migrating the comment to `format.sh`.

test/format/format.sh
tammybailey: @keithbostic it looks like 1-7 are used in `src/btree/bt_split.c` and 8 is used in `src/btree/bt_walk.c`. I found this in JIRA, might be related? [WT-4171: Enabling tree walk timing stress causes excessive slowdown](https://jira.mongodb.org/browse/WT-4171)

tammybailey:
@keithbostic regarding the comment Luke mentioned above, I did not include it in the migration as it appears to refer to a different config parameter, `timing_stress_checkpoint`, that isn't being set or modified in this test. I think Luke and I are both confused as to the intent of the comment - I assumed it was an unintentional leftover of a previous version of this test.

test/evergreen.yml
tammybailey: `CXX` was not set in Jenkins and isn't relevant to the test so I opted to keep the task definition as clean as possible.

test/format/format.sh
tammybailey: Yes, good point, I'll add the echo line. I'm not sure how useful it is to print the `timing_stress_split_` values for every run, particularly without the rest of the config or any other information related to the test run. If the test fails, the artifacts will have the full config plus all other relevant information. But please let me know if there's something I'm missing!

PR #5113, opened 2020-01-06
WT-5391 merge develop

luke-pearson:
So I branched wt-5931-merge-develop off develop, then merged pm-1521-durable-history into it. As such this branch should merge cleanly into pm-1521-durable-history.


src/reconcile/rec_visibility.c
```diff
@@ -223,6 +223,13 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
             r->update_prepared = list_uncommitted = true;
             if (upd->start_ts > max_ts)
                 max_ts = upd->start_ts;
+
+            /*
+             * Track the oldest update not on the page, used to decide whether reads can use the
+             * page image, hence using the start rather than the durable timestamp.
+             */
+            if (upd->start_ts < r->min_skipped_ts)
```
luke-pearson: @keithbostic, now that we select the newest version in __wt_rec_upd_select is this code still valid?

keithbostic: @luke-pearson, all I did in that code was change it so we maintained `r->min_skipped_ts` in one code path where we weren't doing that. If the other instance of `r->min_skipped_ts` has been removed, then I would guess this one can be removed as well.

luke-pearson:
Thanks for reviewing everyone, I've merged the branch manually locally and pushed to pm-1521-durable-history. As such I'm closing this PR.

PR #5114, opened 2020-01-07
WT-5345 Add dump timestamp to wt verify config options

jeremytay90:


jeremytay90:
retest this please

lukech:
retest this please

luke-pearson:
Could you comment on the jira with an example of the old output and the new output on the same table?

jeremytay90:
> Could you comment on the jira with an example of the old output and the new output on the same table?

Will do. There's no real difference with old output per-se but rather just adds output.

dist/api_data.py
```diff
@@ -1319,11 +1319,15 @@ def __ge__(self, other):
         Display the contents of in-memory pages as they are verified,
         using the application's message handler, intended for debugging''',
         type='boolean'),
+    Config('dump_timestamps', 'false', r'''
```
luke-pearson: Would it be better to call it dump_timepairs?

sulabhM: I will go with whatever we call (timestamp + txn-id) in durable history branch.

luke-pearson: Yeah timepair is better.

dist/api_data.py
```diff
@@ -1319,11 +1319,15 @@ def __ge__(self, other):
...
+    Config('dump_time_pairs', 'false', r'''
+        Display aggregated time pair ranges as pages are verified,
+        using the application's message handler, intended for debugging''',
+        type='boolean'),
     Config('strict', 'false', r'''
         Treat any verification problem as an error; by default, verify will
         warn, but not fail, in the case of errors that won't affect future
         behavior (for example, a leaked block)''',
-        type='boolean')
+        type='boolean'),
```
luke-pearson: Remove added comma

luke-pearson: Never mind, this file generally calls for a comma here.

src/btree/bt_vrfy.c
```diff
@@ -360,16 +367,25 @@ __verify_tree(WT_SESSION_IMPL *session, WT_REF *ref, WT_CELL_UNPACK *addr_unpack
...
-          __wt_page_type_string(page->type)));
+    /* Optionally dump the address/timestamp information. */
+    if (vs->dump_address || vs->dump_time_pairs) {
+        WT_RET(__wt_msg(session, "%s", __wt_page_type_string(page->type)));
+        if (vs->dump_address)
+            WT_RET(__wt_msg(session, "\t%s", __wt_page_addr_string(session, ref, vs->tmp1)));
+        if (vs->dump_time_pairs) {
+            addr = ref->addr;
+            if (addr != NULL)
+                WT_RET(__wt_msg(session, "\t<%lx:%lx, %lx:%lx>", addr->oldest_start_ts,
```
luke-pearson: @sulabhM you may be interested in this. We're printing timestamps as Hex which I think makes more sense.

agorrod: @luke-pearson This isn't aimed at me - I think it's important that we print timestamps in the same format that they are printed from MongoDB. Let me know if you need help finding an example of how that is done.

src/btree/bt_vrfy.c
```diff
@@ -360,16 +367,25 @@ __verify_tree(WT_SESSION_IMPL *session, WT_REF *ref, WT_CELL_UNPACK *addr_unpack
 
     bm = S2BT(session)->bm;
     page = ref->page;
+    addr = NULL;
```
sulabhM: variable declarations and assignments should be in the same order. So ```addr``` before ```bm```

src/btree/bt_vrfy.c
sulabhM: I second @agorrod , it will be best to use ```__wt_timestamp_to_string()``` which prints in MongoDB ways.

jeremytay90: In keeping with the current debug printing format in `__debug_cell`, I've changed it to ts/txn,ts/txn

jeremytay90:
retest this please

PR #5117, opened 2020-01-08
WT-5277 Cursor key out-of-order detected in the lookaside file (#5018) [BACKPORT-v4.0]

jeremytay90:
* If the WT_SESSION.update leaf-page search returns either the first or last slot of the page,
discard the search results and search the entire tree, an inserted item might belong on an
entirely different page.

* If a search-near or update based leaf-page search returns an exact match, then we
should be able to use it, no matter what page slots are involved.

* Fix two cases where successful pinned page searches can result in attempting to modify
an object before starting the necessary implicit transaction: WT_SESSION.truncate and
WT_SESSION.update (with the WT_CURSTD_UPDATE_LOCAL flag set).

We haven't seen the failure in truncate before because a "read" of any page normally
results in starting the transaction, it's only pinned pages that avoid that operation,
and it's relatively unlikely both the start/stop cursors involved in a range truncate
operation are the result of successful pinned page searches.

We haven't seen the failure in update because it's currently only possible when updating
the lookaside table, which uses explicit transactions.

* clang format & typo in error handling.

* Rework the comments to match the code.

* Fix a comment.

(cherry picked from commit 47fdb87459c726873d6af326e38771504f8a91d8)

### DIFF
--- a/src/btree/bt_cursor.c
+++ b/src/btree/bt_cursor.c
@@ -630,35 +630,40 @@ __wt_btcur_search_near(WT_CURSOR_BTREE *cbt, int *exactp)
     __cursor_state_save(cursor, &state);
 
     /*
-     * If we have a row-store page pinned, search it; if we don't have a
-     * page pinned, or the search of the pinned page doesn't find an exact
-     * match, search from the root. Unlike WT_CURSOR.search, ignore pinned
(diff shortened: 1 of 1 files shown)


luke-pearson:
LGTM.

PR #5121, opened 2020-01-10
WT-5317 Fix eviction writing uncommitted metadata updates to disk

quchenhao:


src/reconcile/rec_visibility.c
```diff
@@ -212,9 +212,15 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
          * point can move forward during reconciliation so we use a cached copy to avoid races when
          * a concurrent transaction commits or rolls back while we are examining its updates. As
          * prepared transaction IDs are globally visible, need to check the update state as well.
+         *
+         * We avoid the checkpoint transaction pinning the oldest txn id. Therefore, the
+         * r->las_running can move beyond checkpoint transaction id. Need to do a proper visiblity
```
luke-pearson: Could be worded better as: The checkpoint transaction doesn't pin the oldest txn id, therefore the r->last_running can move beyond the checkpoint transaction id.

Also last_running not las_running.

src/reconcile/rec_visibility.c
```diff
@@ -212,9 +212,15 @@ __wt_rec_upd_select(WT_SESSION_IMPL *session, WT_RECONCILE *r, WT_INSERT *ins, v
...
+         * check for metadata pages. Otherwise, eviction may select uncommitted metadata updates to
+         * write to disk
```
luke-pearson: Fullstop at tend of comment.

src/reconcile/rec_visibility.c
quchenhao: Thanks, will reword that.

PR #5122, opened 2020-01-10
WT-5385 Use WT_UPDATE_HISTORY_STORE to avoid repeatedly inserting updates to lookaside

luke-pearson:


src/cache/cache_las.c
```diff
@@ -754,22 +754,23 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
+                        modify_value = NULL;
+                    } else
+                        WT_ERR(__las_insert_record(session, cursor, btree_id, key, upd,
+                          WT_UPDATE_STANDARD, full_value, stop_ts_pair));
+
+                    /* Flag the update as now in the lookaside file. */
+                    F_SET(upd, WT_UPDATE_HISTORY_STORE);
+                    ++insert_cnt;
+                }
                 if (squashed) {
```
luke-pearson: @quchenhao do you think the (squashed) check should be inside the other check? Do you also think this change is roughly what you'd expect?

src/cache/cache_las.c
```diff
@@ -754,22 +754,23 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
                  * updates in the middle.
                  */
                 nentries = MAX_REVERSE_MODIFY_NUM;
```
quchenhao: Why don't we skip the updates flagged when we push them into stack? We can push the first flagged updates on stack and break.

quchenhao: Then when we pop the first update on stack, we can check if it is flagged. If it is flagged, we get the full value from lookaside and skip inserting it to lookaside. The rest of the code should remain the same. This is how I think it should work.

quchenhao: Your solution requires every change in lookaside to be flagged and we still have to apply forward deltas to construct the full values of each update.

src/cache/cache_las.c
```diff
@@ -754,22 +791,23 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
-                    modify_value = NULL;
-                } else
-                    WT_ERR(__las_insert_record(session, cursor, btree_id, key, upd,
-                      WT_UPDATE_STANDARD, full_value, stop_ts_pair));
-
-                /* Flag the update as now in the lookaside file. */
-                F_SET(upd, WT_UPDATE_HISTORY_STORE);
-                ++insert_cnt;
-
+                if (!F_ISSET(upd, WT_UPDATE_HISTORY_STORE)) {
```
luke-pearson: This check is still needed otherwise you insert the first update. aka the full value update.

src/cache/cache_las.c
```diff
@@ -709,25 +723,48 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
         prev_upd = NULL;
```
quchenhao: Why don't we retrieve the full value from las here?

test/suite/test_las08.py
```diff
@@ -0,0 +1,131 @@
+
```
quchenhao: Why not write the test in las06?

quchenhao: You can keep it as a separate test. I'm not going to argue against that. But I feel this should be in las06 as they all test the same code.

src/cache/cache_las.c
```diff
@@ -549,23 +549,26 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
     WT_TIME_PAIR stop_ts_pair;
+    wt_off_t las_size;
+    wt_timestamp_t durable_timestamp;
     uint64_t insert_cnt, max_las_size;
     uint32_t mementos_cnt, btree_id, i;
-    uint8_t *p;
+    uint8_t *p, prepare_state, upd_type;
     int nentries;
-    bool las_key_saved, local_txn, squashed;
+    bool las_key_saved, local_txn, retrieve_modify, squashed;
```
quchenhao: retrieve_modify may be redundant if we move the place we retrieve value from las

test/suite/test_las08.py
luke-pearson: Honestly I prefer tests to be broken and out and test as few things in one test as possible rather than one big file of many tests. I'm happy to keep it like this if you are.

src/cache/cache_las.c
luke-pearson: We don't have the previous updates value to get the stop_time_pair.

quchenhao: OK, can we peek the value here? We can add a peek function in modify_vector and get the stop timestamp. I think the code will be cleaner if we can search las here.

test/suite/test_las08.py
quchenhao: That is fine for me.

src/cache/cache_las.c
luke-pearson: Agreed will add peek.

src/cache/cache_las.c
```diff
@@ -572,8 +578,32 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
+             * iterate the full update list associated with the modify and recalculating the reverse
+             * deltas. Here we need to set the read timestamp of the transaction to be the start
+             * timestamp of the update, otherwise when we search we will see the tombstone value
+             * associated with the update and return not found.
+             */
+            txn->read_timestamp = upd->start_ts;
+            F_SET(txn, WT_TXN_HAS_TS_READ);
+            cursor->set_key(cursor, btree_id, key, upd->start_ts, upd->txnid, prev_upd->start_ts,
+              prev_upd->txnid);
+            WT_ERR(cursor->search(cursor));
```
quchenhao: Should we reset read_timestamp if error?

src/cache/cache_las.c
```diff
@@ -585,25 +615,24 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
 
-        prev_upd = NULL;
         squashed = false;
 
-        /* Flush the updates on stack */
+        /* Flush the updates on stack. */
         for (; modifies.size > 0; tmp = full_value, full_value = prev_full_value,
                                   prev_full_value = tmp, upd = prev_upd) {
-            /* Should not see BIRTHMARK or TOMBSTONE */
+            /* Should not see BIRTHMARK or TOMBSTONE. */
```
agorrod: You could just remove this comment - it's just restating what the code clearly says. Similarly to `Skip TOMBSTONE` below.

src/cache/cache_las.c
```diff
@@ -680,7 +710,11 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
           (uint64_t)las_size, max_las_size);
 
 err:
-    /* Resolve the transaction. */
+    /* Resolve the transaction. Clear the read timestamp if we got an error after setting it. */
+    if (F_ISSET(txn, WT_TXN_HAS_TS_READ)) {
+        F_CLR(txn, WT_TXN_HAS_TS_READ);
+        txn->read_timestamp = 0;
```
agorrod: Is there a special value for WT_TS_NONE?

src/cache/cache_las.c
```diff
@@ -572,8 +578,32 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
+            WT_ASSERT(session, F_ISSET(upd, WT_UPDATE_HISTORY_STORE));
+            __wt_modify_vector_peek(&modifies, &prev_upd);
+            /*
+             * Retrieve the full value of the modify from the history store. This avoid us having to
+             * iterate the full update list associated with the modify and recalculating the reverse
+             * deltas. Here we need to set the read timestamp of the transaction to be the start
+             * timestamp of the update, otherwise when we search we will see the tombstone value
+             * associated with the update and return not found.
+             */
+            txn->read_timestamp = upd->start_ts;
```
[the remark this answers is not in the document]
luke-pearson: I moved it into a function, let me know if thats what you expected. I'm not a huge fan of the function handling the txn->read_timestamp semantics itself however I don't see a nicer way of doing it either. Without making it kinda pointless creating its own function.

src/cache/cache_las.c
```diff
@@ -572,8 +608,23 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
          */
         WT_ASSERT(session, modifies.size > 0);
         __wt_modify_vector_pop(&modifies, &upd);
-        /* The key didn't exist back then, which is globally visible. */
-        WT_ASSERT(session, upd->type == WT_UPDATE_STANDARD || upd->type == WT_UPDATE_TOMBSTONE);
+
+        /* If we popped a modify then it should be flagged as in the history store. */
+        if (upd->type == WT_UPDATE_MODIFY) {
+            WT_ASSERT(session, F_ISSET(upd, WT_UPDATE_HISTORY_STORE));
```
tetsuo-cpp: If you update restore, apply modifies on top and then reconcile again, won't this happen?

src/cache/cache_las.c
```diff
@@ -572,8 +608,23 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
...
+            __wt_modify_vector_peek(&modifies, &prev_upd);
+            /*
+             * Retrieve the full value of the modify from the history store. This avoid us having to
+             * iterate the full update list associated with the modify and recalculating the reverse
+             * deltas.
+             */
+            WT_ERR(__las_get_value(
```
tetsuo-cpp: I just had a thought since this made me wonder. If this modify is on the same lookaside page that we're writing updates to then it makes sense, but if we had to read a page into cache to do this, isn't it likely that this is actually worse?

src/cache/cache_las.c
quchenhao: Currently it happens. But for eviction, we can free everything inserted into lookaside right after commit. Thus, in the future, it won't happen.

src/cache/cache_las.c
quchenhao: It's the same key so likely we are on the same page here.

tetsuo-cpp: Yeah, I realize it's probably going to be the same page but it's not guaranteed. And in the case where it isn't, it's going to be very costly no?

Hard to say how often each scenario occurs vs the cost so no need to address it unless there are issues later.

src/cache/cache_las.c
tetsuo-cpp: Yeah but doesn't it happen now as of the current state of the branch?

quchenhao: No, in current state of the branch, the first update on stack is either a standard update or a tombstone.

src/cache/cache_las.c
```diff
@@ -562,6 +596,8 @@ __wt_las_insert_updates(WT_CURSOR *cursor, WT_BTREE *btree, WT_PAGE *page, WT_MU
                 continue;
```
quchenhao: @luke-pearson We also need to update the comments above.

PR #5123, opened 2020-01-10
WT-4636 Fix strace test (#4537)

ddanderson:
Added O_CLOEXEC flag to the template that the test matches, to correspond with changes to WT-4602.  Also changed the strace test to work with glibc versions > 2.26.  These versions implement the open syscall in terms of openat, so the strace test template needs a conditional version of open.

(cherry picked from commit 2971a48)


ddanderson:
@quchenhao, not sure why, but cherry pick did the wrong thing - that is, it gave a merge conflict that I resolved in a sensible way, but then there were about 5 other lines that it got wrong in `base.run`.  This is probably what you experienced too.  Anyway, the diffs look right and the test passes.  Please review.

PR #5124, opened 2020-01-10
WT-4636 Fix strace test (#4537)

ddanderson:
Added O_CLOEXEC flag to the template that the test matches, to correspond with changes to WT-4602.  Also changed the strace test to work with glibc versions > 2.26.  These versions implement the open syscall in terms of openat, so the strace test template needs a conditional version of open.

(cherry picked from commit 2971a485639120b9580707b56eba309b8b56355c)


ddanderson:
@quchenhao, here's the matching 3.6 backport.  Can you review please?

